PR #12847, opened 2023-11-28
Report the time it took for building the FST in Test2BFST

dungba88:
### Description

- Report the time it took for building the FST
- Report the FST actual size, as it can differ from the RAM bytes used once the test is moved to off-heap

```
100000: 39308104 RAM bytes used; 39256528 FST bytes; 19189166 nodes; took 25 seconds
200000: 78583336 RAM bytes used; 78520655 FST bytes; 38378178 nodes; took 49 seconds
300000: 117858568 RAM bytes used; 117784024 FST bytes; 57567011 nodes; took 75 seconds
400000: 157133800 RAM bytes used; 157048270 FST bytes; 76756271 nodes; took 101 seconds
```

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -83,6 +83,8 @@ Improvements
 
 * GITHUB#12447: Hunspell: speed up the dictionary enumeration on suggestion (Peter Gromov)
 
+* GITHUB#12847: Test2BFST now reports the time it took to build the FST and the real FST size (Anh Dung Bui)
+
 Optimizations
 ---------------------
 
@@ -171,6 +173,9 @@ API Changes
 * GITHUB#12799: Make TaskExecutor constructor public and use TaskExecutor for concurrent
   HNSW graph build. (Shubham Chaudhary)
 
+* GITHUB#12758, GITHUB#12803: Remove FST constructor with DataInput for metadata. Please
+  use the constructor with FSTMetadata instead. (Anh Dung Bui)
+
 New Features
 ---------------------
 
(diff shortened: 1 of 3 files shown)


lucene/core/src/java/org/apache/lucene/util/fst/FSTCompiler.java
```diff
@@ -867,6 +867,10 @@ public long fstRamBytesUsed() {
     return fst.ramBytesUsed();
   }
 
+  public long fstSize() {
```
mikemccand, line 870: Hmm should we just re-use the existing `fstRamBytesUsed`?  Confusing to have two such similar methods?

dungba88: I think they will be similar for on heap use case but will be very different for off heap? In such case the ram bytes used should stay flat while the size should keep increasing.

mikemccand: Ahh OK that indeed makes sense!

Could we rename to `fstSizeInBytes`?

dungba88: Good idea. I have changed the method name

dungba88:
The test failed with just `Error: The operation was canceled.` but I can't tell why it happened. The same PR in my local branch works: https://github.com/dungba88/lucene/pull/20

mikemccand:
> The test failed with just `Error: The operation was canceled.` but I can't tell why it happened. The same PR in my local branch works: [dungba88#20](https://github.com/dungba88/lucene/pull/20)

Weird.  I restarted the failed job.

mikemccand (approved):
Thanks @dungba88.

lucene/CHANGES.txt
```diff
@@ -176,7 +178,8 @@ API Changes
 * GITHUB#12799: Make TaskExecutor constructor public and use TaskExecutor for concurrent
   HNSW graph build. (Shubham Chaudhary)
 
-*
+* GITHUB#12758, GITHUB#12803: Remove FST constructor with DataInput for metadata. Please
```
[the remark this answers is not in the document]
dungba88: Thank you!

PR #12848, opened 2023-11-28
Fix *HnswVectorsFormat.testIndexedValueNotAliased test flakiness

benwtrent:
periodic and random merge policies can cause the docs iterated to be in a different order (as they are merged). 

This commit reduces the randomness of the merge policy for more consistent vector value iteration.

### DIFF
--- a/lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseKnnVectorsFormatTestCase.java
+++ b/lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseKnnVectorsFormatTestCase.java
@@ -42,6 +42,7 @@ import org.apache.lucene.index.IndexWriter;
 import org.apache.lucene.index.IndexWriterConfig;
 import org.apache.lucene.index.LeafReader;
 import org.apache.lucene.index.LeafReaderContext;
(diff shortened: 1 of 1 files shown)


lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseKnnVectorsFormatTestCase.java
```diff
@@ -732,7 +733,13 @@ public void testIndexedValueNotAliased() throws Exception {
...
     float[] v = {0};
     try (Directory dir = newDirectory();
-        IndexWriter iw = new IndexWriter(dir, newIndexWriterConfig())) {
+        IndexWriter iw =
+            new IndexWriter(
+                dir,
+                newIndexWriterConfig()
+                    .setMergePolicy(NoMergePolicy.INSTANCE)
+                    .setMaxBufferedDocs(3)
+                    .setRAMBufferSizeMB(-1))) {
```
ChrisHegarty, line 742: Trivially, `IndexWriterConfig.DISABLE_AUTO_FLUSH` could be used here rather than the literal.

ChrisHegarty (approved):
LGTM

PR #12849, opened 2023-11-28
Fix bug in UnescapedCharSequence and add basic unit tests

### LINKED ISSUE #9049: UnescapedCharSequence Bugs [LUCENE-8001]
There are a couple of issues with UnescapedCharSequence:

1. The [private constructor](https://github.com/apache/lucene-solr/blob/32ed8520c706a865b94e03644ae4e4435e0f7d35/lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/core/util/UnescapedCharSequence.java#L52-L63) is not used anywhere (and if it were, it would throw exceptions)
2. The ToEscapedString() overload has an [invalid condition](https://github.com/apache/lucene-solr/blob/32ed8520c706a865b94e03644ae4e4435e0f7d35/lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/core/util/UnescapedCharSequence.java#L96) that will only evaluate to true if the string has a length of 0.

There are no tests for UnescapedCharSequence so these issues have gone unnoticed for quite some time.

---
Migrated from [LUCENE-8001](https://issues.apache.org/jira/browse/LUCENE-8001) by Shad Storhaug, updated Oct 21 2017
Linked issues:
 - [LUCENENET-592](https://issues.apache.org/jira/browse/LUCENENET-592)

slow-J:
Fixing a basic bug in UnescapedCharSequence https://github.com/apache/lucene/blob/2bb69f3246218dd8176cf92d8064623688c5272c/lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/core/util/UnescapedCharSequence.java#L86
and adding some basic unit tests.
Also removing an unused private constructor.

This is not very urgent since the issue is over 6 years old, but we may as well clean this code up.

Closes #9049.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -138,6 +138,8 @@ Other
 
 * GITHUB#12239: Hunspell: reduced suggestion set dependency on the hash table order (Peter Gromov)
 
+* GITHUB#9049: Removing unused private method in UnescapedCharSequence. (Jakub Slowinski)
+
 ======================== Lucene 9.9.0 =======================
 
 API Changes
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -138,7 +138,7 @@ Other
 
 * GITHUB#12239: Hunspell: reduced suggestion set dependency on the hash table order (Peter Gromov)
 
-* GITHUB#9049: Removing unused private method in UnescapedCharSequence. (Jakub Slowinski)
+* GITHUB#9049: Removing unused private constructor in UnescapedCharSequence. (Jakub Slowinski)
 
 ======================== Lucene 9.9.0 =======================
 
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -138,7 +138,7 @@ Other
 
 * GITHUB#12239: Hunspell: reduced suggestion set dependency on the hash table order (Peter Gromov)
 
-* GITHUB#9049: Removing unused private constructor in UnescapedCharSequence. (Jakub Slowinski)
+* GITHUB#9049: Fixing bug in UnescapedCharSequence#toStringEscaped() (Jakub Slowinski)
 
 ======================== Lucene 9.9.0 =======================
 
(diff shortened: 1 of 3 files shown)


lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/core/util/UnescapedCharSequence.java
```diff
@@ -101,7 +90,7 @@ public String toStringEscaped() {
    */
   public String toStringEscaped(char[] enabledChars) {
     // TODO: non efficient implementation, refactor this code
```
slow-J, line 92: In the Lucene codebase this is only ever called with array of length 2, so I didn't refactor......

lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/core/util/UnescapedCharSequence.java
```diff
@@ -82,11 +71,11 @@ public String toString() {
...
-      if (this.chars[i] == '\\') {
+    StringBuilder result = new StringBuilder(this.length());
+    for (int i = 0; i < this.length(); i++) {
+      if (this.chars[i] == '\\' || this.wasEscaped[i]) {
         result.append('\\');
-      } else if (this.wasEscaped[i]) result.append('\\');
+      }
 
       result.append(this.chars[i]);
     }
```
shubhamvishu, line 81: I didn't find any usage of this function as well. Seems its unused too?

shubhamvishu (approved):
Looks good to me @slow-J

lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/core/util/UnescapedCharSequence.java
slow-J: Thanks for the review @shubhamvishu. I only removed the other method because it was `private` and `@SuppressWarnings("unused")`, while this is a potentially useful public method.

lucene/CHANGES.txt
```diff
@@ -138,6 +138,8 @@ Other
 
 * GITHUB#12239: Hunspell: reduced suggestion set dependency on the hash table order (Peter Gromov)
 
+* GITHUB#9049: Removing unused private constructor in UnescapedCharSequence. (Jakub Slowinski)
```
gsmiller, line 141: Let's have the changes entry highlight the bug fix and not the ctor removal? I don't think we need to mention the removal of an unused, private ctor in CHANGES personally, but users might care about the bug fix :)

slow-J: Good point @gsmiller! Modified the changes.txt to mention the bug fix instead.

gsmiller (approved):
LGTM! Thanks @slow-J !

slow-J:
Thanks for the review @gsmiller and @shubhamvishu !

PR #12850, opened 2023-11-29
Fix intermittently failing TestSortedSetFieldSource

ChrisHegarty:
This commit fixes the intermittently failing TestSortedSetFieldSource.

The test assertions depend on doc order which may be affected by merging. The fix is to trivially avoid merging for the very small index, with just two docs.

### DIFF
--- a/lucene/queries/src/test/org/apache/lucene/queries/function/TestSortedSetFieldSource.java
+++ b/lucene/queries/src/test/org/apache/lucene/queries/function/TestSortedSetFieldSource.java
@@ -17,7 +17,6 @@
 package org.apache.lucene.queries.function;
 
 import java.util.Collections;
-
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field;
 import org.apache.lucene.document.SortedSetDocValuesField;
--- a/lucene/queries/src/test/org/apache/lucene/queries/function/TestSortedSetFieldSource.java
+++ b/lucene/queries/src/test/org/apache/lucene/queries/function/TestSortedSetFieldSource.java
@@ -17,12 +17,15 @@
 package org.apache.lucene.queries.function;
 
 import java.util.Collections;
+
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field;
 import org.apache.lucene.document.SortedSetDocValuesField;
 import org.apache.lucene.index.DirectoryReader;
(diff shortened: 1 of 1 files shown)


ChrisHegarty:
The test previously has been seen to fail with:

```
org.junit.ComparisonFailure: expected:<ba[z]> but was:<ba[r]>
  at __randomizedtesting.SeedInfo.seed([45D4C56DF2092811:7D67E193D5FAFCC0]:0)
  at junit@4.13.1/org.junit.Assert.assertEquals(Assert.java:117)
  at junit@4.13.1/org.junit.Assert.assertEquals(Assert.java:146)
  at org.apache.lucene.queries.function.TestSortedSetFieldSource.testSimple(TestSortedSetFieldSource.java:59)
  ...
```

After the change the test has not failed in several hundreds of thousands of runs.

mikemccand:
> The test assertions depend on doc order which may be affected by merging.

Hmm, I don't fully understand the test, but did the sort order truly rely on `docid` tie-breaking?  If so then this is the right fix.

I just want to make sure we are not hiding a bug that e.g. causes this sorter to incorrectly sort / compare documents across segments that were not in fact ties by the sort criteria.

ChrisHegarty:
My understanding of this part of the test is that it asserts the sort order of the multi-valued string values in the `value` field. The changes here do not affect this assertion, the sort order is still checked.

There are just two docs in the index. The first doc has a single value for the `value` field, `baz`. The second doc has two values for the `value` field, `foo` and `bar`.  The test expects to find `baz` and `bar` (since `bar` is less than `foo`), for the `values` field in doc1 and doc2 respectively.  

 The test assertions are based on doc order in the index.  Does this address your concern? Or have I missed the point?

jpountz (approved):
This new failure looks like it's due to the fact that MockRandomMergePolicy now randomly reverses the doc ID order on merge? If so the fix looks correct to me.

ChrisHegarty:
> This new failure looks like it's due to the fact that MockRandomMergePolicy now randomly reverses the doc ID order on merge?

That's it exactly. I confirmed this through local instrumentation of MockRandomMergePolicy.

mikemccand:
Thanks for the detailed explanation @ChrisHegarty and @jpountz -- fix looks good!  Sneaky random merge policies...

PR #12853, opened 2023-11-29
Ensure #finish is called on all drill-sideways FacetCollectors even when no hits are scored

gsmiller:
In GH#12558, @Shradha26 discovered that faceting implementations are sometimes getting empty MatchingDoc lists, resulting in undesired behavior. We should consider it a bug for MatchingDocs to not be populated, even if there are no collected hits. After chasing the issue a bit, it appears that drill-sideways can miss calling `#finish` on the FacetsCollectors it manages if it end up not scoring any docs. This PR addresses it.

I made a couple related changes in this PR to get at fixing the bug:
1. Added `DrillSideways#createDrillSidewaysFacetsCollectorManager` as a protected method that users can extend to create specific facet collecting implementations. This seems reasonable to me as we already have a similar hook for the drill-down collector. I found this hook useful for testing. If we don't want to expose this API surface, I can adjust the testing approach, but I think it's reasonable to expose personally.
2. Changed the `DrillSidewaysScorer` to accept `LeafCollectors` instead of `FacetCollectors`. This seems like a better division of responsibilities anyway, and is simpler, but it also lets `DrillSidewaysQuery` manage the creation of the leaf collectors when setting up the `BulkScorer`. This is necessary if `DrillSidewaysQuery` is going to call `#finish` in no-scoring cases since `#finish` relies on a leaf scorer having been created already from the collector.
3. I also added a TODO to go remove `DrillSideways#createDrillDownFacetsCollector` since it is no longer used and is a trappy hook for users. I'll tackle this in a separate PR though since we need a deprecation back port and it's not really related to this (just something I came across).

### DIFF
--- a/lucene/facet/src/java/org/apache/lucene/facet/DrillSidewaysQuery.java
+++ b/lucene/facet/src/java/org/apache/lucene/facet/DrillSidewaysQuery.java
@@ -23,13 +23,13 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
-import org.apache.lucene.facet.DrillSidewaysScorer.DocsAndCost;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.search.BulkScorer;
 import org.apache.lucene.search.ConstantScoreScorer;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.Explanation;
 import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.LeafCollector;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.QueryVisitor;
 import org.apache.lucene.search.ScoreMode;
@@ -175,6 +175,17 @@ class DrillSidewaysQuery extends Query {
 
         int drillDownCount = drillDowns.length;
 
+        FacetsCollector drillDownCollector;
+        LeafCollector drillDownLeafCollector;
+        if (drillDownCollectorManager != null) {
+          drillDownCollector = drillDownCollectorManager.newCollector();
+          managedDrillDownCollectors.add(drillDownCollector);
+          drillDownLeafCollector = drillDownCollector.getLeafCollector(context);
+        } else {
+          drillDownCollector = null;
+          drillDownLeafCollector = null;
+        }
+
         FacetsCollector[] sidewaysCollectors = new FacetsCollector[drillDownCount];
         managedDrillSidewaysCollectors.add(sidewaysCollectors);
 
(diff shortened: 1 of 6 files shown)


lucene/facet/src/java/org/apache/lucene/facet/DrillSidewaysQuery.java
```diff
@@ -193,42 +204,29 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
...
-        }
-
-        FacetsCollector drillDownCollector;
-        if (drillDownCollectorManager != null) {
-          drillDownCollector = drillDownCollectorManager.newCollector();
-          managedDrillDownCollectors.add(drillDownCollector);
-        } else {
-          drillDownCollector = null;
-        }
+        Arrays.sort(dims, Comparator.comparingLong(o -> o.approximation.cost()));
```
gautamworah96, line 226: Thanks for these small cleanups as well. LGTM

lucene/facet/src/java/org/apache/lucene/facet/DrillSidewaysQuery.java
```diff
@@ -193,42 +204,29 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
...
-          dims[dim] = new DrillSidewaysScorer.DocsAndCost(scorer, sidewaysCollector);
+          dims[dim] =
+              new DrillSidewaysScorer.DocsAndCost(
+                  scorer, sidewaysCollector.getLeafCollector(context));
         }
 
-        // If more than one dim has no matches, then there
-        // are no hits nor drill-sideways counts.  Or, if we
-        // have only one dim and that dim has no matches,
-        // same thing.
```
gautamworah96, line 202: unlrelated but how exactly is `(nullCount > 1 || (nullCount == 1 && dims.length == 1))` equivalent to `nullCount > 1`?

gautamworah96 (approved):

lucene/facet/src/java/org/apache/lucene/facet/DrillSidewaysQuery.java
gsmiller: Yeah, they're not :). We _do_ need to still run the query in the case where there's a single dim but it's null because we need to collect all the "sideways" hits for that null dim. Looks like this comment (along with the commented out condition) came from @mikemccand back in 2014. I doubt he recalls much 9 years later. Maybe he realized this is a bug, commented out the condition (and replaced it with the correct one) and left the comment untouched? That's my best guess. But no need to keep the comment around anymore.

gsmiller:
Thanks @gautamworah96  for taking a look!

PR #12855, opened 2023-11-29
Remove DrillSideways#createDrillDownFacetsCollector in favor of the manager-based hook

gsmiller:
This extension hook is no longer used on the main branch, so let's remove it. Opened a corresponding PR to mark this deprecated on 9x (#12854)

More specifically, the `DrillSideways` class used to delegate to this method for creating the drill-down collector, but a while back was switched over to delegating to `createDrillDownFacetsCollectorManager` instead (doing "manger-based" collection instead). So users can still override the drill-down collection behavior, but they have to use `createDrillDownFacetsCollectorManager`, not `createDrillDownFacetsCollector`. Since nothing actually delegates to `createDrillDownFacetsCollector`, this is really trappy for users.


stefanvodita (approved):
I left a small comment in the related #12854, but this one makes sense.

Issue #12414, opened 2023-07-04
Wrap MatchedDocs and LiveDocs in a single Bits instance in createBitSet function of AbstractKnnVectorQuery


Pulkitg64:
### Description

**Context**

In KNN Prefiltering we require a ```BitSet``` of docs which matched with prefilter query. This BitSet is used during HNSW Graph Search to check whether a node is accepted for matching or not. To create this BitSet we call [createBitSet](https://github.com/apache/lucene/blob/9ffc625b2ec43481e2cfba29828efbddf12e6852/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L161) function which creates a BitSet with following conditions:

1. When there are **no deletions** in the index and ```DocIdSetIterator``` is of ```BitSetIterator``` instance, then directly return the BitSet.

2. For other cases create a ```FilteredDocIdSetIterator``` first then create ```BitSet``` by iterating over docs from the ```FilteredDocIdSetIterator```.

```
private BitSet createBitSet(DocIdSetIterator iterator, Bits liveDocs, int maxDoc)
      throws IOException {
    if (liveDocs == null && iterator instanceof BitSetIterator bitSetIterator) {
      // If we already have a BitSet and no deletions, reuse the BitSet
      return bitSetIterator.getBitSet();
    } else {
      // Create a new BitSet from matching and live docs
      FilteredDocIdSetIterator filterIterator =
          new FilteredDocIdSetIterator(iterator) {
            @Override
            protected boolean match(int doc) {
              return liveDocs == null || liveDocs.get(doc);
            }
          };
      return BitSet.of(filterIterator, maxDoc);
    }
  }

```
**Problem**

When the ```DocIdSetIterator``` is of ```BitSetIterator``` instance and there are **deletions** in the index, we create a FilteredDocIdSetIterator which is later used to create a BitSet. This is a time consuming process because to create the BitSet we iterate over all matching docs from the iterator.

**Proposed Optimisation**
 
For the above case, we can wrap matchedDocs and liveDocs under single Bits instance something like below which can be directly passed for HNSW search. So, now during HNSW Search when a node is explored, we will check if the doc is accepted or not by checking bits of both matchedDocs and liveDocs which is constant time operation.

```
public class KNNBitSet implements Bits{

    Bits liveDocs;
    Bits matchedDocs;
    public KNNBitSet(Bits matchedDocs, Bits liveDocs) {
        this.matchedDocs = matchedDocs;
[...]

PR #12857, opened 2023-11-30
Reuse BitSet when there are deleted documents in the index instead of creating new BitSet

Pulkitg64:
### Description

Fixes issue: #12414

Before this change we were creating new BitSet every time when there are deletions in the index with use of matched Docs and Live Docs, which required iteration over all matched docs which is a time consuming process with linear time complexity. This is not required when the iterator is of **BitSetIterator** instance.

With this change we have wrapped matching Docs and live Docs under single Bits instance which can be directly passed for HNSW search. So, now during HNSW Search when a node is explored, we will check if the doc is accepted or not by checking bits of both matchedDocs and liveDocs which is constant time operation. Cost (```int cost = acceptDocs.length()```) of the new acceptedDocs is not exactly accurate but gives an upper bound on number, as it is not considering live-docs count.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
@@ -118,13 +118,38 @@ abstract class AbstractKnnVectorQuery extends Query {
       return NO_RESULTS;
     }
 
-    BitSet acceptDocs = createBitSet(scorer.iterator(), liveDocs, maxDoc);
-    int cost = acceptDocs.cardinality();
+    DocIdSetIterator iterator = scorer.iterator();
+    BitSet bitSet =
+        iterator instanceof BitSetIterator
+            ? ((BitSetIterator) iterator).getBitSet()
+            : BitSet.of(iterator, maxDoc);
+    Bits acceptDocs =
+        new Bits() {
+          @Override
+          public boolean get(int index) {
+            return liveDocs != null ? liveDocs.get(index) & bitSet.get(index) : bitSet.get(index);
+          }
+
+          @Override
+          public int length() {
+            return bitSet.cardinality();
+          }
+        };
+
+    FilteredDocIdSetIterator filteredDocIdSetIterator =
+        new FilteredDocIdSetIterator(new BitSetIterator(bitSet, maxDoc)) {
+          @Override
+          protected boolean match(int doc) {
+            return liveDocs == null || liveDocs.get(doc);
+          }
+        };
+
+    int cost = acceptDocs.length();
 
     if (cost <= k) {
       // If there are <= k possible matches, short-circuit and perform exact search, since HNSW
       // must always visit at least k documents
-      return exactSearch(ctx, new BitSetIterator(acceptDocs, cost));
+      return exactSearch(ctx, filteredDocIdSetIterator);
     }
 
     // Perform the approximate kNN search
@@ -133,25 +158,7 @@ abstract class AbstractKnnVectorQuery extends Query {
       return results;
     } else {
       // We stopped the kNN search because it visited too many nodes, so fall back to exact search
-      return exactSearch(ctx, new BitSetIterator(acceptDocs, cost));
-    }
-  }
-
-  private BitSet createBitSet(DocIdSetIterator iterator, Bits liveDocs, int maxDoc)
-      throws IOException {
-    if (liveDocs == null && iterator instanceof BitSetIterator bitSetIterator) {
-      // If we already have a BitSet and no deletions, reuse the BitSet
-      return bitSetIterator.getBitSet();
-    } else {
-      // Create a new BitSet from matching and live docs
-      FilteredDocIdSetIterator filterIterator =
-          new FilteredDocIdSetIterator(iterator) {
-            @Override
-            protected boolean match(int doc) {
-              return liveDocs == null || liveDocs.get(doc);
-            }
-          };
-      return BitSet.of(filterIterator, maxDoc);
+      return exactSearch(ctx, filteredDocIdSetIterator);
     }
   }
 


lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -118,13 +118,38 @@ private TopDocs getLeafResults(LeafReaderContext ctx, Weight filterWeight) throw
...
+          @Override
+          public boolean get(int index) {
+            return liveDocs != null ? liveDocs.get(index) & bitSet.get(index) : bitSet.get(index);
+          }
+
+          @Override
+          public int length() {
+            return bitSet.cardinality();
+          }
+        };
```
shubhamvishu, line 137: Why are we creating this when we could  directly do `int cost = bitSet.cardinality();`

shubhamvishu (commented):
Thanks for working on this @Pulkitg64 . I went through the change but I didn't understand how are we not reusing the bitset in the current approach. We do wrap the  `BitSetIterator` with a `FilteredDocIdSetIterator` when there are deleted docs right which would eventually use the bitset to advance the inner iterator(See [this](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/FilteredDocIdSetIterator.java#L72-L87)). 

Instead in this PR we are unnecessarily always wrapping the `BitSetIterator` with a `FilteredDocIdSetIterator` even when there are no  deletions which is just an added overhead and is avoided in the current code. Additionally, I find the current approach(separate function) more cleaner too. WDYT?

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
Pulkitg64: Here we are implementing the ```Bits``` interface that's why we need to override both functions.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -118,13 +118,38 @@ private TopDocs getLeafResults(LeafReaderContext ctx, Weight filterWeight) throw
...
+        new Bits() {
+          @Override
+          public boolean get(int index) {
+            return liveDocs != null ? liveDocs.get(index) & bitSet.get(index) : bitSet.get(index);
+          }
+
+          @Override
+          public int length() {
+            return bitSet.cardinality();
+          }
```
kaivalnp, line 136: `length()` is more like the [maximum doc](https://github.com/apache/lucene/blob/0e96b9cd8c1267acee02684bcac5be89d60b3bf4/lucene/core/src/java/org/apache/lucene/util/Bits.java#L73) you can request, this should be `bitSet.length()`?

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -118,13 +118,38 @@ private TopDocs getLeafResults(LeafReaderContext ctx, Weight filterWeight) throw
...
+    DocIdSetIterator iterator = scorer.iterator();
+    BitSet bitSet =
+        iterator instanceof BitSetIterator
+            ? ((BitSetIterator) iterator).getBitSet()
+            : BitSet.of(iterator, maxDoc);
+    Bits acceptDocs =
+        new Bits() {
+          @Override
+          public boolean get(int index) {
+            return liveDocs != null ? liveDocs.get(index) & bitSet.get(index) : bitSet.get(index);
```
kaivalnp, line 130: nit: Can we make this cleaner by `return (liveDocs == null || liveDocs.get(index)) && bitSet.get(index)` ?

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -118,13 +118,38 @@ private TopDocs getLeafResults(LeafReaderContext ctx, Weight filterWeight) throw
...
 
-    BitSet acceptDocs = createBitSet(scorer.iterator(), liveDocs, maxDoc);
-    int cost = acceptDocs.cardinality();
+    DocIdSetIterator iterator = scorer.iterator();
+    BitSet bitSet =
+        iterator instanceof BitSetIterator
+            ? ((BitSetIterator) iterator).getBitSet()
+            : BitSet.of(iterator, maxDoc);
+    Bits acceptDocs =
+        new Bits() {
```
kaivalnp, line 127: Perhaps we can wrap the `BitSet` in a new `Bits` only for the case we're trying to optimize (when `iterator instanceof BitSetIterator`) -- not changing the `BitSet.of` flow like @shubhamvishu also mentioned?

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -118,13 +118,38 @@ private TopDocs getLeafResults(LeafReaderContext ctx, Weight filterWeight) throw
...
+
+    FilteredDocIdSetIterator filteredDocIdSetIterator =
+        new FilteredDocIdSetIterator(new BitSetIterator(bitSet, maxDoc)) {
+          @Override
+          protected boolean match(int doc) {
+            return liveDocs == null || liveDocs.get(doc);
+          }
+        };
+
+    int cost = acceptDocs.length();
```
kaivalnp, line 147: The `cost` here determines what limit to set for `#approximateSearch`

If we use `acceptDocs.length()`, this will be equal to `maxDoc` (and always complete graph search without falling back to exact search, even when we want to..)

Perhaps this should be `acceptDocs.cardinality()`?

kaivalnp (commented):
Interesting! So instead of greedily collecting all matching + live docs into a `BitSet`, we're saving on the filter collection step at the cost of running `#approximateSearch` with an upper bound of `visitLimit`

Can you run some benchmarks for different filters to measure this tradeoff?

benwtrent:
Is our goal memory usage or speed?

We could use `FixedBitSet#intersectionCount` and keep from having to create a new bit set that is the intersection. 

I am honestly not sure if the implementation here is any faster than just creating the bit set upfront and checking it. During search, you now have to check two bitsets now instead of one. 

If the filter happens to be `< number of docs visited in a typical search`, your implementation here seems like it would be slower.

benwtrent:
Broad feedback: any "optimizations" without benchmarking aren't optimizations, they are just guesses.

I am curious to see if this helps CPU usage in anyway. I could see it helping memory usage.

mikemccand:
I don't fully understand this change, but it looks like it is stalled on proving it shows lower CPU and/or heap/GC load?

Could we benchmark this change using luceneutil?  It's able to create vector indices that have X% deletions and then run `KNNByte/FloatVectorQuery`...

Pulkitg64:
Thanks @mikemccand for the pointers. Will try to run benchmarks on this change.

PR #12862, opened 2023-11-30
Add Facets#getBulkSpecificValues method

epotyom:
Add `Facets#getBulkSpecificValues` method and implemented it in every class that implements `getSpecificValue`.

Also moved FacetLabel class one level up from `org.apache.lucene.facet.taxonomy` to `org.apache.lucene.facet` to be able to use it as argument for `getBulkSpecificValues`.

Issue #12919


lucene/CHANGES.txt
```diff
@@ -67,6 +67,8 @@ API Changes
 
 * GITHUB#11023: Adding -level param to CheckIndex, making the old -fast param the default behaviour. (Jakub Slowinski)
 
+* GITHUB#12180: Add Facets#getBulkSpecificValues method to get specific values for multiple FacetLabel at once. (Egor Potemkin)
```
mikemccand, line 70: `FacetLabel` -> `FacetLabels`?

lucene/facet/src/java/org/apache/lucene/facet/Facets.java
```diff
@@ -44,11 +50,17 @@ public abstract FacetResult getTopChildren(int topN, String dim, String... path)
       throws IOException;
 
   /**
-   * Return the count or value for a specific path. Returns -1 if this path doesn't exist, else the
-   * count.
+   * Return the count or value for a specific path. Returns -1 {@link #MISSING_SPECIFIC_VALUE} if
```
mikemccand, line 53: Hmm remove the `-1`?  We want users to use the constant not -1 right?

lucene/facet/src/java/org/apache/lucene/facet/LongValueFacetCounts.java
```diff
@@ -568,6 +568,12 @@ public Number getSpecificValue(String dim, String... path) {
     throw new UnsupportedOperationException();
   }
 
+  @Override
+  public Number[] getBulkSpecificValues(FacetLabel[] facetLabels) {
+    // TODO: should we impl this?
```
mikemccand, line 573: Couldn't we emulate it by calling `getSpecificValue` N times and collating?

lucene/facet/src/java/org/apache/lucene/facet/MultiFacets.java
```diff
@@ -77,6 +80,39 @@ public Number getSpecificValue(String dim, String... path) throws IOException {
...
+  @Override
+  public Number[] getBulkSpecificValues(FacetLabel[] facetLabels) throws IOException {
+    if (facetLabels.length == 0) {
+      return new Number[0];
+    }
+    // Split facetLabels into chunks by the Facets they belog to
+    Map<Facets, IntArrayList> labelsPerFacet = new HashMap<>();
+    for (int i = 0; i < facetLabels.length; i++) {
+      String dim = facetLabels[i].components[0];
+      Facets facets = dimToFacets.get(dim);
```
mikemccand, line 92: Hmm does the `Facets` class have a reasonable `hashCode/equals`?

mikemccand (commented):
Thanks @epotyom -- this was a lot of work clearly and it looks great!  I left minor comments.

lucene/CHANGES.txt
epotyom: Done! I'm never sure if I'm supposed to use plurals for class names in sentences like this...

lucene/facet/src/java/org/apache/lucene/facet/MultiFacets.java
```diff
@@ -77,6 +80,39 @@ public Number getSpecificValue(String dim, String... path) throws IOException {
...
+      labelsPerFacet.computeIfAbsent(facets, f -> new IntArrayList()).add(i);
+    }
+    // Call getBulkSpecificValues for each chunk and reconcile results
+    Number[] result = new Number[facetLabels.length];
+    for (Map.Entry<Facets, IntArrayList> fl : labelsPerFacet.entrySet()) {
+      FacetLabel[] labelsForFacet = new FacetLabel[fl.getValue().size()];
+      for (IntCursor ordI : fl.getValue()) {
+        labelsForFacet[ordI.index] = facetLabels[ordI.value];
+      }
+      Number[] ords = fl.getKey().getBulkSpecificValues(labelsForFacet);
```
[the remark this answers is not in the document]
epotyom: It is because there are Facet implementations for integer and float values that share the same base class `Facets`.

Oh, `ords` is a bad name for this variable - renamed.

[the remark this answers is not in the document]
epotyom: Hmm, re-read your message. I don't think we can change `getSpecificValue` behavior as users might not check for `null`. But we can choose to implement `null` for the new method. I think it's more intuitive to use `null`, but It could be a little weird that 2 sibling methods behave differently?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyWriter.java
```diff
@@ -32,8 +32,8 @@
 import org.apache.lucene.document.Field;
 import org.apache.lucene.document.NumericDocValuesField;
 import org.apache.lucene.document.StringField;
+import org.apache.lucene.facet.FacetLabel;
```
mikemccand, line 35: This is a public API break (moving `FacetLabel` up one package).  I think that's fine, but I think that means we should only merge this into 10.0 / main?

mikemccand (approved):
Thanks @epotyom -- this looks great!

I just left one comment about the API break... I think this must be 10.0 only?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyWriter.java
epotyom: Oh, I thought it's ok to move it because it is labelled as `@lucene.internal`. I think merging to 10.0/main only is fine. The alternatives are:
1. Pass 2d array of path components - `String[][] pathComponents`, cons: I think it makes API slightly less intuitive/readable?
2. Create a new interface in `org.apache.lucene.facet` and pass its array - `FacetComponents[]`, cons: exposing both new interface and `FacetLabel` as public API is kind of confusing?
3. Don't move the class and import it from `...taxonomy` in all non-taxonomy facet implementations, cons: doesn't look great;
But I think I still prefer current implementation a little more.

epotyom:
@mikemccand oh I just realised that replying to code line comment doesn't send a notification!

Thank you for reviewing, I listed some alternatives above, but I still think that current approach is better and I think it's ok to only merge to 10.0, but please let me know if you think otherwise.

I've rebased the branch to resolve conflicts.

gsmiller:
@epotyom I'd be happy to do a pass on this and help get it merged if you like. Would you mind resolving the current merge conflicts then I'll have a look?

epotyom:
> [gsmiller](/gsmiller)



> @epotyom I'd be happy to do a pass on this and help get it merged if you like. Would you mind resolving the current merge conflicts then I'll have a look?

Thanks @gsmiller , I've rebased the branch.

lucene/facet/src/java/org/apache/lucene/facet/LongValueFacetCounts.java
gsmiller: I am a bit curious around whether-or-not we should have a default `getBulkSpecificValues` implementations in the `Facets` class that delegates to `getSpecificValue`. We have this pattern in other places. Then again, looking at all your implementations, they do something inherently different than just delegating in this way, so maybe that's too trappy. What if we added to the `Facets#getBulkSpecificValues` javadoc to mention that the results should be identical to calling `getSpecificValue` repeatedly. That's probably enough of a hint to someone implementing the `Facets` class on their own that they could delegate in this way if they want? Personally, I think I'd be more in favor of providing a default implementation that delegates along with javadoc hinting to implementers that they should consider overriding it with a more performant implementation, but I don't feel super strongly.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyWriter.java
gsmiller: Let's also add a note to the `MIGRATE.md` file calling out that the location of the `FacetLabel` class has moved to a new package.

gsmiller:
@epotyom I saw Mike already approved. I did my own pass anyway so I understand the changes. They look good. Thanks! I left one small question for you about a default implementation. Let's resolve that open question then I'll merge.

lucene/facet/src/java/org/apache/lucene/facet/LongValueFacetCounts.java
epotyom: My first instinct was to do the default implementation, the same way it works for `TaxonomyReader.getBulkPath`. But the difference is that `TaxonomyReader` has only one implementation as of now, which makes it easier to track that the method is overriden and it is faster than calling `getPath` in a loop. But `Facets` class has 10+ implementations, so I'm leaning towards not having default method to make sure each implementation optimizes it. It will be easy to add the default method later if necessary.

I've added a note to the javadoc.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyWriter.java
epotyom: Done!

epotyom:
Thank you for reviewing @gsmiller ! I've addressed your comments and updated the branch.

lucene/CHANGES.txt
```diff
@@ -87,6 +87,10 @@ API Changes
 * GITHUB#13146, GITHUB#13148: Remove ByteBufferIndexInput and only use MemorySegment APIs
   for MMapDirectory.  (Uwe Schindler)
 
+* GITHUB#12919: Add Facets#getBulkSpecificValues method to get specific values for multiple FacetLabels at once. (Egor Potemkin)
+
+* GITHUB#12919: Add Facets#getBulkSpecificValues method to get specific values for multiple FacetLabels at once. (Egor Potemkin)
```
gsmiller, line 92: I know it's a big change but we probably don't need double changes entries ;)

lucene/facet/src/java/org/apache/lucene/facet/Facets.java
```diff
@@ -58,6 +58,9 @@ public abstract FacetResult getTopChildren(int topN, String dim, String... path)
   /**
    * Return the counts or values for specific paths. Returns {@link #MISSING_SPECIFIC_VALUE} if
    * corresponding path doesn't exist, else the count.
+   *
+   * <p>{@link #getSpecificValue} for each path should produce similar results, but the bulk method
```
gsmiller, line 62: Can we make this stronger? It should produce identical results right (not "similar")? Or maybe that's not true?

gsmiller (commented):
@epotyom thanks for the revision. Looks like the changes entry snuck in twice and I have one minor comment about the javadoc language. Otherwise looks ready to merge and I'll do so as soon as we address the changes dup. Thanks!

lucene/CHANGES.txt
epotyom: Oh, thanks for catching that! Updated the branch.

lucene/facet/src/java/org/apache/lucene/facet/Facets.java
epotyom: That is true, changed👍

mikemccand:
Hmm it looks like this change was nearly ready to be merged?  But code has since drifted and the PR now has some rust (conflicts)?

PR #12837, opened 2023-11-23
[10.0] Removing TermInSetQuery varargs ctor

slow-J:
Creating a PR for the items discussed in https://github.com/apache/lucene/issues/12243.

Currently, calling the [`KeywordField#newSetQuery`](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/document/KeywordField.java#L174-L180) which takes an array and creates:  2 TermInSetQueries (ctors used:[1](https://github.com/apache/lucene/blob/569533bd76a115e239d8c621d2540462bc12d891/lucene/core/src/java/org/apache/lucene/search/TermInSetQuery.java#L89), [2](https://github.com/apache/lucene/blob/569533bd76a115e239d8c621d2540462bc12d891/lucene/core/src/java/org/apache/lucene/search/TermInSetQuery.java#L102)) with arrays, the TiSQ then immediately changes the array to a list `Arrays.asList(terms)`.

I initially wanted to add another `KeywordField#newSetQuery` ctor which takes `Collection` as a param to remove the needless calls to change the input to array (currently needed to call KeywordField#newSetQuery), which then gets immediately changed back to collection (when creating TiSQ).

The discussion in the issue, mentioned that since TermInSetQuery has constructors which take both array and collection, we should consolidate them (unless I am misunderstanding).

So in this PR, I am removing the both TiSQ array ctor taking varargs (since order does not matter in this case) and changing the `KeywordField#newSetQuery` param to collection.

Closes https://github.com/apache/lucene/issues/12243.

Issue #12243, opened 2023-04-26
Add a new static method for KeywordField#newSetQuery to support collections parameter


slow-J:
### Description

Currently, the [KeywordField](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/document/KeywordField.java) class provides a public static method `newSetQuery(String field, BytesRef... values)`. However, this method only supports varargs parameter for values. I would like to propose adding a new static method that allows for a collections parameter for values, as this would provide greater flexibility.

Proposed method signature:
`public static Query newSetQuery(String field, Collection<BytesRef> values)`

I am willing to make the changes and submit a pull request.
Please let me know if there are any concerns or feedback. Thank you!

rmuir:
"flexibility" is usually a dirty word when it comes to apis, is there a better explanation of the use-case?

Especially given that Collection has .toArray already, it seems really easy for the user to use as-is?

rmuir:
and frankly, i have big concerns that this method will somehow warp into something other than calling .toArray to help the user.

database users be database users and this kind of complexity has crept in here before. 

Let's keep it simple with basic arrays.

slow-J:
Thanks for the quick reply @rmuir , I'll try explain the use case better.

Calling the existing [`newSetQuery`](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/document/KeywordField.java#L174-L180) takes the array and constructs: [1](https://github.com/apache/lucene/blob/569533bd76a115e239d8c621d2540462bc12d891/lucene/core/src/java/org/apache/lucene/search/TermInSetQuery.java#L89),[2](https://github.com/apache/lucene/blob/569533bd76a115e239d8c621d2540462bc12d891/lucene/core/src/java/org/apache/lucene/search/TermInSetQuery.java#L102) TermInSetQueries which both then wrap the array with a list `Arrays.asList(terms)`.

So in the case of having a List of BytesRef, we would first call `.toArray`, and then `TermInSetQuery` immediately calls `Arrays.asList`.
[...]

PR #12864, opened 2023-12-01
[9.10] Mark TermInSetQuery ctors with varargs terms as deprecated

slow-J:
Issue: https://github.com/apache/lucene/issues/12243.
The deprecated items here are being removed in `main` branch through PR:https://github.com/apache/lucene/pull/12837.

For methods calling this, also adding new methods taking a collection param and deprecating previous ones with varargs param: SortedSetDocValuesField#newSlowSetQuery, SortedDocValuesField#newSlowSetQuery, KeywordField#newSetQuery

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -7,7 +7,8 @@ http://s.apache.org/luceneversions
 
 API Changes
 ---------------------
-(No changes)
+* GITHUB#12243: Mark TermInSetQuery ctors with varargs terms as @Deprecated. SortedSetDocValuesField#newSlowSetQuery,
+SortedDocValuesField#newSlowSetQuery, KeywordField#newSetQuery now take a collection of terms as a param. (Jakub Slowinski)
 
 New Features
 ---------------------
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -8,7 +8,7 @@ http://s.apache.org/luceneversions
 API Changes
 ---------------------
 * GITHUB#12243: Mark TermInSetQuery ctors with varargs terms as @Deprecated. SortedSetDocValuesField#newSlowSetQuery,
-SortedDocValuesField#newSlowSetQuery, KeywordField#newSetQuery now take a collection of terms as a param. (Jakub Slowinski)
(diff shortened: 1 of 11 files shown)


lucene/CHANGES.txt
```diff
@@ -7,7 +7,8 @@ http://s.apache.org/luceneversions
 
 API Changes
 ---------------------
-(No changes)
+* GITHUB#12243: Mark TermInSetQuery ctors with varargs terms as @Deprecated. SortedSetDocValuesField#newSlowSetQuery,
+SortedDocValuesField#newSlowSetQuery, KeywordField#newSetQuery now take a collection of terms as a param. (Jakub Slowinski)
```
gsmiller, line 11: minor: let's left-pad with some whitespace like we do with other changes entries

lucene/core/src/java/org/apache/lucene/document/KeywordField.java
```diff
@@ -170,7 +171,9 @@ public static Query newExactQuery(String field, String value) {
    * @param values the set of values to match
    * @throws NullPointerException if {@code field} is null.
    * @return a query matching documents with this exact value
+   * @deprecated Use the method with a collection of values instead.
```
gsmiller, line 174: minor: let's use the `{@link ...}` javadoc to reference the methods users should use (in all places where we're using this annotation)?

gsmiller (commented):
This looks right to me as a back-port, but I think we should also bring in the testing changes you made on `main` too? It would be nice if there were no more references in our codebase to these methods on 9.x, and we only keep them around for users.

slow-J:
> Mark TermInSetQuery ctors with varargs terms as @Deprecated.

Ah I initially thought of doing this, but didn't know which was the right way, I'll add now.

lucene/core/src/java/org/apache/lucene/document/KeywordField.java
slow-J: Will do!

lucene/CHANGES.txt
slow-J: Good catch :D You mean on the second line, right?

slow-J:
Thanks for the review @gsmiller !

gsmiller (approved):
Looks good to me. Thanks @slow-J!

PR #12865, opened 2023-12-01
Fix intermittently failing TestParallelLeafReader

ChrisHegarty:
This commit fixes the intermittently failing TestParallelLeafReader.

The ParallelLeafReader requires the document order to be consistent across indexes - each document contains the union of the fields of all documents with the same document number. The test asserts this. But now, with MockRandomMergePolicy potentially reversing the doc ID order while merging, this invalidates the assumption of the test indexes and assertions. The solution is to just ensure that no merging actually happens in these tiny test indexes.

Similar root cause as #12850.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/index/TestParallelLeafReader.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestParallelLeafReader.java
@@ -301,7 +301,11 @@ public class TestParallelLeafReader extends LuceneTestCase {
 
   private Directory getDir1(Random random) throws IOException {
     Directory dir1 = newDirectory();
-    IndexWriter w1 = new IndexWriter(dir1, newIndexWriterConfig(new MockAnalyzer(random)));
+    IndexWriterConfig conf =
+        newIndexWriterConfig(new MockAnalyzer(random()))
+            .setMaxBufferedDocs(2) // generate few segments
+            .setMergePolicy(NoMergePolicy.INSTANCE); // prevent merges for this test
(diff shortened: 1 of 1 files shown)


uschindler (changes requested):
Hi, looks fine.

The alternative fix is to use LogDocsMergePolicy (which should always be used when somebody uses ParallelReader, because it is the only merge policy that preserves the exact structure of segments as it only relies on number of docs).

Maybe it would be better to use this MP as it is a good example for implementors. So maybe keep the forceMerge and use hard coded LogDocsMergePolicy?

Uwe

uschindler:
See the documentation of class: https://lucene.apache.org/core/9_8_0/core/org/apache/lucene/index/ParallelCompositeReader.html

ChrisHegarty:
That's a good idea @uschindler - updated. The test passes successfully when run in a repeat loop.

uschindler (approved):

jpountz:
Thanks for fixing! FYI LuceneTestCase even has `newLogMergePolicy()` for cases like this when the test doesn't want the doc ID order to change through merging because of either TieredMergePolicy's non-adjacent merges or the new reordering that MockRandomMergePolicy performs. It uses one of LogDocMergePolicy or LogByteSizeMergePolicy with random merge factors and other parameters.

uschindler:
For ParallelCompositeReader tests the LogDoc policy is the only correct one (see documentation).

But I agree, _this_ test does forceMerge to create a single segment index, so segment structure doesn't matter, only document order matters.

The composite reader test already uses LogDoc.

PR #12866, opened 2023-12-01
Prevent extra similarity computation for single-level graphs

kaivalnp:
### Description

[`#findBestEntryPoint`](https://github.com/apache/lucene/blob/4bc7850465dfac9dc0638d9ee782007883869ffe/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java#L151) is used to determine the entry point for the last level of HNSW search

It finds the single best-scoring node from [all upper levels](https://github.com/apache/lucene/blob/4bc7850465dfac9dc0638d9ee782007883869ffe/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java#L159) - but performs an [unnecessary computation](https://github.com/apache/lucene/blob/4bc7850465dfac9dc0638d9ee782007883869ffe/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java#L157) (along with [recording one visited node](https://github.com/apache/lucene/blob/4bc7850465dfac9dc0638d9ee782007883869ffe/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java#L154)) when the graph just has 1 level (so the entry node is just the overall graph's entry node)

Also added a test to demonstrate this (fails without the changes in PR) -- where we visit `graph.size() + 1` nodes when the `topK` is high (should be a maximum of `graph.size()`)

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java
+++ b/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java
@@ -100,19 +100,10 @@ public class HnswGraphSearcher {
       HnswGraphSearcher graphSearcher,
       Bits acceptOrds)
       throws IOException {
-    int initialEp = graph.entryNode();
-    if (initialEp == -1) {
-      return;
+    int ep = graphSearcher.findBestEntryPoint(scorer, graph, knnCollector);
+    if (ep != -1) {
+      graphSearcher.searchLevel(knnCollector, scorer, 0, new int[] {ep}, graph, acceptOrds);
     }
-    int[] epAndVisited = graphSearcher.findBestEntryPoint(scorer, graph, knnCollector.visitLimit());
-    int numVisited = epAndVisited[1];
-    int ep = epAndVisited[0];
-    if (ep == -1) {
-      knnCollector.incVisitedCount(numVisited);
-      return;
-    }
-    knnCollector.incVisitedCount(numVisited);
-    graphSearcher.searchLevel(knnCollector, scorer, 0, new int[] {ep}, graph, acceptOrds);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java
```diff
@@ -100,19 +100,10 @@ private static void search(
       HnswGraphSearcher graphSearcher,
       Bits acceptOrds)
       throws IOException {
-    int initialEp = graph.entryNode();
-    if (initialEp == -1) {
-      return;
+    int ep = graphSearcher.findBestEntryPoint(scorer, graph, knnCollector);
```
benwtrent, line 103: I think you should still check `collector.earlyTerminated()` after finding the best entry point as we increment the visited count after each check. Meaning we could have visited too many by one, and there is no point in continuing.

benwtrent (commented):
Nice find on the small off-by-one error.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java
kaivalnp: Makes sense! Added now..

benwtrent (approved):
This is a nice little edge case bug fix. I don't think the optimization will show up in benchmarks. But a CHANGES entry for Lucene 9.10 is warranted IMO :).

Maybe under bugfix?

benwtrent:
@kaivalnp you can add it :)

kaivalnp:
Replicated [this commit](https://github.com/apache/lucene/commit/94b879a5) (which added the 9.9.0 entry) for 9.10.0
Please let me know if I missed something..

PR #12870, opened 2023-12-01
[Minor] Shorten getOrdinal synchronized loop

stefanvodita:
### Description

`DirectoryTaxonomyReader.getOrdinal` seems to have a wider synchronized loop than needed when checking if the label we've asked about exists in the cache. `getBulkOrdinal` already has a [smaller corresponding synchronized loop](https://github.com/apache/lucene/blob/65d30ca1af15ecbec28618338aebe097e5516afc/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/DirectoryTaxonomyReader.java#L336).

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -165,7 +165,8 @@ New Features
 
 Improvements
 ---------------------
-(No changes)
+
+* GITHUB#12870: Tighten synchronized loop in DirectoryTaxonomyReader#getOrdinal. (Stefan Vodita)
 
 Optimizations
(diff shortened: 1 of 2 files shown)


gsmiller (approved):
Looks right to me. Thanks! @stefanvodita want to add a CHANGES entry? Even though it's a small change, probably still good to track there.

mikemccand (approved):
Thanks @stefanvodita -- looks right.

gsmiller (approved):
Thanks @stefanvodita !

PR #12872, opened 2023-12-03
Throw CorruptSegmentInfoException on encountering missing segment info (_N.si) file in CheckIndex

gokaai:
### Description

Changes: 

- Throw CorruptSegmentInfoException on encountering missing segment info (_N.si) file in CheckIndex. This is a step towards enabling #7820

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
+++ b/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
@@ -947,7 +947,7 @@ public final class CheckIndex implements Closeable {
 
     final Version version = info.info.getVersion();
     if (info.info.maxDoc() <= 0) {
-      throw new CheckIndexException(" illegal number of documents: maxDoc=" + info.info.maxDoc());
+      throw new CheckIndexException("illegal number of documents: maxDoc=" + info.info.maxDoc());
     }
 
     int toLoseDocCount = info.info.maxDoc();
--- a/lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
+++ b/lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
@@ -43,10 +43,7 @@ import org.apache.lucene.store.DataOutput;
 import org.apache.lucene.store.Directory;
 import org.apache.lucene.store.IOContext;
 import org.apache.lucene.store.IndexOutput;
-import org.apache.lucene.util.CollectionUtil;
-import org.apache.lucene.util.IOUtils;
-import org.apache.lucene.util.StringHelper;
-import org.apache.lucene.util.Version;
+import org.apache.lucene.util.*;
 
 /**
  * A collection of segmentInfo objects with methods for operating on those segments in relation to
@@ -389,13 +386,25 @@ public final class SegmentInfos implements Cloneable, Iterable<SegmentCommitInfo
     }
 
     long totalDocs = 0;
+    SegmentInfo info;
     for (int seg = 0; seg < numSegments; seg++) {
       String segName = input.readString();
       byte[] segmentID = new byte[StringHelper.ID_LENGTH];
       input.readBytes(segmentID, 0, segmentID.length);
       Codec codec = readCodec(input);
-      SegmentInfo info =
-          codec.segmentInfoFormat().read(directory, segName, segmentID, IOContext.READ);
+      try {
+        info = codec.segmentInfoFormat().read(directory, segName, segmentID, IOContext.READ);
+      } catch (ThreadInterruptedException e) {
+        throw e;
+      } catch (Exception e) {
+        if (infoStream != null) {
+          message(e.getMessage());
+        }
+        throw new CorruptSegmentInfoException(
+            segName,
+            "segment info file: " + segName + ".si cannot be read - it may be missing or corrupt",
+            input);
+      }
       info.setCodec(codec);
       totalDocs += info.maxDoc();
       long delGen = CodecUtil.readBELong(input);
--- a/lucene/core/src/java/org/apache/lucene/index/CorruptSegmentInfoException.java
+++ b/lucene/core/src/java/org/apache/lucene/index/CorruptSegmentInfoException.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.index;
+
+import org.apache.lucene.store.DataInput;
+import org.apache.lucene.store.DataOutput;
+
+/**
+ * This exception is thrown when Lucene is unable to read a SegmentInfo file _N.si (When it is
+ * missing or corrupt)
+ */
+public class CorruptSegmentInfoException extends CorruptIndexException {
+
+  // The segment name of the missing .si file is always stored for accurate error reporting
+  String segmentName;
+
+  /** Create exception with the segmentName and message only */
+  public CorruptSegmentInfoException(String segmentName, String message, DataInput input) {
+    super(message, input);
+    this.segmentName = segmentName;
+  }
+
+  /** Create exception with the segmentName and message only */
+  public CorruptSegmentInfoException(String segmentName, String message, DataOutput output) {
+    super(message, output);
+    this.segmentName = segmentName;
+  }
+
+  /** Create exception with the segmentName, message and root cause. */
+  public CorruptSegmentInfoException(
+      String segmentName, String message, DataInput input, Throwable cause) {
+    super(message, input, cause);
+    this.segmentName = segmentName;
+  }
+
+  /** Create exception with the segmentName, message and root cause. */
+  public CorruptSegmentInfoException(
+      String segmentName, String message, DataOutput output, Throwable cause) {
+    super(message, output, cause);
+    this.segmentName = segmentName;
+  }
+
+  /** Create exception with the segmentName and message only */
+  public CorruptSegmentInfoException(
+      String segmentName, String message, String resourceDescription) {
+    super(message, resourceDescription);
+    this.segmentName = segmentName;
+  }
+
+  /** Create exception with the segmentName, message and root cause. */
+  public CorruptSegmentInfoException(
+      String segmentName, String message, String resourceDescription, Throwable cause) {
+    super(message, resourceDescription, cause);
+    this.segmentName = segmentName;
+  }
+}
--- a/lucene/core/src/test/org/apache/lucene/index/TestTransactions.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestTransactions.java
@@ -217,7 +217,10 @@ public class TestTransactions extends LuceneTestCase {
         } catch (Exception e) {
           // can be rethrown as RuntimeException if it happens during a close listener
           if (!e.getMessage().contains("on purpose")) {
-            throw e;
+            // Caught "on-purpose" IOException can be rethrown as CorruptSegmentInfoException
+            if (!(e instanceof CorruptSegmentInfoException)) {
+              throw e;
+            }
           }
           // release resources
           IOUtils.closeWhileHandlingException(r1, r2);


lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -4270,11 +4290,21 @@ public int doCheck(Options opts) throws IOException, InterruptedException {
...
                 + " documents would be lost, if -exorcise were specified\n");
       } else {
-        opts.out.println("WARNING: " + result.totLoseDocCount + " documents will be lost\n");
-        opts.out.println(
-            "NOTE: will write new segments file in 5 seconds; this will remove "
-                + result.totLoseDocCount
-                + " docs from the index. YOU WILL LOSE DATA. THIS IS YOUR LAST CHANCE TO CTRL+C!");
+        if (result.numBadSegmentsWithoutSegmentInfo > 0) {
+          opts.out.println(
+              "WARNING: " + result.totLoseDocCount + " or more documents will be lost\n");
```
mikemccand, line 4295: Can we change the wording to `at least ...`?  I think that's stronger / more alarming on casual read than ` or more`.

Or maybe `more than XX documents will be lost`, since we (almost certainly) know the lost segment(s) have `maxDoc > 0`?

Whichever wording we use, let's be consistent in the thrown exception message above too.

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
```diff
@@ -389,13 +389,39 @@ private static void parseSegmentInfos(
...
     for (int seg = 0; seg < numSegments; seg++) {
       String segName = input.readString();
       byte[] segmentID = new byte[StringHelper.ID_LENGTH];
       input.readBytes(segmentID, 0, segmentID.length);
       Codec codec = readCodec(input);
-      SegmentInfo info =
-          codec.segmentInfoFormat().read(directory, segName, segmentID, IOContext.READ);
+      try {
+        info = codec.segmentInfoFormat().read(directory, segName, segmentID, IOContext.READ);
+      } catch (FileNotFoundException | NoSuchFileException e) {
```
mikemccand, line 400: Hmm can we upgrade this to catch any "normal" (not crazy things like OOME) `Throwable`?  E.g. if the `.si` file had some bits scrambled or what not we will get a more exotic exception or maybe an exception because the checksum is mismatched.

Can we somehow return the root cause exception here and include it in `CheckIndexException` that we raise above?

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
```diff
@@ -389,13 +389,39 @@ private static void parseSegmentInfos(
...
+        // https://github.com/apache/lucene/issues/7820 - If we throw a CorruptIndex exception here,
+        // a missing segment info will cause parseSegmentInfos() to short-circuit and stop reading
+        // the rest of the segments. So, when encountering a segment with a missing
+        // segment info (.si) file, we initialize an empty segment info (containing 0 docs)
+        // to keep track of the segment name, so that we can continue reading other segments, and
+        // the segment with the missing .si file can be later validated or exorcised by CheckIndex()
+        info =
```
mikemccand, line 407: Hmm I don't like that this will suppress corruption when `SegmentInfos` is being loaded NOT from `CheckIndex`?  This is the wrong layer to have such leniency?  This method should throw the exception back up to the caller and caller should handle properly?

Could we instead 1) create a subclass of `CorruptIndexException`, maybe `CorruptSegmentInfoException` or so, 2) require that caller pass in the segment name, and root cause exception, when creating it, and 3) catch any `Throwable` here and re-throw that subclass?

Then we can fix `CheckIndex` to specifically catch that, pull the segment name out, and do its thing.

mikemccand (commented):
Thank you for tackling this @gokaai!  I added some comments -- I'm hoping we can find a less invasive way to know which `.si` failed.

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
gokaai: Makes sense! Will rephrase as 'at least' in all occurrences

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
gokaai: Yes - will catch all exceptions and throw a new, specific `CorruptSegmentInfoException`

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
gokaai: Thanks for reviewing @mikemccand ! What I find tricky here are two things - 
 - Throwing an exception in `CheckIndex` short-circuits the flow that makes it possible for `-exorcise` to do it's thing
 - Since we don't do a deeper level check on segments which aren't the latest segment, the recently added unit test where we expect prior commits to throw an exception if corrupt, does not pass

I'll make a new revision trying to address both of these things + your comments

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -664,7 +667,7 @@ public int compare(String a, String b) {
     // failure (fastFail == false).  so if we get here, we should // always have a valid lastCommit:
     assert lastCommit != null;
 
-    if (lastCommit == null) {
+    if (lastCommit == null) { // TODO: Remove/move unreachable if statement?
```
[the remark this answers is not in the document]
gokaai: But we already assert `lastCommit != null` just before

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -957,6 +974,9 @@ private Status.SegmentInfoStatus testSegment(
     SegmentReader reader = null;
 
     try {
+      if (info.info.maxDoc() == 0) {
+        throw new CheckIndexException(" this segment has missing or corrupt segment info");
```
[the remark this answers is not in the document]
gokaai: > Remove the extra space before this

I was keeping it consistent with [this exception message](https://github.com/apache/lucene/blob/ba81826951c1f628efb60c6165add27601173bd9/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java#L949) - should I change this one too?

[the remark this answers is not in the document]
mikemccand: Ahh yes please?

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
mikemccand: > * Throwing an exception in `CheckIndex` short-circuits the flow that makes it possible for `-exorcise` to do it's thing

Hmm this should still work.  The exception should be thrown at a low level, when the codec is trying to read the `.si` file.  This is then caught by `CheckIndex`, and it knows which segment is affected by checking `instanceof` of the exception and then extracting the segment name.  Then it can report the root cause and/or exorcise just fine?

> * Since we don't do a deeper level check on segments which aren't the latest segment, the recently added unit test where we expect prior commits to throw an exception if corrupt, does not pass

Hmm why is that?  Codec should throw an exception when we try to load that (non-latest) SIS too?

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
gokaai: How about `IOException`? Tried catching all `Throwable t` here but it causes `TestIndexWrtier#testThreadInterruptDeadlock` to fail as it catches `ThreadInterruptExceptions` when it shouldn't

gokaai:
Decided to tackle this in smaller steps - 
- Made a new (simpler) revision where we throw a more specific error on encountering missing segment info. Created a new exception `CorruptSegmentInfoException.java`
- Will create a new commit (or issue?) to add in fixes and unit tests for exorciseIndex

mikemccand (commented):
Thanks @gokaai!

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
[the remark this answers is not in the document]
mikemccand: Ahh you are right!  And looking at the code, unless there is a bug, `lastCommit` should never be null.  So I think an `assert` is appropriate and we should remove the `if`.

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
mikemccand: Exception handling is tricky!  I'm not sure what is the right exact set of exceptions to catch here ... but it should be broader than just the "no such file" exceptions I think.  Corruption in `.si` can result in exotic exceptions...

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
```diff
@@ -389,13 +389,26 @@ private static void parseSegmentInfos(
...
+      } catch (Throwable t) {
+        if (infoStream != null) {
+          message(t.getMessage());
+        }
+        if (t instanceof NoSuchFileException || t instanceof FileNotFoundException) {
+          throw new CorruptSegmentInfoException(
+              "segment info file: " + segName + ".si cannot be read - it may be missing or corrupt",
+              input);
+        } else {
+          throw t;
```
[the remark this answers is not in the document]
gokaai: I tried catching all Throwable t here but it causes `TestIndexWrtier#testThreadInterruptDeadlock` to fail as it doesn't throw `ThreadInterruptExceptions`

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
```diff
@@ -389,13 +386,25 @@ private static void parseSegmentInfos(
...
       byte[] segmentID = new byte[StringHelper.ID_LENGTH];
       input.readBytes(segmentID, 0, segmentID.length);
       Codec codec = readCodec(input);
-      SegmentInfo info =
-          codec.segmentInfoFormat().read(directory, segName, segmentID, IOContext.READ);
+      try {
+        info = codec.segmentInfoFormat().read(directory, segName, segmentID, IOContext.READ);
+      } catch (ThreadInterruptedException e) {
+        throw e;
+      } catch (Exception e) {
```
gokaai, line 399: Catching a general-purpose exception here was tricky !
- `TestIndexWrtier#testThreadInterruptDeadlock` expected a ThreadInterruptedException to be thrown as is
- `TestTransactions#testTransactions` expected an 'on-purpose' IOException to be thrown as is

gokaai:
I edited the PR description to more accurately represent what is being done. We aren't dealing with the exorcism issue yet, but throwing the right exception on encountering missing segment info. I will tackle the final step (actually making exorcism possible) in a subsequent PR

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
mikemccand: Thanks @gokaai!

mikemccand:
Thanks @gokaai -- I'll try to review soon!

If possible please try not to force-push: it removes the history of the past commits and makes it harder to see what changed on this iteration :)

gokaai:
> Thanks @gokaai -- I'll try to review soon!
> 
> If possible please try not to force-push: it removes the history of the past commits and makes it harder to see what changed on this iteration :)

Sorry about that @mikemccand . I messed up my workspace and couldn't salvage it - I'll keep it in mind while making any new revisions.

mikemccand:
> > Thanks @gokaai -- I'll try to review soon!
> > If possible please try not to force-push: it removes the history of the past commits and makes it harder to see what changed on this iteration :)
> 
> Sorry about th

Ahh OK no worries.  I experience such `git` disasters myself far more often than I would hope :)

dweiss:
> > > Thanks @gokaai -- I'll try to review soon!
> > > If possible please try not to force-push: it removes the history of the past commits and makes it harder to see what changed on this iteration :)
> > 
> > 
> > Sorry about th
> 
> Ahh OK no worries. I experience such `git` disasters myself far more often than I would hope :)

Look at what git reflog does and how it may help. It is really, really difficult to not be able to recover your work with git.

serhiy-bzhezytskyy:
This PR looks stalled — no longer mergeable, and the last comment from a human is from February 2024 — but the problem it fixes is still there, so I have carried it forward in #16476 rather than let it sit.

Your commit is the first one there, rebased onto `main` with your authorship intact, and you are credited in `CHANGES.txt`. Two hunks needed adjusting: `codec.segmentInfoFormat().read(...)` now takes `IOContext.READONCE` on `main`, and the wildcard import had to go back to explicit ones for the ecj lint. `CheckIndex.java` and `TestTransactions.java` merged cleanly.

On top of it I finished the two review points from this thread that the current diff does not cover:

- The root cause was being dropped. `CorruptSegmentInfoException` has a constructor taking a `Throwable cause`, but the throw site calls the three-argument one — and that cause is the only thing naming the file (`NoSuchFileException` with the full path for a deleted `.si`, a suppressed `EOFException` naming the `MemorySegmentIndexInput` for a truncated one). It is passed now, and the catch is `Exception | AssertionError` per @mikemccand's comment about exotic exceptions.
- `segmentName` was package-private with no accessor, so nothing outside `org.apache.lucene.index` could read the name the exception exists to carry, and no test could assert on it. There is a `getSegmentName()` now, and a test covering a deleted and a truncated `.si`.

@gokaai if you would rather finish it here, say so and I will close #16476 — the point is the fix landing, not where. The `exorciseIndex` half that you deferred is #16474.

PR #12038, opened 2022-12-24
remove non-NRT replication support

rmuir:
Lucene's replicator/ module has really two replication APIs: NRT and the older non-NRT.

The NRT replication is nice, it is actually JUST an API, hence there's no network support or anything like that.

The non-NRT replication has some issues:
* Uses HTTP but in a non-standard way: binary blobs etc. Responses can't be cached or leverage CDNs or anything.
* legacy plaintext HTTP 1.x only, No support for HTTPS, HTTP/2, etc
* Giant security hole: uses java (de)serialization
* legacy web apis (servlet). it is 2022
* drags in third party http client, doesn't use the one in the standard library
* drags in third party logging jars, doesn't use the one in the standard library

I'd like to deprecate the non-NRT support in 9.x and remove in 10.x. I suspect anyone using this module is using the newer NRT mode? If anyone is still using the legacy non-NRT mode, please let me know on this issue and give me your IP address, so I can try to pop a shell.

Please see #11381 for more information (I copied the description from there). I opened the issue about a year ago and there has not been any negative feedback.

Closes #11381

PR #12873, opened 2023-12-03
Rewrite JavaScriptCompiler to use modern JVM features (Java 17)

### LINKED ISSUE #8933: Maybe expression compiler should cache recently compiled expressions? [LUCENE-7882]
I've been running search performance tests using a simple expression (`_score + ln(1000+unit_sales)`) for sorting and hit this odd bottleneck:

```
"pool-1-thread-30" `#70` prio=5 os_prio=0 tid=0x00007eea7000a000 nid=0x1ea8a waiting for monitor entry [0x00007eea867dd000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at org.apache.lucene.expressions.js.JavascriptCompiler$CompiledExpression.evaluate(_score + ln(1000+unit_sales))
	at org.apache.lucene.expressions.ExpressionFunctionValues.doubleValue(ExpressionFunctionValues.java:49)
	at com.amazon.lucene.OrderedVELeafCollector.collectInternal(OrderedVELeafCollector.java:123)
	at com.amazon.lucene.OrderedVELeafCollector.collect(OrderedVELeafCollector.java:108)
	at org.apache.lucene.search.MultiCollectorManager$Collectors$LeafCollectors.collect(MultiCollectorManager.java:102)
	at org.apache.lucene.search.Weight$DefaultBulkScorer.scoreAll(Weight.java:241)
	at org.apache.lucene.search.Weight$DefaultBulkScorer.score(Weight.java:184)
	at org.apache.lucene.search.BulkScorer.score(BulkScorer.java:39)
	at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:658)
	at org.apache.lucene.search.IndexSearcher$5.call(IndexSearcher.java:600)
	at org.apache.lucene.search.IndexSearcher$5.call(IndexSearcher.java:597)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

I couldn't see any `synchronized` in the sources here, so I'm not sure which object monitor it's blocked on.

I was accidentally compiling a new expression for every query, and that bottleneck would cause overall QPS to slow down drastically (\~4X slower after \~1 hour of redline tests), as if the JVM is getting slower and slower to evaluate each expression the more expressions I had compiled.

I tested JDK 9-ea and it also kept slowing down over time as the performance test ran.

Maybe we should put a small cache in front of the expressions compiler to make it less trappy?  Or maybe we can get to the root cause of why the JVM slows down more and more, the more expressions you compile?

I won't have time to work on this in the near future so if anyone else feels the itch, please scratch it!

---
Migrated from [LUCENE-7882](https://issues.apache.org/jira/browse/LUCENE-7882) by Michael McCandless (@mikemccand), updated Sep 17 2020
Attachments: [demo.patch](https://apache.github.io/lucene-jira-archive/attachments/LUCENE-7882/demo.patch)
Pull requests: https://github.com/apache/lucene-solr/pull/1837

uschindler:
This PR rewrites the JavascriptCompiler for the main branch to use more modern JVM features to be easier configurable and does not need a separate classloader per expression (uses hidden classes).

Hidden classes are a new feature of Java 15: It allows to define a hidden class with byte code without using a classloader using the access constraints of a given private lookup. This is the same technique used by lambdas and other invokeDynamic features and replaces the old Unsafe API that Lucene never used. The downsides with a separate on-class classloader are described in #8933. Mainly with many expressions compiled in parallel there's a lock in parent classloader while looking up methods/fields and loading classes.

The new Java 15 hidden classes are unloaded by the JVM if the get out of scope, so the workaround with a separate classloader is no longer needed.

A second reason why separate classloaders were needed is to allow code access functions form other classloaders. This stopped me doing the whole optimization at first until I rewrote the whole function lookup to be more modern:

- The function array is now a `Map<String, MethodHandle>`. This allows to create arbitrary functions anyhere in code like Elasticsearch/Opensearch (the code may even be private or in another classloader). The access to those method is no longer the deal of the compiler, if the MethodHandle is there, it can be used. Access is the deal of the user of the expressions API. In addition you can also use "synthetic" methodhandles like those generated by `MethodHandles#filterArguments` or you can use non-static methods as long as you bind the receiver to a fixed instance. Some examples were added to the tests.
- The problem with MethodHandles is that they should be `static final` constants in the class or must appear in the constant pool of the class. This is now the second trick used here: Since Java 11 there is the feature of dynamic constants: Basically the LDC instruction can load a MethodHandle constant (which is treated as the "perfect" MethodHandle for Hotspot) using a bootstrap method (invokedynamic v2). The Javascript compiler now implements a bootstrap method that looks up the function dynamically from the above `Map<String,MethodHandle>`.
- Java 16 brought another feature that allows to add `classData` to hidden classes (Java 15). This classData is saved in the private bootstrapper `Lookup` and is accessible during bootstrapping constants. When loading the hidden class we pass our function map as classdata to the lookup and this allows the constant bootstrapping to return the MethodHandle.
- I have to do more benchmarking, but generally the dynamic constants and using LDC and `MethodHandle.invokeExact()` should be enough to call ALL functions. For now I added a shortcut: If the `MethodHandle` in our map links directly to a static Method in a public class in same classloader, the bytecode generator for now uses the `MethodHandleInfo` to generate a conventional `invokestatic`. I plan to remove this "optimization" after more benchmarking. You can easily disable the optimization by setting `directFunctions` to an empty Map.

There is one downside in the hidden class feature that I added a workaround for: Hidden classes never appear in stack traces so my initial version did not print the actual JS function in stack trace. I worked around this by adding "exception patching" in the base class. It catches all Throwables throws by the generated code in the subclass and changes the stack trace to include the source code and hidden class name. This allows same user experience before.

The current code is not backward compatible and needs an entry in `MIGRATE.txt`. I will investiagte how to add some helper methods for those people using `java.lang.reflect.Method` maps. Those can be easily converted to `MethodHandle` with some wrapper method. Support for separate classloader was removed. This is not a security issue, because the methods available for calling are defined externally and the API user is still responsible to only supply methods which should be accessible by users. The access checking is done by the provider of function map. For safety you can still let run whole of Lucene or the expressions module in a separate module/classloader.

@rjernst, @jdconrad: Maybe this is also an idea for Painless.

This PR replaces the outdated https://github.com/apache/lucene-solr/pull/1837. This closes #8933.


lucene/expressions/src/java/org/apache/lucene/expressions/js/JavascriptCompiler.java
```diff
@@ -201,48 +210,75 @@ private static void unusedTestCompile() throws IOException {
...
+          @SuppressWarnings("unused")
+          IllegalArgumentException iae) {
+        continue;
+      }
+      if (info.getReferenceKind() == MethodHandleInfo.REF_invokeStatic
+          && REACHABLE_FROM_OUR_CLASSLOADER.get(info.getDeclaringClass())) {
+        map.put(e.getKey(), info);
+      }
+    }
+    this.directFunctions = Collections.unmodifiableMap(map);
```
uschindler, line 235: set this to `Map.empty()` to test using dynamic constants and `MethodHandles#invokeExact` for all functions.

uschindler:
The latest commit changed the stack trace patching to be implemented inside the compiled expression. The original benchmarks had shown that the extra `evaluate()` method in the base class caused some slowdown. By doing the stack trace patching the "new way", all is sane again. Due to the change the original `Expression` class is now unmodified.

rmuir:
I like the change!

Can be a followup ticket, but i think we should remove `ClassLoader` usage from `TestCustomFunctions.java`. It doesn't need to use classloader to test anymore, since it can just pass a MethodHandle: so maybe the test is easy to fix. With that test fixed, `createClassLoader` permission can be removed from tests, lucene code really shouldn't be messing with classloaders if we have a more optimal way?

uschindler:
I will also post my benchmark, as it allows to test the performance of expressions.

uschindler:
Benchmark was added and classloader code removed.

I will add a helper function to support some compatibility with legacy code still using `java.lang.reflect.Method`, but after that it's ready. I will also add a MIGRATE information how to convert old code.

uschindler:
While writing the benchmark I figured out that the Expression base class has some problems with a missing `IOException` on the `evaluate(DoubleValues[])` method. As bytecode knows no exceptions in throws clauses, the evaluation method may throw `IOException` unexpected because `DoubleValues#doubleValue()` is defined to throw `IOException`. This is a bug and should be fixed separately. I will open an issue to fix this in main.

uschindler:
I added documentation and migration guide.

lucene/expressions/src/java/org/apache/lucene/expressions/js/JavascriptCompiler.java
```diff
@@ -201,48 +226,57 @@ private static void unusedTestCompile() throws IOException {
...
           exception);
     }
   }
 
+  private Expression invokeConstructor(
+      Lookup lookup, Class<?> expressionClass, Map<String, Integer> externalsMap)
+      throws ReflectiveOperationException {
+    final MethodHandle ctor = lookup.findConstructor(expressionClass, MT_EXPRESSION_CTOR_LOOKUP);
+    try {
+      return (Expression) ctor.invoke(sourceText, externalsMap.keySet().toArray(String[]::new));
```
rmuir, line 272: any easy way to change this to `invokeExact()` ? I realize its just for creating the expression but you know how these things go. I tried changing it locally and got WrongMethodTypeException so some adaptation is happening....

uschindler:
> I like the change!
> 
> Can be a followup ticket, but i think we should remove `ClassLoader` usage from `TestCustomFunctions.java`. It doesn't need to use classloader to test anymore, since it can just pass a MethodHandle: so maybe the test is easy to fix. With that test fixed, `createClassLoader` permission can be removed from tests, lucene code really shouldn't be messing with classloaders if we have a more optimal way?

I cleaned up the policy files, but the `createClassLoader` permission is still needed by the the Jacoco agent, which uses no AccessController.... So I can't limit access. But I moved the permission to the Jacoco ones. We may open a bug report for Jacoco to fix this.

I was successful with removing `getClassLoader` permission. This one is no longer needed as we are no longer inspecting classloaders of unknown classes. I think this permission can also be removed in 9.x.

Finally I removed some permission from replication in main branch caused by servlet API.

lucene/expressions/src/java/org/apache/lucene/expressions/js/JavascriptCompiler.java
uschindler: this is a one-time call and manually adapting makes no sense here. We have more of those in Lucene.

You can adapt it, but this is not worth it.

rmuir:
some of the permissions involved are probably only safe to remove in 10.x due to #12038

rmuir (approved):
Thanks for modernizing this, no more classloaders or reflection :)

rjernst:
Thanks for the ping @uschindler , great idea! I opened https://github.com/elastic/elasticsearch/issues/102955 to track the same work in Elasticsearch.

dweiss (approved):

PR #12750, opened 2023-11-02
BaseTokenStreamTestCase.assertAnalyzesTo fails when Analyzer contains…

lukas-vlcek:
… PathHierarchy tokenizer

### Description

This PR is expected to fail. It demonstrates issue with `BaseTokenStreamTestCase.assertAnalyzesTo()` method in connection to `PathHierarchyTokenizer`.

Is there any reason why `PathHierarchyTokenizer` shall not be used in the test like this? There are definitely other tokenizers that are being tested like this, ie. they are wrapped in Analyzer and then `assertAnalyzesTo()` method is called to check the tokens. What is special about PathHierarchy tokenizer that it does not work?

I think the problem might not be in the tokenizer but in the test method itself or in the way I call it (maybe I need to pass in more parameters/flags to get rid of the issue?). The testing method is complex, especially when it gets to `checkAnalysisConsistency()` part.

I am looking for any useful tips. Thank you!

PR #12875, opened 2023-12-04
Fix position increment in (Reverse)PathHierarchyTokenizer

lukas-vlcek:
### Description

Incrementing position attribute for each token in both `PathHieararchyTokenizer` and `ReversePathHieararchyTokenizer`.
This change makes it possible to use both tokenizers in `BaseTokenStreamTestCase.assertAnalyzesTo()` method (test cases extended to demonstrate this).

This PR solves and surpasses https://github.com/apache/lucene/pull/12750.

@msfroh I included your original commit from your Lucene fork to keep proper commit attribution.

Shall we squash all commits?

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -77,6 +77,9 @@ API Changes
 
 * GITHUB#12855: Remove deprecated DrillSideways#createDrillDownFacetsCollector extension method. (Greg Miller)
 
+* GITHUB#12875: Ensure token position is always increased in PathHierarchyTokenizer and ReversePathHierarchyTokenizer
+  and resulting tokens do not overlap. (Michael Froh, Lukáš Vlček)
+
 New Features
 ---------------------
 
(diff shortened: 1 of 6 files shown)


mikemccand:
> I am looking at `TestUnifiedHighlighter*` tests. Does it mean that I need to use specific fieldType? Can I use any fieldType(s) from existing `UHTestHelper.parametersFactoryList()`? Such as `postingsType` and `postingsWithTvType`?

Hi @lukas-vlcek -- I don't think we need to add a specific unit test to `UnifiedHighlighter` tests for this.  I think it's fine to say something general in the `MIGRATE.txt` like "`(Reverse)PathHieararchyTokenizer` now produces sequential (instead of overlapping) tokens with accurate offsets, making positional queries and highlighters possible for fields tokenized with this tokenizer" or so?

lukas-vlcek:
@mikemccand Thanks for review. Done!
Shall I squash all commits are will you do it when merging in GH?

mikemccand:
Thanks @lukas-vlcek -- looks great -- no need to squash, I'll take care during merge.  Exciting improvement, finally!

mikemccand (approved):

PR #12797, opened 2023-11-11
CheckIndex - Adding a `-level` parameter to give ability to control index check detail programmatically

slow-J:
Closes #11023 

I found this old issue and reached out to Zach @zacharymorn, who said that he doesn't plan on picking it up currently so I had a go at tackling this.

Adding a -detailLevel parameter to have 3 levels of detail in checkIndex. The higher the value, the more checks are done.
1 - (Default) Checksum checks only.
2 - All level 1 checks + logical integrity checks.
3 - All level 2 checks + slow checks.

### Test

Command executed: `java  -server -Xms2g -Xmx2g --add-modules jdk.incubator.vector -XX:+HeapDumpOnOutOfMemoryError -XX:+UseParallelGC  -classpath ./lucene/core/build/libs/lucene-core-10.0.0-SNAPSHOT.jar -ea org.apache.lucene.index.CheckIndex -threadCount 16 myindex_dir/index`

-level 1: `Took 1.379 sec total.`
-level 2: `Took 74.824 sec total.`
-level 3: `Took 136.988 sec total.`

—
Behaviour in initial commit before the ~~detailLevel~~ -level param was added:
~~1. Making -fast the new default.~~
~~2. The previous -slow is moved to -slower~~
~~3. The previous default behavior (checksum + segment file content) is activated by -slow.~~

Issue #11023, opened 2021-06-01
Make CheckIndex doChecksumsOnly / -fast as default [LUCENE-9984]


asfimport:
This issue is a spin-off from discussion in https://github.com/apache/lucene/pull/128

Currently _CheckIndex_ defaults to checking both checksum as well as content inside each segment files for correctness, and requires _-fast_ flag to be explicitly passed in to do checksum only. However, this default setting was there due to lack of checksum feature historically, and is slow for most end-users nowadays as they probably only care about their indices being intact (from random bit flipping for example).

This issue is to change the default settings for CheckIndex so that they are more appropriate for end-users. One proposal from @rmuir is the following:
1. Make `-fast` the new default.
1. The previous `-slow` could be moved to `-slower` 
1. The current behavior (checksum + segment file content - slow check) could be activated by `-slow`.

---
Migrated from [LUCENE-9984](https://issues.apache.org/jira/browse/LUCENE-9984) by Zach Chen (@zacharymorn), updated Aug 16 2021

slow-J:
I'll clean up the deprecated CheckIndex params in Lucene 11.

PR #12876, opened 2023-12-04
CheckIndex - Removal of some dead code

slow-J:
When I was working on `CheckIndex` in PR: https://github.com/apache/lucene/pull/12797, Issue: https://github.com/apache/lucene/issues/11023, I came across some dead code that I wanted to clean up as a follow-up.

List of changes:
* Collections.sort -> List.sort, Collections.sort simply delegates to List.sort anyways
* println(String.format) -> printf
* Removal of some unused params in private methods
* Adding more usage of existing `numDocs` variable
* ~~Removed `testPostings` and `testTermVectors`, these methods with 2 params are never called in Lucene~~ (no longer part of this PR)

Please let me know if any of these changes are not needed or are Lucene coding conventions and I will change them back.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
+++ b/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
@@ -2448,6 +2448,16 @@ public final class CheckIndex implements Closeable {
    *
    * @lucene.experimental
    */
+  public static Status.TermIndexStatus testPostings(CodecReader reader, PrintStream infoStream)
+      throws IOException {
+    return testPostings(reader, infoStream, false, Level.MIN_LEVEL_FOR_SLOW_CHECKS, false);
+  }
+
+  /**
+   * Test the term index.
+   *
+   * @lucene.experimental
+   */
   public static Status.TermIndexStatus testPostings(
       CodecReader reader, PrintStream infoStream, boolean verbose, int level, boolean failFast)
       throws IOException {
@@ -3611,6 +3621,16 @@ public final class CheckIndex implements Closeable {
       default:
         throw new AssertionError();
     }
+  }
+
+  /**
+   * Test term vectors.
+   *
+   * @lucene.experimental
+   */
+  public static Status.TermVectorStatus testTermVectors(CodecReader reader, PrintStream infoStream)
+      throws IOException {
+    return testTermVectors(reader, infoStream, false, Level.MIN_LEVEL_FOR_INTEGRITY_CHECKS, false);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


jpountz (commented):
I'm not sure about the removal of these two functions, they're quite nice, only taking a CodecReader and an InfoStream. Removing them is forcing users to call the more complex functions and understand what a level is, or what the boolean means. Other changes look good to me.

slow-J:
> I'm not sure about the removal of these two functions, they're quite nice, only taking a CodecReader and an InfoStream. Removing them is forcing users to call the more complex functions and understand what a level is, or what the boolean means. Other changes look good to me.

Thanks @jpountz, will add these functions back in then.

I was wondering about the use of `@lucene.experimental` throughout this class. It has been experimental for > 10 years. Will the entire class or some parts, ever be classified as non-experimental?

mikemccand:
> I was wondering about the use of `@lucene.experimental` throughout this class. It has been experimental for > 10 years. Will the entire class or some parts, ever be classified as non-experimental?

LOL 10 years of experimental API!  I bet Lucene has even older experimental APIs :)  Too bad we don't have an "experimental API age reporting tool"...

Yeah this is a great question, when to remove the experimental API tag.  We have no well defined process -- it's just devs (like you, thanks!) noticing and raising the question on a case by case basis.  I do think some of these tags in `CheckIndex.java` could be removed, e.g. on each of the `testXXX` methods?  But likely we should leave it on each of the `XXXStatus` returned classes, so we are free to break/change what metadata these classes return?  And let's do it separately -- this PR looks like a great cleanup already.  Progress not perfection!

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -97,11 +96,11 @@
  */
 public final class CheckIndex implements Closeable {
 
+  private final Directory dir;
+  private final Lock writeLock;
```
mikemccand, line 100: Ahh thanks for adding missing `final` tags too.  I wish we had a static checker for this.

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -3216,11 +3213,11 @@ public static Status.DocValuesStatus testDocValues(
   }
 
   @FunctionalInterface
-  private static interface DocValuesIteratorSupplier {
```
mikemccand, line 3219: Heh, the other PR about this.

mikemccand (approved):
Thanks @slow-J -- this looks great.  I'll merge.

PR #12878, opened 2023-12-05
Fix the declared Exceptions of Expression#evaluate to match those of DoubleValues#doubleValue

uschindler:
This PR fixes the issue found while coding the benchmark of Expressions module in #12873: The expressions module looks up the variables using the `DoubleValues#doubleValue` method, which throws `IOException`. Because the JVM does not do any exceptions checks, only the type system of javac compiler, some code calling `Expression#evaluate` could suddenly be confronted with a checked `IOException`, although this is not declared. It is not even possible to catch the Exception because compiler complains that it is not declared.

Actually this should also be fixed in 9.x but this changes the method signature of the `Expression#evaluate` method and the impact is not too high. So it is for Lucene 10 only.

The fix is to add the exception to the declaration in the abstract base class (and also the bytecode generator). This also rewrites the sanity check in JavascriptCompiler to check the method signatures.

### DIFF
--- a/lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/ExpressionsBenchmark.java
+++ b/lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/ExpressionsBenchmark.java
@@ -17,7 +17,6 @@
 package org.apache.lucene.benchmark.jmh;
 
 import java.io.IOException;
-import java.io.UncheckedIOException;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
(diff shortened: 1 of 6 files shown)


dweiss (commented):
LGTM. I think an explicit IOException is better than wrapping in UncheckedIOException.

dweiss (approved):

uschindler:
> LGTM. I think an explicit IOException is better than wrapping in UncheckedIOException.

The UncheckedIOException was just a workaround when I implemented the benchmark code. Wrapping inside ASM bytecode generation would be a desaster, so fix the bug correct.

lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/ExpressionsBenchmark.java
```diff
@@ -83,12 +82,8 @@ private static double ident(double v) {
   private static final Expression NATIVE_IDENTITY_EXPRESSION =
       new Expression(NATIVE_IDENTITY_NAME, new String[] {"x"}) {
         @Override
-        public double evaluate(DoubleValues[] functionValues) {
-          try {
-            return functionValues[0].doubleValue();
-          } catch (IOException e) {
-            throw new UncheckedIOException(e);
```
uschindler, line 90: this was just the workaround I applied that made me aware of this bug in the expressions module.

PR #12624, opened 2023-10-04
Allow FST builder to use different writer (#12543) 

dungba88:
### Description

Split the scratch byte operations inside `BytesStore` out of FSTWriter and allow the DataOutput to be configurable from `FSTCompiler.Builder`. Those operations are for building the frontier node. Only the final bytes will be written to the DataOutput. The Builder will still maintain `bytesPageBits(int)` for backward-compatibility, in which it will use a `ByteBuffersDataOutput`.

Also extract the saving of metadata as a standalone method, as there can be use case where some already saved the FST to a DataOutput when building and only need to save the metadata here.

Issue: https://github.com/apache/lucene/issues/12543

The FST produced by this DataOutput-backed FST is not readable directly, and users need to construct a corresponding DataInput and pass it to the FST public constructor. In short the FSTCompiler only compile the FST but not read it.

The write-then-read-immediately use case is still supported with ByteBuffersDataOutput.

PR #12879, opened 2023-12-06
Optimize FST on-heap BytesReader

dungba88:
### Description

This is the same with #12624, except for a slight change in implementation of `ReadWriteDataOutput`. This PR keeps the original implementation that BytesStore did (just the `getReverseBytesReader()` part), to make sure there is no regression. My theory is that ReverseRandomAccessReader need to seek the correct buffer on every `readByte`, and reading from the ByteBuffer has some (small) overhead comparing to accessing from the byte array directly. But these should have insignificant impact on the performance. More extensive benchmark is needed.

See the diff between 2 approaches at https://github.com/dungba88/lucene/pull/13

Using Test2BFST, there is some improvement over the `ByteBuffersDataInput`:

### ReverseRandomAccessReader with ByteBuffersDataInput

```
  1> TEST: now verify [fst size=4621076364; nodeCount=2252341486; arcCount=2264078585]
  1> 0...: took 0 seconds
  1> 1000000...: took 27 seconds
  1> 2000000...: took 54 seconds
  1> 3000000...: took 82 seconds
  1> 4000000...: took 109 seconds
  1> 5000000...: took 137 seconds
  1> 6000000...: took 165 seconds
  1> 7000000...: took 192 seconds
  1> 8000000...: took 219 seconds
  1> 9000000...: took 247 seconds
  1> 10000000...: took 275 seconds
  1> 11000000...: took 300 seconds
```

### This PR

```
  1> TEST: now verify [fst size=4621076364; nodeCount=2252341486; arcCount=2264078585]
  1> 0...: took 0 seconds
  1> 1000000...: took 22 seconds
  1> 2000000...: took 44 seconds
  1> 3000000...: took 66 seconds
  1> 4000000...: took 89 seconds
  1> 5000000...: took 111 seconds
  1> 6000000...: took 133 seconds
  1> 7000000...: took 155 seconds
  1> 8000000...: took 178 seconds
  1> 9000000...: took 200 seconds
  1> 10000000...: took 222 seconds
  1> 11000000...: took 245 seconds
```

I ran the test several times and the results are consistent.


dungba88:
I think we'll likely go with the approach in #12624 (simpler despite some potential regression). But this approach is to have some comparison, and in case someone has strong argument for favoring the latency.

dungba88:
Note: We already merged #12624 , but this PR can be used as benchmark candidate for https://github.com/apache/lucene/issues/12884

mikemccand:
Actually 300 -> 245 seconds is quite a performance gain.  It seems FST's own store is quite a bit faster than the `ByteBuffer` backed store.  I think we should make this change?  I'll try to review soon.

lucene/core/src/java/org/apache/lucene/util/fst/ReadWriteDataOutput.java
```diff
@@ -56,14 +66,59 @@ public long ramBytesUsed() {
...
   public FST.BytesReader getReverseBytesReader() {
-    assert dataInput != null; // freeze() must be called first
-    return new ReverseRandomAccessReader(dataInput);
+    assert byteBuffers != null; // freeze() must be called first
+    if (byteBuffers.size() == 1) {
+      // use a faster implementation for single-block case
+      return new ReverseBytesReader(byteBuffers.get(0).array());
+    }
+    return new FST.BytesReader() {
+      private byte[] current = byteBuffers.get(0).array();
```
dungba88, line 81: `array()` returns the internal array and there's no byte-copy here

lucene/core/src/java/org/apache/lucene/util/fst/ReadWriteDataOutput.java
```diff
@@ -56,14 +66,59 @@ public long ramBytesUsed() {
...
   @Override
   public FST.BytesReader getReverseBytesReader() {
-    assert dataInput != null; // freeze() must be called first
-    return new ReverseRandomAccessReader(dataInput);
+    assert byteBuffers != null; // freeze() must be called first
+    if (byteBuffers.size() == 1) {
+      // use a faster implementation for single-block case
+      return new ReverseBytesReader(byteBuffers.get(0).array());
+    }
+    return new FST.BytesReader() {
```
mikemccand, line 80: Do tests exercise this path sometimes?  We would need tests  to either make large FSTs, or, randomize the block size so sometimes it's tiny?

mikemccand (commented):
I left some minor comments -- I think this is nearly ready? (though it it still marked as draft?) -- given the gains of this optimization I think it makes sense to specialize the single buffer vs multiple buffer cases.

lucene/core/src/java/org/apache/lucene/util/fst/ReadWriteDataOutput.java
dungba88: I believe `testRealFST` did this sometimes

lucene/core/src/java/org/apache/lucene/util/fst/ReadWriteDataOutput.java
```diff
@@ -56,14 +66,59 @@ public long ramBytesUsed() {
...
+      private byte[] current = byteBuffers.get(0).array();
+      private int nextBuffer = -1;
+      private int nextRead = 0;
```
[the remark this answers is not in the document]
dungba88: Good catch! I've removed it

lucene/core/src/java/org/apache/lucene/util/fst/ReadWriteDataOutput.java
```diff
@@ -56,14 +66,59 @@ public long ramBytesUsed() {
...
   }
 
   @Override
   public FST.BytesReader getReverseBytesReader() {
-    assert dataInput != null; // freeze() must be called first
-    return new ReverseRandomAccessReader(dataInput);
+    assert byteBuffers != null; // freeze() must be called first
+    if (byteBuffers.size() == 1) {
+      // use a faster implementation for single-block case
+      return new ReverseBytesReader(byteBuffers.get(0).array());
```
mikemccand, line 78: Can we `assert hasArray()` before this?   `.array()` is optional (only valid for `DirectByteBuffer` I think).

lucene/core/src/java/org/apache/lucene/util/fst/ReadWriteDataOutput.java
mikemccand: Hmm I'm still worried about test coverage of this.  Conditionals like this are dangerous for Lucene, since our tests only test tiny FSTs, this code path would be rarely/never executed, likely even by our nightly benchmarks.  Ideally we would find a way to randomize the page size in many tests, or at least, one test?

dungba88: Sorry I meant TestFSTs.testRealTerms. The default block size is 32KB and the FST is 55-80KB, thus 2-3 blocks.

lucene/core/src/java/org/apache/lucene/util/fst/ReadWriteDataOutput.java
dungba88: Since we call toWriteableBufferList the array should be accessible. I think if the array is somehow not accessible HeapByteBuffer (the one being used) would throw an exception.

mikemccand: Yeah I realize the code is safe but it takes some thinking to confirm it -- maybe add the assert with your awesome above explanation?  So future code readers know why this `ByteBuffer` is for sure backed by an array.

dungba88: That's a good idea. I've added the assertion. I decided to put it right after the ByteBuffer list is created (in `freeze()`) and assert all ByteBuffer in the list.

mikemccand: Awesome, thanks, looks great @dungba88!

mikemccand (approved):
Thanks @dungba88 -- I plan to merge some time today.

dungba88:
Thanks @mikemccand for merging

PR #12880, opened 2023-12-06
Remove some redundant modifiers from code

shubhamvishu:
### Description

This PR removes(or cleans up) some of the redundant modifiers from the code. Addresses the below warning in my IDE :

- Modifier ‘static’ is redundant for inner interfaces
- Modifier ‘public’ is redundant for interface members
- Modifier 'abstract' is redundant for interface methods
- Modifier ‘private’ is redundant for enum constructors

### DIFF
--- a/lucene/analysis/common/src/test/org/apache/lucene/analysis/compound/TestCompoundWordTokenFilter.java
+++ b/lucene/analysis/common/src/test/org/apache/lucene/analysis/compound/TestCompoundWordTokenFilter.java
@@ -528,7 +528,7 @@ public class TestCompoundWordTokenFilter extends BaseTokenStreamTestCase {
     assertTokenStreamContents(tf8, new String[] {"fußball"});
   }
 
-  public static interface MockRetainAttribute extends Attribute {
+  public interface MockRetainAttribute extends Attribute {
     void setRetain(boolean attr);
 
     boolean getRetain();
--- a/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
+++ b/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
@@ -3216,7 +3216,7 @@ public final class CheckIndex implements Closeable {
   }
 
   @FunctionalInterface
-  private static interface DocValuesIteratorSupplier {
+  private interface DocValuesIteratorSupplier {
     DocValuesIterator get(FieldInfo fi) throws IOException;
   }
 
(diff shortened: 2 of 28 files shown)


gsmiller:
Thanks @shubhamvishu! I personally like this cleanup, but I'd be curious if others have some reasons I'm not aware of why they may prefer differently? Maybe more importantly though, I wonder what this looks like going forward. It's fine to do a cleanup now, but will these things creep back in? It would be nice if a tool like "spotless" could handle this automatically (assuming the community agrees with the convention), but I'm not sure if it's capable of that.

As for this one-off cleanup, I'm fine with it but would like to leave it open for a couple days to see if others have more general issues with these changes. Things our IDEs flag don't always align 100% with our conventions.

dweiss:
I don't think we have any conventions for this. And I don't know if tools can automate this. 

I remember reading a thread somewhere on the openjdk mailing list that touched on a similar topic - whether to "clean up" existing codebase now that IDEs have an absolute knowledge of all language quirks. The response was: it's not necessarily a great idea - makes backporting to historical branches more difficult, makes reading the code more difficult for people not _that_ familiar with the language spec... and there's the human factor - I personally prefer to have "public static" redundancy in interface definitions, even if it's not required. I don't have a strong opinion though, we may clean this up - I don't mind./

shubhamvishu:
Thanks @gsmiller @dweiss for taking a look. I see some great points raised here by both of you(and I agree to all). A couple of points why I think we should go with this change :

- Backporting : AFAIK backporting this particular change at least shouldn't be a problem because we are only removing the redundant modifiers here which are not adding any value(maybe other folks might know better if i'm missing something). This was one of the reason I thought this might be quick code cleanup.

- Less cognitive load : I think unnecessarily adding these modifiers increases the cognitive load the reader goes through while reading code. Sorry if I'm being subjective but I feel adding public to interface methods for eg just adds extra complexity/confusion and my perception is this particular change simplifies that but Idk if its the opposite for any folks who are not much familiar with the language.

- Consistency : I agree some might prefer to have "public static" redundancy as Dawid mentioned but I think should we try to have the same convention to other interfaces as well(currently its a mix of both which I believe only contributes to adding more unnecessary confusion/complexity to code). Maybe we could try to stick to one whichever community feel would be better(like we try to do with false conditionals, `x == false` and not `!x`) and this might be a good start. 

Though I understand its difficult to enforce such things in fast growing open source packages and not possible or at least not a good idea to automate as mentioned by Dawid but we should try to cleanup if we see some opportunity to simply things. If we believe its ok to leave it as is on the individual preference, I'm completely fine with that too but personally I'd be in favour of this one(if not the general cleanup changes which involve other complexities sometimes).

gsmiller:
Went ahead and merged this on to main/branch_9x since I saw some support for this and no real blocking concerns.

PR #12881, opened 2023-12-06
Performance improvements to MatchHighlighter and MatchRegionRetriever

dweiss:
This patch provides a number of small improvements aimed at improving performance of MatchHighlighter (and MatchRegionRetriever), especially in corner cases like:

* queries that result in a large number of hits, especially in long fields. This causes hit passage scoring to be time-consuming, only to result in a few "best" passages. A configurable `maxHitsPerField` limit is added to allow capping the number of matches retrieved (and scored) to a reasonable number.

* queries that require highlighting of hundreds of documents. The major performance bottleneck here is field value loading. MatchRegionRetriever now allows specifying which fields to load unconditionally, as well as filtering fields that contain hits (to skip computing highlights for fields which are used for filtering or are never displayed). Another improvement here is that highlights are now computed in parallel (using index searcher's task executor).

There are a few minor API changes so I think it should be targeted for 10.

### DIFF
--- a/lucene/highlighter/src/java/org/apache/lucene/search/matchhighlight/MatchRegionRetriever.java
+++ b/lucene/highlighter/src/java/org/apache/lucene/search/matchhighlight/MatchRegionRetriever.java
@@ -39,6 +39,7 @@ import org.apache.lucene.index.LeafReader;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.index.StoredFieldVisitor;
 import org.apache.lucene.index.StoredFields;
+import org.apache.lucene.search.FilterMatchesIterator;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.Matches;
 import org.apache.lucene.search.MatchesIterator;
@@ -106,7 +107,9 @@ public class MatchRegionRetriever {
    * This constructor uses the default offset strategy supplier from {@link
    * #computeOffsetRetrievalStrategies(IndexReader, Analyzer)}.
    *
-   * @param searcher The {@link IndexSearcher} to use for executing the query.
+   * @param searcher The {@link IndexSearcher} used to execute the query. The index searcher's
+   *     {@linkplain IndexSearcher#getTaskExecutor() task executor} is also used for computing
+   *     highlights concurrently.
    * @param query The query for which highlights should be returned.
    * @param analyzer An analyzer that may be used to reprocess (retokenize) document fields in the
    *     absence of position offsets in the index. Note that the analyzer must return tokens
@@ -139,7 +142,9 @@ public class MatchRegionRetriever {
   }
 
   /**
-   * @param searcher The {@link IndexSearcher} to use for executing the query.
+   * @param searcher The {@link IndexSearcher} used to execute the query. The index searcher's
+   *     {@linkplain IndexSearcher#getTaskExecutor() task executor} is also used for computing
+   *     highlights concurrently.
    * @param query The query for which matches should be retrieved. The query should be rewritten
    *     against the provided searcher.
    * @param fieldOffsetStrategySupplier A custom supplier of per-field {@link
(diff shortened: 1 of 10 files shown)


lucene/highlighter/src/java/org/apache/lucene/search/matchhighlight/MatchRegionRetriever.java
```diff
@@ -53,99 +58,173 @@
...
   public void highlightDocuments(TopDocs topDocs, MatchOffsetsConsumer consumer)
       throws IOException {
     highlightDocuments(
         Arrays.stream(topDocs.scoreDocs).mapToInt(scoreDoc -> scoreDoc.doc).sorted().iterator(),
-        consumer);
+        consumer,
+        field -> Integer.MAX_VALUE);
+  }
+
+  static class DocHighlightData {
```
romseygeek, line 209: If this is for 10 only you can use a `record` here I think?

lucene/highlighter/src/java/org/apache/lucene/search/matchhighlight/MatchRegionRetriever.java
```diff
@@ -53,99 +58,173 @@
...
+     */
+    List<String> getValues(String field);
   }
 
   /**
-   * A constructor with the default offset strategy supplier.
+   * This constructor uses the default offset strategy supplier from {@link
+   * #computeOffsetRetrievalStrategies(IndexReader, Analyzer)}.
    *
+   * @param searcher The {@link IndexSearcher} to use for executing the query.
```
romseygeek, line 109: I wonder if its worth calling out that we'll use the IndexSearcher's task executor to try and parallelize the work here?  It may be that in some applications you don't want highlighting to use threads that could otherwise be running simple queries, in which case clients should use different IndexSearcher instances for querying and highlighting.

romseygeek (approved):
A couple of questions but LGTM otherwise!

lucene/highlighter/src/java/org/apache/lucene/search/matchhighlight/MatchRegionRetriever.java
dweiss: I actually had the system's fork join executor, which is easier to work with (submit/ consume)... but this causes some headaches with tests (runaway system threads) and of course there's no control over how many threads are in use, etc. So I opted to use the default IndexSearcher instead. You're right that this may be problematic - I'll clarify the javadocs. It's a pity we don't know the actual parallelism of the executor - it'd make things easier to work with.

lucene/highlighter/src/java/org/apache/lucene/search/matchhighlight/MatchRegionRetriever.java
```diff
@@ -199,34 +343,95 @@ public void highlightDocument(
...
+
+        List<OffsetRange> ranges = offsetStrategy.get(matchesIterator, doc);
+        if (!ranges.isEmpty()) {
+          outputHighlights.put(field, ranges);
+        }
+      }
+    }
+  }
+
+  private static class MatchesIteratorWithLimit implements MatchesIterator {
```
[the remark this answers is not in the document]
dweiss: Will correct it, thanks Alan.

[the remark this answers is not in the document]
dweiss: Done!

lucene/highlighter/src/java/org/apache/lucene/search/matchhighlight/MatchRegionRetriever.java
dweiss: Added more documentation, hopefully the wording is clear. These are rather low-level utilities but they're very helpful in getting better highlighting performance.

dweiss:
Thanks, @romseygeek !

PR #12885, opened 2023-12-07
Fix for the bug where JapaneseReadingFormFilter cannot convert some hiragana to romaji

kuramitsu:
### Description
I found a bug in using JapaneseReadingFormFilter that some hiragana are not converted to romaji.
(For example, "ぐ" does not become "gu". I noticed this because "マスキング" did not get any hits when searching for "ますきんぐ".)
I believe this is due to the existence of hiragana whose readings are not explicitly defined in the kuromoji dictionary.

### Draft
For hiragana, which is an OOV term, how about treating its conversion to katakana as a reading?


msfroh:
This is really interesting. It looks like the filter logic is already trying to conversion to katakana before converting to romaji.

Specifically in https://github.com/apache/lucene/blob/main/lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseReadingFormFilter.java#L49-L63, my take is:

1. The `reading` variable gets populated with (what I understand should be) the katakana reading (which gets returned if `useRomaji` is `false`).
2. Assuming that reading is populated and `useRomaji` is `true`, then we convert the katakana to romaji.

So, I'm wondering if maybe there's a bug in `JaMorphData.getReading()` implementation? It looks like there's already supposed to be a hiragana -> katakana shift here: https://github.com/apache/lucene/blob/main/lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/dict/TokenInfoMorphData.java#L115

(This is my first time reading this code, back by my very limited, Duolingo-based knowledge of Japanese, so I might be wrong.)

kuramitsu:
@msfroh 
Thank you for your confirmation.
As you said, I think it would be better to check and fix the bug in the JaMorphData.getReading() part.
I will try to check and fix it.

kuramitsu:
I found that one of the implementations of getReading, UnknownMorphData.getReading(), always returns null.
Since the problem seems to be that termAttr is used as it is for OOV Term, I will try to insert a process to convert hiragana into katakana.

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseReadingFormFilter.java
```diff
@@ -43,10 +43,30 @@ public JapaneseReadingFormFilter(TokenStream input) {
...
+      }
+    }
+    return true;
+  }
+
   @Override
   public boolean incrementToken() throws IOException {
     if (input.incrementToken()) {
       String reading = readingAttr.getReading();
+      if (reading == null & isAllHiragana(termAttr)) {
```
zhaih, line 61: Maybe use `&&` instead?

zhaih, line 61: What if we have hiragana and katagana mixed text? I feel like we still need to convert the hiragana part to katagana?

kuramitsu: Thank you. I fixed it.

kuramitsu: Fixed to support partial hiragana. 
However, it is difficult to describe a test for oov terms that contain a mixture of hiragana and katakana, because the system seems to basically split different character types for oov terms.

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseReadingFormFilter.java
```diff
@@ -43,10 +43,38 @@ public JapaneseReadingFormFilter(TokenStream input) {
     this(input, false);
   }
 
+  private boolean isHiragana(char ch) {
+    return ch >= 0x3041 && ch <= 0x3096;
```
dungba88, line 47: I think these can be extracted to constants for readability, e.g HIRAGANA_START, HIRAGANA_END

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseReadingFormFilter.java
```diff
@@ -43,10 +43,38 @@ public JapaneseReadingFormFilter(TokenStream input) {
...
+  }
+
+  private boolean containsHiragana(CharSequence s) {
+    int len = s.length();
+    for (int i = 0; i < len; i++) {
```
dungba88, line 52: I think we can just use `i < s.length()`. It's also computed only once, and don't need an additional temporary variable.

lucene/analysis/kuromoji/src/test/org/apache/lucene/analysis/ja/TestJapaneseReadingFormFilter.java
```diff
@@ -88,6 +88,11 @@ protected TokenStreamComponents createComponents(String fieldName) {
     a.close();
   }
 
+  public void testKatakanaReadingsHiragana() throws IOException {
+    assertAnalyzesTo(
+        katakanaAnalyzer, "が ぎ ぐ げ ご ぁ ゔ", new String[] {"ガ", "ギ", "グ", "ゲ", "ゴ", "ァ", "ヴ"});
```
dungba88, line 93: I think we could add some test for half-width form as well?

dungba88, line 93: I'm wondering what would happen if these characters are in decomposed form? It seems like the か part will be converted to カ but the dakuten is not. Maybe we could have test to see how it looks like?

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseReadingFormFilter.java
kuramitsu: Thank you. I fixed it.
https://github.com/apache/lucene/pull/12885/commits/9ad2cbe113a3164109cd90181dfb7fa598207c82

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseReadingFormFilter.java
kuramitsu: Thank you. I fixed it.
https://github.com/apache/lucene/pull/12885/commits/9773668a763d48869fc8482479d82bfa1a594c43

lucene/analysis/kuromoji/src/test/org/apache/lucene/analysis/ja/TestJapaneseReadingFormFilter.java
kuramitsu: Checking the getRomanization function in ToStringUtil.java, it does not seem to support split muddle even in katakana. I am sorry, but please let this matter be a separate issue.
https://github.com/apache/lucene/blob/main/lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/dict/ToStringUtil.java#L246

dungba88 (approved):

zhaih (approved):
LGTM, I'm not professional on JP analyses but this seems right to me.

zhaih:
Thank you bot I obviously forgot to merge this one. @kuramitsu could you please add an CHANGES.txt entry under Lucene 9.10?

kuramitsu:
@zhaih 
> could you please add an CHANGES.txt entry under Lucene 9.10?

Thank you. I added it.

zhaih:
Merged and backported, thanks for the contribution @kuramitsu !

PR #12887, opened 2023-12-07
Add test that tickles a JVM JIT crash on JDK's less than 21.0.1

ChrisHegarty:
Add test that tickles a JVM JIT crash on JDK's less than 21.0.1

Work in Progress. Trying to repro a C2 crash, for now 

[This PR is not intended to be commit. At least not as it is.]

### DIFF
--- a/gradle/testing/defaults-tests.gradle
+++ b/gradle/testing/defaults-tests.gradle
@@ -51,7 +51,7 @@ allprojects {
            includeInReproLine: false
           ],
           [propName: 'tests.jvmargs',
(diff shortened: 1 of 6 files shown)


gradle/testing/defaults-tests.gradle
```diff
@@ -51,7 +51,7 @@ allprojects {
            includeInReproLine: false
           ],
           [propName: 'tests.jvmargs',
-           value: { -> propertyOrEnvOrDefault("tests.jvmargs", "TEST_JVM_ARGS", isCIBuild ? "" : "-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1") },
```
dweiss, line 54: These defaults were added on purpose to speed up test runs - was this change intentional?

PR #12753, opened 2023-11-03
Bump minimum required Java version to 21

ChrisHegarty:
With all prerequisites complete, this PR bumps the minimum Java release version to Java 21.

~This PR is intended to help tease out potential issues that may arise from compiling with JDK 21. We can use it to identify and pick out the individual issues. So far we know:~

1. ~~we use deprecated java.net URL constructors - usage should likely be replaced with URI::toURL
    Can proceed independent of the version bump~~  ✅ 

2. we use deprecated java.util Locale constructor - ~~usage should likely be replaced with Locale:of
    Locale:of factories are added in Java 19, so this kinda the change to the version bump. Maybe evaluate using Locale.Builder~~ Replace with Locale.Builder ✅ 

3. ~Refactor usage of runFinalization~  ✅ 

4. ~we need a new yet-to-be released ecj jdt - but it is good to validate the unreleased snapshot 
    Await release of 3.36.0~ ✅

PR #12888, opened 2023-12-07
Upgrade ECJ to 3.36.0

ChrisHegarty:
This commit upgrades ECJ to 3.36.0, as it has support for more recent Java versions, like Java 21.

The motivation for this change is to better allow build and testing with more recent JDK.s


rmuir (approved):

ChrisHegarty:
relates: #12753

Issue #11041, opened 2021-06-14
Remove IndexSearcher#search(Query,Collector) in favor of IndexSearcher#search(Query,CollectorManager) [LUCENE-10002]


asfimport:
It's a bit trappy that you can create an IndexSearcher with an executor, but that it would always search on the caller thread when calling `IndexSearcher#search(Query,Collector)`.

 Let's remove `IndexSearcher#search(Query,Collector)`, point our users to `IndexSearcher#search(Query,CollectorManager)` instead, and change factory methods of our main collectors (e.g. `TopScoreDocCollector#create`) to return a `CollectorManager` instead of a `Collector`?

---
Migrated from [LUCENE-10002](https://issues.apache.org/jira/browse/LUCENE-10002) by Adrien Grand (@jpountz), updated Apr 06 2022
Pull requests: https://github.com/apache/lucene/pull/240

wardle:
Hi all. There is a small overhead with using CollectorManager over Collector. In my own usage, I have a read-only index which uses only a single slice, and I've chosen not to provide an executor service because my usage uses lots of cores for concurrent requests.

Consequently, it seems to be that this is more of an issue of documentation, and best practices. Additional usage information in IndexSearcher#search to point users to use a CollectorManager might well suffice, and prevent what is presumably a concern about incorrect usage, while permitting users who know the nature of their index, and their usage patterns, to leverage the APIs as they need. 

Clearly, it is trivial to use CollectorManager and wrap my existing (custom) Collector, but it is an unnecessary overhead, albeit a very small one, but at the cost of losing backwards compatibility and with little gain as fundamentally both could co-exist without complication, but documentation steering most users to prefer CollectorManager.

javanna:
Only the actual removal of the deprecated search(Query, Collector) is left to do. That is tracked by #12892 hence this issue can be closed/

PR #12890, opened 2023-12-07
[WIP] LUCENE-10002: Deprecate FacetsCollector#search helper methods as they internally use IndexSearcher#search(Query, Collector) API

zacharymorn:
This is a WIP / discussion only PR. 

As the next step for https://github.com/apache/lucene/issues/11041, I'm thinking to deprecate FacetsCollector#search methods as they internally use IndexSearcher#search(Query, Collector) API. However, I'm a bit torn on whether we should create new replacement methods for them, or could just deprecate them without replacement, since the logic looks like something application developers could naturally come up with as well (and I don't like creating an extra `Pair` class just for returning multiple values for these helper methods)? 

Please let me know if you have any suggestion on this @mikemccand @jpountz @javanna  @gsmiller.

### DIFF
--- a/lucene/facet/src/java/org/apache/lucene/facet/FacetUtils.java
+++ b/lucene/facet/src/java/org/apache/lucene/facet/FacetUtils.java
@@ -18,7 +18,25 @@
 package org.apache.lucene.facet;
 
 import java.io.IOException;
+import java.util.Objects;
+
+import org.apache.lucene.search.Collector;
+import org.apache.lucene.search.CollectorManager;
 import org.apache.lucene.search.DocIdSetIterator;
+import org.apache.lucene.search.FieldDoc;
+import org.apache.lucene.search.IndexSearcher;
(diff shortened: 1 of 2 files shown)


gsmiller:
IMO we should deprecate these without replacement. I agree that users should be able to implement this logic pretty easily in their application layer, and would probably be better suited to do so. For example, if a user doesn't want to collect actual hits and just do faceting, these methods are a little trappy since they internally decide to also do a full (accurate) count of hits. They're baking in some assumptions like this that users may not expect.

Just my opinion of course. I don't feel all that strongly about it, and if we want some convenience functionality somewhere for this, that's fine.

mikemccand:
Yeah I agree it's OK to deprecate without replacement, but maybe in the deprecated javadocs (and in `MIGRATE.txt` for 10.0) add a short explanation about using `MultiCollectorManager` for this?

PR #240, opened 2021-08-12
LUCENE-10002: Deprecate IndexSearch#search(Query, Collector) in favor of IndexSearcher#search(Query, CollectorManager) - TopFieldCollectorManager & TopScoreDocCollectorManager

zacharymorn:
# Description / Solution
Deprecate IndexSearch#search(Query, Collector) in favor of IndexSearcher#search(Query, CollectorManager), with the following changes:
1. Refactor out TopScoreDocCollectorManager, TopFieldCollectorManager, ~~TotalHitCountCollectorManager and FixedBitSetCollector~~
2. Refactor some tests to use the above collector manager
~~3. Refactor DrillSideways to use extracted out collector managers~~

#11041

# Tests
Passed updated tests with `./gradlew clean; ./gradlew check -Ptests.nightly=true` (currently there are nocommit messages that are failing precommit though)

# Checklist

Please review the following and check all that apply:

- [x] I have reviewed the guidelines for [How to Contribute](https://wiki.apache.org/lucene/HowToContribute) and my code conforms to the standards described there to the best of my ability.
- [x] I have created a Jira issue and added the issue ID to my pull request title.
- [x] I have given Lucene maintainers [access](https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [x] I have developed this patch against the `main` branch.
- [x] I have run `./gradlew check`.
- [ ] I have added tests for my changes.

PR #12891, opened 2023-12-07
[Backporting] LUCENE-10002: Deprecate IndexSearch#search(Query, Collector) in favor of IndexSearcher#search(Query, CollectorManager) - TopFieldCollectorManager & TopScoreDocCollectorManager

zacharymorn:
This PR backports https://github.com/apache/lucene/pull/240 to `branch_9x`


mikemccand (approved):
Thank you @zacharymorn!

javanna:
Thanks for merging this @zacharymorn . I am updating Elasticsearch to this change and I came to wonder if we should deprecate the static methods that create collectors. I do agree on exposing top docs collector managers and deprecating the createShared methods, but there are still scenarios in which users create collectors where collector managers are not needed, for instance situations where IndexSearcher#search is not even called. 

I wanted to double check what folks are thinking around this. The plan isn't to deprecate direct usage of collectors, correct? I wonder if it is excessive to to now have to create a collector manager to get an instance of a collector manager as follows:

```
  @Deprecated
  public static TopScoreDocCollector create(int numHits, int totalHitsThreshold) {
    return new TopScoreDocCollectorManager(numHits, null, totalHitsThreshold, false).newCollector();
  }
  ```

This method is now deprecated in favour of interacting with Collector managers rather than collectors. I'd like to double check this we are all aligned on this direction.

PR #12893, opened 2023-12-08
Removing @lucene.experimental tags in testXXX methods in CheckIndex

slow-J:
Following up on @mikemccand's comment in previous CheckIndex PR:https://github.com/apache/lucene/pull/12876 related to my question of cleaning up the >10 year old @lucene.experimental tags in CheckIndex.

>  I do think some of these tags in CheckIndex.java could be removed, e.g. on each of the testXXX methods? But likely we should leave it on each of the XXXStatus returned classes, so we are free to break/change what metadata these classes return? And let's do it separately -- this PR looks like a great cleanup already. Progress not perfection!

Targetting Lucene 9.10.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -181,6 +181,8 @@ Other
 
 * GITHUB#11023: Removing some dead code in CheckIndex. (Jakub Slowinski)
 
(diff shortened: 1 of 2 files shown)


mikemccand (approved):
Thanks @slow-J!

PR #12897, opened 2023-12-09
Output well-formed UTF-8 bytes in SimpleTextCodec's segmentinfos

msfroh:
### Description

The SimpleTextSegmentInfoFormat was writing the random byte array used as a segment's ID directly -- not converting to a simple text representation of the byte array. As a result, the segment infos were often malformed (as UTF-8 text).

The included test was failing before the change to write out the text representation of the byte array.

### DIFF
--- a/lucene/codecs/src/test/org/apache/lucene/codecs/simpletext/TestSimpleTextSegmentInfoFormat.java
+++ b/lucene/codecs/src/test/org/apache/lucene/codecs/simpletext/TestSimpleTextSegmentInfoFormat.java
@@ -17,16 +17,16 @@
 package org.apache.lucene.codecs.simpletext;
 
 import java.io.IOException;
-import java.nio.ByteBuffer;
-import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import org.apache.lucene.codecs.Codec;
 import org.apache.lucene.index.IndexFileNames;
 import org.apache.lucene.index.SegmentInfo;
+import org.apache.lucene.index.Term;
 import org.apache.lucene.store.ChecksumIndexInput;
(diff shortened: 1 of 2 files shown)


msfroh:
If needed, I'm happy to add versions of `testFileIsUTF8()` for the other SimpleTextCodec format unit tests.

lucene/codecs/src/test/org/apache/lucene/codecs/simpletext/TestSimpleTextSegmentInfoFormat.java
```diff
@@ -33,4 +43,39 @@ protected Version[] getVersions() {
...
+        IndexFileNames.segmentFileName("_123", "", SimpleTextSegmentInfoFormat.SI_EXTENSION);
+    try (ChecksumIndexInput input = dir.openChecksumInput(segFileName)) {
+      long length = input.length();
+      if (length > 5_000) {
+        // Avoid allocating a huge array if the length is wrong
+        fail("SegmentInfos should not be this large");
+      }
+      byte[] bytes = new byte[(int) length];
+      input.readBytes(bytes, 0, bytes.length);
+      StandardCharsets.UTF_8.newDecoder().decode(ByteBuffer.wrap(bytes));
```
jpountz, line 77: Maybe use `Term#toString(BytesRef)` which includes such checks already and configures a few more flags on the decoder that look useful?

jpountz (commented):
The change looks good, I just left a minor suggestion about the test.

msfroh:
I implemented a similar change for binary doc values at https://github.com/apache/lucene/pull/12987

Issue #12895, opened 2023-12-08
Corruption read on term dictionaries in Lucene 9.9


benwtrent:
### Description

It seems that https://github.com/apache/lucene/pull/12699/ has inadvertantly broken reading term dictionaries created in Lucene 9.8<=.

To replicate a bug, one can index wikibigall with LuceneUtil & Lucene 9.8 & force-merge.

Then attempt to read the created index using a wildcard query:

```
    Path path = Paths.get("/data/local/lucene/indices/wikibigall.lucene-main.opt.Lucene90.dvfields.nd6.72652M/index");
    try (FSDirectory dir = FSDirectory.open(path);
        DirectoryReader reader = DirectoryReader.open(dir)) {
      IndexSearcher searcher = new IndexSearcher(reader); 
      searcher.count(new WildcardQuery(new Term("body", "*fo*")));
    }
```

This will result in a trace similar to below:
```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: Index 3 out of bounds for length 3
	at org.apache.lucene.store.ByteArrayDataInput.readByte(ByteArrayDataInput.java:136)
	at org.apache.lucene.store.DataInput.readVInt(DataInput.java:110)
	at org.apache.lucene.store.ByteArrayDataInput.readVInt(ByteArrayDataInput.java:114)
	at org.apache.lucene.codecs.lucene90.blocktree.IntersectTermsEnumFrame.load(IntersectTermsEnumFrame.java:158)
	at org.apache.lucene.codecs.lucene90.blocktree.IntersectTermsEnumFrame.load(IntersectTermsEnumFrame.java:149)
	at org.apache.lucene.codecs.lucene90.blocktree.IntersectTermsEnum.pushFrame(IntersectTermsEnum.java:203)
	at org.apache.lucene.codecs.lucene90.blocktree.IntersectTermsEnum._next(IntersectTermsEnum.java:531)
	at org.apache.lucene.codecs.lucene90.blocktree.IntersectTermsEnum.next(IntersectTermsEnum.java:373)
	at org.apache.lucene.search.MultiTermQueryConstantScoreBlendedWrapper$1.rewriteInner(MultiTermQueryConstantScoreBlendedWrapper.java:111)
	at org.apache.lucene.search.AbstractMultiTermQueryConstantScoreWrapper$RewritingWeight.rewrite(AbstractMultiTermQueryConstantScoreWrapper.java:179)
[...]

PR #12900, opened 2023-12-10
Push and pop OutputAccumulator as IntersectTermsEnumFrames are pushed and popped

gf2121:
Relate to : https://github.com/apache/lucene/issues/12895


mikemccand:
OK I opened #12901 to create a better BWC test revealing these read-time exceptions -- can we do that, first, and then confirm #12900 indeed fixes it, then push the test, then push the fix?  I can try to work on creating that test case, but won't have much time until early tomorrow AM.  If anyone else wants to crack it, feel free!

mikemccand:
I've confirmed the new (failing) BWC test from #12901 now passes with this PR.  I'll review ...

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestBackwardsCompatibility.java
```diff
@@ -2245,7 +2244,6 @@ public void testFailOpenOldIndex() throws IOException {
   // #12895: test on a carefully crafted 9.8.0 index (from a small contiguous subset
   // of wikibigall unique terms) that shows the read-time exception of
   // IntersectTermsEnum (used by WildcardQuery)
-  @Ignore("re-enable once we merge #12900")
```
mikemccand, line 2248: Yay :)

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/IntersectTermsEnumFrame.java
```diff
@@ -89,6 +89,9 @@ final class IntersectTermsEnumFrame {
 
   final ByteArrayDataInput bytesReader = new ByteArrayDataInput();
 
+  // Cumulative outputs so far
+  BytesRef[] outputPrefix;
```
mikemccand, line 93: It's sort of strange to have a separate array for tracking the accumulated `BytesRef` outputs when this is the purpose of the `outputAccumulator`?  Wouldn't we be able to push/pop as we push/pop the frames, into the single accumulator, so we don't need make extra arrays as we push deeper into the FST?

We could assert when we pop that the output we are popping exactly matches (`.equals()`) the `FST.Arc`'s output from that frame.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/IntersectTermsEnum.java
```diff
@@ -198,6 +204,7 @@ private IntersectTermsEnumFrame pushFrame(int state) throws IOException {
     }
 
     f.arc = arc;
+    f.outputPrefix = outputAccumulator.bytesRefs();
```
mikemccand, line 207: Hmm shouldn't we sometimes set this `outputPrefix` back to `null`?  The frames are re-used, so we could push deep into the FST, where there is an `outputPrefix`, then pop back out and push into a new part of the FST that does not have an `outputPrefix` and illegally share / use the old leftover stale `outputPrefix`?

mikemccand (commented):
Thanks @gf2121!  I left a few head scratching comments ...

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/IntersectTermsEnum.java
gf2121: > Hmm shouldn't we sometimes set this outputPrefix back to null

We never set `outputPrefix` back to null. But a `BytesRef[0]` here is equivalent, which means all prefix output (including current frame) are all `NO_OUTPUT`. 

> push into a new part of the FST that does not have an outputPrefix and illegally share / use the old leftover stale outputPrefix?

I thought the `outputPrefix` will be always overwritten by `pushFrame` when we push into a new part of the FST. Did i miss something ?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/IntersectTermsEnumFrame.java
gf2121: > Wouldn't we be able to push/pop as we push/pop the frames, into the single accumulator, so we don't need make extra arrays as we push deeper into the FST? 

I'm not sure. Looking at `pushFrame`, it seems like one single frame could be related to more than one `FST#Arc`. Maybe we need to record the number of outputs pushed into the accumulator when pushing a frame ?

gf2121: I like this idea to avoid the array copy. Implemented in [e387ec9](https://github.com/apache/lucene/pull/12900/commits/e387ec93e92e5a83505419f88e261bc31fb9ab85).

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
```diff
@@ -495,7 +495,7 @@ public boolean seekExact(BytesRef target) throws IOException {
       targetUpto = 0;
       outputAccumulator.push(arc.nextFinalOutput());
       currentFrame = pushFrame(arc, 0);
-      outputAccumulator.pop();
+      outputAccumulator.pop(arc.nextFinalOutput());
```
gf2121, line 498: This seems like another potential bug : If `arc.nextFinalOutput() == NO_OUTPUT`, we will not add the output into 
 the accumulator but we pop one out.

Not sure why we have not seen any bug report for this. Maybe our output format protect it implicitly? This is in `SegmentTermsEnum` so it could possibly affect merge. I'll dig to see if it is a real bug.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/IntersectTermsEnum.java
mikemccand: > I thought the `outputPrefix` will be always overwritten by `pushFrame` when we push into a new part of the FST. Did i miss something ?

Ahh OK you are right -- and we could simply rewrite to empty `BytesRef` array which means the same thing as `null`.  Good.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/IntersectTermsEnum.java
```diff
@@ -198,9 +199,11 @@ private IntersectTermsEnumFrame pushFrame(int state) throws IOException {
     }
 
     f.arc = arc;
+    f.outputNum = outputAccumulator.outputCount() - initOutputCount;
```
mikemccand, line 202: OK we need this because a single frame might traverse multiple FST arcs, so we need to accumulate possibly / often more than one output.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
mikemccand: Good catch!  And the new check you added in `pop` will prevent this bug, right?

But I also think this was (luckily!?) not a bug, because this case is when the `STE` is not yet seek'd to any term, so we are starting at the FST entry arc / root block, and the empty string FST input must always have a final output (the "root block"), which will never be `NO_OUTPUT`.  Are there any other places in `STE` where we were blindly popping like this?

mikemccand: It looks like the other places where we "blindly" pop are fine -- we conditionalize by `arc.isFinal()` check, and we know FST for BlockTree terms index will always have a `nextFinalOutput` on final nodes since it holds the floor data for that terms block.  And there there is one other place in `seekCeil` where we also rely on empty string having a final output, which is fine.  So net/net I don't think this was a bug, but it smelled close to one!

mikemccand (approved):
This looks good to me!  I like the added safety of confirming on `pop` that we are in fact `pop`ing what we think we are (except when we pop multiple items).  And it's cleaner than the first cut by simple using the output accumulator consistently.

I think we should merge this to `main`, `9x` and `99x` and let's let CI chew on it for a bit (day or so) before cutting 9.9.1?

gf2121:
Thanks for review and great advice @mikemccand !

> I think we should merge this to main, 9x and 99x and let's let CI chew on it for a bit (day or so) before cutting 9.9.1?

+1. I'll merge and backport soon.

PR #12903, opened 2023-12-10
[branch_9_9] Reflow computeCommonPrefixLengthAndBuildHistogram to avoid crash

### LINKED ISSUE #12898: JVM SIGSEGV crash when compiling computeCommonPrefixLengthAndBuildHistogram Lucene 9.9.0
We're running in a JVM JIT bug (leading to a crash) when running Lucene 9.9.0.

The crash happens with (at least) all JDK releases between 19 and 21.0.1. There is no crash in the currently unreleased JDK 21.0.2 and JDK 22. The JDK issue tracking this is https://bugs.openjdk.org/browse/JDK-8321370.

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x0000000103cf75a4, pid=83290, tid=30211
#
# JRE version: OpenJDK Runtime Environment (21.0.1+12) (build 21.0.1+12-29)
# Java VM: OpenJDK 64-Bit Server VM (21.0.1+12-29, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, bsd-aarch64)
# Problematic frame:
# V  [libjvm.dylib+0x74f5a4]  PhaseIdealLoop::build_loop_late_post_work(Node*, bool)+0x30c
#
# No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /Users/chegar/git/lucene/lucene/core/build/tmp/tests-cwd/hs_err_pid83290.log
#
# Compiler replay data is saved as:
# /Users/chegar/git/lucene/lucene/core/build/tmp/tests-cwd/replay_pid83290.log
#
# If you would like to submit a bug report, please visit:
#   https://bugreport.java.com/bugreport/crash.jsp
#
```

```
  // This simple test tickles a JVM JIT crash on JDK's less than 21.0.1
  // Needs to be run with C2, so with the environment variable `CI` set
  public void testCrash() throws IOException {
    for (int i = 0; i < 100; i++) {
      try (Directory dir = newDirectory();
          IndexWriter iw = new IndexWriter(dir, newIndexWriterConfig(null))) {
        for (int d = 0; d < 20_000; d++) {
          iw.addDocument(
              List.of(new IntPoint("bar", 0), new SortedNumericDocValuesField("foo", 0)));
        }
      }
    }
  }
```

Since running into this with Elasticsearch we've excluded the offending two methods from compilation, see https://github.com/elastic/elasticsearch/pull/103112 and https://github.com/elastic/elasticsearch/issues/103004

ChrisHegarty:
This commit reflows the code in the method body of `computeCommonPrefixLengthAndBuildHistogram`, so as to avoid a JVM JIT crash. The purpose of this change is to workaround the JVM bug which is somewhat fragile, but the best that we can do for now and appears to be working well. Further details of the JDK issue can be found in https://bugs.openjdk.org/browse/JDK-8321370 and linked issues.

The newly added test fails around 9 out of 10 times for me, before the fix. With the fix the test has not been seen to fail in several hundreds of runs. Though, further test runs are on going. 

The JDK issue is already fixed in the currently unreleased JDK 21.0.2 and JDK 22, but we need this workaround to avoid the crash on current released JDKs.

closes #12898

### DIFF
--- a/gradle/testing/randomization/policies/tests.policy
+++ b/gradle/testing/randomization/policies/tests.policy
@@ -63,6 +63,9 @@ grant {
   permission java.lang.RuntimePermission "getFileStoreAttributes";
   permission java.lang.RuntimePermission "writeFileDescriptor";
 
+  // needed to check if C2 (implied by the presence of the CI env) is enabled
+  permission java.lang.RuntimePermission "getenv.CI";
+
   // TestLockFactoriesMultiJVM opens a random port on 127.0.0.1 (port 0 = ephemeral port range):
   permission java.net.SocketPermission "127.0.0.1:0", "accept,listen,resolve";
 
--- a/lucene/core/src/java/org/apache/lucene/util/MSBRadixSorter.java
+++ b/lucene/core/src/java/org/apache/lucene/util/MSBRadixSorter.java
@@ -214,6 +214,12 @@ public abstract class MSBRadixSorter extends Sorter {
    * @see #buildHistogram
    */
   private int computeCommonPrefixLengthAndBuildHistogram(int from, int to, int k, int[] histogram) {
+    int commonPrefixLength = computeInitialCommonPrefixLength(from, k);
+    return computeCommonPrefixLengthAndBuildHistogramPart1(
+        from, to, k, histogram, commonPrefixLength);
+  }
+
+  private int computeInitialCommonPrefixLength(int from, int k) {
     final int[] commonPrefix = this.commonPrefix;
     int commonPrefixLength = Math.min(commonPrefix.length, maxLength - k);
     for (int j = 0; j < commonPrefixLength; ++j) {
@@ -224,7 +230,12 @@ public abstract class MSBRadixSorter extends Sorter {
         break;
       }
     }
+    return commonPrefixLength;
+  }
 
+  private int computeCommonPrefixLengthAndBuildHistogramPart1(
+      int from, int to, int k, int[] histogram, int commonPrefixLength) {
+    final int[] commonPrefix = this.commonPrefix;
     int i;
     outer:
     for (i = from + 1; i < to; ++i) {
@@ -239,7 +250,12 @@ public abstract class MSBRadixSorter extends Sorter {
         }
       }
     }
+    return computeCommonPrefixLengthAndBuildHistogramPart2(
+        from, to, k, histogram, commonPrefixLength, i);
+  }
 
+  private int computeCommonPrefixLengthAndBuildHistogramPart2(
+      int from, int to, int k, int[] histogram, int commonPrefixLength, int i) {
     if (i < to) {
       // the loop got broken because there is no common prefix
       assert commonPrefixLength == 0;
--- a/lucene/core/src/java/org/apache/lucene/util/MSBRadixSorter.java
+++ b/lucene/core/src/java/org/apache/lucene/util/MSBRadixSorter.java
@@ -213,12 +213,15 @@ public abstract class MSBRadixSorter extends Sorter {
    *
    * @see #buildHistogram
    */
+  // This method, and its namesakes, have been manually split to work around a JVM crash.
+  // See https://github.com/apache/lucene/issues/12898
   private int computeCommonPrefixLengthAndBuildHistogram(int from, int to, int k, int[] histogram) {
     int commonPrefixLength = computeInitialCommonPrefixLength(from, k);
     return computeCommonPrefixLengthAndBuildHistogramPart1(
         from, to, k, histogram, commonPrefixLength);
   }
 
+  // This method, and its namesakes, have been manually split to work around a JVM crash.
   private int computeInitialCommonPrefixLength(int from, int k) {
     final int[] commonPrefix = this.commonPrefix;
     int commonPrefixLength = Math.min(commonPrefix.length, maxLength - k);
@@ -233,6 +236,7 @@ public abstract class MSBRadixSorter extends Sorter {
     return commonPrefixLength;
   }
 
+  // This method, and its namesakes, have been manually split to work around a JVM crash.
   private int computeCommonPrefixLengthAndBuildHistogramPart1(
       int from, int to, int k, int[] histogram, int commonPrefixLength) {
     final int[] commonPrefix = this.commonPrefix;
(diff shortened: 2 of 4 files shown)


ChrisHegarty:
For reviewers or other interested parties. I run the test as follows:

```
export CI=true; export x=; while ./gradlew --no-build-cache cleanTest :lucene:core:test --rerun --tests "org.apache.lucene.util.bkd.TestDocIdsWriter.testCrash"; do echo $x | wc -c; export x=x$x; done
```

We need the `CI=true` because, by default in our gradle environment C2 compilation is disabled - and that is where the JIT crash happens.

lucene/core/src/test/org/apache/lucene/util/bkd/TestDocIdsWriter.java
```diff
@@ -150,4 +154,18 @@ public Relation compare(byte[] minPackedValue, byte[] maxPackedValue) {
...
+    for (int i = 0; i < 100; i++) {
+      try (Directory dir = newDirectory();
+          IndexWriter iw = new IndexWriter(dir, newIndexWriterConfig(null))) {
+        for (int d = 0; d < 20_000; d++) {
+          iw.addDocument(
+              List.of(new IntPoint("foo", 0), new SortedNumericDocValuesField("bar", 0)));
+        }
+      }
+    }
+  }
```
jpountz, line 170: How long does this test take? If it's multiple seconds, should it only run nightly?

lucene/core/src/java/org/apache/lucene/util/MSBRadixSorter.java
```diff
@@ -214,6 +214,12 @@ protected int getBucket(int i, int k) {
    * @see #buildHistogram
    */
   private int computeCommonPrefixLengthAndBuildHistogram(int from, int to, int k, int[] histogram) {
+    int commonPrefixLength = computeInitialCommonPrefixLength(from, k);
+    return computeCommonPrefixLengthAndBuildHistogramPart1(
+        from, to, k, histogram, commonPrefixLength);
+  }
+
+  private int computeInitialCommonPrefixLength(int from, int k) {
```
jpountz, line 222: Can you add a comment here and in other places that you touched, explaining that this works around a C2 compiler bug?

lucene/core/src/test/org/apache/lucene/util/bkd/TestDocIdsWriter.java
ChrisHegarty: The test takes around 3.5 seconds on my laptop. I can probably reduce the iterations. Or make the number of iterations depend on whether nightly or not? I’ll take a look.

lucene/core/src/java/org/apache/lucene/util/MSBRadixSorter.java
ChrisHegarty: Yes. I’ll add a comment to that effect.

lucene/core/src/test/org/apache/lucene/util/bkd/TestDocIdsWriter.java
```diff
@@ -150,4 +154,18 @@ public Relation compare(byte[] minPackedValue, byte[] maxPackedValue) {
     }
     dir.deleteFile("tmp");
   }
+
+  // This simple test tickles a JVM JIT crash on JDK's less than 21.0.1
+  // Needs to be run with C2, so with the environment variable `CI` set
```
dweiss, line 159: If it's triggered by the CI env. variable (and effectively c2 compilation) then perhaps we should check for System.getProperty("CI") is assumption-ignore the test if it can't be triggered?

dweiss:
> export CI=true; export x=; while ./gradlew --no-build-cache cleanTest :lucene:core:test --rerun --tests "org.apache.lucene.util.bkd.TestDocIdsWriter.testCrash"; do echo $x | wc -c; export x=x$x; done

Just a suggestion that this call to gradle should have a similar outcome and be a tad faster (it does fork multiple test JVMs after one compilation - used to run in parallel too but not anymore - gradle's constraint).
```
gradlew --no-build-cache :lucene:core:beast -Ptests.dups=100 --tests "org.apache.lucene.util.bkd.TestDocIdsWriter.testCrash"
```

More info:
https://github.com/apache/lucene/blob/main/help/tests.txt#L89-L123

ChrisHegarty:
> ```
> gradlew --no-build-cache :lucene:core:beast -Ptests.dups=100 --tests "org.apache.lucene.util.bkd.TestDocIdsWriter.testCrash"
> ```

Thanks @dweiss , that's nicer, and makes my life a little easier.

lucene/core/src/test/org/apache/lucene/util/bkd/TestDocIdsWriter.java
ChrisHegarty: Done. [16f91f6](https://github.com/apache/lucene/pull/12903/commits/16f91f6c7dc8a05b3c05de888362999aea01d4f2)

lucene/core/src/java/org/apache/lucene/util/MSBRadixSorter.java
ChrisHegarty: Done. [16f91f6](https://github.com/apache/lucene/pull/12903/commits/16f91f6c7dc8a05b3c05de888362999aea01d4f2)

ChrisHegarty:
I've addressed all comments so far.  The use of System::getenv is a little annoying, as it required a test-only permission, but it's minimal and make the test very straightforward to run.

gradle/testing/randomization/policies/tests.policy
```diff
@@ -63,6 +63,9 @@ grant {
   permission java.lang.RuntimePermission "getFileStoreAttributes";
   permission java.lang.RuntimePermission "writeFileDescriptor";
 
+  // needed to check if C2 (implied by the presence of the CI env) is enabled
+  permission java.lang.RuntimePermission "getenv.CI";
```
dweiss, line 67: Ah, right. Yes, this will be needed - sorry I didn't mention it. I think it's fine since we can't detect c2 in any other way (sun.management.compiler is probably not worth relying on).

dweiss (approved):
Thank you!

uschindler:
> I've addressed all comments so far. The use of System::getenv is a little annoying, as it required a test-only permission, but it's minimal and make the test very straightforward to run.

I made a comment in the main PR (use `Constants#IS_CLIENT_VM`, which was added for vectorization API). The `CI` env var is alos not so useful as Jenkins does not set it.

PR #12904, opened 2023-12-10
Add tests for Lucene90PostingsFormat back

jpountz:
I just noticed that the move from FOR to PFOR did all the work to make the old format (FOR) writeable, but missed keeping an instance of `BasePostingsFormatTestCase` for this format.


ChrisHegarty (approved):
LGTM. Assuming the lost of this test in the rename to xxx99Postingxxx was a mistake.

mikemccand:
> I just noticed that the move from FOR to PFOR did all the work to make the old format (FOR) writeable, but missed keeping an instance of `BasePostingsFormatTestCase` for this format.

Woops!  Thanks for catching this @jpountz.  Codec format migrations are so tricky ... I wish we had a HOWTO somewhere.

mikemccand (approved):

Issue #12901, opened 2023-12-10
Improve BWC tests to reveal #12895 and confirm its fix


mikemccand:
### Description

Spinoff from #12895 where we inadvertently introduced read-time exceptions in `MultiTermQueries` (e.g. `WildcardQuery` `*fo*`) in 9.9.0 when reading pre-9.9.0 written indices.

Our `TestBackwardsCompatibility` test should have caught this but clearly did not.

Let's:
  1. Make a test revealing #12895 
  2. Confirm the test (and all other tests) pass when applying the proposed PR fix #12900
  3. Merge the new test case
  4. Merge the fix in #12900 

Hopefully in that order?

We could simply snapshot the `wikibigall` index into BWC zip files, but, that's way too massive.  Can we instead make a randomized test that indexes fewer-ish terms and reveals the bug?  With "just so" set of terms, the bug should reproduce with a tiny index.  We could maybe iterate to "just so" by tweaking the set of terms' prefixes until the on-disk file pointers of each block share the last 6 bits of their respective pointers?

jpountz:
IMO one thing to improve here is that we changed file formats for the MSB vlong optimization by bumping the file format version number, changing the writer to write the new version and adding version checks to the reader. But we sometimes also change file formats by forking the code, like we did with `Lucene90PostingsFormat` -> `Lucene99PostingsFormat`. A benefit of the latter approach is that we keep existing tests in place, specifically `BasePostingsFormatTestCase` which has pretty good coverage, probably more than what `TestBackwardsCompatibility` will ever be able to check? We should look into more systematically forking the code when we do a file format change, or figuring out other ways to keep testing prior formats with version bumps, e.g. by retaining version checks on the write side and making sure there is a separate `BasePostingsFormatTestCase` impl for every version number?

jpountz:
> A benefit of the latter approach is that we keep existing tests in place, specifically BasePostingsFormatTestCase

Actually, it seems like we lost the test as part of refactorings, I'm adding it back in #12904.

mikemccand:
> We should look into more systematically forking the code when we do a file format change, or figuring out other ways to keep testing prior formats with version bumps, e.g. by retaining version checks on the write side and making sure there is a separate `BasePostingsFormatTestCase` impl for every version number?

+1 -- normally forking/duplicating code is a bad thing, but for BWC testing, and for code clarity and our sanity keeping track of code changes that correspond to our on-disk format changes, it is the lesser evil (to fork).

[...]

PR #12908, opened 2023-12-11
Add tests for the 9.0->9.8 block tree terms dict format back.

jpountz:
This builds on top of #12904 and adds back the write logic for the terms
dictionary format we used between 9.0 and 9.8, so that we can more easily test
it.

Relates #12901.

### DIFF
--- a/lucene/backward-codecs/src/test/org/apache/lucene/backward_codecs/lucene90/Lucene90RWPostingsFormat.java
+++ b/lucene/backward-codecs/src/test/org/apache/lucene/backward_codecs/lucene90/Lucene90RWPostingsFormat.java
@@ -75,7 +75,11 @@ public final class Lucene90RWPostingsFormat extends PostingsFormat {
     try {
       FieldsConsumer ret =
           new Lucene90BlockTreeTermsWriter(
-              state, postingsWriter, minTermBlockSize, maxTermBlockSize);
+              state,
+              postingsWriter,
+              minTermBlockSize,
+              maxTermBlockSize,
+              Lucene90BlockTreeTermsReader.VERSION_START);
       success = true;
       return ret;
     } finally {
--- a/lucene/core/src/java/org/apache/lucene/util/fst/FST.java
+++ b/lucene/core/src/java/org/apache/lucene/util/fst/FST.java
@@ -109,10 +109,19 @@ public final class FST<T> implements Accountable {
 
   // Increment version to change it
   private static final String FILE_FORMAT_NAME = "FST";
+
+  /** First supported version, this is the version that was used when releasing Lucene 7.0. */
   public static final int VERSION_START = 6;
+
   private static final int VERSION_LITTLE_ENDIAN = 8;
+
+  /** Version that started storing continuous arcs. */
   public static final int VERSION_CONTINUOUS_ARCS = 9;
+
+  /** Current version. */
   public static final int VERSION_CURRENT = VERSION_CONTINUOUS_ARCS;
+
+  /** Version that was used when releasing Lucene 9.0. */
   public static final int VERSION_90 = VERSION_LITTLE_ENDIAN;
 
   // Never serialized; just used to represent the virtual
--- a/lucene/core/src/java/org/apache/lucene/util/fst/FST.java
+++ b/lucene/core/src/java/org/apache/lucene/util/fst/FST.java
@@ -113,6 +113,9 @@ public final class FST<T> implements Accountable {
   /** First supported version, this is the version that was used when releasing Lucene 7.0. */
   public static final int VERSION_START = 6;
 
+  // Version 7 introduced direct addressing for arcs, but it's not recorded here because it doesn't
+  // need version checks on the read side, it uses new flag values on arcs instead.
+
   private static final int VERSION_LITTLE_ENDIAN = 8;
 
   /** Version that started storing continuous arcs. */
--- a/lucene/core/src/java/org/apache/lucene/util/fst/FST.java
+++ b/lucene/core/src/java/org/apache/lucene/util/fst/FST.java
@@ -109,10 +109,11 @@ public final class FST<T> implements Accountable {
 
   // Increment version to change it
   private static final String FILE_FORMAT_NAME = "FST";
-  private static final int VERSION_START = 6;
+  public static final int VERSION_START = 6;
   private static final int VERSION_LITTLE_ENDIAN = 8;
-  private static final int VERSION_CONTINUOUS_ARCS = 9;
-  static final int VERSION_CURRENT = VERSION_CONTINUOUS_ARCS;
+  public static final int VERSION_CONTINUOUS_ARCS = 9;
+  public static final int VERSION_CURRENT = VERSION_CONTINUOUS_ARCS;
+  public static final int VERSION_90 = VERSION_LITTLE_ENDIAN;
 
   // Never serialized; just used to represent the virtual
   // final node w/ no arcs:
--- a/lucene/core/src/java/org/apache/lucene/util/fst/FSTCompiler.java
+++ b/lucene/core/src/java/org/apache/lucene/util/fst/FSTCompiler.java
@@ -243,7 +243,7 @@ public class FSTCompiler<T> {
     private boolean allowFixedLengthArcs = true;
     private DataOutput dataOutput;
     private float directAddressingMaxOversizingFactor = DIRECT_ADDRESSING_MAX_OVERSIZING_FACTOR;
-    private int version;
+    private int version = FST.VERSION_CURRENT;
 
     /**
      * @param inputType The input type (transition labels). Can be anything from {@link INPUT_TYPE}
--- a/lucene/core/src/java/org/apache/lucene/util/fst/FSTCompiler.java
+++ b/lucene/core/src/java/org/apache/lucene/util/fst/FSTCompiler.java
@@ -330,10 +330,10 @@ public class FSTCompiler<T> {
 
     /** Expert: Set the codec version. * */
     public Builder<T> setVersion(int version) {
-      if (version < FST.VERSION_START || version > FST.VERSION_CURRENT) {
+      if (version < FST.VERSION_90 || version > FST.VERSION_CURRENT) {
         throw new IllegalArgumentException(
             "Expected version in range ["
-                + FST.VERSION_START
+                + FST.VERSION_90
                 + ", "
                 + FST.VERSION_CURRENT
                 + "], got "
--- a/lucene/core/src/java/org/apache/lucene/util/fst/FSTCompiler.java
+++ b/lucene/core/src/java/org/apache/lucene/util/fst/FSTCompiler.java
@@ -29,7 +29,6 @@ import static org.apache.lucene.util.fst.FST.BIT_STOP_NODE;
 import static org.apache.lucene.util.fst.FST.BIT_TARGET_NEXT;
 import static org.apache.lucene.util.fst.FST.FINAL_END_NODE;
 import static org.apache.lucene.util.fst.FST.NON_FINAL_END_NODE;
-import static org.apache.lucene.util.fst.FST.VERSION_CURRENT;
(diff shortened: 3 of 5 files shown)


ChrisHegarty (approved):
Adding back tests here is good. LGTM

mikemccand:
The `precommit` failure looks unrelated -- good 'ol `TestRandomChains`.

mikemccand:
Hmm the retried job failed with something scary and clearly FST related:

```
java.lang.AssertionError: mismatch frozenHash=1634033861501336 vs hash=6385840662311
   >         at org.apache.lucene.core@10.0.0-SNAPSHOT/org.apache.lucene.util.fst.NodeHash.add(NodeHash.java:148)
   >         at org.apache.lucene.core@10.0.0-SNAPSHOT/org.apache.lucene.util.fst.FSTCompiler.compileNode(FSTCompiler.java:384)
   >         at org.apache.lucene.core@10.0.0-SNAPSHOT/org.apache.lucene.util.fst.FSTCompiler.compile(FSTCompiler.java:976)
   >         at org.apache.lucene.analysis.common@10.0.0-SNAPSHOT/org.apache.lucene.analysis.charfilter.NormalizeCharMap$Builder.build(NormalizeCharMap.java:114)
   >         at org.apache.lucene.analysis.common@10.0.0-SNAPSHOT/org.apache.lucene.analysis.charfilter.MappingCharFilterFactory.inform(MappingCharFilterFactory.java:78)
   >         at org.apache.lucene.benchmark.byTask.tasks.AnalyzerFactoryTask.createAnalysisPipelineComponent(AnalyzerFactoryTask.java:443)
   >         at org.apache.lucene.benchmark.byTask.tasks.AnalyzerFactoryTask.setParams(AnalyzerFactoryTask.java:274)
   >         at org.apache.lucene.benchmark.byTask.utils.Algorithm.<init>(Algorithm.java:149)
   >         at org.apache.lucene.benchmark.byTask.Benchmark.<init>(Benchmark.java:65)
   >         at org.apache.lucene.benchmark.BenchmarkTestCase.execBenchmark(BenchmarkTestCase.java:72)
   >         at org.apache.lucene.benchmark.byTask.TestPerfTasksLogic.testAnalyzerFactory(TestPerfTasksLogic.java:983)
```

I'll try to understand what is happening here ...

mikemccand:
And in fact the first run's failure was NOT spurious!  It failed with this root cause:

```
org.apache.lucene.analysis.tests.TestRandomChains > test suite's output saved to /home/runner/work/lucene/lucene/lucene/analysis.tests/build/test-results/test/outputs/OUTPUT-org.apache.lucene.analysis.tests.TestRandomChains.txt, copied below:
   >     java.lang.AssertionError: mismatch frozenHash=930432529 vs hash=3634519
   >         at __randomizedtesting.SeedInfo.seed([18C0E39CA25896EF:2521CAFDE54A8B2F]:0)
   >         at org.apache.lucene.core@10.0.0-SNAPSHOT/org.apache.lucene.util.fst.NodeHash.add(NodeHash.java:148)
   >         at org.apache.lucene.core@10.0.0-SNAPSHOT/org.apache.lucene.util.fst.FSTCompiler.compileNode(FSTCompiler.java:384)
   >         at org.apache.lucene.core@10.0.0-SNAPSHOT/org.apache.lucene.util.fst.FSTCompiler.freezeTail(FSTCompiler.java:819)
   >         at org.apache.lucene.core@10.0.0-SNAPSHOT/org.apache.lucene.util.fst.FSTCompiler.compile(FSTCompiler.java:967)
   >         at org.apache.lucene.analysis.common@10.0.0-SNAPSHOT/org.apache.lucene.analysis.charfilter.NormalizeCharMap$Builder.build(NormalizeCharMap.java:114)
   >         at org.apache.lucene.analysis.tests@10.0.0-SNAPSHOT/org.apache.lucene.analysis.tests.TestRandomChains$1.lambda$new$17(TestRandomChains.java:327)
   >         at org.apache.lucene.analysis.tests@10.0.0-SNAPSHOT/org.apache.lucene.analysis.tests.TestRandomChains.newRandomArg(TestRandomChains.java:681)
   >         at org.apache.lucene.analysis.tests@10.0.0-SNAPSHOT/org.apache.lucene.analysis.tests.TestRandomChains.newCharFilterArgs(TestRandomChains.java:700)
   >         at org.apache.lucene.analysis.tests@10.0.0-SNAPSHOT/org.apache.lucene.analysis.tests.TestRandomChains$MockRandomAnalyzer.newCharFilterChain(TestRandomChains.java:836)
   >         at org.apache.lucene.analysis.tests@10.0.0-SNAPSHOT/org.apache.lucene.analysis.tests.TestRandomChains$MockRandomAnalyzer.toString(TestRandomChains.java:750)
   >         at java.base/java.lang.StringConcatHelper.stringOf(StringConcatHelper.java:453)
   >         at java.base/java.lang.StringConcatHelper.simpleConcat(StringConcatHelper.java:408)
   >         at org.apache.lucene.analysis.tests@10.0.0-SNAPSHOT/org.apache.lucene.analysis.tests.TestRandomChains.testRandomChains(TestRandomChains.java:954) 
```

`NeedMoreCaffeineException`!!

jpountz:
Don't worry @mikemccand, I'll look.

jpountz:
I think I found the bug, the FST version was mistakenly initialized to 0 instead of Version.CURRENT in some cases, it should be good now.

lucene/core/src/java/org/apache/lucene/util/fst/FSTCompiler.java
```diff
@@ -325,6 +328,21 @@ public Builder<T> directAddressingMaxOversizingFactor(float factor) {
       return this;
     }
 
+    /** Expert: Set the codec version. * */
+    public Builder<T> setVersion(int version) {
+      if (version < FST.VERSION_START || version > FST.VERSION_CURRENT) {
```
mikemccand, line 333: Hmm should we make this tighter, to `version < FST.VERSION_LITTLE_ENDIAN`?  I.e. this code cannot write in the really old format before that?  Just the pre and post continuous arcs format?

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
```diff
@@ -109,10 +109,20 @@ public enum INPUT_TYPE {
 
   // Increment version to change it
   private static final String FILE_FORMAT_NAME = "FST";
-  private static final int VERSION_START = 6;
+
+  /** First supported version, this is the version that was used when releasing Lucene 7.0. */
+  public static final int VERSION_START = 6;
+
   private static final int VERSION_LITTLE_ENDIAN = 8;
```
mikemccand, line 116: Hmm what happened to `7`?  Are we unable to count properly lol :)  Or more likely we hard-removed this version because `main` no longer supports it?  Or it was only used during SNAPSHOT builds, never released?

jpountz:
It's actually a bad news that all tests pass here, as this means that our `BasePostingsFormatTestCase` is not good enough to uncover the recent problem with `Terms#intersect`... So we could add a similar bug in the future, and we'd only learn about it after releasing when @mikemccand 's new bw compat test from #12901 starts running?

mikemccand:
> I think I found the bug, the FST version was mistakenly initialized to 0 instead of Version.CURRENT in some cases, it should be good now.

Phew!!

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
jpountz: Let me do some spelunking.

mikemccand:
> It's actually a bad news that all tests pass here, as this means that our `BasePostingsFormatTestCase` is not good enough to uncover the recent problem with `Terms#intersect`... So we could add a similar bug in the future, and we'd only learn about it after releasing when @mikemccand 's new bw compat test from #12901 starts running?

Well, at least for this particular `Terms.intersect` bug (#12895), the bug only exists when 9.9.0 reads <= 9.8 written indices, i.e. the bug is sidestepped when 9.9.0 reads 9.9.0 written indices because of the separate LSB -> MSB `writeVLong` change.  It masks/prevents the read-time bug entirely.  So there is / was no hope for `BasePostingsFormatTestCase` to catch this hidden / lurking bug, I fear.

jpountz:
Hmmm... I agree we can't expect `BasePostingsFormatTestCase` to catch all bw compat problems, but the `TestLucene90PostingsFormat` from this PR writes data in the 9.8 format of the terms dictionary, and then reads the data with the 9.9 logic, only relying on version checks to read outputs with regular vlongs. So it should be exposing the bug?

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
jpountz: So version 7 was introduced by https://github.com/apache/lucene/commit/5dd9c4c04bcab9911b9a9f0b092eadc50262ee2c but it was later removed because it did not need version checks on the read side for bw compat. I'll add a comment explaining the gap.

mikemccand:
> Hmmm... I agree we can't expect `BasePostingsFormatTestCase` to catch all bw compat problems, but the `TestLucene90PostingsFormat` from this PR writes data in the 9.8 format of the terms dictionary, and then reads the data with the 9.9 logic, only relying on version checks to read outputs with regular vlongs. So it should be exposing the bug?

Oh!  I didn't realize it did this, OK.  Then indeed it has a chance of uncovering the bug, but failed to do so :(

Note that the bug was quite hard to repro -- I started with various random simple strings, than realistic unicode strings, varying counts, term lengths, etc.. (using one Java tool to write the index using 9.8.0, and another to randomly search the index using 9.9.0), but bug would not repro after a great many iterations.

So then I switched to contiguous slices of `wikibigall` terms, and had to also randomize the `WildcardQuery` substring, and finally it repros quite readily.

Too many of our tests use unrealistic random terms ... realistic term distributions tickle the many `if` statements better.  Does `BasePostingsFormatTestCase` index from `LineFileDocs`?  If so, the nightly line docs file is quite a bit beefier and may have a better shot at repro?  (And they are somewhat extracted from `enwiki` text...).

jpountz:
OK, let me try feeding LineFileDocs into this test case.

lucene/backward-codecs/src/test/org/apache/lucene/backward_codecs/lucene90/Lucene90RWPostingsFormat.java
```diff
@@ -75,7 +75,11 @@ public FieldsConsumer fieldsConsumer(SegmentWriteState state) throws IOException
     try {
       FieldsConsumer ret =
           new Lucene90BlockTreeTermsWriter(
-              state, postingsWriter, minTermBlockSize, maxTermBlockSize);
+              state,
+              postingsWriter,
+              minTermBlockSize,
+              maxTermBlockSize,
+              Lucene90BlockTreeTermsReader.VERSION_START);
```
mikemccand, line 82: Hmm test should now be angry about this unsupported (too early) version?

jpountz: No, because `Lucene90BlockTreeTermsReader.VERSION_START` maps to `FST.VERSION_90`.

mikemccand: Oh!  Sorry, I was confusing the `VERSION_X` constants between `Lucene90BlockTreeTermsReader` vs `FST`, sigh.

jpountz: No worries.

jpountz:
> OK, let me try feeding LineFileDocs into this test case.

FTR I will look into it, but it's probably best done in a follow-up PR rather than this one, let's merge this PR first?

mikemccand (approved):
Thanks @jpountz

PR #12912, opened 2023-12-11
Add BWC test to reveal #12895

gf2121:
This patch get a 40000-terms subset (~200KB) of `wikibigall` terms dict, which can reproduce #12895. I have verified #12900 can fix the expection.


lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestBackwardsCompatibility.java
```diff
@@ -2265,4 +2268,47 @@ public void testReadNMinusTwoSegmentInfos() throws IOException {
...
+
+    writer.flush();
+    writer.commit();
+    writer.forceMerge(1);
+    writer.close();
+    dir.close();
+
+    // Gives you time to copy the index out!: (there is also
+    // a test option to not remove temp dir...):
+    Thread.sleep(100000);
```
uschindler, line 2296: this must go away, please add `nocommit`.

uschindler (commented):
We couldthink of recreating older backwards indexes, too (at least for 9.x series).

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestBackwardsCompatibility.java
gf2121: Thanks for review @uschindler @benwtrent !

I closed this since I say @mikemccand [did](https://github.com/apache/lucene/issues/12901#issuecomment-1849981059) similar things and his dictionary is much smaller than mine :)

> this must go away, please add nocommit.

I'm very new to this part. I copied this pattern from another [existing test ](https://github.com/apache/lucene/blob/069c04877011a7e305dfd9e23e194fe4fd5b59e6/lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestBackwardsCompatibility.java#L228). I can open another PR to clean this up if we are sure it makes no sense.

uschindler: wwwwuuuuuuaaaaaaah. This must go away.

mikemccand:
Oh thanks for working on this @gf2121 -- sorry that we both did it at the same time!

mikemccand:
... but at least it lead to discovering a horrifying `Thread.sleep` in our test code!

Can we ban `Thread.sleep` throughout our code?  Or are there actually useful places for it... I do like sleeping myself.

uschindler:
Solr used that, as solr is no longer part of our tree we could add sleep to fobiddenaps (maybe globally for both tests and main code). Maybe Lucene does not sleep in tests, so it could be an easy PR. Let's investigate. 🫣

mikemccand:
> Solr used that, as solr is no longer part of our tree we could add sleep to fobiddenaps (maybe globally for both tests and main code). Maybe Lucene does not sleep in tests, so it could be an easy PR. Let's investigate. 🫣

Awesome, I opened #12946 for this.

PR #12913, opened 2023-12-11
#12901: add TestBackwardsCompatibility test case that reveals the block tree IntersectTermsEnum bug #12895

mikemccand:
This is a smallish repro of #12895.  The test should fail currently with the glorious `java.lang.ArrayIndexOutOfBoundsException` -- let's confirm GitHub actions agrees.  Then I'll add an `@Ignore` before pushing ...


uschindler (commented):
+1

I am not super-happy that there's no way to recreate that index automatically, but it looks fine. The same could also happen with 9.7 indexes or before, but those all use same BlockTermsCodec, right?

PR #12915, opened 2023-12-11
Add new token filters for Japanese sutegana (捨て仮名)

daixque:
### Description

Sutegana (捨て仮名) is small letter of hiragana and katakana in Japanese. In the old Japanese text, sutegana (捨て仮名) is not used unlikely to modern one. For example:

- "ストップウォッチ" is written as "ストツプウオツチ"
- "ちょっとまって" is written as "ちよつとまつて"

So it's meaningful to normalize sutegana to normal (uppercase) characters if we search against the corpus which includes old Japanese text such as patents, legal documents, contract policies, etc.

This pull request introduces 2 token filters:

- JapaneseHiraganaUppercaseFilter for hiragana
- JapaneseKatakanaUppercaseFilter for katakana

so that user can use either one separately. Each. filter make all the sutegana (small characters) into normal kana (uppercase character) to normalize the token.

### Why it is needed

This transformation must be done as token filter. There have already been [MappingCharFilter](https://lucene.apache.org/core/8_0_0/analyzers-common/org/apache/lucene/analysis/charfilter/MappingCharFilter.html), but if we apply this character filter to normalize sutegana, it will impact to tokenization and it is not expected.


lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
```diff
@@ -0,0 +1,65 @@
...
+      for (int i = 0; i < src.length; i++) {
+        Character c = s2l.get(src[i]);
+        if (c != null) {
+          result[i] = c;
+        } else {
+          result[i] = src[i];
+        }
+      }
+      String resultTerm = String.copyValueOf(result);
+      termAttr.setEmpty().append(resultTerm);
```
mikemccand, line 59: You can avoid making `String` here by appending the `char[] result` instead.

mikemccand (approved):
This looks great!  Thanks @daixque.  I left some small comments.

I do not know Japanese myself, so I'll leave this open for a few days to give others a chance to review.  After that I'll merge (lazy consensus).

kojisekig:
From a Japanese perspective, the necessity sounds reasonable. Thank you for the contribution!

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
```diff
@@ -0,0 +1,65 @@
+package org.apache.lucene.analysis.ja;
```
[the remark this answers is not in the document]
daixque: I'm happy to do, thanks!

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseKatakanaUppercaseFilter.java
```diff
@@ -0,0 +1,83 @@
...
+    super(input);
+  }
+
+  @Override
+  public boolean incrementToken() throws IOException {
+    if (input.incrementToken()) {
+      String term = termAttr.toString();
+      // Small letter "ㇷ゚" is not single character, so it should be converted to "プ" as String
+      term = term.replace("ㇷ゚", "プ");
+      char[] src = term.toCharArray();
```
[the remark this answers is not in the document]
daixque: Thanks, but it will affect length of result character array and break the tests. So let me keep current implementation.

Here is the example of test result.
```
term 0 expected:<ちよつと[]> but was:<ちよつと[sTerm]>
Expected :ちよつと
Actual   :ちよつとsTerm
```

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
daixque: I couldn't find `append ` method signature which accept char[]. (There is CharSequence instead)

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
```diff
@@ -0,0 +1,81 @@
...
+
+  public JapaneseHiraganaUppercaseFilter(TokenStream input) {
+    super(input);
+  }
+
+  @Override
+  public boolean incrementToken() throws IOException {
+    if (input.incrementToken()) {
+      String term = termAttr.toString();
+      char[] src = term.toCharArray();
```
dungba88, line 64: I think you can iterate through the term attribute directly. These methods require byte-copy so might be inefficient

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * A {@link TokenFilter} that normalizes small letters (捨て仮名) in hiragana into normal letters. For
+ * instance, "ちょっとまって" will be translated to "ちよつとまつて".
+ *
+ * <p>This filter is useful if you want to search against old style Japanese text such as patents,
+ * legal, contract policies, etc.
+ */
+public final class JapaneseHiraganaUppercaseFilter extends TokenFilter {
+  private static final Map<Character, Character> s2l;
```
dungba88, line 33: I think the parameter should be in all-uppercase as it's a constant?

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
```diff
@@ -0,0 +1,81 @@
...
+  @Override
+  public boolean incrementToken() throws IOException {
+    if (input.incrementToken()) {
+      String term = termAttr.toString();
+      char[] src = term.toCharArray();
+      char[] result = new char[src.length];
+      for (int i = 0; i < src.length; i++) {
+        Character c = s2l.get(src[i]);
+        if (c != null) {
+          result[i] = c;
```
dungba88, line 69: It seems all small characters are just 1 position ahead of the normal characters, so you can use `result[i] = src[i] + 1;`, and you can use a Set instead of Map: https://en.wikipedia.org/wiki/Hiragana_(Unicode_block)

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseKatakanaUppercaseFilter.java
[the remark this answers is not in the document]
dungba88: The buffer return the internal byte[] of the CharTermAttribute, which might has more bytes than the actual term length. You need to use term.length() as well.

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseKatakanaUppercaseFilter.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.lucene.analysis.tokenattributes.CharTermAttribute;
+
+/**
+ * A {@link TokenFilter} that normalizes small letters (捨て仮名) in katakana into normal letters. For
+ * instance, "ストップウォッチ" will be translated to "ストツプウオツチ".
+ *
+ * <p>This filter is useful if you want to search against old style Japanese text such as patents,
+ * legal, contract policies, etc.
+ */
+public final class JapaneseKatakanaUppercaseFilter extends TokenFilter {
```
dungba88, line 32: This seems to be mostly the same as the other filter, so maybe we can combine them?

E.g you can either pass the mapping as a constructor parameter and provide 2 constants mapping

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
dungba88: Also s2l is a bit cryptic, maybe we could use LETTER_MAPPINGS or something

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseKatakanaUppercaseFilter.java
daixque: @dungba88  How should the constructor look like?

Like this?
```
public JapaneseKanaUppercaseFilter(TokenStream input, bool hiragana, bool katakana)
```

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
daixque: No. See `ゕ` for example.

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
daixque: Thanks, let me do that.

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
daixque: Thanks, let me do that.

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
dungba88: I see, that makes sense. Thank you

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
dungba88: It seems you can modify the `CharTermAttribute` directly by accessing `buffer()`, which will return the internal buffer.

```
byte[] buffer = termAttr.buffer();
buffer[i] = LETTER_MAPPINGS.get(buffer[i]);
```

This will eliminate all of the byte copy. I don't know if we are supposed to do that (but the API allow). Maybe @mikemccand could have some thought here.

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseKatakanaUppercaseFilter.java
dungba88: You are right, maybe we can consolidate them with a base class as a follow-up. This LGTM.

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
mikemccand: > This will eliminate all of the byte copy. I don't know if we are supposed to do that (but the API allow). Maybe @mikemccand could have some thought here.

This is indeed the intended usage for high performance -- directly alter that underlying `char[]` buffer, asking the term att to grow if needed, and setting the length when you are done.

dungba88:
Besides the optimization of manipulating the internal byte[] directly, I think this is good to go.

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
```diff
@@ -60,15 +60,13 @@ public JapaneseHiraganaUppercaseFilter(TokenStream input) {
   @Override
   public boolean incrementToken() throws IOException {
     if (input.incrementToken()) {
-      String term = termAttr.toString();
-      char[] src = term.toCharArray();
-      char[] result = new char[src.length];
-      for (int i = 0; i < src.length; i++) {
-        Character c = s2l.get(src[i]);
+      char[] result = new char[termAttr.length()];
```
mikemccand, line 63: I think this could instead be something like:

```
    char[] termBuffer = termAttr.buffer();
    int termLength = termAttr.length();
    for(int i=0; i<termLength; i++) {
        Character c = LETTER_MAPPINGS.get(termBuffer[i]);
        if (c != null) {
          termBuffer[i] = c;
        }
    }

    return true;
```

I.e. you can just directly manipulate the underlying buffer.

But really this is all just optimizing -- not urgent for the first commit of this awesome contribution.  It can be done in follow-on PRs.

daixque: @mikemccand @dungba88 Yeah, thanks for your suggestion. I reflected this, so please check it.

dungba88 (approved):
Thank you for the change and optimization. LGTM!

daixque:
I did refactoring to apply a same kind of enhancement to Katakana filter as well.

mikemccand:
Looks great @daixque -- would you like to add a `lucene/CHANGES.txt` entry dscribing this awesome new capability?  Be sure to put it under the `9.10.0` section since we can backport this change (it is not a 10.0.0-only feature).

daixque:
@mikemccand @dungba88 Let me ping. Do I still have anything to do for this PR? If not, could you merge it?

lucene/CHANGES.txt
```diff
@@ -174,12 +174,14 @@ API Changes
...
 ---------------------
-
 * GITHUB#12679: Add support for similarity-based vector searches using [Byte|Float]VectorSimilarityQuery. Uses a new
   VectorSimilarityCollector to find all vectors scoring above a `resultSimilarity` while traversing the HNSW graph till
   better-scoring nodes are available, or the best candidate is below a score of `traversalSimilarity` in the lowest
   level. (Aditya Prakash, Kaival Parikh)
 
+* GITHUB#12915: Add new token filters for Japanese sutegana (捨て仮名). This introduces JapaneseHiraganaUppercaseFilter
+  and JapaneseKatakanaUppercaseFilter. (Dai Sugimori)
+
```
benwtrent, line 184: We have since released 9.10. Could you add your changes to 9.11 and remove from 9.10?

Issue #12921, opened 2023-12-12
Reproducible failure in TestFloatVectorSimilarityQuery.testApproximate


easyice:
### Description

Seems related to https://github.com/apache/lucene/pull/12679

### Gradle command to reproduce

./gradlew :lucene:core:test --tests "org.apache.lucene.search.TestFloatVectorSimilarityQuery.testApproximate" -Ptests.jvms=2 -Ptests.jvmargs= -Ptests.seed=46DD7D18EE65A0A4 -Ptests.gui=true -Ptests.file.encoding=US-ASCII -Ptests.vectorsize=128

benwtrent:
Fixed via: https://github.com/apache/lucene/pull/12922

PR #12679, opened 2023-10-14
Add support for similarity-based vector searches

kaivalnp:
### Description

Background in #12579

Add support for getting "all vectors within a radius" as opposed to getting the "topK closest vectors" in the current system

### Considerations

I've tried to keep this change minimal and non-invasive by not modifying any APIs and re-using existing HNSW graphs -- changing the graph traversal and result collection criteria to:
1. Visit all nodes (reachable from the entry node in the last level) that are within an outer "traversal" radius
2. Collect all nodes that are within an inner "result" radius

### Advantages

1. Queries that have a high number of "relevant" results will get all of those (not limited by `topK`)
2. Conversely, arbitrary queries where many results are not "relevant" will not waste time in getting all `topK` (when some of them will be removed later)
3. Results of HNSW searches need not be sorted - and we can store them in a plain list as opposed to min-max heaps (saving on `heapify` calls). Merging results from segments is also cheaper, where we just concatenate results as opposed to calculating the index-level `topK`

On a higher level, finding `topK` results needed HNSW searches to happen in `#rewrite` because of an interdependence of results between segments - where we want to find the index-level `topK` from multiple segment-level results. This is kind of against Lucene's concept of segments being independently searchable sub-indexes?

Moreover, we needed explicit concurrency (#12160) to perform these in parallel, and these shortcomings would be naturally overcome with the new objective of finding "all vectors within a radius" - inherently independent of results from another segment (so we can move searches to a more fitting place?)

### Caveats

I could not find much precedent in using HNSW graphs this way (or even the radius-based search for that matter - please add links to existing work if someone is aware) and consequently marked all classes as `@lucene.experimental`

For now I have re-used lots of functionality from `AbstractKnnVectorQuery` to keep this minimal, but if the use-case is accepted more widely we can look into writing more suitable queries (as mentioned above briefly)

### Next steps

Run benchmarks with this new query to see how it compares to the `topK` based search

PR #12922, opened 2023-12-12
Fix failing BaseVectorSimilarityQueryTestCase#testApproximate

kaivalnp:
Discovered in #12921, and introduced in #12679 

The first issue is that we weren't advancing the `VectorScorer` [here](https://github.com/apache/lucene/blob/cf13a9295052288b748ed8f279f05ee26f3bfd5f/lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java#L257-L262) -- so it was still un-positioned while trying to compute the similarity score

Earlier in the PR, the underlying delegate of the `FilteredDocIdSetIterator` was `scorer.iterator()` (see [here](https://github.com/apache/lucene/blob/cad565439be512ac6e95a698007b1fc971173f00/lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java#L107)) -- so we didn't need to explicitly advance it

Later, we decided to maintain parity to `AbstractKnnVectorQuery` and introduce filtering in `AbstractVectorSimilarityQuery` (see [this commit](https://github.com/apache/lucene/commit/5096790f281e477c529a7c8311aeb353ccdffdeb)) to determine the `visitLimit` of approximate search -- after which the underlying iterator changed to the accepted docs (see [here](https://github.com/apache/lucene/blob/5096790f281e477c529a7c8311aeb353ccdffdeb/lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java#L255)) and I missed advancing the `VectorScorer` explicitly..

After doing so, we no longer get the original `java.lang.ArrayIndexOutOfBoundsException` -- but the `BaseVectorSimilarityQueryTestCase#testApproximate` starts failing because it falls back to exact search, as the limit of the prefilter is met during graph search

Relaxed the parameters of the test to fix this (making the filter less restrictive, and trying to visit a fewer number of nodes so that approximate search completes without hitting its limit)

Sorry for missing this earlier!

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java
@@ -255,6 +255,11 @@ abstract class AbstractVectorSimilarityQuery extends Query {
           new FilteredDocIdSetIterator(acceptDocs) {
             @Override
             protected boolean match(int doc) throws IOException {
+              // Advance the scorer
+              if (!scorer.advanceExact(doc)) {
+                return false;
+              }
+
               // Compute the dot product
               float score = scorer.score();
               cachedScore[0] = score * boost;
--- a/lucene/core/src/java/org/apache/lucene/search/VectorScorer.java
+++ b/lucene/core/src/java/org/apache/lucene/search/VectorScorer.java
@@ -87,6 +87,7 @@ abstract class VectorScorer {
 
     @Override
     public float score() throws IOException {
+      assert values.docID() != -1 : getClass().getSimpleName() + " is not positioned";
       return similarity.compare(query, values.vectorValue());
     }
   }
@@ -117,6 +118,7 @@ abstract class VectorScorer {
 
     @Override
     public float score() throws IOException {
+      assert values.docID() != -1 : getClass().getSimpleName() + " is not positioned";
       return similarity.compare(query, values.vectorValue());
     }
   }
--- a/lucene/core/src/test/org/apache/lucene/search/BaseVectorSimilarityQueryTestCase.java
+++ b/lucene/core/src/test/org/apache/lucene/search/BaseVectorSimilarityQueryTestCase.java
@@ -433,8 +433,8 @@ abstract class BaseVectorSimilarityQueryTestCase<
 
   public void testApproximate() throws IOException {
     // Non-restrictive filter, along with similarity to visit a small number of nodes
-    int numFiltered = random().nextInt((numDocs * 4) / 5, numDocs);
-    int targetVisited = random().nextInt(numFiltered / 10, numFiltered / 8);
+    int numFiltered = numDocs - 1;
+    int targetVisited = random().nextInt(1, numFiltered / 10);
 
     V[] vectors = getRandomVectors(numDocs, dim);
     V queryVector = getRandomVector(dim);


lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java
```diff
@@ -255,6 +255,11 @@ static VectorSimilarityScorer fromAcceptDocs(
           new FilteredDocIdSetIterator(acceptDocs) {
             @Override
             protected boolean match(int doc) throws IOException {
+              // Advance the scorer
+              if (!scorer.advanceExact(doc)) {
+                return false;
+              }
+
```
benwtrent, line 262: This tells me that none of the filtered cases drop to using exact search? If they did they would match on docs that were actually filtered out. 

Could you add a test exercising this path?

kaivalnp: > This tells me that none of the filtered cases drop to using exact search?

We do have a test case that covers this: [`testRandomFilter`](https://github.com/apache/lucene/blob/e18f9b1eb04a70d3fe5d8431a3d6724f187f050c/lucene/core/src/test/org/apache/lucene/search/BaseVectorSimilarityQueryTestCase.java#L158) randomly [filters a sub-range](https://github.com/apache/lucene/blob/e18f9b1eb04a70d3fe5d8431a3d6724f187f050c/lucene/core/src/test/org/apache/lucene/search/BaseVectorSimilarityQueryTestCase.java#L160-L161) of documents and [expects all of them to be found](https://github.com/apache/lucene/blob/e18f9b1eb04a70d3fe5d8431a3d6724f187f050c/lucene/core/src/test/org/apache/lucene/search/BaseVectorSimilarityQueryTestCase.java#L183-L184) -- which always falls back to exact search (because a `traversalSimilarity` of `Float.NEGATIVE_INFINITY` [here](https://github.com/apache/lucene/blob/e18f9b1eb04a70d3fe5d8431a3d6724f187f050c/lucene/core/src/test/org/apache/lucene/search/BaseVectorSimilarityQueryTestCase.java#L172) will visit the entire graph and exhaust any limit)

While running this test, I put debug points [here](https://github.com/apache/lucene/blob/e18f9b1eb04a70d3fe5d8431a3d6724f187f050c/lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java#L259) to check the `scorer`, and it was always unpositioned (`docid` = -1) but doesn't throw an error in *many* cases

The error is thrown when the underlying [`values`](https://github.com/apache/lucene/blob/e18f9b1eb04a70d3fe5d8431a3d6724f187f050c/lucene/core/src/java/org/apache/lucene/search/VectorScorer.java#L96) are randomly chosen to `SimpleTextFloatVectorValues` - try `./gradlew test --tests TestFloatVectorSimilarityQuery.testRandomFilter -Dtests.seed=119135B1F0803918` for a failing case (as mentioned [here](https://github.com/apache/lucene/pull/12679#issuecomment-1851062374), thanks @epotyom!)

In other cases (for example when `values` are `DenseOffHeapVectorValues`), the scorer *does not fail even when unpositioned* and returns some (garbage) value - but the count of results is still correct (and [asserted](https://github.com/apache/lucene/blob/e18f9b1eb04a70d3fe5d8431a3d6724f187f050c/lucene/core/src/test/org/apache/lucene/search/BaseVectorSimilarityQueryTestCase.java#L180-L184))

I wonder if we can put an `assert values.docID() != -1` before [this](https://github.com/apache/lucene/blob/1ac1b1cadc66364c8baca48e8334aa2855cd18b6/lucene/core/src/java/org/apache/lucene/search/VectorScorer.java#L90) and [this](https://github.com/apache/lucene/blob/1ac1b1cadc66364c8baca48e8334aa2855cd18b6/lucene/core/src/java/org/apache/lucene/search/VectorScorer.java#L120) line to ensure it is positioned before trying to compute scores? -- I tried this offline, and the issue was caught immediately

kaivalnp: I added a commit to demonstrate this^, please let me know if we can improve this / do it some other way?

benwtrent: adding an assert is nice :). In reality it would throw trying to read garbage.

Issue #12911, opened 2023-12-11
Require bundled FSTs to be on the current FST version


jpountz:
### Description

In Lucene 9.9 we bumped the FST version, but we did not re-generate our bundled FSTs such as the ones that are used by our Kuromoji analyzer. IMO we should systematically re-generate such dictionaries when bumping the FST version. Can we fail the build if the version of these dictionaries gets out-of-date?

Note: I don't think it is a bug (at least it should not cause bugs), it's more about making sure our dictionaries get any improvement we make to FSTs immediately.

mikemccand:
Since we will respin shortly 9.9.1, should we do this specifically for 9.9.1 now?  And leave this issue open for failing the (some) build when the generated FSTs are stale?

Hmm is there a top-level `gradle` target to do this ...

mikemccand:
> Hmm is there a top-level `gradle` target to do this ...

Yes, yes there is: `./gradlew regenerate`.

mikemccand:
I'll open a PR shortly with the regenerated FSTs ... seems only `kuromoji` and `nori` build FSTs.

mikemccand:
Hmm this is a bit of a spooky crab.  If I use OpenJDK 17 (`openjdk full version "17.0.9+8"`) to `./gradlew regenerate` on current `branch_9_9_0` I get this horrifying failure:

```
/s1/l/99x/gradle/generation/extract-jdk-apis/ExtractJdkApis.java:192: error: illegal character: '\ufffd'
    public void visitPermittedSubclass???(String c) {
                                      ^
```

The error does not happen with OpenJDK 21.

dweiss:
./gradlew regenerate regenerates everything, everywhere. I think you can be more selective by passing -p (project path).

dweiss:
https://github.com/apache/lucene/blob/main/help/regeneration.txt#L64-L76

mikemccand:
Thanks @dweiss -- I regenerated everything, but only the two dicts were `git diff`, which I took to be a good sign (we haven't missed to regenerated any of the other many things).

But the scary `\UFFFD` error when using JDK 17 on 9.9.x was weird.  Maybe @uschindler has an idea?  Did we maybe generate very new Unicode characters into that method name that JDK 17 can't understand?

uschindler:
[...]

PR #12924, opened 2023-12-12
#12911: regenerate binary FSTs for Nori and Kuromoji dictionaries to match current FST format

mikemccand:
This is the result of top-level `./gradlew regenerate` to rewrite all generated stuff in our source tree.  The only resulting `git diff` were the Nori and Kuromoji FST dictionaries.

Note that I had to use OpenJDK 21 to do this.  When I tried with JDK 17 (`openjdk full version "17.0.9+8"`) I hit these spooky errors:

```
/s1/l/99x/gradle/generation/extract-jdk-apis/ExtractJdkApis.java:192: error: illegal character: '\ufffd'
    public void visitPermittedSubclass???(String c) {
                                      ^
```
I think we should do this for 9.9.1?


ChrisHegarty (approved):
I dunno how to review this, so I generated the files myself and compared against those in your branch - they are identical. SO, LGTM.

dweiss (approved):
I don't think there's anything to review; what Chris did - run regenerate locally, see if there are any changes)  is the most you can do.

mikemccand:
Thanks @ChrisHegarty and @dweiss -- I'll merge to 9.9.x, 9.x and main shortly.  Hmm, rather, I'll regen on the other two branches (not certain FST format is identical everywhere), and merge this onto 9.9.x.

rmuir:
if it regenerates successfully and the test suite passes afterwards (which it seems they do), that's basically it

mikemccand:
OK I regenerated these two FSTs on 9.9.x, 9.x, and main.

PR #12933, opened 2023-12-12
Ensure Nori/Kuromoji shipped binary FST is the latest version

### LINKED ISSUE #12911: Require bundled FSTs to be on the current FST version

mikemccand:
Closes #12911

This just adds specific unit tests for the binary FST for Nori's and Kuromoji's `TokenInfoDictionary`.  I had to promote some APIs from private -> package private for test visibility.

I backed out this AM's FST upgrade confirmed the tests fail, with something like this:

```
    java.lang.AssertionError: TokenInfoDictionary's FST is not the latest version: expected 9 but got: 8; run "./gradlew regenerate" to regenerate this FST and all other resources expected:<9> but was:<8>
        at __randomizedtesting.SeedInfo.seed([32EBD74005BEF4E1:C57DC409B3CCC027]:0)
        at junit@4.13.1/org.junit.Assert.fail(Assert.java:89)
        at junit@4.13.1/org.junit.Assert.failNotEquals(Assert.java:835)
        at junit@4.13.1/org.junit.Assert.assertEquals(Assert.java:647)
        at org.apache.lucene.analysis.ja.dict.TestTokenInfoDictionary.testBinaryFSTIsLatestFormat(TestTokenInfoDictionary.java:192)
```

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/fst/FST.java
+++ b/lucene/core/src/java/org/apache/lucene/util/fst/FST.java
@@ -1236,5 +1236,9 @@ public final class FST<T> implements Accountable {
       this.version = version;
       this.numBytes = numBytes;
     }
+
+    public int getVersion() {
+      return version;
+    }
   }
 }
--- a/lucene/core/src/java/org/apache/lucene/util/fst/FST.java
+++ b/lucene/core/src/java/org/apache/lucene/util/fst/FST.java
@@ -1208,7 +1208,7 @@ public final class FST<T> implements Accountable {
   }
 
   /**
-   * Represent the FST metadata
+   * Represents the FST metadata.
    *
    * @param <T> the FST output type
    */
@@ -1237,6 +1237,11 @@ public final class FST<T> implements Accountable {
       this.numBytes = numBytes;
     }
 
+    /**
+     * Returns the version constant of the binary format this FST was written in. See the {@code
+     * static final int VERSION} constants in FST's javadoc, e.g. {@link
+     * FST#VERSION_CONTINUOUS_ARCS}.
+     */
     public int getVersion() {
       return version;
     }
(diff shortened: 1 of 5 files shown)


uschindler:
It would be better to give a more detailed Gradle line like:

```sh
./gradlew :lucene:analysis:nori:regenerate
```

The test is not the nicest looking thing, but I accept it, because it doesn't break classloading of resources. 👍

uschindler:
I never run all regeneration - it doesn't work on Windows and takes very long....

dungba88 (approved):

mikemccand:
> The test is not the nicest looking thing, but I accept it, because it doesn't break classloading of resources. 👍

Ha!  I take this a strong positive feedback @uschindler ;)

> It would be better to give a more detailed Gradle line like:

OK -- I fixed the exception message to be specific about precisely the regenerate command to run to re-build the one FST (each for Nori and Kuromoji).

I think it's ready -- I'll merge soon and port forward.

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
```diff
@@ -1236,5 +1236,9 @@ public FSTMetadata(
       this.version = version;
       this.numBytes = numBytes;
     }
+
+    public int getVersion() {
```
uschindler, line 1240: I think as this is a public class, shouldn't there be javadocs?

uschindler (approved):
One minor Javadocs related thing, but otherwise it is fine.

uschindler:
> > The test is not the nicest looking thing, but I accept it, because it doesn't break classloading of resources. 👍
> 
> Ha! I take this a strong positive feedback @uschindler ;)

hehe, I was initially thinking about some method in the FST class like `isFSTLatestVersion(InputStream in)`, but I think this is too much abstraction. I was just worried about the code duplication. If we get more Jarred FSTs anywhere else, we can rethink this and move that to LuceneTestCase class.

So all is fine, lets merge this!

lucene/core/src/java/org/apache/lucene/util/fst/FST.java
mikemccand: Ahh good catch -- I'll add.

PR #12936, opened 2023-12-13
enable error-prone's DisableUnicodeInCode check

### LINKED ISSUE #12931: Build should statically detect when an invisible unicode character, such as U+200B (zero width space), sneak into our sources
### Description

Spinoff from #12911 where I hit weird/scary exceptions during `./gradlew regenerate` with certain JDK version(s) and not others due to a lurking zero width space Unicode character.  @uschindler fixed that one case, but perhaps we have others?  And perhaps copy/apply of a patch file is another vector for them to invade us.

Can we statically assert we don't accidentally invite these little buggers in?

rmuir:
Previously this error-prone check was not enabled, requiring code to be simple ascii, because of violations in spatial3d. I used find-replace on the greek letters there and the tests pass.

Closes #12931


dweiss (approved):

mikemccand:
Thanks @rmuir!

PR #12937, opened 2023-12-13
Prevent the common zero-width code points and detect invalid UTF-8 encoding in our sources and selected resource files

### LINKED ISSUE #12931: Build should statically detect when an invisible unicode character, such as U+200B (zero width space), sneak into our sources

uschindler:
This would have prevented the problem with our gradle file. This is just another extension for `validateSourcePatterns` task. We have other bad codepoints already in there, so I'd suggest to add those.

The uFEFF is there two times (first it detects it at begin of file, so we prevent byte order marks, but later it also disallows it; it is duplicate to have different messages).

I also added a strict UTF-8 decoding check when opening the files thanks to @rmuir Code. When a file was accidentally saved in ISO-8859-x or similar bullshit now the source code patterns check reports a violation.

The only violation for zero-width codepoints was a copy-pasted comment in jflex, not relevant to code. There were also 2 CSS files with wrong encoding (changed them to UTF-8).

This fixes #12931

### DIFF
--- a/lucene/queryparser/docs/xml/cctree.js
+++ b/lucene/queryparser/docs/xml/cctree.js
@@ -1,16 +1,16 @@
 /* This code is based on the one originally provided by
-   Geir Landrö in his dTree 2.05 package. You can get it
+   Geir LandrÃ¶ in his dTree 2.05 package. You can get it
    at : www.destroydrop.com/javascript/tree/.
    
    Therefore, the DTDDoc team considers that this code is 
-   Copyright (c) 2002-2003 Geir Landrö. Since the original
+   Copyright (c) 2002-2003 Geir LandrÃ¶. Since the original
    author didn't clearly forbids copies of this part, we
    assume we're not doing anything wrong in porviding it
    to you, in a modified or non-modified form.
 */
 
 /*   
-   Geir Landrö : Orignal version, for dTree.
+   Geir LandrÃ¶ : Orignal version, for dTree.
    
    Michael Koehrsen (10/2004) : Original modification to
       allow DTDDoc to use this.
--- a/gradle/validation/validate-source-patterns.gradle
+++ b/gradle/validation/validate-source-patterns.gradle
@@ -222,8 +222,9 @@ class ValidateSourcePatternsTask extends DefaultTask {
         }
 
         invalidPatterns.each { pattern, name ->
-          if (pattern.matcher(text).find()) {
-            reportViolation(f, name);
+          def matcher = pattern.matcher(text);
+          if (matcher.find()) {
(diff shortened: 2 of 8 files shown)


dweiss (approved):

uschindler:
Hi, I further improved the source patterns checker and found another violation:

```
Execution failed for task ':lucene:queryparser:validateSourcePatterns'.
> Found 2 source violation(s):
    C:\Users\Uwe Schindler\Projects\lucene\lucene\lucene\queryparser\docs\xml\cctree.js: incorrect UTF-8 encoding [java.nio.charset.MalformedInputException: Input length = 1]
    C:\Users\Uwe Schindler\Projects\lucene\lucene\lucene\queryparser\docs\xml\dtreeStyle.css: incorrect UTF-8 encoding [java.nio.charset.MalformedInputException: Input length = 1]
```

Actually the code now decodes all source files with a REPORTING UTF-8 decoder, so any incorrectly saved ISO-8859-x file will be reported.

This is especially for Robert to be happy (copied his code)!

lucene/queryparser/docs/xml/cctree.js
```diff
@@ -1,16 +1,16 @@
...
-   Copyright (c) 2002-2003 Geir Landr�. Since the original
+   Copyright (c) 2002-2003 Geir Landrö. Since the original
    author didn't clearly forbids copies of this part, we
    assume we're not doing anything wrong in porviding it
    to you, in a modified or non-modified form.
 */
 
 /*   
-   Geir Landr� : Orignal version, for dTree.
+   Geir Landrö : Orignal version, for dTree.
```
rmuir, line 13: what is happening here????

rmuir: ok, i think this is just an artifact of the github reviewing feature...

rmuir (approved):

uschindler:
Thanks. I will merge this later to 9.x and main. I'd like to make the reporting a bit better. When a regex matches I'd like to show the fragment around (similar to highlighting).

I think next to the hidden whitespace the best here is that we detect incorrectly encoded UTF-8.

lucene/queryparser/docs/xml/cctree.js
uschindler: Github tries to autodetect the charset and then shows both sides in the detected one. So it looks wrong because on my laptop it detects ISO-8859-1 on the "broken side" and so shows bullshit on the "correct" side.

PR #12938, opened 2023-12-13
Made the UnifiedHighlighter's hasUnrecognizedQuery function processes FunctionQuery the same way as MatchAllDocsQuery and MatchNoDocsQuery queries for performance reasons.

ljak:
### Description

At [Lexum](https://lexum.com/en/), we deploy Solr with a slightly modified version of the UnifiedHighlighter: the `FunctionQuery` type is added as a knownLeaf to the `QueryVisitor` of the `hasUnrecognizedQuery` function (next to both `MatchAllDocsQuery` and `MatchNoDocsQuery` types). I wasn't there when the change was identified but it seems necessary for performance reasons. I propose the change, hoping that we don't need to maintain a slighty modified version of that class in our code base.

### DIFF
--- a/lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/UnifiedHighlighter.java
+++ b/lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/UnifiedHighlighter.java
@@ -49,6 +49,7 @@ import org.apache.lucene.index.StoredFieldVisitor;
 import org.apache.lucene.index.StoredFields;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.index.TermVectors;
+import org.apache.lucene.queries.function.FunctionQuery;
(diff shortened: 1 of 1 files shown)


romseygeek:
This seems reasonable - FunctionQuery doesn't actually match things, it just scores, so it shouldn't be used for highlighting in any case.

Could you add a test to the highlighting module showing that including a FunctionQuery as a clause in a boolean query that has other highlightable clauses doesn't prevent those clauses from being correctly highlighted?

It might also be worth pulling the set of queries that get visited via `visitLeaf()` but are not unrecognized into an explicit set which we check, rather than using a list of `instanceof` clauses which might get longer in future.

ljak:
Closing as @vletard continued the conversation over https://github.com/apache/lucene/pull/13165

Issue #12932, opened 2023-12-12
Let's run our Monster tests, at least once?


mikemccand:
### Description

I don't think our monster tests even run?

I think we should run them at least for the 9.9.1 release.  I'm especially interested in ensuring `Test2BTerms` is happy with our block tree optimization.

I will try.

slow-J:
Is there a specific method to follow for running `Test20NewsgroupsClassification`?
For me it fails when running: `./gradlew check  -Ptests.heapsize=16g -Dtests.monster=true` with
```
   >     java.security.AccessControlException: access denied ("java.io.FilePermission" "/path/to/lucene-solr/lucene/classification/20n" "read")
```

gsmiller:
@slow-J I think you’re hitting a know issue on AWS ec2 hosts with that exception. Have you tried on a non-ec2 host?

mikemccand:
> For me it fails when running: `./gradlew check -Ptests.heapsize=16g -Dtests.monster=true` with

It fails for me too -- it's silly (trying to write to an arbitrary "example" directory).

I've fixed this in #12942 but then the test hits another error (see that PR for details).

All other monster tests pass for me on 9.9, using that PR, yay!

PR #12942, opened 2023-12-14
#12932: get monsters tests compiling/running again

mikemccand:
This just fixes a bit of rust on the monster tests so they run again ...

With this change I was able to run all monster tests, except for `Test20NewsgroupsClassification` which now fails with:

```
org.apache.lucene.classification.Test20NewsgroupsClassification > test20Newsgroups FAILED
    java.io.IOException: java.lang.IllegalArgumentException: topNGroups must be >= 1 (got 0)
        at __randomizedtesting.SeedInfo.seed([9C94D9A6FA5F55ED:D92A9B02BDA49C6C]:0)
        at org.apache.lucene.classification.utils.DatasetSplitter.split(DatasetSplitter.java:175)
        at org.apache.lucene.classification.Test20NewsgroupsClassification.test20Newsgroups(Test20NewsgroupsClassification.java:155)
```

I'm not quite sure how to fix this one so I leave it be.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/index/Test2BPoints.java
+++ b/lucene/core/src/test/org/apache/lucene/index/Test2BPoints.java
@@ -143,6 +143,6 @@ public class Test2BPoints extends LuceneTestCase {
   }
 
   private static Codec getCodec() {
-    return Codec.forName("Lucene84");
+    return Codec.getDefault();
   }
 }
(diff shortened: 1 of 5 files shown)


lucene/core/src/test/org/apache/lucene/index/Test2BPoints.java
```diff
@@ -143,6 +143,6 @@ public void test2D() throws Exception {
   }
 
   private static Codec getCodec() {
-    return Codec.forName("Lucene84");
+    return Codec.getDefault();
```
benwtrent, line 146: 🙈

benwtrent (approved):

slow-J (approved):

PR #12943, opened 2023-12-14
Fix flaky tests that are caused by small float vectors

### LINKED ISSUE #12940: Test failure in TestKnnGraph.testMultiThreadedSearch
### Description

@msokolov Please take a look when you get a chance!

./gradlew test --tests TestKnnGraph.testMultiThreadedSearch -Dtests.seed=15C41D6B3378DF2B -Dtests.locale=qu-Latn-PE -Dtests.timezone=CST6CDT -Dtests.asserts=true -Dtests.file.encoding=UTF-8

benwtrent:
While quantization generally works well, when the number of dimensions is tiny (just two like in our tests), and we are indexing a circle, and we have random merge policies, we can end up getting unexpected ordering on the resulting vectors.

closes: https://github.com/apache/lucene/issues/12940

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/index/TestKnnGraph.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestKnnGraph.java
@@ -100,22 +100,18 @@ public class TestKnnGraph extends LuceneTestCase {
           }
         };
 
-    if (vectorEncoding == VectorEncoding.FLOAT32) {
-      float32Codec = codec;
-    } else {
-      float32Codec =
-          new FilterCodec(TestUtil.getDefaultCodec().getName(), TestUtil.getDefaultCodec()) {
-            @Override
-            public KnnVectorsFormat knnVectorsFormat() {
-              return new PerFieldKnnVectorsFormat() {
-                @Override
-                public KnnVectorsFormat getKnnVectorsFormatForField(String field) {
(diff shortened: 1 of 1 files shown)


lucene/core/src/test/org/apache/lucene/index/TestKnnGraph.java
```diff
@@ -100,22 +100,18 @@ public KnnVectorsFormat getKnnVectorsFormatForField(String field) {
           }
         };
 
-    if (vectorEncoding == VectorEncoding.FLOAT32) {
-      float32Codec = codec;
```
msokolov, line 104: so this was relying on the default codec being a FLOAT32 codec? TBH I haven't kept up with how we now select whether or not to use quantization. Is it the default? Do we need to override the codec to select it?

benwtrent: @msokolov yeah, two tests are really relying on perfect scores & things being float. I mistakenly turned on quantization which adds some error bands to the scores (obviously, because its lossy) and thus its flaky.

benwtrent: It could be, I was going to look into that one next.

msokolov: OK that does repro for me, but also on branch_9_8, so I think it must be a different issue - I'll open a new one to track

msokolov (approved):

PR #12947, opened 2023-12-14
Update int array growth calls

### LINKED ISSUE #12941: Find and replace uses of unbounded array growth with `growInRange`
### Description

#12844 introduced upper bounds to array growth to avoid overallocation in cases where we can determine a bound.
This task is to go through uses of the grow API, identify other cases where we know the upper bound, and replace with `growInRange`.

stefanvodita:
`LSBRadicSorter.sort` doesn't need the buffer to preserve the data that was written to it for a previous sort.

`TaskSequence` doesn't need to grow arrays beyond the number of iterations it will do to add to the array.

Closes #12941


lucene/benchmark/src/java/org/apache/lucene/benchmark/byTask/tasks/TaskSequence.java
epotyom, line 1: Maybe I'm missing something, but isn't `countByTime` limited by duration rather than number of repetitions in the 1st and 2nd change, and by `max` of `sub.countsByTime.length` in the 3rd?

stefanvodita: My reasoning must have been way off the day I wrote this. Thank you for noticing the mistake! I've reverted the changes to this file.

lucene/core/src/java/org/apache/lucene/util/LSBRadixSorter.java
```diff
@@ -92,7 +92,7 @@ public void sort(int numBits, final int[] array, int len) {
       return;
     }
 
-    buffer = ArrayUtil.grow(buffer, len);
+    buffer = ArrayUtil.growNoCopy(buffer, len);
```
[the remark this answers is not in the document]
stefanvodita: I think the idea is that we expect an application to call this method multiple times. As an optimisation, we keep a buffer allocated instead of reallocating every time the method is called. `buffer` is a permanent reference to that buffer. `buf` is a temporary one, which we might lose when we swap arrays below.

lucene/CHANGES.txt
```diff
@@ -217,6 +217,8 @@ Other
 
 * GITHUB#12934: Cleaning up old references to Lucene/Solr. (Jakub Slowinski)
 
+* GITHUB#12941: Don't preserve auxiliary buffer in LSBRadixSorter if it grows. (Stefan Vodita)
```
slow-J, line 220: FYI: Should be moved to Lucene 9.12 :D

mikemccand (approved):

stefanvodita:
Thank you for the reviews on this forgotten PR!

PR #12594, opened 2023-09-26
Add length method to RandomAccessInput

iverase:
Implementation is straight forward as we always have the length of the data when constructing a RandomAccessInput. The only odd thing is that `ByteBuffersDataInput` already has a method called `size()` so I deprecated it in favour of this new one.

closes https://github.com/apache/lucene/issues/12592

PR #12948, opened 2023-12-14
Replace usage of deprecated size() with length() in ByteBuffersDataInput

easyice:
In https://github.com/apache/lucene/pull/12594, we mark `ByteBuffersDataInput#size()`  as `Deprecated`. For simplicity, maybe we should replace the usage of deprecated `size()` with `length()` ?


dungba88:
I think we can remove the `size()` method in 10.0?

PR #12950, opened 2023-12-15
Use hppc IntIntHashMap to avoid Integer box/unbox when remapping vector ordinals during merge

msokolov:


msokolov:
actually I think we can use hppc's cursor to avoid containsKey/get having to do some duplicate work here ... I'll post a patch in a second

benwtrent (approved):

gsmiller (approved):

msokolov:
also cherry-picked to 9x branch

PR #12953, opened 2023-12-18
An improved check for ignoring the c2-crash test if running on a client compiler.

dweiss:
Follow-up to #12905

### DIFF
--- a/gradle/testing/randomization/policies/tests.policy
+++ b/gradle/testing/randomization/policies/tests.policy
@@ -60,9 +60,6 @@ grant {
   permission java.lang.RuntimePermission "getFileStoreAttributes";
   permission java.lang.RuntimePermission "writeFileDescriptor";
 
-  // needed to check if C2 (implied by the presence of the CI env) is enabled
-  permission java.lang.RuntimePermission "getenv.CI";
-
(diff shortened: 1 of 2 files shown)


uschindler:
FYI, there's also `Constants.IS_HOTSPOT_VM`. Just in case :-) But I don't think that's usefil here as additional constraint.

uschindler (changes requested):
Remove this method, too (it is dead code):

https://github.com/apache/lucene/pull/12953/files#diff-a55d6b740bc9510cd53e003a0f209951ea914ce3f9998fa333df95f7445a46d4R181

uschindler:
This works and you can still run the tests with `export CI=true`.

https://github.com/apache/lucene/blob/f6582ce04830b0122de90a045f1abc595a411aa1/gradle/testing/defaults-tests.gradle#L54

dweiss:
I've removed the unused method and added hotspot check, thanks Uwe.

uschindler (approved):

dweiss:
I've applied to main and branch_9x.

ChrisHegarty:
I'm way too slow here, sorry.  Belated LGTM. And thanks for following up on this.

PR #12958, opened 2023-12-20
Reduce contention on flushControl.isFullFlush().

jpountz:
`flushControl.isFullFlush()` is a surprising source of contention with documents that are cheap to index and many indexing threads. If I slightly modify luceneutil's `IndexGeoNames` benchmark to configure a 4GB indexing buffer and disable `TextField` fields, which are more costly to index than `KeywordField` or `IntField` fields, this brings the time to load all the dataset in the `IndexWriter` buffers from 8.0s to 7.0s with 30 threads.


s1monw (approved):
LGTM

PR #12959, opened 2023-12-21
Make sure `DocumentsWriterPerThread#getAndLock` never returns `null` on a non-empty queue.

### LINKED ISSUE #12649: TestIndexWriterThreadsToSegments.testSegmentCountOnFlushRandom fails randomly
### Description

This has failed multiple times in continuous testing environment. I have a couple different reproduction lines to try, but none seem to replicate the failure locally. The CI on which these failed is a Linux box using OpenJDK17.

trace:

```


org.apache.lucene.index.TestIndexWriterThreadsToSegments > testSegmentCountOnFlushRandom FAILED
--
  | com.carrotsearch.randomizedtesting.UncaughtExceptionError: Captured an uncaught exception in thread: Thread[id=618, name=Thread-542, state=RUNNABLE, group=TGRP-TestIndexWriterThreadsToSegments]
  |  
  | Caused by:
  | java.lang.RuntimeException: java.util.concurrent.BrokenBarrierException
  | at __randomizedtesting.SeedInfo.seed([F09778B329E7B1C9]:0)
  | at org.apache.lucene.index.TestIndexWriterThreadsToSegments$2.run(TestIndexWriterThreadsToSegments.java:239)
  |  
  | Caused by:
  | java.util.concurrent.BrokenBarrierException
  | at java.base/java.util.concurrent.CyclicBarrier.dowait(CyclicBarrier.java:252)
  | at java.base/java.util.concurrent.CyclicBarrier.await(CyclicBarrier.java:364)
  | at org.apache.lucene.index.TestIndexWriterThreadsToSegments$2.run(TestIndexWriterThreadsToSegments.java:236)
  |  
  | com.carrotsearch.randomizedtesting.UncaughtExceptionError: Captured an uncaught exception in thread: Thread[id=626, name=Thread-550, state=RUNNABLE, group=TGRP-TestIndexWriterThreadsToSegments]
  |  
  | Caused by:
  | java.lang.AssertionError
  | at __randomizedtesting.SeedInfo.seed([F09778B329E7B1C9]:0)
  | at org.junit.Assert.fail(Assert.java:87)
  | at org.junit.Assert.assertTrue(Assert.java:42)
  | at org.junit.Assert.assertTrue(Assert.java:53)
  | at org.apache.lucene.index.TestIndexWriterThreadsToSegments$CheckSegmentCount.run(TestIndexWriterThreadsToSegments.java:150)
  | at java.base/java.util.concurrent.CyclicBarrier.dowait(CyclicBarrier.java:222)
  | at java.base/java.util.concurrent.CyclicBarrier.await(CyclicBarrier.java:364)
  | at org.apache.lucene.index.TestIndexWriterThreadsToSegments$2.run(TestIndexWriterThreadsToSegments.java:236)
```

Followed by all the traces for the threads complaining about:

```


com.carrotsearch.randomizedtesting.UncaughtExceptionError: Captured an uncaught exception in thread: Thread[id=619, name=Thread-543, state=RUNNABLE, group=TGRP-TestIndexWriterThreadsToSegments]
--
  |  
  | Caused by:
  | java.lang.RuntimeException: java.util.concurrent.BrokenBarrierException
  | at __randomizedtesting.SeedInfo.seed([F09778B329E7B1C9]:0)
  | at org.apache.lucene.index.TestIndexWriterThreadsToSegments$2.run(TestIndexWriterThreadsToSegments.java:239)
  |  
  | Caused by:
  | java.util.concurrent.BrokenBarrierException
  | at java.base/java.util.concurrent.CyclicBarrier.dowait(CyclicBarrier.java:252)
  | at java.base/java.util.concurrent.CyclicBarrier.await(CyclicBarrier.java:364)
  | at org.apache.lucene.index.TestIndexWriterThreadsToSegments$2.run(TestIndexWriterThreadsToSegments.java:236)
```

```
./gradlew test --tests TestIndexWriterThreadsToSegments.testSegmentCountOnFlushRandom -Dtests.seed=F09778B329E7B1C9 -Dtests.nightly=true -Dtests.locale=ee-Latn-GH -Dtests.timezone=WET -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```

```
 ./gradlew test --tests TestIndexWriterThreadsToSegments.testSegmentCountOnFlushRandom -Dtests.seed=8D41B10F3B89A988 -Dtests.nightly=true -Dtests.locale=bez-Latn-TZ -Dtests.timezone=SST -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```

jpountz:
Before this change, `DocumentsWriterPerThread#getAndLock` could sometimes
return `null` even though the queue was empty at no point in time. The
practical implication is that we can end up with more DWPTs in memory than
indexing threads, which, while not strictly a bug, may require doing more
merging than we'd like later on.

I ran luceneutil's `IndexGeonames` with this change, and
`DocumentsWriterPerThread#getAndLock` was not the main source of
contention.

Closes #12649 #12916

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/ConcurrentApproximatePriorityQueue.java
+++ b/lucene/core/src/java/org/apache/lucene/index/ConcurrentApproximatePriorityQueue.java
@@ -16,6 +16,8 @@
  */
 package org.apache.lucene.index;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Predicate;
@@ -116,20 +118,32 @@ final class ConcurrentApproximatePriorityQueue<T> {
         }
       }
     }
-    for (int i = 0; i < concurrency; ++i) {
-      final int index = (threadHash + i) % concurrency;
-      final Lock lock = locks[index];
-      final ApproximatePriorityQueue<T> queue = queues[index];
-      lock.lock();
-      try {
+
+    // We want to make sure we return a non-null entry if this queue is not empty. This requires us
+    // to not release locks until we're done, otherwise if there is a single non-empty sub queue, as
+    // we iterate through all sub queues, there is a chance that an entry gets added to a queue we
+    // just checked and that the existing entry gets removed from a queue we haven't checked yet.
+    // This would make this method return `null` even though the queue was empty at no point in
+    // time.
+
+    final List<Lock> toUnlock = new ArrayList<>();
+    try {
+      for (int index = 0; index < concurrency; ++index) {
+        final Lock lock = locks[index];
+        final ApproximatePriorityQueue<T> queue = queues[index];
+        lock.lock();
+        toUnlock.add(lock);
         T entry = queue.poll(predicate);
         if (entry != null) {
           return entry;
         }
-      } finally {
+      }
+    } finally {
+      for (Lock lock : toUnlock) {
         lock.unlock();
       }
     }
+
     return null;
   }
 
--- a/lucene/core/src/java/org/apache/lucene/index/ConcurrentApproximatePriorityQueue.java
+++ b/lucene/core/src/java/org/apache/lucene/index/ConcurrentApproximatePriorityQueue.java
@@ -16,8 +16,6 @@
  */
 package org.apache.lucene.index;
 
-import java.util.ArrayList;
-import java.util.List;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Predicate;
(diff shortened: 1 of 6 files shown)


uschindler:
It looks like this does not fix the issue in #12916. The bug is there but it does not look like the cause for #12916.

dweiss:
For what it's worth, I've tried reproducing this on Adrien's branch with:
```
gradlew -p lucene/core -Dtests.seed=F7B4CD7A5624D5EC
beast --tests TestIndexWriterThreadsToSegments.testSegmentCountOnFlushRandom -Dtests.jvmargs="-XX:+UseCompressedOops" -Ptests.iters=1000 -Ptests.dups=100
```
but nope, no fails.

uschindler:
I get it to fail with both OpenJ9 and Hotspot on this branch, see #12916

jpountz:
OK, I think I understand why the test is still failing, this seems to be because we unlock the DWPT after putting it back into the queue. But there is also a good reason to do so. I need to think more about it.

jpountz:
I have a new iteration of this change, which now also accounts for how `DocumentsWriterPerThreadPool` may lock `DocumentsWriterPerThread` instances while they are still in the pool. I'm at 2 hours of beasting with J9 without a failure so far.

jpountz:
If there are no concerns, I plan on merging this PR soon.

uschindler:
Hi,
sorry I missed to test this. I started my beasting and try to reproduce it.
Will report back.

uschindler (approved):
I beasted this PR and it produced on my Ryzen CPU under load no failures. So it looks like it fixed the issue.

The `AtomicInteger` is only incremented so at some point it overflows. But when analyzing the code, the actual number does not matter, it just needs to be different to retry, correct? So when it overflows it does not do any harm.

uschindler:
FYI, I did not try to run the beasting for hours because previously it was failing very fast on a heavy loaded Ryzen CPU with both Hotspot and OpenJ9. Now its stable, so I trust the results also after 2 hours (1 hour with Hotspot, 1 hour with OpenJ9).

jpountz:
Thanks a lot @uschindler !

PR #12794, opened 2023-11-10
Speedup concurrent multi-segment HNWS graph search

mayya-sharipova:
Speedup concurrent multi-segment HNWS graph search by exchanging 
the global minimum similarity collected so far across segments. As the global
similarity is used as a minimum threshold that candidates need to pass to be considered, 
this allows earlier stopping for segments that don't have good candidates.

Implementation details:
- `TopKnnCollector` has a new argument `MaxScoreAccumulator` used to keep track of the global minimum similarity collected so far.
-  As soon as a `TopKnnCollector` for a segment collected the local k results, the global similarity is used as minimum required similarity for knn search.

---

Having a global queue is a crude approach and the 1st step for concurrent search.
We don’t want to stop ourselves reaching a portion of a graph which is a good neighbourhood 
of the query if we use a bound from a graph where the search is much further forward. 
Because of this, we would like to  continuing exploring local graphs with some limitations 
even if the candidates don't pass  the global min similarity threshold. For this reason,
we have added a second queue that has a size of (1-g) k, where g is greediness.

A node will be considered a candidate if it has similarity with a query that is larger 
than the current bottom similarity of the 2nd queue.

PR #12962, opened 2023-12-21
Speedup concurrent multi-segment HNSW graph search 2

mayya-sharipova:
A second implementation of #12794 using Queue instead of MaxScoreAccumulator.

Speedup concurrent multi-segment HNWS graph search by exchanging the global top scores  collected so far across segments. These global top scores set the minimum threshold that candidates need to pass to be considered. This allows earlier stopping for segments that don't have good candidates.


mayya-sharipova:
###  1M vectors of 100 dims

k=10, fanout=90

|                                 |Avg visited nodes |   QPS    |   Recall| 
|  :---                           |    ---:          |     ---: |    ---: |  
| Baseline Single segment         |               980|      2336|    0.739|        
| Baseline 3 segments concurrent  |              2627|      2857|    0.772|     
| Candidate1_with_min_score       |              2458|      2816|    0.766| 
| Candidate2_with_queue           |              2477|      2865|    0.767| 


k=100, fanout=900

|                                 |Avg visited nodes |   QPS    |   Recall| 
|  :---                           |    ---:          |     ---: |    ---: |  
| Baseline Single segment         |              6722|       430|    0.921|             
| Baseline 3 segments concurrent  |             17595|       438|    0.949|     
| Candidate1_with_min_score       |             16386|       464|    0.947|  
| Candidate2_with_queue           |             13483|       469|    0.940| 

Candidate2_with_queue VS Baseline: 
- ${\color{green}Recall}$ is better than single segment
- ${\color{green} QPS}$  are 0.3-7% better than multiple segments

Candidate2_with_queue VS Candidate1_with_min_score: 
- ${\color{red}Recall}$ is slighlty worse
- ${\color{green} QPS}$  are 1-2% better 


###  10M vectors of 100 dims

k=10, fanout=90

|                                 |Avg visited nodes |   QPS    |   Recall| 
|  :---                           |    ---:          |     ---: |    ---: |  
| Baseline Single segment         |              1081|      1798|    0.634|    
| Baseline 13 segments concurrent |             11869|      1371|    0.680|   
| Candidate1_with_min_score       |              7660|      1845|    0.600| 
| Candidate2_with_queue           |              7942|      1776|    0.606| 

k=100, fanout=900

|                                 |Avg visited nodes |   QPS    |   Recall| 
|  :---                           |    ---:          |     ---: |    ---: |   
| Baseline Single segment         |              7213|        272|   0.824| 
| Baseline 13 segments concurrent |             78069|        239|   0.894|  
| Candidate1_with_min_score       |             50649|        301|   0.868| 
| Candidate2_with_queue           |             33139|        361|   0.834| 

k=100, fanout=9900
|                                 |Avg visited nodes |   QPS    |   Recall| 
|  :---                           |    ---:          |     ---: |    ---: |   
| Baseline Single segment         |             55168|         36|   0.916| 
| Baseline 13 segments concurrent |            521851|         29|   0.962| 
| Candidate1_with_min_score       |            331252|         34|   0.954| 
| Candidate2_with_queue           |            193082|         55|   0.938| 

Candidate2_with_queue VS Baseline:
- ${\color{black}Recall}$ is slightly worse or better than single segment
- ${\color{green} QPS}$ are 30-52% better than multiple segments


Candidate2_with_queue VS Candidate1_with_min_score: 
- ${\color{red}Recall}$ is slighlty worse
- ${\color{green} QPS}$ is 20-60% better (but is -6% worse on small k)


### 10M vectors of 768 dims

k=10, fanout=90

|                                 |Avg visited nodes |   QPS    |   Recall| 
|  :---                           |    ---:          |     ---: |    ---: |  
| Baseline Single segment         |              1095|       974|    0.542|   
| Baseline 19 segments concurrent |             18091|       569|    0.541|   
| Candidate1_with_min_score       |             10700|       824|    0.474| 
| Candidate2_with_queue           |             11007|       783|    0.479|


k=100, fanout=900

|                                 |Avg visited nodes |   QPS    |   Recall| 
|  :---                           |    ---:          |     ---: |    ---: |   
| Baseline Single segment         |              7118|       152|    0.688|  
| Baseline 19 segments concurrent |            120453|        88|    0.732| 
| Candidate1_with_min_score       |             65778|       139|    0.693| 
| Candidate2_with_queue           |             44775|       183|    0.650| 


k=100, fanout=9900

|                                 |Avg visited nodes |   QPS    |   Recall| 
|  :---                           |    ---:          |     ---: |    ---: |   
| Baseline Single segment         |             59308|        19|    0.797|  
| Baseline 19 segments concurrent |            818268|        11|    0.847|  
| Candidate1_with_min_score       |            463464|        18|    0.828| 
| Candidate2_with_queue           |            251612|        33|    0.779| 

Candidate2_with_queue VS Baseline:
- ${\color{red}Recall}$ is slightly worse than than single segment
- ${\color{green} QPS}$ are 38-300% better than  multiple segments

Candidate2_with_queue VS Candidate1_with_min_score: 
- ${\color{red}Recall}$ is slighlty worse (but better on small k)
- ${\color{green} QPS}$ is 31-83% better (but is 5% worse on small k)

tveasey:
IMO we shouldn't focus too much on recall since the greediness of non-competitive search allows us to tune this. My main concern is does contention on the queue updates cause slow down. This aside, I think the queue is strictly better.

The search might wind up visiting fewer vertices for min score sharing, because of earlier decisions might mean it by chance gets transiently better bounds, but this should be low probability particularly when the search has to visit many vertices. And indeed these cases are where we see big wins from using a queue.

There appears to be some evidence of contention. This is suggested by looking at the runtime vs expected runtime from vertices visited, e.g.

| scenario | QPS(score) / QPS(queue) | Visited(queue) / Visited(score) |
| --- | --- | --- |
| n=10M, dim=100, k = 100, fo = 900 | 0.83 | 0.65 |
| n=10M, dim=768, k = 100, fo = 900 | 0.76 | 0.68 |

Note that the direction of this effect is consistent, but the size is not (fo = 900 shows the largest effect). However, all that said we still get significant wins in performance, so my vote would be to use the queue and work on strategies for reducing contention, there are various ideas we had for ways to achieve this.

mayya-sharipova:
I have also done experiments using Cohere dataset, as as seen below for 10M docs dataset, the speedups with the proposed approach are 1.7-2.5x times.

## Cohere/wikipedia-22-12-en-embeddings

- [Cohere/wikipedia-22-12-en-embeddings](https://huggingface.co/datasets/Cohere/wikipedia-22-12-en-embeddings) dataset
- 768 dims

### 1M vectors 
k=10, fanout=90

|                                 |Avg visited nodes |   QPS    |   Recall| 
|  :---                           |    ---:          |     ---: |    ---: |  
| Baseline Single segment         |               804|      3225|    0.454|   
| Baseline 8 segments concurrent  |              1807|      1831|    0.887|   
| Candidate2_with_queue           |              1807|      1872|    0.887|

k=100, fanout=900
|                                 |Avg visited nodes |   QPS    |   Recall| 
|  :---                           |    ---:          |     ---: |    ---: |  
| Baseline Single segment         |              4555|       527|    0.477|   
| Baseline 8 segments concurrent  |              9119|       261|    0.923|   
| Candidate2_with_queue           |              9119|       265|    0.923|

### 10M vectors 
k=10, fanout=90

|                                 |Avg visited nodes |   QPS    |   Recall| 
|  :---                           |    ---:          |     ---: |    ---: |  
| Baseline Single segment         |                  |          |         |   
| Baseline 19 segments concurrent |             37726|       293|    0.971|   
| Candidate2_with_queue           |             20199|       501|    0.960|


k=100, fanout=900
|                                 |Avg visited nodes |   QPS    |   Recall| 
|  :---                           |    ---:          |     ---: |    ---: |  
| Baseline Single segment         |                  |          |         |   
| Baseline 19 segments concurrent |            234047|        47|    0.992|   
| Candidate2_with_queue           |             74995|       118|    0.979|

lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java
```diff
@@ -27,25 +29,72 @@
  */
 public final class TopKnnCollector extends AbstractKnnCollector {
 
+  // greediness of globally non-competitive search: [0,1]
+  private static final float DEFAULT_GREEDINESS = 0.9f;
   private final NeighborQueue queue;
+  private final float greediness;
+  private final FloatHeap nonCompetitiveQueue;
+  private final FloatHeap updatesQueue;
+  private final int interval = 0x3ff; // 1023
```
jimczi, line 38: This interval is important since we need to ensure that we don't use the global queue too early. Did you check with different values? I'd expect that a smaller interval could affect recall.

lucene/core/src/java/org/apache/lucene/index/LeafReader.java
```diff
@@ -240,11 +241,19 @@ public final PostingsEnum postings(Term term) throws IOException {
...
    * @lucene.experimental
    */
   public final TopDocs searchNearestVectors(
-      String field, float[] target, int k, Bits acceptDocs, int visitedLimit) throws IOException {
+      String field,
+      float[] target,
+      int k,
+      Bits acceptDocs,
+      int visitedLimit,
+      BlockingFloatHeap globalScoreQueue)
```
jimczi, line 255: I wonder if we could introduce a `CollectorManager` similar to what we have for `TopDocs`? This global queue should be an implementation detail of the collector imo.

lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java
jimczi: It's also important to check the order of execution. For instance what happens if all segments are executed serially (rather than in parallel), does it changes the recall?

lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java
```diff
@@ -27,25 +29,72 @@
...
+
+    boolean globalSimUpdated = false;
+    if (globalSimilarityQueue != null) {
+      updatesQueue.offer(similarity);
+      globalSimUpdated = nonCompetitiveQueue.offer(similarity);
+
+      if (kResultsCollected) {
+        // as we've collected k results, we can start do periodic updates with the global queue
+        if (firstKResultsCollected || (visitedCount & interval) == 0) {
+          cachedGlobalMinSim = globalSimilarityQueue.offer(updatesQueue.getHeap());
```
jimczi, line 78: This way of updating the global queue periodically doesn't bring much if `k` is close to the size of the queue. For instance if `k` is 1000, we only save 24 updates with this strategy. That's fine I guess especially considering that the interval is not only to save concurrent update in the queue but also to ensure that we are far enough in the walk of the graph.

jimczi (commented):
Thanks @mayya-sharipova for all the testing. I left some comments. 
Some of the recalls for the single segment baseline seem seem quite off (`0.477`?). Are you sure that there was no issue during the testing?

mayya-sharipova:
@jimczi Thanks for your feedback.

> Some of the recalls for the single segment baseline seem seem quite off (0.477?). Are you sure that there was no issue during the testing?

Sorry, I made a mistake. I've updated the results. 

I will work on more experiments to address your other comments/

lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java
mayya-sharipova: @jimczi Sorry I did not understand this comment, can you please clarify it. 

What does it mean `k` is close to the size of queue?  The `globalSimilarityQueue` is of size `k`. 

Also I am not clear how you derive the value of `24`?

jimczi: Sorry I meant if `k` is close to the interval (1024). In such case delaying the update to the global queue doesn't save much.

lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java
tveasey: This seems quite large to me, based on the total node visited counts in the Cohere results. For example, for fo = 90 we'd only refresh around twice.

lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java
tveasey: There is a subtlety: note that `firstKResultsCollected` is only true the first time we visit k nodes, so this is saying only refresh every 1024 iterations thereafter. The refresh schedule is `k, k + 1024, k + 2048, ...`. (The || threw me initially.)

As per my comment above, 1024 seems infrequent to me. (Of course you may have tested smaller values and determined this to be a good choice.) If we think it is risky sharing information too early, I would be inclined to share on the schedule `max(k, c1), max(k, c1) + c2, max(k, c1) + 2 * c2, ...` with `c1 > c2` and decouple the concerns.

Also, there maybe issues with *using* information too early, in which case `minCompetitiveSimilarity` would need a check that `visitedCount > max(k, c1)` before it starts modifying `minSim`.

In any case, I can see results being rather sensitive to these choices so if you haven't done a parameter exploration it might be worth trying a few choices.

lucene/core/src/java/org/apache/lucene/util/hnsw/BlockingFloatHeap.java
```diff
@@ -0,0 +1,190 @@
...
+   * @return the new 'top' element in the queue.
+   */
+  public float offer(float[] values) {
+    lock.lock();
+    try {
+      for (int i = values.length - 1; i >= 0; i--) {
+        if (size < maxSize) {
+          push(values[i]);
+        } else {
+          if (values[i] >= heap[1]) {
```
tveasey, line 84: This should break otherwise if the values are sorted.

mayya-sharipova:
I have done more experiments with different `interval` values: 

Cohere 786 dims:

1M vectors, k=10, fanout=90
|  Interval  | Avg visited nodes |   QPS    |   Recall| 
|  ---:      |    ---:           |     ---: |    ---: |    
| 255        |              10283|       959|    0.946|
| 511        |              11455|       912|    0.956|
| 1023       |              12241|       888|    0.963|


1M vectors, k=100, fanout=900
|  Interval  | Avg visited nodes |   QPS    |   Recall| 
|  ---:      |    ---:           |     ---: |    ---: |  
| 255        |              28175|       286|    0.978|
| 511        |              32410|       231|    0.982|
| 1023       |              38929|       193|    0.985|


10M vectors, k=10, fanout=90
|  Interval  | Avg visited nodes |   QPS    |   Recall| 
|  ---:      |    ---:           |     ---: |    ---: |   
| 255        |              18325|       567|    0.906|
| 511        |              20108|       518|    0.919|
| 1023       |              22041|       472|    0.932|

10M vectors, k=100, fanout=900
|  Interval  | Avg visited nodes |   QPS    |   Recall| 
|  ---:      |    ---:           |     ---: |    ---: |    
| 255        |              58543|       157|    0.958|
| 511        |              66518|       134|    0.966|
| 1023       |              78740|       115|    0.972|

lucene/core/src/java/org/apache/lucene/util/hnsw/BlockingFloatHeap.java
mayya-sharipova: @tveasey Thanks for your feedback. I am not sure what do you meant by "break".  In this method, it is good if the passed `values` are sorted in asc order, but it is not necessary.

This method is for bulk update of the queue. For a faster performance it is better to start with larger values, so that if `values[i] < heap[1]`, we don't need to do anything.

mayya-sharipova:
> It's also important to check the order of execution. For instance what happens if all segments are executed serially (rather than in parallel), does it changes the recall?

@jimczi  The results are below


---
Sequential run cohere: 768 dims; interval: 255

### 1M vectors 
k=10, fanout=90

|                                 |Avg visited nodes |   QPS    |   Recall| 
|  :---                           |    ---:          |     ---: |    ---: |  
| Baseline Single segment         |              2033|      1212|    0.880|   
| Baseline 8 segments sequential  |             13927|       175|    0.974|   
| Candidate_with_queue sequential |              9407|       269|    0.923|

k=100, fanout=900
|                                 |Avg visited nodes |   QPS    |   Recall| 
|  :---                           |    ---:          |     ---: |    ---: |  
| Baseline Single segment         |             13083|       167|    0.964|   
| Baseline 8 segments sequential  |             81824|        28|    0.997|   
| Candidate_with_queue sequential |             36851|        62|    0.983|

### 10M vectors 
k=10, fanout=90

|                                  |Avg visited nodes |   QPS    |   Recall| 
|  :---                            |    ---:          |     ---: |    ---: |  
| Baseline Single segment          |              2189|       984|    0.929|   
| Baseline 19 segments sequential  |             37656|        59|    0.951|   
| Candidate_with_queue sequential  |             17748|       121|    0.884|


k=100, fanout=900
|                                 |Avg visited nodes |   QPS    |   Recall| 
|  :---                           |    ---:          |     ---: |    ---: |  
| Baseline Single segment         |             15040|       122|    0.950|   
| Baseline 19 segments sequential |            229945|         9|    0.990|   
| Candidate_with_queue sequential |             77910|        27|    0.964|

lucene/core/src/java/org/apache/lucene/util/hnsw/BlockingFloatHeap.java
tveasey: But I mean we don't need to keep iterating the loop, i.e. we keep iterating all the way down to 0 but know we will never push anything, instead we can just exit immediately. This is all done under a lock so it seems best to avoid this work, even if it is just the fast path of the loop.

tveasey:
> This makes an interval of 255 a reasonable choice.

I agree. This looks better to me. One thing I would be intrigued to try is the slight change in schedules as per [this](https://github.com/apache/lucene/pull/12962#discussion_r1453323532). Particularly, what happens if we delay using the information in `minCompetitiveSimilarity`. However, these results are already very good and we could push them out to a follow on PR.

One last thing I think we should consider is exploring the variance we get in recall as a result of this change. Specifically, if we were to run with some random waits in the different segment searches what is the variation in the recalls we see?

The danger is we get unlikely in ordering of searches and prune segment searches which contain the true nearest neighbours more aggressively sometimes. On average this shouldn't happen, but if we also see low variation in recall for the 1M test case in such a test case it would reassuring.

lucene/core/src/java/org/apache/lucene/index/LeafReader.java
```diff
@@ -280,12 +289,20 @@ public final TopDocs searchNearestVectors(
...
    */
   public final TopDocs searchNearestVectors(
-      String field, byte[] target, int k, Bits acceptDocs, int visitedLimit) throws IOException {
+      String field,
+      byte[] target,
+      int k,
+      Bits acceptDocs,
+      int visitedLimit,
+      BlockingFloatHeap globalScoreQueue)
+      throws IOException {
```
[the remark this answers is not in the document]
benwtrent: I agree, we can have a global queue passed around in the various collectors. I don't think a new overloaded function is necessary.

benwtrent:
I ran my own experiment, which showed some interesting and frustrating results. 

I adjusted the indexing to randomly commit() on every 500 docs or so. I indexed the first 10M docs of cohere-wiki and used max-inner product over the raw float32. This showed that we have some graph building problems, will include those results as well.

# Graph

image

The python code & raw data used:

<detail>

```python
```

</detail>

benwtrent:
OK, I reran my tests, but over the first 500k docs. Randomly committing & merging with a 500MB buffer. The numbers are much saner. I wonder if I have a bug in my data and start sending garbage once I get above 1M vectors...

This shows that having the dynamic `k` based on segment size brings us closer to baseline single segment number of visited. We are still in a different magnitude, but both static & dynamic `k` with the queue are much much better (2-3x) than the baseline multi-segment search.

```
plt.plot([985, 1225, 1398], [0.993, 1.0, 1.0], marker='o', label='baseline_single')
plt.plot([37505, 57693, 72611], [1.0, 1.0, 1.0], marker='o', label='baseline_multi')
plt.plot([12099, 18411, 24452], [0.999, 1.0, 1.0], marker='o', label='candidate_static_k')
plt.plot([10078, 13369, 16769], [0.992, 0.999, 1.0], marker='o', label='candidate_dynamic_k')
```

image

benwtrent:
I fixed my data and ran with 1.5M cohere:

static_k is this PR
dynamic_k is this PR + scaling the `k` explored by
```
loat v = (float)Math.log(sumVectorCount / (double) leafSize);
float filterWeightValue = 1/v;
```

Scaling the `k` is another nice incremental change. Up and to the left is best. We get better recall with visiting fewer consistently.

```
plt.plot([2304, 2485, 3189, 4028, 5610, 16165], [0.875, 0.886, 0.916, 0.938, 0.960, 0.992], marker='o', label='baseline_single')
plt.plot([43015, 45946, 56864, 69149, 90772, 210727], [0.980, 0.982, 0.989, 0.992, 0.996, 1.000], marker='o', label='baseline_multi')
plt.plot([22706, 23749, 27142, 30893, 37162, 76032], [0.959, 0.962, 0.970, 0.976, 0.983, 0.996], marker='o', label='candidate_static_k')
plt.plot([16099, 17183, 20788, 24809, 31334, 64921], [0.937, 0.945, 0.962, 0.973, 0.983, 0.996], marker='o', label='candidate_dynamic_k')
```

image

lucene/core/src/java/org/apache/lucene/index/LeafReader.java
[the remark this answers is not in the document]
mayya-sharipova: Addressed in the last commit

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -79,24 +82,30 @@ public Query rewrite(IndexSearcher indexSearcher) throws IOException {
       filterWeight = null;
     }
 
+    BlockingFloatHeap globalScoreQueue = new BlockingFloatHeap(k);
```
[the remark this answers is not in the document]
mayya-sharipova: Addressed in the last commits

mayya-sharipova:
@benwtrent Thanks for running additional tests. Looks like running with dynamic `k` can speed up searches, but I think we need to run more experiments on smaller dims datasets as well, so I think we should leave this in the follow up.

@jimczi @tveasey I've addressed your comments.  Are we ok to merge as it is now.

---

The following is left for the follow up work:

1.  Dynamic k based on segment size. Currently we gather at least k documents for every segment before starting synchronizing with the global queue. We need to adjust k based on the segment size.
2. Improve synchronization schedule with the global queue. Currently we do synchronization with the global queue every 255 visited docs.  Can we improve schedule as per as per [comment](https://github.com/apache/lucene/pull/12962#discussion_r1453323532).

benwtrent:
> but I think we need to run more experiments on smaller dims datasets as well, how about we leave this for the follow up?

I am 100% fine with this. It was a crazy idea and it only gives us an incremental change. This PR gives a much better improvement as it is already.

tveasey:
> @jimczi @tveasey I've addressed your comments. Are we ok to merge as it is now.

I'm happy

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -79,24 +83,32 @@ public Query rewrite(IndexSearcher indexSearcher) throws IOException {
...
     for (LeafReaderContext context : leafReaderContexts) {
-      tasks.add(() -> searchLeaf(context, filterWeight));
+      tasks.add(() -> searchLeaf(context, filterWeight, knnCollectorManager));
     }
     TopDocs[] perLeafResults = taskExecutor.invokeAll(tasks).toArray(TopDocs[]::new);
 
     // Merge sort the results
     TopDocs topK = mergeLeafResults(perLeafResults);
+    if (infoStream.isEnabled("KnnVectorQuery")) {
+      infoStream.message("KnnVectorQuery", "visited:" + topK.totalHits.value);
```
jimczi, line 100: looks like a leftover too?

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -79,24 +83,32 @@ public Query rewrite(IndexSearcher indexSearcher) throws IOException {
       filterWeight = null;
     }
 
+    final boolean supportsConcurrency = indexSearcher.getSlices().length > 1;
```
jimczi, line 86: A slice can have multiple segments so it should rather check the reader's leaves here and maybe call the boolean `isMultiSegments`?

jimczi (commented):
Thank @mayya-sharipova it's getting close. I left more comments. Can you set the pr to ready for review in case others want to chime in. 
@benwtrent can you look at the changes in the knn query with the KnnCollectorManager?

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
mayya-sharipova: I copied this how we do that for [top docs collectors](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java#L478)

But you are right, because we see speedups even in sequential run, you suggestions make sense.

lucene/core/src/java/org/apache/lucene/index/LeafReader.java
```diff
@@ -236,27 +235,24 @@ public final PostingsEnum postings(Term term) throws IOException {
...
    * @param acceptDocs {@link Bits} that represents the allowed documents to match, or {@code null}
    *     if they are all allowed to match.
-   * @param visitedLimit the maximum number of nodes that the search is allowed to visit
+   * @param knnCollector collector with settings for gathering the vector results.
    * @return the k nearest neighbor documents, along with their (searchStrategy-specific) scores.
    * @lucene.experimental
    */
   public final TopDocs searchNearestVectors(
-      String field, float[] target, int k, Bits acceptDocs, int visitedLimit) throws IOException {
+      String field, float[] target, Bits acceptDocs, KnnCollector knnCollector) throws IOException {
```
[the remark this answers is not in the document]
benwtrent: I agree with Jim, this should be `String field, float[] target, int k, Bits acceptDocs, int visitedLimit` at least for this PR, and not use a queue.

lucene/core/src/java/org/apache/lucene/index/LeafReader.java
```diff
@@ -277,27 +273,24 @@ public final TopDocs searchNearestVectors(
...
-   * @param k the number of docs to return
    * @param acceptDocs {@link Bits} that represents the allowed documents to match, or {@code null}
    *     if they are all allowed to match.
-   * @param visitedLimit the maximum number of nodes that the search is allowed to visit
+   * @param knnCollector collector with settings for gathering the vector results.
    * @return the k nearest neighbor documents, along with their (searchStrategy-specific) scores.
    * @lucene.experimental
    */
   public final TopDocs searchNearestVectors(
-      String field, byte[] target, int k, Bits acceptDocs, int visitedLimit) throws IOException {
```
benwtrent, line 288: same here, this shouldn't be mutated at all.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnCollector.java
```diff
@@ -23,7 +23,7 @@
  */
 public abstract class AbstractKnnCollector implements KnnCollector {
 
-  private long visitedCount;
+  long visitedCount;
```
benwtrent, line 26: I think this should be protected, not package private. Only sub-classes should be able to read it.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
benwtrent: > because we see speedups even in sequential run

Do you mean speed ups without concurrency via sharing information? That is interesting, I wonder why that is.

lucene/core/src/java/org/apache/lucene/search/knn/KnnCollectorManager.java
```diff
@@ -0,0 +1,38 @@
...
+import java.io.IOException;
+import org.apache.lucene.search.KnnCollector;
+import org.apache.lucene.util.BitSet;
+
+/**
+ * KnnCollectorManager responsible for creating {@link KnnCollector} instances. Useful to create
+ * {@link KnnCollector} instances that share global state across leaves, such a global queue of
+ * results collected so far.
+ */
+public abstract class KnnCollectorManager<C extends KnnCollector> {
```
benwtrent, line 29: Do we need these generics? This also seems like it should be an `interface`

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -79,24 +83,32 @@ public Query rewrite(IndexSearcher indexSearcher) throws IOException {
...
+    KnnCollectorManager<?> knnCollectorManager = getKnnCollectorManager(k, supportsConcurrency);
```
benwtrent, line 87: I think this interface should accept the `indexSearcher` as the parameter and not `supportsConcurrency` or `multipleLEaves`

This way it can build whatever internal state it needs, this is particularly useful for `DiversifyingChildrenFloatKnnVectorQuery` etc.

lucene/core/src/java/org/apache/lucene/search/knn/KnnCollectorManager.java
```diff
@@ -0,0 +1,38 @@
...
+
+  /**
+   * Return a new {@link KnnCollector} instance.
+   *
+   * @param visitedLimit the maximum number of nodes that the search is allowed to visit
+   * @param parentBitSet the parent bitset, {@code null} if not applicable
+   */
+  public abstract C newCollector(int visitedLimit, BitSet parentBitSet) throws IOException;
```
benwtrent, line 37:
```suggestion
  public abstract C newCollector(int visitedLimit, LeafReaderContext context) throws IOException;
```

Also, I am not even sure `visitedLimit` should be there. It seems like something the manager should already know about (as in this instance its static) and we just need to know about the context (the context is for `DiversifyingChildrenFloatKnnVectorQuery` so that its collector manager can create `BitSet parentBitSet` from its encapsulated `BitSetProducer`).

I also think this method could return `null` if collection is not applicable for that given leaf context.

lucene/join/src/java/org/apache/lucene/search/join/DiversifyingChildrenByteKnnVectorQuery.java
```diff
@@ -123,7 +124,16 @@ protected TopDocs exactSearch(LeafReaderContext context, DocIdSetIterator accept
   }
 
   @Override
-  protected TopDocs approximateSearch(LeafReaderContext context, Bits acceptDocs, int visitedLimit)
+  protected KnnCollectorManager<?> getKnnCollectorManager(int k, boolean supportsConcurrency) {
+    return new DiversifyingNearestChildrenKnnCollectorManager(k);
```
benwtrent, line 128: If we adjust the interface, this manager could know about `BitSetProducer parentsFilter;` and abstract that away from this query.

lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java
```diff
@@ -27,25 +29,78 @@
  */
 public final class TopKnnCollector extends AbstractKnnCollector {
 
+  // greediness of globally non-competitive search: (0,1]
+  private static final float DEFAULT_GREEDINESS = 0.9f;
+  // the local queue of the results with the highest similarities collected so far in the current
+  // segment
```
benwtrent, line 35: I think this should be a separate collector. Something like `MultiLeafTopKnnCollector`. 

There is such very little code from the original collector still around, it seems weird to me. We should have two, one that shares information, another that doesn't. 

This allows us to remove all the `null` values in the ctor.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
tveasey: It is because later graph searches are faster: they use information gathered in earlier ones. (Note though the speed up is only relative to not sharing information and searching multiple graphs not relative to searching a single graph.)

jimczi: For the dynamic k follow up we can also look at the order of the segments. For instance if we start with largest first we can be more aggressive on smaller segments.

lucene/core/src/java/org/apache/lucene/search/knn/KnnCollectorManager.java
mayya-sharipova: `visitedLimit` changes per segment as different [cost](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L122)  of using filter in a segment context

lucene/core/src/java/org/apache/lucene/index/LeafReader.java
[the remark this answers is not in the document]
mayya-sharipova: Address in 3e5b0a4

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
mayya-sharipova: Addressed

lucene/core/src/java/org/apache/lucene/index/LeafReader.java
mayya-sharipova: Addressed

lucene/core/src/java/org/apache/lucene/search/knn/KnnCollectorManager.java
mayya-sharipova: Addressed

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
mayya-sharipova: Addressed

lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java
mayya-sharipova: Great idea, addressed with latest changes!

mayya-sharipova:
@benwtrent @jimczi Thanks for your great feedback, with the latest changes I tried to address your comments, please check if need something else.

benwtrent (approved):
One minor thing, but this looks much better.

jpountz:
FYI I pushed an annotation, this yielded a major speedup: http://people.apache.org/~mikemccand/lucenebench/VectorSearch.html.

uschindler:
Is it called HNSW or HNWS? I just noticed the title of this PR and differing changes entries.

msokolov:
HNSW stands for "hierarchical navigable small world" - that should make it easy to remember :)

mikemccand:
I think this was released in 9.10.0?  I added a milestone.

wurui90:
Dear experts: I wonder 

1) Will this PR introduce query-time non determinism? When we issue the same query multiple times on the same index, will the query result be identical?

2) Is this PR automatically on? i.e. we still use the `IndexSearcher.search(vectorQuery)` without extra args for executor or boolean flags, and this feature will be on. (Based on our reading of the codes, the answer is yes.)

mayya-sharipova:
@wurui90  Answering your questions
1. Yes, indeed, there is some non-determinism, but I think it would not be much noticed in practice, as the algorithm ensures that each segment still collects some good results before exchanging globally with other segments.
2. Yes, indeed, it automatically on. If your indexSearcher is using an executor with a single thread, than you will get deterministic behaviour. Also, you can set the greediness to 0 [here](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/knn/MultiLeafKnnCollector.java#L76), will effectively disable algorithm.

PR #12965, opened 2023-12-21
Remove unnecessary fields loop from extractWeightedSpanTerms()

sabi0:


sabi0:
It seems the method could be skipped completely if the query does not contain `fieldName` or `defaultField`:
```
  protected void extractWeightedSpanTerms(Map<String,WeightedSpanTerm> terms, SpanQuery spanQuery, float boost) throws IOException {
    Set<String> queryFieldNames = new HashSet<>();
    collectSpanQueryFields(spanQuery, queryFieldNames);
    if (fieldName != null
        && !queryFieldNames.contains(fieldName)
        && (defaultField == null || !queryFieldNames.contains(defaultField))) {
      return;
    }
```

Should I add this to the PR?

dweiss:
Yep, I think this early termination condition makes sense. Could you add it, please? Thank you.

sabi0:
I've rebased on the latest main, added the early termination check and a separate commit with minor cleanups.
Please let me know if you'd like me to squash them all before the merge.

dweiss:
Thank you!

PR #12967, opened 2023-12-23
Add missing spaces in concatenated strings

sabi0:
Mostly changing code like this
`"expected: " + currentFullFlushDelQueue + "but was: "`
to
`"expected: " + currentFullFlushDelQueue + " but was: "`

But also fixing bugs like `"" + i + numDocs` => `"" + (i + numDocs)`.

I took the liberty to also do some minor housekeeping in the affected files:
- corrected grammar and spelling mistakes
- removed unnecessary type parameters
- optimized API usage (e.g. "UTF-8" => `UTF_8`, `assertThat()` => `assertEquals()`, etc.)

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
@@ -58,7 +58,7 @@ import org.apache.lucene.util.InfoStream;
  *
  * <p>When flush is called by IndexWriter we check out all DWPTs that are associated with the
  * current {@link DocumentsWriterDeleteQueue} out of the {@link DocumentsWriterPerThreadPool} and
- * write them to disk. The flush process can piggy-back on incoming indexing threads or even block
+ * write them to disk. The flush process can piggyback on incoming indexing threads or even block
  * them from adding documents if flushing can't keep up with new documents being added. Unless the
  * stall control kicks in to block indexing threads flushes are happening concurrently to actual
  * index requests.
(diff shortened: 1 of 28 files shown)


lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
```diff
@@ -236,7 +236,7 @@ synchronized void abort() throws IOException {
     }
   }
 
-  final boolean flushOneDWPT() throws IOException {
+  boolean flushOneDWPT() throws IOException {
```
mikemccand, line 239: Ahh because the class itself is already final, great!

lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseDocValuesFormatTestCase.java
```diff
@@ -3631,9 +3631,9 @@ public void testHighOrdsSortedSetDV() throws Exception {
   }
 
   private interface FieldCreator {
-    public Field next();
+    Field next();
 
-    public DocIdSetIterator iterator(IndexReader r) throws IOException;
```
mikemccand, line 3636: There is another PR about removing the not-needed `public` in interfaces I think.

mikemccand (approved):
Wow, thank you for the attention to detail @sabi0!

I'm curious: how did you uncover all of these corrections?  Is there some static analysis tool that you started from?

mikemccand:
I'll backport for 9.10 soon.

sabi0:
Thank you for checking this so quickly!

I started with a "Whitespace may be missing in string concatenation" inspection in IntelliJ IDEA.
After applying those fixes I went through the changed files again fixing file-level problems reported by the default inspections profile.

mikemccand:
> I started with a "Whitespace may be missing in string concatenation" inspection in IntelliJ IDEA.  After applying those fixes I went through the changed files again fixing file-level problems reported by the default inspections profile.

Cool!  Thanks for your attention to detail.  I wish we could somehow fold these checks into Lucene's static code analysis tooling...

sabi0:
https://www.jetbrains.com/qodana/

Open source projects can get a free license for any JetBrains products:
https://www.jetbrains.com/community/opensource/#support

PR #12969, opened 2023-12-24
Fix broken testAllVersionHaveCfsAndNocfs()

sabi0:


lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestBackwardsCompatibility.java
```diff
@@ -731,6 +732,7 @@ public void testAllVersionHaveCfsAndNocfs() {
         String prefix = prevFile.replace("-cfs", "");
         assertEquals("Missing -nocfs for backcompat index " + prefix, prefix + "-nocfs", file);
       }
+      prevFile = file;
```
sabi0, line 735: The assert in the loop was never executed because this line was missing.

PR #12972, opened 2023-12-24
Remove obsolete 'mappingRules' in Tokenizer tests

sabi0:


lucene/analysis/common/src/test/org/apache/lucene/analysis/wikipedia/TestWikipediaTokenizerFactory.java
```diff
@@ -145,19 +140,17 @@ public void testBogusArguments() throws Exception {
...
+                tokenizerFactory(WIKIPEDIA, "bogusArg", "bogusValue")
+                    .create(newAttributeFactory()));
     assertTrue(expected.getMessage().contains("Unknown parameters"));
   }
 
   public void testIllegalArguments() throws Exception {
     IllegalArgumentException expected =
         expectThrows(
             IllegalArgumentException.class,
-            () -> {
```
dweiss, line 158: Sometimes those extra brackets make the code easier to read. Please avoid to urge to just apply
everything intellij says can be "fixed". In this case (and the one above), I'd pick the bracketed version - easier on the eyes, especially after the code is tidied up.

sabi0: Thank you for this remark.
I try to be rather picky about anything IntelliJ says. But sometimes the code I see when working on a file is further changed by `gradlew tidy`. And I overlooked the unfortunate line split in the above case.

dweiss: > I try to be rather picky about anything IntelliJ says.

It's a very useful tool. But I wouldn't do automatically - many of the refactorings it suggests are subjective (whoever devised their code inspections was a human, after all). They're occasionally also wrong, introducing bugs - I've had that more than one time in my life (escape or visibility analysis gone wrong, for example).

PR #12979, opened 2023-12-25
Replace "UTF-8" with StandardCharsets.UTF_8

sabi0:


lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -409,8 +408,8 @@ public static final class IndexSortStatus {
     }
 
     /** Status from testing soft deletes */
-    public static final class SoftDeletsStatus {
-      SoftDeletsStatus() {}
+    public static final class SoftDeletesStatus {
```
dweiss, line 411: Just FYI - I know it's tempting to do such cleanups as you browse the files for other things but it makes reviewing more complex. It'd be nice if the patches were focused on one thing - if it says replace "utf-8" constant with StandardCharsets.UTF-8, it'd be easier to review and apply if the patch didn't do other things, like renaming public classes, removing final, or access scope changes.

dweiss: I think a single PR for one type of "mechanical" changes would be nice. Those changes apply to one aspect of the source code but are not likely to touch the API or the runtime behavior. For example, this PR could only touch the UTF-8 refactoring - even though the patch affects a large number of files, it's simpler to skim through them and perhaps cherry pick it to another branch. When things are aggregated, it may be more difficult to untangle (or miss something important among trivial replacements). 

Thank you for doing these cleanups, they're definitely needed and there's lots of cruft left over the years that could be improved (or modernized to new APIs).

PR #12980, opened 2023-12-26
Make FSTPostingFormat to build FST off-heap

dungba88:
### Description

This is an attempt to make FSTPostingFormat to write the FST off-heap. Instead of write it on-heap then save to disk, we configure the compiler to write the FST off-heap right from the start.

Some additional changes:
- As we can't write the FST metadata and FST data on the same file, now we need to break the `tfp` file into 2 files: `tfp.meta` and `tfp.data`
- We need to write the starting address of the FST data in the posting metadata file, then seek to that address when read

An alternative way is to copy the written FST back into metadata file, but that will slow down the writing.


lucene/codecs/src/java/org/apache/lucene/codecs/memory/FSTTermsWriter.java
```diff
@@ -278,6 +298,7 @@ public void finish(long sumTotalTermFreq, long sumDocFreq, int docCount) throws
       // save FST dict
       if (numTerms > 0) {
         final FST<FSTTermOutputs.TermData> fst = fstCompiler.compile();
+        fst.saveMetadata(metaOut);
```
dungba88, line 301: This seems to be incorrect, we don't need to save the metadata here

dungba88:
I temporarily made this PR to just split the meta and data into 2 files, but still use on-heap DataOutput. Seems like the test passed.

lucene/codecs/src/java/org/apache/lucene/codecs/memory/FSTTermsWriter.java
```diff
@@ -187,33 +200,38 @@ public void write(Fields fields, NormsProducer norms) throws IOException {
...
-            out.writeVLong(field.sumTotalTermFreq);
+            metaOut.writeVLong(field.sumTotalTermFreq);
           }
-          out.writeVLong(field.sumDocFreq);
-          out.writeVInt(field.docCount);
-          field.dict.save(out, out);
+          metaOut.writeVLong(field.sumDocFreq);
+          metaOut.writeVInt(field.docCount);
+          // write the starting file pointer
+          metaOut.writeVLong(dataOut.getFilePointer());
```
dungba88, line 220: Oh I think I found the bug, and why using an on-heap DataOutput works.

This `close()` method is called after the FST for all fields have been saved (streamed), and thus `dataOut.getFilePointer()` always points to the same pointer (EOF).

dungba88:
I found another quite tricky issue:

If we write the FST directly to the IndexOutput, there might be a chance that there's no term accepted by the FST, in that case we still write the padding 0 byte. This padding byte is to ensure no node having the 0 address: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/fst/FSTCompiler.java#L172-L174

However, since we are writing the FST consecutively for each field, appending to the same file, that means there could be a case we still write that additional padding byte, which is mapped to no field:
[ FST_Field_1 ] [ 0 ] [ FST_Field_2 ]

dungba88:
There is only 1 failed test left: TestFSTPostingFormat.testRandomException

```
   >         Caused by:
   >         java.lang.RuntimeException: unclosed IndexOutput: _s_FST50_0.tfp.meta
   >             at org.apache.lucene.tests.store.MockDirectoryWrapper.addFileHandle(MockDirectoryWrapper.java:783)
```

Seems like some file might not be closed correctly when there are exception

dungba88:
Fixed the above unclosed issue by moving `openInput` and `createOutput` to try-catch block

dungba88:
I added the change log, increased the FSTPostingsFormat version (isn't entirely related to this PR, but it seems the naming convention is outdated). The change for FSTCompiler can be merged as part of #12981 first. Will published the PR when the tests finished

dungba88:
I'm not sure why FSTPostingsFormat is different from the rest, that it write both the metadata and data to the same file. I think writing to separate files would be cleaner and more consistent.

dungba88:
Thanks @mikemccand for the clarification!

Do you think we should still make this change? One benefit is that it can be used for reference. Otherwise I'll close this PR

mikemccand:
> Thanks @mikemccand for the clarification!
> 
> Do you think we should still make this change? One benefit is that it can be used for reference. Otherwise I'll close this PR

Yes I think we should!  The FSTs this postings format creates can easily be massive, so they really should build off-heap now that Lucene FST impl has this capability (thank you Tantivy for this inspiration!).

dungba88:
Thanks @mikemccand! I rebased the PR with the latest Lucene 11. One question I have is that do we need to maintain the backward compatibility for this format?

PR #12981, opened 2023-12-26
Lazily write the FST padding byte

dungba88:
### Description

Lazily write the FST padding byte, so that in case the FST is empty (no accepted nodes) nothing will be written. This is important for off-heap writing, as we don't want to add that extra byte when the FST would be thrown away. Found while working on https://github.com/apache/lucene/pull/12980

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/fst/FSTCompiler.java
+++ b/lucene/core/src/java/org/apache/lucene/util/fst/FSTCompiler.java
@@ -109,6 +109,9 @@ public class FSTCompiler<T> {
 
   private final IntsRefBuilder lastInput = new IntsRefBuilder();
 
+  // indicates whether we are not yet to write the padding byte
+  private boolean paddingBytePending;
+
   // NOTE: cutting this over to ArrayList instead loses ~6%
   // in build performance on 9.8M Wikipedia terms; so we
   // left this as an array:
@@ -171,6 +174,7 @@ public class FSTCompiler<T> {
     // pad: ensure no node gets address 0 which is reserved to mean
     // the stop state w/ no arcs. the actual byte will be written lazily
     numBytesWritten++;
+    paddingBytePending = true;
     this.dataOutput = dataOutput;
     fst =
         new FST<>(
@@ -551,7 +555,7 @@ public class FSTCompiler<T> {
 
     reverseScratchBytes();
     // write the padding byte if needed
-    if (numBytesWritten == 1) {
+    if (paddingBytePending) {
       writePaddingByte();
     }
     scratchBytes.writeTo(dataOutput);
@@ -561,9 +565,14 @@ public class FSTCompiler<T> {
     return numBytesWritten - 1;
   }
 
+  /**
+   * Write the padding byte, ensure no node gets address 0 which is reserved to mean the stop state
+   * w/ no arcs
+   */
   private void writePaddingByte() throws IOException {
-    assert numBytesWritten == 1;
+    assert paddingBytePending;
     dataOutput.writeByte((byte) 0);
+    paddingBytePending = false;
   }
 
   private void writeLabel(DataOutput out, int v) throws IOException {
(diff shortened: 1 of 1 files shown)


mikemccand:
Egads, thank you bot!  This one had already fallen past the event horizon of my email box.  I'll try to review soon.

lucene/core/src/java/org/apache/lucene/util/fst/FSTCompiler.java
```diff
@@ -164,14 +164,12 @@ private FSTCompiler(
...
-      int version)
-      throws IOException {
+      int version) {
     this.allowFixedLengthArcs = allowFixedLengthArcs;
     this.directAddressingMaxOversizingFactor = directAddressingMaxOversizingFactor;
     this.version = version;
     // pad: ensure no node gets address 0 which is reserved to mean
-    // the stop state w/ no arcs
-    dataOutput.writeByte((byte) 0);
+    // the stop state w/ no arcs. the actual byte will be written lazily
```
mikemccand, line 172: Hmm could we instead add a boolean `paddingBytePending` or so?  Set it to true, here, then when the byte is lazily written, set it to false, write the byte, and increment `numBytesWritten` at that point?  Otherwise it's sort of weird to increment `numBytesWritten` when we didn't actually write the byte yet?

lucene/core/src/java/org/apache/lucene/util/fst/FSTCompiler.java
```diff
@@ -968,6 +975,9 @@ public FST<T> compile() throws IOException {
     if (root.numArcs == 0) {
       if (fst.metadata.emptyOutput == null) {
         return null;
```
mikemccand, line 977: Maybe add a comment that this means a completely empty FST?  Accepts nothing?

lucene/core/src/java/org/apache/lucene/util/fst/FSTCompiler.java
```diff
@@ -552,13 +550,22 @@ long addNode(FSTCompiler.UnCompiledNode<T> nodeIn) throws IOException {
...
+    }
     scratchBytes.writeTo(dataOutput);
     numBytesWritten += scratchBytes.getPosition();
 
     nodeCount++;
     return numBytesWritten - 1;
   }
 
+  private void writePaddingByte() throws IOException {
+    assert numBytesWritten == 1;
```
mikemccand, line 565: Maybe add a comment explaining what this padding byte even is for?  I myself cannot remember :)

mikemccand (commented):
Thanks @dungba88 -- I left a few small comments.

How serious is this, during off-heap writing?  Are we using off-heap writing anywhere in Lucene yet (e.g. block tree)?

lucene/core/src/java/org/apache/lucene/util/fst/FSTCompiler.java
dungba88: numBytesWritten is used for determine the address of the to be written nodes, so if we don't increment it here, it would mess up the address.

dungba88:
> How serious is this, during off-heap writing? Are we using off-heap writing anywhere in Lucene yet (e.g. block tree)?

This will break the off-heap writing, as we are writting byte which won't belong to any FST, and it would mess up subsequent reads (if multiple FST are written to the same file)

We are not using off-heap, but I have 2 other PR to start doing that:
- https://github.com/apache/lucene/pull/12980
- https://github.com/apache/lucene/pull/12985

mikemccand:
> > How serious is this, during off-heap writing? Are we using off-heap writing anywhere in Lucene yet (e.g. block tree)?
> 
> This will break the off-heap writing, as we are writting byte which won't belong to any FST, and it would mess up subsequent reads (if multiple FST are written to the same file)

Oh no, I see.  Hmm, well could we increment the `numBytesWritten` but use a separately boolean indicating that the lazy byte hasn't been written?  It's spooky to have `numBytesWritten = 1` mean it may or may not have been written.

mikemccand:
> We are not using off-heap, but I have 2 other PR to start doing that:

OK, phew, thanks.  Those will be next up :)

dungba88:
Thanks @mikemccand for reviewing, I've addressed those in the latest revision.

mikemccand (approved):
Thanks, it looks great @dungba88, I'll merge!

PR #12982, opened 2023-12-26
Get rid of deprecated assertThat() usages

sabi0:


lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestSuggestField.java
```diff
@@ -485,7 +470,7 @@ public void testNRTDeletedDocFiltering() throws Exception {
     PrefixCompletionQuery query =
         new PrefixCompletionQuery(analyzer, new Term("suggest_field", "abc_"));
     TopSuggestDocs suggest = indexSearcher.suggest(query, numLive, false);
-    assertSuggestions(suggest, expectedEntries.toArray(new Entry[expectedEntries.size()]));
+    assertSuggestions(suggest, expectedEntries.toArray(new Entry[0]));
```
dweiss, line 473: This is a weird idiom. Could we use the Entry::new instead? It's short and, I think, more intuitive.

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestSuggestField.java
```diff
@@ -685,8 +669,8 @@ public void testMultipleSegments() throws Exception {
     PrefixCompletionQuery query =
         new PrefixCompletionQuery(analyzer, new Term("suggest_field", "abc_"));
     TopSuggestDocs suggest =
-        indexSearcher.suggest(query, (entries.size() == 0) ? 1 : entries.size(), false);
-    assertSuggestions(suggest, entries.toArray(new Entry[entries.size()]));
+        indexSearcher.suggest(query, entries.isEmpty() ? 1 : entries.size(), false);
+    assertSuggestions(suggest, entries.toArray(new Entry[0]));
```
dweiss, line 673: The same?

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestSuggestField.java
```diff
@@ -766,12 +749,12 @@ public void testScoring() throws Exception {
           assertTrue(topScore >= scoreDoc.score);
         }
         topScore = scoreDoc.score;
-        assertThat((float) mappings.get(scoreDoc.key.toString()), equalTo(scoreDoc.score));
+        assertEquals(scoreDoc.score, (float) mappings.get(scoreDoc.key.toString()), 1.0e-8);
```
dweiss, line 752: I'm not a fan of those magic constants or deltas in assertions in particular... Maybe it'd be better to add assertEquals(float, float) to LuceneTestCase and do a Float.compare assertion there (with proper exception in case of a mismatch)?

dweiss (commented):
The deprecation of assertThat in junit points at the same method in MatcherAssert (hamcrest) - you've changed the methods to junit assertions instead, which I actually like more (I never got used to hamcrest conventions, assertj is closer to my heart).

I left two or three comments, but I think it's fine to go in.

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestSuggestField.java
sabi0: I am not sure how do you expect `Entry::new` to be used here?
`toArray()` accepts an array and not an allocator.

As for the 0 size - it's more performant and generally safer since Java 6 than an array with a pre-calculated size.
See https://shipilev.net/blog/2016/arrays-wisdom-ancients/

dweiss: Sorry, I though it was a stream. I know that but performance analysis but for one-offs such as this one the performance gain is non-existent and the idiom is awkward. Maybe we can change assertSuggestions to accept whatever expectedEntries are (is it a collection, iterable?); I don't have the code in front of me.

sabi0: Currently it's a vararg method allowing both
```
assertSuggestions(suggest, entries.toArray(new Entry[0]));
```
and
```
assertSuggestions(suggest, new Entry("abc_" + num, num));
```

I could change it to accept a Collection and wrap all the single-Entry calls with `List.of()`.
Would that be ok?

dweiss: Yep, I think that'd be nicer.

sabi0: I had a closer look at the `assertSuggestions()` method. It has 56 usages in 6 different classes.
So here are the options I see:
1. return pre-calculated array sizes and keep this part as is
2. overload the method to accept either `Entry...` or `Collection<Entry>` and update the 13 usages in the `TestSuggestField` (leaving the other classes for later)
3. proceed with the refactoring to `Collection<Entry>` and update all 6 classes (in a separate PR)

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestSuggestField.java
sabi0: There are 33 usages of `assertEquals(String, double, double, double)` and 653 usages of `assertEquals(String, float, float, float)` across the project.
The deltas vary considerably, from 0 to 0.0001 to 0.0000001 to 1e-10.

Maybe I should just use 0 delta in this PR?

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestSuggestField.java
dweiss: Given the number of use cases, I'd go with option 3 or 1 - maybe I'm overly sensitive about that new Foo[0]...

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestSuggestField.java
dweiss: Yep, please do. Seems like this is what it was before (without any delta). Let's see if CI passes.

lucene/core/src/test/org/apache/lucene/geo/TestTessellator.java
```diff
@@ -841,20 +839,19 @@ public void testComplexPolygon50() throws Exception {
...
-    assertThat("Only one polygon", polygons.length, equalTo(1));
+    assertEquals("Only one polygon", 1, polygons.length);
     Polygon polygon = polygons[0];
     TestCountingMonitor monitor = new TestCountingMonitor();
     Tessellator.tessellate(polygon, true, monitor);
-    assertThat("Expected many monitor calls", monitor.count, greaterThan(390));
-    assertThat("Expected specific number of splits", monitor.splitsStarted, equalTo(3));
-    assertThat(
-        "Expected splits to start and end", monitor.splitsStarted, equalTo(monitor.splitsEnded));
+    assertTrue("Expected many monitor calls", monitor.count > 390);
```
dweiss, line 846: The downside of converting to an explicit comparison is that you won't know what the variable (monitor.count, etc.) was. Maybe these should be converted to hamcrest MatcherAssert.assertThat(...), just as JUnit deprecation message suggests?

sabi0: Done.

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestSuggestField.java
sabi0: Done.

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestSuggestField.java
sabi0: Reverted to pre-calculated array sizes.

dweiss:
Thank you! Please use regular merges, not force-pushes, unless absolutely necessary. A linear merge/change commit is easier to follow-up to (compared to a new forced commit). All pull requests are typically squashed into a single commit at the end, so intermediate commits are only preserved in pr history.

PR #12987, opened 2023-12-29
Output binary doc values as hex array in SimpleTextCodec

msfroh:
### Description

Binary doc values were being written directly in SimpleTextCodec, though they may not be valid UTF-8 (i.e. they may not be "text"). This change encodes them as a string representing an array of hexadecimal bytes.

### DIFF
--- a/lucene/codecs/src/java/org/apache/lucene/codecs/simpletext/SimpleTextDocValuesReader.java
+++ b/lucene/codecs/src/java/org/apache/lucene/codecs/simpletext/SimpleTextDocValuesReader.java
@@ -333,12 +333,7 @@ class SimpleTextDocValuesReader extends DocValuesProducer {
               termByteArray.grow(len);
               termByteArray.setLength(len);
               in.readBytes(termByteArray.bytes(), 0, len);
-              if (len > 2) {
-                term.copyBytes(
-                    SimpleTextUtil.fromBytesRefString(termByteArray.get().utf8ToString()));
-              } else {
-                term.setLength(0);
-              }
+              term.copyBytes(SimpleTextUtil.fromBytesRefString(termByteArray.get().utf8ToString()));
               return term.get();
             } catch (IOException ioe) {
               throw new RuntimeException(ioe);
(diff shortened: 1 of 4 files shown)


lucene/codecs/src/java/org/apache/lucene/codecs/simpletext/SimpleTextDocValuesReader.java
```diff
@@ -329,9 +330,15 @@ public BytesRef apply(int docID) {
...
-              in.readBytes(term.bytes(), 0, len);
+              termByteArray.grow(len);
+              termByteArray.setLength(len);
+              in.readBytes(termByteArray.bytes(), 0, len);
+              if (len > 2) {
+                term.copyBytes(
+                    SimpleTextUtil.fromBytesRefString(termByteArray.get().utf8ToString()));
+              } else {
+                term.setLength(0);
+              }
```
jpountz, line 341: Why do we need this condition? Can't `fromBytesRefString` read empty arrays?

msfroh: ~~The issue is that on the writing side, if `stringVal == null`, we write the length as `0` and don't output anything. So, `termByteArray.get().utf8ToString()` is the empty string and `fromBytesRefString` throws an exception on that. It was showing up as a unit test failure.~~

~~I added this condition here, but I suppose it would also be fine to output `[]` followed by `F` for missing binary values (though we'll waste a couple of bytes). Of course, now I'm wondering how we get here if the doc doesn't have the given field...~~

Nope... after debugging it, it wasn't the `null` case (which seems to work fine and skips any doc where the "has value" is `F`). It turns out that `fromBytesRefString` **can't** read empty arrays. The logic extracts from within the `[]`, then tries to split by `,`. The result is the empty string, so it thinks it has 1 "part".

I'll follow up with a fix for `fromBytesRefString` instead.

msfroh:
Test failure is reproducible on main: https://github.com/apache/lucene/issues/13009

PR #12995, opened 2024-01-06
Split taxonomy arrays across chunks

### LINKED ISSUE #12989: Taxonomy facets: can we change massive `int[]` for parent/child/sibling tree to paged/block `int[]` to reduce RAM pressure?
### Description

At Amazon product search we use taxonomy facets for the facet filtering we show customers, but this causes high RAM pressure on every refresh as the current implementation allocates a new massive `int[]` on each refresh, requiring ~2X the transient RAM usage until the old taxonomy reader is fully closed / dereferenced, causing us to over-size our heaps just to handle this short RAM surge.

Yet, on each refresh, all that really is happening is a few new ints might be effectively appended to the end of the old `int[]` -- it is inherently a write once append only data structure.  Reallocating the full massive `int[]` every time is silly.

I think we could switch to a paged `int[]` structure?  This way the new reader could share nearly all of the old `int[]` pages, and only make a new last page to hold the few new append ints?

msfroh:
### Description

Taxonomy ordinals are added in an append-only way.

Instead of reallocating a single big array when loading new taxonomy ordinals and copying all the values from the previous arrays over individually, we can keep blocks of ordinals and reuse blocks from the previous arrays.

Resolves https://github.com/apache/lucene/issues/12989

### DIFF
--- a/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
+++ b/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
@@ -109,13 +109,19 @@ class TaxonomyIndexArrays extends ParallelTaxonomyArrays implements Accountable 
     if (size == 0) {
       return new int[0][];
     }
-    int chunkCount = size / CHUNK_SIZE + 1;
+    int chunkCount = size >> CHUNK_SIZE_BITS;
+    int fullChunkCount;
     int lastChunkSize = size & CHUNK_MASK;
+    if (lastChunkSize == 0) {
+      fullChunkCount = chunkCount; // Size is a multiple of CHUNK_SIZE, so all arrays are full-sized
+    } else {
+      fullChunkCount = chunkCount++;
+    }
     int[][] array = new int[chunkCount][];
-    if (array.length > 0) {
-      for (int i = startFrom; i < chunkCount - 1; i++) {
-        array[i] = new int[CHUNK_SIZE];
-      }
+    for (int i = startFrom; i < fullChunkCount; i++) {
+      array[i] = new int[CHUNK_SIZE];
+    }
+    if (lastChunkSize > 0) {
       array[chunkCount - 1] = new int[lastChunkSize];
     }
     return array;
@@ -208,7 +214,7 @@ class TaxonomyIndexArrays extends ParallelTaxonomyArrays implements Accountable 
               "Missing parent data for category " + (doc + leafContext.docBase), reader.toString());
         }
         int pos = doc + leafContext.docBase;
-        parentsArray[pos / CHUNK_SIZE][pos % CHUNK_SIZE] =
+        parentsArray[pos >> CHUNK_SIZE_BITS][pos & CHUNK_MASK] =
             Math.toIntExact(parentValues.longValue());
       }
     }
@@ -226,7 +232,7 @@ class TaxonomyIndexArrays extends ParallelTaxonomyArrays implements Accountable 
           allocateChunkedArray(
               ArrayUtil.oversize(ordinal + 1, Integer.BYTES), parents.values.length - 1);
       copyChunkedArray(parents.values, newParents);
-      newParents[ordinal / CHUNK_SIZE][ordinal % CHUNK_SIZE] = parentOrdinal;
+      newParents[ordinal >> CHUNK_SIZE_BITS][ordinal & CHUNK_MASK] = parentOrdinal;
       return new TaxonomyIndexArrays(newParents);
     }
     parents.set(ordinal, parentOrdinal);
--- a/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
+++ b/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
@@ -39,7 +39,7 @@ import org.apache.lucene.util.RamUsageEstimator;
  */
 class TaxonomyIndexArrays extends ParallelTaxonomyArrays implements Accountable {
   private static final int CHUNK_SIZE_BITS = 13;
-  private static final int CHUNK_SIZE = 1 << CHUNK_SIZE_BITS;
+  static final int CHUNK_SIZE = 1 << CHUNK_SIZE_BITS;
   private static final int CHUNK_MASK = CHUNK_SIZE - 1;
 
   private final ChunkedIntArray parents;
@@ -51,8 +51,8 @@ class TaxonomyIndexArrays extends ParallelTaxonomyArrays implements Accountable 
   private volatile boolean initializedChildren = false;
   private ChunkedIntArray children, siblings;
 
-  private static class ChunkedIntArray extends ParallelTaxonomyArrays.IntArray {
-    private final int[][] values;
+  static class ChunkedIntArray extends ParallelTaxonomyArrays.IntArray {
+    final int[][] values;
 
     private ChunkedIntArray(int[][] values) {
       this.values = values;
@@ -73,8 +73,8 @@ class TaxonomyIndexArrays extends ParallelTaxonomyArrays implements Accountable 
     }
   }
 
-  /** Used by {@link #add(int, int)} after the array grew. */
-  private TaxonomyIndexArrays(int[][] parents) {
+  /** Used by {@link #add(int, int)} after the array grew. Also, used for testing. */
+  TaxonomyIndexArrays(int[][] parents) {
     this.parents = new ChunkedIntArray(parents);
   }
 
@@ -244,7 +244,7 @@ class TaxonomyIndexArrays extends ParallelTaxonomyArrays implements Accountable 
    * {@code i}.
    */
   @Override
-  public IntArray parents() {
+  public ChunkedIntArray parents() {
     return parents;
   }
 
@@ -254,7 +254,7 @@ class TaxonomyIndexArrays extends ParallelTaxonomyArrays implements Accountable 
    * taxonomy as an immediate child of {@code i}.
    */
   @Override
-  public IntArray children() {
+  public ChunkedIntArray children() {
     if (!initializedChildren) {
       initChildrenSiblings(null);
     }
@@ -268,7 +268,7 @@ class TaxonomyIndexArrays extends ParallelTaxonomyArrays implements Accountable 
    * {@code i}. The sibling is defined as the previous youngest child of {@code parents[i]}.
    */
   @Override
-  public IntArray siblings() {
+  public ChunkedIntArray siblings() {
     if (!initializedChildren) {
       initChildrenSiblings(null);
     }
--- a/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
+++ b/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
@@ -171,10 +171,13 @@ class TaxonomyIndexArrays extends ParallelTaxonomyArrays implements Accountable 
     }
 
     for (int i = first; i < length; i++) {
+      int parent = parents.get(i);
+      // The existing youngest child of the parent is the next older sibling of i.
       // note that parents[i] is always < i, so the right-hand-side of
       // the following line is already set when we get here
-      siblings.set(i, children.get(parents.get(i)));
-      children.set(parents.get(i), i);
+      siblings.set(i, children.get(parent));
+      // The new youngest child of the parent is i.
+      children.set(parent, i);
     }
   }
 
--- a/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
+++ b/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
@@ -38,49 +38,53 @@ import org.apache.lucene.util.RamUsageEstimator;
  * @lucene.experimental
  */
 class TaxonomyIndexArrays extends ParallelTaxonomyArrays implements Accountable {
-  private static final int CHUNK_SIZE = 8192;
+  private static final int CHUNK_SIZE_BITS = 13;
+  private static final int CHUNK_SIZE = 1 << CHUNK_SIZE_BITS;
+  private static final int CHUNK_MASK = CHUNK_SIZE - 1;
 
-  private final ChunkedArray parents;
+  private final ChunkedIntArray parents;
 
   // the following two arrays are lazily initialized. note that we only keep a
   // single boolean member as volatile, instead of declaring the arrays
   // volatile. the code guarantees that only after the boolean is set to true,
   // the arrays are returned.
   private volatile boolean initializedChildren = false;
-  private ChunkedArray children, siblings;
+  private ChunkedIntArray children, siblings;
 
-  private static class ChunkedArray extends ParallelTaxonomyArrays.IntArray {
+  private static class ChunkedIntArray extends ParallelTaxonomyArrays.IntArray {
     private final int[][] values;
 
-    private ChunkedArray(int[][] values) {
+    private ChunkedIntArray(int[][] values) {
       this.values = values;
     }
 
     @Override
     public int get(int i) {
-      return values[i / CHUNK_SIZE][i % CHUNK_SIZE];
+      return values[i >> CHUNK_SIZE_BITS][i & CHUNK_MASK];
     }
 
     public void set(int i, int val) {
-      values[i / CHUNK_SIZE][i % CHUNK_SIZE] = val;
+      values[i >> CHUNK_SIZE_BITS][i & CHUNK_MASK] = val;
     }
 
     @Override
     public int length() {
-      return (values.length - 1) * CHUNK_SIZE + values[values.length - 1].length;
+      return ((values.length - 1) << CHUNK_SIZE_BITS) + values[values.length - 1].length;
     }
   }
 
   /** Used by {@link #add(int, int)} after the array grew. */
   private TaxonomyIndexArrays(int[][] parents) {
-    this.parents = new ChunkedArray(parents);
+    this.parents = new ChunkedIntArray(parents);
   }
 
   public TaxonomyIndexArrays(IndexReader reader) throws IOException {
-    int[][] parentArray = allocateChunkedArray(reader.maxDoc());
-    initParents(parentArray, reader, 0);
-    parentArray[0][0] = TaxonomyReader.INVALID_ORDINAL;
-    parents = new ChunkedArray(parentArray);
+    int[][] parentArray = allocateChunkedArray(reader.maxDoc(), 0);
+    if (parentArray.length > 0) {
+      initParents(parentArray, reader, 0);
+      parentArray[0][0] = TaxonomyReader.INVALID_ORDINAL;
+    }
+    parents = new ChunkedIntArray(parentArray);
   }
 
   public TaxonomyIndexArrays(IndexReader reader, TaxonomyIndexArrays copyFrom) throws IOException {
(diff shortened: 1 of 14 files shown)


lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
```diff
@@ -98,26 +144,28 @@ private void computeChildrenSiblings(int first) {
...
+      siblings.set(0, TaxonomyReader.INVALID_ORDINAL);
     }
 
-    for (int i = first; i < parents.length; i++) {
+    for (int i = first; i < length; i++) {
       // note that parents[i] is always < i, so the right-hand-side of
       // the following line is already set when we get here
-      siblings[i] = children[parents[i]];
-      children[parents[i]] = i;
+      siblings.set(i, children.get(parents.get(i)));
```
stefanvodita, line 161: Let's extract out `parents.get(i)` to a new variable, since it's a bit more than a simple array lookup now, and it's used twice.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
```diff
@@ -68,25 +90,49 @@ public TaxonomyIndexArrays(IndexReader reader, TaxonomyIndexArrays copyFrom) thr
...
+    return array;
+  }
+
+  private static void copyChunkedArray(int[][] oldArray, int[][] newArray) {
+    // Copy all but the last (maybe partial) chunk from the old array
+    if (oldArray.length > 1) {
+      System.arraycopy(oldArray, 0, newArray, 0, oldArray.length - 1);
+    }
+    int[] lastCopyChunk = oldArray[oldArray.length - 1];
+    System.arraycopy(lastCopyChunk, 0, newArray[oldArray.length - 1], 0, lastCopyChunk.length);
```
stefanvodita, line 121: Why are we making this a deep copy while the other arrays are shallow copies?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
```diff
@@ -68,25 +90,49 @@ public TaxonomyIndexArrays(IndexReader reader, TaxonomyIndexArrays copyFrom) thr
     // it may be caused if e.g. the taxonomy segments were merged, and so an updated
     // NRT reader was obtained, even though nothing was changed. this is not very likely
     // to happen.
-    int[] copyParents = copyFrom.parents();
-    this.parents = new int[reader.maxDoc()];
-    System.arraycopy(copyParents, 0, parents, 0, copyParents.length);
-    initParents(reader, copyParents.length);
-
+    int[][] parentArray = allocateChunkedArray(reader.maxDoc());
```
stefanvodita, line 93: Hmm, maybe I'm missing something, but if we're allocating here are we any better off than before? Most of this memory will be discarded anyway when we shallowly copy the old arrays.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
```diff
@@ -68,25 +90,49 @@ public TaxonomyIndexArrays(IndexReader reader, TaxonomyIndexArrays copyFrom) thr
...
+    }
+    int[] lastCopyChunk = oldArray[oldArray.length - 1];
+    System.arraycopy(lastCopyChunk, 0, newArray[oldArray.length - 1], 0, lastCopyChunk.length);
+  }
+
   private synchronized void initChildrenSiblings(TaxonomyIndexArrays copyFrom) {
     if (!initializedChildren) { // must do this check !
-      children = new int[parents.length];
-      siblings = new int[parents.length];
+      int[][] childrenArray = allocateChunkedArray(parents.length());
```
stefanvodita, line 126: Similar comment as in the constructor - are we getting any memory back by allocating like this and the copying?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
```diff
@@ -153,12 +203,13 @@ private void initParents(IndexReader reader, int first) throws IOException {
    * <p><b>NOTE:</b> you should call this method from a thread-safe code.
    */
   TaxonomyIndexArrays add(int ordinal, int parentOrdinal) {
-    if (ordinal >= parents.length) {
-      int[] newarray = ArrayUtil.grow(parents, ordinal + 1);
-      newarray[ordinal] = parentOrdinal;
-      return new TaxonomyIndexArrays(newarray);
+    if (ordinal >= parents.length()) {
+      int[][] newParents = allocateChunkedArray(ArrayUtil.oversize(ordinal + 1, Integer.BYTES));
```
stefanvodita, line 207: Won't this make `parents.length()` invalid since we have empty space at the end of the chunked array and `length()` just looks at the size of the chunked array?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
```diff
@@ -38,27 +38,49 @@
...
+    this.parents = new ChunkedArray(parents);
   }
 
   public TaxonomyIndexArrays(IndexReader reader) throws IOException {
-    parents = new int[reader.maxDoc()];
-    if (parents.length > 0) {
-      initParents(reader, 0);
-      parents[0] = TaxonomyReader.INVALID_ORDINAL;
-    }
+    int[][] parentArray = allocateChunkedArray(reader.maxDoc());
```
stefanvodita, line 80: Is it better to work with `int[][]` in `TaxonomyArrays` rather than add support to the chunked array for the operations we want and only work with it?

stefanvodita (changes requested):

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
```diff
@@ -38,27 +38,49 @@
...
+  private static class ChunkedArray extends ParallelTaxonomyArrays.IntArray {
+    private final int[][] values;
+
+    private ChunkedArray(int[][] values) {
+      this.values = values;
+    }
+
+    @Override
+    public int get(int i) {
+      return values[i / CHUNK_SIZE][i % CHUNK_SIZE];
```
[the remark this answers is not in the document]
msfroh: That's a great idea!

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
```diff
@@ -68,25 +90,49 @@ public TaxonomyIndexArrays(IndexReader reader, TaxonomyIndexArrays copyFrom) thr
...
+    copyChunkedArray(copyFrom.parents.values, parentArray);
+    initParents(parentArray, reader, copyFrom.parents.length());
+    parents = new ChunkedArray(parentArray);
     if (copyFrom.initializedChildren) {
       initChildrenSiblings(copyFrom);
     }
   }
 
+  private static int[][] allocateChunkedArray(int size) {
+    int chunkCount = size / CHUNK_SIZE + 1;
```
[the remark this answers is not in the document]
msfroh: In practice, we always have `size >= 1` (because the `TaxonomyWriter` writes the root ordinal on startup) -- but  relying on that behavior may bite us. I'll add a check for the `size == 0` case.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
msfroh: The last chunk is the only one that changes -- the other chunks can be reused.

Suppose the old array had size 20,000. It would look like:

```
0 -> 8192 element chunk (c0)
1 -> 8192 element chunk (c1)
2 -> 3616 element chunk (c2)
```

Suppose  I want to add 10 more elements. In the old way, that would mean allocating a whole new 20,010 element array and copying over the 20,000 elements from the original array. In the new, chunked world, we allocate a new 3-element `int[][]`, and populate it like:

```
0 -> c0 (reusing the old array)
1 -> c1 (reusing the old array)
2 -> c2' (new 3626-element array).
```

Now I copy over the 3616 elements from `c2` to `c2'`. The total "temporary" extra space is 3 references for the new top-level array, plus 3616 integers (since we're holding onto `c2` until we've copied it). The old model needed to allocate a temporary 20,000. elements.

In general, where the old model required `n` temporary extra `int`s (so `4n` bytes plus a few bytes of array overhead), the new model needs `ceil(n/8192)` temporary pointers (the new `int[][]` array) plus up to 8192 extra `int`s for the last chunk.

In terms of minimizing waste, assuming references are 32-bit, I suppose the ideal chunk size is the square root of `n` (if references are 64-bit, we would be better off with bigger chunks). If we want to more efficiently handle more than 67 million ordinals, I guess we could set chunk size to something bigger than 8192. I just copied the value from `IntBlockPool`.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
msfroh: Oof -- you're right. That was a terrible attempt at code reuse on my part, and indeed it would make the whole thing useless. That's embarrassing!

I originally had some of the code from `allocateChunkedArray` in `copyChunkedArray` and I tried to make it reusable in the "fresh reader, no copying" use-case.

Thank you for catching that!

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
msfroh: This is actually consistent with the existing behavior:

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
msfroh: I was a little torn on this. 

I figured I would try to avoid method calls where possible (assuming method calls have more overhead than array access), and could happily muck about in the `ChunkedArray` internals within `TaxonomyIndexArrays`(since it's a private implementation detail).

We could move `allocateChunkedArray` and `copyChunkedArray` logic into `ChunkedArray` itself (and call them from `ChunkedArray` constructors). It's still in the same file, so I don't have strong opinions either way.

stefanvodita:
As far as testing, can we add some unit tests that allocate more than one chunk and exercise the new functionality?
Should we also run some benchmarks to understand if there's any sort of performance regression?

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
stefanvodita: Actually, I guess the real concern here is that the last array is not properly sized, so always deep-copying it makes sense, since it's capacity requirements will have always changed.

msfroh:
> As far as testing, can we add some unit tests that allocate more than one chunk and exercise the new functionality? Should we also run some benchmarks to understand if there's any sort of performance regression?

I can take care of the first part.

@stefanvodita, do you mind running the Lucene benchmarks against this change to see how it performs?

stefanvodita:
The results are in. I don't see any significant p-values (< 0.05).
`python3 src/python/localrun.py -source wikimediumall -r`

```
                       MedPhrase       27.55      (7.2%)       27.08      (5.8%)   -1.7% ( -13% -   12%) 0.408
                          IntNRQ       19.20     (10.6%)       18.89      (7.1%)   -1.6% ( -17% -   18%) 0.566
           BrowseMonthTaxoFacets        9.67      (8.9%)        9.52     (10.7%)   -1.6% ( -19% -   19%) 0.613
     BrowseRandomLabelTaxoFacets        4.01     (25.1%)        3.97     (26.0%)   -1.0% ( -41% -   66%) 0.902
                HighSloppyPhrase       28.64      (3.5%)       28.41      (2.4%)   -0.8% (  -6% -    5%) 0.406
                      AndHighLow      325.06      (3.8%)      322.67      (7.0%)   -0.7% ( -11% -   10%) 0.680
     BrowseRandomLabelSSDVFacets        2.92      (1.8%)        2.90      (2.6%)   -0.6% (  -4% -    3%) 0.403
       BrowseDayOfYearSSDVFacets        4.08      (3.8%)        4.06      (3.7%)   -0.6% (  -7% -    7%) 0.625
                         Prefix3      142.86      (5.0%)      142.32      (5.4%)   -0.4% ( -10% -   10%) 0.819
            HighTermTitleBDVSort        7.40      (6.3%)        7.38      (6.4%)   -0.3% ( -12% -   13%) 0.862
                 LowSloppyPhrase       11.47      (1.9%)       11.44      (1.7%)   -0.3% (  -3% -    3%) 0.657
                     LowSpanNear        6.86      (2.0%)        6.85      (1.8%)   -0.2% (  -3% -    3%) 0.754
                      HighPhrase       67.07      (3.1%)       66.94      (3.1%)   -0.2% (  -6% -    6%) 0.850
                       LowPhrase        7.77      (3.2%)        7.76      (3.4%)   -0.1% (  -6% -    6%) 0.888
                    HighSpanNear        3.81      (3.0%)        3.80      (2.8%)   -0.1% (  -5% -    5%) 0.875
            HighIntervalsOrdered        1.61      (2.5%)        1.61      (2.4%)   -0.1% (  -4% -    4%) 0.853
                      AndHighMed       76.92      (6.7%)       76.83      (8.0%)   -0.1% ( -13% -   15%) 0.959
                 MedSloppyPhrase       15.30      (1.9%)       15.29      (1.7%)   -0.1% (  -3% -    3%) 0.850
                      OrHighHigh       20.33      (7.3%)       20.31      (6.4%)   -0.1% ( -12% -   14%) 0.961
             LowIntervalsOrdered        7.12      (2.5%)        7.12      (2.4%)   -0.0% (  -4% -    4%) 0.955
         AndHighMedDayTaxoFacets       11.03      (2.2%)       11.04      (2.2%)    0.0% (  -4% -    4%) 0.959
                    OrNotHighMed      216.20      (2.6%)      216.28      (3.0%)    0.0% (  -5% -    5%) 0.967
             MedIntervalsOrdered       23.70      (1.8%)       23.71      (1.6%)    0.0% (  -3% -    3%) 0.944
                     MedSpanNear       13.33      (1.9%)       13.34      (1.7%)    0.1% (  -3% -    3%) 0.910
        AndHighHighDayTaxoFacets        7.73      (3.5%)        7.74      (3.2%)    0.1% (  -6% -    7%) 0.906
            MedTermDayTaxoFacets       15.49      (2.2%)       15.54      (1.6%)    0.3% (  -3% -    4%) 0.590
               HighTermTitleSort      121.92      (1.8%)      122.33      (1.6%)    0.3% (  -2% -    3%) 0.533
                      TermDTSort      142.74      (2.1%)      143.26      (3.4%)    0.4% (  -5% -    5%) 0.683
                        PKLookup      172.65      (3.0%)      173.29      (2.8%)    0.4% (  -5% -    6%) 0.687
                         Respell       43.69      (2.3%)       43.87      (2.7%)    0.4% (  -4% -    5%) 0.613
           HighTermDayOfYearSort      210.29      (2.8%)      211.25      (3.1%)    0.5% (  -5% -    6%) 0.626
                     AndHighHigh       23.87     (11.4%)       23.99     (11.7%)    0.5% ( -20% -   26%) 0.888
            BrowseDateSSDVFacets        1.19      (5.2%)        1.19      (4.2%)    0.6% (  -8% -   10%) 0.698
                        Wildcard      186.59      (2.6%)      187.68      (2.9%)    0.6% (  -4% -    6%) 0.505
                       OrHighMed       90.56      (5.3%)       91.09      (4.9%)    0.6% (  -9% -   11%) 0.715
                   OrNotHighHigh      354.86      (3.6%)      356.98      (3.9%)    0.6% (  -6% -    8%) 0.611
                   OrHighNotHigh      306.95      (3.8%)      308.88      (4.2%)    0.6% (  -7% -    9%) 0.620
                    OrHighNotMed      363.76      (3.9%)      366.46      (4.0%)    0.7% (  -6% -    8%) 0.548
            BrowseDateTaxoFacets        4.97     (27.3%)        5.00     (27.3%)    0.8% ( -42% -   76%) 0.929
       BrowseDayOfYearTaxoFacets        5.06     (26.8%)        5.09     (26.9%)    0.8% ( -41% -   74%) 0.927
                    OrHighNotLow      244.26      (5.3%)      246.17      (5.3%)    0.8% (  -9% -   12%) 0.641
           BrowseMonthSSDVFacets        4.38      (3.9%)        4.41      (3.1%)    0.8% (  -5% -    8%) 0.479
                       OrHighLow      264.51      (5.4%)      266.91      (4.5%)    0.9% (  -8% -   11%) 0.568
               HighTermMonthSort     2111.04      (1.9%)     2131.88      (3.2%)    1.0% (  -4% -    6%) 0.236
                    OrNotHighLow      543.72      (2.9%)      549.12      (2.3%)    1.0% (  -4% -    6%) 0.229
                          Fuzzy2       51.05      (2.7%)       51.61      (2.6%)    1.1% (  -4% -    6%) 0.191
                          Fuzzy1       66.48      (3.0%)       67.31      (3.1%)    1.3% (  -4% -    7%) 0.199
                         MedTerm      508.60      (7.4%)      515.34      (7.2%)    1.3% ( -12% -   17%) 0.568
                        HighTerm      399.43      (7.2%)      405.35      (7.0%)    1.5% ( -11% -   16%) 0.509
          OrHighMedDayTaxoFacets        3.92      (4.2%)        3.98      (3.4%)    1.6% (  -5% -    9%) 0.194
                         LowTerm      350.89      (8.4%)      358.77      (9.4%)    2.2% ( -14% -   21%) 0.427

```

stefanvodita (approved):

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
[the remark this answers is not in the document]
stefanvodita: I missed this thread that was still pending. @msfroh - what do you think, can we simplify this method?

lucene/facet/src/test/org/apache/lucene/facet/taxonomy/TestTaxonomyCombined.java
```diff
@@ -669,20 +668,26 @@ public void testChildrenArraysInvariants() throws Exception {
...
       }
-      assertEquals(j, olderSiblingArray[i]);
+      assertEquals(j, olderSiblingArray.get(i));
     }
 
     tr.close();
     indexDir.close();
   }
 
+  private static void assertArrayEquals(int[] expected, ParallelTaxonomyArrays.IntArray actual) {
```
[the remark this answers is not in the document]
msfroh: This was the only place where we assert anything about the full contents of the `ChunkedArray`. I would lean toward not adding `equals` to `ChunkedArray`, since it's not something we would normally want to call, IMO. (In general, it's going to be pretty expensive.)

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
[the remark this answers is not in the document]
msfroh: Got it! 

I handled the multiple of `CHUNK_SIZE` case without allocating an empty array. I also added a unit test that specifically flexes ChunkedIntArray.

(And I cleaned up a few more cases where I was still dividing/moduloing instead of bit-shifting and masking.)

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
```diff
@@ -68,25 +94,66 @@ public TaxonomyIndexArrays(IndexReader reader, TaxonomyIndexArrays copyFrom) thr
...
   }
 
+  private static int[][] allocateChunkedArray(int size, int startFrom) {
+    if (size == 0) {
+      return new int[0][];
+    }
+    int chunkCount = size >> CHUNK_SIZE_BITS;
+    int fullChunkCount;
+    int lastChunkSize = size & CHUNK_MASK;
+    if (lastChunkSize == 0) {
```
stefanvodita, line 115: Thank you for persisting while we're iterating over this method.

Since `fullChunkCount` is assigned `chunkCount` on both branches, why not do this:
```java
int fullChunkCount = chunkCount;
if (lastChunkSize != 0) {
  chunkCount++;
}
```

On a higher level, I think I still wasn't specific enough in my previous comment. I didn't mind that we would sometimes have an empty array at the end if `size` was a multiple of `CHUNK_SIZE`, but we had if-statements that didn't seem to me like they were adding something. In this case, I prefer spending those extra bytes if we can make the code simpler. If you think the implementation we already have is better, we can keep it, but here is my preferred solution written out if you want to consider it:

```java
private static int[][] allocateChunkedArray(int size, int startFrom) {
    int chunkCount = (size >> CHUNK_SIZE_BITS) + 1;
    int[][] array = new int[chunkCount][];
    for (int i = startFrom; i < chunkCount - 1; i++) {
        array[i] = new int[CHUNK_SIZE];
    }
    array[chunkCount - 1] = new int[size & CHUNK_MASK];
    return array;
}
```

stefanvodita (commented):
@msfroh - I left just one more nitpicky comment. Can you also add a CHANGES entry? I would be happy to merge the PR after.

lucene/facet/src/java/org/apache/lucene/facet/taxonomy/directory/TaxonomyIndexArrays.java
msfroh: Oh, that's much cleaner, thank you!

msfroh:
> @msfroh - I left just one more nitpicky comment. Can you also add a CHANGES entry? I would be happy to merge the PR after.

Done! 

I also tweaked the bounds for `testThousandsOfCategories`, since it would take ~3.5 seconds on my (relatively powerful) work computer with a chain length of 3-6. Dropping to a length of 2-4 cut the time to 1.25 seconds.

stefanvodita:
Thank you! I think the PR is in good shape. I'll leave it up for a couple days and then merge if no one else comments.

PR #12996, opened 2024-01-06
Reduce ArrayUtil#grow in decompress

easyice:


lucene/core/src/java/org/apache/lucene/codecs/lucene90/LZ4WithPresetDictCompressionMode.java
```diff
@@ -128,10 +128,12 @@ public void decompress(DataInput in, int originalLength, int offset, int length,
       }
 
       // Read blocks that intersect with the interval we need
+      if (offsetInBlock < offset + length) {
```
dungba88, line 131: Wondering if growExact would be better here. I think grow will try to oversize it?

easyice: You are right, thank you!

jpountz: I'd rather keep using `grow` over `growExact`. This helps make sure we don't keep allocating a new array in cases lengths grow in small increments.

easyice: Okay, it's fixed :)

vigyasharma (approved):
This is a neat and simple optimization. We grow the `bytes` array to hold all the decompressed data in one shot, instead of doing it in multiples of `blockLength`.

Thanks for this improvement @easyice !

vigyasharma:
@easyice Let's add a changes.txt entry for this?

easyice:
Thank you for reviewing! @vigyasharma   

I have added the CHANGES entry under Lucene 9.11.0

jpountz:
@vigyasharma I wonder if you missed to backport this change to `branch_9x`?

vigyasharma:
@jpountz my bad, I think I missed the back-port. Merged it now.

jpountz:
Thank you!

PR #12813, opened 2023-11-15
Introduce workflow for stale PRs

stefanvodita:
PRs get stale and we miss out on good contributions. This workflow will mark PRs that are becoming stale.

Addresses #12796

PR #12998, opened 2024-01-08
Increase stale PRs action budget and mark not debug-only

stefanvodita:
With this change, we will start marking stale PRs.

Follow-up to #12813.


mikemccand (approved):
Thanks @stefanvodita -- I'll merge once the checks finish!

PR #13001, opened 2024-01-09
Forbidden Thread.sleep API

### LINKED ISSUE #12946: Can we ban `Thread.sleep`?
### Description

Spinoff from #12912.

`Thread.sleep` should ideally never appear in our main and test sources.  Let's add it to forbidden APIs?

shubhamvishu:
### Description

This API mark `Thread.sleep` API as forbidden for futures uses in the codebase and suppresses the existing usages.

Closes #12946


gradle/validation/forbidden-apis/defaults.all.txt
```diff
@@ -74,3 +74,8 @@ javax.sql.rowset.spi.SyncFactory
 @defaultMessage Math.fma is insanely slow (2500x) in many environments (e.g. VMs). Use multiply/add and suffer the extra rounding
 java.lang.Math#fma(float,float,float)
 java.lang.Math#fma(double,double,double)
+
+@defaultMessage We should not use Thread.sleep in lucene code
+java.lang.Thread#sleep(long)
```
uschindler, line 79: I'd use `java.lang.Thread#sleep(**)`.

This prevents any sleep variant coming soon. E.g., the Duration one not yet available in JDK 17.

uschindler (changes requested):

gradle/validation/forbidden-apis/defaults.all.txt
uschindler: you should also move the message into the signature. I'd also prefer a more meaningful message.

shubhamvishu: Makes sense. I updated these in the new revision

uschindler (commented):
In general this looks fine, although I think we should also work on remove some of those sleeps. For all others there should be an explanation, why the test needs sleeping.

The description in the forbiddenmessage is now a bit too verbose, but much better!

uschindler:
P.S.: You don't need to force-push, it is better to keep track what was changed in a PR. The PR will get squashed anyways.

mikemccand:
+1 to merge this, first, so we stop the bleeding (no more `Thread.sleep` added to the code base), and in follow-on PR we can start whittling down the existing grandfather'd `Thread.sleep` calls.

uschindler (approved):

shubhamvishu:
@uschindler @mikemccand I think we could merge this now unless I'm missing something?

mikemccand (approved):
+1, let's merge this!  Progress not perfection!  It has already picked up one conflict because we were too slow merging ...

uschindler (changes requested):
You added merge artifacts. Please remove them.

shubhamvishu:
@uschindler I have removed the added file. Thanks!

uschindler (approved):

uschindler:
I moved the changes entry to Lucene 10.0, as to me it makes no sense to apply this to Lucene 9.x which is not used for active development. New code enters main first and I had trouble in merging this (and 9.x has many more sleeps).

PR #12829, opened 2023-11-21
Add support for index sorting with document blocks

s1monw:
Today index sorting with likely break document blocks added with `IndexWriter#addDocuments(...)` and friends since the index sorter has no indication of what documents are part of a block. This change proposes a marker field as a requirement for parent documents if the block API is used in conjunction with index sorting. 
At this point this change requires a `NumericDV` field only present on the last document of a block or on every document indexed as an individual document iff a parent field is configured on the Sort. This can potentially be extended to a `Term` which is not as straight forward since validations might be more difficult and today postings are not available when we sort the flushed segment. 

Relates to #12711

PR #13013, opened 2024-01-15
Backport #12829 to 9.x

s1monw:
This is a backport PR for #12829. It add the optional ability to index parent fields in 9.x. While 9.x can still use index sorting with doc blocks without using a parent field this allows users to migrate to this mandatory feature earlier than 10.x. Only indices with min segment version 9.9 and older will be able to be configured with a parent field. These versions record if the index already has blocks which makes it safe to add the parent field after the fact.


s1monw:
this doesn't work as I expected. I will close this one for now

PR #12784, opened 2023-11-08
Cache buckets to speed up BytesRefHash#sort

gf2121:
Following https://github.com/apache/lucene/pull/12775, this PR tries another approach to speed up `BytesRefHash#sort`:
The idea is that since we have extra ints in this map, we can cache the bucket when building the histograms, and reuse them when `reorder`.  I checked this approach on intel chip, showing ~30% speed up. I'll check M2 chip and wikimedium data tomorrow.

```
BASELINE:  sort 5169965 terms, build histogram took: 1968ms, reorder took: 2132ms, total took: 5470ms.
BASELINE:  sort 5169965 terms, build histogram took: 1975ms, reorder took: 2133ms, total took: 5526ms.
BASELINE:  sort 5169965 terms, build histogram took: 1999ms, reorder took: 2157ms, total took: 5573ms.
BASELINE:  sort 5169965 terms, build histogram took: 1955ms, reorder took: 2138ms, total took: 5446ms.
BASELINE:  sort 5169965 terms, build histogram took: 1990ms, reorder took: 2161ms, total took: 5528ms.
BASELINE:  sort 5169965 terms, build histogram took: 1997ms, reorder took: 2175ms, total took: 5571ms.
BASELINE:  sort 5169965 terms, build histogram took: 2004ms, reorder took: 2119ms, total took: 5477ms.
BASELINE:  sort 5169965 terms, build histogram took: 1978ms, reorder took: 2155ms, total took: 5501ms.
BASELINE:  sort 5169965 terms, build histogram took: 2015ms, reorder took: 2169ms, total took: 5572ms.
BASELINE:  sort 5169965 terms, build histogram took: 1941ms, reorder took: 2138ms, total took: 5400ms.
BASELINE:  sort 5169965 terms, build histogram took: 2000ms, reorder took: 2155ms, total took: 5558ms.

CANDIDATE:  sort 5169965 terms, build histogram took: 1996ms, reorder took: 133ms, total took: 3734ms.
CANDIDATE:  sort 5169965 terms, build histogram took: 1989ms, reorder took: 142ms, total took: 3655ms.
CANDIDATE:  sort 5169965 terms, build histogram took: 2031ms, reorder took: 155ms, total took: 3762ms.
CANDIDATE:  sort 5169965 terms, build histogram took: 2016ms, reorder took: 145ms, total took: 3739ms.
CANDIDATE:  sort 5169965 terms, build histogram took: 1994ms, reorder took: 142ms, total took: 3667ms.
[...]

PR #13014, opened 2024-01-16
Rollback the tmp storage of BytesRefHash to -1 after sort

gf2121:
In #12784 we cache buckets into extra slots in `BytesRefHash` to speed up `BytesRefHash#sort`. This causes an [AssertError](https://lists.apache.org/thread/lxk2lm88twfzdl549974vov1984kj995) for [`TermsQuery`](https://github.com/apache/lucene/blob/c746bea233fd14fe81d9805633fcce0f7b9681b6/lucene/join/src/java/org/apache/lucene/search/join/TermsQuery.java#L68) because it could call `sort` more than once on a `BytesRefHash` instance. Not sure if any other users were relying on this but it would be great to keep consistent with before.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/BytesRefHash.java
+++ b/lucene/core/src/java/org/apache/lucene/util/BytesRefHash.java
@@ -214,6 +214,7 @@ public final class BytesRefHash implements Accountable {
         pool.fillBytesRef(result, bytesStart[compact[i]]);
       }
     }.sort(0, count);
+    Arrays.fill(compact, tmpOffset, compact.length, -1);
     return compact;
   }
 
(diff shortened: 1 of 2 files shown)


ChrisHegarty (approved):
LGTM.

ChrisHegarty:
@gf2121  I've added the 9.9.2 milestone to this PR.  Do you agree? If so, is it possible to merge and backport to the `branch_9x` and `branch_9_9` branches.

jpountz (approved):
Thanks @gf2121 !

mikemccand:
Thanks for finding/fixing this @gf2121!

> because it could call `sort` more than once on a `BytesRefHash` instance

Hmm -- that sounds buggy (on `TermsQuery` part)?  Shouldn't callers be prevented from calling `.sort()` more than once (I think it is a destructive operation?).  Or at least let's fix `TermsQuery` to not do this?

cpoerschke:
> ... Or at least let's fix `TermsQuery` to not do this?

Thanks for the input! Wondering if perhaps I was too hasty closing out the naive https://github.com/apache/lucene/pull/13019 change ... WDYT?

mikemccand:
> Wondering if perhaps I was too hasty closing out the naive #13019 change ... WDYT?

+1 -- let's reopen?  I would rather move towards making this API prevent such abuse than being lenient about it, if we can.  Thanks @cpoerschke.

PR #13016, opened 2024-01-16
Prevent parent field from added to existing index

s1monw:
This change prevents users from adding a parent field to an existing index. Parent field must be added before any documents are added to the index to prevent documents without the parent field from being indexed and later to be treated as child documents upon merge.

Relates to #12829


jpountz (approved):
LGTM

Issue #13015, opened 2024-01-16
DV update files referenced by merge will be deleted by concurrent flush


guojialiang92:
### Description

DV update files referenced by merge will be deleted by concurrent flush.
For example, according to the following execution sequence, the DV update file referenced by merge will be deleted.

1. `IndexWriter#addDocument` doc1 = {"id": "1", "version": "1"}
2. `IndexWriter#flush`, which will produce segment0 = {_0.cfe, _0.cfs, _0.si}
3. `IndexWriter#addDocument` = {"id": "2", "version": "1"}
4. `IndexWriter#softUpdateDocument` updates doc2 to {"id": "2", "version": "2"} 
5. `IndexWriter#flush`, which will produce segment1 = {_1.cfe, _1.cfs, _1.si, _1_1.fnm, _1_1_Lucene80_0.dvd, _1_1_Lucene80_0.dvm}
6. `IndexWriter#forceMerge`, merge segment0 and segment1 into segment2. The merge process will reference segment0 = {_0.cfe, _0.cfs, _0.si} and segment1 = {_1.cfe, _1.cfs, _1.si, _1_1.fnm, _1_1_Lucene80_0.dvd, _1_1_Lucene80_0.dvm}
7. During merge, `IndexWriter#softUpdateDocument` updates doc2 to {"id": "2", "version": "3"}
8. `IndexWriter#flush`，which will produce segment3 = {_3.cfe, _3.cfs, _3.si}，and update segment1 to {_1.cfe, _1.cfs, _1.si, _1_2.fnm, _1_2_Lucene80_0.dvd, _1_2_Lucene80_0.dvm}
9. The files = {_1_1.fnm, _1_1_Lucene80_0.dvd, _1_1_Lucene80_0.dvm} referenced by merge is deleted

In the Linux environment, an open file is still readable after being deleted, so it does not cause file reading exception.
However, in other environment that do not have the above guarantee, an exception will be caused because the file does not exist.

Later I will file a PR for this. Add related tests and fix the problem. 

### Version and environment details

I found the problem in releases/lucene-solr/8.7.0, and confirmed that the problem still exists in main branch.

guojialiang92:
I saw a similar issue [11751](https://github.com/apache/lucene/issues/11751) a year ago.

guojialiang92:
I saw a similar issue [11751](https://github.com/apache/lucene/issues/11751).

mikemccand:
[...]

PR #13017, opened 2024-01-16
Fix DV update files referenced by merge will be deleted by concurrent flush

guojialiang92:
### Description

This PR aims to address issue #13015.
A more detailed explanation of the issue and the reasoning behind the fix can be found in the report link above.

### Solution

- Before the file merge starts, execute `IndexFileDeleter#incRef` on the files referenced by merge to ensure that the file will not be deleted.
- After the file merge is completed/abort, execute `IndexFileDeleter#decRef`.
- In `ReadersAndUpdates#writeFieldUpdates`, `ReadersAndUpdates#swapNewReaderWithLatestLiveDocs` should be executed after `SegmentCommitInfo#setDocValuesUpdatesFiles`, ensuring that `SegmentReader` contains a deep clone of the `SegmentCommitInfo#dvUpdatesFiles`.

### Tests

- I introduced `MockAssertFileExistIndexInput` which validate file exists before reading it.
- I introduced `testForceMergeDVUpdateFileWithConcurrentFlush` which fails on `main` branch since the DV update file not exist, but passes on this branch.
- I added corresponding `testMergeDVUpdateFileOnGetReaderWithConcurrentFlush` which passes on both `main` and this branch. Because before merge, `IndexFileDeleter#incRef` is executed in `StandardDirectoryReader#open`.
- I added corresponding `testMergeDVUpdateFileOnCommitWithConcurrentFlush` which passes on both `main` and this branch. Because before merge, `IndexFileDeleter#incRef` is executed in `IndexWriter#prepareCommitInternal`.

### Checklist

- [x]  I have reviewed the guidelines for [How to Contribute](https://github.com/apache/lucene/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.
- [x]  I have given Lucene maintainers [access](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [x]  I have developed this patch against the `main` branch.
- [x]  I have run `./gradlew check`.
- [x]  I have added tests for my changes.


lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -5136,7 +5141,13 @@ private int mergeMiddle(MergePolicy.OneMerge merge, MergePolicy mergePolicy) thr
           sci -> {
             final ReadersAndUpdates rld = getPooledInstance(sci, true);
             rld.setIsMerging();
-            return rld.getReaderForMerge(context);
+            synchronized (this) {
+              return rld.getReaderForMerge(
+                  context,
+                  mr -> {
+                    deleter.incRef(mr.reader.getSegmentInfo().files());
```
soren-xzk, line 5148: There is an assert judgment in incRef/decRef, which must have the IndexWriter's lock. There is no lock in this place, may there be a problem?

guojialiang92: Hello, This PR has noticed that `IndexFileDeleter#incRef/IndexFileDeleter#decRef` needs to have the `IndexWriter` lock. I will explain in detail:

1. `IndexFileDeleter#incRef` is called in `IndexWriter#mergeMiddle`. it has been locked by `synchronized (this)` before the `IndexFileDeleter#incRef` is called.
2. Another call of `IndexFileDeleter#incRef` in `IndexWriter#preparePointInTimeMerge`. `IndexWriter#preparePointInTimeMerge` is called by `IndexWriter#prepareCommitInternal` or `IndexWriter#getReader`. `synchronized (this)` is performed in both `IndexWriter#prepareCommitInternal` and `IndexWriter#getReader`, so the `IndexFileDeleter#incRef` in  `IndexWriter#preparePointInTimeMerge` does not need to be locked.
3. `IndexFileDeleter#decRef` is called in `IndexWriter#closeMergeReaders`. `IndexWriter#closeMergeReaders` is a `synchronized` function.

stefanvodita (changes requested):
Thank you for identifying and addressing this issue and sorry for the long wait in review.
I appreciate the thorough tests, I ran them too with and without the change and saw the same success/failure pattern you described.

To commit this, we need to add an entry to `CHANGES.txt`. The release process for 9.10 is starting on Monday. I don’t know if we would delay for this bug fix, so maybe it’s safer to aim for 9.11 - I think you would have to add a new section in `CHANGES.txt`.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMergePolicy.java
```diff
@@ -950,4 +957,354 @@ public void setMergeInfo(SegmentCommitInfo info) {
...
+
+    public MockAssertFileExistIndexInput(String name, IndexInput delegate, Path filePath) {
+      super("MockAssertFileExistIndexInput(name=" + name + " delegate=" + delegate + ")");
+      this.name = name;
+      this.delegate = delegate;
+      this.filePath = filePath;
+    }
+
+    private void checkFileExist() throws IOException {
+      if (!Files.exists(filePath)) {
```
stefanvodita, line 974: Nit: We try not to use `!` in the Lucene codebase because it's easy to miss. We can say `Files.exists(filePath) == false` instead.

stefanvodita:
Thanks! Could you also look at the CHANGES entry I had mentioned? We want to record the fix and credit you for the work.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3935,7 +3935,11 @@ public void setMergeInfo(SegmentCommitInfo info) {
                 // updates
                 // in a separate map in order to be applied to the merged segment after it's done
                 rld.setIsMerging();
-                return rld.getReaderForMerge(context);
+                return rld.getReaderForMerge(
+                    context,
+                    mr -> {
```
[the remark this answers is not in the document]
guojialiang92: Nice catch! I've adapted the code as suggested.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMergePolicy.java
guojialiang92: Sounds good! I have fix it.

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
```diff
@@ -782,13 +784,17 @@ synchronized MergePolicy.MergeReader getReaderForMerge(IOContext context) throws
     }
 
     SegmentReader reader = getReader(context);
-    if (pendingDeletes.needsRefresh(reader)) {
+    if (pendingDeletes.needsRefresh(reader)
+        || reader.getSegmentInfo().getDelGen() != pendingDeletes.info.getDelGen()) {
```
guojialiang92, line 788: I found that when both `IndexWriter#updateDocument` and `IndexWriter#softUpdateDocument` exist, the new refCount control logic may delete the `.liv` file that no longer exists.

I have fixed this problem and added test cases, please code review, thank you!
@stefanvodita 

### Test
In order to reproduce and solve the problem, I introduced `testForceMergeWithPendingHardAndSoftDeleteFile`.

1. `IndexWriter#addDocument` doc1 = {"id": "1", "version": "1"}
2. `IndexWriter#commit`, which will produce segment0 = {_0.cfe, _0.cfs, _0.si}
3. `IndexWriter#addDocument` doc2 = {"id": "2", "version": "1"}
4. `IndexWriter#addDocument` doc3 = {"id": "3", "version": "1"}
5. `IndexWriter#addDocument` doc4 = {"id": "4", "version": "1"}
6. `IndexWriter#addDocument` doc5 = {"id": "5", "version": "1"}
7. `IndexWriter#commit`, which will produce segment1 = {_1.cfe, _1.cfs, _1.si}
8. `IndexWriter#updateDocument` updates doc2 to {"id": "2", "version": "2"}
9. `IndexWriter#commit`, which will produce segment2 = {_2.cfe, _2.cfs, _2.si}, and update segment1 to {_1.cfe, _1.cfs, _1.si, _1_1.liv}
10. `IndexWriter#updateDocument` updates doc3 to {"id": "3", "version": "2"}
11. `IndexWriter#softUpdateDocument` updates doc4 to {"id": "4", "version": "2"}
12. `DirectoryReader reader = IndexWriter#getReader(true, false)`, which will produce segment3 = {_3.cfe, _3.cfs, _3.si}, and update segment1 to {_1.cfe, _1.cfs, _1.si, _1_1.liv, _1_1.fnm, _1_1_Lucene80_0.dvd, _1_1_Lucene80_0.dvm}
13. `reader.close()`
14. `IndexWriter#commit`, update segment1 to {_1.cfe, _1.cfs, _1.si, _1_2.liv, _1_1.fnm, _1_1_Lucene80_0.dvd, _1_1_Lucene80_0.dvm}
15. `IndexWriter#forceMerge(1)`

### Analyze
**In step 12**
Because parameter `writeAllDeletes` of method `IndexWriter#getReader` has a value of `false`, `IndexWriter#writeReaderPool`, `ReaderPool#writeAllDocValuesUpdates`, and `ReadersAndUpdates#writeFieldUpdates` will be executed in sequence.

In method `ReadersAndUpdates#writeFieldUpdates`, since `ReadersAndUpdates#pendingDVUpdates` is not empty, `ReadersAndUpdates#swapNewReaderWithLatestLiveDocs` and `ReadersAndUpdates#createNewReaderWithLatestLiveDocs` will be executed in sequence.

`ReadersAndUpdates#createNewReaderWithLatestLiveDocs` will generate a new `SegmentReader` based on `pendingDeletes`. The new `SegmentReader` will hold the latest `liveDocs` and `numDocs`, but the value of `SegmentReader.si.delGen `will still be 1.

**In step 14**
_1_2.liv will be created, _1_1.liv will be deleted, and `pendingDeletes.info.delGen` will be updated to 2. 

**In step 15**
`ReadersAndUpdates#getReaderForMerge` will determine whether `SegmentReader` needs to be refreshed based on `pendingDeletes`. If `delGen` is not compared, `SegmentReader` will not be refreshed, and the refCount of _1_1.liv will still be +/- during the merge. When the refCount reaches 0, _1_1.liv will be deleted and `NoSuchFileException` will be thrown.

### Solution
In order to solve this problem, `ReadersAndUpdates#getReaderForMerge` is necessary to add comparison logic to `delGen`.
`SegmentReader` needs to be refreshed when `pendingDeletes.needsRefresh(reader) || reader.getSegmentInfo().getDelGen() != pendingDeletes.info.getDelGen()` returns `true`.

guojialiang92:
> Thanks! Could you also look at the CHANGES entry I had mentioned? We want to record the fix and credit you for the work.

Thanks, I have add entry to CHANGES.txt.

stefanvodita (approved):
I think the change is good, but there's people who understand this code much better than I. I'll approve tentatively, but leave the PR up over the weekend to see if anyone else wants to have a look. Amazing work @guojialiang92!

PR #13026, opened 2024-01-22
Propagate minimum competitive score in ReqOptSumScorer.

jpountz:
If the required clause doesn't contribute scores, which typically happens if the required clause is a `FILTER` clause, then the minimum competitive score can be propagated directly to the optional clause.


jimczi (approved):
lgtm

PR #13027, opened 2024-01-22
Fix NPE when sampling for quantization in Lucene99HnswScalarQuantizedVectorsFormat

benwtrent:
When merging `Lucene99HnswScalarQuantizedVectorsFormat` a NPE is possible when deleted documents are present.

`ScalarQuantizer#fromVectors` doesn't take deleted documents into account. This means using `FloatVectorValues#size` may actually be larger than the actual size of live documents. Consequently, when iterating for sampling iteration too far is possible and an NPE will be thrown.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -243,6 +243,13 @@ Other
 
 * GITHUB#12934: Cleaning up old references to Lucene/Solr. (Jakub Slowinski)
 
+======================== Lucene 9.9.2 =======================
+
+Bug Fixes
+---------------------
+
+* GITHUB#13027: Fix NPE when sampling for quantization in Lucene99HnswScalarQuantizedVectorsFormat (Ben Trent)
+
 ======================== Lucene 9.9.1 =======================
 
 Bug Fixes
--- a/lucene/core/src/java/org/apache/lucene/util/ScalarQuantizer.java
+++ b/lucene/core/src/java/org/apache/lucene/util/ScalarQuantizer.java
@@ -192,6 +192,38 @@ public class ScalarQuantizer {
 
   private static final Random random = new Random(42);
 
+  static int[] reservoirSampleIndices(int numFloatVecs, int sampleSize) {
+    int[] vectorsToTake = IntStream.range(0, sampleSize).toArray();
+    for (int i = sampleSize; i < numFloatVecs; i++) {
+      int j = random.nextInt(i + 1);
+      if (j < sampleSize) {
+        vectorsToTake[j] = i;
+      }
+    }
+    Arrays.sort(vectorsToTake);
+    return vectorsToTake;
+  }
+
+  static float[] sampleVectors(FloatVectorValues floatVectorValues, int[] vectorsToTake)
+      throws IOException {
+    int dim = floatVectorValues.dimension();
+    float[] values = new float[vectorsToTake.length * dim];
+    int copyOffset = 0;
+    int index = 0;
+    for (int i : vectorsToTake) {
+      while (index <= i) {
+        // We cannot use `advance(docId)` as MergedVectorValues does not support it
+        floatVectorValues.nextDoc();
+        index++;
+      }
+      assert floatVectorValues.docID() != NO_MORE_DOCS;
+      float[] floatVector = floatVectorValues.vectorValue();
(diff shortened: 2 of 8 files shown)


jpountz:
Good catch. I wonder what is the best place to compute size() correctly. I see you fixed the merge instances, but this is not how it's done elsewhere, see e.g. `FieldsConsumer#write`, which clearly states that index statistics must not be pulled from the merged `Terms` instance but recomputed. Should we follow a similar approach and re-compute the size in `ScalarQuantizer#fromVectors`? I see that it needs to linearly scan all vectors anyway, so this shouldn't come at a performance penalty?

benwtrent:
@jpountz 

> I see that it needs to linearly scan all vectors anyway, so this shouldn't come at a performance penalty?

We would then need to linearly scan all the vectors twice. Because we do not have random access, we can only sample while iterating directly and we don't want to read&copy in vectors that ultimately don't get sample anyways (this would happen if we randomly sampled while iterating).

So, we would need to iterate them twice:

 - Once to get the size
 - Then again to sample the vectors

Instead of passing in two iterators, passing in a live vector count seemed more sensible as this is already provided.



Adding an `@Override` to `MergedVectorValues` that asserts `size()` is never called is a good protection idea.

jpountz:
OK, taking an int in that method sounds fine to me. Can we still preserve the signature of `mergeOneField` to return `void` and re-compute the doc count before passing the doc count to `ScalarQuantizer#fromVectors`? Returning the doc count feels a bit arbitrary.

benwtrent:
@jpountz, I gotcha, pushed a change

jpountz (approved):

ChrisHegarty (approved):
The code changes LGTM

lucene/core/src/java/org/apache/lucene/util/ScalarQuantizer.java
```diff
@@ -201,13 +233,26 @@ public String toString() {
...
    * @param floatVectorValues the float vector values from which to calculate the quantiles
    * @param confidenceInterval the confidence interval used to calculate the quantiles
+   * @param totalVectorCount the total number of live float vectors in the index. This is vital for
+   *     accounting for deleted documents when calculating the quantiles.
    * @return A new {@link ScalarQuantizer} instance
    * @throws IOException if there is an error reading the float vector values
    */
   public static ScalarQuantizer fromVectors(
-      FloatVectorValues floatVectorValues, float confidenceInterval) throws IOException {
+      FloatVectorValues floatVectorValues, float confidenceInterval, int totalVectorCount)
```
ChrisHegarty, line 242: Alternatively, and if we're concerned about binary compatibility with 9.9.0, then this could be structured as a new additional override accepting the `totalVectorCount`, while retaining the the original signature that passes the `floatVectorValues.size()` and asserting that there are no deleted docs or something. This could be a bit "trappy", but no unreasonable if we think that we need to maintain binary compatibility.

lucene/CHANGES.txt
```diff
@@ -243,6 +243,13 @@ Other
 
 * GITHUB#12934: Cleaning up old references to Lucene/Solr. (Jakub Slowinski)
 
+======================== Lucene 9.9.2 =======================
+
+Bug Fixes
+---------------------
+
+* GITHUB#13027: Fix NPE when sampling for quantization in Lucene99HnswScalarQuantizedVectorsFormat (Ben Trent)
```
ChrisHegarty, line 251: This is a significant issue and needs addressing with some urgency. However, we need to decide exactly where this should land.

lucene/core/src/java/org/apache/lucene/util/ScalarQuantizer.java
benwtrent: Yeah, changing a public definition in a bug fix stinks. I can overload it for the bug fix and remove the overload in a minor release. These are experimental, so we should be able to change them in minors.

lucene/CHANGES.txt
ChrisHegarty: After given it some thought, I agree with targeting this for 9.9.2. This kinda bug is exactly why we have patch releases.

ChrisHegarty (approved):

PR #13031, opened 2024-01-24
Propagate topLevelScoringClause from QueryProfiler

mrkm4ntr:
### Description
The topLevelScoringClause is not propagated from QueryProfiler. This causes some optimizations are skipped during profiling.


jpountz (approved):
Good catch!

jpountz:
@mrkm4ntr Can you add a CHANGES entry for your change?

mrkm4ntr:
@jpountz Thank you for your review. Where should I add an entry? 9.9.2 or 9.10.0?

jpountz:
@mrkm4ntr 9.10 please

mrkm4ntr:
@jpountz Sure. Added.

PR #13033, opened 2024-01-25
Early exit range queries on non-matching segments.

jpountz:
This change makes `PointRangeQuery` exit early when it knows that it doesn't match a segment. In the case when this query is part of a conjunction, this helps make sure `ScorerSupplier#get` doesn't get called on other required clauses, which is sometimes an expensive operation (e.g. multi-term queries).

This is especially relevant for time-based data combined with `LogByteSizePolicy`, which gives non-adjacent ranges of timestamps to each segment, which in-turn increases the likelihood that some segments won't match a range query at all.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
@@ -290,6 +290,24 @@ public abstract class PointRangeQuery extends Query {
           return null;
         }
 
+        if (values.getDocCount() == 0) {
+          return null;
+        } else {
+          final byte[] fieldPackedLower = values.getMinPackedValue();
+          final byte[] fieldPackedUpper = values.getMaxPackedValue();
+          for (int i = 0; i < numDims; ++i) {
+            int offset = i * bytesPerDim;
+            if (comparator.compare(lowerPoint, offset, fieldPackedUpper, offset) > 0
+                || comparator.compare(upperPoint, offset, fieldPackedLower, offset) < 0) {
+              // If this query is a required clause of a boolean query, then returning null here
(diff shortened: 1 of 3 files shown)


iverase (approved):
LGTM

lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
```diff
@@ -290,6 +290,24 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
           return null;
         }
 
+        if (values.getDocCount() == 0) {
+          return null;
+        } else {
```
benwtrent, line 295: From what I can tell, `ScorerSupplier` is an optional interface, which is probably just fine for most things (I am unsure which queries satisfy it or if we care).

I am not 100% sure the expensive part we are trying to bypass. But in a `BooleanQuery`, if we hit an "expensiveScorerSupplier" first, we will call `scorerSupplier.cost()` on that "expensiveScorerSupplier". If this is fine & still allows for the biggest optimizations (bypassing `scorerSupplier.get(leadCost)`), then this solution is rather nice!

jpountz:
> I am not 100% sure the expensive part we are trying to bypass

Typically it's browsing terms dictionaries and/or building bitsets across the entire term/doc ID space. So multi-term queries, term-in-set queries and a few more.

> we will call scorerSupplier.cost() on that "expensiveScorerSupplier". If this is fine

This is fine, `ScorerSupplier` was introduced in the first place to be able to do a quick estimation of the cost without building a scorer. This is then used in `IndexOrDocValuesQuery` to know whether the query will be used to drive iteration (in which case the index query is used), or just to validate matches (in which case the doc-values query is used).

benwtrent (approved):
:shipit:

PR #13037, opened 2024-01-27
Removing thread sleep calls from TestIndexWriter.testThreadInterruptDeadlock and TestDirectoryReader.testStressTryIncRef

iamsanjay:
### Description

As suggested in  #12946, I tried to tackle two test cases and replacing `Thread.sleep` calls appearing in `TestIndexWriter.testThreadInterruptDeadlock` and  `TestDirectoryReader.testStressTryIncRef`

And of course, #13001 is going to precedes this one. For now seeking suggestions from community regarding new code that would replace the `Thread.sleep` calls.


lucene/core/src/test/org/apache/lucene/index/TestDirectoryReader.java
```diff
@@ -1005,55 +1010,59 @@ public void testTryIncRef() throws IOException {
...
-
-  static class IncThread extends Thread {
-    final IndexReader toInc;
-    final Random random;
-    Throwable failed;
-
-    IncThread(IndexReader toInc, Random random) {
-      this.toInc = toInc;
-      this.random = random;
-    }
```
iamsanjay, line 1044: I refactored `testStressTryIncRef` test bit further, removed the IncThread class and convert it into a runnable. And with that use the call to `random()` which we do not use anymore. 
Also introduce the try-with-resources statement. I would love to get some feedback on it!

I observed that usually we are using `Thread.sleep`, in many cases, where we want to delay the execution of some lines of code in the main thread. For instance, closing IndexReader in above case before stress test the `tryIncRef()` and `decRef()` via running two threads simultaneously. To replace `Thread.sleep`, I inspect the code for which we want the delayed execution and then schedule that block via `ScheduledExecutorService`. Having said that, I have my suspicion that whether we want to pivot in that direction.

PR #13038, opened 2024-01-28
Remove concatenation in String.format() calls

sabi0:
Using concatenation in the format string like this:
```
              String.format(
                  Locale.ROOT,
                  "top: warning: still warming merge "
                      + info
                      + " to "
                      + preCopy.connections.size()
                      + " replicas for %.1f sec...",
```

- is inconsistent
- is error prone (the `info.toString()` result above might happen to contain some `100%s`)
- results in longer code
- that is more difficult to modify

Replacing the concatenation with the format placeholders results in a cleaner more compact code:
```
              String.format(
                  Locale.ROOT,
                  "top: warning: still warming merge %s to %d replicas for %.1f sec...",
                  info,
                  preCopy.connections.size(),
```


uschindler:
Thanks for this. The code also has safety problems. If the concatted text para contain percent signs it would break. So format strings should never ever be constructed with variables containing inches ked strings.

uschindler (approved):

PR #12295, opened 2023-05-15
Use `instanceof` pattern-matching where possible

JarvisCraft:
### Description

This PR enables the usage of `instanceof` pattern matching wherever possible (without changing semantics) reducing error-proneness and potentially enhancing readability.

### Important notes

1. It is also woth mentioning that whenever used with inverted conditions (such as `!(foo instanceof Bar bar)` this **must** use prefix invertion notation (`!expr`) instead of comparisoj to `false` (`expr == false`).

2. There are multiple occurrences of the following pattern:

  ```java
  assert foo instanceof Bar;
  Bar bar = (Bar) foo;
  ```

  Formally, this **cannot** be rewritten as

  ```java
  assert foo instanceof Bar;
  Bar bar = (Bar) foo;
  ```

  because the assertion may be skipped if not enabled.

  I have kept such cases untouched although I suggest
  to update them to something like

  ```java
  if (!foo instanceof Bar bar) {
      throw new AssertionError();
  }
  ```

  which seems to have no performance penalty considering that typecheck is done anyway
  when trying to cast `foo` to `Bar` so toggleable assertion seems to not be helping with anything

  (please write in the comments, if this should be done).

PR #13039, opened 2024-01-28
Align instanceof check with type cast

sabi0:


lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/standard/processors/AnalyzerQueryNodeProcessor.java
```diff
@@ -107,10 +105,9 @@ public QueryNode process(QueryNode queryTree) throws QueryNodeException {
   @Override
   protected QueryNode postProcessNode(QueryNode node) throws QueryNodeException {
 
-    if (node instanceof TextableQueryNode
+    if (node instanceof FieldQueryNode
```
sabi0, line 108: 'node' is cast below to `FieldQueryNode`. But the condition here was checking for a base type `TextableQueryNode`, excluding some sub-types. If a new `TextableQueryNode` sub-type was added the code below could run into a CCE.

Having the `instanceof FieldQueryNode` condition here makes the type cast below safe. And also eliminates the need to exclude the `RegexpQueryNode` as it does not subclass `FieldQueryNode`.

lucene/core/src/java/org/apache/lucene/analysis/tokenattributes/PayloadAttributeImpl.java
```diff
@@ -62,8 +62,7 @@ public boolean equals(Object other) {
       return true;
     }
 
-    if (other instanceof PayloadAttribute) {
-      PayloadAttributeImpl o = (PayloadAttributeImpl) other;
+    if (other instanceof PayloadAttributeImpl o) {
```
uschindler, line 65: This looks like an unrelated change. There is a PR about this type of code change already. It can only be applied to main branch as it requires Java 17.

uschindler: See this PR: #12295

sabi0: Thank you for linking to the pattern-matching PR.
I've removed the pattern variable part of this change.

uschindler:
I added the changes entry to the existing one for you.

sabi0:
Thank you. I will make sure the new PRs have them out of the box.

PR #13041, opened 2024-01-28
Cleanup TokenizedPhraseQueryNode code

sabi0:
- harmonize implementations of the methods working with children
- save `getChildren()` in a variable and re-use it
- use more efficient `isEmpty()` instead of `size() == 0`
- fix a typo in the tag name: `tokenized_t_phrase` => `tokenizedphrase`
- remove unnecessary `toString()` call
- remove trailing comment referring to another class (copy paste? renaming that did not update the comment?)


lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/core/nodes/TokenizedPhraseQueryNode.java
```diff
@@ -70,10 +70,8 @@ public QueryNode cloneTree() throws CloneNotSupportedException {
   @Override
   public CharSequence getField() {
     List<QueryNode> children = getChildren();
-
-    if (children == null || children.size() == 0) {
+    if (children == null || children.isEmpty()) {
       return null;
-
     } else {
       return ((FieldableNode) children.get(0)).getField();
```
sabi0, line 76: According to the `setField(CharSequence)` method below children are not necessarily `FieldableNode`.
I.e. this line might throw `ClassCastException`?

stefanvodita: I see your point. Maybe it's worth adding an assertion?

stefanvodita (approved):
The clean-up looks good. Sorry it took so long to get a review.

lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/core/nodes/TokenizedPhraseQueryNode.java
sabi0: I think the two methods should be aligned.
If one has an assert (assumes all the children are `FieldableNode`) then the other should do so too.
Of they both should silently ignore children of other types.

I would probably go with "silently ignore children of other types" as the least surprising option.

stefanvodita: I think having the two methods aligned, the way they are in your latest commit, makes sense.

lucene/CHANGES.txt
```diff
@@ -258,6 +258,8 @@ Improvements
   Tests are running with random byte order to ensure that the order does not affect correctness
   of code. Native order was enabled for LZ4 compression.  (Uwe Schindler)
 
+* GITHUB#13041: TokenizedPhraseQueryNode code cleanup (Dmitry Cherniachenko)
```
stefanvodita, line 261: This would have to move up to version 9.11 because 9.10 is already in the release process.

PR #13042, opened 2024-01-28
Use short-circuit boolean expressions

sabi0:


jpountz:
Thanks, eagerly evaluating all conditions looks unintended indeed.

PR #13043, opened 2024-01-29
Support getMaxScore of ConjunctionScorer for non top level scoring clause

mrkm4ntr:
### Description
After introducing topLevelScoringClause, ConjunctionScorer with multiple scorers can be used for non top level scoring clause conjunctions instead of BlockMaxConjunctionScorer even requiredScorers is empty. In such case, ConjunctionScorer returns Infinity as maxScore and it ruins some optimizations like parent WANDScorer.
https://github.com/apache/lucene/blob/7d35ae485807147460f63ea58ae495124e972e13/lucene/core/src/java/org/apache/lucene/search/Boolean2ScorerSupplier.java#L218C77-L218C98

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -202,6 +202,8 @@ Improvements
 
 * GITHUB#12999: Use Automaton for SurroundQuery prefix/pattern matching (Michael Gibney)
 
+* GITHUB#13043: Support getMaxScore of ConjunctionScorer for non top level scoring clause. (Shintaro Murakami)
+
 Optimizations
 ---------------------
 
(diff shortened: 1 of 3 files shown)


jpountz:
Thanks for looking into this, can you explain what kind of queries perform better with this change?

mrkm4ntr:
For this case. Suppose ScorerA, B, and C can return valid maxScore. If the ScorerD is dominant, larger minCompetitiveScore is set to WANDScorer. But ConjunctionScorer returns Infinity as maxScore and scaledMaxScore becomes 0. Then WANDScorer cannot skip docs well. Is my understanding correct?
```
DisjunctionMaxScorer(scorers=[
  WANDScorer(scorers=[ConjunctionScorer(ScorerA, ScorerB), ScorerC]),
  ScorerD
], tieBreakerMultiplier=0.0)
```

jpountz:
OK, I see, it's about conjunctions within disjunctions. Thanks for explaining.

mrkm4ntr:
Thanks. 9.10 as well?

jpountz:
Yes please.

PR #13046, opened 2024-01-29
Modernize BWC testing with parameterized tests

s1monw:
After working on #12829 I was quite surprised about the complexity our BWC have and how ancient the appear compared to the rest of our test suite. I took a step back and tried to modernize the tests to leverage RandomizedRunner Parameterized Tests that allow us to have more structured and hopefully more extendible BWC tests in the future. This change doesn't add any new tests but tries to make the ones we have more structured and support growth down the road. 
Basically, every index type got it's own Test class that doesn't require to loop over all the indices in each test. Each test case is run with all versions specified. Several sanity checks are applied in the base class to make individual tests smaller and much easier to read.


lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestDVUpdateBackwardsCompatibility.java
```diff
@@ -0,0 +1,268 @@
...
+  static final String SUFFIX = "";
+
+  public TestDVUpdateBackwardsCompatibility(Version version, String pattern) {
+    super(version, pattern);
+  }
+
+  @ParametersFactory(argumentFormatting = "Lucene-Version:%1$s; Pattern: %2$s")
+  public static Iterable<Object[]> testVersionsFactory() {
+    List<Object[]> params = new ArrayList<>();
+    // NOCOMMIT: why are we only testing one version here?
```
s1monw, line 59: maybe somebody has an answer to that

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestBinaryBackwardsCompatibility.java
```diff
@@ -0,0 +1,87 @@
...
+    for (Version version : BINARY_SUPPORTED_VERSIONS) {
+      params.add(new Object[] {version, createPattern(INDEX_NAME, SUFFIX_CFS)});
+      params.add(new Object[] {version, createPattern(INDEX_NAME, SUFFIX_NO_CFS)});
+    }
+    return params;
+  }
+
+  @Override
+  void verifyUsesDefaultCodec(Directory dir, String name) throws IOException {
+    // don't this will fail since the indices are not supportd
```
jpountz, line 55:
```suggestion
    // don't this will fail since the indices are not supported
```

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestEmptyIndexBackwardsCompatibility.java
```diff
@@ -0,0 +1,65 @@
...
+            .setMergePolicy(NoMergePolicy.INSTANCE);
+    try (IndexWriter writer = new IndexWriter(directory, conf)) {
+      writer.flush();
+    }
+  }
+
+  @ParametersFactory(argumentFormatting = "Lucene-Version:%1$s; Pattern: %2$s")
+  public static Iterable<Object[]> testVersionsFactory() {
+    List<Object[]> params = new ArrayList<>();
+    // NOCOMMIT: why are we only testing one version here?
```
jpountz, line 54: I would guess it's because it's more about testing the upgrade logic on empty indexes than about testing the format of old empty indexes.

jpountz (commented):
I like the approach. I'm curious of @dweiss' opinion since he's more familiar with the testing framework.

dweiss:
I like it. Whenever there is test repetition that can be driven by data, it should be driven by data. 

The only downside to using ParametersFactory is that it's something that is implemented by the randomizedtesting framework (not JUnit itself) and people may be unfamiliar with it - perhaps a comment on the class (or the parameterized constructor) pointing at where the arguments come from would make it easier to grasp how this thing works?

Plain JUnit alternatives do exist but they rely on a different test runner (Parameterized) - you wouldn't be able to extend from LuceneTestCase anymore (and you'd be outside of the randomization context). I don't see any other, more elegant, solutions.

dweiss:
Oh, one more problem is that you can't "see" the structure of tests before you actually run them (in an IDE). Don't know how much of an issue this is in practice but it's impossible to solve with JUnit4. Some IDEs may also have problems repeating a particular test if its name contains parameters and their names. But I don't think we should think in terms of IDE support - the final outcome and benefit for maintainability matters more.

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestIndexUpgradeBackwardsCompatibility.java
```diff
@@ -0,0 +1,260 @@
...
+  public static Iterable<Object[]> testVersionsFactory() throws IllegalAccessException {
+    Iterable<Object[]> allSupportedVersions =
+        allVersion(
+            TestBasicBackwardsCompatibility.INDEX_NAME,
+            TestBasicBackwardsCompatibility.SUFFIX_CFS,
+            TestBasicBackwardsCompatibility.SUFFIX_NO_CFS);
+    return allSupportedVersions;
+  }
+
+  /** Randomizes the use of some of hte constructor variations */
```
mikemccand, line 63: `hte` -> `the`

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestIndexUpgradeBackwardsCompatibility.java
```diff
@@ -0,0 +1,260 @@
...
+            TestBasicBackwardsCompatibility.INDEX_NAME,
+            TestBasicBackwardsCompatibility.SUFFIX_CFS))) {
+      TestBasicBackwardsCompatibility.createIndex(directory, true, false);
+    } else {
+      TestBasicBackwardsCompatibility.createIndex(directory, false, false);
+    }
+  }
+
+  public void testUpgradeOldSingleSegmentIndexWithAdditions() throws Exception {
+    // NOCOMMIT we use to have single segment indices but we stopped creating them at some point
```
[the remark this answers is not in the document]
mikemccand: I don't know why we stopped!  We should start again?  But maybe as followon?

mikemccand (approved):
This looks great!  Thank you for modernizing our ancient BWC tests @s1monw.  I left minor comments.  For the `// nocommit` questions let's make followon issues to address?  We should PnP this!

s1monw:
> We should PnP this!

What on earth means PnP? Mike, check out this search: https://www.google.com/search?q=pnp+acronym   wikipedia FTW

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/BackwardsCompatibilityTestBase.java
```diff
@@ -0,0 +1,252 @@
...
+        for (Object p : patterns) {
+          versionAndPatterns.add(new Object[] {v, p});
+        }
+      }
+    }
+    return versionAndPatterns;
+  }
+
+  public String indexName(Version version) {
+    return indexPattern.formatted(version);
```
uschindler, line 201: default locale...

uschindler (changes requested):
Looks fine to me. Theres only a forbiddenapis problem due to `String#formatted`.

Replace by `String.format(Locale.ROOT, indexPattern, version)`

mikemccand:
> > We should PnP this!
> 
> What on earth means PnP? Mike, check out this search: https://www.google.com/search?q=pnp+acronym wikipedia FTW

PnP = progress not perfection!  Not plug and play, not transistors, heh.

s1monw:
@uschindler I need to hear you speaking german to tell, you know that ;)

uschindler:
There is no formatted() with locale.

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/BackwardsCompatibilityTestBase.java
```diff
@@ -0,0 +1,253 @@
...
+        for (Object p : patterns) {
+          versionAndPatterns.add(new Object[] {v, p});
+        }
+      }
+    }
+    return versionAndPatterns;
+  }
+
+  public String indexName(Version version) {
+    return indexPattern.formatted(Locale.ROOT, version);
```
uschindler, line 202: This method does not exist with locale. You just format the locale and the version number :-)

`String#formatted()` is looking nice but locale dependent bullshit. You may need to change the comment above, too.

uschindler (changes requested):

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/BackwardsCompatibilityTestBase.java
uschindler: This is the only correct variant: `String.format(Locale.ROOT, indexPattern, version)`

uschindler: See also: https://github.com/elastic/elasticsearch/pull/81603

uschindler (approved):
Looks fine. All nocommits are todos now.

I think it's ready for a first round. P💊P

P.S.: About the code style I would prefer the list of versions as a `Stream.of(...)` and converting them to those sets of versions using a stream mapping. Those array declarations look always strange.

dweiss (approved):

s1monw:
thanks everybody... I will go and backport this to 9.x as well

HoustonPutman:
So I just finished the 8.11.3 release, and had some problems with the back-compat testing code. Obviously the ant/gradle switch and jdk versions meant that I had to do a lot of it manually, but I'm pretty sure the regex matching in the `update_backcompat_tests()` function in `addBackcompatIndexes.py` was broken by this commit.

I updated the files myself manually for 8.11.3 (on main and 9x), and I think they are correct (and the tests pass), but I would recommend checking them and re-checking that the python script works with these changes...

PR #13051, opened 2024-01-29
Replace `new HashSet<>(Arrays.asList())` with `EnumSet.of()`

sabi0:
For sets with enum values `EnumSet` provides a more efficient implementation.
Besides, `EnumSet.of()` results in a shorter, easier to understand code than `new HashSet<>(Arrays.asList())`.


uschindler (approved):

uschindler:
When backporting this to Lucene 9.x, it confliced because the sets have different contents in older version. I fixed this.

Nevertheless, the static final constants should be unmodifiable sets, so we should possibly use Java 9+ `Set.of()` instead of `EnumSet.of()`(which is modifiable) or wrap the `EnumSet` with `Collections.unmodifiableSet()`.

Could you open an issue about this?

sabi0:
> static final constants should be unmodifiable sets
> Could you open an issue about this?

#13055

PR #13052, opened 2024-01-29
Change `set.removeAll(list)` to `list.forEach(set::remove)`

sabi0:
When list size is greater than set size, `list.contains()` will be called for each set element resulting in O(n^2) complexity.
When list size is smaller the default `removeAll()` implementation will iterate the list anyway.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/UpgradeIndexMergePolicy.java
+++ b/lucene/core/src/java/org/apache/lucene/index/UpgradeIndexMergePolicy.java
@@ -106,6 +106,10 @@ public class UpgradeIndexMergePolicy extends FilterMergePolicy {
       // the resulting set contains all segments that are left over
       // and will be merged to one additional segment:
       for (final OneMerge om : spec.merges) {
+        // om.segments.forEach(::remove) is used here instead of oldSegments.keySet().removeAll()
+        // for performance reasons; om.segments size is always greater than oldSegments size, and
+        // then the AbstractSet#removeAll() implementation will iterate the set elements calling
+        // list.contains() for each of them, resulting in O(n^2) performance
         om.segments.forEach(oldSegments::remove);
       }
     }
--- a/lucene/core/src/java/org/apache/lucene/index/UpgradeIndexMergePolicy.java
+++ b/lucene/core/src/java/org/apache/lucene/index/UpgradeIndexMergePolicy.java
@@ -106,7 +106,7 @@ public class UpgradeIndexMergePolicy extends FilterMergePolicy {
       // the resulting set contains all segments that are left over
       // and will be merged to one additional segment:
       for (final OneMerge om : spec.merges) {
-        oldSegments.keySet().removeAll(om.segments);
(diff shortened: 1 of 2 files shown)


uschindler:
I am not sure if I like this code. Unfortunately the spec of `Set#removeAll()` is a bit stupid and should only use `argument.contains()` if it is a set (so almost constant contains is possible).

I'd like to ask to open a bug in OpenJDK so this gets fixed at some point.

Did you make a benchmarking if this affects Lucene at all?

uschindler:
I think we should apply this PR. When merging, Lucene always has a larger set before the merge so the new set set is always smaller. So `AbstractSet#removeAll` will always use the slow path.

We should open a bug report in JDK, that the optimization in `AbstractSet` should only be used if the collection passed as argument is also a set. In all other cases it is for sure faster to iterate over the collection passed as parameter and remove from actual set. The JDK impl has also some bugs with TreeSet as the contains of a TreeSet may not be symmetric with contains of a List.

uschindler (approved):
This is indeed a problem of jdk impl.

uschindler:
Maybe put a comment here describing why foreach is faster, especially because merging will always produce smaller sets so always triggering slow path.

@jpountz what do you think?

sabi0:
I've added a comment explaining the `forEach`.

[Opening a bug](https://bugs.openjdk.org/) in OpenJDK requires "OpenJDK Author" status which I do not have.

uschindler:
> I've added a comment explaining the `forEach`.
> 
> [Opening a bug](https://bugs.openjdk.org/) in OpenJDK requires "OpenJDK Author" status which I do not have.

I can do that.

lucene/core/src/java/org/apache/lucene/index/UpgradeIndexMergePolicy.java
```diff
@@ -106,7 +106,11 @@ public MergeSpecification findForcedMerges(
       // the resulting set contains all segments that are left over
       // and will be merged to one additional segment:
       for (final OneMerge om : spec.merges) {
-        oldSegments.keySet().removeAll(om.segments);
+        // om.segments.forEach(::remove) is used here instead of oldSegments.keySet().removeAll()
+        // for performance reasons; om.segments size is always greater than oldSegments size, and
```
jpountz, line 110: This sounds incorrect? `om.segments` is a subset of `oldSegments.keySet()` since the merge policy is only allowed to merge segments that are in `oldSegments.keySet()`. So the problem you are describing only occurs when `oldSegments.size() == om.segments.size()`, ie. the merge policy merges all mergeable segments in one go, which is probably a common case.

jpountz (commented):
The comment that explains the bug looks incorrect to me, but the fix looks good.

lucene/core/src/java/org/apache/lucene/index/UpgradeIndexMergePolicy.java
sabi0: Fixed

jpountz (approved):

PR #13053, opened 2024-01-29
Clean up AnyQueryNode code

sabi0:
- removed redundant field initializers
- fixed a typo in the field name: `minimumMatching_m_Elements`
- removed redundant `instanceof` checks
- replaced type casts with a pattern variable
- changed `size() == 0` to `isEmpty()`
- removed unnecessary explicit `toString()` calls


dweiss:
Thank you, LGTM. Could you add a changes entry as well?

lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/core/nodes/AnyQueryNode.java
```diff
@@ -21,37 +21,29 @@
...
-        if (clause instanceof FieldQueryNode) {
-
-          if (clause instanceof QueryNodeImpl) {
-            ((QueryNodeImpl) clause).toQueryStringIgnoreFields = true;
-          }
-
-          if (clause instanceof FieldableNode) {
-            ((FieldableNode) clause).setField(field);
-          }
+        if (clause instanceof FieldQueryNode qn) {
```
dweiss, line 37: This won't backport to 9x, unfortunately. I'll just correct it manually when backporting. Or should we leave this on main only, @uschindler - wdyt?

sabi0: I could undo the pattern variable part, leaving it to #12295.

uschindler: Yes please.

uschindler (approved):
Looks fine. Thanks for also fixing the typo in field name. 🤩

PR #13055, opened 2024-01-30
Make static final Set immutable

sabi0:
EnumSet.of() returns a mutable Set that should not be used for static final constants.


uschindler (approved):

uschindler:
Please add a changes.txt, as this is theoretically a backwards incompatible change.

uschindler:
I backported it. Like the previous PR this caused a conflict due to different stop tags, but this was easy to solve.

PR #13058, opened 2024-01-30
Fix knn vector visit limit fence post error

benwtrent:
I noticed while experimenting with brute-force search that our visitation limit is EXACTLY the number of filtered docs to hit. Consequently, if we happen to do brute force search and visit that exact number of vectors, we will fall back again to do brute-force a second time. This struck me as weird.

This commit adjusts the visit limit threshold to be a pure `gt` instead of a `gte` it was before.


jpountz:
I'm not sure about this one. Intuitively, if I configure a limit on the number of visited nodes, I may consider it a bug if we end up visiting more nodes than this limit.

Maybe we should instead fix the queries to set a limit of `cost+1`? (plus logic to protect from overflows)

benwtrent:
@jpountz good point. I can update the query limit to adjust instead of allowing the visit limit to increase above :).

jpountz:
Hmm, sorry it still doesn't feel completely right... It feels like the issue is that the collector doesn't distinguish between "exactly visitLimit hits have been collected but I collected all that I needed" and "exactly visitLimit hits have been collected but I would need to collect more to be done"?

benwtrent:
> "exactly visitLimit hits have been collected but I collected all that I needed" and "exactly visitLimit hits have been collected but I would need to collect more to be done"?

@jpountz I suppose the key thing is that the collector doesn't know if more candidates were left to explore. Being able to tell it this would extend its API and require some more code changed in the searcher. Basically, "earlyTerminated" would no longer be tied to numVisited, but whatever is doing the exploring would have to increment the visited count and when the visit limit is reached flag "earlyTerminated" if indeed it did.

This seems like a bigger complication to me with little benefit.

jpountz:
Yeah, it's annoying, but I agree that adding more APIs to fix this is overkill. Can you update your change to keep cost = cardinality, and only do the +1 when calling `approximateSearch()`?

benwtrent:
@jpountz there you go :). Only for `approximateSearch`

jpountz (approved):
Can you leave a quick comment explaining the +1? Otherwise LGTM.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -135,7 +135,7 @@ private TopDocs getLeafResults(
     }
 
     // Perform the approximate kNN search
-    TopDocs results = approximateSearch(ctx, acceptDocs, cost, knnCollectorManager);
+    TopDocs results = approximateSearch(ctx, acceptDocs, cost + 1, knnCollectorManager);
```
benwtrent, line 138:
```suggestion
    // We pass cost + 1 here to account for the edge case when we explore exactly cost vectors
    TopDocs results = approximateSearch(ctx, acceptDocs, cost + 1, knnCollectorManager);
```

jpountz (approved):

PR #13059, opened 2024-01-30
Fix normalization in TeluguAnalyzer

sabi0:
DecimalDigitFilter and IndicNormalizationFilter were mistakenly omitted.


rmuir (approved):
thank you for cetching this

PR #13060, opened 2024-01-30
Remove redundant field initializers

sabi0:
In many cases this allowed to mark the field as `final`.


lucene/analysis/phonetic/src/java/org/apache/lucene/analysis/phonetic/DaitchMokotoffSoundexFilter.java
```diff
@@ -32,10 +32,10 @@
  */
 public final class DaitchMokotoffSoundexFilter extends TokenFilter {
   /** true if encoded tokens should be added as synonyms */
-  protected boolean inject = true;
+  final boolean inject;
```
sabi0, line 35: I would make these fields `private`, but not sure if breaking backward compatibility here would be ok?

lucene/analysis/phonetic/src/java/org/apache/lucene/analysis/phonetic/PhoneticFilter.java
```diff
@@ -32,13 +32,13 @@
  */
 public final class PhoneticFilter extends TokenFilter {
   /** true if encoded tokens should be added as synonyms */
-  protected boolean inject = true;
+  final boolean inject;
```
sabi0, line 35: Could be changed to `private` too?

lucene/analysis/phonetic/src/java/org/apache/lucene/analysis/phonetic/DaitchMokotoffSoundexFilter.java
stefanvodita: Since this class is marked with `@lucene.experimental`, we can make the change.

lucene/analysis/phonetic/src/java/org/apache/lucene/analysis/phonetic/PhoneticFilter.java
stefanvodita: It's a strange case. We're not really breaking backwards compatibility, since no one could have accessed this field anyway with the class being final.

stefanvodita (approved):

PR #13066, opened 2024-02-02
Support getMaxScore of DisjunctionSumScorer for non top level scoring clause

mrkm4ntr:
### Description
Same as https://github.com/apache/lucene/pull/13043. <s>WANDScorer</s> BlockMaxConjunctionScorer doesn't work for disjunctions within <s>disjunctions</s> conjunctions.


mrkm4ntr:
@jpountz Could you take a look?

lucene/core/src/java/org/apache/lucene/search/DisjunctionSumScorer.java
```diff
@@ -43,10 +47,20 @@ protected float score(DisiWrapper topList) throws IOException {
...
+  }
+
   @Override
   public float getMaxScore(int upTo) throws IOException {
-    // It's ok to return a bad upper bound here since we use WANDScorer when
-    // we actually care about block scores.
-    return Float.MAX_VALUE;
+    double maxScore = 0;
+    for (Scorer scorer : scorers) {
+      maxScore += scorer.getMaxScore(DocIdSetIterator.NO_MORE_DOCS);
```
jpountz, line 62: Computing the max score up to `upTo` should be enough, no need to use `NO_MORE_DOCS`?

lucene/core/src/java/org/apache/lucene/search/DisjunctionSumScorer.java
```diff
@@ -43,10 +47,20 @@ protected float score(DisiWrapper topList) throws IOException {
     return (float) score;
   }
 
+  @Override
+  public int advanceShallow(int target) throws IOException {
+    for (Scorer scorer : scorers) {
+      scorer.advanceShallow(target);
+    }
+    return super.advanceShallow(target);
```
jpountz, line 55: Should we take the min across all clauses or something like that instead of returning NO_MORE_DOCS?

jpountz (approved):
I see that we have a rewrite rule that inlines nested disjunctions, so presumably this is more about disjunctions within conjunctions than disjunctions within disjunctions?

We have tasks in nightly benchmarks that should be impacted by this change, e.g. http://people.apache.org/~mikemccand/lucenebench/AndHighOrMedMed.html, I'm curious if you tried to run them?

mrkm4ntr:
@jpountz Thank you for your review.

> I see that we have a rewrite rule that inlines nested disjunctions, so presumably this is more about disjunctions within conjunctions than disjunctions within disjunctions?

Ah, yes. That's true.

> We have tasks in nightly benchmarks that should be impacted by this change, e.g. http://people.apache.org/~mikemccand/lucenebench/AndHighOrMedMed.html, I'm curious if you tried to run them?

I'm not sure how to run these benchmarks. Is there any good instruction?

mrkm4ntr:
This? https://github.com/mikemccand/luceneutil

jpountz:
This indeed! It's a bit manual, but it's what powers nightly benchmarks and what we run to check the performance impact of the changes that we merge.

mrkm4ntr:
Thanks. The result is here. 
```
                AndMedOrHighHigh       21.24      (7.1%)       20.61      (3.5%)   -3.0% ( -12% -    8%) 0.091
                 AndHighOrMedMed       30.97      (4.0%)       30.51      (3.5%)   -1.5% (  -8% -    6%) 0.216
```
I run this with `verifyCount=false`. The hit counts were smaller than baseline. This means we can skip more items with this change, right?

jpountz:
> This means we can skip more items with this change.

Indeed!

lucene/core/src/java/org/apache/lucene/search/DisjunctionSumScorer.java
```diff
@@ -43,10 +47,23 @@ protected float score(DisiWrapper topList) throws IOException {
...
+  }
+
   @Override
   public float getMaxScore(int upTo) throws IOException {
-    // It's ok to return a bad upper bound here since we use WANDScorer when
-    // we actually care about block scores.
-    return Float.MAX_VALUE;
+    double maxScore = 0;
+    for (Scorer scorer : scorers) {
+      maxScore += scorer.getMaxScore(upTo);
```
jpountz, line 65: We should not call `getMaxScore` on scorers whose doc ID is already beyond `upTo`. See `WANDScorer#getMaxScore`.

jpountz (commented):
Sorry I just found a small problem, otherwise the change looks good to me.

mrkm4ntr:
@jpountz Thank you for your review! I fixed the issue.

lucene/core/src/java/org/apache/lucene/search/DisjunctionSumScorer.java
```diff
@@ -43,10 +47,25 @@ protected float score(DisiWrapper topList) throws IOException {
     return (float) score;
   }
 
+  @Override
+  public int advanceShallow(int target) throws IOException {
+    int min = DocIdSetIterator.NO_MORE_DOCS;
+    for (Scorer scorer : scorers) {
+      if (scorer.docID() <= target) {
+        min = Math.max(min, scorer.advanceShallow(target));
```
jpountz, line 55: I think you meant to do `Math.min` here, not `Math.max`?

```suggestion
        min = Math.min(min, scorer.advanceShallow(target));
```

mrkm4ntr: Indeed! Thank you for catching this!

PR #13067, opened 2024-02-02
Enable parent field in sorted bwc tests

s1monw:
This enables the optional parent field in BWC tests from 9.10 on. This will need to be forward ported to main branch where the parent field is required to these tests since they add document blocks during tests.


lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestIndexSortBackwardsCompatibility.java
```diff
@@ -82,14 +84,20 @@ public void testSortedIndexAddDocBlocks() throws Exception {
...
       // add 10 docs
       for (int i = 0; i < 10; i++) {
         Document child = new Document();
         child.add(new StringField("relation", "child", Field.Store.NO));
         child.add(new StringField("bid", "" + i, Field.Store.NO));
-        child.add(new NumericDocValuesField("dateDV", i));
+        if (version.onOrAfter(FIRST_PARENT_DOC_VERSION)
+            == false) { // only add this to earlier versions
+          child.add(new NumericDocValuesField("dateDV", i));
+        }
```
jpountz, line 100: Why do we need this change?

s1monw: if we are testing against an earlier version we need to add this to the child document otherwise it will not sort correctly. With parent field added this is the behavior we want test. We can remove it once we move to 11.x

jpountz: Ah thanks, that makes sense. I had read the condition under the if statement backwards.

jpountz (approved):

jpountz:
FYI I had not paid attention to the timing but this got merged after I cut the branch for 9.10, so I just bumped the version in this change to 9.11.

PR #13068, opened 2024-02-03
Move `brToString(BytesRef)` to `ToStringUtils`

sabi0:
I noticed there are `brToString()` methods in 16 different classes.
What do you think of providing a shared implementation in the `ToStringUtils` class?

If you find the proposal useful I will proceed with removing the other copies.

Otherwise the `ToStringUtils` class could probably be deleted - the two methods there `byteArray()` and `longHex()` are not used anywhere.


sabi0:
That's what I meant with
> I will proceed with removing the other copies

Did not want to waste time on this in case the suggestion would be rejected.

I will push the changes shortly.

lucene/core/src/java/org/apache/lucene/util/ToStringUtils.java
```diff
@@ -32,11 +32,37 @@ public static void byteArray(StringBuilder buffer, byte[] bytes) {
...
   }
+
+  @SuppressWarnings("unused")
+  public static String brToString(BytesRef b) {
+    if (b == null) {
+      return "null";
+    }
+    try {
+      return b.utf8ToString() + " " + b;
+    } catch (Throwable t) {
```
uschindler, line 54: This method should not catch Throwable, can we not make a multi-catch out of it and explicitely:

- `AssertionFailedError` (if we hit an Assertion)
- `RuntimeException` (anything is wrong and we have wrong offsets or bytes array is too short or incomplete surrogates

uschindler (changes requested):

lucene/core/src/java/org/apache/lucene/util/ToStringUtils.java
sabi0: Nice catch. Thank you.

lucene/core/src/java/org/apache/lucene/util/ToStringUtils.java
```diff
@@ -32,11 +32,37 @@ public static void byteArray(StringBuilder buffer, byte[] bytes) {
...
   public static String longHex(long x) {
     char[] asHex = new char[16];
     for (int i = 16; --i >= 0; x >>>= 4) {
       asHex[i] = HEX[(int) x & 0x0F];
     }
     return "0x" + new String(asHex);
   }
+
+  @SuppressWarnings("unused")
+  public static String brToString(BytesRef b) {
```
mikemccand, line 48: Could we rename it to `bytesRefToString`?

mikemccand (commented):
Thank you @sabi0 -- I can't believe how many (badly) forked copies of this method had proliferated.

I just left one naming comment, otherwise this looks awesome.  Naming is the hardest part :)

lucene/core/src/java/org/apache/lucene/util/ToStringUtils.java
mikemccand: Or maybe just `toString(BytesRef b)`?

mikemccand, line 48: Maybe also add a javadoc explaining how this differs from `BytesRef.toString`?  E.g. this one does not assume valid UTF-8 but the other one days (I think)?

sabi0: I think `toString(br)` will look confusingly similar to `br.toString()`.

How about `format(br)`? Or `prettyPrint(br)` maybe?
(`bytesRefToString(br)` does look a bit long)

dweiss: I think this method could be the default toString on BytesRef, actually... in most cases when you wish to dump it using toString it's for debugging purposes: an array of bytes would always be there, utf8 as an additional helpful hint... why add another method? Even the fact that this additional method exists points at the need to actually dump BytesRef in a more verbose way...

sabi0: I thought of that too. But reckoned that could be too much of a change if all BytesRef's start spitting out text.

I guess the format will have to be changed somewhat to avoid the BytesRef's text "blending in" with the preceding text:
```
... previous " + bytesRef;  // ... previous term [ww xx yy zz]
```

Also there is an unfortunate side-effect of an exception being thrown for "binary" data.

dweiss: Ok. Small steps. We can always redirect and inline this method later on.

sabi0: Shall I rename it to `bytesRefToString()` ?

dweiss: Yes, please do. Better a bit more verbose than cryptic.

uschindler (approved):
Looks fine to me. 

We must move the changes entry to next version as the release process already started.

uschindler:
Please move entry to 9.11.0

uschindler:
To the other reviewers: I would like to merge this as a first step. To me this looks fine as it removes the code duplication and we have a better method name. Ok?

If we remove the static method and instead use `toString()` for that, should possibly another issue. I like the discussion started by @mikemccand and @dweiss.

I plan to merge this late afternoon German time.

uschindler:
Backport was easy, no conflicts. I also checked with `fgrep -R brToString *` but no occurrences found.

Thanks @sabi0 !

PR #13069, opened 2024-02-04
Fix NPE in TestReqOptSumScorer.testFilterRandomRareOpt

easyice:
The value `C` of field `f` may not appear in all segments.

```
   >     java.lang.NullPointerException: Cannot invoke "org.apache.lucene.search.Scorer.twoPhaseIterator()" because "s2" is null
   >         at __randomizedtesting.SeedInfo.seed([E000CB24BEEF3CDA:634D0259C7E92DEB]:0)
   >         at org.apache.lucene.tests.search.CheckHits.doCheckMaxScores(CheckHits.java:738)
   >         at org.apache.lucene.tests.search.CheckHits.checkTopScores(CheckHits.java:698)
   >         at org.apache.lucene.search.TestReqOptSumScorer.doTestRandom(TestReqOptSumScorer.java:379)
   >         at org.apache.lucene.search.TestReqOptSumScorer.testFilterRandomRareOpt(TestReqOptSumScorer.java:267)
```

It cannot be reproduced stably, but there is a high probability.

```
./gradlew test --tests TestReqOptSumScorer.testFilterRandomRareOpt -Dtests.seed=E000CB24BEEF3CDA -Dtests.nightly=true -Dtests.locale=hsb-DE -Dtests.timezone=Africa/Djibouti -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```

### DIFF
--- a/lucene/test-framework/src/java/org/apache/lucene/tests/search/CheckHits.java
+++ b/lucene/test-framework/src/java/org/apache/lucene/tests/search/CheckHits.java
@@ -733,6 +733,9 @@ public class CheckHits {
         assertTrue(s2 == null || s2.iterator().nextDoc() == DocIdSetIterator.NO_MORE_DOCS);
         continue;
       }
+      if (s2 == null) {
+        continue;
+      }
       TwoPhaseIterator twoPhase1 = s1.twoPhaseIterator();
       TwoPhaseIterator twoPhase2 = s2.twoPhaseIterator();
       DocIdSetIterator approx1 = twoPhase1 == null ? s1.iterator() : twoPhase1.approximation();


easyice:
This will also fix test failure for TestReqOptSumScorer.testFilterRandomFrequentOpt

```
./gradlew test --tests TestReqOptSumScorer.testFilterRandomFrequentOpt -Dtests.seed=70A6B2258BEC1D96 -Dtests.nightly=true -Dtests.locale=mni-Beng-IN -Dtests.timezone=America/Indianapolis -Dtests.asserts=true -Dtests.file.encoding=UTF-8
``` 


```
org.apache.lucene.search.TestReqOptSumScorer > testFilterRandomFrequentOpt FAILED
    java.lang.NullPointerException: Cannot invoke "org.apache.lucene.search.Scorer.twoPhaseIterator()" because "s2" is null
        at __randomizedtesting.SeedInfo.seed([70A6B2258BEC1D96:94B3EF7331EC2DC4]:0)
        at org.apache.lucene.tests.search.CheckHits.doCheckMaxScores(CheckHits.java:737)
        at org.apache.lucene.tests.search.CheckHits.checkTopScores(CheckHits.java:698)
        at org.apache.lucene.search.TestReqOptSumScorer.doTestRandom(TestReqOptSumScorer.java:379)
        at org.apache.lucene.search.TestReqOptSumScorer.testFilterRandomFrequentOpt(TestReqOptSumScorer.java:263)
```

dweiss (approved):
This looks good to me, thank you for investigating. I am not that familiar with this part of the code so it'd be good to have a second opinion though. @jpountz , maybe?

jpountz (approved):
LGTM. I looked into this failure, this is due to the weight wrapper that we use for caching, which never produces `null` scorers. So s1 is not `null` because its required clause (which doesn't match anything) is cached, but s2 is `null` because its required clause (which is the same, so it doesn't match anything either) doesn't get cached.

easyice:
Thanks for reviewing! @dweiss @jpountz

jpountz:
Hmm actually we should probably assert that s1.iterator().nextDoc() returns NO_MORE_DOCS in this `if` block, to make sure that both scorers produce the same hits? @easyice Can you open a follow-up PR?

easyice:
Nice catch! I will fix it.

PR #13072, opened 2024-02-05
Terminate automaton when it can match all suffixes, and match suffixes directly.

vsop-479:
For PrefixQuery, we can terminate the automaton on current term if we have matched the whole prefix, and match this term directly.
Furthermore, if there is a subBlock, we could match all its' sub terms.


vsop-479:
@jpountz Please take a look when you get a chance.

jpountz:
`IntersectsTermsEnum` is a bit scary to me, maybe @mikemccand can take a look, I expect him to be more familiar with it.

mikemccand:
I will have a look -- thanks for the ping @jpountz.

lucene/core/src/java/org/apache/lucene/util/automaton/Automaton.java
```diff
@@ -70,6 +70,7 @@ public class Automaton implements Accountable, TransitionAccessor {
   private int[] states;
 
   private final BitSet isAccept;
+  private final BitSet terminable;
```
mikemccand, line 73: At first I couldn't understand what `terminable` means.

If a bit is set for a state, does that mean this state accepts everything from now on (all possible suffixes)?

Could we maybe rename it to `isMatchAllSuffix` or so?

lucene/core/src/java/org/apache/lucene/util/automaton/Automaton.java
```diff
@@ -92,6 +93,7 @@ public Automaton() {
   public Automaton(int numStates, int numTransitions) {
     states = new int[numStates * 2];
     isAccept = new BitSet(numStates);
+    terminable = new BitSet(numStates);
```
mikemccand, line 96: One spooky thing about this new `BitSet` is it is "best effort" now?  I.e. one could create an Automaton that indeed has some states that match all suffixes, but forget to set the bit here?  E.g. if I build a `RegexpQuery` that is actually a `PrefixQuery` we won't set this?

Everything else about `Automaton` today is fundamental (states, transitions, isAccept) and necessary, but this new member is more a best effort optimization?

lucene/core/src/java/org/apache/lucene/util/automaton/RunAutomaton.java
```diff
@@ -67,12 +68,16 @@ protected RunAutomaton(Automaton a, int alphabetSize) {
     points = a.getStartPoints();
     size = Math.max(1, a.getNumStates());
     accept = new FixedBitSet(size);
+    terminable = new FixedBitSet(size);
```
mikemccand, line 71: Perhaps, instead of storing this concept in `Automaton`, we could solely store it in `RunAutomaton`, if we can efficiently find final states that effectively have `.*` transitions to themselves?

mikemccand (commented):
This is a clever optimization!  You recognize that this Automaton will match all possible suffixes in this state, and so more efficiently enumerate all terms from block tree under that state.

I have concerns about storing this in `Automaton` itself, and the naming was confusing to me :)  Could we instead store it in `RunAutomaton`?  Or, possibly, do it on the fly in `IntersectEnum` by detecting a state that is both accept and has a `.*` transition back onto itself?

Have you tried to measure any performance change with this?  E.g. you could run a `luceneutil` benchy with just `PrefixQuery`, or, `Regexp/WildcardQuery` that also have this property (match-all states in their automata).

vsop-479:
@mikemccand Thanks for you suggestion, I am trying to implement it. 
> Have you tried to measure any performance change with this? E.g. you could run a luceneutil benchy with just PrefixQuery, or, Regexp/WildcardQuery that also have this property (match-all states in their automata).

I am working on this too.

@jpountz Thanks for you reply.

rmuir:
I think the optimization may be similar to the one done in AutomatonTermsEnum?
 https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/AutomatonTermsEnum.java#L149-L153

When "ping-ponging" the term dictionary against the automaton, it tracks `visited` bitset and looks for such loops in the automaton. when it finds one, it temporarily acts like a TermRangeQuery.

I think, it works a bit more general than just prefixquery and also helps with regex and wildcard queries too.

vsop-479:
> I think the optimization may be similar to the one done in AutomatonTermsEnum?

Thanks for reminding that, I will dig into `AutomatonTermsEnum`'s optimization.

lucene/core/src/java/org/apache/lucene/util/automaton/Automaton.java
vsop-479: > If a bit is set for a state, does that mean this state accepts everything from now on (all possible suffixes)?

Yes, you are right about this.

> Could we maybe rename it to isMatchAllSuffix or so?

I will rename it (and remove it to `RunAutomaton`)

lucene/core/src/java/org/apache/lucene/util/automaton/Automaton.java
vsop-479: > E.g. if I build a RegexpQuery that is actually a PrefixQuery we won't set this?

Yes, I just implemented this in `PrefixQuery` in current version.

> Everything else about Automaton today is fundamental (states, transitions, isAccept) and necessary, but this new member is more a best effort optimization?

Yes, I will remove it.

vsop-479:
There is still problem in the implementation of detecting whether accept state can match all suffix (`RunAutomaton.canMatchAllSuffix`), I am trying to figure it out.

On the other hand, I will dig into AutomatonTermsEnum's optimization.

vsop-479:
@mikemccand 
I renamed the field used to indicate wether an accept state can match all suffixes, and detected it in `RunAutomaton`.

vsop-479:
> I think the optimization may be similar to the one done in AutomatonTermsEnum?

It seems the optimization of `AutomatonTermsEnum` is to improve iterating term mode, from seeking bytes from DFA and seek termsEnum(`seekCeil`), to simply sequential reads the termsEnum, after finding a loop(`setLinear`).
In both mode, `AutomatonTermsEnum` needs to check if the term is accepted by running automaton.

If I am not mistaken, this optimization is different from `AutomatonTermsEnum`'s. It directly matches all reminding suffixes and sub blocks, after detecting an accept state with a `.*` transition back onto itself.

Or maybe you mean we can improve `AutomatonTermsEnum`'s optimization, to implement this optimization's effect? @rmuir

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/IntersectTermsEnum.java
```diff
@@ -534,6 +555,9 @@ private BytesRef _next() throws IOException {
         // Match!  Recurse:
         copyTerm();
         currentFrame = pushFrame(state);
+        if (matchAllSuffix) {
```
mikemccand, line 558: Couldn't you just do `currentFrame.matchAllSuffix = matchAllsuffix`?

lucene/core/src/java/org/apache/lucene/util/automaton/RunAutomaton.java
```diff
@@ -67,12 +68,16 @@ protected RunAutomaton(Automaton a, int alphabetSize) {
...
     size = Math.max(1, a.getNumStates());
     accept = new FixedBitSet(size);
+    matchAllSuffix = new FixedBitSet(size);
     transitions = new int[size * points.length];
     Arrays.fill(transitions, -1);
     Transition transition = new Transition();
     for (int n = 0; n < size; n++) {
       if (a.isAccept(n)) {
         accept.set(n);
+        if (canMatchAllSuffix(n)) {
```
mikemccand, line 78: Maybe rename to `isMatchAllSuffix`?

lucene/core/src/java/org/apache/lucene/util/automaton/RunAutomaton.java
```diff
@@ -96,6 +101,35 @@ protected RunAutomaton(Automaton a, int alphabetSize) {
...
+      Transition transition = new Transition();
+      automaton.getTransition(state, 0, transition);
+      if (transition.dest == state && transition.min == 0 && transition.max == 255) {
+        return true;
+      }
+    }
+
+    // Apply to RegexpQuery, WildcardQuery.
+    for (int i = 0; i < numTransitions; i++) {
+      Transition transition = new Transition();
```
mikemccand, line 119: Maybe allocate this scratch `Transition` once up front at the start of the method?

lucene/core/src/java/org/apache/lucene/util/automaton/RunAutomaton.java
```diff
@@ -96,6 +101,35 @@ protected RunAutomaton(Automaton a, int alphabetSize) {
     }
   }
 
+  /** Returns true if this state can accept everything(all remaining suffixes). */
+  private boolean canMatchAllSuffix(int state) {
+    assert automaton.isAccept(state);
+    int numTransitions = automaton.getNumTransitions(state);
+    // Apply to PrefixQuery, TermRangeQuery.
+    if (numTransitions == 1) {
```
mikemccand, line 109: Can we remove this special case?  Just let the `for` loop below handle the 1-transition case too?

Edit: hmm, I see, it is subtly different: this is checking for max label `255` but the loop below is checking `127`, hmmm.  This is a bit messy -- this low level of code shouldn't be specializing to different automata that come from the  high level queries.  Can we use `alphabetSize-1` as the `transition.max` check instead?  But, separately, we need to figure out why `Regexp/WildcardQuery` are compiling down to `127` as their max on `.*` suffix transitions?  That is not even correct for matching UTF-8 encoded terms.

Perhaps we could also add tests cases for custom `Automata` passed to `AutomatonQuery` matching sometimes binary (non-UTF8) terms?

lucene/core/src/java/org/apache/lucene/util/automaton/RunAutomaton.java
```diff
@@ -96,6 +101,35 @@ protected RunAutomaton(Automaton a, int alphabetSize) {
...
+      if (transition.dest == state && transition.min == 0 && transition.max == 255) {
+        return true;
+      }
+    }
+
+    // Apply to RegexpQuery, WildcardQuery.
+    for (int i = 0; i < numTransitions; i++) {
+      Transition transition = new Transition();
+      automaton.getTransition(state, i, transition);
+      if (transition.dest == state && transition.min == 0 && transition.max == 127) {
```
mikemccand, line 121: Maybe refactor this to:

```
  if (transition.min == 0 && transition.max == 127) {
    if (transition.dest == state) {
      return true;
    } else if (automaton.isAccept(transition.dest) {
      // recurse
      return isMatchAllSuffix(transition.dest);
    }
  }
```

mikemccand (commented):
Thanks @vsop-479 -- this looks closer.  I like that the opto is now contained under `RunAutomaton`, but I'm confused/concerned about sometimes checking for `255` max label and other times `127` depending on which query.

lucene/core/src/java/org/apache/lucene/util/automaton/RunAutomaton.java
vsop-479: > Maybe rename to isMatchAllSuffix?

I don't like that name too. But, there is a method named isMatchAllSuffix, which indicates whether a state can accept all remaining suffixes (similar to `isAccept`). 

````
public final boolean isMatchAllSuffix(int state) {
    return matchAllSuffix.get(state);
  }
````

File: lucene/core/src/java/org/apache/lucene/util/automaton/RunAutomaton.java (hunk 7 above)
Comment (vsop-479) [reply lucene/core/src/java/org/apache/lucene/util/automaton/RunAutomaton.java:110]: > we need to figure out why Regexp/WildcardQuery are compiling down to 127 as their max on .* suffix transitions?

These queries' (including `AutomatonQuery`)`Automaton` like this: [0, 127]: 3 -> 3; [194, 194]: 3 -> 4; [128, 191]: 4 -> 3. assume 3 is an accept state. 
It is more complex to detect wether a state can accept all remaining suffixes for these queries, because its accept states split many transitions like:[0, 127], [194, 223], [224, 239], [240, 243], [244], etc.

I am still working on this, any suggestion is welcome @mikemccand.

> Perhaps we could also add tests cases for custom Automata passed to AutomatonQuery matching sometimes binary (non-UTF8) terms?

Added.

**Comment by vsop-479** [comment]
> Have you tried to measure any performance change with this? E.g. you could run a luceneutil benchy with just PrefixQuery, or, Regexp/WildcardQuery that also have this property (match-all states in their automata).

I measured it with current implementation with `wikimedium1m`, the result is unstable:

````
Prefix3      879.20      (9.1%)      995.37      (4.2%)   13.2% (   0% -   29%) 0.062
Prefix3      924.98      (9.9%)     1042.17      (7.9%)   12.7% (  -4% -   33%) 0.083
```` 

````
Prefix3     1480.94      (8.8%)     1559.89      (6.4%)    5.3% (  -9% -   22%) 0.195
Prefix3     1242.80      (6.9%)     1307.30      (5.3%)    5.2% (  -6% -   18%) 0.299
Prefix3      177.54      (1.3%)      202.74      (6.3%)   14.2% (   6% -   22%) 0.000
````

File: lucene/core/src/java/org/apache/lucene/util/automaton/RunAutomaton.java (hunk 7 above)
Comment (vsop-479) [reply lucene/core/src/java/org/apache/lucene/util/automaton/RunAutomaton.java:110]: > These queries' (including AutomatonQuery)Automaton like this: 3 -> 3: [0, 127]; 3 -> 4: [194, 194]; 4 -> 3: [128, 191]. assume 3 is an accept state.

@mikemccand 
I can track an accept state's other transitions, to check wether these transitions can finally ended on an accept (typically transited by  [128, 191]). But i am not sure wether it is  enough to judge an state can match all suffix, even not sure wether it is necessary, since maybe it is equivalent  to just check the [0, 127] transition's dest is an accept state.

Comment (vsop-479) [reply lucene/core/src/java/org/apache/lucene/util/automaton/RunAutomaton.java:110]: > we need to figure out why Regexp/WildcardQuery are compiling down to 127 as their max on .* suffix transitions?

I think we split the transition([0, 1114111]) with `utf8` edges in `UTF32ToUTF8.convertOneEdge`.

Comment (vsop-479) [reply lucene/core/src/java/org/apache/lucene/util/automaton/RunAutomaton.java:110]: I think I can detecting a match all suffix state for `Regexp/WildcardQuery`, in `UTF32ToUTF8.convert` after `convertOneEdge` like this:
````
// Writes new transitions into pendingTransitions:
convertOneEdge(utf8State, destUTF8, scratch.min, scratch.max);
        
// Set match all suffix state.
if(scratch.min == 0 && scratch.max == 1114111 && utf8.isAccept(utf8State) && utf8.isAccept(destUTF8)){
  utf8.setMatchAllSuffix(utf8State, true);
}
````
Which is simple and reliable, but will violate the rule below:

> Everything else about Automaton today is fundamental (states, transitions, isAccept) and necessary, but this new member is more a best effort optimization?

Also: I can check whether a candidate state can finally ended on an accept by [128, 191], which is added in `UTF32ToUTF8.adll`, in `RunAutomaton`.

File: lucene/core/src/java/org/apache/lucene/util/automaton/RunAutomaton.java (hunk 5 above)
Comment (vsop-479) [reply lucene/core/src/java/org/apache/lucene/util/automaton/RunAutomaton.java]: > Maybe rename to isMatchAllSuffix?

Since there is a method named `isMatchAllSuffix`, I renamed it to `detectMatchAllSuffix`.

**Comment by vsop-479** [comment]
@mikemccand 
I think this change is close for `PrefixQuery`, `TermRangeQuery`, and custom binary `AutomatonQuery`.

As for `RegexpQuery` and `WildcardQuery`, whose automata converted by `UTF32ToUTF8`. We just get accept transitions [0, 127] in `RunAutomaton`, I am not sure whether it is enough that just check the accept state is transition is [0, 127].

Another approach is set matchAllSuffix in `UTF32ToUTF8`, but need to add a unFundamental member to `Automaton`.

Or, can we just push this change for `Prefix/TermRangeQuery`, and leave a `TODO` for `Regexp/WildcardQuery` temporarily?

**Comment by vsop-479** [comment]
Conflicts resolved.

**Comment by vsop-479** [comment]
Hi @mikemccand , Can we push on this change?

File: lucene/core/src/java/org/apache/lucene/util/automaton/RunAutomaton.java [hunk 9]
```diff
@@ -96,6 +101,35 @@ protected RunAutomaton(Automaton a, int alphabetSize) {
...
+      if (transition.min == 0 && transition.max == 127) {
+        if (transition.dest == state) {
+          return true;
+        } else if (automaton.isAccept(transition.dest)) {
+          // recurse
+          return detectMatchAllSuffix(transition.dest);
+        }
+      }
+    }
+    return false;
```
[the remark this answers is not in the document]
vsop-479: That is incorrect, after UTF32ToUTF8 conversion, transitions intentionally have gaps for illegal UTF-8 byte sequences (e.g. [128, 193], surrogate range). These bytes will never appear in indexed terms, so gap-based rejection would incorrectly disqualify valid matchAllSuffix states.

PR #13075, opened 2024-02-05
Make date parsing more flexible for linedocsfile (europarl, enwiki)

### LINKED ISSUE #13073: java.lang.AssertionError in backward compat tests ("failed to parse ... as date")
### Description

These failures do reproduce but you need the linedocsfile. I'll take a look.

_No response_

dweiss:
Fixes #13073.

### DIFF
--- a/lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestIndexSortBackwardsCompatibility.java
+++ b/lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestIndexSortBackwardsCompatibility.java
@@ -20,7 +20,6 @@ import com.carrotsearch.randomizedtesting.annotations.ParametersFactory;
 import java.io.IOException;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
-import java.time.LocalTime;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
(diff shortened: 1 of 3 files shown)


dweiss:
This makes date parsing accept both europarl and enwiki. The tests still assume the docs come from europarl though and fail on the large enwiki.random.lines.txt:
```
gradlew -p lucene\backward-codecs test -Dtests.linedocsfile=/../enwiki.random.lines.txt
```

uschindler (approved):
Looks fine. Why did you create a `Function<String,LocalDateTime>` for parsing instead of a simple static method? I think you did this to hide the formatter instances?

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestIndexSortBackwardsCompatibility.java
```diff
@@ -147,6 +150,36 @@ public void testSortedIndex() throws Exception {
...
+            new DateTimeFormatterBuilder()
+                .parseStrict()
+                .parseCaseInsensitive()
+                .appendPattern("dd-MMM-uuuu HH:mm:ss['.'SSS]")
+                .toFormatter(Locale.ROOT);
+
+        @Override
+        public LocalDateTime apply(String s) {
+          if (s.matches("^[0-9]{4}-[0-9]{2}-[0-9]{2}$")) {
+            return euroParl.parse(s, LocalDate::from).atTime(LocalTime.MIDNIGHT);
```
uschindler, line 176: `.atStartOfDay()` would be shorter.

uschindler:
P.S.: Maybe add a quick test for the parsing logic to check that both formats are accepted.

dweiss:
> Looks fine. Why did you create a `Function<String,LocalDateTime>` for parsing instead of a simple static method? I think you did this to hide the formatter instances?

Yes, correct.

dweiss:
> P.S.: Maybe add a quick test for the parsing logic to check that both formats are accepted.

Maybe it'd be better to move this logic into LineFileDocs so that the date field's value is normalized somehow? Then we can add TestLineFileDocs and test it there. Adding a test method to TestIndexSortBackwardsCompatibility will multiply it over all the parameters, which doesn't make sense.

dweiss:
I've moved that utility function to LineFileDocs and added a basic test case to TestLineFileDocs (good idea). I feel tempted to normalize the date field's value in LineFileDocs but this may break other stuff so better to leave it for a follow-up issue.

dweiss:
I'll merge this in so that we can avoid jenkins failures. If there has to be a follow-up, I'll open another issue.

s1monw:
LGTM thanks

PR #13087, opened 2024-02-07
Make `static final Set` constants immutable

sabi0:
For the mutable "global" Sets changed `Collections.synchronizedSet()` to `ConcurrentHashMap.newKeySet()` for better performance.


kkmr (approved):

lucene/core/src/java/org/apache/lucene/store/NativeFSLockFactory.java
```diff
@@ -65,7 +64,7 @@ public final class NativeFSLockFactory extends FSLockFactory {
   /** Singleton instance */
   public static final NativeFSLockFactory INSTANCE = new NativeFSLockFactory();
 
-  private static final Set<String> LOCK_HELD = Collections.synchronizedSet(new HashSet<String>());
+  private static final Set<String> LOCK_HELD = ConcurrentHashMap.newKeySet();
```
uschindler, line 67: I think LOCK_HELP should be `Set<Path>`. This would make more sense as it would also be working with different slashes.

I know this isn't related to this cleanup task, but I just have seen it.

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/BackwardsCompatibilityTestBase.java
```diff
@@ -80,7 +81,7 @@ public abstract class BackwardsCompatibilityTestBase extends LuceneTestCase {
       // make sure we never miss a version.
       assertTrue("Version: " + version + " missing", binaryVersions.remove(version));
     }
-    BINARY_SUPPORTED_VERSIONS = binaryVersions;
+    BINARY_SUPPORTED_VERSIONS = Collections.unmodifiableSet(binaryVersions);
```
uschindler, line 84: We changed that code recently, maybe check also the other Ancient indexes test.

lucene/core/src/java/org/apache/lucene/util/VirtualMethod.java
```diff
@@ -74,14 +73,13 @@
...
-  private static final Set<Method> singletonSet =
-      Collections.synchronizedSet(new HashSet<Method>());
+  private static final Set<Method> singletonSet = ConcurrentHashMap.newKeySet();
 
   private final Class<C> baseClass;
   private final String method;
   private final Class<?>[] parameters;
   private final ClassValue<Integer> distanceOfClass =
-      new ClassValue<Integer>() {
+      new ClassValue<>() {
```
uschindler, line 82: this is unrelated and does not work with Java 11 (when you subclass on "new" you had to be explicit in former times). I'd like to be explicit here.

lucene/CHANGES.txt
```diff
@@ -260,6 +260,8 @@ Improvements
 
 * GITHUB#13092: `static final Map` constants have been made immutable (Dmitry Cherniachenko)
 
+* GITHUB#13087: Some `static final Set` constants were mutable - rectified. (Dmitry Cherniachenko)
```
uschindler, line 263: Move to 9.11

uschindler: Also add the new set impl.

uschindler (changes requested):
This PR has 2 different types of changes: One is to use ConcurrentHashMap for some synchronized sets, the other one is making stuff final and unmodifiable.

I generally agree, but we could better communicate that in the changes log.

uschindler:
I checked backwards compatibility: Except the scandinavian analyzer everything is tests or implementation private.

I think for scandinavian normalizer we need a short note in the changes, because the public set is no longer mutable. Although it would be a downstream bug to actually change it.

uschindler:
Hi @sabi0,
what's going on with this? Should I merge this one after resolving the conflict? The changes entry needs to be moved to 9.11.

sabi0:
Hi @uschindler,
I was away from computer last week. I will take care of the conflicts and your remarks later today.
Sorry for the delay.

uschindler:
> Hi @uschindler, I was away from computer last week. I will take care of the conflicts and your remarks later today. Sorry for the delay.

There seems to be overlap with the latest PR regarding concurrent sets. Maybe revert those changes from this PR.

sabi0:
> There seems to be overlap with the latest PR regarding concurrent sets. Maybe revert those changes from this PR.

This was a good idea. Thank you.
Now there is a clean separation between "immutable" (this PR) and "concurrent" sets (#13142).

lucene/core/src/java/org/apache/lucene/util/VirtualMethod.java
sabi0: Reverted

lucene/CHANGES.txt
sabi0: Done

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/BackwardsCompatibilityTestBase.java
sabi0: `TestAncientIndicesCompatibility.UNSUPPORTED_INDEXES` is already immutable:
https://github.com/apache/lucene/blob/390c109e673c9164c5359eda34223688875e4b66/lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestAncientIndicesCompatibility.java#L84

Or did you mean something else?

lucene/core/src/java/org/apache/lucene/store/NativeFSLockFactory.java
sabi0: I will open a separate PR for this.

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/BackwardsCompatibilityTestBase.java
uschindler: ok all fine.

uschindler (approved):

uschindler:
Thanks!

dweiss (commented):
Thanks!

PR #13090, opened 2024-02-08
Prevent humongous allocations when calculating scalar quantiles

benwtrent:
The initial release of scalar quantization would periodically create a humongous allocation, which can put unwarranted pressure on the GC & on the heap usage as a whole.

This commit adjusts this by only allocating a float array of 20*dimensions and averaging the discovered quantiles from there. 

Why does this work?

 - Quantiles based on confidence intervals are (generally) unbiased and doing an average gives statistically good results
 - The selector algorithm scales linearly, so the cost is just about the same
 - We need to do more than `1` vector at a time to prevent extreme confidence intervals interacting strangely with edge cases

I benchmarked this over 500k vectors.

candidate
```
Force merge done in: 691533 ms
0.817	 0.04	500000	0	16	250	2343	596410	1.00	post-filter
```
baseline
```
Force merge done in: 685618 ms
0.818	 0.04	500000	0	16	250	2346	582242	1.00	post-filter
```

100k vectors
candidate
```
0.855	 0.03	100000	0	16	250	2207	144173	1.00	post-filter
```

baseline
```
0.858	 0.03	100000	0	16	250	2205	141578	1.00	post-filter
```

There does seem to be a slight increase in merge time (these are single threaded numbers) and a slight change in recall. 

But to me, these seem acceptable given we are no longer allocating a ginormous array.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/ScalarQuantizer.java
+++ b/lucene/core/src/java/org/apache/lucene/util/ScalarQuantizer.java
@@ -66,6 +66,7 @@ import org.apache.lucene.index.VectorSimilarityFunction;
 public class ScalarQuantizer {
 
   public static final int SCALAR_QUANTIZATION_SAMPLE_SIZE = 25_000;
+  private static final int SCRATCH_SIZE = 20;
 
   private final float alpha;
   private final float scale;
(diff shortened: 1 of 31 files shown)


lucene/core/src/java/org/apache/lucene/util/ScalarQuantizer.java
```diff
@@ -281,24 +270,57 @@ static ScalarQuantizer fromVectors(
...
+      int i = 0;
       while (floatVectorValues.nextDoc() != NO_MORE_DOCS) {
-        float[] floatVector = floatVectorValues.vectorValue();
-        System.arraycopy(floatVector, 0, values, copyOffset, floatVector.length);
-        copyOffset += dim;
+        float[] vectorValue = floatVectorValues.vectorValue();
+        System.arraycopy(
+            vectorValue, 0, quantileGatheringScratch, i * vectorValue.length, vectorValue.length);
+        i++;
+        if (i == scratchSize) {
```
mayya-sharipova, line 286: Are we are ok to discard the last left-over (< 20) vectors for calculations?

mayya-sharipova (approved):
@benwtrent Thanks, the Lucene implementation LGTM as long as we are ok with the math and decreased recall that it brings.

lucene/core/src/java/org/apache/lucene/util/ScalarQuantizer.java
benwtrent: I think so. If we are sampling these few vectors already, I wouldn't want to be negatively impacted by extreme confidence intervals & quantiles. Meaning, extreme confidence intervals can add small biases when calculated over few points (this is why we do our best to calculate these over `scratchSize * dim`).

tveasey (approved):
LGTM

PR #13091, opened 2024-02-08
Moving quantization logic to make future quantizer work simpler

benwtrent:
While digging around and working on future scalar quantization work, I noticed that how these formats are is out of line with our other formats. Additionally, I am moving some things that were only into the codec into `util.quantization` to align with our hnsw stuff.


mayya-sharipova (approved):
@benwtrent Thanks, indeed looks better organization!

ChrisHegarty (commented):
LGTM

PR #13095, opened 2024-02-12
Fix addBackcompatIndexes.py to properly generate missing versions

### LINKED ISSUE #13094: Python backward compat scripts need updating
### Description

As @HoustonPutman wrote on the mailing list:

> So I just finished the 8.11.3 release, and had some problems with the back-compat testing code. Obviously the ant/gradle switch and jdk versions meant that I had to do a lot of it manually, but I'm pretty sure the regex matching in the update_backcompat_tests() function in addBackcompatIndexes.py was broken by this commit.

A review of these release scripts would be needed to make sure no regressions lurk after #13046 

_No response_

s1monw:
In #13046 several changes broke the addBackcompatIndexes.py script to properly add and test the unreleased version. This updates the script to again properly add the new version.

Closes #13094


s1monw:
this fixes the release script but I can't get tidy to not wrap the array [here](https://github.com/apache/lucene/pull/13095/files#diff-8946e557343c70aadb734c2f1bdceeed366f32c6f667ea75b89b89ae2895887fR50)  @uschindler or @rjernst can you help me to find out what's going on....

uschindler:
What's the exact problem with tidy? 
I think after adding the new version, we should run tidy to let it cleanup the code.

I will check if there are some comment signals to tidy so it excludes some code from reformatting. Personally I'd like to do this sometimes because reformatting of code breaks readability.

uschindler:
I'd like for the reasons described before to enable this option:

https://stackoverflow.com/a/76221803

s1monw:
@uschindler this works. it's funky if one of the strings has a certain length it will allow the bock-like strucuture.

uschindler:
Great! I haven't seen we use `//spotless:off` already and its enabled in the config. So I think this is best option here. This also allows the regexes for the script to not break suddenly.

I don't know why the PR's precommit checks fail. It complains about missing classes....

uschindler (approved):

s1monw:
@uschindler i think it's because @jpountz added a new version on branch_9x I spoke to him already I think it breaks tests as well on the stable branch

uschindler:
> @uschindler i think it's because @jpountz added a new version on branch_9x I spoke to him already I think it breaks tests as well on the stable branch

Yeah, I also see this one on Jenkins:

```
C:\Users\jenkins\workspace\Lucene-MMAPv2-Windows\lucene\core\src\java\org\apache\lucene\util\Version.java:139: warning: [dep-ann] deprecated item is not annotated with @Deprecated
  @Deprecated public static final Version LUCENE_9_10_0 = new Version(9, 10, 0);
                                          ^
1 error
1 warning
```

It has a duplicate annotation:

https://github.com/apache/lucene/blob/314c553bdce22bc1bc08e46b218691575c2287e0/lucene/core/src/java/org/apache/lucene/util/Version.java#L133-L139

uschindler:
@jpountz fixed the duplicate `@Deprecated` but it still fails because it can't find v9.11.0. I have no idea how it handles the last version of a previous stable branch. For sure from main's point of view 9.11.0 was never released.

At least for this PR I think it looks fine. The spotless trick solves the issue. +1

jpountz:
@uschindler I have been blindly following the release wizard, I'll look into the failure in a bit and see how we can address it.

jpountz (approved):

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/BackwardsCompatibilityTestBase.java
```diff
@@ -47,25 +47,64 @@
 
 public abstract class BackwardsCompatibilityTestBase extends LuceneTestCase {
 
-  protected final Version version;
+  // spotless:off
```
dweiss, line 50: I don't think this is such a great idea... in general, google java format does not have any way to prevent formatting - it's all or nothing (this is a deliberate decision).

I think a better way to fix this would be to:
* read this from a resource, which can be a plain text file formatted in any way you like,
* use // comments on each line to force line breaks for google java formatter (ugly but will work).

dweiss: thinking about it, having these versions in separate resources could also make it easier to manipulate from scripts?

uschindler: I agree, maybe 2 resource files would be better.

To your first comment: there is `spotless:off` and `spotless:on` and we use it already in another source file. Why is it all or nothing although it is documented in a different way? It this because of a Google Java Format special case?

dweiss: When you use google java formatter from intellij, for example, it will ignore spotless markup and reformat the source anyway.

Any such remaining markup is a mistake. It will keep piling up - it's almost never a good idea to have it. 

I think those versions should be kept in plain text resources - then updating them will be easier as well?

dweiss: In case I'm not clear - GJF (google java formatter) is "used by" spotless but you can also use it independently. As for exclusion markup - here is the faq entry from GJF concerning on/off markup (there are also discussions on their issue tracker).

https://github.com/google/google-java-format/wiki/FAQ#why-cant-i-use-magic-comments-to-disable-formatting-for-certain-sections-of-code

uschindler: It looks like `spotless:off` works fine.

I tried it out in https://github.com/apache/lucene/blob/main/lucene/queries/src/test/org/apache/lucene/queries/payloads/TestPayloadCheckQuery.java and it worked as expected:
- I changed indent and code outside the block: formatting was restored
- I changed indent in the code of those array sections: formatting was not updated by formatter

So it looks like GJF does not support the tag, but spotless does.

uschindler: I understand you comment about IntelliJ, but why should I appliy the Google Java format there when gradle can do it?

uschindler: P.S.: In this text above the formatting seems to have a reason because the upper and lower line somehow aligned to each other (the lower line has 1000 added to all entries and the first one is missing). So it looks like this demonstrates some gap in the positions: https://github.com/apache/lucene/blob/027240f522e32001d2daf6a622aeb7fbe87f8869/lucene/queries/src/test/org/apache/lucene/queries/payloads/TestPayloadCheckQuery.java#L270-L287

uschindler: Anyways, I agree to have those 2 lists to be a resource file and loaded using `Class#getResourceAsStream()` and split by lines.

dweiss: > I understand you comment about IntelliJ, but why should I appliy the Google Java format there when gradle can do it?

Accidently. I format the code in intellij all the time - I may not see the formatter-exclusion tags and intellij would reformat that part - a subsequent gradle tidy run wouldn't touch this section but it would remain reformatted.

uschindler: ...then it's good that you have not touched that test file :-)

dweiss: I feel inclined to remove those existing two sections and add it to the rat pre-check, actually. I can think of better ways of conveying the meaning of those "aligned" source lines.

Should we move these constants to a resource file as part of this issue or a separate one?

uschindler: > I feel inclined to remove those existing two sections and add it to the rat pre-check, actually. 

It's enough to disable the setting in the gradle build: https://github.com/apache/lucene/blob/027240f522e32001d2daf6a622aeb7fbe87f8869/gradle/validation/spotless.gradle#L31

It was enabled in #2185

uschindler: > Should we move these constants to a resource file as part of this issue or a separate one?

I think this should be part of this PR. Especially as it requires changes in the scripts, too.

dweiss: Thanks, Uwe. I'll revert that option. It's a problem waiting to happen. I may help straightening this issue too but in the evening - on the road today.

s1monw:
@uschindler @dweiss I moved all version to property files and updated the python script

dweiss:
I pushed a simple patch replacing properties with txt files - no need to worry about property file syntax and easier to read too, I think.

I like how python script got simpler too.

dweiss (approved):

uschindler (approved):
Looks fine.

I only have one comment about the "unsupported_versions.txt" which are not versions but filenames.

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/BackwardsCompatibilityTestBase.java
```diff
@@ -48,25 +52,28 @@
...
-        };
-
+    String name = "versions.txt";
+    try (LineNumberReader in =
+        new LineNumberReader(
+            IOUtils.getDecodingReader(
+                IOUtils.requireResourceNonNull(
+                    TestAncientIndicesCompatibility.class.getResourceAsStream(name), name),
+                StandardCharsets.UTF_8))) {
+      OLD_VERSIONS = in.lines().collect(Collectors.toCollection(LinkedHashSet::new));
```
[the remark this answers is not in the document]
uschindler: One comment: Should we not have a `filter(Predicate.not(String::isBlank)`?

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/BackwardsCompatibilityTestBase.java
```diff
@@ -50,7 +52,7 @@
 
 public abstract class BackwardsCompatibilityTestBase extends LuceneTestCase {
 
-  static final String[] OLD_VERSIONS;
+  static final Set<String> OLD_VERSIONS;
```
uschindler, line 55: I think the String sets can be private, as we only want to expose the Version numbers, correct?

s1monw:
@uschindler I did separate out the specific indices and added a new file. that makes maintenance simpler and moves the `-cfs` `-nocfs` duplication to the code. Also applied you comments.

uschindler (approved):
Looks fine. Thanks.

P.S.: I don't know if we still need those 4 special case indexes. They seem to be a relic from earlier times. As we only check that the indexes throw an Exception on loading, the already existing 5.x and 4.x indexes are enough.

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/BackwardsCompatibilityTestBase.java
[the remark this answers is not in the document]
dweiss: I see Simon already applied that (thank you!).

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/BackwardsCompatibilityTestBase.java
```diff
@@ -48,25 +53,31 @@
...
-          "8.11.2", "8.11.3", "8.11.3", "8.12.0", "9.0.0", "9.1.0", "9.2.0", "9.3.0", "9.4.0",
-          "9.4.1", "9.4.2", "9.5.0", "9.6.0", "9.7.0", "9.8.0", "9.9.0", "9.9.1", "9.9.2"
-        };
-
+    String name = "versions.txt";
+    try (LineNumberReader in =
+        new LineNumberReader(
+            IOUtils.getDecodingReader(
+                IOUtils.requireResourceNonNull(
+                    TestAncientIndicesCompatibility.class.getResourceAsStream(name), name),
```
uschindler, line 70: Small nit: Copy-paste error, wrong class name.

dweiss: Darn.

dweiss: Fixed.

s1monw:
> P.S.: I don't know if we still need those 4 special case indexes. They seem to be a relic from earlier times. As we only check that the indexes throw an Exception on loading, the already existing 5.x and 4.x indexes are enough.

true, I will merge this one and open a new pr to remove these

uschindler:
This was merged to 9.x only. Should we not have it in main, too?

s1monw:
@uschindler dude I need to press buttons it took me 14 mins to port I guess that's ok?

uschindler:
Sorry, I was just irritated, because the original PR was opened against stable branch. 😜

PR #13097, opened 2024-02-12
Change the links to Lucene's homepage in the DOAP file to https://

uschindler:
This PR changes all references to the Lucene homepage from the DOAP file to use the canonic HTTPS link.

P.S.: This cannot change the namespace links as they are identifiers in XML world. Also the official Apache License URL is without https at moment.


rmuir (approved):

uschindler:
I will commit this only to main branch, as the file is only read from there. Or I am missing something (does it need to be backported for another reason?)

PR #13098, opened 2024-02-13
Revert the addition of spotless:on/off regions

dweiss:
This disables spotless exclusion zones (on/off comments). These comments are only honored by spotless, not by google java format so accidental reformatting is still possible and you can't recover from that. 

This patch also shows one possible alternative that is formatter fool-proof - if you really need to show some alignment in test data.

### DIFF
--- a/gradle/validation/spotless.gradle
+++ b/gradle/validation/spotless.gradle
@@ -28,7 +28,6 @@ configure(project(":lucene").subprojects) { prj ->
 
     spotless {
       java {
-        toggleOffOn() // obviously, only to be used sparingly.
         // TODO: Work out how to support multiple different header files (we have
         // classes in the codebase that have original headers). We currently use
         // Apache RAT to enforce headers so this is of lesser priority.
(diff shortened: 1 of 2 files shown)


rmuir (approved):

uschindler (commented):
This looks fine to me, although I was expecting a solution without this extra method.

Downside: Won't work in 9.x (Java 11 has no multiline strings), but I think we can leave the old state there.

I don't give an explicit plus or minus. To me it is one solution.

dweiss:
I could try to  play with spaces, comments, etc. to align those two lines but in the end it's just a hack - looking for cracks in gjf. What I suggested as an alternative is guaranteed to work and fulfills the intention of what that code's author had in mind (hopefully).

No magic wand exists here, I'm afraid. It's one of the tradeoffs of using gjf - the code will be shaped this way or another - if you need precise alignment or formatting, use data that is external to the source code or data types that have a defined layout (like here). 

I still think overall it's well worth it. Consistent and most of the time better than what I'd do (especially with more complex nestings or logical conditions).

uschindler:
Yeah, let's apply this for main branch. branch_9x is Java 11 only and I think we can keep it as is, it won't survive for too long anymore (I hope).

PR #13108, opened 2024-02-15
Allow multiple JDKs in smoke test

### LINKED ISSUE #13107: Improve smoketester JDK versions support
### Description

Currently it is difficult for smoketester to "keep up" with supported java versions, which are moving faster these days.

@uschindler ran into it with the release vote for 9.10, and I ran into it on #12753

From Uwe:

> JAVA_HOME must run be Java 11 (in 9.x)
At moment you can pass "--test-java17 <dir>", but this one is also checked to be really java 17 (by parsing strings from its version output), but I'd like to pass "--test-alternative-java <dir>" multiple times and it would just run all those as part of smoking, maxbe the version number can be extracted to be printed out.

So I think the smoketester should require a specific version and check it, for JAVA_HOME. This should be the minimal required java version. But we should also allow passing other JDKs too.

This would make it easier to maintain the logic going forwards.

stefanvodita:
Introduce `--test-alternative-java` to`smokeTestRelease.py`, allowing the tests to run on multiple JDKs.

### DIFF
--- a/dev-tools/scripts/smokeTestRelease.py
+++ b/dev-tools/scripts/smokeTestRelease.py
@@ -639,8 +639,6 @@ def verifyUnpacked(java, artifact, unpackPath, gitRevision, version, testArgs):
     testDemo(java.run_java, isSrc, version, BASE_JAVA_VERSION)
     if java.run_alt_javas:
       for run_alt_java, alt_java_version in zip(java.run_alt_javas, java.alt_java_versions):
-        print("DELETEME")
-        print("    run test demo w/ Java %s..." % alt_java_version)
         testDemo(run_alt_java, isSrc, version, alt_java_version)
 
   testChangesText('.', version)
@@ -929,9 +927,13 @@ def make_java_config(parser, alt_java_homes):
     actual_version = re.search(r'version "([1-9][0-9]*)', s).group(1)
     print('Java %s JAVA_HOME=%s' % (actual_version, java_home))
 
-    if (is_base_version and BASE_JAVA_VERSION != actual_version) \
-    or int(BASE_JAVA_VERSION) > int(actual_version):
-      parser.error('got wrong version for java %s:\n%s' % (BASE_JAVA_VERSION, s))
+    # validate Java version
+    if is_base_version:
+      if BASE_JAVA_VERSION != actual_version:
+        parser.error('got wrong base version for java %s:\n%s' % (BASE_JAVA_VERSION, s))
+    else:
+      if int(actual_version) < int(BASE_JAVA_VERSION):
+        parser.error('got wrong version for java %s, less than base version %s:\n%s' % (actual_version, BASE_JAVA_VERSION, s))
 
     def run_java(cmd, logfile):
       run('%s; %s' % (cmd_prefix, cmd), logfile)
--- a/dev-tools/scripts/smokeTestRelease.py
+++ b/dev-tools/scripts/smokeTestRelease.py
@@ -39,6 +39,8 @@ import zipfile
 from collections import namedtuple
 import scriptutil
 
+BASE_JAVA_VERSION = "11"
+
 # This tool expects to find /lucene off the base URL.  You
 # must have a working gpg, tar, unzip in your path.  This has been
 # tested on Linux and on Cygwin under Windows 7.
@@ -144,10 +146,10 @@ def checkJARMetaData(desc, jarFile, gitRevision, version):
       'Implementation-Vendor: The Apache Software Foundation',
       'Specification-Title: Lucene Search Engine:',
       'Implementation-Title: org.apache.lucene',
-      'X-Compile-Source-JDK: 11',
-      'X-Compile-Target-JDK: 11',
+      'X-Compile-Source-JDK: %s' % BASE_JAVA_VERSION,
+      'X-Compile-Target-JDK: %s' % BASE_JAVA_VERSION,
       'Specification-Version: %s' % version,
-      'X-Build-JDK: 11.',
+      'X-Build-JDK: %s.' % BASE_JAVA_VERSION,
       'Extension-Name: org.apache.lucene'):
       if type(verify) is not tuple:
         verify = (verify,)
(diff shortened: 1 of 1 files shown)


uschindler:
Can we remove the java 17 one? Or at least map the java 17 command line parameter as additional alternative.

stefanvodita:
We can, that would be cleaner. But this should work for testing 9.10, right?
Marking as draft, since I don’t think we would merge as-is.

uschindler:
It's already in testing: https://jenkins.thetaphi.de/job/Lucene-Release-Tester-v2/1/console

uschindler:
waaah it doesnt like the version :-(

Can you comment out the version check in your branch?

uschindler:
Thanks, will retrigger the job!

uschindler:
Runs: https://jenkins.thetaphi.de/job/Lucene-Release-Tester-v2/2/console

uschindler:
Smoke tester succeeded (see mailing list).

stefanvodita:
That's great! If we extract the JDK version number as well, maybe we can commit this.
@uschindler - would you want the parameter for a specific JDK version removed on both `main` and `branch_9x` i.e. remove `--test-java19` from `main` and remove `--test-java17` from `branch_9x`?

uschindler:
> That's great! If we extract the JDK version number as well, maybe we can commit this.

Extracting the JDK version number is already implemented, because there's a check for Java 11 and Java 17 already.

> @uschindler - would you want the parameter for a specific JDK version removed on both `main` and `branch_9x` i.e. remove `--test-java19` from `main` and remove `--test-java17` from `branch_9x`?

I would remove all of them. Only the base version should be checked (so in 9.x base version must be 11). The other versions passed as cmd line parameter should maybe just checked to be >= base version.

If you want backwards compatibility, you can keep the old cmd line parameters and just add the arguments from the specific parameters to the list of alternative versions. But as this is a developer tools only, I don't think we need any backwards compatibility.

ChrisHegarty:
> > That's great! If we extract the JDK version number as well, maybe we can commit this.
> 
> Extracting the JDK version number is already implemented, because there's a check for Java 11 and Java 17 already.
> 
> > @uschindler - would you want the parameter for a specific JDK version removed on both `main` and `branch_9x` i.e. remove `--test-java19` from `main` and remove `--test-java17` from `branch_9x`?

+1

> I would remove all of them. Only the base version should be checked (so in 9.x base version must be 11). The other versions passed as cmd line parameter should maybe just checked to be >= base version.

+1 

What's the latest status of this PR? It looks like it is in reasonable shape and just needs to address some comments, right?

stefanvodita:
@uschindler - I addressed your first three points. On the last two points I think I'm missing some of the context to understand the issues. I'll try to dig a bit to understand those better. Is this a "progress not perfection" situation, where we can treat the last 2 points in a separate PR or are they tightly bound with the rest of this work?

As far as testing, it's a bit annoying to do, but I think I've got the script working correctly. I used the latest [available RC](https://dist.apache.org/repos/dist/dev/lucene/) (9.9.0):
```
JAVA_HOME=/usr/lib/jvm/java-11-amazon-corretto python3 -u dev-tools/scripts/smokeTestRelease.py --test-alternative-java /usr/lib/jvm/java-20-amazon-corretto --test-alternative-java /usr/lib/jvm/java-21-amazon-corretto \
https://dist.apache.org/repos/dist/dev/lucene/lucene-9.9.0-RC1-rev-92a5e5b02e0e083126c4122f2b7a02426c21a037

```
There was an error about 8.11.3 back compatibility not being tested, but I think that has to do with it being released after 9.9.0, and not with the changes to the script.

dev-tools/scripts/smokeTestRelease.py
```diff
@@ -633,9 +636,12 @@ def verifyUnpacked(java, artifact, unpackPath, gitRevision, version, testArgs):
 
     checkAllJARs(os.getcwd(), gitRevision, version)
 
-    testDemo(java.run_java11, isSrc, version, '11')
-    if java.run_java17:
-      testDemo(java.run_java17, isSrc, version, '17')
+    testDemo(java.run_java, isSrc, version, BASE_JAVA_VERSION)
+    if java.run_alt_javas:
+      for run_alt_java, alt_java_version in zip(java.run_alt_javas, java.alt_java_versions):
+        print("DELETEME")
```
uschindler, line 642: Yes, DELETE ME ! :-)

dev-tools/scripts/smokeTestRelease.py
```diff
@@ -911,33 +917,45 @@ def crawl(downloadedFiles, urlString, targetDir, exclusions=set()):
...
     s = subprocess.check_output('%s; java -version' % cmd_prefix,
                                 shell=True, stderr=subprocess.STDOUT).decode('utf-8')
-    if s.find(' version "%s' % version) == -1:
-      parser.error('got wrong version for java %s:\n%s' % (version, s))
+
+    actual_version = re.search(r'version "([1-9][0-9]*)', s).group(1)
+    print('Java %s JAVA_HOME=%s' % (actual_version, java_home))
+
+    if (is_base_version and BASE_JAVA_VERSION != actual_version) \
+    or int(BASE_JAVA_VERSION) > int(actual_version):
```
uschindler, line 933: I think the messages are a bit misleading depending on context. Maybe differentiate between `is_base_version` and the alternate version mode:
- if `is_base_version == true` then compare with exact same version and complain about that
- if `is_base_version == false` print a message about `actual_version`need to be `>= BASE_JAVA_VERSION`

The `>=` is important, because I want to test OpenJ9 vs Hotspot.

uschindler (commented):
I just found a minor problem and suggest to change the messages a bit. Otherwise looks fine +1

We can open another issue about the problem with Java 22+ (Gradle does not start with Java 22, so you can't test it). What you do is the following (Jenkins does this for example): `JAVA_HOME` always points to base version, but `RUNTIME_JAVA_HOME` is used to instruct Gradle to run all compilation and tests with that JDK.

I agree, this can be a separate issue. The workaround at moment is to run the smoketester with `RUNTIME_JAVA_HOME` env var :-)

uschindler:
> There was an error about 8.11.3 back compatibility not being tested, but I think that has to do with it being released after 9.9.0, and not with the changes to the script.

If you run with Lucene 9.10.0 (just released last week), it should work.

dev-tools/scripts/smokeTestRelease.py
stefanvodita: Thanks for noticing, I forgot about this one.

dev-tools/scripts/smokeTestRelease.py
stefanvodita: Sure, I'll do separate messages for the two cases.

stefanvodita:
> If you run with Lucene 9.10.0 (just released last week), it should work.

I think so, but the 9.10.0 RC artefacts were no longer available at https://dist.apache.org/repos/dist/dev/lucene/. Not sure why some versions are still there and some aren't. It was easier just to use the 9.9.0 RC that is still there.

uschindler (approved):

dweiss:
Just FYI - I've returned to trying to add a nightly smoketester workflow check. The main branch required some changes to the script as well, which sort of proves Rob's point that we need to run it regularly, otherwise it goes out of sync with the code quickly...

I'll try to forward-port @stefanvodita 's changes to main and will create a pr when I think it presents a reasonable state (still working my way through some python checks).

uschindler:
I checked the script. Forward port should be more or less:
- copy file to main branch
- change BASE_VERSION constant in the script

The new script is no longer a search replace Desaster.

dweiss:
I don't think it's going to be so easy - some things just didn't work for me with local artifacts. I'll follow up, perhaps tomorrow.

dweiss (approved):

dweiss:
I think we can merge this?

uschindler:
Will you forward port the patch to main branch? I can help with that, have some freetime.

This will allow us to make everything Java 21 ready tomorrow, so we have one file less to search/replace version numbers.

stefanvodita:
Here it is: #13139.
I wasn't sure about the difficulties @dweiss was encountering - I expect this to just work.

uschindler:
> Here it is: #13139. I wasn't sure about the difficulties @dweiss was encountering - I expect this to just work.

I think Dawid's problems were preexisting, not related to this PR and porting it forward.

uschindler:
The other PR looks straight-forward to me. Have you compared the two files directly after merging? They should be mostly identical and only the BASE_VERSION line should be different.

stefanvodita:
I have. The diff looks correct.

PR #13109, opened 2024-02-15
Add WrappedCandidateMatcher for composing matchers

bjacobowitz:
### Description

Create a `WrappedCandidateMatcher`, along with a factory for creating these objects and a default implementation of the factory, to assist users in creating their own composite matchers.

### Problem

The current access protections on `CandidateMatcher` make it infeasible for a user of the Lucene Monitor library to create their own composite `CandidateMatcher` (for example, an alternate version of `ParallelMatcher`) on top of existing `CandidateMatcher` instances if they are outside the package.

As an example, the existing implementation of `ParallelMatcher` makes use of delegate `CandidateMatcher` instances, but a user would not be able to write their own version of this in their own package, because `matchQuery()`, `finish()` and `reportError()` all have package-private access ([see this gist](https://gist.github.com/bjacobowitz/be9a54663217c46cc239cc9f9f580de8#file-parallelmatcher-java), comments labeled "PROBLEM").

### Solution

The simplest solution to this problem would be to increase the visibility of those functions in `CandidateMatcher` (either to `protected` or `public`), but it runs the risk of violating encapsulation, so I've tried to avoid that here.

The solution I've put together here introduces a `WrappedCandidateMatcher`, which can be used to increase visibility when creating a composite matcher. `WrappedCandidateMatcher` does **NOT** extend from `CandidateMatcher` but has nearly the same interface, with increased visibility on `matchQuery()`, `finish()` and `reportError()`. A user building a new `CandidateMatcher` by composing existing `CandidateMatcher` instances could make full use of those `CandidateMatchers` by wrapping them with `WrappedCandidateMatcher`, allowing them to be used internally to the composite matcher ([see this gist](https://gist.github.com/bjacobowitz/be9a54663217c46cc239cc9f9f580de8#file-parallelmatcherusingwrapped-java), comments labeled "SOLUTION").

On a certain level, this solution also violates existing encapsulation, but it requires the user to jump through a few hoops to get there, so it would limit the potential for misuse of a `CandidateMatcher`. If we would instead prefer to increase the visibility on `CandidateMatcher`, let me know and I can do that instead.

### Merge Request

If this PR gets merged, can you please use my `bjacobowitz1@bloomberg.net` email address for the squash+merge. Thank you.

### DIFF
--- a/lucene/monitor/src/java/org/apache/lucene/monitor/QueryMatch.java
+++ b/lucene/monitor/src/java/org/apache/lucene/monitor/QueryMatch.java
@@ -48,6 +48,21 @@ public class QueryMatch {
             }
           };
 
+  public static final WrappedMatcherFactory<QueryMatch> SIMPLE_WRAPPED_MATCHER_FACTORY =
+      searcher ->
+          new WrappedCandidateMatcher<QueryMatch>(
+              new CollectingMatcher<QueryMatch>(searcher, ScoreMode.COMPLETE_NO_SCORES) {
+                @Override
+                public QueryMatch resolve(QueryMatch match1, QueryMatch match2) {
+                  return match1;
+                }
+
+                @Override
+                protected QueryMatch doMatch(String queryId, int doc, Scorable scorer) {
+                  return new QueryMatch(queryId);
+                }
+              });
+
   /**
    * Creates a new QueryMatch for a specific query and document
    *
--- a/lucene/monitor/src/java/org/apache/lucene/monitor/WrappedCandidateMatcher.java
+++ b/lucene/monitor/src/java/org/apache/lucene/monitor/WrappedCandidateMatcher.java
@@ -22,45 +22,92 @@ import java.util.List;
 import java.util.Map;
 import org.apache.lucene.search.Query;
 
-// Wrapper class for an existing candidate matcher exposing finish, matchQuery and reportError
-// for composition of matchers
+/**
+ * Wrapper class for an existing candidate matcher exposing finish, matchQuery and reportError for
+ * composition of matchers
+ */
 public class WrappedCandidateMatcher<T extends QueryMatch> {
 
   private CandidateMatcher<T> delegateTo;
 
+  /**
+   * Creates a new WrappedCandidateMatcher delegating to the supplied CandidateMatcher
+   *
+   * @param wrapThis the CandidateMatcher to which functions should be delegate
+   */
   public WrappedCandidateMatcher(CandidateMatcher<T> wrapThis) {
     delegateTo = wrapThis;
   }
 
+  /**
+   * Runs the supplied query against the delegate CandidateMatcher's set of documents, storing any
+   * resulting match, and recording the query in the presearcher hits
+   *
+   * @param queryId the query id
+   * @param matchQuery the query to run
+   * @param metadata the query metadata
+   * @throws IOException on IO errors
+   */
   public void matchQuery(String queryId, Query matchQuery, Map<String, String> metadata)
       throws IOException {
     delegateTo.matchQuery(queryId, matchQuery, metadata);
   }
 
+  /**
+   * @return the matches from this matcher
+   */
   public MultiMatchingQueries<T> finish(long buildTime, int queryCount) {
     return delegateTo.finish(buildTime, queryCount);
   }
 
+  /**
+   * Record a match into delegate
+   *
+   * @param match a QueryMatch object
+   */
   protected void addMatch(T match, int doc) {
     delegateTo.addMatch(match, doc);
   }
 
+  /**
+   * If two matches from the same query are found (for example, two branches of a disjunction),
+   * combine them using the delegate's resolve function.
+   *
+   * @param match1 the first match found
+   * @param match2 the second match found
+   * @return a Match object that combines the two
+   */
   public T resolve(T match1, T match2) {
     return delegateTo.resolve(match1, match2);
   }
 
+  /** If running a query throws an Exception, this function will add error into delegate */
   public void reportError(String queryId, Exception e) {
     delegateTo.reportError(queryId, e);
   }
 
+  /** Called when all monitoring of a batch of documents is complete */
   protected void doFinish() {
     delegateTo.doFinish();
   }
 
+  /** Copy all matches from another CandidateMatcher */
   protected void copyMatches(CandidateMatcher<T> other) {
     delegateTo.copyMatches(other);
   }
 
+  /**
+   * Build a MultiMatchingQueries
+   *
+   * @param matches the matches to include, mapping from queryId to match
+   * @param errors any errors thrown while evaluating matches
+   * @param queryBuildTime how long (in ns) it took to build the Presearcher query for the matcher
+   *     run
+   * @param searchTime how long (in ms) it took to run the selected queries
+   * @param queriesRun the number of queries passed to this CandidateMatcher during the matcher run
+   * @param batchSize the number of documents in the batch
+   * @return a MultiMatchingQueries object with the results of matching a batch of Documents
+   */
   public static <T1 extends QueryMatch> MultiMatchingQueries<T1> buildMultiMatchingQueries(
       List<Map<String, T1>> matches,
       Map<String, Exception> errors,
--- a/lucene/monitor/src/java/org/apache/lucene/monitor/WrappedCandidateMatcher.java
+++ b/lucene/monitor/src/java/org/apache/lucene/monitor/WrappedCandidateMatcher.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.lucene.monitor;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import org.apache.lucene.search.Query;
+
+// Wrapper class for an existing candidate matcher exposing finish, matchQuery and reportError
+// for composition of matchers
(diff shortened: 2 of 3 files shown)


cpoerschke:
Thanks @bjacobowitz for a very detailed pull request here!

I'm not really familiar with this area of the code but intuitively would be curious what the alternative (you mentioned) of increasing the visibility on `CandidateMatcher` would look like.

bjacobowitz:
> The `protected` visibility on `matchQuery()` should already be fine here - you can override or call protected methods from within subclasses. I think making `reportError()` and `finish()` `protected final` would be the neatest solution.

@romseygeek On further reflection, I think protected visibility may not be sufficient for a practical solution here, even when using subclasses. I'll again use `ParallelMatcher` as an example, since I think it's a fairly representative use case.

`ParallelMatcher` [takes in a `MatcherFactory`](https://github.com/apache/lucene/blob/75e1ebc4501733a6a9857858b3fd9c30304234d0/lucene/monitor/src/java/org/apache/lucene/monitor/ParallelMatcher.java#L66) to [create instances of `CandidateMatcher`](https://github.com/apache/lucene/blob/75e1ebc4501733a6a9857858b3fd9c30304234d0/lucene/monitor/src/java/org/apache/lucene/monitor/ParallelMatcher.java#L120), with the intent to [take advantage of the `matchQuery` implementation in the factory's generated instances of `CandidateMatcher`](https://github.com/apache/lucene/blob/75e1ebc4501733a6a9857858b3fd9c30304234d0/lucene/monitor/src/java/org/apache/lucene/monitor/ParallelMatcher.java#L129). If we try to reimplement `ParallelMatcher` outside the Lucene Monitor package, we can't call `matchQuery` on the factory-built `CandidateMatcher` from outside (i.e. on an instance other than `this`) because of `matchQuery`'s protected access -- the only reason the existing `ParallelMatcher` implementation can do this is because it is in the Lucene Monitor and therefore has package-local access to the protected functions on another object in the same package.

To access and invoke the implementation of matchQuery, we would need to create a subclass of the concrete `CandidateMatcher` implementation generated by the factory, which isn't directly possible, because the factory pattern creates a situation where the concrete class isn't known.

We could approach a solution using an alternate factory implementation that would return a subclass of a known concrete type with hidden functions exposed. For example, we could create a subclass of `CollectingMatcher` (let's call it `ExposedCollectingMatcher`) with a public function `exposedMatchQuery` which calls `this.matchQuery()`, and then we could create a factory to return that subclass (it might be called `ExposedCollectingMatcherFactory`).

The problem with that approach is that it breaks down in cases where the concrete implementation is not available to be subclassed, as is the case for several anonymous `CandidateMatcher` implementations (e.g. instances generated by [`MATCHER` factory in `ExplainingMatch`](https://github.com/apache/lucene/blob/75e1ebc4501733a6a9857858b3fd9c30304234d0/lucene/monitor/src/java/org/apache/lucene/monitor/ExplainingMatch.java#L32-L54)).

Ideally we would like a factory that returns a more generic class that extends the functionality of a `CandidateMatcher`, which is what `WrappedCandidateMatcher` provides.

Of course another solution would be to simply make these functions public, but I understand the desire to avoid that if possible.

cpoerschke:
> ... Of course another solution would be to simply make these functions public, but I understand the desire to avoid that if possible.

Again, I'm not really familiar with this area of the code but would be curious what making the functions public would look like and if making them public is necessarily problematic.

And echo-ing @mkhludnev's https://github.com/apache/solr/pull/2315#issuecomment-1968545878 comment w.r.t. just-compile test to retain extensibility in future.

bjacobowitz:
@romseygeek I'm wondering if maybe we should make those functions `protected final` as you suggest, but also make some of the `CandidateMatcher` implementations public.

Right now, outside of the `org.apache.lucene.monitor` package, there is no basic concrete implementation of CandidateMatcher that is publicly availble to be subclassed. Today we have the following CandidateMatcher implementations, and none of the basic ones are public:
- [output of factory ExplainingMatch::MATCHER](https://github.com/apache/lucene/blob/acbd7141404d503e7ca89bec520a175482d41bce/lucene/monitor/src/java/org/apache/lucene/monitor/ExplainingMatch.java#L30-L54) (anonymous `CandidateMatcher`, can't be subclassed)
- [output of factory HighlightsMatch::MATCHER](https://github.com/apache/lucene/blob/acbd7141404d503e7ca89bec520a175482d41bce/lucene/monitor/src/java/org/apache/lucene/monitor/HighlightsMatch.java#L44-L96) (anonymous `CandidateMatcher`, can't be subclassed)
- [output of factory QueryMatch::SIMPLE_MATCHER](https://github.com/apache/lucene/blob/acbd7141404d503e7ca89bec520a175482d41bce/lucene/monitor/src/java/org/apache/lucene/monitor/QueryMatch.java#L37-L49) (anonymous `CollectingMatcher`, can't be subclassed)
- [output of factory ScoringMatch::matchWithSimilarity](https://github.com/apache/lucene/blob/acbd7141404d503e7ca89bec520a175482d41bce/lucene/monitor/src/java/org/apache/lucene/monitor/ScoringMatch.java#L29-L50) (anonymous `CollectingMatcher`, can't be subclassed)
- [output of factory QueryTimeListener::timingMatcher](https://github.com/apache/lucene/blob/acbd7141404d503e7ca89bec520a175482d41bce/lucene/monitor/src/java/org/apache/lucene/monitor/QueryTimeListener.java#L36-L61) (anonymous `CandidateMatcher`, can't be subclassed, plus the factory is package-private)
- [CollectingMatcher](https://github.com/apache/lucene/blob/acbd7141404d503e7ca89bec520a175482d41bce/lucene/monitor/src/java/org/apache/lucene/monitor/CollectingMatcher.java#L28) (package-private and abstract, not available to be subclassed)
- [ParallelMatcher](https://github.com/apache/lucene/blob/acbd7141404d503e7ca89bec520a175482d41bce/lucene/monitor/src/java/org/apache/lucene/monitor/ParallelMatcher.java#L47) (public but not basic, already composed of other `CandidateMatcher` instances via factory)
- [PartitionMatcher](https://github.com/apache/lucene/blob/acbd7141404d503e7ca89bec520a175482d41bce/lucene/monitor/src/java/org/apache/lucene/monitor/PartitionMatcher.java#L44) (public but not basic, but already composed of other `CandidateMatcher` instances via factory)

If we're thinking about making `CandidateMatcher` functions protected to give subclasses have access to them, I think we would need to make some of the basic implementations of `CandidateMatcher` public, so that they're available to be extended outside the `org.apache.lucene.monitor` package. 

Again I'll use `ParallelMatcher` as an example. First we [make the `reportError` and `finish` functions `protected final` in `CandidateMatcher`](https://gist.github.com/bjacobowitz/28f6b0b62b3a8828f54ec1d544563cfd#file-candidatematcher-java) as you suggested. If we then [expose a basic implementation of CandidateMatcher](https://gist.github.com/bjacobowitz/28f6b0b62b3a8828f54ec1d544563cfd#file-simplecandidatematcher-java) (I've extracted the CandidateMatcher returned by QueryMatch.SIMPLE_MATCHER, calling it SimpleCandidateMatcher), then outside the package we can [create a subclass of that](https://gist.github.com/bjacobowitz/28f6b0b62b3a8828f54ec1d544563cfd#file-candidatematcherextension-java) (I'm calling it CandidateMatcherExtension here) with public functions that call the now-protected functions of the superclass. That gives our `ParallelMatcher` implementation the ability to [make full use of the basic candidate matcher through the subclass](https://gist.github.com/bjacobowitz/28f6b0b62b3a8828f54ec1d544563cfd#file-parallelmatcher-java) (see the comments labeled "SOLUTION").

I think this would solve this issue without requiring us to dramatically change function permissions. The only additional requirement would be to expose some implementing CandidateMatcher classes, but I think that might be okay (we've already exposed ParallelMatcher and PartitionMatcher). Should I go ahead with this approach?

bjacobowitz:
Sure. It's roughly the same approach as ParallelMatcher, but with a reactive setup instead of an ExecutorService.

`monitor.match(Document, MatcherFactory)` runs a bunch of queries with a `CandidateMatcher` from `MatcherFactory` and returns `MatchingQueries`, but it won't return results until all the queries have run. I am trying to create a custom `CandidateMatcher` (and a factory to wrap it) that delegates to an existing matcher to do the matching work, but puts its results into a Flux, so that I can start processing the individual results as soon as they are available.

It's something sort of like this:

```java
package com.my.custom.package;

public class MyReactiveMatcher<T extends QueryMatch> extends CandidateMatcher<T> {

  public static class MatchTask {
    final String queryId;
    final Query matchQuery;
    final Map<String, String> metadata;
    public MatchTask(String queryId, Query matchQuery, Map<String, String> metadata) {
      this.queryId = queryId;
      this.matchQuery = matchQuery;
      this.metadata = metadata;
    }
  }
  
  private final List<MatchTask> tasks = new ArrayList<>();
  
  // Factory of CandidateMatcher to which we will delegate matching
  private final MatcherFactory<T> factory;
  
  public MyReactiveMatcher(IndexSearcher searcher, MatcherFactory<T> factory) {
    super(searcher);
    this.factory = factory;
  }

  @Override
  public void matchQuery(String queryId, Query matchQuery, Map<String, String> metadata) {
    // Register a query to try to match later
    tasks.add(new MatchTask(queryId, matchQuery, metadata));
  }
  
  // Run matchQuery
  private static <T extends QueryMatch> MultiMatchingQueries<U> getResult(
      MatchTask task, CandidateMatcher<T> matcher) {
    try {
      // This call is not possible outside the lucene.monitor package:
      matcher.matchQuery(task.queryId, task.matchQuery, task.metadata);
    } catch (IOException e) {
      // This call is not possible outside the lucene.monitor package:
      matcher.reportError(task.queryId, e);
    }
    // This call is not possible outside the lucene.monitor package:
    return matcher.finish(0, 0);
  }

  @Override
  protected void doFinish() {
    // Kick off matching, with results coming back through the Flux as they are finished
    Flux.fromIterable(tasks).flatMap(t -> getResult(t, factory.createMatcher(searcher))).doOnNext(::aBunchOfStuff).subscribe();
  }
```

The `getResult` function is where the issue comes up. I can't really make use of the CandidateMatcher generated by the factory if I'm outside of the `lucene.monitor` package, due to the access protections in place.

To use your suggestion of a subclass, I could create a `MyReactiveMatcher` with a factory that returns my custom subclass of `CandidateMatcher` (then I could just cast to it and call extra its functions), but to be able to create that custom subclass in the first place I would need to extend a concrete implementation of `CandidateMatcher`, and there really aren't any available at the moment.

romseygeek:
Thanks, that's very helpful.  Given that composite matchers need to call those methods for CandidateMatcher to be useful, I think the simplest way forward here is to make `matchQuery()`, `reportError()` and `finish()` public.  Do you want to open another PR for that @bjacobowitz?

bjacobowitz:
Great, I'll open another PR shortly with those changes (`matchQuery()`, `reportError()` and `finish()` public), and will reference this one for context. Thanks!

bjacobowitz:
Closing, see PR #13632 with new approach

PR #13110, opened 2024-02-15
github workflow cleanup and minor improvements

dweiss:
This extracts a common build setup composite action and cleans up existing workflows a bit. A functional change is that gradle/actions/setup-gradle is used to set up gradle instead of a manually-crafted generic gh cache.


dweiss:
I'll commit this shortly. There may be follow-ups, I'll keep an eye on those.

PR #13111, opened 2024-02-15
Adding a getter to FeatureField to allow value access

benwtrent:
Getting the access of a FeatureField#value is useful for deduplicating. If you have a sparse vector model and you want to handle multiple inputs from them, you want flexibility in how you de-duplicate the feature dimensions.


jpountz (approved):
We accepted this kind of changes in the past for highlighting (hopefully this is no longer needed) or lucene/monitor, so no objections.

PR #13115, opened 2024-02-19
Short circuit queued flush check when flush on update is disabled

### LINKED ISSUE #13079: [Suggestion] Short circuit check for queued flushes in preUpdate() when checkPendingFlushOnUpdate is disabled
### Description

`numQueuedFlushes()` is a blocking function which gets called in `DocumentWriter#preUpdate()` to check if there are any queued flushes. If `checkPendingFlushOnUpdate` is disabled i.e. its false then the overall condition `flushControl.numQueuedFlushes() > 0 && config.checkPendingFlushOnUpdate` could be short circuited if we reverse the check to `config.checkPendingFlushOnUpdate && flushControl.numQueuedFlushes() > 0`. This will prevent a blocking call.
Code pointer: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java#L387

Currently while benchmarking I can see this particular issue significantly contributing to locking of write threads which is increasing indexing latency.

Screenshot 2024-02-06 at 12 58 51 PM

CaptainDredge:
### Description
Small change which reduces locking of write threads when flush on update is disabled by short circuiting a check
Resolves #13079

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocumentsWriter.java
@@ -384,7 +384,7 @@ final class DocumentsWriter implements Closeable, Accountable {
     ensureOpen();
     boolean hasEvents = false;
     while (flushControl.anyStalledThreads()
-        || (flushControl.numQueuedFlushes() > 0 && config.checkPendingFlushOnUpdate)) {
+        || (config.checkPendingFlushOnUpdate && flushControl.numQueuedFlushes() > 0)) {
       // Help out flushing any queued DWPTs so we can un-stall:
       // Try pickup pending threads here if possible
       // no need to loop over the next pending flushes... doFlush will take care of this


CaptainDredge:
@msfroh sorry to tag you directly but can you please take a look at this change?

mikemccand (approved):
Neat idea -- move the invocation of the `synchronized` `numQueuedFlushes` method to the end so that if the app has disabled `IWC.checkPendingFlushOnUpdate` (it defaults to `true`) then we avoid the `sync` cost.

The change is safe -- these two methods have no weird side effects.

Thanks @CaptainDredge -- I'll merge soon.

uschindler:
Thanks Mike!

jpountz:
FYI I played with this change and [IndexGeoNames](https://github.com/mikemccand/luceneutil/blob/master/src/main/perf/IndexGeoNames.java) which has been good at detecting contention in the past.

I first modified the benchmark to call `iwc.setCheckPendingFlushUpdate(false);` and saw little difference. Then I disabled the 3 text fields, which is where most indexing time goes, and could see a noticeable difference when indexing with 30 threads. Before this change it took ~6.5s to fill up IndexWriter's buffer. After this change it took only ~5.5s to fill up IndexWriter's buffer, a significant improvement!

PR #13119, opened 2024-02-20
FieldInfosFormat translation should be independent of VectorSimilartyFunction enum

ChrisHegarty:
This commit updates the FieldInfosFormat translation of vector similarity functions to be independent of the `VectorSimilartyFunction` enum.

The `VectorSimilartyFunction` enum lives outside of the codec format, and the format should not inadvertently depend upon the declaration order or values in VectorSimilartyFunction. The format should be in charge of the translation of similarity function to format ordinal (and visa versa). In reality, and for now, the translation remains the same as the declaration order, but this may not be the case in the future.

Note: did we introduce a potential index corruption issue when adding maximum inner product in 9.8.0? since the format was not updated when the enum value was added - the ordinal for maximum inner product is unknown to Lucene 9.7.0, which uses the same format.


ChrisHegarty:
This PR is a prerequisite for future work to make the similarity function symbolic and lookup-able, see https://github.com/apache/lucene/pull/13076#issuecomment-1930363479.

lucene/core/src/java/org/apache/lucene/codecs/lucene94/Lucene94FieldInfosFormat.java
```diff
@@ -111,6 +112,8 @@
  *         <li>0: EUCLIDEAN distance. ({@link VectorSimilarityFunction#EUCLIDEAN})
  *         <li>1: DOT_PRODUCT similarity. ({@link VectorSimilarityFunction#DOT_PRODUCT})
  *         <li>2: COSINE similarity. ({@link VectorSimilarityFunction#COSINE})
+ *         <li>3: MAXIMUM_INNER_PRODUCT similarity. ({@link
+ *             VectorSimilarityFunction#MAXIMUM_INNER_PRODUCT})
```
ChrisHegarty, line 116: I'm not sure whether or not to add this here, since maximum inner product was added in 9.8.0!

tteofili: probably it'd make sense to do it in a different PR, for easier tracking

tteofili (commented):
Index format wise, I think the index corruption can occur when reading a Lucene 9.8.0 index with Lucene 9.7.0, as the format would allow that, but I am not sure this is an expected scenario.

uschindler (commented):
Hi,
as stated in the other issue: I am not really happy to have that enum at all! The similarity/distance functions should be pluggable using `NamedSPILoader`. To implement that, the ordinals must removed in a new file format version and instead names be written using the Codec utility classes.

As a first step this PR is fine as it does not change file format and just decouples the ordinals from the enum. In future, when we have SPI, we can use the current code of the ordinals

In my opinion, the strings as lookup keys are not needed: Just define it as `List<VectorSimilarityFunction>` to get the link between them. At a later stage the backwards layer could then fallback to the list with SPI instances to lookup the legacy ordinals. The coec and the enum are still *enough* decoupled.

ChrisHegarty:
> Hi, as stated in the other issue: I am not really happy to have that enum at all! The similarity/distance functions should be pluggable using `NamedSPILoader`. To implement that, the ordinals must removed in a new file format version and instead names be written using the Codec utility classes.

Agreed on where we wanna get to. Just trying to get there incrementally, since format changes are quite noisy.

> As a first step this PR is fine as it does not change file format and just decouples the ordinals from the enum. In future, when we have SPI, we can use the current code of the ordinals

Exactly, this is just a first step. It (for the most part) encapsulates the translation in the format. When we add a new format and/or evolve VectorSimilarityFunction, this format should be largely immune to the change.

> In my opinion, the strings as lookup keys are not needed: Just define it as `List<VectorSimilarityFunction>` to get the link between them. At a later stage the backwards layer could then fallback to the list with SPI instances to lookup the legacy ordinals. The coec and the enum are still _enough_ decoupled.

Yeah, that's probably good enough for now. Updated.

ChrisHegarty:
> > Index format wise, I think the index corruption can occur when reading a Lucene 9.8.0 index with Lucene 9.7.0, as the format would allow that, but I am not sure this is an expected scenario.
> 
> This is perfectly fine.

Ok cool. I was worried for nothing then.

uschindler:
> > Hi, as stated in the other issue: I am not really happy to have that enum at all! The similarity/distance functions should be pluggable using `NamedSPILoader`. To implement that, the ordinals must removed in a new file format version and instead names be written using the Codec utility classes.
> 
> Agreed on where we wanna get to. Just trying to get there incrementally, since format changes are quite noisy.
> 
> > As a first step this PR is fine as it does not change file format and just decouples the ordinals from the enum. In future, when we have SPI, we can use the current code of the ordinals
> 
> Exactly, this is just a first step. It (for the most part) encapsulates the translation in the format. When we add a new format and/or evolve VectorSimilarityFunction, this format should be largely immune to the change.
> 
> > In my opinion, the strings as lookup keys are not needed: Just define it as `List<VectorSimilarityFunction>` to get the link between them. At a later stage the backwards layer could then fallback to the list with SPI instances to lookup the legacy ordinals. The coec and the enum are still _enough_ decoupled.
> 
> Yeah, that's probably good enough for now. Updated.

The comment is a bit outdated. I was thinking of making it even more vebose by having 2 maps for the lookup...

+1 looks fine.

benwtrent (approved):
Depending on enum order was always trappy. Thanks for decoupling!

uschindler (approved):
if you fix the comment and remove "names" from 

maybe be sure to explicitly say: add new ones always at end of list :-)

ChrisHegarty:
I see now that we have a similar dependency in `Lucene99HnswVectorsReader`. I'll update in a similar way.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java
```diff
@@ -171,15 +172,25 @@ private void validateFieldEntry(FieldInfo info, FieldEntry fieldEntry) {
     }
   }
 
+  // List of vector similarity functions. This list is defined here, in order
+  // to avoid an undesirable dependency on the declaration and order of values
+  // in VectorSimilarityFunction. The list values and order have been chosen to
+  // match that of VectorSimilarityFunction in, at least, Lucene 9.10. Values
+  // should only be appended to the end of the list.
+  public static final List<VectorSimilarityFunction> SIMILARITY_FUNCTIONS =
```
ChrisHegarty, line 180: There's a little duplication here, but we want the codecs to be largely independent.

benwtrent: Since (for some unknown reason to me), we already write out the values to the codec files & we don't want to break the current codec, this is fine with me. 

I think if/when we create a new HNSW/vector codecs in the future, we should consider not writing the danged similarity functions in two places.

benwtrent: Could you add/complete your comment indicating that the values here need to be in order and the SAME order as the field info ones for bwc?

benwtrent (approved):

ChrisHegarty:
Thanks for the reviews. All comments have been addressed.

msokolov:
>  OK, this is really weird to me. For some reason, we are writing the dimension & similarity into the vector metdata but that information is retained in the field info already.

@benwtrent honestly don't remember, but I do know that early on we tried things a few different ways. There was some discussion about whether the similarity function and dimensions should be in the codec vs in the field info. I suspect it evolved and we did not end up removing the redundant version?

benwtrent:
@msokolov thanks for clarifying. I just wanted to make sure there wasn't an important reason that I missed.

PR #13121, opened 2024-02-20
Add mult-leaf optimizations for diversify children collector

benwtrent:
This adds multi-leaf optimizations for diversified children collector. This means as children vectors are collected within a block join, we can share information between leaves to speed up vector search.

To make this happen, I refactored the multi-leaf collector slightly. Now, instead of inheriting from `TopKnnCollector`, we inject a inner collector.


mayya-sharipova (approved):
@benwtrent Thanks Ben, nice addition!

PR #13123, opened 2024-02-21
Change BP reordering logic to help support document blocks later on.

jpountz:
The current logic for reordering splits a slice of doc IDs into a left side and a right side, and for each document it computes the expected gain of moving to the other side. Then it swaps documents from both sides as long as the sum of the gain of moving the left doc to the right and the right doc to the left is positive.

This works well, but I would like to extend BP reordering to also work with blocks, and the swapping logic is challenging to modify as two parent documents may have different numbers of children.

One of the follow-up papers on BP suggested to use a different logic, where one would compute a bias for all documents that is negative when a document is attracted to the left and positive otherwise. Then we only have to partition doc IDs around the mid point, e.g. with quickselect.

A benefit of this change is that it will make it easier to generalize BP reordering to indexes that have blocks, e.g. by using a stable sort on biases.

### DIFF
--- a/lucene/misc/src/java/org/apache/lucene/misc/index/BPIndexReorderer.java
+++ b/lucene/misc/src/java/org/apache/lucene/misc/index/BPIndexReorderer.java
@@ -355,8 +355,7 @@ public final class BPIndexReorderer {
       // Computing biases is typically a bottleneck, because each iteration needs to iterate over
       // all postings to recompute biases, and the total number of postings is usually one order of
       // magnitude or more larger than the number of docs. So we try to parallelize it.
-      ComputeBiasTask biasTask =
-          new ComputeBiasTask(
+      new ComputeBiasTask(
               docIDs.ints,
               biases,
               docIDs.offset,
@@ -364,12 +363,8 @@ public final class BPIndexReorderer {
               leftDocFreqs,
               rightDocFreqs,
               threadLocal,
-              depth);
-      if (shouldFork(docIDs.length, docIDs.ints.length)) {
-        invokeAll(biasTask);
-      } else {
-        biasTask.compute();
-      }
+              depth)
+          .compute();
 
       float maxLeftBias = Float.NEGATIVE_INFINITY;
       for (int i = docIDs.offset; i < midPoint; ++i) {
--- a/lucene/misc/src/java/org/apache/lucene/misc/index/BPIndexReorderer.java
+++ b/lucene/misc/src/java/org/apache/lucene/misc/index/BPIndexReorderer.java
@@ -49,7 +49,7 @@ import org.apache.lucene.util.ArrayUtil;
 import org.apache.lucene.util.BytesRef;
 import org.apache.lucene.util.CloseableThreadLocal;
 import org.apache.lucene.util.IOUtils;
-import org.apache.lucene.util.IntroSorter;
+import org.apache.lucene.util.IntroSelector;
 import org.apache.lucene.util.IntsRef;
 import org.apache.lucene.util.packed.PackedInts;
 
@@ -251,17 +251,17 @@ public final class BPIndexReorderer {
   private class IndexReorderingTask extends BaseRecursiveAction {
 
     private final IntsRef docIDs;
-    private final float[] gains;
+    private final float[] biases;
     private final CloseableThreadLocal<PerThreadState> threadLocal;
 
     IndexReorderingTask(
         IntsRef docIDs,
-        float[] gains,
+        float[] biases,
         CloseableThreadLocal<PerThreadState> threadLocal,
         int depth) {
       super(depth);
       this.docIDs = docIDs;
-      this.gains = gains;
+      this.biases = biases;
       this.threadLocal = threadLocal;
     }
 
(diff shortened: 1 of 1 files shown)


lucene/misc/src/java/org/apache/lucene/misc/index/BPIndexReorderer.java
```diff
@@ -341,116 +344,94 @@ protected void compute() {
...
-          pivotGain = gains[i];
+          pivotDoc = docIDs.ints[i];
+          pivotBias = biases[i];
         }
 
         @Override
         protected int comparePivot(int j) {
-          // Compare in reverse order to get a descending sort
-          int cmp = Float.compare(gains[j], pivotGain);
+          int cmp = Float.compare(pivotBias, biases[j]);
```
rishabhmaurya, line 399: I see that you are terminating early before shuffling by checking ` if (gain <= iter)`; 
Should we also include `iter` into the equation here `compare(pivotBias - iter, biases[j])`?

rishabhmaurya: @jpountz To be precise, this is what I tried to avoid swaps as per the paper - 
```java
    int cmp = Float.compare(pivotBias-iter, biases[j]);
    if (cmp > 0) {
      return cmp;
    }
    cmp = Float.compare(pivotBias, biases[j]);
    if (cmp >= 0) {
      // Tie break on the doc ID to preserve doc ID ordering as much as possible
      cmp = pivotDoc - docIDs.ints[j];
    }
    return cmp;
```
It won't be a precise kth largest but will be close to it with tolerance factor governed by `iter`. 

I used [http_logs](https://github.com/opensearch-project/opensearch-benchmark-workloads/blob/main/http_logs) workload and I see similar results as described in the paper with respect to time taken to perform reordering.

For benchmark setup, I setup just a single shard and try to force merge into one segment after indexing and the corpus used is `logs-231998`. 

Here is the result - 

### With this change
```shell
Final index size: 686.3mb
curl localhost:9200/_cat/indices
green open logs-231998 T-HpeIWARuSZFGa74PWOig 1 0 11961342 0 686.3mb 686.3mb
```
Some additional metrics I'm publishing [here](https://github.com/rishabhmaurya/lucene/commit/fda8f19ac691af832e4bd7b4722324a4f14971bf#diff-43ddf233a75c4de53d99ec83e5b52a5eff0c39e8d601ae6816db49c7c14f785aR1154), you can see **number of swaps to number of docs ratio is 8**
```
Reorder stats
totalDocs: 11961342; numTerms: 964; totalPostings: 88726436; totalSwapsPerformed: 96188516; postingsDocRatio: 7; swapsDocsRatio: 8
12816206 9489609 12087839 12496439 9235544 9035264 8148933 6569703 5403994 3803447 2775037 1757719 1117468 668304 374775 216838 120914 70483 0 0 0 0 0 0 0
```
|                                                         Metric |                     Task |       Value |   Unit |
|---------------------------------------------------------------:|-------------------------:|------------:|-------:|
|                        Cumulative merge time of primary shards |                          |     9.23488 |    min |
|                       Cumulative merge count of primary shards |                          |          20 |        |

### Without this change
```shell
Final index size: 840.2mb
curl localhost:9200/_cat/indices
green open logs-231998 wZ8LNHeDQaCiLgOkJTjKxg 1 0 11961342 0 840.2mb 840.2mb
```
**number of swaps to number of docs ratio is 47**
```
Reorder stats
totalDocs: 11961342; numTerms: 964; totalPostings: 88726436; totalSwapsPerformed: 571132294; postingsDocRatio: 7; swapsDocsRatio: 47
47254463 55941681 48415352 38245831 39694052 36558063 37907697 33234795 32316838 30361191 27886028 26467715 24669276 22745648 21136997 17510104 16757532 14029031 0 0 0 0 0 0 0
```
| Metric | Task | Value | Unit |
| ---: | ---: | ---: | ---: |
| Cumulative merge time of primary shards |  | 15.5215 | min |
| Cumulative merge count of primary shards |  | 20 |  |

So overall time to perform merge reduced from 15.5 mins to 9.23 mins. Also, number of swaps performed to docs ratio reduced significantly here. 

### Without reordering

```
Final index size: 923.8mb
curl localhost:9200/_cat/indices
green open logs-231998 O0bppXZUREadOqVz80qJBw 1 0 11961342 0 923.8mb 
```

| Metric | Task | Value | Unit |
| ---: | ---: | ---: | ---: |
| Cumulative merge time of primary shards |  | 1.13032 | min |
| Cumulative merge count of primary shards |  | 21 |  |

rishabhmaurya: @jpountz let me know your thoughts on it, happy to contribute and improve it further. Thanks

jpountz: @rishabhmaurya Yes! Please feel free to open a PR and let's discuss there!

rishabhmaurya: @jpountz done https://github.com/apache/lucene/pull/13186. I wasn't sure if its worthy a PR, thus started discussion here.

Issue #12740, opened 2023-10-31
Better wire up HNSW concurrent merge config


zhaih:
### Description

Follow up of #12660, currently the HNSW concurrent merge need 2 parameters to make it work: `numMergeWorker` (num of thread spawn per merge) and an `ExecutorService` that is used to execute the tasks. Currently I put them as part of `Lucene99HnswVectorFormat` such that user have to override the Codec and create their own format to use it.
We're thinking whether this could eventually be put into IWC, such that user don't need to know about Codec stuff. One thing I'm not sure is that since Lucene does not need to always use HNSW and we could very much have some other vector impl in the future. So it doesn't make sense to me if we make this config too general.
Would like to open this issue to gather some ideas, any comment is welcomed!

jpountz:
Tangentially, maywe we should think of reviving @dweiss 's #9626. We did not move forward with concurrent merging because this approach used to parallelize across data structures, and usually the bottleneck would be terms/postings, and parallelizing within terms/postings sounded challenging. But now that vectors can be a typical bottleneck for merging and that we have ways to parallelize within a single field, we could look back into passing an executor (or ForkJoinPool?) that merges can use to run concurrently?

benwtrent:
@jpountz @zhaih Looking at @dweiss old proposal I am thinking we should add a `TaskExecutor` to  `IndexWriter`. The only thing thats weird is that with `ConcurrentMergeScheduler`, we already have one layer of parallelism from the `IndexWriter`. 

Do we think its cool to have a `TaskExecutor parallelMergeExecutor` that is passed to the `IndexWriter` and then is passed to all merges kicked off by the `ConcurrentMergeScheduler`?

My concern would be if the `parallelMergeExecutor.poolSize() <= ConcurrentMergeScheduler#maxThreadCount` & every merge had a vector field. Wouldn't this effectively make HNSW merging single threaded? Or do I misunderstand how `ConcurrentMergeScheduler` actually works?

msokolov:
Thinking about holistically as someone trying to tune the performance of the indexer I want as few knobs as possible that enable me to get the best performance for my particular workload. This is kind of tricky since some folks will be indexing and searching on the same node, whileothers will have dedicated a host entirely to indexing. In any case indexing clients can currently control the number of threads dedicated to index by: (1) setting the number of threads given to CMS and (2) controlling the number of indexing threads they create. Adding a third pool seems to make this a tricky problem. I wonder if we could leverage the same thread pool that is used by CMS?

[...]

PR #13124, opened 2024-02-21
Add new parallel merge task executor for parallel actions within a single merge action

benwtrent:
Opening this PR for discussion. I took a stab at https://github.com/apache/lucene/issues/12740

The idea is this:

 - Add new task executor to IWC & num parallel worker actions
 - Extend knn writer to accept the executor during merge
 - Throw if both executors (via format & writer) are provided

If this direction seems good, I can finish cleaning up things and write some tests.

@zhaih @jpountz y'all might be interested here.


benwtrent:
>  it would be nice to fully encapsulate the merging concurrency there instead of having two sources of merging concurrency that are not aware of one another.

To take advantage of the executor, etc. it needs to be accessible from the scheduler. So, we would have to update the `MergeScheduler` to have some methods to return the executor for us to use and pass to `MergeState` (which is only created via the `SegmentMerger` object). This means that the scheduler for individual merges and the parallelism available to those individual merges are independent.

I don't know how they would ever be "aware" of one another. 

> I'm also keen on keeping the codec API as simple as possible and adding the executor as a member of the MergeState rather than a new parameter of all codec write APIs.

I updated the PR to do just this. Its a much cleaner API.

zhaih:
+1 to move executor away from the Codec API (altho it's me who placed them there LOL)

> it would be nice to fully encapsulate the merging concurrency there instead of having two sources of merging concurrency that are not aware of one another.

I like the idea, but I'm a little bit worry about the situation where the HNSW merging threads are taking most of the resource such that original CMS threads cannot be executed. As the `numWorkers` we configured is a per-merge limit, which means if there are several big merges happening at the same time, say `n` merges, they will use `n * (numWorkers + 1)` threads, the thread pool can run out of threads very fastly and result in smaller merges be blocked by large merges.

So I think it is still better to have two separate thread pools for inter-segment merge and inner segment merge, but I wonder whether we can have a `ThreadPoolManager` which configures both inter and inner pool. And CMS and HNSW merge (and any future merges) will source from this manager?

Also I heard there's a usecase where they're currently sharing one CMS with multiple IW in a same process for better resource management, this new `ThreadPoolManager` can be a solution for them as well maybe?

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java
```diff
@@ -152,7 +153,25 @@ public Lucene99HnswVectorsFormat() {
    * @param beamWidth the size of the queue maintained during graph construction.
    */
   public Lucene99HnswVectorsFormat(int maxConn, int beamWidth) {
-    this(maxConn, beamWidth, DEFAULT_NUM_MERGE_WORKER, null);
```
zhaih, line 155: If there will be more future inner segment concurrent merge, I think it might still be better to leave the `numMergeWorkers` here rather than put it in IWC? Otherwise we'll either have too many similar configures in IWC or hard to allocate weight on different types of merges?

benwtrent:
> So I think it is still better to have two separate thread pools for inter-segment merge and inner segment merge, but I wonder whether we can have a ThreadPoolManager which configures both inter and inner pool. And CMS and HNSW merge (and any future merges) will source from this manager?

Maybe? But this doesn't seem to simplify anything for users. They will still need two things to configure (number of merges that can happen at a time, number of threads available to all those merges to take action). 

There is always the issue of best using resources and the potential of over-allocating the CPU cores. I don't think any finagling we do will ultimately change that.

Selfishly, I really don't want to mess with the CMS code at all. Even if we switch it to use an executor of some sort, the complexities of I/O throttling & executable queuing would still all exist.

jpountz:
> So, we would have to update the MergeScheduler to have some methods to return the executor for us to use and pass to MergeState (which is only created via the SegmentMerger object). This means that the scheduler for individual merges and the parallelism available to those individual merges are independent.

I agree with passing some executor from the scheduler to the `MergeState`, but I'm not sure I agree that this implies that inter-merge and intra-merge parallelism would be independent. For instance the merge scheduler could return a custom `Executor` that dynamically decides to run a new task in the current thread or to fork to a separate thread depending on how many threads are currently busy across all merges vs. the current value of `ConcurrentMergeScheduler.maxThreadCount`?

benwtrent:
OK, my head is spinning a bit trying to grok the side-effects of the CMS. I think I understand most of it now. 

Currently we have two adjustable parameters. 

`maxThreadCount` dictates merge through put, or now many merges we have at a time.

`mergeWorkerCount` dictates back pressure, or when we stop allowing merges to be queued at all & block upstream indexing from continuing. 

Throughout the life of a merge, it can become (un)paused numerous times, this is controlled via the `RateLimitedIndexOutput` which only rate limits WRITING results, not reading or anything else related to threads. 

Now we want to add a new configuration, `maxMergeThreads`, which will control merge latency by adding parallelism to each individual thread. 

Since merge thread pausing only has to do with index output, I am not sure we need to add any individual thread throttling other than what’s already there. The Directory wrapper will pause/throttle all writing occurring for the merge. This is acceptable even if the merge is using multiple threads. 

I also think that small merges (<50MB) should never be allowed to run over multiple threads. Similar to how we never throttle those because they are so small, the benefit in latency reduction will be minuscule and we should reserve the extra thread usage to larger merges. 

What I am stuck on is on this: 
 - Should we use a common pool for all merges? Thus restricting the total threads used by merging to be `maxThreadCount` + `maxMergeThreads`? This will simplify the logic in the CMS significantly as a single task executor can be used for all merges.
 - Or should we use an individual executor/pool per merge? Thus restricting total threads by `maxMergeThreads` * `maxMergeThreads` (or some fraction of `maxMergeThreads`)? This could get interesting… How do we determine how many threads each merge can get? Are we ok with creating a new task executor on every larger merge and then closing it?

What do you think @jpountz?

One other thing we need to fix is this idea of “numWorkers” in HNSW. It seems like it should just pick optimal slices given the number of vectors it has (similar to multi-segment search stuff). Chunk itself into those slices and then be at the behest of the task executor. What say you @zhaih ? It seems weird to use `numWorkers` as a way to say “only use these many threads” when we have no knowledge of how many threads the task executor actually has.

benwtrent:
> Intra-merge concurrency would take advantage of the fact that there will sometimes be fewer active merges than threads to enable intra-merge concurrency. 

Sorry for being so dense, `Executor#execute` finally clicked it for me. `Executor#execute` can check for `currentRunningThreads + mergeThreads.size()`. If its larger than `>= maxThreadCount` execute in the current thread, otherwise spawn or give a thread from a pool. This would then increment `currentRunningThreads`.

I will spin a little bit on this to see what I can come up with.

zhaih:
So the current way of HNSW concurrent merge implemented is: each worker will try to use an AtomicInteger to coordinate and only do a small batch of work (1024 documents) each time. The advantage is we are able to load balance between workers and I remember this did brings some (5-10%) performance gain when I was testing it. 

Maybe, instead of specify a numWorkers per merge, we can default a expected work load per thread (like 10K), and then allocate numWorkers dynamically? But still keep the current way of merge to keep the performance?

One thing I'm worried about putting all things into CMS is that we're binding intra segment merge with CMS. But to my understanding using CMS means we're using background thread to merge and merge become indeterministic, such that there are still some part of users are using SMS (SerialMS) or similar thing to keep the deterministic of merging. But on the other hand the HNSW concurrent merge does not affect that aspect at all, no matter how many threads you're using it won't affect the determinism of merge result. So if we bind those two together whether we potentially prevent a part of users using the intra-segment merges?

benwtrent:
@zhaih I don't see any reason why we also cannot extend the SerlialMS and allow multiple threads per merge ran. 

We will have to update the base `MergeScheduler` class anyways as the merges being ran don't know anything about who kicked off the merge (and shouldn't).

The reason for all the CMS discussion is that it is the hardest to implement correctly (to me anyways...). For SerialMS, users could provide a number and its a static executor with that number of threads.

> But still keep the current way of merge to keep the performance?

I really don't think users should configure numWorkers or workPerThread at all. I would much prefer us supply good defaults and remove configuration.

If we did the parallelism outside the codec to being with, I don't think we would have added any configurable values to the HNSW codec.

> So if we bind those two together whether we potentially prevent a part of users using the intra-segment merges?

I am not thinking about binding them. I think that `MergeScheduler` itself should be extended to return a TaskExecutor (probably defaulting to `null` to indicate none, or maybe `SameThreadExecutorService`).

CMS is just the most difficult one to figure out.

zhaih:
> I am not thinking about binding them. I think that MergeScheduler itself should be extended to return a TaskExecutor (probably defaulting to null to indicate none, or maybe SameThreadExecutorService).

Sounds good

> I really don't think users should configure numWorkers or workPerThread at all. I would much prefer us supply good defaults and remove configuration.

No I'm not suggesting that either, I'm open to let user specify or we provide default, just want to make sure the performance is not affected (by not pre-dividing the doc space into each thread).

benwtrent:
OK, I took a stab at this @jpountz latest commit has a POC. Need to test. But early performance testing shows indexing speed is better and forcemerge speed is just about the same. I am still using the numWorkers setting from HNSW to determine workers. This is mainly to prevent testing too many performance changes at a time.

lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
```diff
@@ -910,4 +936,58 @@ public void setSuppressExceptions(ConcurrentMergeScheduler cms) {
           }
         });
   }
+
+  private class ScaledExecutor extends ThreadPoolExecutor {
```
jpountz, line 940: nit: I usually have a preference for composition over inheritance, ie. could we wrap the thread-pool executor instead of wrapping it?

lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
```diff
@@ -910,4 +936,58 @@ public void setSuppressExceptions(ConcurrentMergeScheduler cms) {
...
+    AtomicInteger activeCount = new AtomicInteger(0);
+
+    public ScaledExecutor() {
+      super(
+          Math.max(0, maxThreadCount - 1),
+          Math.max(1, maxThreadCount - 1),
+          Long.MAX_VALUE,
+          TimeUnit.NANOSECONDS,
+          new SynchronousQueue<>());
+    }
```
jpountz, line 951: Thinking out loud: for the case when tens of index writers are open in the same JVM, we may want to configure a timeout on threads in order to avoid spending too much heap on idle threads?

lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
```diff
@@ -910,4 +936,58 @@ public void setSuppressExceptions(ConcurrentMergeScheduler cms) {
...
+
+    boolean incrementUpTo(int max) {
+      while (true) {
+        int value = activeCount.get();
+        if (value >= max) {
+          return false;
+        }
+        if (activeCount.compareAndSet(value, value + 1)) {
+          return true;
+        }
```
jpountz, line 972: If I read correctly, this tries to keep `activeCount <= maxThreadCount`. I was thinking we should try to keep `activeCount <= maxThreadCount - mergeThreadCount()`. Otherwise we're effectively using more than `maxThreadCount` for merging in total, which I find a bit surprising?

lucene/core/src/java/org/apache/lucene/index/MergeScheduler.java
```diff
@@ -52,6 +56,14 @@ public Directory wrapForMerge(OneMerge merge, Directory in) {
     return in;
   }
 
+  /**
+   * Provides an executor for parallelism during a single merge operation. By default, this method
+   * returns an executor that runs tasks in the calling thread.
+   */
+  public Executor getInterMergeExecutor(OneMerge merge) {
```
jpountz, line 63: Should it be get**Intra**MergeExecutor?

lucene/core/src/java/org/apache/lucene/index/NoMergeScheduler.java
```diff
@@ -52,4 +53,9 @@ public Directory wrapForMerge(OneMerge merge, Directory in) {
   public MergeScheduler clone() {
     return this;
   }
+
+  @Override
+  public Executor getInterMergeExecutor(OneMerge merge) {
+    return null;
```
jpountz, line 59: nit: throw instead?

lucene/core/src/java/org/apache/lucene/index/SegmentMerger.java
```diff
@@ -56,13 +58,19 @@ final class SegmentMerger {
...
       throw new IllegalArgumentException(
           "IOContext.context should be MERGE; got: " + context.context);
     }
-    mergeState = new MergeState(readers, segmentInfo, infoStream);
+    mergeState =
+        new MergeState(
+            readers,
+            segmentInfo,
+            infoStream,
+            parallelMergeTaskExecutor == null ? null : new TaskExecutor(parallelMergeTaskExecutor));
```
jpountz, line 73: Nit: it's a bit weird to use a class from the search package for merging (TaskExecutor). Should merging get access to the raw Executor, which is a bit more flexible (I don't know if all formats will be able to split work into a list of tasks up-front)? Vectors could still wrap inside a `TaskExecutor` for convenience?

jpountz (commented):
I left some nitpicks but I find the change promising/exciting!

lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
benwtrent: Yeah no doubt.

lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
benwtrent: > Thinking out loud: for the case when tens of index writers are open in the same JVM, we may want to configure a timeout on threads in order to avoid spending too much heap on idle threads?

The timeout would only apply to non-core threads. But, your point is taken, it is possible for merges to be idle for a while and we don't want threads just sitting around taking up unnecessary space. 

So, I can make `core` always `0`, and rely on the pool dynamically adding threads up to max.

lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
benwtrent: @jpountz you are correct, my logic here isn't what it needs to be. I will fix it up.

lucene/core/src/java/org/apache/lucene/index/MergeScheduler.java
benwtrent: 🤦 `inter` vs `intra` is just as bad as `affect` and `effect`.

msokolov:
ooh exciting! I left some comments in a related issue that were maybe a little clueless given all the discussion here that I missed until now. Still I'm happy about the direction this seems to be going re: moving the configuration out of codec and into merge policy, and coordinating thread usage across inter/intra segment merges

lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
```diff
@@ -910,4 +936,58 @@ public void setSuppressExceptions(ConcurrentMergeScheduler cms) {
...
+    }
+
+    private void updatePoolSize() {
+      int newMax = Math.max(0, maxThreadCount - 1);
+      if (newMax > getCorePoolSize()) {
+        setMaximumPoolSize(Math.max(newMax, 1));
+        setCorePoolSize(newMax);
+      } else {
+        setCorePoolSize(newMax);
+        setMaximumPoolSize(Math.max(newMax, 1));
```
zhaih, line 960: Aren't those two branch the same?

benwtrent: The order of setting the max is important. 

But I am gonna change this. I think the core thread count should be `0` and we shouldn't update it at all and allow the threadpool to scale.

lucene/core/src/java/org/apache/lucene/index/SegmentMerger.java
shubhamvishu: > it's a bit weird to use a class from the search package for merging (TaskExecutor)

+1 ...... @jpountz since now we are using `TaskExecutor` on both search and indexing use cases does it make sense to move it into `org.apache.lucene.util` package maybe?

lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
```diff
@@ -910,4 +936,68 @@ public void setSuppressExceptions(ConcurrentMergeScheduler cms) {
...
+        synchronized (ConcurrentMergeScheduler.this) {
+          int max = maxThreadCount - mergeThreads.size();
+          int value = activeCount.get();
+          if (value < max) {
+            activeCount.incrementAndGet();
+            isThreadAvailable = true;
+          } else {
+            isThreadAvailable = false;
+          }
+        }
```
benwtrent, line 987: One major decision I made here is that CMS thread throttling knows nothing about intra-merge threads.

This means, if there are intra-merge threads running, it is possible that concurrency creeps above `maxThreadCount` until the intra-merge thread(s) finish(es).

Making throttling & overall CMS behavior more aware of intra-merge thread usage will be a much trickier change. I wasn't sure it was worth it. I am open to opinions on this @msokolov @jpountz .

msokolov: I am going to just make stuff up here since I don't know this code well at all -- but I think I was assuming there was a ThreadPoolExecutor somewhere that was doling out threads for both of these operations. Now I'm realizing perhaps that's not the case and CMS is doing all that tracking on its own. I wonder if we should open a separate issue to migrate CMS to use JDK's thread pool abstractions? It could make some of this easier to handle.

lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
```diff
@@ -910,4 +941,58 @@ public void setSuppressExceptions(ConcurrentMergeScheduler cms) {
...
+              0, Math.max(1, maxThreadCount), 1L, TimeUnit.MINUTES, new SynchronousQueue<>());
+    }
+
+    void shutdown() {
+      executor.shutdown();
+    }
+
+    private void updatePoolSize() {
+      executor.setMaximumPoolSize(Math.max(1, maxThreadCount));
+    }
```
jpountz, line 962: It feels like we don't actually need this since we already have control on the thread pool size via `maxThreadCount`. What about setting a high max thread count in the ctor and never updating it (ie. making it a "cached" executor rather than a "scaled" executor).

jpountz (commented):
I like the change a lot.

Currently this is only used by vectors. Could we also take advantage of this new executor to paralellize at the data structure level and submit separate tasks for postings, points, norms, doc values, etc. from `SegmentMerger.merge`? Earlier experiments suggested that it did not help much before there typically is one bottleneck but this would be good dogfooding for this change?

lucene/core/src/java/org/apache/lucene/index/MergeScheduler.java
```diff
@@ -52,6 +53,14 @@ public Directory wrapForMerge(OneMerge merge, Directory in) {
     return in;
   }
 
+  /**
+   * Provides an executor for parallelism during a single merge operation. By default, the method
+   * returns `null` indicating that there is no parallelism during a single merge operation.
+   */
+  public Executor getIntraMergeExecutor(OneMerge merge) {
+    return null;
```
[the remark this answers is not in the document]
benwtrent: My concern is that certain things might actually be slower trying to do parallel work when there aren't any threads to be had. I suppose we will end up in this situation when all the merge threads are actually being used by other merges...

I need to benchmark this with vectors to see if things are measurably worse.

benwtrent:
@jpountz 

OK, so I did some benchmarking to look at the impact of this change. 500k docs, flushing segments every 1MB and then force merging. 

For all of these, the number of workers is `8` and the threads available are `8`.

Baseline single threaded (no concurrency in HNSW)
```
Indexed: 467617ms
Force merge: 593037 ms
```

Baseline multi-threaded (executor separate from CMS)
```
Indexed: 173143ms
Force merge: 120341 ms
```

Candidate, CMS with 8 merge threads, sharing with intra-merge (what this PR is doing). Shows that we share the threads with the CMS background threads. 

There is a ton of merging work being done, so usually just 1 thread is being passed to the intra-merge executor. This is an expected result and working as intended.
However, we see a speed up in force-merge as no other merge activity is occurring and it can use all the provided CMS threads.
```
Indexed: 424924ms
Force merge: 121705 ms
```

To confirm that this is indeed the case and that indexing wasn’t slowed down due to some other weird overhead, I gave 2x as many threads to intra-merging. This effectively removes the limit that CMS is providing to intra-merges. 
Shows that that the indexing is now inline with how it is now, basically using 2x as many threads as configured for CMS (8 merge threads and 8 intra-merge threads).
```
Indexed: 171825ms
Force merge: 121886 ms
```

To make sure we don’t slow down too much compared to baseline single threaded, I gave 8 workers to HNSW merge, but only provided a SameThreadExecutorService.

I did this to justify the default to a SameThreadExecutorService, even when HNSW has > 1 worker.

Well within the runtime of baseline.
```
Indexed: 418529ms
Force merge: 524468 ms
```

lucene/core/src/java/org/apache/lucene/index/MergeScheduler.java
```diff
@@ -52,6 +56,14 @@ public Directory wrapForMerge(OneMerge merge, Directory in) {
     return in;
   }
 
+  /**
+   * Provides an executor for parallelism during a single merge operation. By default, the method
+   * returns `null` indicating that there is no parallelism during a single merge operation.
```
zhaih, line 61: I think it's ok to force this method always return non-null? Since we already have provided a good default?

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java
```diff
@@ -152,7 +154,25 @@ public Lucene99HnswVectorsFormat() {
...
+      throw new IllegalArgumentException(
+          "beamWidth must be positive and less than or equal to "
+              + MAXIMUM_BEAM_WIDTH
+              + "; beamWidth="
+              + beamWidth);
+    }
+    this.maxConn = maxConn;
+    this.beamWidth = beamWidth;
+    this.mergeExec = null;
+    this.numMergeWorkers = 1;
```
zhaih, line 175: Can remove those 2?

zhaih: Oh I see, you haven't yet put in the change for auto figure out number of workers

zhaih (commented):
A very neat approach! The code change is way less than I expected, nice work!

benwtrent:
@zhaih @jpountz I am going to create a separate issue around making HNSW worker slicing automatic. It will require a bunch of its own benchmarking and work and honestly seems orthogonal to this particular change as HNSW merging still works fine when statically setting the number of workers & using the CMS thread pool.

jpountz:
Thanks for sharing performance numbers @benwtrent, very interesting. Also double checking if you saw my above comment: https://github.com/apache/lucene/pull/13124#pullrequestreview-1930418114, I'd have more confidence about the way merge concurrency is exposed if we took advantage of it in more than one place. If there's some difficulty with it, I'm happy with looking into it in a follow-up.

jpountz:
Yes exactly, something very simple, mostly to exercise intra-merge concurrency with more than just vectors.

benwtrent:
> Yes exactly, something very simple, mostly to exercise intra-merge concurrency with more than just vectors.

Latest commit adds `TaskExecutor` actions to merge to allow different merging field types to be merged in parallel. 

I needed to use `TaskExecutor` as it abstracts out the exception handling (all these methods throw `IOException`).

I still need to benchmark it with Lucene-util

benwtrent:
@jpountz ok, that naive attempt failed as norms & terms apparently need to be merged in order (one or the other would fail due to missing files...). 

I am not sure if this is true with other things. But when I adjusted for this (latest commit) I did see a speed improvement.

My test was over IndexGeoNames, I did no segment merging and flushed every 12MB to ensure I got many segments. I also set CMS merge threads to `8` threads. 

Baseline forcemerge(1): 17793 ms
Candidates forcemerge(1): 13739 ms

lucene/core/src/java/org/apache/lucene/index/SegmentMerger.java
```diff
@@ -130,19 +135,31 @@ MergeState merge() throws IOException {
...
             segmentWriteState.segmentSuffix);
 
+    TaskExecutor taskExecutor = new TaskExecutor(mergeState.intraMergeTaskExecutor);
+    List<Callable<Void>> mergingTasks = new ArrayList<>();
+
     if (mergeState.mergeFieldInfos.hasNorms()) {
       mergeWithLogging(this::mergeNorms, segmentWriteState, segmentReadState, "norms", numMerged);
     }
 
     mergeWithLogging(this::mergeTerms, segmentWriteState, segmentReadState, "postings", numMerged);
```
jpountz, line 145: Can we have a parallel task that handles norms + terms so that the order is respected?

benwtrent: OK, I added a commit that does this. The candidate merge time is now:

`forceMerge took 4965 ms`
vs. baseline (i re-ran to make sure)
`forceMerge took 15783 ms`

So, 3x faster? Seems pretty good. The major downside is that now memory usage on merge might increase as we are potentially doing all this merge activity at the same time. If we are cool with that, this seems like a nice speed improvement.

jpountz (approved):

lucene/core/src/java/org/apache/lucene/index/SegmentMerger.java
jpountz: Presumably, the worst-case scenario is not much worse than if you have many concurrent merges, so I think it's fine.

lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
```diff
@@ -902,12 +932,52 @@ private static String getSegmentName(MergePolicy.OneMerge merge) {
...
-        new ConcurrentMergeSchedulerAccess() {
-          @Override
-          public void setSuppressExceptions(ConcurrentMergeScheduler cms) {
-            cms.setSuppressExceptions();
-          }
-        });
+    TestSecrets.setConcurrentMergeSchedulerAccess(ConcurrentMergeScheduler::setSuppressExceptions);
+  }
+
+  private class CachedExecutor implements Executor {
```
zhaih, line 938: Can we have some simple javadoc explaining what this executor is doing? E.g. will use a spare thread if it is within configured max thread and will run directly on caller thread if we have no extra thread available?

zhaih (approved):
Very exciting PR!

mikemccand:
This looks like a great change! ... Lucene finally catching up to modern concurrency :)  I have not looked closely yet.

But nightly benchy is upset, prolly related to this, when running `runFacetsBenchmarks.py`.  I haven't looked closely yet, and will be offline mostly next couple days (sorry for the hit&run!), but here's the exception is popped out:

```
Exception in thread "main" org.apache.lucene.store.AlreadyClosedException: this IndexWriter is closed
        at org.apache.lucene.index.IndexWriter.ensureOpen(IndexWriter.java:913)
        at org.apache.lucene.index.IndexWriter.ensureOpen(IndexWriter.java:926)
        at org.apache.lucene.index.IndexWriter.commit(IndexWriter.java:4070)
        at perf.facets.IndexFacets.main(IndexFacets.java:127)
Caused by: java.lang.RuntimeException: Only the merge owner thread can call pauseNanos(). This thread: pool-1-thread-1, owner thread: Thread[#290,Lucene Merge Thread #12,5,main]
        at org.apache.lucene.index.MergePolicy$OneMergeProgress.pauseNanos(MergePolicy.java:142)
        at org.apache.lucene.index.MergeRateLimiter.maybePause(MergeRateLimiter.java:147)
        at org.apache.lucene.index.MergeRateLimiter.pause(MergeRateLimiter.java:92)
        at org.apache.lucene.store.RateLimitedIndexOutput.checkRate(RateLimitedIndexOutput.java:86)
        at org.apache.lucene.store.RateLimitedIndexOutput.writeBytes(RateLimitedIndexOutput.java:55)
        at org.apache.lucene.store.ByteBuffersDataOutput.copyTo(ByteBuffersDataOutput.java:339)
        at org.apache.lucene.codecs.lucene90.blocktree.Lucene90BlockTreeTermsWriter$TermsWriter.writeBlock(Lucene90BlockTreeTermsWriter.java:1022)
        at org.apache.lucene.codecs.lucene90.blocktree.Lucene90BlockTreeTermsWriter$TermsWriter.writeBlocks(Lucene90BlockTreeTermsWriter.java:760)
        at org.apache.lucene.codecs.lucene90.blocktree.Lucene90BlockTreeTermsWriter$TermsWriter.pushTerm(Lucene90BlockTreeTermsWriter.java:1133)
        at org.apache.lucene.codecs.lucene90.blocktree.Lucene90BlockTreeTermsWriter$TermsWriter.write(Lucene90BlockTreeTermsWriter.java:1089)
        at org.apache.lucene.codecs.lucene90.blocktree.Lucene90BlockTreeTermsWriter.write(Lucene90BlockTreeTermsWriter.java:399)
        at org.apache.lucene.codecs.FieldsConsumer.merge(FieldsConsumer.java:95)
        at org.apache.lucene.codecs.perfield.PerFieldPostingsFormat$FieldsWriter.merge(PerFieldPostingsFormat.java:205)
        at org.apache.lucene.index.SegmentMerger.mergeTerms(SegmentMerger.java:236)
        at org.apache.lucene.index.SegmentMerger.mergeWithLogging(SegmentMerger.java:325)
        at org.apache.lucene.index.SegmentMerger.lambda$merge$0(SegmentMerger.java:147)
        at org.apache.lucene.search.TaskExecutor$TaskGroup.lambda$createTask$0(TaskExecutor.java:117)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
        at org.apache.lucene.index.ConcurrentMergeScheduler$CachedExecutor.lambda$execute$0(ConcurrentMergeScheduler.java:978)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
        at java.base/java.lang.Thread.run(Thread.java:1583)
        Suppressed: java.lang.RuntimeException: Only the merge owner thread can call pauseNanos(). This thread: pool-1-thread-2, owner thread: Thread[#290,Lucene Merge Thread #12,5,main]
                at org.apache.lucene.index.MergePolicy$OneMergeProgress.pauseNanos(MergePolicy.java:142)
                at org.apache.lucene.index.MergeRateLimiter.maybePause(MergeRateLimiter.java:147)
                at org.apache.lucene.index.MergeRateLimiter.pause(MergeRateLimiter.java:92)
                at org.apache.lucene.store.RateLimitedIndexOutput.checkRate(RateLimitedIndexOutput.java:86)
                at org.apache.lucene.store.RateLimitedIndexOutput.writeBytes(RateLimitedIndexOutput.java:55)
                at org.apache.lucene.store.DataOutput.writeBytes(DataOutput.java:54)
                at org.apache.lucene.util.packed.DirectWriter.flush(DirectWriter.java:97)
                at org.apache.lucene.util.packed.DirectWriter.add(DirectWriter.java:83)
                at org.apache.lucene.codecs.lucene90.Lucene90DocValuesConsumer.writeValuesSingleBlock(Lucene90DocValuesConsumer.java:349)
                at org.apache.lucene.codecs.lucene90.Lucene90DocValuesConsumer.writeValues(Lucene90DocValuesConsumer.java:328)
                at org.apache.lucene.codecs.lucene90.Lucene90DocValuesConsumer.doAddSortedNumericField(Lucene90DocValuesConsumer.java:707)
                at org.apache.lucene.codecs.lucene90.Lucene90DocValuesConsumer.addSortedSetField(Lucene90DocValuesConsumer.java:772)
                at org.apache.lucene.codecs.DocValuesConsumer.mergeSortedSetField(DocValuesConsumer.java:853)
                at org.apache.lucene.codecs.DocValuesConsumer.merge(DocValuesConsumer.java:148)
                at org.apache.lucene.codecs.perfield.PerFieldDocValuesFormat$FieldsWriter.merge(PerFieldDocValuesFormat.java:154)
                at org.apache.lucene.index.SegmentMerger.mergeDocValues(SegmentMerger.java:205)
                at org.apache.lucene.index.SegmentMerger.mergeWithLogging(SegmentMerger.java:325)
                at org.apache.lucene.index.SegmentMerger.lambda$merge$1(SegmentMerger.java:155)
                ... 6 more
                Suppressed: java.lang.RuntimeException: Only the merge owner thread can call pauseNanos(). This thread: pool-1-thread-2, owner thread: Thread[#290,Lucene Merge Thread #12,5,main]
                        at org.apache.lucene.index.MergePolicy$OneMergeProgress.pauseNanos(MergePolicy.java:142)
                        at org.apache.lucene.index.MergeRateLimiter.maybePause(MergeRateLimiter.java:147)
                        at org.apache.lucene.index.MergeRateLimiter.pause(MergeRateLimiter.java:92)
                        at org.apache.lucene.store.RateLimitedIndexOutput.checkRate(RateLimitedIndexOutput.java:86)
                        at org.apache.lucene.store.RateLimitedIndexOutput.writeByte(RateLimitedIndexOutput.java:48)
                        at org.apache.lucene.codecs.CodecUtil.writeBEInt(CodecUtil.java:654)
                        at org.apache.lucene.codecs.CodecUtil.writeFooter(CodecUtil.java:410)
                        at org.apache.lucene.codecs.lucene90.Lucene90DocValuesConsumer.close(Lucene90DocValuesConsumer.java:116)
                        at org.apache.lucene.codecs.perfield.PerFieldDocValuesFormat$ConsumerAndSuffix.close(PerFieldDocValuesFormat.java:86)
                        at org.apache.lucene.util.IOUtils.close(IOUtils.java:85)
                        at org.apache.lucene.codecs.perfield.PerFieldDocValuesFormat$FieldsWriter.close(PerFieldDocValuesFormat.java:249)
                        at org.apache.lucene.index.SegmentMerger.mergeDocValues(SegmentMerger.java:204)
                        ... 8 more
Traceback (most recent call last):
  File "/l/util.nightly/src/python/nightlyBench.py", line 1957, in <module>
    run()
  File "/l/util.nightly/src/python/nightlyBench.py", line 533, in run
    runFacetsBenchmark.run_benchmark(f'{constants.BASE_DIR}/{NIGHTLY_DIR}',
  File "/l/util.nightly/src/python/runFacetsBenchmark.py", line 56, in run_benchmark
```

benwtrent:
Ah, that is likely due to this change. I bet throttling is occurring but in one of the other threadpool threads. 

I didn't override the thread construction, so maybe that is enough? 

I can look more closely next week

PR #13125, opened 2024-02-21
Make BP work on indexes that have blocks.

jpountz:
This is similar to the work we did on supporting index sorting on indexes that
have blocks, but for index reordering this time.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -186,7 +186,9 @@ API Changes
 
 New Features
 ---------------------
-(No changes)
+
+* GITHUB#13125: Recursive graph bisection is now supported on indexes that have blocks, as long as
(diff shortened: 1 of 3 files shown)


jpountz:
Draft for now because it builds on #13125 which has not been merged yet.

When we have support for BP + blocks, then we can enable BP on githubsearch cc @mikemccand.

mikemccand:
Yay, I confirmed that githubsearch [picked up that mention of me](https://githubsearch.mikemccandless.com/search.py?sort=recentlyUpdated&dd=status%3AOpen&dd=mentioned_users%3Amikemccand) above @jpountz!

And yes +1 to enable this in githubsearch: some good dog food here, yum yum!  I'll open an issue in githubsearch.

jpountz:
The PR is ready for review in case anyone is interested.

PR #13126, opened 2024-02-21
factor out (public static) MultiCollector.scoreMode(Collector[]) method

cpoerschke:
To allow re-use rather than duplication of the logic e.g. see https://github.com/apache/solr/pull/2248/commits/3cee3e591995520cdfed41a5228a2e01c3b5cc0f#r1485486388 thread.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/MultiCollector.java
+++ b/lucene/core/src/java/org/apache/lucene/search/MultiCollector.java
@@ -101,8 +101,7 @@ public class MultiCollector implements Collector {
     this.cacheScores = numNeedsScores >= 2;
   }
 
-  @Override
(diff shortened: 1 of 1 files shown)


jpountz:
I'm not too familiar with the Solr code but I see that it duplicates logic in `setScorer` as well, could it reuse `MultiCollector` directly instead of exposing the method that combines score modes?

cpoerschke:
> ... could it reuse `MultiCollector` directly instead of exposing the method that combines score modes?

I'd wondered the same-ish i.e. `MultiCollectorManager` re-use -- https://github.com/apache/solr/pull/2248/files#r1489857592 -- but so far don't see a way, perhaps just needs more looking at and into. @chatman or @dsmiley might also have more context or thoughts on this?

PR #13133, opened 2024-02-26
Vector accelerated GroupVInt decoder for MemorySegmentIndexInput

mccullocht:
### Vector accelerated GroupVInt decoder for MemorySegmentIndexInput

Use Vector APIs when decoding GroupVInt for `MemorySegmentIndexInput`. In most cases we can read the
maximum group length (16 bytes) from the segment and then use a small static table and `Vector.rearrange()`
to produce the output `long` values. This is about 35-40% faster than the existing scalar/unaligned load strategy.

Benchmarks run on M1 Mac
```
main
Benchmark                                                  (size)   Mode  Cnt   Score   Error   Units
GroupVIntBenchmark.benchMMapDirectoryInputs_readGroupVInt      64  thrpt    5  10.031 ± 0.094  ops/us

group-varint-simd
Benchmark                                                  (size)   Mode  Cnt   Score   Error   Units
GroupVIntBenchmark.benchMMapDirectoryInputs_readGroupVInt      64  thrpt    5  14.666 ± 0.267  ops/us
```

Benchmarks run on github codespace (x86_64)
```
main
Benchmark                                                  (size)   Mode  Cnt  Score   Error   Units
GroupVIntBenchmark.benchMMapDirectoryInputs_readGroupVInt      64  thrpt    5  8.138 ± 0.223  ops/us

group-varint-simd
Benchmark                                                  (size)   Mode  Cnt   Score   Error   Units
GroupVIntBenchmark.benchMMapDirectoryInputs_readGroupVInt      64  thrpt    5  11.095 ± 0.487  ops/us
```


rmuir:
Hi, a couple suggestions:

1. Somehow, we need to avoid Vector API code inside the MemorySegment code. Just because MemorySegment is available, does not mean Vector API is usable, one of them is "preview" and the other is "incubator". For example, user may have a recent JVM that supports MemorySegment, which is automatically enabled, but not have passed special command-line flags to their JVM to enable incubator vector api, hardware may not support vector api, etc.
2. Need to guard use of vector api with checks, and fall back to scalar implementation when vector api would be very slow, see how it is currently done: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/internal/vectorization/VectorizationProvider.java . There are many checks already there, they should be re-used. Especially for integer vectors such as ByteVector operations it is necessary to check e.g. that actual AVX is available: https://github.com/apache/lucene/blob/main/lucene/core/src/java20/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java#L74-L78 . These checks prevent code from executing 100x slower, which is unfortunately what happens when the stars do not align in hotspot. You can pass `-XX:UseAVX=0` to your benchmark to see this slowness in action.

uschindler:
I have to agree with Robert. We can't glue that code inside MemorySegmentIndexInput.

The only way to do this is to move the vector optimized decoding to a special VectorizationProvider interface which is loaded dynamically by our safety guards.

mccullocht:
@rmuir @uschindler Would it be sufficient to add a factory function to `VectorizationProvider` for this like `GroupVIntUtil.Decoder createGroupVIntDecoder(MemorySegment segment)`? I anticipated that his might be necessary so I already have a scalar implementation available.

I'm surprised by how slow this is with AVX off given that this can be implemented with SSE2 :(.

I don't think this can be auto-vectorized by the VM (and FWIW LLVM can't autovectorize it either). Given that the loop would be to a constant bound it is likely something would unroll it anyway and we'd end up with the same code.

uschindler:
Hi, you can try it out. Two important comments:
- In Java 21, MemorySegment is still a "preview" class, so it can't be used in any public method signatures. So actually to pass the MemorySegment around, you have to use `java.lang.Object` and cast it in the receiver. The coe compiled for Java 19-21 is stripping the preview flag, but the problem is that the public API in the proivider and the interface can't use the class. Sorry about this. Unfortunately this is still the case in Java 21, so it won't get better with Lucene 10 (which will bei Java 21 minimum).
- Please implement the vector code in the "main20" folder against Java 20 (where the other vector code lives). As the vector API did not change between those versions, we don't need to extract the api for Java 20, so please revert the changes to API extractor.

uschindler:
> I'm surprised by how slow this is with AVX off given that this can be implemented with SSE2 :(.

This is why we try to avoid the incubating vector API as much as possible. The code needs to be tested with all platforms and bitsizes using extensive benchmarking. The problem is:
- If Hotspot does not have an optimization for the actual CPU -> slow
- If you use a JDK with Graal -> slow
- If you use non-Hotspot (e.g., OpenJ9) -> slow
- If you disable teiered compilation / use client compiler -> slow

Actually the code is slow because without support in hardware, the code executes as written in Java code, producing hundreds of instances.

Actually another thing: Once you fixed the code, make sure to show a benchmark on real queries. Just because the group vint decoding is 30% faster, it does not mean you would see any difference in production. Normally we would only accept incubator vector optimizations if the results are at least 4 times faster than scalar code (e.g., the float dotproduct is 12 to 16 times faster, but still the effect on query/merging performance is not 16 times faster, just about 15%).

So if the effect on queries is <5% on query performance, I would disagree to merge this. Sorry!

rmuir:
> I'm surprised by how slow this is with AVX off given that this can be implemented with SSE2 :(.

Yes, it is surprising: we found the same situation with VectorUtil byte[] methods. Seems hotspot just doesn't in fact implement/plumb the SSE2 for integer stuff (they do it for float though!).

Also the "fallback" scalar code used by vector API is very very very slow. So you get things like 30x slowdowns in these situations: real performance trap. Traps that are easy to fall into with situations like virtualization where maybe AVX isn't plumbed to VMs. That's why we have so many checks in the code...

PR #13135, opened 2024-02-26
Fix TestLucene90FieldInfosFormat.testRandom

### LINKED ISSUE #13129: Reproducible test failure in TestLucene90FieldInfosFormat.testRandom
### Description

bisect shows df6bd25ce44ead0b38cffca6ef9bc3a9ce63dbea is the first bad commit.

```
org.apache.lucene.codecs.lucene90.TestLucene90FieldInfosFormat > testRandom FAILED
    java.lang.IllegalArgumentException: parent document and soft-deletes field can't be the same field ""
        at __randomizedtesting.SeedInfo.seed([E9414A90E55BE2D:7CD831A6BF35085E]:0)
        at org.apache.lucene.index.FieldInfos$FieldNumbers.<init>(FieldInfos.java:402)
        at org.apache.lucene.index.IndexWriter$5$1.<init>(IndexWriter.java:6626)
        at org.apache.lucene.index.IndexWriter$5.newFieldInfosBuilder(IndexWriter.java:6625)
        at org.apache.lucene.tests.index.BaseFieldInfoFormatTestCase.testRandom(BaseFieldInfoFormatTestCase.java:281)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
```

```
./gradlew test --tests TestLucene90FieldInfosFormat.testRandom -Dtests.seed=E9414A90E55BE2D -Dtests.nightly=true -Dtests.locale=da -Dtests.timezone=Japan -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```

_No response_

shubhamvishu:
### Description

Failing seed (`E9414A90E55BE2D`) generates empty string with `TestUtil#randomUnicodeString` whereas both field names must not be same as per [this](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/FieldInfos.java#L399-L405). This fix adds an extra check that prevents this condition.

Test : passes test with failing seed

Closes #13129

### DIFF
--- a/lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseFieldInfoFormatTestCase.java
+++ b/lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseFieldInfoFormatTestCase.java
@@ -278,6 +278,9 @@ public abstract class BaseFieldInfoFormatTestCase extends BaseIndexFileFormatTes
 
     String parentField = random().nextBoolean() ? TestUtil.randomUnicodeString(random()) : null;
 
+    if (softDeletesField != null && softDeletesField.equals(parentField)) {
+      parentField = null;
+    }
     var builder = INDEX_PACKAGE_ACCESS.newFieldInfosBuilder(softDeletesField, parentField);
 
     for (String field : fieldNames) {


lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseFieldInfoFormatTestCase.java
```diff
@@ -278,46 +278,50 @@ public void testRandom() throws Exception {
...
-      IndexableFieldType fieldType = randomFieldType(random(), field);
-      boolean storeTermVectors = false;
-      boolean storePayloads = false;
-      boolean omitNorms = false;
-      if (fieldType.indexOptions() != IndexOptions.NONE) {
-        storeTermVectors = fieldType.storeTermVectors();
-        omitNorms = fieldType.omitNorms();
-        if (fieldType.indexOptions().compareTo(IndexOptions.DOCS_AND_FREQS_AND_POSITIONS) >= 0) {
-          storePayloads = random().nextBoolean();
+    if (softDeletesField == null
```
msokolov, line 281: this seems like somewhat complex logic here to handle a problem we created ( parentField name the same as softDeletesField name - empty?) Would it be cleaner to add a loop generating a parentFieldName that is unique (or null)? Or maybe, if parentFIeld = softDeletesField, set it to null instead?

Or - am I missing something here - did you fix some other potential edge cases too?

shubhamvishu: Your understanding is right, we are only fixing the edge case when both are empty and obviously not null.

> Would it be cleaner to add a loop generating a parentFieldName that is unique (or null)? Or maybe, if parentFIeld = softDeletesField, set it to null instead?

That seems fair to me, it'll be cleaner if we set one to null

stefanvodita (approved):
With this new revision, we are addressing all the cases where `softDeletesField` and `parentField` are equal by setting `parentField` to `null`. I think that makes sense.

PR #13141, opened 2024-02-27
Add a nightly workflow to run and verify buildAndPushRelease.py and smokeTestRelease.py

dweiss:
This is an initial attempt to run a gh workflow to at least make sure these two scripts work and things are building. For the time being, I've omitted the tests entirely - nightly tests OOM on gh infrastructure and they're run/ verified elsewhere anyway (jenkins jobs). This is just to have a regular check that the python scripts pass.

The workflow is at the moment quite simple:
* it runs a local "release" in dev-mode (so also omits running gradle checks and signing artifacts),
* it then runs the smoke tester against the locally produced release. The smoke tester is run with the minimum default JDK but gradle is run with newer, head-released javas (17, 21 and 22-ea).

Remaining things to correct:
* [x] The workflow should run nightly (on schedule). Perhaps limit it to run just on lucene repo (and not on all forks?),
* [x] ~I'm tempted to run one job to produce a release, which would be followed by smoke-tester jdk matrix. Not sure whether it's worth it.~ I don't think this is worth complicating the workflow; the gain seems minimal.
* [x] ~it'd be nice to allow triggering this workflow manually (with an explicit artifact url).~ There are differences between dev mode and full (signed) mode that make it tricky and would require conditionals. Maybe a separate workflow is needed, if people would like to use it (but github can't handle nightly tests anyway, so the benefit is minimal).
* [ ] the demo and other run_java checks always run with the baseline (minimum) jdk - either we'd have to duplicate the checks on each matrix job or we'd have to modify the smoke tester somehow. **Perhaps in a follow-up issue, I think.**
* [ ] the folder "gradle assembleRelease" produces should also pass with the smoke tester, I think - currently it does not because it includes archives and other version references with the snapshot suffix (not the "base" version produced by buildAndPushRelease). I think it's odd that an "assembleRelease" target actually requires the python script to be properly used. **Again - perhaps worth another issue.**
* [ ] the buildAndPushRelease and the smoke tester script assume a lone 'lucene' subfolder in the candidate directory, which is a legacy from joint solr/lucene distribution. I think this could be nuked. **A follow-up issue too.**


dev-tools/scripts/smokeTestRelease.py
```diff
@@ -158,9 +158,9 @@ def checkJARMetaData(desc, jarFile, gitRevision, version):
           break
       else:
         if len(verify) == 1:
-          raise RuntimeError('%s is missing "%s" inside its META-INF/MANIFEST.MF' % (desc, verify[0]))
+          raise RuntimeError('%s is missing "%s" inside its META-INF/MANIFEST.MF: %s' % (desc, verify[0], s))
```
dweiss, line 161: I added dumping manifest content here, since I had some trouble getting X-Build-JDK to pass.

dev-tools/scripts/smokeTestRelease.py
```diff
@@ -577,7 +577,7 @@ def verifyUnpacked(java, artifact, unpackPath, gitRevision, version, testArgs):
   #       raise RuntimeError('lucene: file "%s" is missing from artifact %s' % (fileName, artifact))
   #     in_root_folder.remove(fileName)
 
-  expected_folders = ['analysis', 'analysis.tests', 'backward-codecs', 'benchmark', 'classification', 'codecs', 'core', 'core.tests',
+  expected_folders = ['analysis', 'analysis.tests', 'backward-codecs', 'benchmark', 'benchmark-jmh', 'classification', 'codecs', 'core', 'core.tests',
```
dweiss, line 580: Added benchmark-jmh, not present on branch_9x.

dweiss:
This is already working (https://github.com/dweiss/lucene/actions/runs/8067320693) but I need to polish a few things before it's worth of your time to review. Hopefully later tonight or tomorrow.

dweiss:
Ok. I think things can be improved but let's not make it too complicated at the beginning. Everything seems to work:

image

and since the workflow collects logs from each jvm, it's easy to verify these "runtime" JVMs are used for compilation and testing (even though all tests are actually skipped):

image

I'd like to merge this in to main and let it run a day or two.

dweiss:
I confirm it's working on schedule, as intended:
https://github.com/apache/lucene/actions/runs/8090015214

PR #13142, opened 2024-02-27
Replace Collections.synchronizedSet() with ConcurrentHashMap.newKeySet()

sabi0:
`ConcurrentHashMap.newKeySet()` provides better multi-threaded performance thanks to judicious use of synchronization.
Also it does not require "external" synchronization for iteration.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/TrackingDirectoryWrapper.java
+++ b/lucene/core/src/java/org/apache/lucene/store/TrackingDirectoryWrapper.java
@@ -17,14 +17,14 @@
 package org.apache.lucene.store;
 
 import java.io.IOException;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 /** A delegating Directory that records which files were written to and deleted. */
 public final class TrackingDirectoryWrapper extends FilterDirectory {
 
-  private final Set<String> createdFileNames = Collections.synchronizedSet(new HashSet<String>());
+  private final Set<String> createdFileNames = ConcurrentHashMap.newKeySet();
 
   public TrackingDirectoryWrapper(Directory in) {
     super(in);
@@ -61,10 +61,8 @@ public final class TrackingDirectoryWrapper extends FilterDirectory {
   @Override
   public void rename(String source, String dest) throws IOException {
     in.rename(source, dest);
-    synchronized (createdFileNames) {
-      createdFileNames.add(dest);
-      createdFileNames.remove(source);
-    }
+    createdFileNames.add(dest);
+    createdFileNames.remove(source);
   }
 
   /** NOTE: returns a copy of the created files. */
--- a/lucene/core/src/java/org/apache/lucene/util/VirtualMethod.java
+++ b/lucene/core/src/java/org/apache/lucene/util/VirtualMethod.java
@@ -26,7 +26,7 @@ import java.util.concurrent.ConcurrentHashMap;
  * A utility for keeping backwards compatibility on previously abstract methods (or similar
  * replacements).
  *
- * <p>Before the replacement method can be made abstract, the old method must kept deprecated. If
+ * <p>Before the replacement method can be made abstract, the old method must be kept deprecated. If
  * somebody still overrides the deprecated method in a non-final class, you must keep track, of this
  * and maybe delegate to the old method in the subclass. The cost of reflection is minimized by the
  * following usage of this class:
--- a/lucene/core/src/java/org/apache/lucene/util/VirtualMethod.java
+++ b/lucene/core/src/java/org/apache/lucene/util/VirtualMethod.java
@@ -19,9 +19,8 @@ package org.apache.lucene.util;
 import java.lang.reflect.Method;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.util.Collections;
-import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * A utility for keeping backwards compatibility on previously abstract methods (or similar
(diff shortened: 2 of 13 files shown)


lucene/core/src/java/org/apache/lucene/util/VirtualMethod.java
```diff
@@ -100,7 +98,7 @@ public VirtualMethod(Class<C> baseClass, String method, Class<?>... parameters)
     this.method = method;
     this.parameters = parameters;
     try {
-      if (!singletonSet.add(baseClass.getDeclaredMethod(method, parameters)))
```
dweiss, line 103: This isn't enforced, but many people in this project prefer the verbose comparison instead of the negation - please don't correct this, thanks.

lucene/replicator/src/java/org/apache/lucene/replicator/nrt/PrimaryNode.java
```diff
@@ -158,10 +158,7 @@ public long getPrimaryGen() {
    */
   public boolean flushAndRefresh() throws IOException {
     message("top: now flushAndRefresh");
-    Set<String> completedMergeFiles;
```
dweiss, line 161: I don't think this is semantically the same as iteration may be concurrent to updates (while previously it was blocked for the time the set was cloned)... I wonder whether it'll have an impact on runtime.

lucene/core/src/java/org/apache/lucene/util/VirtualMethod.java
sabi0: I changed _to_ verbose comparison.
Should I still revert it?

lucene/replicator/src/java/org/apache/lucene/replicator/nrt/PrimaryNode.java
sabi0: Code here is written safely: it makes a copy of a shared set, processes the copy and then removes the processed elements from the shared set.
I believe it should be fine.

Unlike, for instance something like this:
https://github.com/sabi0/lucene/blob/0b3f2886eb9715b6d55b4df4f755ef87c8e9a5cc/lucene/replicator/src/java/org/apache/lucene/replicator/nrt/ReplicaNode.java#L695-L698
(Just an example, this code is also fine as all operations on the 'pendingMergeFiles' in that class are inside `synchronized (this)` blocks. But this is rather fragile. It takes a single mutation outside 'synchronized' to introduce a hard to find bug.)

lucene/core/src/java/org/apache/lucene/util/VirtualMethod.java
dweiss: Uh, sorry. Need another coffee.

lucene/replicator/src/java/org/apache/lucene/replicator/nrt/PrimaryNode.java
dweiss: The code is safe but how it works is different, especially with concurrency involved. The iteration over a concurrent hash map itself is not blocking updates to the collection (or other iterators) - often it doesn't matter, but sometimes it may - depends on the logic outside.

Seeing the patch diff only, it's hard to figure out the scope of such a change. Even when you see the entire code, it's sometimes tricky to figure out how it may behave. If this wasn't an automatic replacement and you've reviewed how those collections are used, I'm fine with the change.

dweiss:
I'm out of office this week. If anybody can pick this up, please do. Otherwise I'll return to it next wek.

lucene/core/src/java/org/apache/lucene/store/TrackingDirectoryWrapper.java
```diff
@@ -61,10 +61,8 @@ public void copyFrom(Directory from, String src, String dest, IOContext context)
   @Override
   public void rename(String source, String dest) throws IOException {
     in.rename(source, dest);
-    synchronized (createdFileNames) {
-      createdFileNames.add(dest);
-      createdFileNames.remove(source);
-    }
+    createdFileNames.add(dest);
+    createdFileNames.remove(source);
```
benwtrent, line 65: This means that another thread can mutate `createdFileNames` between `add` and `remove`. Are we sure this doesn't add any weird race conditions?

lucene/replicator/src/java/org/apache/lucene/replicator/nrt/ReplicaNode.java
```diff
@@ -625,19 +624,17 @@ public void run(CopyJob job) {
...
-            message(
-                "top: now cancel merge copy job="
-                    + mergeJob
-                    + ": file "
-                    + fileName
-                    + " is now being copied via NRT point");
-            mergeJob.cancel("newNRTPoint is copying over the same file", null);
-          }
+      for (CopyJob mergeJob : mergeCopyJobs) {
+        if (mergeJob.getFileNames().contains(fileName)) {
```
benwtrent, line 628: Since this no longer locks on `mergeCopyJobs`, this means during the iteration, the collection can be mutated. Are we sure this is OK?

benwtrent: `mergeCopyJobs` is protected, implying its use with sub-classes. You can see this in `SimpleReplicaNode`, which takes action while `synchronized(mergeCopyJobs)`. 

A quick search on github shows that other repositories take similar action. 

Are we sure this doesn't add a race condition to library user's code that sub-class the `ReplicaNode` class & synchronize on `mergeCopyJobs`?

lucene/core/src/java/org/apache/lucene/store/TrackingDirectoryWrapper.java
uschindler: I have the same problem. We need to avoid that. Unless there's an add+remove atomic call to ConcurrentSet we need to synchronize.

lucene/replicator/src/java/org/apache/lucene/replicator/nrt/PrimaryNode.java
uschindler: I agree, better keep old code.

uschindler:
Hi,
I am fine to apply the "safe" changes where iteration or an atomic add+remove is not required.
But all others should be reverted and external synchronization using synchronizedSet should be used.
Let's only do this for simple sets where we never iterate and where we only check and add single entries from different threads.
Uwe

PR #12806, opened 2023-11-14
Utilize exact kNN search when gathering k > numVectors in a segment

benwtrent:
When requesting for k >= numVectors, it doesn't make sense to go through the HNSW graph. Even without a user supplied filter, we should not explore the HNSW graph if it contains fewer than `k` vectors.

One scenario where we may still explore the graph if k >= numVectors is when not every document has a vector and there are deleted docs. But, this commit significantly improves things regardless.

Issue #12505, opened 2023-08-11
Re-explore the logic around when Vector search should be Exact


benwtrent:
### Description

Lucene always does an approximate nearest neighbors search when no filter is provided. 

This seems like unnecessary work. Some benchmarks would have to be done, but some ideas I had around options to explore:

 - Why not always do exact when `maxDoc < k`?
 - Should the "when to do exact" calculation consider `byte` vs `float` vectors?

It seems weird to go through all the work of going to the graph if there are only 10 documents in a segment.

benwtrent:
One thing to consider is that we should test some various graphs to see how many vectors we actually visit. 

I suspect its around `Math.log(graphSize) * vectorsCollected`. We could use some derivative of this equation to determine if we should use brute force or not.

For testing purposes, since we want to verify graph exploration, etc. all work without having to build huge indices, we might have an "override" to always do HNSW search first.

benwtrent:
As pointed out in other issue conversations, Cassandra keeps track of the visited ratio over the lifetime of the index and its searches: https://github.com/apache/cassandra/blob/2f2bb70ccb2657a75abc5aa691cfa28924f98d10/src/java/org/apache/cassandra/index/sai/disk/v1/segment/VectorIndexSegmentSearcher.java#L288

I am not sure if this is a thing we want to do in Lucene directly or not.

Another option is to keep what we are doing (drop out of graph search if we see we are worse than brute-force).

PR #13143, opened 2024-02-28
Remove unnecessary `AbstractKnnVectorQuery.exactSearch()`

mccullocht:
### Description

As of #12806 the hnsw codec has implemented a more complete version of this logic that may trigger without a pre-filter query. `exactSearch()` might also have some odd behavior surrounding scalar quantization -- IIUC any segment in
which we hit this path may score against the unquantized vectors which may then be mixed with quantized scored results.

This is related to #12505


kkmr (approved):

benwtrent:
@mccullocht How does this work against segments that were created before 9.9? I know the new reader will support this. But my concern is the SPI loading Lucene95HNSWReader and that code not handling the under sampling. 

Maybe we should add my change to the other backwards codec readers to make your change possible?

mccullocht:
@benwtrent Segments before 9.9 would perform a graph search. In 9c3679bf14b0d75cd58e1fda3ca4b8a76aa033b9 I back ported your change from #12806 to 9.1, 9.2, 9.4, and 9.5 since it wasn't too much effort.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -125,24 +124,11 @@ private TopDocs getLeafResults(
...
-
-    // Perform the approximate kNN search
-    // We pass cost + 1 here to account for the edge case when we explore exactly cost vectors
-    TopDocs results = approximateSearch(ctx, acceptDocs, cost + 1, knnCollectorManager);
-    if (results.totalHits.relation == TotalHits.Relation.EQUAL_TO) {
-      return results;
-    } else {
-      // We stopped the kNN search because it visited too many nodes, so fall back to exact search
-      return exactSearch(ctx, new BitSetIterator(acceptDocs, cost));
-    }
```
benwtrent, line 145: How do you handle the case when graph search early exits due to the visitation limit? Right now your change will return incomplete results to the user because we stopped exploring the graph.

benwtrent (changes requested):
I have read this more carefully. This change doesn't address the following edge case: when there are more than `k` docs in the filter set but the filter is restrictive enough that more than `cost + 1` nodes are visited.

I am not sure the codecs should handle this. The searcher asked for a visit limit to be set and the codec should respect it when possible and stop searching and return.

It seems the real fix is to have a better abstraction in the leaf reader. One for approximate search and the other for iterating the vectors & scoring based on the stored vector similarity function. 

This tangentially relates to the ideas around making vector similarity functions pluggable as the "exact search" use-case shouldn't even need to know the similarity function used by the codec.

mccullocht:
@benwtrent I see a couple of possibilities to internalize this in `KnnVectorsReader`:
* add a `public abstract void exhaustiveSearch(field, target, knnCollector, acceptDocs)` call that internalizes the exact search logic that lives in the query today.
* add a `public VectorScorer createVectorScorer(field, target)` call to internalize creation of the `VectorScorer` and handling of the similarity function but otherwise leave much of the logic the same.

benwtrent:
> add a public VectorScorer createVectorScorer(field, target) call to internalize creation of the VectorScorer and handling of the similarity function but otherwise leave much of the logic the same.

I personally prefer this. It should allow it to combine with other docset iterators. I am not 100% sure of the API/name/etc. But it would have to be added to the leafReader.

vigyasharma:
I was looking at this PR since it's marked as stale by our bots. I see that we've already added the [fallback to exact search](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java#L258-L265) when `k > maxOrd()`, and we also have the [createVectorScorer()](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L177) that uses the similarity configured in FieldInfo.

Do we plan to make any other changes as part of this PR? If not, we can close this one.

benwtrent:
I agree, I think it can be closed. We will still need the `exactSearch` method, even with refactorings that make it play well with quantization.

Issue #12968, opened 2023-12-24
[BUG] FSDirectory stuck at open(Path path) method when ran from .jar file


setokk:
### Description

Hello!

I use the FSDirectory.open(Path path) method when building the index.

When I run this method while **not** being in a jar, it works as intended.
But as soon as I package into a .jar file, it gets stuck at FSDirectory.open(Path path).

I outputted the absolute paths of the files (dataFile and idxDir) and they were correct.  

The weird thing is no exception is thrown. It just gets stuck. The method is called in the background from a SwingWorker but that shouldn't affect it really since it works well when I just run it from my IDE.

Not 100% sure if this is a bug but it seems weird that it just gets stuck and nothing is thrown while the file paths are correct.

Thanks for reading!

Here is my search engine init method:
```
public static void init(boolean clearIndex) {
        File dataFile = new File(System.getProperty("user.dir") + File.seperator + "data.txt");
        File idxDir = new File(System.getProperty("user.dir") + File.seperator + "index");

		// Set up analyzer
		SearchEngine.analyzer = new EnglishAnalyzer();

        try {
            System.out.println(idxDir.getAbsolutePath());
            System.out.println(dataFile.getAbsolutePath());
            SearchEngine.indexDir = FSDirectory.open(idxDir.toPath());

            if (clearIndex) {
                System.out.println("Rebuilding index...");

                IndexWriterConfig idxConfig = new IndexWriterConfig(SearchEngine.analyzer);
                idxConfig.setOpenMode(IndexWriterConfig.OpenMode.CREATE);
                IndexWriter idxWriter = new IndexWriter(indexDir, idxConfig);

                // Now, populate the index
                int docs = 0;
                JsonParser jParser = new JsonParser();

                for (String line : Files.readAllLines(dataFile.toPath(), StandardCharsets.UTF_8)) {
                    // On large amounts of data, this can take a while
                    if (docs % 10000 == 0) {
                        System.out.println(docs);
                    }
                    docs++;

                    // Parse JSON
                    // Each line of the input file is a serialized JSON object
                    Map j = jParser.parse(line);

                    // Get title
                    String title = (String) j.get("title");

                    // Get description
                    String ab = (String) j.get("abstract");

                    // Iterate through each author object and get "surname" and "given_names"
                    List<Map<String, String>> authors = (List<Map<String, String>>) j.get("authors");
                    StringBuilder authorsConcat = new StringBuilder();
                    String prefix = "";
[...]

PR #13148, opened 2024-02-29
Do not place Panama Java 21 class files in MR-JAR section of core.jar file

uschindler:
This is a followup after: https://github.com/apache/lucene/pull/13146#issuecomment-1971597031

This places the compilation unit for Java 21 MRJAR classes in the main section of JAR file. As we currentlly only have Java 21 classes, the MR-JAR file is then automatically disabled.

It is fully dynamic, so when we add later versions of vector code it will be placed in MRJAR sections again.

### DIFF
--- a/gradle/java/core-mrjar.gradle
+++ b/gradle/java/core-mrjar.gradle
@@ -46,15 +46,20 @@ configure(project(":lucene:core")) {
     }
     
     tasks.named('jar').configure {
+      boolean needMRJAR = false;
       mrjarJavaVersions.each { jdkVersion ->
-        into("META-INF/versions/${jdkVersion}") {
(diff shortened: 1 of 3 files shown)


dweiss:
Right. A bit of puzzling magic - I'm sure people will wonder what happens here but LGTM.

dweiss (approved):

uschindler:
The main reason why I like to do this it to make the har file more compatible with "Maven Shade plugin lovers", e.g., #12968.

The change makes at least mmap usable without correct development support for MRJARs.

uschindler:
> Right. A bit of puzzling magic - I'm sure people will wonder what happens here but LGTM.

I added a comment.

uschindler:
Thanks for fixing the typo.

dweiss:
No worries at all.

PR #13150, opened 2024-03-01
Make Hnsw99Codecs Extensible

Shibi-bala:
### Description
Lucene99HnswVectorsFormat and Lucene99HnswScalarQuantizedVectorsFormat are not extensible. With the new changes to how max vector dimension is set in the codec, users should be able to extend these classes to override the `getMaxDimensions` method. The current alternative is to make a delegating class that extends from KnnVectorsFormat which isn't the nicest UX.


uschindler:
Lucene prefers the delegation pattern. To prevent uncontrolled subclassing most of the components in Lucene are final.

PR #13155, opened 2024-03-04
hunspell: allow ignoring exceptions on duplicate ICONV/OCONV mappings

donnerpeter:


donnerpeter:
To backport to Lucene 9_x

dweiss:
LGTM. A changes.txt entry is missing. Did you mean to assign this to me or just assign me as a reviewer? :)

donnerpeter:
> LGTM. A changes.txt entry is missing. Did you mean to assign this to me or just assign me as a reviewer? :)

Oh sorry. I meant you as a reviewer :)
Is the changes.txt entry needed if I backport this to 9_x (and add the entry there)? AFAIU the changes for 10.0 shouldn't include stuff that's in 9_x? Or am I mistaken?

dweiss:
They do go to main with changes under the corresponding 9.x section - then you can typically cherry pick the same commit when backporting.

dweiss:
So, on main they'd go somewhere below this place:
https://github.com/apache/lucene/blob/main/lucene/CHANGES.txt#L184

dweiss (approved):

rmuir:
Does it make sense to allow this by default? I don't know how `hunspell` behaves here, but I've seen crazy stuff in these dictionaries...

donnerpeter:
> Does it make sense to allow this by default? I don't know how `hunspell` behaves here, but I've seen crazy stuff in these dictionaries...

@rmuir An interesting idea. Such duplication appears to be a relatively rare condition, not present in the two major dictionary collections I've been testing against. On the other hand, people would need to recompile their code if they find they face such dictionaries. On yet another hand, people might be more motivated to fix the faulty dictionaries. I'm unsure :(

PR #13159, opened 2024-03-06
Rename TestWildcard to TestWildcardQuery.

vsop-479:
Similar to `TestPrefixQuery` and `TestRegexpQuery`, this class tests `WildcardQuery`.


rmuir (approved):

vsop-479:
@rmuir 
If there is no problem with this change, would you like to merge it?

PR #13163, opened 2024-03-06
Fix demo application, if no knn dict was provided and over 100 docs are indexed

### LINKED ISSUE #12330: Demo application throws an exception when following documentation.
### Description

If you clone the lucene core repo and follow the instruction at https://lucene.apache.org/core/9_6_0/demo/index.html to run the demo app, you get the following exception:

```
Exception in thread "main" java.lang.RuntimeException: Are you (ab)using the toy vector dictionary? See the package javadocs to understand why you got this exception.
        at org.apache.lucene.demo.IndexFiles.main(IndexFiles.java:181)
```

Looking at the code, this will happen any time you don't pass `-knn_dict` as `vectorDictSize` will always be zero in that case.

_No response_

timgrein:
### Description

If you index over 100 documents with the demo application without providing a KnnVector dictionary via the `-knn-dict` parameter you'll run always into this [code path](https://github.com/apache/lucene/blob/49db29e59f1d9c7de33bc09d42334a6afab943ac/lucene/demo/src/java/org/apache/lucene/demo/IndexFiles.java#L178-L183), which throws an exception. This PR extends the condition to only be executed, if an actual KnnVector dictionary was provided.

Note: As `test-files/docs` only contains 16 documents I hesitated to introduce additional 85 docs just for testing this specific functionality of the demo application. Happy to add it anyway, if needed. We could also parameterise the thresholds via an additional constructor, but again don't know, how extensively this should be tested.

Closes https://github.com/apache/lucene/issues/12330

### DIFF
--- a/lucene/demo/src/java/org/apache/lucene/demo/IndexFiles.java
+++ b/lucene/demo/src/java/org/apache/lucene/demo/IndexFiles.java
@@ -28,6 +28,7 @@ import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.Date;
+import java.util.Objects;
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.analysis.standard.StandardAnalyzer;
 import org.apache.lucene.demo.knn.DemoEmbeddings;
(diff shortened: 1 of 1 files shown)


msokolov (approved):
Thank you @timgrein! sorry this moldered for a while, looks good, I'll merge

msokolov:
I also cherry-picked to branch_9x: 3d109bdc493b451f967a7ad3102267c9d177ade0

PR #13164, opened 2024-03-06
An eye-gouging way to limit suppressAccessChecks to just the three JARs that need them.

dweiss:


rmuir:
Had to look given the PR title! This doesn't seem so bad?

One benefit of limiting the permission: tests can prevent any more uses from slipping in (including third party dependencies)

gradle/testing/randomization/policies/replicator-tests.policy
```diff
@@ -92,3 +90,14 @@ grant {
...
 };
+
+// Grant reflection permissions to certain JARs (https://github.com/apache/lucene/pull/13146)
+grant codeBase "file:${randomizedtesting.jar}" {
+  permission java.lang.reflect.ReflectPermission "suppressAccessChecks";
+};
+grant codeBase "file:${junit.jar}" {
+  permission java.lang.reflect.ReflectPermission "suppressAccessChecks";
+};
+grant codeBase "file:${lucene.test.framework}${/}-" {
```
uschindler, line 101: Maybe add the .jar suffix to property name to make it consistent.

rmuir:
Another possible improvement as a followup, I can try to find time to look in, i'm not sure replicator-tests.policy needs to exist anymore. A lot of the permissions there are outdated since https://github.com/apache/lucene/pull/12038

In fact tests may pass now with just the standard policy! Sorry, I left this mess and didn't properly clean it up.

uschindler (approved):

rmuir:
I tried it out, i don't want to conflict with your changes here, but if we add this to main policy:
```
--- a/gradle/testing/randomization/policies/tests.policy
+++ b/gradle/testing/randomization/policies/tests.policy
@@ -61,6 +61,8 @@ grant {

   // TestLockFactoriesMultiJVM opens a random port on 127.0.0.1 (port 0 = ephemeral port range):
   permission java.net.SocketPermission "127.0.0.1:0", "accept,listen,resolve";
+  // Replicator tests connect to ephemeral ports
+  permission java.net.SocketPermission "127.0.0.1:1024-", "connect,resolve";

   // read access to all system properties:
   permission java.util.PropertyPermission "*", "read";
```

Then we can remove the entire `replicator-tests.policy` and its associated if-then-else in `randomization.gradle` and just have one of these files. replicator tests must be run with `-Dtests.nightly` as a lot of them are monstrous, but it seems to work. so it might make it less eye-gouging :)

gradle/testing/randomization/policies/replicator-tests.policy
dweiss: It's not just the jar - it's everything under the build folder... extracting the jar name from another subproject is problematic, also because the test framework itself uses Test tasks. I thought this was the easiest way to accomplish the goal - but I admit it also contributed to the name of the PR...

dweiss:
Please feel free to commit to this branch, Robert - no problem at all. I also noticed some of the stuff in policy files is outdated or should be reviewed - like these properties:
```
// Permissions to support ant build
grant {
  permission java.io.FilePermission "${user.home}${/}.ivy2${/}cache${/}-", "read";
  permission java.io.FilePermission "${junit4.tempDir}${/}*", "read,write,delete";
  permission java.io.FilePermission "${clover.db.dir}${/}-", "read,write,delete";
};
```
Just didn't have the time to iterate on these.

gradle/testing/randomization.gradle
```diff
@@ -119,6 +119,18 @@ allprojects {
...
+    }
+    dependencies {
+      secManagerExclusions ( "com.carrotsearch.randomizedtesting:randomizedtesting-runner", {
+          exclude group: "junit"
+      })
+      secManagerExclusions ( "junit:junit", {
+        exclude group: "org.hamcrest"
+      })
+    }
+
```
dweiss, line 133: A short explanation why this block is here at all - it's simpler to just use another configuration to pin the location of those dependencies than extract them from actual test class's classpath arguments (we use modular path extensions, etc.).

The problem is that these locations can be resolved only after the configuration is done (gradle's constraint) and at the same time systemProperty doesn't allow lazy providers so we need to push this into SecurityArgumentProvider. 

I think this could (and should) be simpler but at the moment it's the simplest what I could come up with.

rmuir:
OK I pushed a commit removing that replicator policy file... to ease our burden...

gradle/testing/randomization.gradle
```diff
@@ -189,6 +201,14 @@ allprojects {
           systemProperty 'gradle.lib.dir', Paths.get(project.class.location.toURI()).parent.toAbsolutePath().toString().replace('\\', '/')
           systemProperty 'gradle.worker.jar', Paths.get("${gradleUserHome}/caches/${gradle.gradleVersion}/workerMain/gradle-worker.jar").toAbsolutePath().toString()
           systemProperty 'gradle.user.home', gradleUserHome.toPath().toAbsolutePath().toString()
+
+          securityArgumentProvider.otherProperties.put("randomizedtesting.jar", project.provider { ->
+            return configurations.secManagerExclusions.resolve().find { it.name.startsWith("randomizedtesting-runner-") }.absolutePath
```
uschindler, line 206: Should this not use `.toString().replace('\\', '/')` like seen a few lines above? Haven't tested with windows yet.

gradle/testing/randomization/policies/replicator-tests.policy
uschindler: You're right!

gradle/testing/randomization.gradle
uschindler: Hm, i think native path is correct here. I don't userstand the above gradle chaos.

Maybe we should consistently use `${/}`

rmuir:
ok, sorry for the heavy committing, if one fails lets just revert it, but I tried to do some cleanup here.

gradle/testing/randomization.gradle
uschindler: Do we need the extra config at all. As we parse through the dependencies below like `return configurations.secManagerExclusions.resolve().find { it.name.startsWith("randomizedtesting-runner-") }.absolutePath`, why not use the test configuration from beginning?

rmuir:
To better improve the crushing tests, instead of the child process killing itself with unsafe/halt, we can have the parent kill it with Process.destroyForcibly() and such.

This is even better than Unsafe, in *nix for example it sends `kill -9` and kernel really crashes it for sure: the jvm can't catch this, can't dump `hs_err` or anything else. I will try to look into this tomorrow.

gradle/testing/randomization.gradle
dweiss: I tested with Windows, although without odd paths with spaces, etc... I believe property expansion is guaranteed to work properly in policy files but I can't find a spec reference for this. I'm not sure why these other properties use the complex logic to convert paths.

gradle/testing/randomization.gradle
dweiss: I think it's simpler this way - we resolve classpath/ module paths independently, with different source sets... it's quite complicated. These properties are only needed for two dependencies and we know their versions will be consistent with whatever shows up on classpath/ module path so I didn't bother trying to figure out how to do this in a generic way.

dweiss:
Thanks for the follow-ups, @rmuir! It's a great idea to kill the subprocess from the outside - I'll take a look at that (but not until later, I'm busy throughout the day).

dweiss:
Ok, it wasn't too difficult - I've removed the halt() call and I think it works.

uschindler:
Yes, the ProcessBuilder API can also kill forcibly. Nice idea. Very clean.

Uwe

lucene/replicator/src/test/org/apache/lucene/replicator/nrt/TestSimpleServer.java
```diff
@@ -135,39 +132,12 @@ public void run() {
...
-      // Fallback attempt to Runtime.halt();
-      Runtime.getRuntime().halt(-1);
-    } catch (Exception e) {
-      System.out.println("Couldn't kill the JVM.");
-      e.printStackTrace(System.out);
-    }
-
-    // We couldn't get the JVM to crash for some reason.
-    throw new RuntimeException("JVM refuses to die!");
+    Runtime.getRuntime().halt(1);
```
uschindler, line 140: How about this one?

dweiss: I missed it. I'll do it in the evening.

rmuir:
nice!

lucene/replicator/src/test/org/apache/lucene/replicator/nrt/TestSimpleServer.java
dweiss: There's this one and there's a halt() call in TestCodecLoadingDeadlock. The TestCodecLoadingDeadlock seems reasonable since it tries to exit the forked JVM after dumping stacks. The logic in TestSimpleServer is quite complicated (and the forking code too). I think we should leave it for a separate patch?

dweiss:
I've merged this with main and I'll follow up with the remaining halt methods in a separate PR. Thanks for your help with this!

uschindler:
I can confirm all works well with Windows and whitespace in pathnames. Test counting works and all tests pass.

uschindler:
I think we still have a problem with ECL Emma Coverage: https://ci-builds.apache.org/job/Lucene/job/Lucene-Coverage-main/1063/console

Looks like EMMA wants to do setAccessible, too. Will figure out later today.

dweiss:
Maybe we shouldn't bother and run coverage without the security manager?

dweiss:
I've turned off security manager for these coverage runs. Jacoco jar/ options are resolved/ added by plugins and it's hard to wrap one's head around it (where does the jar come from, what version it is, etc.)

image

uschindler:
Can't we do that in the gradle magic automatically?

uschindler:
I mean when you enable coverage or start the virtual "coverage task", gradle should automatically set security manager to off. When I start coverage run locally it breaks, too, so it would be good to have this automatically.

dweiss:
Probably possible - not sure whether increasing the build logic is worth it. Maybe it (coverage task) should just fail if it detects conflicting options?

uschindler:
I will check a bit later, sorry time is limited.

uschindler:
P.S.: If we remove support for SecurityManager in combination with EMMA, we should remove the already existing policy file entries that are related to EMMA. I think there are a few. Maybe we can also extend them to allow access override?

dweiss:
I'm for simplifying it, if possible (so removing unnecessary stuff). I don't think security manager coverage is that helpful, especially that security manager is going away in the future anyway.

PR #13165, opened 2024-03-06
Made the UnifiedHighlighter's hasUnrecognizedQuery function processes FunctionQuery the same way as MatchAllDocsQuery and MatchNoDocsQuery queries for performance reasons.

vletard:
### Description

This pull request is a follow up to #12938. I took on the task of complying with @romseygeek's comment on behalf of [Lexum](https://lexum.com/en/).
I also rebased to the latest commit of `apache:main`.

Let me know if any other adjustment is needed.

### DIFF
--- a/lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/UnifiedHighlighter.java
+++ b/lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/UnifiedHighlighter.java
@@ -100,11 +100,8 @@ public class UnifiedHighlighter {
 
   public static final int DEFAULT_CACHE_CHARS_THRESHOLD = 524288; // ~ 1 MB (2 byte chars)
 
-  public static final Set<Class<? extends Query>> QUERIES_WITH_NO_HL_EFFECT = Set.of(
-    MatchAllDocsQuery.class,
-    MatchNoDocsQuery.class,
-    FunctionQuery.class
-  );
+  public static final Set<Class<? extends Query>> QUERIES_WITH_NO_HL_EFFECT =
+      Set.of(MatchAllDocsQuery.class, MatchNoDocsQuery.class, FunctionQuery.class);
 
   protected static final LabelledCharArrayMatcher[] ZERO_LEN_AUTOMATA_ARRAY =
       new LabelledCharArrayMatcher[0];
@@ -1138,7 +1135,8 @@ public class UnifiedHighlighter {
           public void visitLeaf(Query query) {
             if (MultiTermHighlighting.canExtractAutomataFromLeafQuery(query) == false) {
               boolean no_effect_query = false;
-              for (Class<? extends Query> queryType: UnifiedHighlighter.QUERIES_WITH_NO_HL_EFFECT) {
+              for (Class<? extends Query> queryType :
+                  UnifiedHighlighter.QUERIES_WITH_NO_HL_EFFECT) {
                 if (queryType.isInstance(query)) {
                   no_effect_query = true;
                   break;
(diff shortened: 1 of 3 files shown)


mkhludnev (approved):

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/UnifiedHighlighter.java
```diff
@@ -1130,7 +1134,16 @@ public boolean acceptField(String field) {
           @Override
           public void visitLeaf(Query query) {
             if (MultiTermHighlighting.canExtractAutomataFromLeafQuery(query) == false) {
-              if (!(query instanceof MatchAllDocsQuery || query instanceof MatchNoDocsQuery)) {
+              boolean no_effect_query = false;
```
romseygeek, line 1137: This would be simpler as `QUERIES_WITH_NO_HL_EFFECT.contains(query.getClass())`

vletard: Would you say that there is no need to capture hypothetical subclasses?

romseygeek: `MatchAllDocsQuery` is already final, but the other two aren't.  Maybe this would be better done as a new protected method, `isIgnorableQuery(Query q)` with a default implementation of `instanceof` checks, and then if somebody has their own extension (or another query implementation that will do weird things here) they can override the method?

vletard: I just edited my branch with your suggestion.

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/UnifiedHighlighter.java
```diff
@@ -1130,7 +1137,7 @@ public boolean acceptField(String field) {
           @Override
           public void visitLeaf(Query query) {
             if (MultiTermHighlighting.canExtractAutomataFromLeafQuery(query) == false) {
-              if (!(query instanceof MatchAllDocsQuery || query instanceof MatchNoDocsQuery)) {
+              if (!UnifiedHighlighter.this.isIgnorableQuery(query)) {
```
romseygeek, line 1140: Can you add a test to TestUnifiedHighlighterExtensibility that checks that overrides to `isIgnorableQuery` work here as well?  I think it will, but to be perfectly honest I can never remember the inheritance rules around class-specific `this` calls...

romseygeek (changes requested):

romseygeek:
Thanks @vletard!  This looks great, I'd just like to add one more test to ensure that inheritance works in the way we expect.

vletard:
I added a test in TestUnifiedHighlighterExtensibility, probably too simple for what you asked though.
From my understanding, performing a more realistic test with adequate UH and query setup would require one of the following:
- cross-package import some code from TestUnifiedHighlighter to TestUnifiedHighlighterExtensibility
- copy this code instead of importing it (we probably want to avoid this as it would duplicate much of the randomized generation process)
- create an extensibility test in the `uhighlight` package instead of `uhighlight.visibility`

Options 1 and 3 that I tend to prefer would require not-so-light edits, so I'd like your opinion.
What is the expected way to do this in Lucene unit testing?

Thanks!

vletard:
@romseygeek, could you write what you would like to see added more precisely? See the three options I listed in my earlier comment.
Thank you!

PR #13171, opened 2024-03-09
Reduce some unnecessary ArrayUtil#grow calls

easyice:
This change will slightly improve the performance of `DataOutput#writeGroupVInts` and some other methods related to `XXXRefBuilder#append`.

Here is a JMH benchmark for `Util.toIntsRef`, using java 21 on my MAC (intel chip).

```
Benchmark                    Mode  Cnt   Score   Error   Units
PR   ToIntsRefBenchMark.run  thrpt    5  38.208 ± 1.054  ops/us
main ToIntsRefBenchMark.run  thrpt    5   7.862 ± 0.832  ops/us
```

JMH benchmark code for `Util#toIntsRef`:

```java
public class ToIntsRefBenchMark {
  IntsRefBuilder intsRefBuilder = new IntsRefBuilder();
  BytesRef bytesRef = new BytesRef(new byte[32]);

  @Benchmark
  public void run() {
    Util.toIntsRef(bytesRef, intsRefBuilder);
  }
}
```

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/BytesRefBuilder.java
+++ b/lucene/core/src/java/org/apache/lucene/util/BytesRefBuilder.java
@@ -57,7 +57,9 @@ public class BytesRefBuilder {
 
   /** Ensure that this builder can hold at least <code>capacity</code> bytes without resizing. */
   public void grow(int capacity) {
-    ref.bytes = ArrayUtil.grow(ref.bytes, capacity);
+    if (ref.bytes.length < capacity) {
+      ref.bytes = ArrayUtil.grow(ref.bytes, capacity);
+    }
   }
 
   /**
--- a/lucene/core/src/java/org/apache/lucene/util/IntsRefBuilder.java
+++ b/lucene/core/src/java/org/apache/lucene/util/IntsRefBuilder.java
@@ -74,7 +74,9 @@ public class IntsRefBuilder {
    * @lucene.internal
    */
   public void grow(int newLength) {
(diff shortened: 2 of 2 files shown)


stefanvodita:
@easyice - I'm probably missing something obvious, but I thought `ArrayUtil.grow` already made the same check. Can you explain why this PR helps?
https://github.com/apache/lucene/blob/04f335ad79cae019d1a552e41e0029d061d8e3d9/lucene/core/src/java/org/apache/lucene/util/ArrayUtil.java#L430-L435

easyice:
@stefanvodita Thank you for looking into this!  when an array does not need to grow, The `ArrayUtil.grow` will return the input `array`, then it will call an extra assignment like `ref.bytes = ref.bytes` in `BytesRefBuilder#grow`. I observed there was some performance impact, which was noticeable in the `append` method. Maybe i'm being overly picky, Overall, it doesn't have a significant impact on performance.

stefanvodita:
Thanks for the explanation! The way I think about it is whether it would be ok to make this check every time we call `grow`. Probably not, it's nice that `grow` does the check for you. Maybe simplicity is worth more than the performance gain here? If you think the performance win is compelling, I won't argue against it.

easyice:
I originally wanted to speed up `writeGroupVInts` in this way. Now there is a better way to speed up it (by https://github.com/apache/lucene/pull/13203) . I agree with you that only make the change where performance is important to keep things simple, Currently, I haven't seen any other places yet, so i will close it. Thank you!

PR #13177, opened 2024-03-10
Remove halt() call in TestSimpleServer (part of TestStressNRTReplication

dweiss:
A follow-up to #13164 . I don't fully understand this code (I get the gist) so I didn't go for any revolutionary changes - just removed the halt() call in favor of an external process killer. I've re-ran the test 10 times. It is currently only run in nightly mode (takes ages) but you can run it in non-nightly mode with:
```
gradlew -p lucene\replicator test --tests TestStressNRTReplication  -Ptests.filter="@Nightly" 
```
because the tests.filter doesn't affect the "nightly" system property (so test parameters are not scaled to nightly standards but still run).


lucene/replicator/src/test/org/apache/lucene/replicator/nrt/TestStressNRTReplication.java
```diff
@@ -256,7 +257,7 @@ public void test() throws Exception {
     if (TEST_NIGHTLY) {
       runTimeSec = RANDOM_MULTIPLIER * TestUtil.nextInt(random(), 120, 240);
     } else {
-      runTimeSec = RANDOM_MULTIPLIER * TestUtil.nextInt(random(), 45, 120);
+      runTimeSec = RANDOM_MULTIPLIER * TestUtil.nextInt(random(), 20, 60);
```
dweiss, line 260: I lowered these a bit intentionally.

rmuir:
Thanks for doing this. I was disappointed that java doesn't allow this:
```
jshell> ProcessHandle.current().destroyForcibly()
|  Exception java.lang.IllegalStateException: destroy of current process not allowed
|        at ProcessHandleImpl.destroyProcess (ProcessHandleImpl.java:369)
|        at ProcessHandleImpl.destroyForcibly (ProcessHandleImpl.java:392)
|        at (#2:1)
```

dweiss:
I think it's in the method's documentation (current()) that the returned handle can't be used to stop yourself. Indeed, I tried it too. ;)

PR #13178, opened 2024-03-12
Support disabling IndexSearcher.maxClauseCount with a value of -1

dsmiley:
This PR allows `IndexSearcher.maxClauseCount` to be -1 with the meaning of there being no limit.  This avoids pointless query tree visiting. 
Tangential:`TermInSetQuery.visit()` is disappointing.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
+++ b/lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
@@ -99,5 +99,11 @@ public abstract class QueryVisitor {
   }
 
   /** A QueryVisitor implementation that does nothing */
-  public static final QueryVisitor EMPTY_VISITOR = new QueryVisitor() {};
+  public static final QueryVisitor EMPTY_VISITOR =
+      new QueryVisitor() {
+        @Override
+        public boolean acceptField(String field) {
+          return false;
+        }
+      };
 }
--- a/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
+++ b/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
@@ -733,7 +733,9 @@ public class IndexSearcher {
         rewrittenQuery = query.rewrite(this)) {
       query = rewrittenQuery;
     }
-    query.visit(getNumClausesCheckVisitor());
+    if (maxClauseCount > -1) {
+      query.visit(getNumClausesCheckVisitor());
+    }
     return query;
   }
 


lucene/core/src/java/org/apache/lucene/search/QueryVisitor.java
```diff
@@ -99,5 +99,11 @@ public void consumeTerms(Query query, Term... terms) {
   }
 
   /** A QueryVisitor implementation that does nothing */
-  public static final QueryVisitor EMPTY_VISITOR = new QueryVisitor() {};
+  public static final QueryVisitor EMPTY_VISITOR =
```
dsmiley, line 102: I'll create a separate PR for this; didn't mean to include it.

dsmiley (commented):
* needs a test to show -1 is a valid value; I think that would be sufficient
* needs javadocs

Looking for feedback!

jpountz:
I'm curious what problem you are trying to address. It looks like you're trying to avoid the overhead of checking the number of clauses, but intuitively this wouldn't help much as we have other operations that need to run in linear time with the number of clauses in the query anyway such as creating the query in the first place, rewriting it, creating a weight, etc.

dsmiley:
Problem: clarity in being able to turn it off; a -1 value would clarify the configuration intent.  It has become even less useful, not that I quite wanted it in the first place in the past either.  In the production clusters I work with, there are other limits (query parsing level enforced) and they are adequate.  I could use a bunch of 9's and add a comment in the configuration but shouldn't Lucene let you turn it off?  Why argue against a simple `if` condition for a check the user doesn't want to enforce.  I understand your point that it probably costs pathetically little compared to executing the query anyway.  However note the sad `TermInSetQuery.visit()` implementation.

Additionally/alternatively, making getNumClausesCheckVisitor protected and non-static could be useful.

Additionally/alternatively, remove this limit.  Keep a sample QueryVisitor impl with instructional javadocs to show how trivial it is for a user to layer this on if they so choose (subclass `rewrite`)

dweiss:
I just ran into this issue. Do you think we could revisit this and maybe merge it in? I've hit it with a large query consisting of multiple intervals - there are no "clauses" as such and it's quite confusing.

dsmiley:
Happy to do so.

dweiss:
An alternative could be to implement this logic in the default visitor returned from getNumClausesCheckVisitor(). This way, if somebody overrides this method, it'll still work for them after the patch is applied. 

I also realized that this could now be implemented downstream (in Solr, for example) as an override of this method. Don't know what's better.

PR #13181, opened 2024-03-13
Add new VectorScorer interface to vector value iterators

benwtrent:
With quantized vectors, and with current vectors, we separate out the "scoring" vs. "iteration", requiring the user to always iterate the raw vectors and provide their own similarity function.

While this is flexible, it creates frustration in:

 - Just iterating and scoring, especially since the field already has a similarity function stored...Why can't we just know which one to use and use it!
 - Iterating and scoring quantized vectors. By default it would be good to be able to iterate and score quantized vectors (e.g. without going through the HNSW graph).

This significantly hampers support for true exact kNN search.

I see two options on providing this:

 - A new top level thing on the LeafReader (getVectorScorer or something).
 - Extend the vector value iterators to be able to return a scorer given some vector value (what this PR demonstrates).

Folks who might be interested: @jpountz @msokolov @mccullocht


lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -190,12 +190,13 @@ protected TopDocs exactSearch(LeafReaderContext context, DocIdSetIterator accept
...
     }
+    DocIdSetIterator vectorIterator = vectorScorer.iterator();
     HitQueue queue = new HitQueue(k, true);
     ScoreDoc topDoc = queue.top();
     int doc;
     while ((doc = acceptIterator.nextDoc()) != DocIdSetIterator.NO_MORE_DOCS) {
-      boolean advanced = vectorScorer.advanceExact(doc);
-      assert advanced;
+      vectorIterator.advance(doc);
+      assert vectorIterator.docID() == doc;
```
mccullocht, line 199: skip the document if it doesn't match? i'm guessing it would be an easily predicted branch and it's possible that the filter query could yield a document that doesn't have a vector.

mccullocht (commented):
Overall this lgtm, just one small thing.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
benwtrent: The iterator already does a filter on the field existing. But I agree, this interaction needs improved. Especially as adding the filter doesn't really seem necessary and could unnecessarily thrash any filter clause caching that exists. 

The iterator query that is actually created upstream of this code:
```java
    if (filter != null) {
      BooleanQuery booleanQuery =
          new BooleanQuery.Builder()
              .add(filter, BooleanClause.Occur.FILTER)
              .add(new FieldExistsQuery(field), BooleanClause.Occur.FILTER)
              .build();
      Query rewritten = indexSearcher.rewrite(booleanQuery);
      filterWeight = indexSearcher.createWeight(rewritten, ScoreMode.COMPLETE_NO_SCORES, 1f);
    } else {
      filterWeight = null;
    }
```

benwtrent:
This is my first attempt at a new interface for exact search.

We are missing performance improvements through quantization and have a fairly complex API for exact search, requiring users to iterate the vectors themselves and choose a correct vector similarity.

This new abstraction allows a simple iterator, scorer, API for dense vector fields.

However, it's sort of inverted from the typical API (scorers create iterators, but here iterators provide scorers).

@msokolov what do you think?

The other option I can easily see is adding another top level thing to the leaf reader that provides an iterative vector scorer.

lucene/core/src/java/org/apache/lucene/search/VectorScorer.java
```diff
@@ -18,64 +18,39 @@
...
+  float score() throws IOException;
 
-  /** Compute the similarity score for the current document. */
-  abstract float score() throws IOException;
-
-  abstract boolean advanceExact(int doc) throws IOException;
+  /**
+   * @return a {@link DocIdSetIterator} over the documents.
+   */
+  DocIdSetIterator iterator();
```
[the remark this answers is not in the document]
benwtrent: @msokolov 

> since presumably the caller already has an iterator over docs with vectors in this field?

I kept it as its nice sugar so that things can rely on `VectorScorer` and not our typed `Float|ByteVectorValues`. An example of how this is a nice interface is: `DiversifyingChildrenVectorScorer`, as we can iterate and score either bytes or floats without caring the types. We just have a scorer & iterator now.

lucene/core/src/java/org/apache/lucene/util/quantization/QuantizedByteVectorValues.java
```diff
@@ -18,13 +18,40 @@
...
+import org.apache.lucene.search.VectorScorer;
 
 /**
  * A version of {@link ByteVectorValues}, but additionally retrieving score correction offset for
  * Scalar quantization scores.
  *
  * @lucene.experimental
  */
-public abstract class QuantizedByteVectorValues extends ByteVectorValues {
+public abstract class QuantizedByteVectorValues extends DocIdSetIterator {
```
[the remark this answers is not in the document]
benwtrent: > it's part of FieldInfo right?

Its part of the Codec.

> Or ... are we trying to allow two-pass scoring where we coarsely use quantized score and then later refine with full-precision scoring?

We already allow this. Approximate search is done over quantized and then folks can rescore later by iterating the raw floats and scoring however they want (un-quantized). 

>  One use case I'm not sure we support (and we should) is a vector field with no associated HNSW graph where we want to rank documents by their vector score "brute force".

This is the sort of use-case I have in mind here as well. Somebody creating their own "Flat" codec is pretty simple given we have separated out the flat vector storage & quantization from the HNSW graph. But, we don't provide a nice way for them to handle exact search. Instead they must override the approximate search API and do an a "top-k" over flat vectors. This requires iterating everything upfront and finding top-k. Instead, they should be able to return a scorer/iterator that can be used in a query.

The other use case is when we do exact search instead of approximate search due to filtering criteria. Right now we drop down to score the raw vectors instead of the quantized ones, that doesn't make sense to me. 

If somebody wants to rescore or search the raw vectors, they can already do that by iterating the `Float|ByteVectorValues` that they provided.

lucene/core/src/java/org/apache/lucene/search/VectorScorer.java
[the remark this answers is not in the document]
benwtrent: I have been thinking more. 

I think it would be good for the VectorScorer to use a "copy" of the iterator. This way the ONLY way to iterate is the iterator returned by the scorer.

Allowing both the VectorValues and a iterator returned from the scorer refer to the same internal iterator seems trappy.

Requiring iteration with the VectorScorer#iterator seems more natural and safer.

I will update the PR accordingly soon.

lucene/core/src/java/org/apache/lucene/index/ByteVectorValues.java
```diff
@@ -75,4 +76,14 @@ public static void checkField(LeafReader in, String field) {
...
+
+  /**
+   * Return a {@link VectorScorer} for the given query vector. The iterator for the scorer is not
+   * the same instance as the iterator for this {@link ByteVectorValues}. It is a copy, and
+   * iteration over the scorer will not affect the iteration of this {@link ByteVectorValues}.
+   *
+   * @param query the query vector
+   * @return a {@link VectorScorer} instance
+   */
+  public abstract VectorScorer scorer(byte[] query) throws IOException;
```
ChrisHegarty, line 88: More of a question than a comment. Is the similarity available in the context where the scorer is created? Could it be passed here to avoid tightly coupling the values and similarity (in the VectorValues interface). That way the same vector values source could be used for different similarities.

benwtrent: >  Is the similarity available in the context where the scorer is created? Could it be passed here to avoid tightly coupling the values and similarity (in this interface). That way the same vector values source could be used for different similarities.

🤔 I will have to think on this. We can maybe allow it, but allow default for just using whats stored in the codec.

One concern would be quantized similarities. For example, if the user initially indexed quantized vectors saying they would use `euclidean`, we don't calculate the error correction, its just `1`.

If they then supply `DOT_PRODUCT`, we don't have the error corrections calculated. Maybe in this case we just error?

msokolov: Is this something we could easily add later if it's needed?

benwtrent: Yes, I think so. These in general are still experimental APIs, so extending the method definition in the future could occur.

ChrisHegarty: Ok, ignore my comment. I was mostly trying to decouple the values from the similarities (as it has been up until this PR), and also forgot about quantization.

ChrisHegarty (approved):
LGTM

PR #13183, opened 2024-03-14
Fix TestIndexWriter.testDeleteUnusedFiles' failure on Windows 11

### LINKED ISSUE #12524: Test failure in TestIndexWriter.testDeleteUnusedFiles on Windows 11
### Description

TestIndexWriter.testDeleteUnusedFiles asserts file exists, after the commit follow the force merge on Windows.
Actually it is deleted after that commit. So, I got a assertion error on Windows.

OS details: Windows 11 10.0 amd64/Oracle Corporation 17.0.8 (64-bit).

gradlew test --tests TestIndexWriter.testDeleteUnusedFiles -Dtests.seed=1E21BC5E99CF60B9 -Dtests.locale=si-LK -Dtests.timezone=Indian/Chagos -Dtests.asserts=true -Dtests.file.encoding=GBK

vsop-479:
Fix https://github.com/apache/lucene/issues/12524

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/Constants.java
+++ b/lucene/core/src/java/org/apache/lucene/util/Constants.java
@@ -41,6 +41,9 @@ public final class Constants {
   /** True iff running on Windows. */
   public static final boolean WINDOWS = OS_NAME.startsWith("Windows");
 
+  /** True iff running on Windows and the version is on or after Windows 11. */
+  public static final boolean ON_OR_AFTER_WINDOWS_11 = onOrAfterWindows(11);
+
   /** True iff running on SunOS. */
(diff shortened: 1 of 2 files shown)


vsop-479:
@vigyasharma Please take a look when you get a chance.

vigyasharma:
Thanks for persisting on it @vsop-479. 

I think we should dig more into what has changed in Windows 11+ before we change the assert condition on this test. Does it impact any other parts of Lucene? Is it some API behavior change that will persist in subsequent versions as well? Please add if you have any more context on these.

vsop-479:
You are right @vigyasharma. I will dig into them.

PR #13184, opened 2024-03-14
Make the HitQueue size more appropriate for KNN exact search

bugmakerrrrrr:
### Description
Currently, when performing KNN exact search, we consistently set the HitQueue size to `k`. However, there may be instances where the number of candidates is actually lower than `k`.

### DIFF
--- a/lucene/join/src/java/org/apache/lucene/search/join/DiversifyingChildrenFloatKnnVectorQuery.java
+++ b/lucene/join/src/java/org/apache/lucene/search/join/DiversifyingChildrenFloatKnnVectorQuery.java
@@ -98,7 +98,8 @@ public class DiversifyingChildrenFloatKnnVectorQuery extends KnnFloatVectorQuery
             parentBitSet,
             query,
             fi.getVectorSimilarityFunction());
-    HitQueue queue = new HitQueue(k, true);
+    final int queueSize = Math.min(k, Math.toIntExact(acceptIterator.cost()));
+    HitQueue queue = new HitQueue(queueSize, true);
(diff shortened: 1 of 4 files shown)


lucene/join/src/java/org/apache/lucene/search/join/DiversifyingChildrenFloatKnnVectorQuery.java
```diff
@@ -98,7 +98,8 @@ protected TopDocs exactSearch(LeafReaderContext context, DocIdSetIterator accept
             parentBitSet,
             query,
             fi.getVectorSimilarityFunction());
-    HitQueue queue = new HitQueue(k, true);
+    final int queueSize = Math.min(k, Math.toIntExact(acceptIterator.cost()));
```
benwtrent, line 101: I don't know a better way, but, since this diversifies over parent doc ids, its possible that the hitqueue is still much smaller than `acceptIterator.cost()` as `acceptIterator.cost()` is the iterator over CHILD docs (e.g. passage vector docs). I think any further optimization (e.g. counting the number of relevant parents) would add undo overhead.

benwtrent (approved):
I think this is a sane optimization. Mind adding a CHANGES entry under optimizations?

lucene/join/src/java/org/apache/lucene/search/join/DiversifyingChildrenFloatKnnVectorQuery.java
bugmakerrrrrr: +1

PR #13186, opened 2024-03-14
Doc reordering avoid iterations: cooling using simulated annealing

rishabhmaurya:
### Description
As described in the paper (Tradeoff Options for Bipartite Graph Partitioning), simulated annealing-type mechanism can be employed to reduce number of swaps with each iteration. If projected advantage swapping 2 docs across partitions isn't as appealing, in this case less than `iter` bits, then swap can be avoided. 
Swapping documents is a heavy operation as it requires recomputing biases for documents which are affected by the swap i.e. have common terms with other documents in the partition. 
Currently, its only employed once before starting the shuffle operation, where we check against the max gain possible for a given iteration and if its at least `iter` bits apart. This condition can be checked with each swap when performing quick select, if its worthy.

### DIFF
--- a/lucene/misc/src/java/org/apache/lucene/misc/index/BPIndexReorderer.java
+++ b/lucene/misc/src/java/org/apache/lucene/misc/index/BPIndexReorderer.java
@@ -472,8 +472,12 @@ public final class BPIndexReorderer {
 
         @Override
         public int comparePivot(int j) {
-          int cmp = Float.compare(pivotBias, biases[j]);
-          if (cmp == 0) {
+          int cmp = Float.compare(pivotBias-iter, biases[j]);
+          if (cmp > 0) {
+            return cmp;
+          }
+          cmp = Float.compare(pivotBias, biases[j]);
+          if (cmp >= 0) {
             // Tie break on the doc ID to preserve doc ID ordering as much as possible
             cmp = pivotDoc - docIDs.ints[j];
           }


jpountz (commented):
Thanks for catching this!

This seems to cover the case when `pivotBias - bias <= iter`, but I would expect to also need to consider the case when `pivotBias - bias >= - iter`?

rishabhmaurya:
@jpountz that's a good point. Since we want tolerance to be `iter`, we can check for - 
`pivot - bias <= iter/2` and `pivot - bias >= -iter/2`.

rishabhmaurya:
@jpountz  +1 on comparison function that is not transitive could be problematic. 
Do you think, given we are recomputing biases on each swap, that also breaks the transitivity of compare function? 

On reading Bentley-McIlroy 3-way partitioning (https://algs4.cs.princeton.edu/lectures/demo/23DemoPartitioning.pdf), I don't think our swap logic makes sense here especially in phase-1 of the algo. 
It does a 3 way partitioning and runs in 2 phases. In phase 1 (https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/IntroSelector.java#L100-L118), it creates 4  partitions - `[elements equal to pivot] [elements smaller than pivot][elements greater than pivot] [elements equal to pivot]`. 
Then in phase-2 (https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/IntroSelector.java#L119-L133) it rearranges these 4 partitions into 3 partitions by swapping elements from the left and right end to the middle, so it looks like - `[smaller elements] [equal elements] [greater element]`. 

All swaps performed in phase-1 is when the element is found equal to the pivot so that they can be swapped towards either left or the right end. Recomputing biases on swap is almost unnecessary in this phase since these elements will be swapped again towards the middle in phase-2. 

If we want to continue using IntroSelector, I propose that we should not perform any recomputation of gains in phase-1 of the algorithm, and only perform recomputation in phase-2 when swaps are being performed. 
For it, we need to distinguish `swap()` from both the phases of the algo. 

For `comparePivot()`, which is only used in phase-1 of the algo, I like your approach of rounding values - `int cmp = Float.compare(Math.round(pivot / iter * 2), Math.round(bias / iter * 2))`, so we can try it out. 

Let me know what do you think?

jpountz:
You are anticipating my next question, which was going to be about whether it helps speed up doc ID reordering.

Your reasoning about why the swap logic is inefficient seems to assume that there would be lots of equal elements, but actually since we tie break on docID, there would never be any equal elements and phase 2 is a no-op in practice?

> and also keep track of original positions

I have a bias against this because it would further increase RAM requirements by 4 bytes (one integer) per doc ID?

If we see evidence that swapping is a bottleneck, I'd be more keen to looking into using something like `RadixSelector`, which should perform less useless swapping.

rishabhmaurya:
> Your reasoning about why the swap logic is inefficient seems to assume that there would be lots of equal elements, but actually since we tie break on docID, there would never be any equal elements and phase 2 is a no-op in practice?

yes! it slipped my head that we tie break on docID. You're right, phase-2 here is a no-op here. 
However, do you see issue with recomputing the bias on [swap(i, j)](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/IntroSelector.java#L110) in phase-1. It probably works since we do `++i` and `--j` in next iteration and never use `i` or `j` ever again while comparison, so its probably harmless to change their bias. 

> I have a bias against this because it would further increase RAM requirements by 4 bytes (one integer) per doc ID?

or a boolean per doc, if it moved across middle as that's we recompute. 

> If we see evidence that swapping is a bottleneck, I'd be more keen to looking into using something like RadixSelector, which should perform less useless swapping.

That makes sense. I have some evidences with `http_logs` workload (https://github.com/apache/lucene/pull/13123#discussion_r1513889998) but again it may not be the best workload to benchmark this feature. I will run benchmark against other workloads too and share the numbers here, then we can see if swapping is a bottleneck and make a call. 
Thanks for reviewing and implementing the reorderer.

Issue #13182, opened 2024-03-13
Making vector comparisons pluggable


benwtrent:
### Description

Opening an issue to continue discussion originating here: 
 https://github.com/apache/lucene/pull/13076#issuecomment-1930363479

Making vector similarities pluggable via SPI will enable users to provide their own specialized similarities without the additional burden of Lucene core having to provide BWC for all the various similarity functions (e.g. hamming, jaccard, cosine).

It is probably best that the plug-in-play aspect is placed on `FieldInfo`, though this would be a bit of work as `FieldInfo` isn't currently pluggable. Attaching it directly to a particular Vector Format would place undue burden on users, requiring a new format for any field that desires a separate similarity.

While I am not 100% sure how to add it to FieldInfo, I do want to try and figure out the API for such a change. 

When used within a particular vector format, the following scenarios would be useful:
 - Indexing, comparing on-heap vectors accessible via ordinals
 - Merging, comparing off-heap vectors, possibly reading them directly on-heap via ordinals
 - During search, an on-heap user provided vector being compared with off-heap vectors via ordinals (potentially reading them on-heap).

Some of the optimizations discussed here: https://github.com/apache/lucene/pull/12703 show some significant gains in being able to simply have a memory segment and an ordinal offset. While we are not there yet in Lucene, it indicates that we shouldn’t force the API to reading off-heap vectors into a `float[]` or `byte[]` arrays..

I was thinking of something *like* this. 100%, this is not finalized, just wanting to start the discussion.

<details>

```java

public abstract class VectorSimilarityInterface implements NamedSPILoader.NamedSPI {
  
  private static final class Holder {
    private static final NamedSPILoader<VectorSimilarityInterface> LOADER =
      new NamedSPILoader<>(VectorSimilarityInterface.class);
    private Holder() {}
    static NamedSPILoader<VectorSimilarityInterface> getLoader() {
      if (LOADER == null) {
        throw new IllegalStateException(
          "You tried to lookup a VectorSimilarityInterface name before all formats could be initialized. "
            + "This likely happens if you call VectorSimilarityInterface#forName from a VectorSimilarityInterface's ctor.");
      }
      return LOADER;
    }
  }
  
  public static VectorSimilarityInterface forName(String name) {
    return VectorSimilarityInterface.Holder.getLoader().lookup(name);
  }
  
  private final String name;
  protected VectorSimilarityInterface(String name) {
    NamedSPILoader.checkServiceName(name);
    this.name = name;
  }
  @Override
[...]

PR #13187, opened 2024-03-18
Pass custom similarity function to similarityToQueryVector API

shubhamvishu:
### Description

This PR allows passing a custom vector similarity function to DVS implementations of `VectorSimilarityValuesSource` as opposed to current behaviour which by default only allows to use the configured `VectorSimilarityFunction` associated with the KNN/vector field.

**Tests :** Added some unit tests

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/ByteVectorSimilarityValuesSource.java
+++ b/lucene/core/src/java/org/apache/lucene/search/ByteVectorSimilarityValuesSource.java
@@ -20,9 +20,9 @@ package org.apache.lucene.search;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Objects;
+import java.util.function.BiFunction;
 import org.apache.lucene.index.ByteVectorValues;
 import org.apache.lucene.index.LeafReaderContext;
-import org.apache.lucene.index.VectorSimilarityFunction;
 
 /**
  * A {@link DoubleValuesSource} which computes the vector similarity scores between the query vector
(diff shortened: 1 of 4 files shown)


msokolov:
There is some discussion how to make similarities more pluggable https://github.com/apache/lucene/issues/13182 that seems relevant. Part of the idea there is to accept ordinal values rather than arrays so that internally we can make use of directory memory access rather than having to materialize values on the heap. This issue could still fit into that framework I think, just providing a query vector but not a document vector? I guess we should think about whether this change would lock us in further to the current situation or not. Maybe we could go ahead with this and then change it later when that other issue lands? But it could make it more difficult to "plug in" implementations if they have to be reimplemented

shubhamvishu:
Thanks for the review @msokolov! The idea to make it pluggable seems relevant and interesting. Currently it was not possible to use any custom vector similarity function other than the configured one which for instance in case of `DOT_PRODUCT` similarity gives the scaled score and not the actual raw dot product. So by that logic I think this change would be good. 

> I guess we should think about whether this change would lock us in further to the current situation or not. Maybe we could go ahead with this and then change it later when that other issue lands? But it could make it more difficult to "plug in" implementations if they have to be reimplemented

Yes, we should definitely take that into account. Though I'm not sure if this change conflicts with or makes things difficult for the ongoing efforts to have pluggability (maybe @benwtrent would be interested in sharing his thoughts?).

benwtrent:
> Though I'm not sure if this change conflicts with or makes things difficult for the ongoing efforts to have pluggability (maybe @benwtrent would be interested in sharing his thoughts?).

I would assume any change here would want to take advantage of any pluggable similarity. If so, this change would 100% be effected. 

What makes this PR doubly worrying is that none of the interfaces are marked experimental. Meaning, if its released and we want to change it, we gotta be compatible, that doesn't seem wise.

shubhamvishu:
@benwtrent So should we instead wait for the pluggability support and discard this for now? or Is it possible to go forward with this?

> What makes this PR doubly worrying is that none of the interfaces are marked experimental. Meaning, if its released and we want to change it, we gotta be compatible, that doesn't seem wise.

I'm not sure if we can mark them experimental at this point but if yes, maybe we could first mark these classes as experimental so that we don't have to worry about backward compatibility later?

benwtrent:
I am sorry this work has stalled. But I have been iterating on https://github.com/apache/lucene/pull/13200 for a week now. Its getting to a palatable place.

PR #13190, opened 2024-03-18
Add new parallel merge task executor for parallel actions within a single merge action

benwtrent:
This commit adds a new interface to all MergeScheduler classes that allows the scheduler to provide an Executor for intra-merge parallelism. The first sub-class to satisfy this new interface is the ConcurrentMergeScheduler (CMS). In particular, the CMS will limit the amount of parallelism based on the current number of mergeThreads and the configured maxThread options. 

Parallelism is now added by default to the SegmentMerger where each merging task is executed in parallel with the others. 

Additionally, the Lucene99 HNSW codec will use the provided MergeScheduler executor if no executor is provided directly.

Relates to: https://github.com/apache/lucene/issues/12740
Relates to: https://github.com/apache/lucene/issues/9626

This is a take 2 of: https://github.com/apache/lucene/pull/13124


benwtrent:
@dweiss @mikemccand I am currently iterating on how to best make `RateLimitedIndexOutput` `MergePolicy` and `MergeRateLimiter` thread safe. 

Right now, it is all assumed that the interactions with all these are from the same thread, this obviously breaks when we add intra-merge parallelism. 

I was hoping y'all had some thoughts on how these should all work together with intra-merge parallelism? 

Is it enough to make these classes threadsafe and remove assertions? 

Do we want to somehow figure out if the ultimate calling thread was a MergeThread? (This is possible, but will require some wrangling on the tp-executor to keep track of which thread belongs where...)

lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
```diff
@@ -281,11 +297,11 @@ public IndexOutput createOutput(String name, IOContext context) throws IOExcepti
...
-        assert mergeThread == Thread.currentThread()
-            : "Not the same merge thread, current="
-                + Thread.currentThread()
-                + ", expected="
-                + mergeThread;
+        /*        assert mergeThread == Thread.currentThread()
+        : "Not the same merge thread, current="
+            + Thread.currentThread()
+            + ", expected="
+            + mergeThread;*/
```
benwtrent, line 304: I don't know about this, it seems like paranoid check that may not be necessary if RLIO was threadsafe?

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -136,13 +136,13 @@ public boolean isAborted() {
...
         throws InterruptedException {
-      if (Thread.currentThread() != owner) {
+      /*      if (Thread.currentThread() != owner) {
         throw new RuntimeException(
             "Only the merge owner thread can call pauseNanos(). This thread: "
                 + Thread.currentThread().getName()
                 + ", owner thread: "
                 + owner);
-      }
+      }*/
```
benwtrent, line 145: Do we care about this at all if the underlying throttling is threadsafe?

lucene/core/src/java/org/apache/lucene/store/RateLimitedIndexOutput.java
```diff
@@ -28,30 +30,30 @@ public final class RateLimitedIndexOutput extends FilterIndexOutput {
   private final RateLimiter rateLimiter;
 
   /** How many bytes we've written since we last called rateLimiter.pause. */
-  private long bytesSinceLastPause;
+  private final AtomicLong bytesSinceLastPause = new AtomicLong(0);
```
benwtrent, line 33: This obviously creates some overhead now as RLIO could be accessed via multiple threads.  I will likely have to benchmark to see the impact.

lucene/core/src/java/org/apache/lucene/store/RateLimitedIndexOutput.java
```diff
@@ -62,30 +64,42 @@ public void writeBytes(byte[] b, int offset, int length) throws IOException {
...
   }
 
   private void checkRate() throws IOException {
-    if (bytesSinceLastPause > currentMinPauseCheckBytes) {
-      rateLimiter.pause(bytesSinceLastPause);
-      bytesSinceLastPause = 0;
-      currentMinPauseCheckBytes = rateLimiter.getMinPauseCheckBytes();
+    AtomicLong localBytesSinceLastPause = new AtomicLong(0);
+    AtomicBoolean shouldPause = new AtomicBoolean(false);
+    bytesSinceLastPause.updateAndGet(
```
benwtrent, line 89: I went with an `updateAndGet` instead of `synchronize` here. Either would work really.

lucene/core/src/java/org/apache/lucene/store/RateLimitedIndexOutput.java
```diff
@@ -62,30 +64,42 @@ public void writeBytes(byte[] b, int offset, int length) throws IOException {
...
+        bytes -> {
+          if (bytes > currentMinPauseCheckBytes.get()) {
+            shouldPause.set(true);
+            currentMinPauseCheckBytes.set(rateLimiter.getMinPauseCheckBytes());
+            localBytesSinceLastPause.set(bytes);
+            return 0;
+          } else {
+            shouldPause.set(false);
+          }
```
benwtrent, line 98: I figured if another thread reached this throttling and is paused & reset the counter, than this current thread shouldn't pause. Throttle tracking is effectively global and each thread uses the same localBytesSinceLastPause, so the work done by another thread could cause a current thread to pause. Or another thread pausing could allow this thread to continue work.

lucene/core/src/java/org/apache/lucene/index/MergeRateLimiter.java
```diff
@@ -118,24 +118,32 @@ private long maybePause(long bytes, long curNS) throws MergePolicy.MergeAbortedE
...
-    // We don't bother with thread pausing if the pause is smaller than 2 msec.
-    if (curPauseNS <= MIN_PAUSE_NS) {
-      // Set to curNS, not targetNS, to enforce the instant rate, not
-      // the "averaged over all history" rate:
-      lastNS = curNS;
+    final double rate = mbPerSec; // read from volatile rate once.
+    final double secondsToPause = (bytes / 1024. / 1024.) / rate;
+
+    AtomicLong curPauseNSSetter = new AtomicLong();
+    lastNS.updateAndGet(
```
benwtrent, line 125: We would need to lock here or do an `updateAndGet`. `lastNS` is only updated in this particular method, so using atomics seemed better to me.

lucene/core/src/java/org/apache/lucene/index/MergeRateLimiter.java
```diff
@@ -118,24 +118,32 @@ private long maybePause(long bytes, long curNS) throws MergePolicy.MergeAbortedE
...
+          // maybe we should also offer decayed recent history one?
+          long targetNS = last + (long) (1000000000 * secondsToPause);
+          long curPauseNS = targetNS - curNS;
+          // We don't bother with thread pausing if the pause is smaller than 2 msec.
+          if (curPauseNS <= MIN_PAUSE_NS) {
+            // Set to curNS, not targetNS, to enforce the instant rate, not
+            // the "averaged over all history" rate:
+            curPauseNSSetter.set(0);
+            return curNS;
+          }
```
benwtrent, line 138: Its possible that other threads request for very short pauses, thus all updating `lastNS` to their respective `curNS`. I honestly don't know if this is OK or not. But to handle this case, I decided to use `updateAndGet` to ensure if `lastNS` is changed, we account for it and update our pause request accordingly.

benwtrent (commented):
Some high level concerns and thoughts I had on the new changes.

@dweiss || @mikemccand y'all might be interested.

jpountz:
I understand that we need to make changes to account for the fact that multiple threads may be contributing to the same merge concurrently, but I would not expect `RateLimitedIndexOutput` to need to be thread-safe: the caller needs to manage synchronization themselves anyway if they want bytes to be written in the correct order?

benwtrent:
@jpountz 
>  the caller needs to manage synchronization themselves anyway if they want bytes to be written in the correct order?

The MT safety isn't really around the bytes. I agree with your assessment there. But its more around the throttling. The way I read the class is that different threads could be writing to different files but both could experience throttling and we should account for that.

lucene/core/src/java/org/apache/lucene/store/RateLimitedIndexOutput.java
benwtrent: Ah, yeah, I could remove all the concurrency checks here. I was just concerned as the above assertions tripped because of the threads. but you are correct, they are still ultimately only written to via one thread.

dweiss:
I don't think rate limiting has to be exact - it's already allowed to be temporarily exceeded (see comments in writeBytes on RateLimitedIndexOutput).

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
benwtrent: > I guess not, but now we need to make sure that the rate is applied across all threads rather than per-thread?

Looking at the code, it seems like it was previously assumed that only one `createOutput` was called at a time for a merge. Now, it could be that more than one is called. 

I guess this means we need a `RateLimitingDirectory` that passes things in to the `RateLimitedIndexOutput` ensure global rate limiting is controlled across all outputs.  Is this what you are talking about?

I am not sure that `pauseNanos` should directly know about other threads and pause them. It seems better to put this up on the directory level.

lucene/core/src/java/org/apache/lucene/index/MergeRateLimiter.java
benwtrent: So, the limits are on a per `RateLimitedIndexOutput` bases, not necessarily per thread. Since all `RateLimitedIndexOutput` constructed share the same rate-limiter, it might pause all threads?

But I don't think so, as each process could have its own `RateLimitedIndexOutput` which is tracking its own bytes locally and while the combined total of all threads might exceed the throttling limit, the output isn't really throttled until a single thread exceeds the limit.

lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
```diff
@@ -446,7 +454,13 @@ private static String rateToString(double mbPerSec) {
 
   @Override
   public void close() {
-    sync();
+    try {
```
jpountz, line 457: should we call `super.close()` as well, to close `MergeScheduler`'s `SameThreadExecutorService` which `ConcurrentMergeScheduler` uses for small merges?

It's not a big deal not to close it, but it would help catch if we ever send tasks to this executor after closing.

lucene/core/src/java/org/apache/lucene/index/MergeRateLimiter.java
jpountz: >  the output isn't really throttled until a single thread exceeds the limit.

This limitation feels ok to me, let's just add a comment about it? Intuitively, the write rate at merge time is rather bursty, so if the sum of the bytes written by all threads running this merge exceeds the limit, then there would often be one thread that exceeds the limit on its own as well.

lucene/core/src/java/org/apache/lucene/index/MergeRateLimiter.java
jpountz: It's hard for me to reason about this as well.

I'm wondering about keeping `maybePause` as-is and making `MergeRateLimiter#pause` synchronized, essentially trying to make the pausing logic behave as if threads were writing bytes sequentially rather than in parallel. (I'm considering making `pause` synchronized rather than `maybePause` so that `System.nanoTime()` is computed within the lock and the pausing logic accounts for the fact that some time may have been spent waiting on the lock already.)

mikemccand: Sorry I'm trying to catch up here and will probably ask a bunch of stupid questions :)  Thank you @benwtrent for persisting in this hairy logic!

> (I'm considering making `pause` synchronized rather than `maybePause` so that `System.nanoTime()` is computed within the lock and the pausing logic accounts for the fact that some time may have been spent waiting on the lock already.)

Couldn't we move the `curNS = System.nanoTime()` inside the locked/updateAndGet'd part of `maybePause`?  I like the thread safety that the `updateAndGet` is giving us here, ensuring we accurately account for all bytes written by N threads within a single merge.  Also, I don't expect the added sync required here will hurt performance much: Lucene is doing much work to produce these bytes being written already, so conflict should be rare-ish.

Also, if we make `pause` sync'd, it will cause scary looking `jstack` thread dumps?  Making it look like one thread is sleeping while holding a lock and blocking other threads (which indeed is what it'd be doing).  Versus the thread stacks we'd see w/ the current approach that make it quite clear that all N threads are intentionally stalling in `mergeProgress.pauseNanos`?  It would reduce the horror reaction we'd see peeking at thread dumps maybe ...

benwtrent:
@mikemccand @jpountz 

So, thinking about this more as I fell asleep.

This is how throttling will work as it is in this PR:

 - Throttling is per thread. Meaning, a intra-merge thread only gets throttled once the bytes it has written get to the rate limit set
 - Consequently, we may get to `rateLimitBytes*numIntraMergeIO` bytes before a single thread gets throttled.

This makes us throttle less often given how many bytes are actually written. Maybe this is OK as the throttling logic could "catch up" if things continue to get backed up? Throttling has always been a "best effort" thing anyways.

Even if we somehow made the RateLimiter throttle every thread (via some global semaphore or something...), we would still only throttle once one of the multiple threads hit the byte throttling limit.

IMO, either we: 

 - account for bytes globally (per directory) & throttle globally (global lock that pauses all threads)
 - accept that throttling is per thread and bytes used is measured in individual threads.

Selfishly, I wish for the second option as its the simplest.

lucene/core/src/java/org/apache/lucene/index/MergeRateLimiter.java
benwtrent: > Couldn't we move the `curNS = System.nanoTime()` inside the locked/updateAndGet'd part of maybePause? 

Yes, we could, otherwise if there are conflicts between threads (thus having to do the `updateAndGet` more than once), the `curNS` gets further in the past.

> is giving us here, ensuring we accurately account for all bytes written by N threads within a single merge

That is not what it is doing. It is ensuring that the throttling timestamps stay in sync :/. Bytes being throttled are still per RateLimitedIndexOutput, which means they are per thread.

mikemccand:
Phew, it took me two hours of strongly caffeinated time to catch up on this!  Thank you @benwtrent.

What an awesome change, modernizing Lucene's merge concurrency so that intra-merge (just HNSW, but maybe other index parts soon) concurrency is enabled by default.

(Separately, the nightly benchy had a silly off-by-one bug preventing it from resuming after you reverted the previous PR ... I think I've fixed that bug now, and kicked off a one-off nightly benchy run that will hopefully finish late today).

Some high level questions (sorry if these were already asked/answered):
  * How do we control the risk that a massive merge with KNN vectors soaks up all available concurrency from the shared Executor for intra-merge concurrency (all threads doing HNSW merging) and then starves smaller merges that would finish quickly?
  * Maybe we don't need the IO write rate limiter anymore?  It's a tricky setting because the `mbPerSec` you set is then multiplied by the number of concurrent merges that are running.  It is a per-merge setting, not a global setting, so it's kinda trappy today.

Finally, I worry that the rate-limiter's simplistic "instantaneous" measure is making it effectively super buggy (there is a TODO about this), because merging that does a lot of CPU work and then writes a lot of bytes will be effectively throttled to far below the target `mbPerSec` in aggregate.  A better solution might be something like the "burst IOPs bucket" that AWS (and likely other cloud providers) offer ("Burst IOPS is a feature of Amazon Web Services (AWS) EBS volume types that allows applications to store unused IOPS in a burst bucket and then drain them when needed" -- thank you Gemini for the summary).  This is clearly not a blocker for this issue, but we really should (separately) fix it.  I'll open a follow-on issue about this ... but to me it's another reason to maybe remove this feature entirely.

mikemccand:
> Selfishly, I wish for the second option as its the simplest.

+1 to keep with the second approach.  It is best effort, and, the "instant vs burst-bucket" bug above is yet more weirdness about it :)

Another thought on the "instant vs burst-bucket" bug: I suppose in some case, e.g. bulk merging of stored fields (no deletions, so we are just copying byte blocks), there is very little CPU and tons of IO and in cases like that, the instant vs burst-bucket approaches would be essentially the same (both wind up throttling based on the instant rate since the burst bucket is quickly exhausted).  I'll open a spinoff ...

mikemccand:
I opened https://github.com/apache/lucene/issues/13193 specifically about the "instant vs burst bucket" IO rate limiting approaches.

benwtrent:
> How do we control the risk that a massive merge with KNN vectors soaks up all available concurrency from the shared Executor for intra-merge concurrency (all threads doing HNSW merging) and then starves smaller merges that would finish quickly?

Intra-merge threads do not count against the `maxThreadCount` and are not tracked in `mergeThreads`. Additionally, intra-merge threads only allow up to `maxThreadCount - mergeThreads.size() - 1` threads to ever run on its pool.

With this change, it is possible that a chunk of work delegated starves an intra-merge thread and counts against that parallelism, but it wouldn't block other `mergeThreads` from running. Those other `mergeThreads` just might have to run on themselves instead of off of the intra-merge thread-pool.

I do think there is some future work here to unify thread tracking between all the inter&intra merge threads, but all the custom logic in `mergeThreads` and how we create, track, etc. just seemed like too much to figure out in addition to adding intra-merge parallelism.

> Maybe we don't need the IO write rate limiter anymore?

I honestly don't know. There is still logic in the CMS that determines if we are on a spinning disk vs. SSD, which is sort of crazy to me :D.

benwtrent:
I re-ran `IndexGeoNames` with the latest commits on this PR. I didn't override CMS settings at all & I allowed merging to occur while indexing. This was done on my macbook, so the numbers might be a little fuzzy. 

I indexed with 4 threads, and did the batching.

RamBufferSize: 12MB
Baseline:
```
8221789: 28.378 sec
forceMerge took 18856 ms
```

Candidate:
```
8221789: 27.428 sec
forceMerge took 6581 ms
```

RamBufferSize: 1MB

Baseline:
```
8221789: 49.209 sec
forceMerge took 17235 ms
```

Candidate:
```
8221789: 46.078 sec
forceMerge took 4638 ms
```

jpountz:
> Maybe we don't need the IO write rate limiter anymore?

I'm curious what you have in mind. Are you considering throttling merges purely based on merge concurrency then? E.g. slow indexing -> single merge at a time, heavy indexing -> many parallel merges? This wouldn't sound crazy to me as I don't recally seeing indexing/merging saturate I/O.

jpountz:
> Maybe we don't need the IO write rate limiter anymore?

Nevermind, I see that you added more details on #13193.

lucene/core/src/java/org/apache/lucene/index/MergeRateLimiter.java
mikemccand: > > is giving us here, ensuring we accurately account for all bytes written by N threads within a single merge
> 
> That is not what it is doing. It is ensuring that the throttling timestamps stay in sync :/. Bytes being throttled are still per RateLimitedIndexOutput, which means they are per thread.

Ahhhhhh gotchya.  Each `RateLimitedIndexOutput` tracks its own `bytesSinceLastPause` and then invokes `pause` with its own private byte count, OK.  So, yes, with the current approach we rate limit MB/sec per thread, not per merge.  I think that's fine.  Best effort!

benwtrent:
If y'all are good with it, I will merge this tomorrow.

jpountz:
@benwtrent FYI I left a few minor comments: on `ConcurrentMergeScheduler#close`, and about adding a comment for the limitation you identified that a merge may not get throttled if the sum of bytes written exceeds the rate but none of the merge threads for the same merge exceed the rate.

benwtrent:
@jpountz added comments to the creation of the ratelimiter in CMS and the limiter itself. Also updated the close interaction.

lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
```diff
@@ -445,8 +456,15 @@ private static String rateToString(double mbPerSec) {
   }
 
   @Override
-  public void close() {
-    sync();
+  public void close() throws IOException {
+    super.close();
```
jpountz, line 460: nit: should we try to close as best as possible in the event of an exception, e.g. doing something like below

```java
    IOUtils.close(
        this::sync,
        super::close,
        intraMergeExecutor == null ? null : intraMergeExecutor::shutdown);
```

jpountz (approved):
Thanks for pushing on this change, I like it. The fact that the extra merge concurrency may not starve merging from threads is good. I'm curious how the nightly benchmarks will react to it, given the high number of CPU cores of beast3.

One question on my mind is whether this change should make us update the default number of merging threads that `ConcurrentMergeScheduler` configures (in a follow-up issue/PR).

mikemccand:
> Thanks for pushing on this change, I like it. The fact that the extra merge concurrency may not starve merging from threads is good. I'm curious how the nightly benchmarks will react to it, given the high number of CPU cores of beast3.

Nightly benchy is still forcing its own thread pool into the HNSW indexing Codec component (`Lucene99Hnsw*VectorsFormat`) -- once this change is merged, I'll remove that so we switch to this awesome default approach.  But maybe we won't see too much change in the nightly indexing throughput since it's already doing intra-merge concurrency "itself".
 
> One question on my mind is whether this change should make us update the default number of merging threads that `ConcurrentMergeScheduler` configures (in a follow-up issue/PR).

+1 to explore this.  Nightly benchy hardwires the maxMergeCount=16, maxThreadCount=12.  Maybe they should be higher :)

But also note that the nightly benchy does not wait for merges on `IndexWriter.close`, so it's not really a fair test of merge performance.

mikemccand (approved):
Thank you @benwtrent -- what an awesome improvement in Lucene's default concurrency!

benwtrent:
@mikemccand 

> I'll remove that so we switch to this awesome default approach. 

For now, you still need to tell the codec how many workers are available for it to use. But you should be able to pass `null` for the executor.

dnhatn:
Elasticsearch CI has identified an issue related to this change. The PerFieldDocValuesFormat and PerFieldPostingsFormat, which mutate and reset the fieldInfos of the mergeState while executing a merge. Consequently, other ongoing merges may fail to access some fieldInfos.

https://github.com/apache/lucene/blob/e2788336d49c56050b3cea5944008bcffe35dcb1/lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldDocValuesFormat.java#L150-L158

https://github.com/apache/lucene/blob/04f335ad79cae019d1a552e41e0029d061d8e3d9/lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldPostingsFormat.java#L196-L214

mikemccand:
It looks like this awesome change was backported for 9.11.0?  I'll add the milestone.  So hard to remember to set the milestones on our issues/PRs...

benwtrent:
@mikemccand oh dang, I haven't been doing that. Thanks for picking up my slack!

PR #11888, opened 2022-10-28
[Fix] Binary search the entries when all suffixes have the same length in a leaf block.

vsop-479:
Fix https://github.com/apache/lucene/pull/11722 's bug.

PR #13192, opened 2024-03-20
Binary search all terms.

vsop-479:
Base on https://github.com/apache/lucene/pull/11888 (binary search `allEqual` term leaf), this change implemented binary search for all term leaf (`allEqual` and `unEqual`). 
I am still trying to do the same thing to `NonLeaf` term.


vsop-479:
Implemented binary search term non leaf (`allEqual` and `unEqual`).

vsop-479:
@jpountz @mikemccand 

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnumFrame.java
```diff
@@ -196,6 +207,90 @@ void loadBlock() throws IOException {
...
+      // Reset suffixLengthsReader's position.
+      suffixLengthsReader.setPosition(0);
+    } else {
+      suffixes = new int[entCount];
+      // TODO: remove postions if it is unnecessary.
+      postions = new int[entCount];
+      if (isLeafBlock) {
+        for (int i = 0; i < suffixes.length; i++) {
+          suffixes[i] = suffixLengthsReader.readVInt();
+          postions[i] = suffixLengthsReader.getPosition();
```
mikemccand, line 262: `postions` -> `positions`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnumFrame.java
```diff
@@ -554,49 +683,128 @@ public SeekStatus scanToTermLeaf(BytesRef target, boolean exactOnly) throws IOEx
...
-    do {
-      nextEnt++;
+    // TODO early terminate when target length unequals suffix + prefix.
+    // But we need to keep the same status with scanToTermLeaf.
+    int start = nextEnt;
+    int end = entCount - 1;
+    // Binary search the entries (terms) in this leaf block:
+    int cmp = 0;
+    while (start <= end) {
+      int mid = (start + end) / 2;
```
mikemccand, line 693: Looks like we need to rebase this change on the latest from #11888?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnumFrame.java
```diff
@@ -196,6 +207,90 @@ void loadBlock() throws IOException {
     suffixLengthsReader.reset(suffixLengthBytes, 0, numSuffixLengthBytes);
     totalSuffixBytes = ste.in.getFilePointer() - startSuffixFP;
 
+    // Prepare suffixes, offsets to binary search.
+    if (allEqual) {
+      if (isLeafBlock) {
+        suffix = suffixLengthsReader.readVInt();
+      } else {
+        // Handle subCode for non leaf block.
+        postions = new int[entCount];
```
mikemccand, line 216: This many new allocations, at such a low level / hot spot (on each block load) seems risky / performance hurting?

I wonder how often we see `allEqual` for non-leaf blocks when the field is indeed fixed-length terms?  It might happen often, depending on how evenly distributed the tokens are across term space?  A random UUID should be quite even, at least on initial indexing, but a predictable ID (just 0-padded incrementing int, like luceneutil) might be less so?

mikemccand (commented):
Thanks @vsop-479 -- it seems to get quite a bit more complex to add binary search to non-leaf blocks, and to non-`allEqual` blocks!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnumFrame.java
vsop-479: Thanks, i will fix this.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnumFrame.java
vsop-479: Yes, i will merge https://github.com/apache/lucene/pull/11888 into this change.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnumFrame.java
vsop-479: Yes, maybe it is less worthy to implement this binary search to non-leaf blocks (both `allEqual` and non-`allEqual` ?).

Can we just take this change to non-`allEqual` leaf blocks? But it still need to allocation `suffixes ` and `offsets` ( If we don't need `positions` to set `suffixLengthsReader`'s position after searching a block).

vsop-479: > assume We don't need positions to set suffixLengthsReader's position after searching a block.

Setting `suffixLengthsReader`'s position is necessary, so we need `positions`.

vsop-479:
> This many new allocations

Maybe we can share these allocations(`suffixes`, `positions`, `positions`) from `searchers`, since they are just immutable and non-stateful data.

vsop-479:
Or, we just supply this (maybe only for `non-allEqual` leaf blocks) as an option, So, users can use it when their applications  are not busy.

PR #13195, opened 2024-03-20
Fix NPE for when fields are missing in `searchNearestVectors`

benwtrent:
Related to: https://github.com/apache/lucene/pull/13162

Since this is unreleased, no changelog entry is necessary.


mayya-sharipova (approved):
Thanks

ChrisHegarty:
Was this behaviour regressed by https://github.com/apache/lucene/pull/13162? I wonder if a test would be warranted to ensure that it does not reoccur.  Also, in some areas `fi.getVectorDimension() == 0` was also checked at the same time.

benwtrent:
@ChrisHegarty added a test :)

ChrisHegarty (approved):
Thanks @benwtrent. LGTM

PR #13196, opened 2024-03-21
Add support for posix_madvise to Java 21 MMapDirectory

uschindler:
This is a first idea how we can use Panama Foreign to pass `madvise()` hints to the kernel when mapping memory segments.

The code looks up the function pointer from stdlib (libc) on Linux and Macos (untested, but should work) and then invokes `madvise()` for all MemorySegments we have mmapped when the following is true:
- IOContext#readOnce is set
- The chunk size is large enough (at least 8192 means `chunkSizePower>=13`) - this prevents TestMultiMMap from failing because for very small mappings (as done by this test), the `FileChannel#map` call will produce unaligned memory segments (it uses some tricks and maps larger segments and returns slices - which are no longer pageSize aligned)
- There is a noop implementation doing nothing which is choosen if you disable native access

Interestingly it works without any extra parameters to command line (at least in Java 21).

This is a draft only to do some performance tests and extend the IOContext interpretation to try out more possibilities. The current "readOnce => MADV_SEQUENTIAL" is just an example as this is the main issue: We merge segments and don't want the soon to be trashed segments be sticky in RAM. MADV_SEQUENTIAL instructs kernel to forget about the mappings and also do readahead which helps during merging.


jpountz:
I remember this kind of things being discussed more than 10 years ago, it's extremely exciting to see it close to being included in the default `Directory`!

> current "readOnce => MADV_SEQUENTIAL" is just an example as this is the main issue: We merge segments and don't want the soon to be trashed segments be sticky in

I believe that we don't use `readOnce` for merging, but for tiny metadata files that we load fully in memory. That said, I don't think that this should block this change, we can later look into the proper way to detect whether to pass `MADV_SEQUENTIAL` based on the `IOContext`.

uschindler:
I refactored the code a bit:
- @jpountz I added the merge context (I checked DirectIODirectory).
- I retrieve the pageSize to check if a segment is correctly aligned. Unfortunately this is mega-stupid. I tend to remove it again (the constant `_SC_PAGESIZE` is part of a C `enum` and its value is different in various platforms as no fixed value is assigned) and maybe keep a hardcoded page size to handle the case. Maybe just check that it is a multiple of 64K (native page size is 8K, but may be different), because it is normally only an issue in this test....

uschindler:
I removed the page size retrieval again. No need for it, we just be safe and only apply the advice, when the chunk size is large enough.

uschindler:
Unfortunately after the problems with other constants I tend to hide the advice constants in the class any use some abstraction, so we can call

We should maybe also simply remove the NoopNativeAccess and do a null check.

lucene/core/src/java21/org/apache/lucene/store/PosixNativeAccess.java
```diff
@@ -0,0 +1,82 @@
...
+import java.lang.invoke.MethodHandle;
+import java.util.Locale;
+import java.util.logging.Logger;
+
+@SuppressWarnings("preview")
+final class PosixNativeAccess extends NativeAccess {
+
+  private static final Logger LOG = Logger.getLogger(PosixNativeAccess.class.getName());
+
+  private final MethodHandle mh$posix_madvise;
```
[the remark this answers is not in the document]
ChrisHegarty: I pushed a small change to refactor this. Revert or refactor as you like, but the MH is now static final.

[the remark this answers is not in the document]
uschindler: I don't like this, because now the Unsupported operation exception is triggered by static initializer and is causing NoClassDefFoundError.

If you look at the caller of ctor you will see the try catch.

I added this because for some reason it could fail to find the symbol (e.g. new libc version,...). So I want it bullet proof.

The non static method handle is not a problem, it's called not at perf critical places. The constructor is only called once.

Can you revert? If you really want it static we would need another logic.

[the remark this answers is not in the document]
ChrisHegarty: I’ll take another look or revert tomorrow.

lucene/core/src/java/org/apache/lucene/store/IOContext.java
```diff
@@ -54,14 +60,16 @@ public enum Context {
 
   public static final IOContext DEFAULT = new IOContext(Context.DEFAULT);
 
-  public static final IOContext READONCE = new IOContext(true, false);
+  public static final IOContext READONCE = new IOContext(true, false, false);
+
+  public static final IOContext READ = new IOContext(false, false, false);
 
-  public static final IOContext READ = new IOContext(false, false);
+  public static final IOContext LOAD = new IOContext(false, true, false);
```
jpountz, line 67: Should `LOAD` set `randomAccess=true` rather than `false`? We use it for small index files like the terms index and the stored fields index that have a random access pattern.

lucene/core/src/java21/org/apache/lucene/store/PosixNativeAccess.java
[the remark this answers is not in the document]
uschindler: Thanks. You're not angry to revert it for now and make a second try? I think to have those static we need a different way to report back to the caller of getInstance or the constructor that the necessary symbols are not available or the lookup failed because the module has no access to native functions.

All this is no longer handled and we may cause problems because reporting the exceptions is only depending on class loading order.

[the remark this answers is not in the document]
uschindler: Reverted in 9ed1da74a2e7eff9de89125b75a25b58b803a68c

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
```diff
@@ -92,10 +110,22 @@ private final MemorySegment[] map(
...
       startOffset += segSize;
     }
     return segments;
   }
+
+  private OptionalInt mapContextToMadvise(IOContext context) {
+    if (context.randomAccess) {
+      return OptionalInt.of(NativeAccess.POSIX_MADV_RANDOM);
+    }
+    if (context.readOnce || context.context == Context.MERGE) {
```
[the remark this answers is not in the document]
uschindler: I moved the check for MERGE context to beginning in 242e5e9064acb70430c9082b82b2258ce75beeb3. This makes it always win. The reason for that is: We often set MERGE context, but the other settings keep their defaults (like random access). So when merging is first, we always bail out and advise kernel: "may be freed soon after they are accessed"

lucene/core/src/java21/org/apache/lucene/store/PosixNativeAccess.java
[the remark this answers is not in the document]
ChrisHegarty: >  You're not angry to revert it for now and make a second try?

No, not at all.  I'll take another look at this later, but as you say it is not performance sensitive, is less significant.

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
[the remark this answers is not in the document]
ChrisHegarty: > when you merge segments you certainly always want to read the files with sequential advise 

I've not yet checked, but is `Context.MERGE` set when merging HNSW graphs? (which would favour random access)

lucene/core/src/java/org/apache/lucene/store/IOContext.java
ChrisHegarty: > Should `LOAD` set `randomAccess=true` rather than `false`? We use it for small index files like the terms index and the stored fields index that have a random access pattern. 

Logically, that seems right to me, but ...

> (We may then want to exclude contexts whose `load=true` from `POSIX_MADV_RANDOM`.

Yeah, this seems counter to the above (`LOAD` set `randomAccess=true`), but I get that it would be needed. Is this not already covered by `preload` ?

jpountz: Looking at the code, this looks already covered by `preload` indeed. Maybe we could make it more explicit in the code.

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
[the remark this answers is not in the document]
jpountz: HNSW merging is fine because the first thing that merging does it to write all vectors to a temporary file. And then it builds the graph on top of this temporary file. So it's ok for the input segments to read sequentially, and we could update the merging logic to open this temporary file with `IOContext.RANDOM` since it's the one that will have a random access pattern.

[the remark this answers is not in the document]
ChrisHegarty: Perfect. That is exactly what I was looking for.

lucene/core/src/java/org/apache/lucene/store/IOContext.java
jpountz: I opened https://github.com/uschindler/lucene/pull/4

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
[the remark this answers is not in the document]
uschindler: Thanks, so the current code looks fine.

One thing, I am doubting: If we delete a file, will the kernel free all those pages asap? If yes, why do we do the whole "sequentical" handling for merges at all?

lucene/core/src/java/org/apache/lucene/store/IOContext.java
uschindler: MMapDirectory's preload feature is currently always taking precendence: https://github.com/apache/lucene/pull/13196/files#diff-8681a7b2259f78a6d0d2efe4894ef79ce4697c3825d24675248bc557617773ceL93

If preload is set, it does not use madvise at all. We can change that, too, but I think MADV_NORMAL fits best here. If we preload files we want them to be in RAM.

jpountz: > I think MADV_NORMAL fits best here

++

uschindler:
I refactored the code a bit:
- The POSIX constants are now part of PosixNativeAcess class
- The abstract method madvise() now only takes a `MemorySegment` and the `IOContext`. So it is now the hook for native implementations to implement the "advise" given by the `IOContext` for the platform (POSIX, Windows, Maybe special case for Linux which has more constants if you use plain low-level non-posix MADVISE
- I removed `OptionalInt`: The values are all small and are in the `valueOf` cache, so we just return null when no madvice is needed

One thing: Should we remove the NoopNativeAccess class and just add a null check?

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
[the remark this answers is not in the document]
uschindler: We may add another abstract method to NativeAcess that is called when we close a file. On Linux we could there give some instructions like "forget everything".

uschindler:
Thanks for the PR contributions. You may also commit directly to this branch, unless the change is a complete rewrite :-)

ChrisHegarty:
I took another run at static final ;-). https://github.com/uschindler/lucene/pull/5 (if we still don't want it, then I'll drop it)

ChrisHegarty:
> One thing: Should we remove the NoopNativeAccess class and just add a null check?

Null is a little easier to reason about in the code at the use-site, since the reader is singled that the action may not happen (if null), rather than assuming the advise be given unconditionally.

uschindler:
It would be good if you always run tests with `./gradlew :lucene:core:test -Ptests.directory=MMapDirectory`

This will use MMapDir for all tests. I have no macos machine, if somebody else could test the whole thing with macos (x86 and arm), it would be cool.

It would also be good to get a printout of the haeder file containing those `POSIX_MADV_*` constants. Just grep on `POSIX_MADV_` through the include files (they are a bit hidden on macos, not in /usr/include, best is to search for `unistd.h` to find their location):

```
serv1-vm2:include jenkins$ xcrun --show-sdk-path
/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk
serv1-vm2:~ jenkins$ cd /Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include
serv1-vm2:include jenkins$ fgrep -R POSIX_MADV *
sys/mman.h:#define POSIX_MADV_NORMAL       0       /* [MC1] no further special treatment */
sys/mman.h:#define POSIX_MADV_RANDOM       1       /* [MC1] expect random page refs */
sys/mman.h:#define POSIX_MADV_SEQUENTIAL   2       /* [MC1] expect sequential page refs */
sys/mman.h:#define POSIX_MADV_WILLNEED     3       /* [MC1] will need these pages */
sys/mman.h:#define POSIX_MADV_DONTNEED     4       /* [MC1] dont need these pages */
sys/mman.h:#define MADV_NORMAL             POSIX_MADV_NORMAL
sys/mman.h:#define MADV_RANDOM             POSIX_MADV_RANDOM
sys/mman.h:#define MADV_SEQUENTIAL         POSIX_MADV_SEQUENTIAL
sys/mman.h:#define MADV_WILLNEED           POSIX_MADV_WILLNEED
sys/mman.h:#define MADV_DONTNEED           POSIX_MADV_DONTNEED
serv1-vm2:include jenkins$
```

uschindler:
> > One thing: Should we remove the NoopNativeAccess class and just add a null check?
> 
> Null is a little easier to reason about in the code at the use-site, since the reader is singled that the action may not happen (if null), rather than assuming the advise be given unconditionally.

So your recommendation is also to remove it. Will do later.

uschindler:
> > > One thing: Should we remove the NoopNativeAccess class and just add a null check?
> > 
> > 
> > Null is a little easier to reason about in the code at the use-site, since the reader is singled that the action may not happen (if null), rather than assuming the advise be given unconditionally.
> 
> So your recommendation is also to remove it. Will do later.

I removed it and used an Optional (looks better than null check).

uschindler:
> > ./gradlew :lucene:core:test -Ptests.directory=MMapDirectory
> 
> All tests pass on my Mac M2.

It is always a good idea to run one of those tests with `-Ptests.verbose=true` to see if the INFO mesage about madvise is enabled is printed on console:

```
thetaphi@serv1:~/repro/lucene$ ./gradlew :lucene:core:test --tests TestDirectoryReader -Ptests.directory=MMapDirectory -Ptests.verbose=true | grep madvise
  2> 信息: posix_madvise() available on this platform
WARNING: A command line option has enabled the Security Manager
WARNING: The Security Manager is deprecated and will be removed in a future release
```

ChrisHegarty:
>..
> It is always a good idea to run one of those tests with `-Ptests.verbose=true` to see if the INFO mesage about madvise is enabled is printed on console:
> 
> ```
> thetaphi@serv1:~/repro/lucene$ ./gradlew :lucene:core:test --tests TestDirectoryReader -Ptests.directory=MMapDirectory -Ptests.verbose=true | grep madvise
>   ...

Yes that's better. I been just running a modified test locally where I eyeball the output. Here is the output on my machine (same was your output)

```
$ ./gradlew :lucene:core:test --tests TestDirectoryReader -Ptests.directory=MMapDirectory -Ptests.verbose=true | grep madvise
  2> INFO: posix_madvise() available on this platform
WARNING: A command line option has enabled the Security Manager
WARNING: The Security Manager is deprecated and will be removed in a future release
```

uschindler:
It looks like we are close for making this ready to be added to main branch. We can decide if we want to backport it (should not be too hard). If yes, I will only backport to Java 21 (not 20 or 19).

I will add some getter to MMapDirectory so tests (we should add a test that it is really enabled for Linux and Mac) or users can figure out if madvise is used. I would add `public static boolean MMapDirectory#supportsMadvise()` and use it in tests around MMapDir. Thoughts?

ChrisHegarty:
> We can decide if we want to backport it (should not be too hard). If yes, I will only backport to Java 21 (not 20 or 19).

++  Java 21 is fine.

> I will add some getter to MMapDirectory so tests (we should add a test that it is really enabled for Linux and Mac) or users can figure out if madvise is used. I would add public static boolean MMapDirectory#supportsMadvise() and use it in tests around MMapDir. Thoughts?

The platform may support madvise, but it is up to the calling code to set the IOContext appropriately for it to have any affect. I'm not sure what the user can do about any of this, other than log it? I wonder if it makes sense to add some short narrative to the IOContext javadoc - which would reference the new MMapDirectory#supportsMadvise.

We kinda need MMapDirectory#supportsMadvise to assert the output of some of the tests - I can only think of checking the log output? Then we just need a handful of new tests to check the various combinations of IOContext. Additionally, I assume we can have unit tests for NativeAccess directly. Lemme take a quick look at this.

lucene/core/src/java/org/apache/lucene/store/IOContext.java
```diff
@@ -44,24 +44,32 @@ public enum Context {
...
+  public final boolean randomAccess;
+
   /**
    * This flag is used for files that are a small fraction of the total index size and are expected
    * to be heavily accessed in random-access fashion. Some {@link Directory} implementations may
    * choose to load such files into physical memory (e.g. Java heap) as a way to provide stronger
-   * guarantees on query latency.
+   * guarantees on query latency. If this flag is set, then {@link #randomAccess} will be true.
    */
   public final boolean load;
```
rmuir, line 59: maybe as a followup we can refactor these booleans? This file is very confusing to me...

lucene/core/src/java/org/apache/lucene/store/IOContext.java
```diff
@@ -124,6 +142,7 @@ public int hashCode() {
     result = prime * result + ((flushInfo == null) ? 0 : flushInfo.hashCode());
     result = prime * result + ((mergeInfo == null) ? 0 : mergeInfo.hashCode());
     result = prime * result + (readOnce ? 1231 : 1237);
+    result = prime * result + (randomAccess ? 1249 : 1259);
```
rmuir, line 145: `load` is missing from hashcode/equals/tostring

ChrisHegarty:
>... Additionally, I assume we can have unit tests for NativeAccess directly. Lemme take a quick look at this.

Dammit! I forgot about how our code is organised. Over in Elasticsearch we added version specific test sets for exactly this reason, so that `test21` could depend upon `java21` (which we call `main21`). For unit tests of NativeAccess, we would need similar.

lucene/core/src/java21/org/apache/lucene/store/PosixNativeAccess.java
```diff
@@ -0,0 +1,123 @@
...
+            .find(name)
+            .orElseThrow(
+                () ->
+                    new UnsupportedOperationException(
+                        "Platform has no symbol for '" + name + "' in libc."));
+    return linker.downcallHandle(symbol, desc);
+  }
+
+  @Override
+  public void madvise(MemorySegment segment, IOContext context) throws IOException {
```
rmuir, line 83: i think this is really strange, that the "native access" takes IOContext and mixes in the "business" logic of mapping IOContext to native integer value. why not make the native access more transparent/straightforward?

rmuir: and no please don't take this as me wanting/asking for more abstractions. maybe if the iocontext was a little less confusing then this would also be less confusing.

uschindler: The problem is that not all platforms have the same constants.... I agree I was also not really deal with that, but this was my best idea.

I was about to define an enum in the base class, but this is an unnecessary abstraction.

Let's go one step after the other. Let's cleanup IOContext for Lucene 10 in a separate issue. Technically, IOContext fits here, it is just the definition.

These are all hidden classes, so it's no public API. I just wanted to have it maintainable.

uschindler:
> > ... Additionally, I assume we can have unit tests for NativeAccess directly. Lemme take a quick look at this.
> 
> Dammit! I forgot about how our code is organised. Over in Elasticsearch we added version specific test sets for exactly this reason, so that `test21` could depend upon `java21` (which we call `main21`). For unit tests of NativeAccess, we would need similar.

When we implemented the MRJAR part for MMapDirectory and Vectorization, I said: Let's have no extra complexity with tests and just *indirectly* test the code. For MMapDirectory this worked fine, because when the test passes, the implementation of the IndexInput is fine.

For vectorization its similar, we have a scalar and a vectorized impl that is compared by tests => also fine.

I agree, here it triggers platform specific madvise(). We may add mocks for testing, but I am not sure if it is worth. Maybe the easiest would be to extend IOContext at a later time to return some madvise-like enum as additional abstraction., so the native code just gets a MAdvice enum constant and maps it straight to integers in a simple `Map<MAdvice,Integer>`.

jpountz:
I'm curious if you've already thought of how we could update the `IOContext` on our merge instances. If I'm not mistaken, given how reader pooling works in `IndexWriter`, if you use a `SegmentReader` for search and then the same segment is included in a merge, then this `SegmentReader` will not use an `IOContext.Context.MERGE` context but a `IOContext.Context.READ` context (because it's reusing the first reader that was created). It would be nice if our file format merge instances (e.g. `StoredFieldsReader#getMergeInstance`) could somehow use a `IOContext.Context.MERGE` even though they might be cloned from a reader that was created with a `IOContext.Context.READ` context. I don't think it should block this change, it's a good step in the right direction, this other problem is quite orthogonal.

uschindler:
I added a basic test that verifies that we have madvise support enabled on Linux and Macos. This also added

One question: In Lucene 9.x we had some system properties to disable MemorySegments. I think in main barnch we can have all of this by default enabled.

But when we backport this to 9.x, I tend to add another system property to disable this feature.

ChrisHegarty:
> I agree, here it triggers platform specific madvise(). We may add mocks for testing, but I am not sure if it is worth. Maybe the easiest would be to extend IOContext at a later time to return some madvise-like enum as additional abstraction., so the native code just gets a MAdvice enum constant and maps it straight to integers in a simple Map<MAdvice,Integer>.

We might be saying the same thing, but I really just wanna get a the `Optional<NativeAccess>` in a unit test and, where present, so as to exercise it's functionalist directly. For now, and it's perfectly fine here, we can test indirectly once we know the various different ways to exercise the code paths. I pushed a trivial test that at least exercises `POSIX_MADV_RANDOM`.

gradle/testing/defaults-tests.gradle
```diff
@@ -132,6 +132,8 @@ allprojects {
       if (rootProject.vectorIncubatorJavaVersions.contains(rootProject.runtimeJavaVersion)) {
         jvmArgs '--add-modules', 'jdk.incubator.vector'
       }
+      
+      jvmArgs '--enable-native-access=' + (project.path == ':lucene:core' ? 'ALL-UNNAMED' : 'org.apache.lucene.core')
```
ChrisHegarty, line 136: 👍

ChrisHegarty:
> But when we backport this to 9.x, I tend to add another system property to disable this feature.

Why is this extra property needed? And not just enable madvise for the MemorySegmentIndexInput on Java 21+ unconditionally? Maybe I'm missing something obvious?

gradle/testing/defaults-tests.gradle
uschindler: 🙏

uschindler:
Do we have any benchmarks by Mike? Some on memory restricted devices. An idea would be a limited VM or by using the memory hog in Mike's benchmark.

lucene/core/src/java/org/apache/lucene/store/IOContext.java
uschindler: I think we can fix this here, but possibly a separate PR would be better. Will do.

uschindler: Here is the PR: #13204

uschindler: Actually we should in main change this class to be a "record". Then equals/hashcode is a no-brainer (as autogenerated by invokedynamic).

rmuir:
It's not the fault of the PR, but it's almost impossible for me to follow what-happens-when thru the iocontext. Maybe we can overhaul it in trunk... I don't think we should add our own layer of abstraction here (I know that's "java"-style to do but it's wrong). It would be clearer if we just used madvise semantics directly: then you could see MADV_RANDOM from the code opening the file and understand what happens.

uschindler:
> It's not the fault of the PR, but it's almost impossible for me to follow what-happens-when thru the iocontext. Maybe we can overhaul it in trunk... I don't think we should add our own layer of abstraction here (I know that's "java"-style to do but it's wrong). It would be clearer if we just used madvise semantics directly: then you could see MADV_RANDOM from the code opening the file and understand what happens.

Please do the change to IOContext separately. We need an abstraction anyways, because not all platforms have same constants. We may add another layer between IOContext and madvise, but that's out of the scope of this issue.

rmuir:
> We need an abstraction anyways, because not all platforms have same constants.

Things can have the same names at least.

uschindler:
Let's meet in the center:
- I will add a public enum `PosixMFAdvise` to store package with enum constants named as the POSIX_MADV / POSIX_FADV constants (they are same, just different numbers)
- IOContext becomes a `toPosixMFAdvise()` method that converts it into that enum (in a similar way like the previously mentioned private/internal method).
- MMapDirectory by defaults uses this method to interpret the IOContext. As it's a `Function<IOContext,PosixMFAdvise>` you can configure it like the other MMapDirectoy. This allows us to play around in benchmarks and also allows to disable it completely.
- Inside the private PosixNativeAccess class we just convert the enum to the low-level integer value.

I can do this later. OK?

ChrisHegarty:
> At a later stage we can ...
> 
> I can do this later. OK?

This seems quite reasonable to me. Native access should be as close to a wrapper around the native functions as possible, so should look and feel familiar to the reader of the code.

As it stands this PR is a significant step forward, we should merge it as is and do further refactoring incrementally.

ChrisHegarty (approved):

rmuir (approved):

rmuir:
I agree, wasn't trying to hold up the change, was just thinking out loud!

uschindler:
Maybe it is easier to see results on benchmarking when it is in main branch. I am waiting for final review by @jpountz and then merge this. Backporting to 9.x is also planned and should be done before #13205 is applied.

I will merge this tomorrow afternoon CET.

uschindler:
@jpountz Are you fine with merging?

jpountz (approved):
Yes indeed!

uschindler:
The test added by @ChrisHegarty sometimes fails on windows: It does not close the file it opened for random access testing, so the directory can't be deleted. Will fix this in a separate commit.

uschindler:
I fixed the test in https://github.com/apache/lucene/commit/ae5d3534e3ef44ff3336dea0308d8a82f0672ff2

uschindler:
I also removed the extra logging included while development from the main branch. In 9.x the log message was adapted to list both features together with the sysprop to disable).

ChrisHegarty:
> The test added by @ChrisHegarty sometimes fails on windows: It does not close the file it opened for random access testing, so the directory can't be deleted. Will fix this in a separate commit.

Oops, sorry about this. Thankfully the fix was straightforward.

uschindler:
> > The test added by @ChrisHegarty sometimes fails on windows: It does not close the file it opened for random access testing, so the directory can't be deleted. Will fix this in a separate commit.
> 
> Oops, sorry about this. Thankfully the fix was straightforward.

About this test and the quite large 8 MB buffer (see https://github.com/apache/lucene/pull/13196/files#diff-ebee319f3691cdb1627e5e9b1dfbdd0c266b1da28ccf0b2a9218dee1d34ff2b7R103):

Is this some limit inside the kernel to trigger something? For this test, we call posix_madvise() always.

Maybe there was a misunderstanding: We only do not call madvise, if the chunk size is too small (< 2 MiB), but by default the chunk size is 16 Gigabytes, except for tests we always call madvise. Theres only one special case: If the file is zero length, then we can't call it.

ChrisHegarty:
I dunno what I was thinking, this is clearly not correct. I opened #13214 to fix the test. ( apologies for the stupid test issues! )

jpountz:
@uschindler Should we open a separate issue for adding `fadvise` support to `NIOFSDirectory`?

uschindler:
Unfortunately fadvise is at moment close to impossible. Reason: we have no file handle!

Chances are good that we also get a Java-based fadvise some time in the future (e.g., through an OpenOption like with O_DIRECT).

uschindler:
As disussed before, for implementing fadvise for reading/writing files, we would need to write a full stack of IO layer natively (OutputStream for writing and FileChannel for NIOFSDir). See https://bugs.openjdk.org/browse/JDK-8292771

uschindler:
Anyways we can open an issue to track what's going on on the JDK (listing all relevant issue numbers like the above one).

uschindler:
> As disussed before, for implementing fadvise for reading/writing files, we would need to write a full stack of IO layer natively (OutputStream for writing and FileChannel for NIOFSDir). See https://bugs.openjdk.org/browse/JDK-8292771

Here is the `posix_fadvise` specific issue.

PR #13197, opened 2024-03-21
Expand scalar quantization with adding half-byte (int4) quantization

benwtrent:
This PR is a culmination of some various streams of work:

 - Confidence interval optimizations, unlocked even smaller quantization bytes.
 - The ability to quantize down smaller than just int8 or int7
 - Adding an optimized int4 (halfbyte) vector API comparison for dot-product.

The idea of further scalar quantization gives users the choice between:

 - Further quantizing to gain space through compressing the bits into single byte values
 - Or allowing quantization to give guarantees around maximal values that afford faster vector operations.

I didn't add more panama vector APIs as I think trying to micro-optimize int4 for anything other than dot-product was a fools errand. Additionally, I only focused on ARM. I experimented with trying to get better performance on other architectures, but didn't get very far, so I fall back to dotProduct.


tteofili:
@benwtrent to me it makes sense to have quantization `bits` configurable in this case.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/OffHeapQuantizedHalfByteVectorValues.java
```diff
@@ -0,0 +1,299 @@
...
+
+  public static void decompressBytes(byte[] compressed, int numBytes) {
+    if (numBytes << 1 != compressed.length) {
+      throw new IllegalArgumentException(
+          "numBytes: " + numBytes + " does not match compressed length: " + compressed.length);
+    }
+    for (int i = 0; i < numBytes; ++i) {
+      compressed[numBytes + i] = (byte) (compressed[i] & 0x0F);
+      compressed[i] = (byte) ((compressed[i] & 0xFF) >> 4);
+    }
```
jpountz, line 54: Cool, this should get auto-vectorized on JDK13+.

jpountz (commented):
For that case, adding an option makes sense to me since it seems extremely similar to int8 scalar quantization.

benwtrent:
I did a bunch of local benchmarking on this. I am adding a parameter to allow optional compression as the numbers without compressing are compelling enough on ARM to justify it IMO. 

To achieve similar recall, `int4` without compression is about 30% faster. With compression its about 30% slower, but with 50% of the memory requirements.

Here are some latency vs. recall for int7, and int4 with this change.

```
plt.plot([1.49, 1.53, 1.54, 1.83, 2.09], [0.952, 0.962, 0.965, 0.974, 0.981], marker='o', label='int7')
plt.plot([1.72, 1.75, 1.79, 2.04, 2.48], [0.897, 0.915, 0.929, 0.971, 0.980 ], marker='o', label='int4_compressed')
plt.plot([1.08, 1.12, 1.12, 1.34, 1.50], [0.897, 0.915, 0.929, 0.971, 0.980 ], marker='o', label='int4')
```
image

int4 with compression gives 2x space improvement over int7, but it comes at an obvious cost as we have to (un)pack bytes during dot-products.

Here are the numbers around index building as well. I committed ever 1MB to ensure merging occurred and that force-merging was adequately exercised. 

Int4 no compression:
```
Indexed 500000 documents in 312090ms
Force merge done in: 76169 ms
```

Int4 compression:
```
Indexed 500000 documents in 326978ms
Force merge done in: 124961 ms
```

Int7:
```
Indexed 500000 documents in 344584ms
Force merge done in: 98311 ms
```

jpountz (commented):
I wonder if this feature should be more opinionated, e.g. should it only accept 4 and 7 as numbers of bits, these look like the two most interesting numbers to me? And maybe we should enforce compression with 4 bits or less, I understand that there is a performance hit, but storing vectors in a wasteful way doesn't feel great?

tteofili:
I tend to agree on being opinionated on a set of allowed configurations for what concerns the number of bits (4 and 7).
Given the speed-space trade-off for packing, I think it's useful to leave that as an option.

tteofili (approved):

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestGenerateBwcIndices.java
```diff
@@ -82,6 +82,16 @@ public void testCreateSortedIndex() throws IOException {
     sortedTest.createBWCIndex();
   }
 
+  public void testCreateInt8HNSWIndices() throws IOException {
```
benwtrent, line 85: @jpountz because the change adds a version metadata difference to Scalar quantized HNSW, I added some BWC tests. I only built BWC indices for 8.10.1. 

The reason for its own BWC class is because the codec here for the particular field isn't the default testing codec, I didn't want to adjust the other tests unnecessarily.

jpountz:
@benwtrent I tried to fix the compilation on luceneutil at https://github.com/mikemccand/luceneutil/commit/027146b05c755c303cac5d95451cc458ed397c51. I could use your help to check if this is the right fix.

benwtrent:
@jpountz I can add the parameters today and fix the compilation. I think your change is the correct one.

PR #13199, opened 2024-03-22
Break point estimate when threshold exceeded

gf2121:
Typically, we estimate the point value count to compare to a threshold and all we need is just a boolean which represents whether the point count is greater than this threshold. This PR proposes to parse the threshold into the intersect logic and break the recursion when the threshold is exceeded.

Dynamic pruning is a case that heavily using estimate point count so i run luceneutil for it. Here is the benchmark result on `wikimedium10m`:

**M2 Chip**
```
           HighTermDayOfYearSort      374.33      (1.8%)      429.74      (0.6%)   14.8% (  12% -   17%) 0.000
                      TermDTSort      388.72      (1.7%)      474.43      (1.5%)   22.1% (  18% -   25%) 0.000
```

**Intel Chip**
```
                      TermDTSort      123.32      (2.6%)      158.34      (2.5%)   28.4% (  22% -   34%) 0.000
           HighTermDayOfYearSort      391.59      (1.8%)      577.73      (2.7%)   47.5% (  42% -   52%) 0.000
```

PS: When profiling i noticed that PointTree construction cost a lot so i tried to make it reusable, this optimization also contributed to this speed-up.


lucene/core/src/java/org/apache/lucene/index/PointValues.java
```diff
@@ -375,16 +375,23 @@ private void intersect(IntersectVisitor visitor, PointTree pointTree) throws IOE
...
-  private long estimatePointCount(IntersectVisitor visitor, PointTree pointTree)
-      throws IOException {
+  /**
+   * Estimate the number of documents that would be matched by {@link #intersect} with the given
+   * {@link IntersectVisitor}. The estimation will terminate when the point count exceeds the upper
+   * bound.
+   *
+   * <p>TODO: Broad-first will help extimation terminate earlier?
+   */
+  public static long estimatePointCount(
```
jpountz, line 393: Nit: To make contracts of these functions clearer, I'd rather make this function private, and them have another `isEstimatedPointCountGreaterThanOrEqualTo` public function (and probably tagged with `@lucene.internal` so that we can evolve it as we want) that calls this private function?

lucene/core/src/java/org/apache/lucene/index/PointValues.java
```diff
@@ -398,8 +405,8 @@ private long estimatePointCount(IntersectVisitor visitor, PointTree pointTree)
         if (pointTree.moveToChild()) {
           long cost = 0;
           do {
-            cost += estimatePointCount(visitor, pointTree);
-          } while (pointTree.moveToSibling());
+            cost += estimatePointCount(visitor, pointTree, upperBound - cost);
+          } while (cost <= upperBound && pointTree.moveToSibling());
```
jpountz, line 409: I believe that we can stop counting if `cost == upperBound`?

```suggestion
          } while (cost < upperBound && pointTree.moveToSibling());
```

lucene/core/src/java/org/apache/lucene/search/comparators/NumericComparator.java
```diff
@@ -301,6 +303,14 @@ public PointValues.Relation compare(byte[] minPackedValue, byte[] maxPackedValue
       updateSkipInterval(true);
     }
 
+    private PointValues.PointTree pointTree() throws IOException {
+      if (pointTree == null) {
+        pointTree = pointValues.getPointTree();
+      }
+      assert !pointTree.moveToParent();
+      return pointTree;
+    }
```
jpountz, line 312: What about pulling the point tree in the constructor instead of doing it lazily (for simplicity)?

jpountz (commented):
This is a big speedup! I left some minor comments.

jpountz (approved):
LGTM

gf2121:
Thanks for review @jpountz !

gf2121:
Nightly benchmark:
https://home.apache.org/~mikemccand/lucenebench/TermDTSort.html
https://home.apache.org/~mikemccand/lucenebench/TermDayOfYearSort.html
https://home.apache.org/~mikemccand/lucenebench/2024.03.26.08.24.36.html

I opened a PR to add annotation:
https://github.com/mikemccand/luceneutil/pull/261/files

PR #13201, opened 2024-03-22
Get better cost estimate on MultiTermQuery over few terms

### LINKED ISSUE #13029: Improve AbstractMultiTermQueryConstantScoreWrapper#RewritingWeight ScorerSupplier cost estimation
### Description

We recently discovered a performance degradation on our project when going from Lucene 9.4 to 9.9. The cause seems to be a side effect of https://github.com/apache/lucene/commit/c6667e709f610669b57a6a1a6ab1cc80f4f0ebaf and https://github.com/apache/lucene/commit/3809106602a9675f4fd217b1090af4505d4ec2a7

The situation is as follows: we have a `WildcardQuery` and a `TermInSetQuery` which are and-combined (within a `BooleanQuery`). This structure gets executed repeatedly, kind of like a nested loop where the `WildcardQuery` remains the same, but the `TermInSetQuery` keeps changing its terms. In the old version, this was fast because the `WildcardQuery` was cached within the `LRUQueryCache`. However in the new version this is no longer the case, so the execution time of our scenario has increased.

Why our `WildcardQuery` is not cached any more? It boils down to [this line](https://github.com/apache/lucene/blob/f16007c3eca7cbf89bf6c2f88907707cf30c0058/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java#L771) in `LRUQueryCache`, where the cache operation won't happen if the cost estimation is too high:
```
final ScorerSupplier supplier = in.scorerSupplier(context);
...
final long cost = supplier.cost();
...
// skip cache operation which would slow query down too much
if (cost / skipCacheFactor > leadCost) {
...
```

Before the upgrade to 9.9, that cost was provided by a `ConstantScoreWeight` returned by the old `MultiTermQueryConstantScoreWrapper` (which was returned by the default `RewriteMethod`), which in the end was just based on the "default" `Weight#scoreSupplier` [implementation](https://github.com/apache/lucene/blob/f16007c3eca7cbf89bf6c2f88907707cf30c0058/lucene/core/src/java/org/apache/lucene/search/Weight.java#L147): basically the cost was the `scorer.iterator().cost();` and in **our case the `WildcardQuery` returns just one document, so cost 1**.

After the upgrade, the default `RewriteMethod` has changed and now this cost is provided by `AbstractMultiTermQueryConstantScoreWrapper#RewritingWeight#scorerSupplier` [here](https://github.com/apache/lucene/blob/f16007c3eca7cbf89bf6c2f88907707cf30c0058/lucene/core/src/java/org/apache/lucene/search/AbstractMultiTermQueryConstantScoreWrapper.java#L263), and for that purpose a private [estimateCost method](https://github.com/apache/lucene/blob/f16007c3eca7cbf89bf6c2f88907707cf30c0058/lucene/core/src/java/org/apache/lucene/search/AbstractMultiTermQueryConstantScoreWrapper.java#L295) was introduced, which bases the estimation on the [MultiTermQuery#getTermsCount](https://github.com/apache/lucene/blob/f16007c3eca7cbf89bf6c2f88907707cf30c0058/lucene/core/src/java/org/apache/lucene/search/MultiTermQuery.java#L316) value. The problem is that, for our `WildcardQuery` (in fact for any sub-class of `AutomatonQuery`), this value is unknown, i.e. `-1`, so the `estimateCost` method just [returns](https://github.com/apache/lucene/blob/f16007c3eca7cbf89bf6c2f88907707cf30c0058/lucene/core/src/java/org/apache/lucene/search/AbstractMultiTermQueryConstantScoreWrapper.java#L309)  `terms.getSumDocFreq()`, which is clearly an overestimation in our case, so it prevents the caching, and leads to a performance degradation.

I understand that I can fix this situation by writing my customized `RewriteMethod`.
The question is: could we improve `AbstractMultiTermQueryConstantScoreWrapper#RewritingWeight#scorerSupplier#cost`  so that, if the MTQ cannot provide a term count (`getTermsCount() == -1`) then we return `scorer.iterator().cost()` ?

msfroh:
### Description

The existing cost estimate for multi-term queries was taking the worst-case scenario (e.g the query matches every term in the field), to provide an upper bound on the cost.

In cases where the MultiTermQuery matches a small number of terms, we can provide a tighter estimate by summing the doc frequencies over those terms.

Fixes #13029


lucene/core/src/java/org/apache/lucene/search/AbstractMultiTermQueryConstantScoreWrapper.java
```diff
@@ -292,7 +292,21 @@ public long cost() {
       };
     }
 
-    private static long estimateCost(Terms terms, long queryTermsCount) throws IOException {
+    private static final int MAX_TERMS_TO_COUNT = 128;
+
+    private long estimateCost(Terms terms, long queryTermsCount) throws IOException {
+
+      // Try to explicitly compute the cost, but only if the term count is sufficiently small.
+      if (queryTermsCount < MAX_TERMS_TO_COUNT) {
```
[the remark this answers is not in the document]
msfroh: Correct -- I thought about making that more explicit, but opted to take advantage of the "fall through" logic.

lucene/core/src/java/org/apache/lucene/search/AbstractMultiTermQueryConstantScoreWrapper.java
```diff
@@ -292,7 +292,21 @@ public long cost() {
       };
     }
 
-    private static long estimateCost(Terms terms, long queryTermsCount) throws IOException {
+    private static final int MAX_TERMS_TO_COUNT = 128;
```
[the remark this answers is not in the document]
msfroh: It's a great question! Maybe we could pass it as a constructor parameter, and the constant `RewriteMethod`s in `MultiTermQuery` could specify the default?

I host a weekly OpenSearch Lucene Study Group on Zoom and we spent this week's meetup talking about this PR (and your issue): https://forum.opensearch.org/t/opensearch-lucene-study-group-meeting-monday-march-25th-2024/18547/3

There were some nice ideas that came up there around how to pick something better than an arbitrary limit -- like maybe using a time threshold instead. (But would that be a threshold per-clause?)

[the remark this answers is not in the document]
rquesada-tibco: That was indeed an interesting discussion. Thanks for sharing @msfroh !

msfroh:
I should check if maybe this has been superseded by https://github.com/apache/lucene/pull/13454

That change seems to look at some of the term states for the first 16 terms during rewrite. Maybe that improves the cost estimate?

Or maybe this change can work on the assumption that the first 16 terms are available and use that to produce a better cost estimate if those (up-to) 16 terms are exhaustive.

msfroh:
There we go -- I reworked the PR to reuse the terms collected during the call to `rewrite` in https://github.com/apache/lucene/pull/13454.

Rather than modifying `estimateCost`, we can say "If the MTQ will rewrite to a BQ, then use the BQ's cost".

lucene/core/src/java/org/apache/lucene/search/AbstractMultiTermQueryConstantScoreWrapper.java
```diff
@@ -154,21 +154,24 @@ protected abstract WeightOrDocIdSetIterator rewriteInner(
         List<TermAndState> collectedTerms)
         throws IOException;
 
-    private IOSupplier<WeightOrDocIdSetIterator> rewrite(LeafReaderContext context, Terms terms)
+    private IOSupplier<WeightOrDocIdSetIterator> rewrite(
+        LeafReaderContext context,
+        Terms terms,
+        List<TermAndState> collectedTerms,
+        boolean[] collectResult)
```
msfroh, line 161: Turning `collectResult` into an out-param (along with `collectedTerms`) feels hacky.

I might try inlining `rewrite` (folding it into `scorerSupplier`) to see how that looks.

msfroh:
@gsmiller, @mayya-sharipova -- What do you both think of this approach? Thanks!

lucene/core/src/java/org/apache/lucene/search/AbstractMultiTermQueryConstantScoreWrapper.java
```diff
@@ -240,9 +219,38 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
...
+      assert terms != null;
+
+      final int fieldDocCount = terms.getDocCount();
+      final TermsEnum termsEnum = q.getTermsEnum(terms);
+      assert termsEnum != null;
+
+      List<TermAndState> collectedTerms = new ArrayList<>();
+      boolean collectResult = collectTerms(fieldDocCount, termsEnum, collectedTerms);
+
+      if (collectResult && collectedTerms.isEmpty()) return null;
```
gsmiller, line 231: nit: stylistically at least, I think we generally prefer always using braces even for one-liners. Mind including them here?

lucene/core/src/java/org/apache/lucene/search/AbstractMultiTermQueryConstantScoreWrapper.java
```diff
@@ -240,9 +219,38 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
...
+
+      final long cost;
+      if (collectResult) {
```
gsmiller, line 234: We end up checking `collectResult` twice in the common case here. We could refactor this slightly to have one conditional block for when `collectResult == true` instead (and pull the `isEmpty` short-circuit check into it). WDYT?

lucene/core/src/java/org/apache/lucene/search/AbstractMultiTermQueryConstantScoreWrapper.java
```diff
@@ -240,9 +219,38 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
...
+        long sumTermCost = 0;
+        for (TermAndState collectedTerm : collectedTerms) {
+          sumTermCost += collectedTerm.docFreq;
```
gsmiller, line 237: Just sort of thinking out loud here, but I wonder how expensive it is to build the boolean query and grab its weight / scoreSupplier at this point (instead of lazily doing it in `scoreSupplier#get`)? If it's cheap, another option would be to invoke `rewriteAsBooleanQuery`, pull its score supplier and then delegate `#cost` to it. That way if the cost logic in `BooleanQuery` evolves in the future, this benefits as well. Probably not worth messing with this now, but maybe worth a TODO comment in here? That is, unless we know it is prohibitively expensive.

gsmiller (commented):
Left a couple very minor comments and thoughts. Thanks @msfroh ! Looks good.

gsmiller:
@msfroh hope you don't mind, but since my PR feedback was pretty minor, I went ahead and made those small changes on your branch and merged. I'll work on getting this backported shortly (a few conflicts to resolve). If you spot anything in the change you disagree with, please let me know and we can work through that. Thanks again!

msfroh:
Thanks, Greg! I can create a follow-up PR with your suggestion on delegating to the rewritten `BooleanQuery` for the cost estimate.

I think if we go down that path (basically doing the rewrite early), we might be able to clean up the code a bit. I think `WeightOrDocIdSetIterator` might become unnecessary (since we'll know ahead of time which one we'll have).

gsmiller:
Thanks @msfroh. Yeah, I think it's worth exploring as long as there isn't something expensive-ish hiding in the weight creation of `BooleanQuery`. Thanks again!

PR #927, opened 2022-05-27
LUCENE-10151: Adding Timeout Support to IndexSearcher 

Deepika0510:
### Problem Statement
Adding optional "timeout" capabilities to `IndexSearcher`. This would enable users to (optionally) specify a maximum time budget for search execution. If the search "times out", partial results would be available. Original Jira [issue](https://issues.apache.org/jira/browse/LUCENE-10151).
### Approach
Created `TimeLimitingBulkscorer` extended from `BulkScorer` class to wrap methods with timeout and delegates the calls to underlying `BulkScorer`. Also created `TestTimeLimitingBulkScorer`. Link for the [JIRA](https://issues.apache.org/jira/browse/LUCENE-10544) where all the approaches have been discussed.
### Note
This implementation is in progress and the current PR consists of basic search functionality with timeout value as additional argument.
### Checklist
Please review the following and check all that apply:

- [x] I have reviewed the guidelines for [How to Contribute](https://github.com/apache/lucene/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.

- [x] I have raised the PR against the main branch.

- [x] I have run ./gradlew check.

- [x] I have added tests for my changes.

PR #13202, opened 2024-03-22
Add timeout support to AbstractKnnVectorQuery

kaivalnp:
### Description

#927 added timeout support to [`IndexSearcher#search`](https://github.com/apache/lucene/blob/75e1ebc4501733a6a9857858b3fd9c30304234d0/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java#L680) where we wrap a scorer in a [`TimeLimitingBulkScorer`](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TimeLimitingBulkScorer.java) which periodically checks whether a query timed out while collecting docs in fixed [chunks](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TimeLimitingBulkScorer.java#L33-L35)

However as #11677 points out, this does not apply to query rewrites -- an important use case being vector searches ([`AbstractKnnVectorQuery`](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java)), where the bulk of computations (HNSW graph searches) happen in [`#rewrite`](https://github.com/apache/lucene/blob/75e1ebc4501733a6a9857858b3fd9c30304234d0/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L68)

If we had set a timeout, and HNSW searches turned out too expensive -- we would not return results even after the search completes (because the timeout is [checked before attempting to score docs](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TimeLimitingBulkScorer.java#L79-L82) -- at which point we have all possible results, but they aren't returned)

In this PR, we're wrapping the `KnnCollector` in another one which additionally checks for `QueryTimeout#shouldExit` in the `KnnCollector#earlyTerminated` function. If this timeout is hit, we immediately return empty results

Also extended this to exact searches, where we check for a timeout before scoring each document


kaivalnp:
### baseline with `null` timeout

```
0.965	 1.10	1000000	50	16	100	1675	0	1.00	post-filter
0.992	 2.82	1000000	400	16	100	4086	0	1.00	post-filter
0.974	 1.21	1000000	50	16	200	1853	0	1.00	post-filter
0.995	 3.10	1000000	400	16	200	4504	0	1.00	post-filter
```

### baseline with `Long.MAX_VALUE` timeout

```
0.965	 1.10	1000000	50	16	100	1675	0	1.00	post-filter
0.992	 2.82	1000000	400	16	100	4086	0	1.00	post-filter
0.974	 1.21	1000000	50	16	200	1853	0	1.00	post-filter
0.995	 3.09	1000000	400	16	200	4504	0	1.00	post-filter
```

### baseline with `1 ms` timeout

```
0.483	 1.08	1000000	50	16	100	1675	0	1.00	post-filter
0.005	 2.78	1000000	400	16	100	4086	0	1.00	post-filter
0.418	 1.18	1000000	50	16	200	1853	0	1.00	post-filter
0.000	 3.05	1000000	400	16	200	4504	0	1.00	post-filter
```

### candidate with `null` timeout

```
0.965	 1.10	1000000	50	16	100	1675	0	1.00	post-filter
0.992	 2.83	1000000	400	16	100	4086	0	1.00	post-filter
0.974	 1.21	1000000	50	16	200	1853	0	1.00	post-filter
0.995	 3.10	1000000	400	16	200	4504	0	1.00	post-filter
```

### candidate with `Long.MAX_VALUE` timeout

```
0.965	 1.17	1000000	50	16	100	1675	0	1.00	post-filter
0.992	 2.99	1000000	400	16	100	4086	0	1.00	post-filter
0.974	 1.28	1000000	50	16	200	1853	0	1.00	post-filter
0.995	 3.28	1000000	400	16	200	4504	0	1.00	post-filter
```

### candidate with `1 ms` timeout

```
0.445	 0.89	1000000	50	16	100	476	0	1.00	post-filter
0.001	 1.01	1000000	400	16	100	12	0	1.00	post-filter
0.374	 0.92	1000000	50	16	200	425	0	1.00	post-filter
0.000	 1.01	1000000	400	16	200	3	0	1.00	post-filter
```

Some observations:
- There is almost no regression when a timeout is not set (= `null` value)
- There is some regression in queries where a timeout is set but not met (\~6%) because of additional time checks on every `#earlyTerminated` call
- There are large gains in queries where a timeout is set and met

vigyasharma:
> If we had set a timeout, and HNSW searches turned out too expensive -- we would not return results even after the search completes (because the timeout is [checked before attempting to score docs](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TimeLimitingBulkScorer.java#L79-L82) -- at which point we have all possible results, but they aren't returned)

Is this because we have already exhausted our query timeout during rewrite, and so we throw a `TimeExceededException` and bail before scoring docs for even the [first chunk](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TimeLimitingBulkScorer.java#L71-L84)?

kaivalnp:
Exactly! This issue is also visible from the above benchmarks, where the `baseline` (having no early termination in `#rewrite`) performs the entire graph search (same number of nodes visited, similar latency) even with a `1 ms` timeout, but the results are eventually discarded -- demonstrated by the low recall compared to high or no timeout

vigyasharma (commented):
Interesting find on KNN search behavior, @kaivalnp! I added some thoughts after a first pass at the PR. 

Apart from those, one divergent behavior is that we won't be raising some form of `TimeExceededException` when we timeout and terminate the search. Wonder if we should have a way to surface that this was a timeout, v/s all the results we could collect within `visitLimit()` node traversal.

vigyasharma:
> ### baseline with `null` timeout
> ```
> 0.965	 1.10	1000000	50	16	100	1675	0	1.00	post-filter
> 0.992	 2.82	1000000	400	16	100	4086	0	1.00	post-filter
> 0.974	 1.21	1000000	50	16	200	1853	0	1.00	post-filter
> 0.995	 3.10	1000000	400	16	200	4504	0	1.00	post-filter
> ```

Could we add headers for each column with these benchmark results please? Thanks!

vigyasharma:
Separately, should we deprecate `TimeLimitingCollector` ? It doesn't use `QueryTimeout` and I don't think we're using it anywhere.

lucene/core/src/java/org/apache/lucene/search/TimeLimitingKnnCollectorManager.java
```diff
@@ -0,0 +1,91 @@
...
+      public float minCompetitiveSimilarity() {
+        return collector.minCompetitiveSimilarity();
+      }
+
+      @Override
+      public TopDocs topDocs() {
+        if (queryTimeout.shouldExit()) {
+          // Quickly exit
+          return TopDocsCollector.EMPTY_TOPDOCS;
+        }
```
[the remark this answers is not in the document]
kaivalnp: Returning partial results makes sense, but I passed the empty results here purely because they would be discarded later (so we could save on computations used for [creating the `TopDocs`](https://github.com/apache/lucene/blob/04f335ad79cae019d1a552e41e0029d061d8e3d9/lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java#L52-L64))

> `TimeLimitingBulkScorer` sees the query as timed out, and doesn't even run the first `interval`
> But for queries with cached results like `DocAndScoreQuery`, maybe we shouldn't?

I don't think returning partial results would add value _today_, can we add a `TODO` to return them once the above behavior is fixed?

> Most KnnCollector implementations (like [this one](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java#L59-L63)) tend to return whatever results we've collected so far, and set the relation as `TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO` if we early terminated

On a separate note, this seems to be a no-op as well -- since we [return fresh results from exact search](https://github.com/apache/lucene/blob/04f335ad79cae019d1a552e41e0029d061d8e3d9/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L140-L145) if approximate results are incomplete. I wonder if we can return empty results here as well to save on some computations?

kaivalnp:
Thanks for the review @vigyasharma!

> Apart from those, one divergent behavior is that we won't be raising some form of `TimeExceededException` when we timeout and terminate the search. Wonder if we should have a way to surface that this was a timeout, v/s all the results we could collect within `visitLimit()` node traversal.

I think this exception will be raised [here](https://github.com/apache/lucene/blob/04f335ad79cae019d1a552e41e0029d061d8e3d9/lucene/core/src/java/org/apache/lucene/search/TimeLimitingBulkScorer.java#L71-L84) automatically, because the timeout has already occurred? Though we won't know / raise this _explicitly because graph search timed out_, but the end goal of marking results as "incomplete" will still be satisfied?

I'm also not sure how we'd do it explicitly, perhaps we require lower-level API changes to allow rewrites to time out more gracefully?

> Could we add headers for each column with these benchmark results please?

Oops, I missed adding the benchmark description and row headers earlier, updated now..

> Separately, should we deprecate `TimeLimitingCollector` ? It doesn't use `QueryTimeout` and I don't think we're using it anywhere.

+1, I only [see](https://github.com/search?q=repo%3Aapache%2Flucene%20TimeLimitingCollector&type=code) occurrences of it in comments. I'll try to create a PR for this soon

lucene/core/src/java/org/apache/lucene/search/TimeLimitingKnnCollectorManager.java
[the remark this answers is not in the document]
vigyasharma: > I don't think returning partial results would add value _today_, can we add a `TODO` to return them once the above behavior is fixed?

Yes, we can handle it in a different change. I'll start a discussion on a separate issue for this.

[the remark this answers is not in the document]
vigyasharma: > On a separate note, this seems to be a no-op as well -- since we [return fresh results from exact search](https://github.com/apache/lucene/blob/04f335ad79cae019d1a552e41e0029d061d8e3d9/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L140-L145) if approximate results are incomplete.

I think we only fall back to exact search if there is a filter provided, i.e. `filterWeight != null` ([code](https://github.com/apache/lucene/blob/04f335ad79cae019d1a552e41e0029d061d8e3d9/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L119-L121)). Otherwise we return the approx search result and callers can use `TotalHits.Relation` to figure out if search terminated early.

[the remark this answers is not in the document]
vigyasharma: I still think `topDocs()` from the collector should return whatever it collected, and set the `relation` to GTE. With `EMPTY_TOPDOCS`, the query gets rewritten as `MatchNoDocsQuery` which is misleads debugging and further masks the timeout occurrence. We can let the results get discarded by `TimeLimitingBulkScorer` for now, while we handle that in a separate change (since it is at par with existing behavior).

I'm open to more opinions on this.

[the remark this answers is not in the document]
kaivalnp: > callers can use `TotalHits.Relation` to figure out if search terminated early.

This information is internal to `AbstractKnnVectorQuery` -- even if we pass some `TotalHits.Relation`, the rewritten `DocAndScoreQuery` treats all results as "complete" (no distinction for the relation [here](https://github.com/apache/lucene/blob/04f335ad79cae019d1a552e41e0029d061d8e3d9/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L239-L254), and does not pass it on to `IndexSearcher#search`)

Though I agree, we could pass those results right now and handle them properly in a follow up issue. I've added a commit to do this

kaivalnp:
> Separately, should we deprecate `TimeLimitingCollector` ? It doesn't use `QueryTimeout` and I don't think we're using it anywhere.

Created #13220 to discuss this

lucene/core/src/java/org/apache/lucene/search/TimeLimitingKnnCollectorManager.java
```diff
@@ -0,0 +1,95 @@
...
+      public TopDocs topDocs() {
+        TopDocs docs = collector.topDocs();
+
+        // Mark results as partial if timeout is met
+        TotalHits.Relation relation =
+            queryTimeout.shouldExit()
+                ? TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO
+                : docs.totalHits.relation;
+
+        return new TopDocs(new TotalHits(docs.totalHits.value, relation), docs.scoreDocs);
```
vigyasharma, line 91: Can we simply return `collector.topDocs();`  and let collectors decide how to handle this? All implementations of `AbstractKnnCollector` already set relation to `GTE` based on `earlyTerminated()` check.

lucene/core/src/test/org/apache/lucene/search/TestKnnByteVectorQuery.java
```diff
@@ -102,14 +103,34 @@ public void testVectorEncodingMismatch() throws IOException {
     }
   }
 
+  public void testTimeout() throws IOException {
```
vigyasharma, line 106: We could also add a test for the partial result case. You could create a mock query timeout that returns `false` only for the first time it's called, and then flips to returning `true`.

lucene/core/src/java/org/apache/lucene/search/TimeLimitingKnnCollectorManager.java
kaivalnp: I don't think the `collector.topDocs()` will set the relation to `GTE` in case of timeouts (it will only set this when the `visitLimit()` is crossed, because it will use its own internal `earlyTerminated()` function that does not have information about the `QueryTimeout`)

lucene/core/src/test/org/apache/lucene/search/TestKnnByteVectorQuery.java
kaivalnp: Makes sense!

There was a consideration here that the number of levels in the HNSW graph should be == 1, because if the timeout is hit while finding the best entry point for the last level, we haven't collected any results yet. I think this should be fine as we're only indexing 3 vectors, and running these tests for a few thousand times did not give an error. Added a note about this as well

Also fixed [another place](https://github.com/apache/lucene/blob/1f909baca5496d5ea7d06d6e83f3b0c54756d8e2/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java#L253-L256) where the timeout needs to be checked

lucene/join/src/java/org/apache/lucene/search/join/DiversifyingChildrenFloatKnnVectorQuery.java
```diff
@@ -100,8 +102,15 @@ protected TopDocs exactSearch(LeafReaderContext context, DocIdSetIterator accept
...
     final int queueSize = Math.min(k, Math.toIntExact(acceptIterator.cost()));
     HitQueue queue = new HitQueue(queueSize, true);
+    TotalHits.Relation relation = TotalHits.Relation.EQUAL_TO;
     ScoreDoc topDoc = queue.top();
     while (vectorScorer.nextParent() != DocIdSetIterator.NO_MORE_DOCS) {
+      // Mark results as partial if timeout is met
+      if (queryTimeout != null && queryTimeout.shouldExit()) {
+        relation = TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO;
+        break;
+      }
```
kaivalnp, line 112: Also wanted some opinions here: we're checking the timeout in exact search of `DiversifyingChildren[Byte|Float]KnnVectorQuery` once per-parent (as opposed to once per document elsewhere)

Should we update this to once per-child as well?

benwtrent:
Looking at the benchmarking, we are adding a 5% overhead to all vector operations when using float32. As vector operations get faster (consider hamming distance with exploring more vectors), I expect this timeout mechanism to have an even bigger impact on search.

What was the dimension count & dataset for your performance testing @kaivalnp ?

Do we really think that timing out a couple of seconds faster is worth the overhead?

kaivalnp:
> What was the dimension count & dataset for your performance testing

It was the enwiki dataset (https://home.apache.org/~sokolov/enwiki-20120502-lines-1k-100d.vec), unit vectors, 100 dimensions, indexed first 1M vectors, searched next 10K vectors, with `DOT_PRODUCT` similarity function

> Looking at the benchmarking, we are adding a 5% overhead to all vector operations when using float32

This overhead is only when a timeout is set but not met, there is no regression when it is not set at all (in the benchmarked case without any filter). IMO this additional check can be useful to time out unexpectedly expensive queries in a production system to keep overall CPU within limits, but looking forward to more opinions here..

benwtrent:
> unit vectors, 100 dimensions, indexed first 1M vectors,

@kaivalnp AH, ok, only 100 dimensions, this explains the overly large impact that checking for timeout had, as there is very little work being done on distance calculations. I would expect the percentage of compute to drop as the amount of work done by vector comparisons to increase.

I have a dumb question about timeouts. I am guessing the `QueryTimeout` object has to do some global syncing to determine the current runtime to check. Is that done on every single vector operation? If so, could we make sure its only done every `N` operations?

Timeouts are already approximate, I understand the need for having a more strict timeout, but I am not sure we should check for timeout on every single vector.

kaivalnp:
> I am guessing the `QueryTimeout` object has to do some global syncing to determine the current runtime to check

I have used the default [`QueryTimeoutImpl`](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/QueryTimeoutImpl.java) class, which makes `System.nanoTime()` calls to check for a timeout at every `shouldExit()` call

> but I am not sure we should check for timeout on every single vector.

Perhaps this can be configured by the end-user themselves, by making actual timeout checks after every TK number of calls, according to acceptable latency / accuracy tradeoffs?

For example, I modified the `shouldExit()` function [here](https://github.com/apache/lucene/blob/69d846f700a66972104ada26f569d34efd14d560/lucene/core/src/java/org/apache/lucene/index/QueryTimeoutImpl.java#L57-L60) from:

```java
return timeoutAt != null && nanoTime() - timeoutAt > 0;
```

to

```java
// count is an integer initialized to 0
return timeoutAt != null && ++count % 100 == 0 && nanoTime() - timeoutAt > 0;
```

to check the timeout every 100 vectors, and the latency overhead with a `Long.MAX_VALUE` timeout got cut in half. Correspondingly, the latency with a `1 ms` timeout increased because we check the timeout less frequently, but all of this is expected

vigyasharma:
> Perhaps this can be configured by the end-user themselves, by making actual timeout checks after every TK number of calls, according to acceptable latency / accuracy tradeoffs?

`TimeLimitingBulkScorer` already optimizes for timeout check frequency outside of `QueryTimeout` impl, by adjusting the interval passed to bulk scorer. I would rather we keep a compatible/similar vector timeout behavior by implementing this in the collector manager, so that users can use the same timeout impl in both places.

That being said, I'm okay if we do it in a separate change.

vigyasharma:
I think this PR's changes are okay to be merged. @kaivalnp: can you please resolve the version conflicts and add a changes entry, and I can merge it in.

lucene/join/src/java/org/apache/lucene/search/join/DiversifyingChildrenFloatKnnVectorQuery.java
vigyasharma: I don't have a strong opinion either way. Also seems like something that we can easily change in a follow up PR is needed.

kaivalnp:
> `TimeLimitingBulkScorer` already optimizes for timeout check frequency outside of `QueryTimeout` impl

Ahh nice catch! You mean something like:

```java
// counter is an integer initialized to 0
return (counter++ == 100 && queryTimeout.shouldExit()) || collector.earlyTerminated();
```

in the `TimeLimitingKnnCollectorManager` itself?

This makes sense to me, we can add this in a follow-up and even make the interval growing like [this commit](https://github.com/apache/lucene/commit/7ce9dbbf65e522f9b463b8518f247a3e7d6332ae)?

lucene/join/src/java/org/apache/lucene/search/join/DiversifyingChildrenFloatKnnVectorQuery.java
kaivalnp: Sounds good, thanks!

vigyasharma (approved):

vigyasharma:
> > `TimeLimitingBulkScorer` already optimizes for timeout check frequency outside of `QueryTimeout` impl
> 
> Ahh nice catch! You mean something like:
> 
> ```java
> // counter is an integer initialized to 0
> return (counter++ == 100 && queryTimeout.shouldExit()) || collector.earlyTerminated();
> ```
> 
> in the `TimeLimitingKnnCollectorManager` itself?
> 
> This makes sense to me, we can add this in a follow-up and even make the interval growing like [this commit](https://github.com/apache/lucene/commit/7ce9dbbf65e522f9b463b8518f247a3e7d6332ae)?

Yes. Let's do it in a follow-up change.

benwtrent:
@kaivalnp I too have had this fail, but it is not deterministic. I created https://github.com/apache/lucene/issues/13272 to track it.

I can mute the test while we wait for the fix.

kaivalnp:
FYI I opened #13285 to add the same timeout support to `AbstractVectorSimilarityQuery`

vigyasharma:
Backported to `branch_9x`

PR #13203, opened 2024-03-23
Speed up writeGroupVInts

easyice:
This change uses `VarHandles` instead of `BytesRefBuilder#append` to speed up `writeGroupVInts`, which is more aligned with group-varint generally implementation.

Main:
```
Benchmark                                (size)   Mode  Cnt  Score   Error   Units
GroupVIntBenchmark.bench_writeGroupVInt      64  thrpt    5  1.050 ± 0.057  ops/us
```

PR:
```
Benchmark                                (size)   Mode  Cnt  Score   Error   Units
GroupVIntBenchmark.bench_writeGroupVInt      64  thrpt    5  2.924 ± 0.049  ops/us
```

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
+++ b/lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
@@ -18,6 +18,7 @@ package org.apache.lucene.util;
 
 import java.io.IOException;
 import org.apache.lucene.store.DataInput;
+import org.apache.lucene.store.DataOutput;
 
 /**
  * This class contains utility methods and constants for group varint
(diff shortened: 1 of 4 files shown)


lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
```diff
@@ -111,4 +112,40 @@ public static int readGroupVInt(
     pos += 1 + n4Minus1;
     return (int) (pos - posStart);
   }
+
+  private static int numBytes(int v) {
+    // | 1 to return 1 when v = 0
+    return Integer.BYTES - (Integer.numberOfLeadingZeros(v | 1) >> 3);
+  }
+
+  public static void writeGroupVInts(DataOutput out, byte[] scratch, long[] values, int limit)
```
jpountz, line 121: Can you add some minimal javadocs?

jpountz (approved):
This looks like a good speedup!

lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
easyice: done!

easyice:
Thanks for reviewing!

jpountz (approved):

PR #13204, opened 2024-03-23
Fix equals/hashcode of IOContext

uschindler:
It also modernizes it. Actualy this class should be changed to be "record", then equals/hashCode is autogenerated and more performant by using invokedynamic.

This was found by @rmuir in #13196


ChrisHegarty (approved):
LGTM.

rmuir (approved):
Thanks!

ChrisHegarty:
Looks like a good candidate for a record class alright, but that can come separate to the bug fix.

uschindler:
Will open a PR directly after this was merged. The change to "record" kills almost half of the code. I just have to restore the missing ctors.

PR #13205, opened 2024-03-23
Convert IOContext, MergeInfo, and FlushInfo to record classes

uschindler:
Followup to #13204: This converts IOContext to a record class. This has several positive effects:
- equals, hashCode and toString are autogenerated by an invokedynamic, so we can't forget to handle it correctly
- it is also immutable
- all parameter checking logic can be moved to some special constructor without any parameters (this is new syntax since Java 17)

The IOContext has some crazy constructors, I left them in, but all have to delegate to the default constructor now.


uschindler:
I needed to change some consumers to call the getter methods instead of accessing fields (which are now hidden).

uschindler:
We should possibly convert other classes to records, too (like FlushInfo or MergeInfo if they are immutable).

uschindler:
I had to fix the missing-doclet javadoc parser to understand records. The easy fix was to treat them as "classes".

lucene/core/src/java/org/apache/lucene/store/IOContext.java
```diff
@@ -111,42 +94,6 @@ private IOContext(Context context, MergeInfo mergeInfo) {
...
-    return "IOContext [context="
-        + context
-        + ", mergeInfo="
-        + mergeInfo
-        + ", flushInfo="
-        + flushInfo
-        + ", readOnce="
-        + readOnce
-        + "]";
+    this(ctxt.context, ctxt.mergeInfo, ctxt.flushInfo, readOnce, ctxt.load);
```
uschindler, line 97: here was a bug in original code. The load info was not copied

uschindler: this ctor should possibly be removed and replaced by a method `asReadOnce()`

uschindler: i removed this ctor and implemented above concept.

uschindler:
I also converted the MergeInfo and FlushInfo classes to records. Very clean now.

rmuir (approved):
thanks for fixing the missing-doclet to deal with records!

uschindler:
> thanks for fixing the missing-doclet to deal with records!

This was a requirement, because building Javadocs failed without that fix. It is really nice that the code throws an exception when it hits an unknown element!

uschindler:
I will apply those changes after #13196, because if doing this before, backporting of the other PR gets harder (#13196 has changes to IOContext, too).

lucene/core/src/java/org/apache/lucene/store/IOContext.java
```diff
@@ -62,91 +55,36 @@ public enum Context {
...
-        + ", mergeInfo="
-        + mergeInfo
-        + ", flushInfo="
-        + flushInfo
-        + ", readOnce="
-        + readOnce
-        + "]";
+  /** Return a copy of this IOContext with {@link #readOnce} set to {@code true}. */
+  public IOContext toReadOnce() {
+    return new IOContext(context, mergeInfo, flushInfo, true, load);
```
jpountz, line 88: Should this set `load` to `false`? The current copy constructor that sets `readOnce` to true also forces `load` to be false.

jpountz (approved):
Much nicer!

ChrisHegarty (approved):
LGTM

lucene/core/src/java/org/apache/lucene/store/IOContext.java
uschindler: I missed to mention this in the issue description: This seems to be a bug in current code, because the copy-ctor did not do what its documentation said.

The code is only used in [VariableGapTermsIndexReader.java](https://github.com/apache/lucene/pull/13205/files#diff-6cc5c40198a0a7cb2746909210a4ad045cb71a970c4d7fa023dbb02079eb33c7) at the moment. I don't know if ever `load=true` there, so changing this here as a bugfix is possibly a noop.

Unless theres a real reason to enforce this, we must document it. What do you think?

uschindler: Actually it would be better to add "modifier" methods like toReadOnce for all booleans?

jpountz: IMO this is a bug in the doc rather than in the code. `load` imlies that the `IndexInput` will be hammered with calls to `seek`/`read`, possibly via the `RandomAccessInput` abstraction, so it's incompatible with `readOnce`. I hope we can refactor all these booleans to an enum as a next step, something like `access = READ_ONCE | READ | LOAD | RANDOM` similarly to the various pre-built `IOContext`s we expose.

uschindler: OK, will change this and the documentation.

Actually so load=true and readOnce=true should be mutually exclusive. I will add a check to the default ctor.

uschindler: I changed this in c2d5f083e5d7ee08fe6cabd9fa2782b5bc7de226.

jpountz (approved):

lucene/core/src/java/org/apache/lucene/store/IOContext.java
jpountz: I opened https://github.com/apache/lucene/issues/13211.

uschindler: Thanks. I fixed the title of the PR, but looks good.

ChrisHegarty:
> Hi @ChrisHegarty,
> can you have another look on the crazy randomAccess flag afetr I merged main into this. Especially the checks in the record's constructor should be checked again. I have no idea what flags are mutually exclusive. Maybe there are more combinations missing.

The record constructor checks, while not pretty, seem right to me. This is a nice cleanup and somewhat easier to reason about.

PR #13206, opened 2024-03-23
Subtract deleted file size from the cache size of NRTCachingDirectory.

jfboeuf:
The size of deleted files is not subtracted from the cache size of NRTCachingDirectory. As a consequence, the cache eventually appears to be full preventing new files from being cached despite there being available memory since files no longer consume memory once they are deleted.
There is a weakness in the fix if the file is concurrently deleted in another thread while being closed but I don't think this pathologic use case deserves the additional synchronization it would require.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/NRTCachingDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/NRTCachingDirectory.java
@@ -73,7 +73,12 @@ public class NRTCachingDirectory extends FilterDirectory implements Accountable 
           new SingleInstanceLockFactory(),
           ByteBuffersDataOutput::new,
           (fileName, content) -> {
-            if (!isCachedFile(fileName)) return null;
+            // Defensive check to handle the case the file has been deleted before this lambda
+            // is called when the IndexOutput is closed. Unsafe in the unlikely case the deletion
+            // happens concurrently on another thread.
+            if (isCachedFile(fileName) == false) {
+              return null;
+            }
             cacheSize.addAndGet(content.size());
             return ByteBuffersDirectory.OUTPUT_AS_MANY_BUFFERS.apply(fileName, content);
           });
@@ -121,7 +126,8 @@ public class NRTCachingDirectory extends FilterDirectory implements Accountable 
     if (cacheDirectory.fileExists(name)) {
       long size = cacheDirectory.fileLength(name);
       cacheDirectory.deleteFile(name);
-      cacheSize.addAndGet(-size);
+      long newSize = cacheSize.addAndGet(-size);
+      assert newSize >= 0;
     } else {
       in.deleteFile(name);
     }
--- a/lucene/core/src/java/org/apache/lucene/store/NRTCachingDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/NRTCachingDirectory.java
@@ -73,6 +73,7 @@ public class NRTCachingDirectory extends FilterDirectory implements Accountable 
           new SingleInstanceLockFactory(),
           ByteBuffersDataOutput::new,
           (fileName, content) -> {
+            if (!isCachedFile(fileName)) return null;
             cacheSize.addAndGet(content.size());
             return ByteBuffersDirectory.OUTPUT_AS_MANY_BUFFERS.apply(fileName, content);
           });
@@ -118,7 +119,9 @@ public class NRTCachingDirectory extends FilterDirectory implements Accountable 
       System.out.println("nrtdir.deleteFile name=" + name);
     }
     if (cacheDirectory.fileExists(name)) {
+      long size = cacheDirectory.fileLength(name);
       cacheDirectory.deleteFile(name);
+      cacheSize.addAndGet(-size);
     } else {
       in.deleteFile(name);
     }
@@ -293,6 +296,10 @@ public class NRTCachingDirectory extends FilterDirectory implements Accountable 
     } catch (@SuppressWarnings("unused") NoSuchFileException | FileNotFoundException e) {
       return false;
     }
+  }
+
+  private synchronized boolean isCachedFile(String fileName) {
+    return cacheDirectory.fileExists(fileName);
   }
 
   private void unCache(String fileName) throws IOException {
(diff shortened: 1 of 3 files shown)


jpountz (commented):
Thanks for catching this. I'm not sure what's the best way of dealing with deletion of files that are open for writing, should we try to make it work properly like you did, or rather fail? @uschindler do you have an opinion?

uschindler:
Hi,

> There is a weakness in the fix if the file is concurrently deleted in another thread while being closed

I don't think that adding synchronization for that is really necessary. Actually NRTCachingDir would use a larger cache for a short amount of time. I prefer to fix NRTCachingDirectory to have an overall better tracking of files. Of course in real filesystems, the size of deleted, but still open files is still counted, but that's hard to implement here without refactoring the whole class to use some "inode" like structure, decoupled from filename.

I wonder why this bug was not seen before? This looks like a serious issue. Maybe it was introduced when we switched to ByteBufferDirectory.

uschindler:
Actually the delete while it gets written to should never appear in Lucene.

The bigger problem is when the file is still open in an NRTReader and gets deleted. I am not sure how we should handle that case? It may be serious depending on size of cache, because servers like Elasticserach or Solr often have files open longer times for reading and IndexWriter deletes them.

Maybe let's think about this a bit more and maybe there's an easy way how to deal with it. Maybe all IndexInputs created by the directory should decrement when they close if the file is "scheduled for delete"?

jpountz:
> The bigger problem is when the file is still open in an NRTReader and gets deleted.

Hmm does it actually happen? I thought index files were ref-counted so that files would only get deleted when the NRT reader gets closed? This is what `StandardDirectoryReader#doClose` suggests.

uschindler:
> > The bigger problem is when the file is still open in an NRTReader and gets deleted.
> 
> Hmm does it actually happen? I thought index files were ref-counted so that files would only get deleted when the NRT reader gets closed? This is what `StandardDirectoryReader#doClose` suggests.

Ah you're right. This is to workaround windows issues.

uschindler:
So I think we are fine then. +1 to merge

lucene/core/src/java/org/apache/lucene/store/NRTCachingDirectory.java
```diff
@@ -73,6 +73,7 @@ public class NRTCachingDirectory extends FilterDirectory implements Accountable
           new SingleInstanceLockFactory(),
           ByteBuffersDataOutput::new,
           (fileName, content) -> {
+            if (!isCachedFile(fileName)) return null;
```
mikemccand, line 76: Could we split this onto three lines and add `{` ... `}` around the body?

And replace the `!` with `== false` to reduce chance of future accidental refactoring bugs?

mikemccand (commented):
Wow, this looks like a serious bug indeed.  Wouldn't this mean that `NRTCachingDirectory` is only doing something on startup, and then gradually reduces the cached files eventually down to no more caching?

lucene/core/src/java/org/apache/lucene/store/NRTCachingDirectory.java
mikemccand: Ahh thank you for the clarification @jpountz.  I am playing catchup on this code ;)  I think I understand it now.

So, this is the `onClose` lambda that is called (when the `IndexOutput` is closed), where we normally increase the `cacheSize` (we do not count the bytes as they are being appended, live, to the file, but rather only when the `IndexOutput` is closed).

And we are (newly, with this PR) checking whether it has already been deleted before it was closed and avoiding adding to `cacheSize` in this case.  Indeed Lucene should never do this, except maybe in exceptional situations (and even then, hopefully not: Lucene should first close any still-open `IndexOutput`s on exception before delete is tried, I think).  +1 to be defensive anyways.

Also this is different from the "delete while file is still open for reading" case...

mikemccand:
> > The bigger problem is when the file is still open in an NRTReader and gets deleted.
> 
> Hmm does it actually happen? I thought index files were ref-counted so that files would only get deleted when the NRT reader gets closed? This is what `StandardDirectoryReader#doClose` suggests.

NRT readers pulled from `IndexWriter` indeed do this ref counting so delete is not attempted until all readers still using that segment/file are closed.

But in the NRT segment replication case, I think this may still happen?  I.e. segments are replicated out to a node that has an open `DirectoryReader`.  That reader is refreshed to switch to the new segments, but the old reader is still alive to finish any in-flight queries or stay open for any long-term leases for deep scrolling purposes.  Oh, nevermind, even in the NRT segment replication case we have `ReplicaFileDeleter` to do the right ref counting.  So yeah I think we don't need to worry about this case either!  Phew, complicated.

lucene/core/src/java/org/apache/lucene/store/NRTCachingDirectory.java
```diff
@@ -118,7 +119,9 @@ public synchronized void deleteFile(String name) throws IOException {
       System.out.println("nrtdir.deleteFile name=" + name);
     }
     if (cacheDirectory.fileExists(name)) {
+      long size = cacheDirectory.fileLength(name);
       cacheDirectory.deleteFile(name);
+      cacheSize.addAndGet(-size);
```
mikemccand, line 124: Could we hold onto the returned (get'd) value and assert that it never drops below zero?

lucene/core/src/java/org/apache/lucene/store/NRTCachingDirectory.java
mikemccand: In addition to the small style comments up above, could you also please add a comment explaining that this check is defensive/paranoia to handle deleting a file that is still open for writing, which Lucene should never do?

jfboeuf: Done!

lucene/core/src/java/org/apache/lucene/store/NRTCachingDirectory.java
jfboeuf: Done

mikemccand (approved):
Thank you @jfboeuf!

uschindler (approved):

uschindler:
Could you add a changes.txt entry in the 9.11 bugfix section? Will merge this PR tomorrow.

uschindler:
I did not notice that changes entry was in wrong section, fixed in main branch (https://github.com/apache/lucene/commit/376ec27a5166ccfad2e68fc22c7543dc38e1fb46). Now backporting.

uschindler:
All backported to 9.x branch.

PR #13208, opened 2024-03-25
Avoid modify merge state in per field mergers

dnhatn:
The PerFieldDocValuesFormat and PerFieldPostingsFormat mutate and reset the fieldInfos of the mergeState during merges. Consequently, other running merge sub-tasks may fail to see some fieldInfos. This was problematic since we introduced concurrency for merge sub-tasks.

Relates #13190

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldMergeState.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldMergeState.java
@@ -31,40 +31,18 @@ import org.apache.lucene.index.IndexOptions;
 import org.apache.lucene.index.MergeState;
 import org.apache.lucene.index.Terms;
 
-/**
- * Utility class to update the {@link MergeState} instance to be restricted to a set of fields.
- *
- * <p>Warning: the input {@linkplain MergeState} instance will be updated when calling {@link
- * #apply(Collection)}.
- *
- * <p>It should be called within a {@code try {...} finally {...}} block to make sure that the
- * mergeState instance is restored to its original state:
- *
- * <pre>
- * PerFieldMergeState pfMergeState = new PerFieldMergeState(mergeState);
- * try {
- *   doSomething(pfMergeState.apply(fields));
- *   ...
- * } finally {
- *   pfMergeState.reset();
- * }
- * </pre>
- */
+/** Utility class to update the {@link MergeState} instance to be restricted to a set of fields. */
 final class PerFieldMergeState {
   private final MergeState in;
-  private final FieldInfos orgMergeFieldInfos;
-  private final FieldInfos[] orgFieldInfos;
-  private final FieldsProducer[] orgFieldsProducers;
+  private final MergeState copy;
 
   PerFieldMergeState(MergeState in) {
     this.in = in;
-    this.orgMergeFieldInfos = in.mergeFieldInfos;
-    this.orgFieldInfos = new FieldInfos[in.fieldInfos.length];
-    this.orgFieldsProducers = new FieldsProducer[in.fieldsProducers.length];
-
-    System.arraycopy(in.fieldInfos, 0, this.orgFieldInfos, 0, this.orgFieldInfos.length);
-    System.arraycopy(
-        in.fieldsProducers, 0, this.orgFieldsProducers, 0, this.orgFieldsProducers.length);
+    this.copy =
+        new MergeState(
+            in,
+            new FieldInfos[in.fieldInfos.length],
+            new FieldsProducer[in.fieldsProducers.length]);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldMergeState.java
```diff
@@ -31,40 +31,18 @@
...
- * }
- * </pre>
- */
+/** Utility class to update the {@link MergeState} instance to be restricted to a set of fields. */
 final class PerFieldMergeState {
   private final MergeState in;
-  private final FieldInfos orgMergeFieldInfos;
-  private final FieldInfos[] orgFieldInfos;
-  private final FieldsProducer[] orgFieldsProducers;
+  private final MergeState copy;
```
jpountz, line 37: Should we dynamically allocate this copy in `apply()` so that each caller gets an independent instance, and we're less subject to a similar kind of bug in the future due to multiple threads trying to modify this `copy` at the same time?

lucene/core/src/java/org/apache/lucene/index/MergeState.java
```diff
@@ -23,13 +23,7 @@
...
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
-import org.apache.lucene.codecs.DocValuesProducer;
-import org.apache.lucene.codecs.FieldsProducer;
-import org.apache.lucene.codecs.KnnVectorsReader;
-import org.apache.lucene.codecs.NormsProducer;
-import org.apache.lucene.codecs.PointsReader;
-import org.apache.lucene.codecs.StoredFieldsReader;
-import org.apache.lucene.codecs.TermVectorsReader;
+import org.apache.lucene.codecs.*;
```
jpountz, line 26: We prefer to avoid star imports.

jpountz (commented):
Thanks Nhat, I had completely forgotten about the fact that we'd mutate the `MergeState` in-place at merge time, this is scary! I like your change better than the current approach, regardless of the concurrency issue.

lucene/core/src/java/org/apache/lucene/index/MergeState.java
dnhatn: @jpountz Sorry, I haven't changed the imports setting for my new Intellij setup.

lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldMergeState.java
dnhatn: ++ I pushed https://github.com/apache/lucene/pull/13208/commits/ee3b0357e07e07cca76662df658d101f927254e4

lucene/core/src/java/org/apache/lucene/index/MergeState.java
```diff
@@ -266,4 +266,35 @@ static PackedLongValues removeDeletes(final int maxDoc, final Bits liveDocs) {
...
+
+  /**
+   * Creates a new state from an existing state with new fieldInfos and FieldsProducer.
+   *
+   * @param state the state to be copied
+   * @param mergeFieldInfos the new field infos of the merged segment
+   * @param fieldInfos the new field infos
+   * @param fieldsProducers the new field producers
+   */
+  public MergeState(
```
jpountz, line 278: I'm not too happy with this constructor being public since it's really an implementation detail of `PerFieldMergeState `. At the same time, I like having the per-field-specific logic in the `.perfield` package rather than here.

I'm not sure how to fix this cleanly... my best idea at this point would be to store all parameters of the main `MergeState` constructor on the `MergeState` object, so that `PerFieldMergeState` can then call this constructor again, wrapping codec readers with a `FilterCodecReader` that only exposes a subset of fields?

dnhatn: That also works for me. I pushed https://github.com/apache/lucene/pull/13208/commits/0f6a566f18521b5810c91442b24a5be7874184bc.

jpountz: I had something different in mind but I realize it wouldn't work without re-computing merge instances over and over again, which we probably don't want. Your approach looks better.

jpountz (approved):

mikemccand:
I just tested this PR on the nightly benchy and indeed it fixes the [merge exception that was blocking the nightly high-cardinality facets test](https://github.com/mikemccand/luceneutil/issues/259)!  Yay, thanks @dnhatn!

dnhatn:
@jpountz @mikemccand Thank you!

benwtrent:
Huzzah! Thanks @dnhatn @jpountz !!!

mikemccand:
Thank you @dnhatn!  I've kicked off a one-off nightly benchy run ...

PR #12172, opened 2023-02-28
Add Romanian stopwords with s&t with comma

Trey314159:
Romanian uses s&t with commas (ș/ț), but for a long time those weren't available, so s&t with cedilla (ş/ţ) were used. Both are still in use, but the comma forms are much more common now. Both should be supported in stopword lists.

PR #13209, opened 2024-03-25
upgrade snowball to 26db1ab9adbf437f37a6facd3ee2aad1da9eba03

rmuir:
I'd like to upgrade snowball to make progress on resolving the Romanian situation, after merging this I'd like to merge #12172 which has been hanging out there FOREVER (over a year).

I've been waiting on an official snowball release, but I've been waiting a long time, and I think we should just move forward? They haven't released since 2021.

This upgrade cuts the size of our "patch" that we apply to snowball's sources in half, as the Armenian and Estonian stemmers that were contributed via patches to lucene are now integrated into snowball codebase. Also the Irish stopwords file was incorporated into snowball codebase.

One downside: it has a somewhat user-impacting change where the `German2` stemmer was folded into the `German` stemmer properly. 

You can read about this more on their website: https://snowballstem.org/algorithms/german2/stemmer.html or via the commit: https://github.com/snowballstem/snowball/commit/b08bdc52e750290eee99373571a8427fa63e691b

To reduce the impact on users, if `German2` is provided, instead of a reflection-exception, we'll map it to `German`, at least for now... we can remove this "if" on our own time? Somehow I suspect Germans will be unhappy with me no matter what I do here :)


rmuir:
The TestModularLayer is angry about the fact that i removed our copy of Irish stopwords to use the snowball one instead. Not sure what is happening here.

rmuir:
ok, i found the issue. I removed the now-unnecessary `opens`. Nice test :)

dweiss (approved):
LGTM.

lucene/analysis/common/src/java/org/apache/lucene/analysis/snowball/SnowballFilter.java
```diff
@@ -70,6 +70,11 @@ public SnowballFilter(TokenStream input, SnowballStemmer stemmer) {
   public SnowballFilter(TokenStream in, String name) {
     super(in);
     Objects.requireNonNull(name, "name");
+    // it was called "German2" for eons, but snowball folded it into "German" and deleted "German2"
+    // for now, don't annoy our users...
+    if (name.equals("German2")) {
```
uschindler, line 75: It's ok to delegate German2 to German.

lucene/analysis/common/src/java/org/apache/lucene/analysis/snowball/SnowballPorterFilterFactory.java
```diff
@@ -59,7 +59,13 @@ public class SnowballPorterFilterFactory extends TokenFilterFactory implements R
   /** Creates a new SnowballPorterFilterFactory */
   public SnowballPorterFilterFactory(Map<String, String> args) {
     super(args);
-    language = get(args, "language", "English");
+    String lang = get(args, "language", "English");
+    // it was called "German2" for eons, but snowball folded it into "German" and deleted "German2"
```
uschindler, line 63: This code is not required as the filter itself renames, but looks still fine to change early.

uschindler (approved):
I did not check the code changes in autogenerated code. Just the double negations are much better.

I agree with the German2 aliazing.

lucene/analysis/common/src/java/org/apache/lucene/analysis/snowball/SnowballPorterFilterFactory.java
rmuir: The filter has two constructors:
```
/** takes explicit instance: we can't provide back compat/user help here unless we generate subclass or something messy */
SnowballFilter(TokenStream, SnowballStemmer)
/** takes string: here we map German2 to German */
SnowBallFilter(TokenStream, String)
```

The factory does reflection itself and uses the first ctor. So it needs the mapping too.

uschindler (approved):

lucene/MIGRATE.md
```diff
@@ -23,6 +23,10 @@
 
 [Apache OpenNLP](https://opennlp.apache.org) 2.x opens the door to accessing various models via the ONNX runtime.  To migrate you will need to update any deprecated OpenNLP methods that you may be using and be running on Java 17.
 
+### Snowball dependency upgrade
+
+Snowball has folded the "German2" stemmer into their "German" stemmer, so there's no "German2" anymore.  For Lucene APIs (TokenFilter, TokenFilterFactory) that accept String, "German2" will be mapped to "German" to avoid breaking users. If you were previously creating German2Stemmer instances, you'll need to change your code to create GermanStemmer instances instead.  For more information see https://snowballstem.org/algorithms/german2/stemmer.html
```
uschindler, line 28: Should we maybe also make a note here that reindexing is required if the Romanian Stemmer was used?

rmuir: I'd like to address the Romanian with one consistent entry after merging a few other PRs, if that works?

rmuir:
one more PR coming as a followup, the snowball maintainers helpfully worked with me to integrate a bunch of our code changes. Time to remove more code!!!!

PR #13216, opened 2024-03-26
Clean up variable-gaps terms format.

jpountz:
The variable-gaps terms format uses the legacy storage layout of storing metadata at the end of the index file, and storing the start pointer of the metadata as the last 8 bytes of the index files (just before the footer). This forces an awkward access pattern at open time when we first need to seek towards the end to check that a footer is present, then seek some more bytes backwards to read metadata, and finally read the content of the index that sits before metadata.

To fix this, meta data and index data are now split into different files. This way, both files have a clean sequential and read-once access pattern, and can take advantage of the `ChecksumIndexInput` abstraction for checksum validation.

This further helps clean up `IOContext` by removing the ability to set `readOnce` to `true` on an existing `IOContext`.


uschindler (commented):
Hi,
I am not fully familar with the index format, so I can't really review the actual read/write code.

My understanding is the following:
- we now use `ChecksumIndexInput` as the file is fully readonce without seeking. Because we removed the `IOContext` from directory's open method (as it is hardcoded to readOnce), we do not need to change the existing IOContext, so the modifier method can go away
- the file is then read in whole and closed, the FST is on-heap.

We do not have backwards compatibility here as it is not the default format, so we can change it in main branch (10.x). This won't be backported, correct?

Uwe

jpountz:
> we now use ChecksumIndexInput as the file is fully readonce without seeking. Because we removed the IOContext from directory's open method (as it is hardcoded to readOnce), we do not need to change the existing IOContext, so the modifier method can go away

This is correct.

> the file is then read in whole and closed, the FST is on-heap.

Correct as well.

> We do not have backwards compatibility here as it is not the default format, so we can change it in main branch (10.x). This won't be backported, correct?

I don't believe that we attempt to not break non-default formats in minor releases, I have broken a few of them in the past in a minor. That said, I'm fine with not backporting, we could still have the `IOContext` ctor removed on 9.x by hardcoding `IOContext.READONCE` when opening the `IndexInput`.

For reference, I bumped the version number to make sure users get an `IndexFormatTooOldException` rather than a cryptic corruption exception.

uschindler:
> > we now use ChecksumIndexInput as the file is fully readonce without seeking. Because we removed the IOContext from directory's open method (as it is hardcoded to readOnce), we do not need to change the existing IOContext, so the modifier method can go away
> 
> This is correct.
> 
> > the file is then read in whole and closed, the FST is on-heap.
> 
> Correct as well.
> 
> > We do not have backwards compatibility here as it is not the default format, so we can change it in main branch (10.x). This won't be backported, correct?
> 
> I don't believe that we attempt to not break non-default formats in minor releases, I have broken a few of them in the past in a minor. That said, I'm fine with not backporting, we could still have the `IOContext` ctor removed on 9.x by hardcoding `IOContext.READONCE` when opening the `IndexInput`.

That's fine. Hardcoding READONCE is also fine for this case, because the MergeContext is not needed at all.

We should at least add a note that the format has changed and older indexes can't be read.

> For reference, I bumped the version number to make sure users get an `IndexFormatTooOldException` rather than a cryptic corruption exception.

Of course, I understood that. Although we provide no backwards compatibility we should always change version numbers when format changes.

jpountz:
I'm merging only to `main` for now.

PR #13219, opened 2024-03-26
Replace boolean flags on `IOContext` with an enum.

### LINKED ISSUE #13211: Replace boolean flags on IOContext with an enum
### Description

`IOContext` has a few boolean flags: `readOnce`, `load`, `randomAccess`. But some combinations of these flags don't make sense, e.g. something can't be `readOnce` and `randomAccess` at the same time. So it would be natural to identify combinations that make sense and represent them as an `enum`.

Separately @rmuir suggested having a more natural translation between `IOContext` and `(m|f)advise`. So maybe this enum could be modeled after the allowed values for `posix_fadvise`:  `NOREUSE` (read-once), `RANDOM`, etc.

jpountz:
This replaces the `load`, `randomAccess` and `readOnce` flags with a `ReadAdvice` enum, whose values are aligned with the allowed values to (f|m)advise.

Closes #13211

### DIFF
--- a/lucene/codecs/src/java/org/apache/lucene/codecs/blockterms/VariableGapTermsIndexReader.java
+++ b/lucene/codecs/src/java/org/apache/lucene/codecs/blockterms/VariableGapTermsIndexReader.java
@@ -27,6 +27,7 @@ import org.apache.lucene.index.CorruptIndexException;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.IndexFileNames;
 import org.apache.lucene.index.SegmentReadState;
+import org.apache.lucene.store.IOContext;
 import org.apache.lucene.store.IndexInput;
 import org.apache.lucene.util.Accountable;
 import org.apache.lucene.util.Accountables;
@@ -53,7 +54,7 @@ public class VariableGapTermsIndexReader extends TermsIndexReaderBase {
             state.segmentInfo.name,
             state.segmentSuffix,
             VariableGapTermsIndexWriter.TERMS_INDEX_EXTENSION);
-    final IndexInput in = state.directory.openInput(fileName, state.context.toReadOnce());
+    final IndexInput in = state.directory.openInput(fileName, IOContext.READONCE);
     boolean success = false;
 
     try {
--- a/lucene/core/src/java/org/apache/lucene/store/IOContext.java
+++ b/lucene/core/src/java/org/apache/lucene/store/IOContext.java
@@ -27,22 +27,11 @@ import java.util.Objects;
  * @param context An object of a enumerator Context type
  * @param mergeInfo must be given when {@code context == MERGE}
  * @param flushInfo must be given when {@code context == FLUSH}
- * @param readOnce This flag indicates that the file will be opened, then fully read sequentially
- *     then closed.
- * @param load This flag is used for files that are a small fraction of the total index size and are
- *     expected to be heavily accessed in random-access fashion. Some {@link Directory}
- *     implementations may choose to load such files into physical memory (e.g. Java heap) as a way
- *     to provide stronger guarantees on query latency.
- * @param randomAccess This flag indicates that the file will be accessed randomly. If this flag is
- *     set, then readOnce will be false.
+ * @param readAdvice Advice regarding the read access pattern. Write operations should disregard
+ *     this field.
  */
 public record IOContext(
-    Context context,
-    MergeInfo mergeInfo,
-    FlushInfo flushInfo,
-    boolean readOnce,
-    boolean load,
-    boolean randomAccess) {
+    Context context, MergeInfo mergeInfo, FlushInfo flushInfo, ReadAdvice readAdvice) {
 
   /**
    * Context is a enumerator which specifies the context in which the Directory is being used for.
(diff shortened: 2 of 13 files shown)


lucene/codecs/src/java/org/apache/lucene/codecs/blockterms/VariableGapTermsIndexReader.java
```diff
@@ -53,7 +54,7 @@ public VariableGapTermsIndexReader(SegmentReadState state) throws IOException {
             state.segmentInfo.name,
             state.segmentSuffix,
             VariableGapTermsIndexWriter.TERMS_INDEX_EXTENSION);
-    final IndexInput in = state.directory.openInput(fileName, state.context.toReadOnce());
+    final IndexInput in = state.directory.openInput(fileName, IOContext.READONCE);
```
jpountz, line 57: A change to this class wouldn't be necessary anymore once #13216 is merged.

lucene/core/src/java/org/apache/lucene/store/IOContext.java
```diff
@@ -54,58 +43,74 @@ public enum Context {
...
+    RANDOM,
+    /** Data is expected to be read sequentially with very little seeking at most. */
+    SEQUENTIAL,
+    /**
+     * Data is treated as random-access memory in practice. {@link Directory} implementations may
+     * explicitly load the content of the file in memory, or provide hints to the system so that it
+     * loads the content of the file into the page cache at open time. This should only be used on
+     * very small files that can be expected to fit in RAM with very high confidence.
+     */
+    LOAD
```
jpountz, line 67: I wonder if there's a better name for this that is more aligned with other constant names. `RANDOM_ACCESS_MEMORY`?

lucene/core/src/java/org/apache/lucene/store/IOContext.java
```diff
@@ -54,58 +43,74 @@ public enum Context {
...
+  public static final IOContext READONCE = new IOContext(ReadAdvice.SEQUENTIAL);
 
-  public static final IOContext READ = new IOContext(false, false, false);
+  public static final IOContext READ = new IOContext(ReadAdvice.NORMAL);
 
-  public static final IOContext LOAD = new IOContext(false, true, true);
+  public static final IOContext LOAD = new IOContext(ReadAdvice.LOAD);
 
-  public static final IOContext RANDOM = new IOContext(false, false, true);
+  public static final IOContext RANDOM = new IOContext(ReadAdvice.RANDOM);
```
jpountz, line 79: FWIW kept these constant names as-is rather than align them with `ReadAdvice` constant names on purpose as they convey stronger expectations.

uschindler: We can still refactor and rename them a bit. The LOAD should really be PRELOAD.

lucene/core/src/java21/org/apache/lucene/store/PosixNativeAccess.java
```diff
@@ -135,12 +135,12 @@ public void madvise(MemorySegment segment, IOContext context) throws IOException
     }
   }
 
-  private Integer mapIOContext(IOContext ctx) {
-    return switch (ctx.readAdvice()) {
+  private Integer mapIOContext(ReadAdvice readAdvice) {
```
uschindler, line 138: should be `mapReadAdvice()` :-)

uschindler (approved):

lucene/core/src/java/org/apache/lucene/store/IOContext.java
```diff
@@ -54,58 +42,50 @@ public enum Context {
...
+      new IOContext(Context.DEFAULT, null, null, ReadAdvice.NORMAL);
 
-  public static final IOContext READONCE = new IOContext(true, false, false);
+  public static final IOContext READONCE = new IOContext(ReadAdvice.SEQUENTIAL);
 
-  public static final IOContext READ = new IOContext(false, false, false);
+  public static final IOContext READ = new IOContext(ReadAdvice.NORMAL);
 
-  public static final IOContext LOAD = new IOContext(false, true, true);
+  public static final IOContext PRELOAD = new IOContext(ReadAdvice.RANDOM_PRELOAD);
```
uschindler, line 52: This may need a MIGRATE.md entry.

jpountz:
> P.S.: Are we using RANDOM at the moment?

Not yet, we'd need to start using it where it makes sense like we do for (PRE)LOAD.

> I also found https://github.com/elastic/elasticsearch/issues/27748, this person suggests to pass RANDOM for everything.

Yeah, Wikimedia also did testing and they [report](https://phabricator.wikimedia.org/T169498) getting best performance with a mmap readahead of 16kB instead of the default of 128kB (it's shared on the same thread). It feels a bit like a bug to me that mmap has such a higher readahead than regular read operations, I wonder if we should recommend lowering this default readahead in our wiki / javadocs instead of trying to work around it by passing RANDOM everywhere. My preference would be to not index too much on how the various hints perform in practice and try to provide what seems to be the correct read advice based on what we know of the access patterns. E.g. postings and doc values data should probably use NORMAL, stored fields, term vectors and vectors data should probably use RANDOM, etc.

uschindler:
> > P.S.: Are we using RANDOM at the moment?
> 
> Not yet, we'd need to start using it where it makes sense like we do for (PRE)LOAD.
> 
> > I also found [elastic/elasticsearch#27748](https://github.com/elastic/elasticsearch/issues/27748), this person suggests to pass RANDOM for everything.
> 
> Yeah, Wikimedia also did testing and they [report](https://phabricator.wikimedia.org/T169498) getting best performance with a mmap readahead of 16kB instead of the default of 128kB (it's shared on the same thread). It feels a bit like a bug to me that mmap has such a higher readahead than regular read operations, I wonder if we should recommend lowering this default readahead in our wiki / javadocs instead of trying to work around it by passing RANDOM everywhere. My preference would be to not index too much on how the various hints perform in practice and try to provide what seems to be the correct read advice based on what we know of the access patterns. E.g. postings and doc values data should probably use NORMAL, stored fields, term vectors and vectors data should probably use RANDOM, etc.

The question that I have about this: How to handle merging then? If we use random access for some files and then want to merge away the segments. As you said before, the problem is with reused NRT readers for merging. I think, we should not hardcode the RANDOM flag now on all files?!

It is good that IOContext with MergeInfo always requires SEQUENTIAL, but is this really used in all cases when we merge? When its hardcoded while opening index files we have a problem. The example of the vargaps reader has exactly that problem: It always uses readOnce.

I think you are more familar with how the merging works, these are just some points to consider.

uschindler (approved):

jpountz:
> The question that I have about this: How to handle merging then?

This is a big question to me too. With reader pooling, if you open a reader and then it gets included in a merge, we'll reuse the same `SegmentReader` and its existing open index inputs, which have likely been open with a `NORMAL` or `RANDOM` advice. Ideally there would be a way for our `getMergeInstance()` APIs to somehow return a clone that has a different read advice.

> It is good that IOContext with MergeInfo always requires SEQUENTIAL, but is this really used in all cases when we merge?

My understanding is that it will only be used if the index input is created with the `IOContext` that is set on the `SegmentReadState` and this reader has been open specifically for merging, said otherwise the index has not been reopened between the time when the segment got written and the time when the same segment got merged away. This is only going to cover a small subset of the segments that we write, this doesn't look good enough to me.

jpountz:
> P.S.: Are we using RANDOM at the moment?

FYI I tried to start switching some files to it at #13222 and discussed some limitations.

PR #13220, opened 2024-03-26
Mark TimeLimitingCollector as deprecated

kaivalnp:
### Description

Follow-up to https://github.com/apache/lucene/pull/13202#issuecomment-2016947960

[`TimeLimitingCollector`](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TimeLimitingCollector.java) only seems to be used from comments and tests. Can we mark it as `@Deprecated` and remove from `main` in another PR?

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
+++ b/lucene/core/src/java/org/apache/lucene/index/ExitableDirectoryReader.java
@@ -31,8 +31,6 @@ import org.apache.lucene.util.automaton.CompiledAutomaton;
  * {@link QueryTimeout} implementation object to be checked periodically to see if the thread should
  * exit or not. If {@link QueryTimeout#shouldExit()} returns true, an {@link ExitingReaderException}
  * is thrown.
- *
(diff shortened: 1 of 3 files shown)


vigyasharma:
As an alternate to deprecating this entirely, we could also change it to start using `QueryTimeout`, instead of its custom time limiting counters. I'd like to get more opinions from the community, on whether we think it's useful to keep a time limiting *collector* around.

jpountz:
It makes sense to me to deprecate it, `IndexSearcher#setTimeout` should be used instead.

jpountz (approved):

jpountz:
@vigyasharma I'd like to double check with you if you're good with deprecating before merging, I'm not sure if your previous comment was candidly checking for consensus, or if you were implicitly raising concerns.

jpountz:
Thanks for confirming, I'll merge.

kaivalnp:
Created a follow-up to remove `TimeLimitingCollector` from `main` after deprecation: #13243

PR #13222, opened 2024-03-26
Use `ReadAdvice#RANDOM` when appropriate.

jpountz:
This switches the following files to `ReadAdvice.RANDOM`:
 - Stored fields data file.
 - Term vectors data file.
 - HNSW graph.
 - Temporary file storing vectors at merge time that we use to construct the merged HNSW graph.
 - Vector data files, including quantized data files.

I hesitated using `IOContext.RANDOM` on terms, since they have a random access pattern when running term queries, but a more sequential access pattern when running multi-term queries. I erred on the conservative side and did not switch them to `IOContext.RANDOM` for now.

For simplicity, I'm only touching the current codec, not previous codecs. There are also some known issues:
 - These files will keep using a `RANDOM` `IOContext` at merge time. We need some way for merge instances to get an updated `IOContext`? We have the same problem with `IOContext#LOAD` today.
 - With quantized vectors, raw vectors don't have random access pattern, but it was challenging to give raw vectors a sequential access pattern when there are quantized vectors and a random access pattern otherwise. So they assume a random access pattern all the time.


ChrisHegarty (approved):
LGTM

uschindler:
Hi, I have some problem regarding merging with it - and a suggestion. Please hold with merging.

jpountz:
No worries, Uwe. Looking forward to your suggestions.

uschindler (commented):
Hi,
sorry for the delay, was a bit busy. I clicked through the code and constructors with my Eclipse IDE and figured out that the hardcoded RANDOM context does not fit well with merging (at least for stored fields).

Merging also uses the normal `Lucene90CompressingStoredFieldReader` ctor, it only sets some "merge=true" as constant.

The `state.iocontext` is correctly initialized for merging (`IOContext.Context#MERGE`), but this is forcefully changed to RANDOM. So actually this patch makes all merging use RANDOM and the pages may not be freed afterwards (see my comment about this below).

What we should better do: The idea in the variable gaps reader before #13216 looks like the solution here: Let's add a method `IOContext#forRandomAccess()` that does the following: If we are in `Context.MERGE` or `Context#FLUSH` this method just returns the original context unmodified, only in cases `Context#READ` or `Context#DEFAULT` it returns a `IOContext#RANDOM`.

By that it is a bit more dynamic: When we merge or flush we keep the context as is (so information about how large the merge is and so on), but on default/read we enable RANDOM access.

We should then maybe also remove the public `RANDOM` constant in `IOContext`?

What do you think?

uschindler:
In addition, we should still look into getting the IOContexts correct when we merge. The current solution is not ideal, but somehow not really changeable. When you clone an indexinput, you can't set another read advice, as the underlying mmaped segments are the same in the clone.

uschindler:
About the announced comment: When we merge we want to use sequential, as the kernel may earlier free the pages. But actually I am not sure, if we really need this: After merging the file is closed and deleted later. Then the pages get freed automatically, so I don't understand the whole issue that Mike has seen.

jpountz:
> What do you think?

I had similar thoughts is mind, so that sounds good to me.

I'm still curious about how to fix the bigger issue wrt reader pooling. Should `getMergeInstance()` open fresh new `IndexInput`s? Or maybe we avoid reusing readers across searching and merging, and we'll open a fresh new reader if we need to merge, even though the same segment may already be open for search?

> After merging the file is closed and deleted later. Then the pages get freed automatically, so I don't understand the whole issue that Mike has seen.

This only happens after the whole merge is finished. So if you're merging a 5GB segment, and it takes time, it's still a fair amount of page cache that merges are taking from searches. If pages could be more aggressively reclaimed while the merge is running, this could help?

jpountz:
I opened #13229, is it what you had in mind?

jpountz:
@uschindler I'm curious what you think of this PR now that `ReadAdvice#RANDOM` is the default. Should we keep calling `state.context.withReadAdvice(ReadAdvice.RANDOM)` to be explicit about the expected access pattern (what the PR is doing now) or should we lean on the default being `ReadAdvice.RANDOM`? In the latter case, we would only need to keep the changes in `Lucene99FlatVectorsWriter`.

uschindler (approved):
Hi,
as said in the other PR, I would still call it explicit if we know that the access will be really RANDOM. The default (which is now configurable after #13264) is then applied only when not explicitly requested.

uschindler:
> I'll look into porting this work to branch_9x via a separate PR.

We should maybe also port some recent changes back, too. Not sure how to best do this. The new record classes and enums make this hard.

Maybe easiest is to use a method like `IOContext#withRandomAccess()`.

jpountz:
This is exactly what I'm doing. :) All in the same PR, but it shouldn't be too big.

jpountz:
Backport PR: #13267.

PR #13223, opened 2024-03-27
Recommend lowering the default mmap readahead.

jpountz:
This is a follow-up of a discussion on #13219. `mmap` has a higher readahead than regular `read()` operations by default, e.g. 128kB instead of 16kB on my Linux box. On indexes that exceed the size of the page cache, this may trigger performance issues due to page cache trashing and additional page cache contention. Rather than forcing `MMapDirectory` to use `MADV_RANDOM` on all files, it would make more sense to configure a lower `mmap` readahead at the system level, e.g. the same readahead value as `read()` operations use.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
@@ -38,6 +38,16 @@ import org.apache.lucene.util.Constants;
  * fragmented address space. If you get an {@link IOException} about mapping failed, it is
  * recommended to reduce the chunk size, until it works.
  *
+ * <p><b>NOTE</b>: On some platforms like Linux, mmap comes with a higher readahead than regular
+ * read() operations, e.g. 128kB for mmap reads and 16kB for regular reads. Such a high default
+ * readahead can hurt performance on indexes that exceed the size of the page cache, by triggering
+ * more page cache trashing as well as more contention on the page cache. For typical workloads
+ * involving Lucene indexes, it is recommended to update the default readahead of mmap to the same
+ * readahead that is used by regular read() operations. This can be done by calling
+ * <code>sudo blockdev --setra 32 &lt;filesystem&gt;</code>, where 32 is the number of 512 byte
+ * sectors to read ahead, ie. 16kB, and &lt;filesystem&gt; is the filesystem where Lucene indexes
+ * are stored.
+ *
  * <p>This class supports preloading files into physical memory upon opening. This can help improve
  * performance of searches on a cold page cache at the expense of slowing down opening an index. See
  * {@link #setPreload(BiPredicate)} for more details.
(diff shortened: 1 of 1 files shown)


uschindler (approved):
If you fix the html violations all fine.

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -38,6 +38,15 @@
  * fragmented address space. If you get an {@link IOException} about mapping failed, it is
  * recommended to reduce the chunk size, until it works.
  *
+ * <p><b>NOTE</b>: On some platforms like Linux, mmap comes with a higher readahead than regular
+ * read() operations, e.g. 128kB for mmap reads and 16kB for regular reads. Such a high default
```
rmuir, line 42: Can you point to place in the kernel where this is happening?

rmuir (changes requested):
I don't think we should ask users to modify these settings on their block devices, at least i'd like to see actual documentation on why this should be adjusted for lucene (also it will impact the entire system)

uschindler:
I am also a bit skeptical why you need to modify the block device. If this would be a file system setting I can imagine it's useful.

@rmuir this came from investigation by Wikimedia on Elasticsearch https://github.com/elastic/elasticsearch/issues/27748

rmuir:
my thoughts here are that issues can be addressed by providing correct advice to `madvise`. IMO this should typically be `MADV_RANDOM` because accesses are in random order: even if "we" think of it as sequential, we are sharing single filemap across multiple threads and they seek to some random place in file and read a little and are done, that's RANDOM!

we should benchmark this stuff to get it right.

jpountz:
For reference, this change is based on similar observations as made on https://biriukov.dev/docs/page-cache/3-page-cache-and-basic-file-operations. `mmap` comes with a 128kB readahead while `read()` only does 16kB readahead. I can reproduce the exact same numbers with code as below, plus dropping caches and using `vmtouch`.

```java
    try (FSDirectory dir = FSDirectory.open(Paths.get("/data/a")); // switch to NIOFSDirectory to test with read()
        IndexInput in = dir.openInput("term-ids__47.tmp", IOContext.READ)) {
      in.readInt();
    }
  }
```

While 16kB has proved workable in practice, we've seen major performance issues with Elasticsearch, a 128kB readahead and indexes that exceed the size of the page cache. My first take was that 128kB feels huge for a default readahead, almost buggy, and it's not clear to me why it's so much higher than with read(). Since this is controversial, I'm ok with the alternative approach of using a MADV_RANDOM all the time for `IOContext.READ`. We should benchmark the impact of a smaller readahead to confirm it performs well, from my testing it only reads one page at a time in that case, but intuitively it should be ok.

jpountz:
> the alternative approach of using a MADV_RANDOM all the time for IOContext.READ

I opened #13244 to show what this could look like.

mikemccand:
The Linux source for readahead is quite wild (WARNING: GPL 2 code -- read at your own risk!): https://github.com/torvalds/linux/blob/master/mm/readahead.c

jpountz:
Superseded by #13244.

PR #13224, opened 2024-03-27
Use singleton for all-zeros DirectMonotonicReader.Meta

original-brownbear:
Having a single block of all zeros is a fairly common case that is using a lot of heap for duplicate instances in some use-cases in ES. => read a singleton for it to save the duplication


lucene/core/src/java/org/apache/lucene/util/packed/DirectMonotonicReader.java
```diff
@@ -39,6 +39,9 @@ public final class DirectMonotonicReader extends LongValues implements Accountab
    * from disk.
    */
   public static class Meta implements Accountable {
+
+    private static final Meta SINGLE_ZERO_BLOCK = new Meta(1L, 16);
```
benwtrent, line 43: My concern is that nothing is preventing other things in this package from mutating the internal arrays (they are package private, not actually private). 

Can we make them all members of the Meta class private and final?

From what I can tell, they are only ever used in the `DirectMonotonicReader` and indeed only mutated once during construction.

original-brownbear: sure thing private it is :)

ChrisHegarty (approved):
LGTM

jpountz:
The idea makes sense to me, but the fact that we're checking for a specific `blockShift` of `16` looks fragile to me. If codecs change the value of `blockShift` tomorrow, this will break this optimization and we won't know about it. I haven't tested it, but I believe that we don't actually need to test for a specific `blockShift` value and could do something like below instead (should be seen as a quick hack):

```diff
diff --git a/lucene/core/src/java/org/apache/lucene/util/packed/DirectMonotonicReader.java b/lucene/core/src/java/org/apache/lucene/util/packed/DirectMonotonicReader.java
index 1f5954ec05f..af43da8b948 100644
--- a/lucene/core/src/java/org/apache/lucene/util/packed/DirectMonotonicReader.java
+++ b/lucene/core/src/java/org/apache/lucene/util/packed/DirectMonotonicReader.java
@@ -39,6 +39,10 @@ public final class DirectMonotonicReader extends LongValues implements Accountab
    * from disk.
    */
   public static class Meta implements Accountable {
+
+    // Use a shift of 63 so that there would be a single block regardless of the number of values.
+    private static final Meta SINGLE_ZERO_BLOCK = new Meta(1, 63);
+
     private static final long BASE_RAM_BYTES_USED =
         RamUsageEstimator.shallowSizeOfInstance(Meta.class);
 
@@ -49,19 +53,32 @@ public final class DirectMonotonicReader extends LongValues implements Accountab
     final byte[] bpvs;
     final long[] offsets;
 
-    Meta(long numValues, int blockShift) {
-      this.blockShift = blockShift;
+    public static Meta of(long numValues, int blockShift) {
       long numBlocks = numValues >>> blockShift;
       if ((numBlocks << blockShift) < numValues) {
         numBlocks += 1;
       }
-      this.numBlocks = (int) numBlocks;
+      return new Meta(Math.toIntExact(numBlocks), blockShift);
+    }
+
+    private Meta(int numBlocks, int blockShift) {
+      this.blockShift = blockShift;
+      this.numBlocks = numBlocks;
       this.mins = new long[this.numBlocks];
       this.avgs = new float[this.numBlocks];
       this.bpvs = new byte[this.numBlocks];
       this.offsets = new long[this.numBlocks];
     }
 
+    public boolean isFullOfZeroes() {
+      for (int i = 0; i < numBlocks; ++i) {
+        if (mins[i] != 0 || bpvs[i] != 0 || avgs[i] != 0) {
+          return false;
+        }
+      }
+      return true;
+    }
+
     @Override
     public long ramBytesUsed() {
       return BASE_RAM_BYTES_USED
@@ -79,13 +96,16 @@ public final class DirectMonotonicReader extends LongValues implements Accountab
    */
   public static Meta loadMeta(IndexInput metaIn, long numValues, int blockShift)
       throws IOException {
-    Meta meta = new Meta(numValues, blockShift);
+    Meta meta = Meta.of(numValues, blockShift);
     for (int i = 0; i < meta.numBlocks; ++i) {
       meta.mins[i] = metaIn.readLong();
       meta.avgs[i] = Float.intBitsToFloat(metaIn.readInt());
       meta.offsets[i] = metaIn.readLong();
       meta.bpvs[i] = metaIn.readByte();
     }
+    if (meta.isFullOfZeroes()) {
+      return Meta.SINGLE_ZERO_BLOCK;
+    }
     return meta;
   }
 

```

benwtrent:
I agree with @jpountz 's concern here. But I think the `zero` checks can be done as things are read in and we can return the static object.

original-brownbear:
Sounds good @jpountz and @benwtrent, how about [f924708](https://github.com/apache/lucene/pull/13224/commits/f924708da93b14a99ca9aa312bab675782b0ddac) for a simpler version of the above that avoid an extra loop? :)

benwtrent (approved):
Looks good. Could you add a CHANGES entry under `optimizations` for 9.11?

original-brownbear:
Thanks @benwtrent! Done in [e579101](https://github.com/apache/lucene/pull/13224/commits/e579101a98c18194e9a455d010efce53a332d698)

jpountz:
Could you add a test when the `blockShift` that is passed at write time is less than `log2(numValues)` to make sure that returning a reader that has a different number of blocks is fine? Otherwise this looks great.

original-brownbear:
@jpountz I did my best: [478f6c3](https://github.com/apache/lucene/pull/13224/commits/478f6c3006e6abc43ac939507be8ce46207e40ec) hope I got this right, I could only really get this path exercised with `0` values? Is that the kind of test you're looking for?

lucene/core/src/test/org/apache/lucene/util/packed/TestDirectMonotonic.java
```diff
@@ -154,6 +154,43 @@ public void testConstantSlope() throws IOException {
...
+        w.add(0);
+      }
+      w.finish();
+      dataLength = dataOut.getFilePointer();
+    }
+
+    try (IndexInput metaIn = dir.openInput("meta", IOContext.READONCE);
+        IndexInput dataIn = dir.openInput("data", IOContext.DEFAULT)) {
+      DirectMonotonicReader.Meta meta =
+          DirectMonotonicReader.loadMeta(metaIn, numValues, blockShift);
```
jpountz, line 182: Can you assert that the whole meta file has been read after this call?

jpountz (approved):
Thanks, this is exactly what I had in mind, just left a minor comment.

lucene/core/src/test/org/apache/lucene/util/packed/TestDirectMonotonic.java
jpountz: Maybe we could also assert that this returns the `SINGLE_ZERO_BLOCK` instance.

original-brownbear: Sure thing, asserted that it's the singleton by reading twice since we made everything private :)

lucene/CHANGES.txt
```diff
@@ -249,6 +249,8 @@ Optimizations
 
 * GITHUB#13203: Speed up writeGroupVInts (Zhang Chao)
 
+* GITHUB#13224: Use singleton for single block, all-zeros DirectMonotonicReader.Meta (Armin Braun)
```
jpountz, line 252: nit: it no longer focuses on the single-block case?

jpountz (approved):

lucene/CHANGES.txt
original-brownbear: ++ removed the "single block" :)

PR #13225, opened 2024-03-27
Test KNN query works seamlessly regardless of underlying format

tteofili:
### Description

This introduces just a test to check that a `KnnVectorQuery` runs the same when a same field is indexed with different `KnnVectorFormats` (e.g. `Lucene99HnswVectorsFormat` and `Lucene99HnswScalarQuantizedVectorsFormat`).

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
+++ b/lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
@@ -956,8 +956,8 @@ abstract class BaseKnnVectorQueryTestCase extends LuceneTestCase {
     try (Directory directory = newDirectory()) {
       MockAnalyzer mockAnalyzer = new MockAnalyzer(random());
       IndexWriterConfig iwc = newIndexWriterConfig(mockAnalyzer);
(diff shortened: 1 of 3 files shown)


lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
```diff
@@ -949,4 +951,58 @@ public int hashCode() {
       return 31 * classHash() + docs.hashCode();
     }
   }
+
+  public void testSameFieldDifferentFormats() throws IOException {
```
mayya-sharipova, line 955: +1 on the test itself. I wonder if a better place for it would be in a different test file (e.g. `TestLucene99HnswVectorsFormat`). I like how `BaseKnnVectorQueryTestCase` is not dependant on the format.

tteofili: thanks @mayya-sharipova. I do see the point.
Another option might be to extend `LuceneTestCase` to have some `#randomVectorFormat` method which gives random `KnnVectorFormats` to be used in the test.

mayya-sharipova (approved):
Thanks @tteofili , LGTM

PR #13227, opened 2024-03-27
upgrade snowball to 34f3612e5e8c (round two)

rmuir:
Forever we've accumulated various performance differences to the snowball code.

These are now integrated upstream, see https://github.com/snowballstem/snowball/pull/195

Since we aren't modifying this snowball code in crazy ways, all the dictionary-based testing is removed: snowball tests this stuff itself well.

I will be following up with more changes, to try to simplify how we integrate with snowball, by sending upstream more odds-and-ends until we can get all this minimal.


rmuir:
@uschindler How can i exclude the org.tartarus code from this check (or really all checks) without touching it?

```
Forbidden method invocation: java.util.Arrays#copyOf(**) [Prefer using ArrayUtil as Arrays#copyOf fills zeros for bad bounds]
> Task :lucene:analysis:common:forbiddenApisMain FAILED
  in org.tartarus.snowball.SnowballProgram (SnowballProgram.java:337)
Scanned 653 class file(s) for forbidden API invocations (in 0.81s), 1 error(s).
```

uschindler:
> @uschindler How can i exclude the org.tartarus code from this check (or really all checks) without touching it?
> 
> ```
> Forbidden method invocation: java.util.Arrays#copyOf(**) [Prefer using ArrayUtil as Arrays#copyOf fills zeros for bad bounds]
> > Task :lucene:analysis:common:forbiddenApisMain FAILED
>   in org.tartarus.snowball.SnowballProgram (SnowballProgram.java:337)
> Scanned 653 class file(s) for forbidden API invocations (in 0.81s), 1 error(s).
> ```

Yes you can, something like that in forbiddenapis.gradle:

```
allprojects { prj ->
  plugins.withId("java", {
     //.....
    if (prj.path==:lucene:analysis:common') {
      forbiddenApisMain.exclude('org/tartarus/**')
    }
    //....
  }
}
```

rmuir:
thanks, i will do this. It isn't an option to call lucene ArrayUtil methods etc from snowball code.

uschindler:
> thanks, i will do this. It isn't an option to call lucene ArrayUtil methods etc from snowball code.

The only downside is that we can't detect anymore if snowball code calls unsafe shit like forgetting locales or charsets.

rmuir:
> The only downside is that we can't detect anymore if snowball code calls unsafe shit like forgetting locales or charsets.

And we also won't if it starts being published as jar and used as a normal dependency. It isn't our code.

rmuir:
I know, that seems crazy, but I think it would be the ultimate goal. Will have to find an "easy" / "dead-simple" way to achieve publishing snowball properly that avoids craziness of maven etc. They have a simple make-based build and support MANY languages, but e.g. they publish their python bindings to pypi.

For now I'm just trying to just deal with smaller changes, reduce the size of these differences and remove hacks and stuff.

dweiss (approved):

uschindler (approved):

ojwb:
> The only downside is that we can't detect anymore if snowball code calls unsafe shit like forgetting locales or charsets.

Speaking for snowball, we'd likely take patches adding useful code quality checks.  If they're rather dependency heavy we could do them only in CI.

PR #13229, opened 2024-03-27
More consistently use a SEQUENTIAL ReadAdvice for merging.

jpountz:
Merging `IOContext`s use a `SEQUENTIAL` `ReadAdvice`. However, some file formats hardcode `IOContext.LOAD` for some of their files, which silences the whole merging context, in particular the `SEQUENTIAL` `ReadAdvice`.

This PR switches file formats to
`ioContext.toReadAdvice(ReadAdvice.RANDOM_PRELOAD)` so that merges will use a `SEQUENTIAL` `ReadAdvice` while searches will use a `RANDOM_PRELOAD` `ReadAdvice`.

This is not a huge deal for `RANDOM_PRELOAD`, which is only used for very small files. However, this change becomes more relevant for the new `RANDOM` `ReadAdvice` as we would like merges to keep using a `SEQUENTIAL` `ReadAdvice`.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/IOContext.java
+++ b/lucene/core/src/java/org/apache/lucene/store/IOContext.java
@@ -49,10 +49,6 @@ public record IOContext(
 
   public static final IOContext READ = new IOContext(ReadAdvice.NORMAL);
 
-  public static final IOContext PRELOAD = new IOContext(ReadAdvice.RANDOM_PRELOAD);
-
-  public static final IOContext RANDOM = new IOContext(ReadAdvice.RANDOM);
-
   @SuppressWarnings("incomplete-switch")
   public IOContext {
     Objects.requireNonNull(context, "context must not be null");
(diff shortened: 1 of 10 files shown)


lucene/core/src/java/org/apache/lucene/store/IOContext.java
```diff
@@ -88,4 +84,18 @@ public IOContext(MergeInfo mergeInfo) {
...
   }
+
+  /**
+   * Return an updated {@link IOContext} that has the provided {@link ReadAdvice} if the {@link
+   * Context} is a {@link Context#READ} context, otherwise return this existing instance. This helps
+   * preserve a {@link ReadAdvice#SEQUENTIAL} advice for merging, which is always the right choice,
+   * while allowing {@link IndexInput}s open for searching to use arbitrary {@link ReadAdvice}s.
+   */
+  public IOContext toReadAdvice(ReadAdvice advice) {
+    if (context == Context.READ) {
```
uschindler, line 95: How should we deal with DEFAULT. I see lots of code using DEFAULT, so I would suggest:
- Remove the enum constant altogether
- Do the same like for READ

I think DEFAULT is mostly used for writing, but I haven't checked closely.

uschindler: I think "withReadAdvice" sounds better, because we do not convert to a readAdvice, but we create a new context with different readAdvice. Maybe native speakers can suggest a better name.

uschindler:
Looks fine now. How should we handle the DEFAULT context?

I think for the purpose here it can be seen as READ.

jpountz:
> Looks fine now. How should we handle the DEFAULT context?

I'm looking into a (separate) change that renames it to `WRITE` and enforces that it's never used for reading. Let's see how it goes. :)

uschindler (approved):

jpountz:
> I'm looking into a (separate) change that renames it to WRITE and enforces that it's never used for reading.

I opened #13236.

uschindler:
Thanks.

PR #13231, opened 2024-03-27
remove now-unnecessary snowball mojibake hack

rmuir:
Remove this hack, to reduce more logic in this script.

It is no longer needed as of https://github.com/snowballstem/snowball-website/commit/b934d6b565e268b3db080140cc145f532cd6e648


rmuir:
Give me some time for a couple more PRs to get this shell script doing less, and I think we'll be able to totally nuke it. 

I kept the `iconv` check here for UTF-8 correctness because such a thing can be done from gradle/java too, it is just safety.

PR #13232, opened 2024-03-28
Introduce singleton for PackedInts.NullReader of size 256

original-brownbear:
Size 256 is very common here through the monotonic long values default page size. In ES we're seeing many MB O(10M) of duplicate instances of this size relatively quickly.
=> adding a singleton for it to save some heap


lucene/core/src/java/org/apache/lucene/util/packed/PackedInts.java
```diff
@@ -566,10 +566,20 @@ public String toString() {
   /** A {@link Reader} which has all its values equal to 0 (bitsPerValue = 0). */
   public static final class NullReader extends Reader {
 
+    private static final NullReader DEFAULT_MONOTONIC_LONG_VALUES_PAGE_SIZE =
+        new NullReader(MonotonicLongValues.DEFAULT_PAGE_SIZE);
```
jpountz, line 570: Can you use `PackedLongValues.DEFAULT_PAGE_SIZE` instead (the parent class)? (and rename the constant)

Separately I wonder if we should override `#ramBytesUsed()` to return 0 on this instance now that it's a singleton.

jpountz (approved):
Thanks, it looks pretty straightforward. Can you add a CHANGES entry?

lucene/core/src/java/org/apache/lucene/util/packed/PackedInts.java
original-brownbear: Renamed + made it return `0` now. Makes sense to me :)

original-brownbear:
Thanks for taking a look Adrien!

>  Can you add a CHANGES entry?

Done :) Hope 9.11.0 was correct?

lucene/core/src/java/org/apache/lucene/util/packed/PackedInts.java
original-brownbear: hmm this causes an annoying test failure .. let me see what I can do about it

original-brownbear: [95fb423](https://github.com/apache/lucene/pull/13232/commits/95fb423e182bc51bd0a3e6bb71f2ae759ee597ec) is pretty much the best I could come up with for the test, wdyt? I don't have a good intuition if the accuracy gain on the ram bytes method is worth the test noise I must admit.

jpountz: Hmm agreed. Sorry for the suggestion. Let's not tweak ramBytesUsed then?

original-brownbear: No worries! ++ not worth the hassle :) -> Reverted the change

PR #13235, opened 2024-03-28
Fix vector type check for diversified knn search

benwtrent:
I repeatably saw some test failures related to `TestParentBlockJoin[Byte|Float]KnnVectorQuery#testVectorEncodingMismatch`. This commit fixes those test failures and actually checks the field type.


benwtrent:
@bugmakerrrrrr ping

lucene/join/src/java/org/apache/lucene/search/join/DiversifyingChildrenByteKnnVectorQuery.java
```diff
@@ -140,6 +140,10 @@ protected TopDocs approximateSearch(
     if (collector == null) {
       return NO_RESULTS;
     }
+    if (context.reader().getByteVectorValues(field) == null) {
```
jpountz, line 143: For my understanding, why do we only perform this validation when byte vector values are `null`?

Also I wonder if we should move this code higher in the method, validation should still run even though the collector might be `null`?

benwtrent: Ah, I could move it higher. 

> why do we only perform this validation when byte vector values are null?

The contract is really that codec EITHER has byte vector values or float vector values.

Null is returned:

 - if the field is missing
 - If there are no vectors in the field at all (dims == 0)
 - If the storage is vectors is the incorrect element type (byte vs. float)

So, if `getByteVectorValues` actually returns non-null, we know that byte vector search is valid and thus we can continue on.

However, if its `null`, we must verify the other conditions aren't the reason why it is null.

jpountz: I'm missing something: how would it differ from calling `ByteVectorValues.checkField(context.reader(), field);` without the enclosing `if` statement?

benwtrent: yeah, you are correct. I shouldn't even grab the vector values at all here, just check the field. 

For the other queries, it makes sense as we also use the vector values later, but here we do not.

jpountz (approved):

bugmakerrrrrr:
@benwtrent Sorry for being unresponsive. The code looks good to me. But I am wondering if we should move related check to the beginning of  `rewrite` to make the knn query fail fast, as sometimes the filter query may be costly.

PR #13236, opened 2024-03-28
Split IOContext.DEFAULT into WRITE and READ.

jpountz:
This effectively removes the `DEFAULT` `IOContext` and forces callers to use the existing `IOContext.READ` context or the new `IOContext.WRITE` context instead.

This comes with some interesting changes:
 - `MockDirectoryWrapper` fails when using a writing context for reading or a reading context for writing.
 - `Directory.copyFrom` no longer takes an `IOContext`.
 - `LuceneTestCase#newIOContext` has been split into `newReadIOContext` and `newWriteIOContext`.


jpountz:
For what it's worth, the change is huge because lots of tests used to use `IOContext.DEFAULT` and had to be fixed. If you filter the diff to only look at the non-test source folders, this change is pretty contained.

uschindler:
Hi,
I was thinking about it. Why do we need to differentiate between read and write? If you open an IndexInput you read, if you open IndexOutput you're going to write. If you copy, you will do both and this sequential.

How about removing READ instead and name it DEFAULT? The others are FLUSH and MERGE.

In the future we may also have a WriteAdvice enum so we have all coordinates for handling IO.

uschindler:
Actually my suggestion is now a rather simple rename so should be easy to apply with IDE. The diff should also get shorter.

jpountz:
That should work as well, let me give it a try.

jpountz:
@uschindler See #13242.

uschindler:
Perfect. Let's close this PR as won't fix.

jpountz:
Closed in favor of #13242.

PR #13237, opened 2024-03-28
Use FieldInfo vector similarity in knn readers

benwtrent:
Both the KnnWriters & FieldInfo keep track of the vector similarity used by a given field. This commit ensures they are the same and utilizes the FieldInfo one (which, while these are enums, are exactly the same).

### DIFF
--- a/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene90/Lucene90HnswVectorsReader.java
+++ b/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene90/Lucene90HnswVectorsReader.java
@@ -158,7 +158,7 @@ public final class Lucene90HnswVectorsReader extends KnnVectorsReader {
         throw new CorruptIndexException("Invalid field number: " + fieldNumber, meta);
       }
 
-      FieldEntry fieldEntry = readField(meta);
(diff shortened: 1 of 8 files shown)


lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene90/Lucene90HnswVectorsReader.java
```diff
@@ -200,9 +200,18 @@ private VectorSimilarityFunction readSimilarityFunction(DataInput input) throws
     return VectorSimilarityFunction.values()[similarityFunctionId];
   }
 
-  private FieldEntry readField(DataInput input) throws IOException {
+  private FieldEntry readField(IndexInput input, FieldInfo info) throws IOException {
     VectorSimilarityFunction similarityFunction = readSimilarityFunction(input);
-    return new FieldEntry(input, similarityFunction);
+    if (similarityFunction != info.getVectorSimilarityFunction()) {
+      throw new IllegalStateException(
```
jpountz, line 206: We usually throw `CorruptIndexException` for such inconsistencies.

jpountz (approved):
That makes sense. I guess that future vector formats should no longer serialize the similarity function and just rely on the one that is recorded on field infos.

tteofili (approved):

PR #13238, opened 2024-03-28
GITHUB-13218: Add migrate entry for Collector to CollectorManager migration

zacharymorn:
This is a PR to add migration entry for `Collector` to `CollectorManager` migration for `IndexSearch#search`. 

Thanks @mikemccand for the suggestions on entry content!

### DIFF
--- a/lucene/MIGRATE.md
+++ b/lucene/MIGRATE.md
@@ -190,11 +190,11 @@ enum.
 `IndexSearch#search(Query, Collector)` is now being deprecated in favor of `IndexSearcher#search(Query, CollectorManager)`,
 as `CollectorManager` implementation would allow taking advantage of intra-query concurrency via its map-reduce API design.
 To migrate, use a provided `CollectorManager` implementation that suits your use cases, or change your `Collector` implementation
-to follow the new API pattern. The straight forward approach would be to wrap the single-threaded `Collector` into a degenerate `CollectorManager`.
+to follow the new API pattern. The straight forward approach would be to instantiate the single-threaded `Collector` in a wrapper `CollectorManager`.
 
 For example
 ```java
-public class WrappingCollectorManager implements CollectorManager<CustomCollector, List<Object>> {
+public class CustomCollectorManager implements CollectorManager<CustomCollector, List<Object>> {
     @Override
     public CustomCollector newCollector() throws IOException {
       return new CustomCollector();
@@ -211,7 +211,7 @@ public class WrappingCollectorManager implements CollectorManager<CustomCollecto
     }
 }
 
-List<Object> results = searcher.search(query, new WrappingCollectorManager());
+List<Object> results = searcher.search(query, new CustomCollectorManager());
 ```
 ## Migration from Lucene 9.0 to Lucene 9.1
 
--- a/lucene/MIGRATE.md
+++ b/lucene/MIGRATE.md
@@ -185,6 +185,34 @@ enum.
 
 `IOContext#LOAD` has been replaced with `IOContext#PRELOAD`.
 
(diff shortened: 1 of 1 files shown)


lucene/MIGRATE.md
```diff
@@ -185,6 +185,34 @@ enum.
 
 `IOContext#LOAD` has been replaced with `IOContext#PRELOAD`.
 
+### IndexSearch#search(Query, Collector) being deprecated in favor of IndexSearcher#search(Query, CollectorManager) (LUCENE-10002)
+
+`IndexSearch#search(Query, Collector)` is now being deprecated in favor of `IndexSearcher#search(Query, CollectorManager)`,
+as `CollectorManager` implementation would allow taking advantage of intra-query concurrency via its map-reduce API design.
+To migrate, use a provided `CollectorManager` implementation that suits your use cases, or change your `Collector` implementation
+to follow the new API pattern. The straight forward approach would be to wrap the single-threaded `Collector` into a degenerate `CollectorManager`.
```
jpountz, line 193: Not a native speaker, but the use of `degenerate` looks surprising to me in that context?

lucene/MIGRATE.md
```diff
@@ -185,6 +185,34 @@ enum.
...
+
+For example
+```java
+public class WrappingCollectorManager implements CollectorManager<CustomCollector, List<Object>> {
```
jpountz, line 197: nit: maybe call it `CustomCollectorManager` to align the name with the name of the collector?

jpountz (approved):
I left minor comments, but the entry looks good to me!

lucene/MIGRATE.md
zacharymorn: Ah yes let me use a different word here.

lucene/MIGRATE.md
zacharymorn: Done.

lucene/MIGRATE.md
```diff
@@ -185,6 +185,34 @@ enum.
 
 `IOContext#LOAD` has been replaced with `IOContext#PRELOAD`.
 
+### IndexSearch#search(Query, Collector) being deprecated in favor of IndexSearcher#search(Query, CollectorManager) (LUCENE-10002)
+
+`IndexSearch#search(Query, Collector)` is now being deprecated in favor of `IndexSearcher#search(Query, CollectorManager)`,
```
mikemccand, line 190: Maybe say "has been removed" instead of "is now being deprecated"?  Since this is the `MIGRATE.md` for 10.0.0, won't (will?) we have removed the `Collector` based search API?

mikemccand (commented):
Thanks @zacharymorn -- I left a small comment.  I love the short code sample -- very helpful for users who might just want to "upgrade quickly, add concurrency later".

lucene/MIGRATE.md
zacharymorn: Thanks Mike for the suggestion! I wonder if `has been deprecated` (instead of `has been removed`) would be better here, if for some reasons we are not able to completely delete the method and its usage by `10.0.0`? `deprecated` also sometimes imply `removed`, so we don't need to update this entry either if we are able to remove the method by `10.0.0`?

mikemccand: OK good point, let's just leave the wording as is?

mikemccand (approved):
Thanks @zacharymorn

lucene/MIGRATE.md
zacharymorn: Sounds good, thanks Mike!

zacharymorn:
After merge, I saw a test run error and this issue is fixing it https://github.com/apache/lucene/issues/13272

PR #13239, opened 2024-03-29
upgrade icu4j to 74.2

rmuir:
Current ICU version is ancient (70.1): Upgrade to the latest 74.2 release supporting Unicode 15.1
Practically, because `testEmojiFromTheFuture()` passes, we don't behave that out-of-date to end-users.
But the algorithms and code here get maintained, improved, etc, bugs get fixed, we should upgrade.
It would also be confusing for 10.0 to support older unicode version than the JDK itself.

* Regenerate all related files, I forced this with `--rerun-tasks` just to be sure. I had to bump groovy version so that regeneration works with java 21.
* Update RBBI grammar, i just replaced with the new one, disabled dictionary break/rules, and added comment on how to do this. All other changes except this were autogenerated.


lucene/analysis/common/src/test/org/apache/lucene/analysis/email/TLDs.txt
```diff
@@ -1,6 +1,5 @@
 # Generated from IANA TLD Database (gradlew generateTlds).aaa
 aarp
-abarth
```
mikemccand, line 3: Curious that all these TLDs are deleted...

lucene/analysis/icu/src/data/uax29/Default.rbbi
```diff
@@ -17,6 +17,12 @@
 # This file is from ICU (with some small modifications, to avoid CJK dictionary break,
 # and status code change related to that)
 #
+# To update this file: grab rule file corresponding to your ICU version, e.g.:
+#   https://github.com/unicode-org/icu/blob/release-74-2/icu4c/source/data/brkitr/rules/word.txt
+#
+# * Prevent dictionary break: disable final rule that would chain han and kana
+# * For kana rules, change from 400 to 300 status (since there's no dictionary break)
```
mikemccand, line 24: Thank you for adding these instructions!

Where do we use this `Default.rbbi`?  Is it for users to pass to `ICUTokenizer` to run (our slightly modified) UAX #29 tokenization?

mikemccand (approved):
Thanks for upgrading our Unicode impl @rmuir!

lucene/analysis/icu/src/data/uax29/Default.rbbi
rmuir: This grammar is not used by default. This grammar is used only when the user overrides defaults to disable ICU's dictionary-based CJK segmentation:

```
@param cjkAsWords true if cjk text should undergo dictionary-based segmentation, otherwise text
        will be segmented according to UAX#29 defaults. If this is true, all Han+Hiragana+Katakana
        words will be tagged as IDEOGRAPHIC.
```

https://github.com/apache/lucene/blob/main/lucene/analysis/icu/src/java/org/apache/lucene/analysis/icu/segmentation/DefaultICUTokenizerConfig.java#L94C1-L96C46

e.g. the user can configure `false` here and use `CJKBigramFilter` if they prefer an n-gram approach.

lucene/analysis/common/src/test/org/apache/lucene/analysis/email/TLDs.txt
rmuir: I just regenerate, it has not been run since September 2022, which is one reason we see many differences.

PR #13242, opened 2024-03-29
Replace (IO)Context#READ with (IO)Context.DEFAULT.

jpountz:
`DEFAULT` doesn't mean much today and could be used whenever `READ` is used. So let's use `DEFAULT` all the time instead and remove `READ`.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/IOContext.java
+++ b/lucene/core/src/java/org/apache/lucene/store/IOContext.java
@@ -36,9 +36,11 @@ public record IOContext(
    * Context is a enumerator which specifies the context in which the Directory is being used for.
    */
   public enum Context {
+    /** Context for reads and writes that are associated with a merge. */
     MERGE,
-    READ,
+    /** Context for writes that are associated with a segment flush. */
     FLUSH,
+    /** Default context, can be used for reading or writing. */
     DEFAULT
   };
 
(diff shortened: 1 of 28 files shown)


lucene/core/src/java/org/apache/lucene/store/IOContext.java
```diff
@@ -87,12 +86,13 @@ public IOContext(MergeInfo mergeInfo) {
...
-   * while allowing {@link IndexInput}s open for searching to use arbitrary {@link ReadAdvice}s.
+   * Context} is a {@link Context#DEFAULT} context, otherwise return this existing instance. This
+   * helps preserve a {@link ReadAdvice#SEQUENTIAL} advice for merging, which is always the right
+   * choice, while allowing {@link IndexInput}s open for searching to use arbitrary {@link
+   * ReadAdvice}s.
    */
   public IOContext withReadAdvice(ReadAdvice advice) {
-    if (context == Context.READ) {
+    if (context == Context.DEFAULT) {
       return new IOContext(advice);
```
uschindler, line 96: This is a bit unrelated, but here we could use precalculated DEFAULT IOContexts for each advice, to not create new instances every time. Will check in  a separate PR.

uschindler (approved):
Perfect! 😍

lucene/core/src/java/org/apache/lucene/store/IOContext.java
uschindler: see #13245

PR #13243, opened 2024-03-29
Delete TimeLimitingCollector after deprecation

kaivalnp:
### Description

Follow up to #13220: Delete `TimeLimitingCollector` on `main` after deprecation


jpountz:
Can you add an entry to `lucene/MIGRATE.txt` suggesting to use `IndexSearcher#setTimeout` instead?

kaivalnp:
Makes sense, added now!

PR #13245, opened 2024-03-29
Optimization: Use precalculated IOContexts for withReadAdvice() to not create new instances all the time

uschindler:
Followup of #13242: This optimizes the `IOContext#withReadAdvice()` method to use precalculated instances of `IOContext` to not create new instances all the time.


uschindler:
I think last one is shortest. EnumMap would have been fine, but as it is all internal cache a simple array on enum's ordinal number is enough (same approach like switch statements on enums use).

jpountz (approved):

PR #13246, opened 2024-03-29
Simplify bytes arrays in NumericLeafComparator to long

gf2121:
Simplify complex byte array comparisons as long comparison.


lucene/core/src/java/org/apache/lucene/search/comparators/NumericComparator.java
```diff
@@ -321,42 +307,52 @@ private void updateSkipInterval(boolean success) {
...
-     * could better tune the {@link NumericLeafComparator#maxValueAsBytes}/{@link
-     * NumericLeafComparator#minValueAsBytes}. For instance, if the sort is ascending and bottom
+     * could better tune the {@link NumericLeafComparator#maxValueAsLong}/{@link
+     * NumericLeafComparator#minValueAsLong}. For instance, if the sort is ascending and bottom
      * value is 5, we will use a range on [MIN_VALUE, 4].
      */
     private void encodeBottom() {
+      if (queueFull == false) {
+        return;
+      }
```
jpountz, line 317: This looks a bit hacky to me, let's instead fix the call site to only call `encodeBottom()` if the queue is full?

lucene/core/src/java/org/apache/lucene/search/comparators/NumericComparator.java
```diff
@@ -321,42 +307,52 @@ private void updateSkipInterval(boolean success) {
...
     /**
      * If {@link NumericComparator#pruning} equals {@link Pruning#GREATER_THAN_OR_EQUAL_TO}, we
-     * could better tune the {@link NumericLeafComparator#maxValueAsBytes}/{@link
-     * NumericLeafComparator#minValueAsBytes}. For instance, if the sort is ascending and top value
+     * could better tune the {@link NumericLeafComparator#minValueAsLong}/{@link
+     * NumericLeafComparator#minValueAsLong}. For instance, if the sort is ascending and top value
      * is 3, we will use a range on [4, MAX_VALUE].
      */
     private void encodeTop() {
+      if (leafTopSet == false) {
```
jpountz, line 338: and likewise here, let's only call this method if `leafTopSet` is true?

jpountz (commented):
Thanks for extracting this change. I left some comments but it looks ok to me. FWIW I'm leaning towards only including it in 10.0 since it's a big change to the comparator API.

lucene/core/src/java/org/apache/lucene/search/comparators/NumericComparator.java
```diff
@@ -85,6 +83,16 @@ public void disableSkipping() {
...
 
+  protected abstract long missingValueAsComparableLong();
+
+  private static long sortableBytesAsLong(byte[] bytes) {
+    return switch (bytes.length) {
+      case 4 -> NumericUtils.sortableBytesToInt(bytes, 0);
+      case 8 -> NumericUtils.sortableBytesToLong(bytes, 0);
+      default -> throw new IllegalStateException("bytes count should be 4 or 8");
+    };
+  }
```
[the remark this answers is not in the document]
gf2121: Good catch! There is no restriction on how numbers encoded as point values so i make this method abstract and delegate it to implementations.

lucene/core/src/java/org/apache/lucene/search/comparators/DoubleComparator.java
```diff
@@ -59,6 +59,11 @@ protected long missingValueAsComparableLong() {
     return NumericUtils.doubleToSortableLong(missingValue);
   }
 
+  @Override
+  protected long sortableBytesToLong(byte[] bytes) {
+    return NumericUtils.sortableBytesToInt(bytes, 0);
```
jpountz, line 64: Hmm, should be `long` since doubles are 8-bytes?

```suggestion
    return NumericUtils.sortableBytesToLong(bytes, 0);
```

lucene/core/src/java/org/apache/lucene/search/comparators/IntComparator.java
```diff
@@ -58,6 +59,11 @@ protected long missingValueAsComparableLong() {
     return missingValue;
   }
 
+  @Override
+  protected long sortableBytesToLong(byte[] bytes) {
+    return NumericUtils.sortableBytesToLong(bytes, 0);
```
jpountz, line 64: ToInt?

jpountz (commented):
Out of curiosity, are you seeing a speedup from this change?

gf2121:
> Out of curiosity, are you seeing a speedup from this change?

On `wikimedium10m`, this patch is slightly faster with little P-value:
```
                      TermDTSort      275.90      (1.2%)      278.55      (0.8%)    1.0% (  -1% -    3%) 0.003
           HighTermDayOfYearSort      459.59      (1.4%)      465.90      (1.1%)    1.4% (  -1% -    3%) 0.000
```

lucene/core/src/java/org/apache/lucene/search/comparators/NumericComparator.java
```diff
@@ -85,6 +82,10 @@ public void disableSkipping() {
     pruning = Pruning.NONE;
   }
 
+  protected abstract long missingValueAsComparableLong();
+
+  protected abstract long sortableBytesToLong(byte[] bytes);
```
jpountz, line 87: Can you add minimal javadocs?

jpountz (approved):

PR #13253, opened 2024-03-30
Avoid reload block when seeking backward in SegmentTermsEnum.

vsop-479:
In `SegmentTermsEnum.seekExact` and `SegmentTermsEnum.seekCeil`, if we seek a smaller target after seeking a greater one, we set `nextEnt` to -1 in `SegmentTermsEnumFrame.rewind`. Which result in `Force reload` a loaded block.

I think we could just set `nextEnt` to 0 to let seek from start, and temporarily set `entCount` to last `nextEnt` to reduce the seeking range, and reset `suffixesReader` and `suffixLengthsReader` 's position. But don't need to reload the block again.


vsop-479:
@mikemccand 
Do you think this change is feasible?
Though this change is still in progress, it would be nice to get your opinion.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnumFrame.java
```diff
@@ -284,6 +284,61 @@ void rewind() {
     */
   }
 
+  // Only rewind, don't force reload block.
+  // Reset reader position, don't read, decompress.
+  // Current term greater than target, reduce endCount.
+  void rewind2() {
```
mikemccand, line 290: Maybe rename to `rewindWithoutReload` or so?  Or if we think we can fully replace all `rewind` with this new approach then we can keep the name `rewind` :)

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
```diff
@@ -309,6 +309,7 @@ private boolean setEOF() {
   @Override
   public boolean seekExact(BytesRef target) throws IOException {
 
+    int originEndCount = -1;
```
mikemccand, line 312: Rename to `origEntCount`?

mikemccand (commented):
This is a nice idea @vsop-479!  It would speed up cases where the caller seeks backwards a bit, remaining in the current frame.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnumFrame.java
vsop-479: Got it.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
vsop-479: I will rename it.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnumFrame.java
```diff
@@ -284,6 +284,61 @@ void rewind() {
...
+
+    // Force reload:
+    fp = fpOrig;
+    // TODO: Reset entCount after this seek.
+    entCount = nextEnt;
```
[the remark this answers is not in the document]
vsop-479: Thanks @mikemccand , I will fix it.

vsop-479:
@mikemccand 

mikemccand:
This might be a needle-moving optimization for apps that reuse a single `TermsEnum` and seek randomly to terms, right?  Because all up and down the stack of `SegmentTermsEnumFrame`s we can `rewindWithoutReload` for many/most of them?

vsop-479:
> This might be a needle-moving optimization for apps that reuse a single TermsEnum and seek randomly to terms, right?

Yes.

> Because all up and down the stack of SegmentTermsEnumFrames we can rewindWithoutReload for many/most of them?

I just implemented `rewindWithoutReload` for non-floor block and first floor block, in current version. Since `target` may exists in more upper floor when it less than last `term`, but `lastFrame` is non-first floor.

mikemccand:
Thanks @vsop-479 I will try to re-engage here soon!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
```diff
@@ -434,8 +436,29 @@ public boolean seekExact(BytesRef target) throws IOException {
...
+        // We got lastFrame by comparing target and term, and target less than last seeked term in
+        // currentFrame. If lastFrame's fp is same with currentFrame's fp, we can reduce entCount to
+        // nextEnt.
+        boolean currentIsLast = currentFrame.fp == lastFrame.fp;
         currentFrame = lastFrame;
-        currentFrame.rewind();
+
+        // Only rewindWithoutReload for non-floor block or first floor block.
+        // TODO: We need currentFrame's first entry to judge whether we can rewindWithoutReload for
+        // non-first floor blocks.
```
mikemccand, line 448: The FST terms index has a lower bound `BytesRef` pointer to the start term of all floor multi-blocks, but I don't think we save this anywhere today.  Let's leave it as a future `TODO`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
```diff
@@ -710,8 +761,28 @@ public SeekStatus seekCeil(BytesRef target) throws IOException {
         // System.out.println("  target is before current (shares prefixLen=" + targetUpto + ");
         // rewind frame ord=" + lastFrame.ord);
         // }
+
+        // We got lastFrame by comparing target and term, and target less than last seeked term in
+        // currentFrame. If lastFrame's fp is same with currentFrame's fp, we can reduce entCount to
+        // nextEnt.
```
mikemccand, line 767: Maybe change to `we can rewind without reloading`?  We don't seem to reduce `entCount` anymore?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
```diff
@@ -518,7 +541,20 @@ public boolean seekExact(BytesRef target) throws IOException {
...
+        // We still stay on withOutReload frame, reduce entCount to nextEnt.
+        int origEntCount = -1;
+        if (currentFrame.fp == withOutReloadFp && origNextEnt != 0) {
+          origEntCount = currentFrame.entCount;
+          currentFrame.entCount = origNextEnt;
+        }
+
         final SeekStatus result = currentFrame.scanToTerm(target, true);
+
+        // Revert entCount to origEntCount.
```
mikemccand, line 553: `Revert` -> `Restore`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
```diff
@@ -573,7 +609,20 @@ public boolean seekExact(BytesRef target) throws IOException {
...
+    // We still stay on withOutReload frame, reduce entCount to nextEnt.
+    int origEntCount = -1;
+    if (currentFrame.fp == withOutReloadFp && origNextEnt != 0) {
+      origEntCount = currentFrame.entCount;
+      currentFrame.entCount = origNextEnt;
+    }
+
     final SeekStatus result = currentFrame.scanToTerm(target, true);
+
+    // Revert entCount to origEntCount.
```
mikemccand, line 621: `Revert` -> `Restore`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
```diff
@@ -434,8 +436,29 @@ public boolean seekExact(BytesRef target) throws IOException {
...
+        // currentFrame. If lastFrame's fp is same with currentFrame's fp, we can reduce entCount to
+        // nextEnt.
+        boolean currentIsLast = currentFrame.fp == lastFrame.fp;
         currentFrame = lastFrame;
-        currentFrame.rewind();
+
+        // Only rewindWithoutReload for non-floor block or first floor block.
+        // TODO: We need currentFrame's first entry to judge whether we can rewindWithoutReload for
+        // non-first floor blocks.
+        if (currentFrame.fp != currentFrame.fpOrig
```
mikemccand, line 449: Could you add comment for each of these reasons to still reload the frame?

E.g.:

`currentFrame.fp != currentFrame.fpOrig // this is a floor multi-block and not the first one``
`currentFrame.entCount == 0 // why does this happen?`

etc.?

I don't understand when/why the 2nd two conditions occur.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
```diff
@@ -434,8 +436,29 @@ public boolean seekExact(BytesRef target) throws IOException {
...
+            || currentFrame.entCount == 0
+            || currentFrame.nextEnt == -1) {
+          currentFrame.rewind();
+        } else {
+          // Prepare to reduce entCount.
+          if (currentIsLast && currentFrame.isLeafBlock) {
```
mikemccand, line 455: I am confused why there are two cases under the `rewindWithoutReload` case.  One case you can roll back `entCount` temporarily, because you know the target term is before the current term?  And in the 2nd case, you don't know that, but you do know the target term is in this current block?  But then why are we even rewinding?

mikemccand (commented):
Thank you for being patient @vsop-479 -- I keep trying to review/understand this change and keep struggling :)

I left some superficial-ish comments.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
vsop-479: > One case you can roll back entCount temporarily, because you know the target term is before the current term? 

Yes.

> but you do know the target term is in this current block? But then why are we even rewinding?

We get `lastFrame` from `stack[0]` firstly, and compare `target` to last seek'd `term` to update `lastFrame` to reuse seek state, and assign it to `currentFrame`. This `currentFrame` only got by `target` and last seek'd `term` 's common prefix, we still need to continue walking the index to seek `target` term's block.

I think we need rewinding to set frame's state (`nextEnt`, `entCount`, etc.) to prepare for seeking. Unlike `rewind`, `rewindWithoutReload` can keep loaded block's data (same `FP`).

vsop-479: `rewindWithoutReload` is called when currentFrame's  is loaded and `fp` equals `fpOrig`.
Doing this can avoid reload a loaded block, when we finally need seek it, and it is still in `stack`.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
vsop-479: > currentFrame.entCount == 0 // why does this happen?`

We push frame to `stack`, and load it when we need to seek it. so we may get a unloaded frame from stack.

vsop-479: Maybe `currentFrame.nextEnt == -1` is enough for both unloaded or rewinded block, I will confirm that.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnumFrame.java
```diff
@@ -287,6 +287,68 @@ void rewind() {
...
+  // Current term greater than target, reduce endCount.
+  void rewindWithoutReload() {
+    // Set nextEnt to 0, to prevent force load.
+    nextEnt = 0;
+    suffixesReader.setPosition(0);
+    suffixLengthsReader.setPosition(0);
+    statsReader.setPosition(0);
+    bytesReader.setPosition(0);
+
+    // TODO: Since we only rewind without reload for fist floor(currentFrame.fp ==
```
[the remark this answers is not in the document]
vsop-479: Done.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
vsop-479: > We don't seem to reduce entCount anymore?

We can reduce entCount to nextEnt if we finally seek the same block.
This has been already implemented in: 
````

// We still stay on withOutReload frame, reduce entCount to nextEnt.
        int origEntCount = -1;
        if (currentFrame.fp == withOutReloadFp && origNextEnt != 0) {
          origEntCount = currentFrame.entCount;
          currentFrame.entCount = origNextEnt;
        }
````

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
vsop-479: Done.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
vsop-479: Done.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
vsop-479: > Could you add comment for each of these reasons to still reload the frame?

Added.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
vsop-479: > Let's leave it as a future TODO?

OK.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
mikemccand: Ahh I had forgotten that `entCount == 0` means unloaded block -- that makes sense.  Thanks.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
vsop-479: Maybe I was not clear. I mean we can `rewindWithoutReload` when `currentFrame.fp == currentFrame.fpOrig`, furthermore, if we finally seek the same block with last `term`, we can roll back `entCount` temporarily.

vsop-479:
Conflicts resolved.

vsop-479:
Hello @mikemccand -- I am trying to simplify this change, it seems too complex. How about remove the reduce and restore entCount part?

vsop-479:
Resolve conflicts.

PR #13255, opened 2024-04-02
Remove Accountable interface in KnnVectorsReader

### LINKED ISSUE #13241: Remove Accountable interface on KnnVectorsReader
### Description

In Lucene 9.0, we removed the `Accountable` interface on our index readers and file formats, because heap usage had become so low that this metric had become less relevant, and also much less accurate as there were no longer a couple things that would account for a majority of the RAM usage, but rather many small objects that don't account for much memory on their own and that were too hard to estimate properly.

We still kept `Accountable` on `KnnVectorsReader` because it used to load lots of things in heap at the time. Since vectors have become much more reasonable since then, I would suggest now removing `Accountable` on `KnnVectorsReader` like we did for other file formats in 9.0.

Pulkitg64:
### Description

Closes #13241 

Remove Accountable interface in ```KnnVectorsReader``` and removed ramBytesUsed function from wherever KNNVectorsReader class is used/extended.


jpountz:
Thanks for looking into this. Can you add an entry to lucene/MIGRATE.txt and lucene/CHANGES.txt under 10.0?

Let's also wait for the discussion to settle on the linked issue as it looks like this change may be controversial.

jpountz:
Separately, I wonder if removing `Accountable` on `KnnVectorsReader` would help remove the `Accountable` on some other utility classes. For instance, it looks like `DirectMonotonicReader.Meta` may be such a class, as the fact that it's accountable is only taken advantage of by `KnnVectorsReader` implementations. (For a separate change.)

jpountz:
I opened #13280 as a follow-up.

PR #13258, opened 2024-04-02
Adds bwc indices for 9.10.1

benwtrent:
Adds bwc indices for 9.10.1.

All I did was run: 

```
./gradlew :lucene:backward-codecs:test -Ptests.useSecurityManager=false --tests TestGenerateBwcIndices
```


benwtrent:
@mikemccand ^ This should fix that build failure. I am guessing for every line in `version.txt` if the appropriate BWC index isn't found, its assumed to be unsupported, and so it looks for the unsupported index. 

I forgot to generate these when I adding the missing 9.10.1. Sorry for the noise

jpountz (approved):

mikemccand:
Super, thank you @benwtrent !

PR #13259, opened 2024-04-03
Rename NodeHash to FSTSuffixNodeCache + Javadoc improvement

dungba88:
### Description

Rename NodeHash to FSTSuffixNodeCache as NodeHash is rather an obscured name which doesn't tell its exact functionality. I also added the class Javadoc to explain what it is for and how it works, so that it would be easier to follow-up its implementation details (which I often struggled when working on it the first time).


dungba88:
Hi Lucene community, would someone kindly take a look at this PR? This is only minor renaming and Javadoc improvement.

msokolov:
Thanks, this does seem much more explicit and helpful

PR #13260, opened 2024-04-03
BaseVectorSimilarityQueryTestCase assumes connected hnsw graph

msokolov:
…rityQueryTestCase

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/BaseVectorSimilarityQueryTestCase.java
+++ b/lucene/core/src/test/org/apache/lucene/search/BaseVectorSimilarityQueryTestCase.java
@@ -26,17 +26,23 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.IntStream;
+import org.apache.lucene.codecs.HnswGraphProvider;
+import org.apache.lucene.codecs.perfield.PerFieldKnnVectorsFormat;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field;
 import org.apache.lucene.document.IntField;
+import org.apache.lucene.index.CodecReader;
 import org.apache.lucene.index.DirectoryReader;
 import org.apache.lucene.index.IndexReader;
 import org.apache.lucene.index.IndexWriter;
+import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.index.VectorSimilarityFunction;
 import org.apache.lucene.store.Directory;
 import org.apache.lucene.tests.index.RandomIndexWriter;
 import org.apache.lucene.tests.util.LuceneTestCase;
+import org.apache.lucene.tests.util.hnsw.HnswTestUtil;
+import org.apache.lucene.util.hnsw.HnswGraph;
 
 @LuceneTestCase.SuppressCodecs("SimpleText")
 abstract class BaseVectorSimilarityQueryTestCase<
@@ -165,6 +171,9 @@ abstract class BaseVectorSimilarityQueryTestCase<
 
     try (Directory indexStore = getIndexStore(getRandomVectors(numDocs, dim));
         IndexReader reader = DirectoryReader.open(indexStore)) {
+      if (graphIsDisconnected(reader)) {
+        return;
+      }
       IndexSearcher searcher = newSearcher(reader);
 
       Query query =
@@ -289,6 +298,9 @@ abstract class BaseVectorSimilarityQueryTestCase<
       w.commit();
 
       try (IndexReader reader = DirectoryReader.open(indexStore)) {
+        if (graphIsDisconnected(reader)) {
+          return;
+        }
         IndexSearcher searcher = newSearcher(reader);
 
         Query query =
@@ -521,5 +533,21 @@ abstract class BaseVectorSimilarityQueryTestCase<
       }
     }
     return dir;
+  }
+
+  private boolean graphIsDisconnected(IndexReader reader) throws IOException {
+    for (LeafReaderContext ctx : reader.leaves()) {
+      HnswGraph graph =
+          ((HnswGraphProvider)
+                  ((PerFieldKnnVectorsFormat.FieldsReader)
+                          ((CodecReader) ctx.reader()).getVectorReader())
+                      .getFieldReader(vectorField))
+              .getGraph(vectorField);
+      if (HnswTestUtil.isFullyConnected(graph) == false) {
+        // for now just bail out.
+        return true;
+      }
+    }
+    return false;
   }
 }
--- a/lucene/core/src/test/org/apache/lucene/search/BaseVectorSimilarityQueryTestCase.java
+++ b/lucene/core/src/test/org/apache/lucene/search/BaseVectorSimilarityQueryTestCase.java
@@ -26,23 +26,18 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.IntStream;
-import org.apache.lucene.codecs.HnswGraphProvider;
-import org.apache.lucene.codecs.perfield.PerFieldKnnVectorsFormat;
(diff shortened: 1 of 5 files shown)


uschindler (changes requested):

lucene/core/src/test/org/apache/lucene/search/BaseVectorSimilarityQueryTestCase.java
```diff
@@ -522,4 +534,20 @@ final Directory getIndexStore(V... vectors) throws IOException {
     }
     return dir;
   }
+
+  private boolean graphIsDisconnected(IndexReader reader) throws IOException {
```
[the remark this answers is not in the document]
msokolov: yeah, maybe we want to use in some other test class too. Ill refactor and pull out the field name

lucene/test-framework/src/java/org/apache/lucene/tests/util/hnsw/HnswTestUtil.java
```diff
@@ -0,0 +1,132 @@
...
+    for (LeafReaderContext ctx : reader.leaves()) {
+      HnswGraph graph =
+          ((HnswGraphProvider)
+                  ((PerFieldKnnVectorsFormat.FieldsReader)
+                          ((CodecReader) ctx.reader()).getVectorReader())
+                      .getFieldReader(vectorField))
+              .getGraph(vectorField);
+      if (HnswTestUtil.isFullyConnected(graph) == false) {
+        return false;
+      }
```
benwtrent, line 128: Do we need to unwrap `FilterLeafReader`? I wonder if tests would randomly wrap the readers and this would fail the casting here. Since its a test, not that big of a deal as we would see the failure as a class cast exception.

I think assuming that all KnnVectorReaders implement `HnswGraphProvider` is OK for this utility as its specifically an HNSW utility.

uschindler: We have Unwrappable#unwrap for that (under the assumption that FilterLeafReader implements that interface).

I added this to some classes, e.g
 It works with directories.

lucene/test-framework/src/java/org/apache/lucene/tests/util/hnsw/HnswTestUtil.java
```diff
@@ -0,0 +1,132 @@
...
+        int next = (i << 6) + Long.numberOfTrailingZeros(word);
+        if (next >= bits.length()) {
+          return NO_MORE_DOCS;
+        } else {
+          return next;
+        }
+      }
+    }
+    return NO_MORE_DOCS;
+  }
```
shubhamvishu, line 115: This function seems to be motivated from `FixedBitSet#nextSetBit` and would be more suitable for `FixedBitSet` class?....I wonder why it does not have it already.

lucene/test-framework/src/java/org/apache/lucene/tests/util/hnsw/HnswTestUtil.java
```diff
@@ -0,0 +1,132 @@
...
+   */
+  public static boolean isFullyConnected(HnswGraph knnValues) throws IOException {
+    return componentSizes(knnValues).size() < 2;
+  }
+
+  /**
+   * Returns the sizes of the distinct graph components on level 0. If the graph is fully-connected
+   * there will only be a single component. If the graph is empty, the returned list will be empty.
+   */
+  public static List<Integer> componentSizes(HnswGraph hnsw) throws IOException {
```
shubhamvishu, line 49: It seems like the purpose of this function is to be a utility for future use cases(if any) because if we only need to know the graph is fully connected we could just pick a node and traverse only once to check if its size is equal to graph size i.e. early terminate instead of going through all components. Alternatively, I think we should keep `componentSizes` as-is but implement `isFullyConnected` like below?

```java
public static boolean isFullyConnected(HnswGraph hnsw) throws IOException {
    FixedBitSet connectedNodes = new FixedBitSet(hnsw.size());
    assert hnsw.size() == hnsw.getNodesOnLevel(0).size();
    System.out.println("size=" + hnsw.size());
    return connectedNodes.length() == traverseConnectedNodes(hnsw, connectedNodes);
  }
  ```

**Review by shubhamvishu — COMMENTED** [review]
Thanks @msokolov for fixing this!.....I left some comments

File: lucene/test-framework/src/java/org/apache/lucene/tests/util/hnsw/HnswTestUtil.java (hunk 3 above)
Comment (msokolov) [reply lucene/test-framework/src/java/org/apache/lucene/tests/util/hnsw/HnswTestUtil.java:115]: Yes it could go there - it's a near clone of `FixedBitSet.nextSetBit` but I guess the motivation is to limit the surface area of supported functions. Perhaps if this gets wider use we could move it?

File: lucene/test-framework/src/java/org/apache/lucene/tests/util/hnsw/HnswTestUtil.java (hunk 2 above)
Comment (msokolov) [reply lucene/test-framework/src/java/org/apache/lucene/tests/util/hnsw/HnswTestUtil.java:128]: LeafReader doesn't seem to implement that unwrapping interface right now. Are you saying `ctx.reader()` might not be a `CodecReader`?

Comment (benwtrent) [reply lucene/test-framework/src/java/org/apache/lucene/tests/util/hnsw/HnswTestUtil.java:128]: The `ctx.reader()` might be a `FilterLeafReader` which isn't a `CodecReader`. I don't know if any of the tests actually wrap it or not.

So, I would do `FilterLeafReader.unwrap(ctx.reader())`.

`FilterLeafReader` doesn't implement `Unwrappable`.

File: lucene/test-framework/src/java/org/apache/lucene/tests/util/hnsw/HnswTestUtil.java (hunk 3 above)
Comment (msokolov) [reply lucene/test-framework/src/java/org/apache/lucene/tests/util/hnsw/HnswTestUtil.java]: I guess one would also want to discourage its use in performance-sensitive code because it requires extra work beyond what `nextSetBit` does

File: lucene/test-framework/src/java/org/apache/lucene/tests/util/hnsw/HnswTestUtil.java (hunk 2 above)
Comment (uschindler) [reply lucene/test-framework/src/java/org/apache/lucene/tests/util/hnsw/HnswTestUtil.java]: Ok. So that's a job for future to implement this interface for all filter or wrappers. Originally this was added for MMapDirectory and some other class.

Comment (msokolov) [reply lucene/test-framework/src/java/org/apache/lucene/tests/util/hnsw/HnswTestUtil.java]: Oh, I see. You can call `Unwrappable.unwrapAll()` on anything it's just a no-op, duh. I suppose I can add that `implements Unwrappable`.

Comment (uschindler) [reply lucene/test-framework/src/java/org/apache/lucene/tests/util/hnsw/HnswTestUtil.java]: Actually it's not implemented yet for any subclass: https://github.com/search?q=repo%3Aapache%2Flucene%20unwrappable&type=code

Comment (uschindler) [reply lucene/test-framework/src/java/org/apache/lucene/tests/util/hnsw/HnswTestUtil.java]: Can you just add the interface to the FilterLeafReader? After that it confirms to the standard and You can use the static unwrapper.

**Review by uschindler — APPROVED** [review]

**Comment by uschindler** [comment]
I think the Unwrappable for FilterReader should be added in a separate PR. For now this is fine.

I remember vaguely that I left our FilterReader because of the static method conflicting with the interface. This may be a good job for cleanup 10.x (remove static and just implement interface).

**Comment by msokolov** [comment]
thanks for the reviews!

**Comment by msokolov** [comment]
hm a failure popped up when I pushed:

```
org.apache.lucene.search.TestKnnByteVectorQuery > testTimeout FAILED
    java.lang.AssertionError: expected:<1> but was:<0>
        at __randomizedtesting.SeedInfo.seed([AE2970724EFAD848:FE11FBC92B12FA54]:0)
 ```

I don't *think* it's related but I'll check

benwtrent:
@msokolov I have had something similar fail in other code, so it isn't related to your change here. I think its a flaky test. It doesn't reliably reproduce for me.

Issue #13207, opened 2024-03-24
Find more classes in main branch that can be converted to record classes


uschindler:
### Description

This is an overview issue about all classes in current Lucene's main branch to be ported over to `record` classes. Record classes are fine for immutable data structures which have no functionality and are not part of class hierarchy (they have no superclass).

Records offer several simplifications for implementing them:
- equals, hashCode and toString are autogenerated by an invokedynamic, so we can't forget to handle it correctly
- it is also immutable
- all parameter checking logic can be moved to some special constructor without any parameters (this is new syntax since Java 17)
- In later java versions they will have an optimized internal layout and arrays of record classes may no longer be pointer arrays (see Project Valhalla with a first preview of it in Java 22). In future they will also not have identity anymore (I don't understand the impact yet).

The first PR #13205 already added support for record classes in our Javadoc linter (`missing-doclet` plugin).

The following classes are candidates (please add more here and add link to PR):
- [x] `IOContext`, `MergeInfo`, `FlushInfo`: #13205
- [x] `BooleanClause` #13261
- [ ] `FieldInfo` ?
- [x] All the various`FieldEntry` classes used by codecs internally: #13296
- [ ] All the various `FieldMetaData`/`FSTMetadata`/`DirectMonotonicReader.Meta` classes used by codecs internally.
- [ ] More classes: #13328

jpountz:
Lucene 10 is out, I'm closing.

PR #13261, opened 2024-04-04
Convert BooleanClause class to record class

Pulkitg64:
### Description

Convert ```BooleanClause``` class to record class which addresses #13207 partially.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/BooleanClause.java
+++ b/lucene/core/src/java/org/apache/lucene/search/BooleanClause.java
@@ -68,9 +68,9 @@ public record BooleanClause(Query query, Occur occur) {
   }
 
   /** Constructs a BooleanClause. */
-  public BooleanClause(Query query, Occur occur) {
-    this.query = Objects.requireNonNull(query, "Query must not be null");
-    this.occur = Objects.requireNonNull(occur, "Occur must not be null");
+  public BooleanClause {
+    Objects.requireNonNull(query, "Query must not be null");
+    Objects.requireNonNull(occur, "Occur must not be null");
   }
 
   public boolean isProhibited() {
--- a/lucene/core/src/java/org/apache/lucene/search/BooleanClause.java
+++ b/lucene/core/src/java/org/apache/lucene/search/BooleanClause.java
@@ -19,7 +19,7 @@ package org.apache.lucene.search;
 import java.util.Objects;
 
 /** A clause in a BooleanQuery. */
-public final class BooleanClause {
+public record BooleanClause(Query query, Occur occur) {
 
   /** Specifies how clauses are to occur in matching documents. */
   public enum Occur {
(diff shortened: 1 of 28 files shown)


jpountz (approved):
Please add a note to `lucene/MIGRATE.txt`, otherwise it looks good to me.

Do you have thoughts @uschindler ?

jpountz:
> Please add a note to lucene/MIGRATE.txt

By the way, I wonder how we should handle all the classes that we are converting to records, should we have a single section in `lucene/MIGRATE.txt` that lists all classes for simplicity, instead of having one entry per class?

Pulkitg64:
That's a good idea @jpountz. I can add this change under this section: https://github.com/apache/lucene/blob/main/lucene/MIGRATE.md#some-classes-converted-to-records-classes-github13207

uschindler:
> That's a good idea @jpountz. I can add this change under this section: https://github.com/apache/lucene/blob/main/lucene/MIGRATE.md#some-classes-converted-to-records-classes-github13207

That was also my plan. This is why I created the section in the migration guide. No need for separate sections, but a new entry in the enumeration is required.

lucene/core/src/java/org/apache/lucene/search/BooleanClause.java
```diff
@@ -67,25 +67,12 @@ public String toString() {
     };
   }
 
-  /** The query whose matching documents are combined by the boolean query. */
-  private final Query query;
-
-  private final Occur occur;
-
   /** Constructs a BooleanClause. */
   public BooleanClause(Query query, Occur occur) {
```
uschindler, line 71: Please replace this by the generic constructor without parameters and brackets.

`public BooleanClause {`

Just check parameters, no assignment.

uschindler (changes requested):

lucene/core/src/java/org/apache/lucene/search/BooleanClause.java
uschindler: Like this: https://github.com/apache/lucene/blob/54a2e11434d81ab92923bccbe651752f81ac1973/lucene/core/src/java/org/apache/lucene/store/IOContext.java#L61

Pulkitg64: Makes sense. On it.

Pulkitg64: Changed in the new commit.

uschindler (approved):

uschindler:
The failing test is a known issue (see mailing list). Just ignore that.

wardle:
Hi all. I understand the change to a record class but this does create a breaking change that would be avoided through the addition of deprecated accessors matching the old class ie getQuery and getOccur, which could be later removed in Lucene 11. Obviously other changes have been easier to manage in moving from Lucene 9->10 because methods have been flagged as deprecated in 9 and it has been straightforward to prepare for these changes. It is tricky because for users of Lucene with codebases that might run on either Java 11, or 21, one previously could have a codebase that would run using different Lucene versions, but this change means one has to choose the correct Lucene version AND keep two branches of their code.

uschindler:
Hi,
This was planned to be backwards incompatible. There is a migration guide entry. 

Lucene generally revises APIs between major versions in incompatible ways. With Lucene 9 this was even more dramatic as a large part of the class and package structure was renamed to be Java module system compatible.

Lucene 10 was released last week and there won't be any changes anymore.

Uwe

PR #13263, opened 2024-04-04
Correct quantized HNSW validation to be in-line with HNSW

benwtrent:
When intra-merge parallelism was introduced, the validation that `numWorkers` must `==1` with a `null` executor service was removed from `Lucene99HnswVectorsFormat`. However, I forgot to remove that validation from `Lucene99HnswScalarQuantizedVectorsFormat`. 

This corrects that mistake, allowing `Lucene99HnswScalarQuantizedVectorsFormat` and `Lucene99HnswVectorsFormat` to take advantage of the merge-schedulers intra-merge threading.


tteofili (approved):
LGTM

PR #13244, opened 2024-03-29
Use ReadAdvice.RANDOM by default.

jpountz:
This switches the default `ReadAdvice` from `NORMAL` to `RANDOM`, which is a better fit for the kind of access pattern that Lucene has. This is expected to reduce page cache trashing and contention on the page table.

`NORMAL` is still available, but never used by any of the file formats.

PR #13264, opened 2024-04-04
Make the default ReadAdvice configurable by sysprop

uschindler:
Followup after #13244: This makes the default `ReadAdvice` for index files configurable via system property `org.apache.lucene.store.defaultReadAdvice`.

I moved the corresponding constant to the `oal.util.Constants` class because this makes reading and error reporting easier as @rmuir and I added some framework to safely support security manager and map those variables (we used this for vectorization).

It would be good to move all reading of system properties to that class so we do not have the splattered AccessControllers everywhere.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -163,9 +163,11 @@ Bug Fixes
 Changes in Runtime Behavior
 ---------------------
 
-* GITHUB#13244: IOContext now uses ReadAdvice#RANDOM by default for read
+* GITHUB#13244, GITHUB#13264: IOContext now uses ReadAdvice#RANDOM by default for read
   operations. An implication is that `MMapDirectory` will use POSIX_MADV_RANDOM
-  on POSIX systems. (Adrien Grand)
+  on POSIX systems. To fallback to OS default behaviour, pass system property via
+  `-Dorg.apache.lucene.store.defaultReadAdvice=normal`. This may be useful on systems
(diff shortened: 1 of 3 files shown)


uschindler:
P.S.: I moved the lookup cache to a new constant name and placed it next to the `withReadAdvice()` method.

jpountz (approved):
This makes sense to me. At first I wondered if we needed this since it's possible to create a `FilterDirectory` that calls `ioContext.withReadAdvice(ReadAdvice.NORMAL)` on `Directory#openIndexInput` or something along these lines. But actually this wouldn't work as well, as this would override explicit decisions made from the code, which is probably not what we want.

uschindler:
> This makes sense to me. At first I wondered if we needed this since it's possible to create a `FilterDirectory` that calls `ioContext.withReadAdvice(ReadAdvice.NORMAL)` on `Directory#openIndexInput` or something along these lines. But actually this wouldn't work as well, as this would override explicit decisions made from the code, which is probably not what we want.

That was exactly my problem with making the default hardcoded.

I thought about making it configurable "per directory", but this is too much changes for little benefit. As this is mostly machine specific, a system property that can be passed at command line was my favourite.

In addition, we should still pass `iocontext.withReadAdvice(ReadAdvice.RANDOM)` in our index files, where we know for sure that the access will be fully random (like HNSW), so the other PR you implemented was still fine.

So the default specified by system property is only used for cases where no explicit read advice is given at all .

PR #13265, opened 2024-04-04
Propagate the flush IOContext to stored fields / term vectors writers when index sorting is enabled.

jpountz:
This fixes index sorting to pass the correct `IOContext` to stored fields and term vectors writers when index sorting is enabled. This is important for things like `NRTCachingDirectory`.


jpountz:
I tried to create a test for it but couldn't create a satisfying one.

uschindler (approved):

PR #13267, opened 2024-04-04
Use IOContext#RANDOM when appropriate.

jpountz:
This switches the following files to `IOContext#RANDOM`:
 - Stored fields data file.
 - Term vectors data file.
 - HNSW graph.
 - Temporary file storing vectors at merge time that we use to construct the merged HNSW graph.
 - Vector data files, including quantized data files.

I hesitated using `IOContext.RANDOM` on terms, since they have a random access pattern when running term queries, but a more sequential access pattern when running multi-term queries. I erred on the conservative side and did not switch them to `IOContext.RANDOM` for now.

For simplicity, I'm only touching the current codec, not previous codecs.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -78,6 +78,12 @@ Optimizations
 * GITHUB#13149: Made PointRangeQuery faster, for some segment sizes, by reducing the amount of virtual calls to
   IntersectVisitor::visit(int). (Anton Hägerstrand)
 
+* GITHUB#13222: Vector data files - including quantized vectors, HNSW graphs,
+  stored fields data files and term vectors data files now advise the
+  underlying `Directory` to optimize for random-access. On recent Java versions
+  (>= 21) and POSIX systems, `MMapDirectory` would use MADV_RANDOM for these
+  files under the hood. (Adrien Grand)
+
 Bug Fixes
 ---------------------
 
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -19,9 +19,12 @@ New Features
 * GITHUB#12915: Add new token filters for Japanese sutegana (捨て仮名). This introduces JapaneseHiraganaUppercaseFilter
   and JapaneseKatakanaUppercaseFilter. (Dai Sugimori)
 
-* GITHUB#13196: Add support for posix_madvise to MMapDirectory: If running on Linux/macOS and Java 21
-  or later, MMapDirectory uses IOContext to pass suitable MADV flags to kernel of operating system.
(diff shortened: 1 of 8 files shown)


lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatVectorsWriter.java
```diff
@@ -315,10 +316,13 @@ public CloseableRandomVectorScorerSupplier mergeOneFieldToIndex(
...
       IOUtils.close(tempVectorData);
 
-      // copy the temporary file vectors to the actual data file
+      // This temp file will be accessed in a random-access fashion to construct the HNSW graph.
+      // Note: don't use the context from the state, which is a flush/merge context, not expecting
+      // to perform random reads.
       vectorDataInput =
           segmentWriteState.directory.openInput(
-              tempVectorData.getName(), segmentWriteState.context);
+              tempVectorData.getName(), IOContext.DEFAULT.withRandomAccess());
```
uschindler, line 324: Here you could also use `IOContext.RANDOM` directly.

lucene/core/src/java/org/apache/lucene/store/IOContext.java
```diff
@@ -132,10 +132,24 @@ public IOContext(IOContext ctxt, boolean readOnce) {
     this.mergeInfo = ctxt.mergeInfo;
     this.flushInfo = ctxt.flushInfo;
     this.readOnce = readOnce;
-    this.randomAccess = ctxt.randomAccess;
+    this.randomAccess = false;
```
uschindler, line 135: I think this should be documented.

uschindler (approved):
Looks fine.

I think changing the default to random (using a sysprop) can be done optionally once we have enough benchmarks.

lucene/CHANGES.txt
```diff
@@ -78,6 +78,12 @@ Optimizations
 * GITHUB#13149: Made PointRangeQuery faster, for some segment sizes, by reducing the amount of virtual calls to
   IntersectVisitor::visit(int). (Anton Hägerstrand)
 
+* GITHUB#13222: Vector data files - including quantized vectors, HNSW graphs,
+  stored fields data files and term vectors data files now advise the
+  underlying `Directory` to optimize for random-access. On recent Java versions
+  (>= 21) and POSIX systems, `MMapDirectory` would use MADV_RANDOM for these
```
uschindler, line 84: It is a bit unclear: both conditions must apply. "On Linux and macOS systems using recent Java versions (>=21), MMapDirectory would use POSIX_MADV_RANDOM"

uschindler (approved):

PR #13268, opened 2024-04-04
UnifiedHighlighter highlight on multiple fields

mayya-sharipova:
Add ability to UnifiedHighlighter to combine matches from multiple fields
 to highlight a single field.

  FastVectorHighlighter for a long time has an option to highlight a single field
  based on matches from several fields. But UnifiedHighlighter was missing this option.
  This adds this ability.


lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/UnifiedHighlighter.java
```diff
@@ -712,17 +712,21 @@ public String[] highlight(String field, Query query, TopDocs topDocs) throws IOE
...
    * @return Array of formatted snippets corresponding to the documents in <code>topDocs</code>. If
    *     no highlights were found for a document, the first {@code maxPassages} sentences from the
    *     field will be returned.
    * @throws IOException if an I/O error occurred during processing
    * @throws IllegalArgumentException if <code>field</code> was indexed without {@link
    *     IndexOptions#DOCS_AND_FREQS_AND_POSITIONS_AND_OFFSETS}
    */
-  public String[] highlight(String field, Query query, TopDocs topDocs, int maxPassages)
+  public String[] highlight(
+      String field, Query query, TopDocs topDocs, int maxPassages, Set<String> matchedFields)
```
jimczi, line 724: Can we use the `Builder` instead of overloading? The field matcher for instance is provided through the builder and custom implementation can override: `protected Predicate<String> getFieldMatcher(String field)`.
So for the new functionality that would mean adding something like `protected Set<String> getMaskedFields(String field)` (naming is hard) and to make it available in the `Builder`.

jimczi (commented):
This is a great addition @mayya-sharipova .
I left a comment regarding the exposure, we should also document the limit of the feature (input should be identical for all masked fields).

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/UnifiedHighlighter.java
mayya-sharipova: @jimczi Thanks, this is a great comment, addressed with recent commit. The amount of changes is much smaller now.

I opted for naming of `matchedFields` similar to `FastVectorHighlighter`, but willing to change it to `maskedFields` it is a better option.

mayya-sharipova:
@jimczi Thanks for your initial review. I've tried to address all your comments, so this is ready for another review.

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/UnifiedHighlighter.java
jimczi: Thanks @mayya-sharipova . I have a slight preference for `maskedFields` since we already have a `fieldMatcher` which exposes a different feature.

mayya-sharipova: @jimczi Makes sense, addressed in 1f7679f540a37324348ce47a245942f636ea92aa

lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/UnifiedHighlighter.java
```diff
@@ -360,6 +366,22 @@ public Builder withFieldMatcher(Predicate<String> value) {
...
+     * Set up a function that given a field retuns a set of masked fields whose matches are combined
+     * to highlight the given field. This is useful when you want to highlight a field based on
+     * matches from several fields.
+     *
+     * <p>Note: All masked fields must share the same source as the field being highlighted,
+     * otherwise their offsets will not correspond to the highlighted field.
+     *
+     * <p>Note: Only the field being highlighted must provide an original source value (e.g. through
+     * stored field), other masked fields don't need it.
+     */
```
jimczi, line 379: It's unclear whether the original field must be added in the list or not? Currently the highlighted fields must be present in the returned set to be considered for highlighting. Should we instead consider these masked fields as extra fields (in addition to the original highlighted fields)?

mayya-sharipova: Great suggestion, addressed in https://github.com/apache/lucene/pull/13268/commits/4dec32a339e95106ee9bc7be668a110ef296bb75

jimczi (approved):
LGTM

PR #13269, opened 2024-04-04
upgrade commons codec from 1.13.0 to 1.16.0

rmuir:
Currently, we depend upon a pre-COVID version of commons-codec (like several other dependencies, I am working the issues).

Upgrade to 1.16.0.

NOTE: I'm unable to upgrade to 1.16.1: it causes test failures. 

```
   $ ./gradlew -p lucene/analysis/phonetic test
   ...
   >     java.lang.ExceptionInInitializerError
   >         at org.apache.commons.codec@1.16.1/org.apache.commons.codec.language.bm.Lang.<clinit>(Lang.java:108)
   >         at org.apache.commons.codec@1.16.1/org.apache.commons.codec.language.bm.PhoneticEngine.<init>(PhoneticEngine.java:300)
   >         at org.apache.commons.codec@1.16.1/org.apache.commons.codec.language.bm.PhoneticEngine.<init>(PhoneticEngine.java:276)
             ...
   >
   >         Caused by:
   >         java.lang.IllegalArgumentException: Unable to resolve required resource: org/apache/commons/codec/language/bm/ash_languages.txt
   >             at org.apache.commons.codec@1.16.1/org.apache.commons.codec.Resources.getInputStream(Resources.java:38)
   >             at org.apache.commons.codec@1.16.1/org.apache.commons.codec.language.bm.Languages.getInstance(Languages.java:254)
   >             at org.apache.commons.codec@1.16.1/org.apache.commons.codec.language.bm.Languages.<clinit>(Languages.java:243)
```

Seems some resources are missing from the jar or similar.
I tried to investigate a bit, by looking at their source code changes from 1.16.0 to 1.16.1, it is overwhelming, not exactly what I'd call a "patch" release :)

https://github.com/apache/commons-codec/compare/rel/commons-codec-1.16.0...rel/commons-codec-1.16.1

Either way, that isn't to be fixed here, so let's at least move from a 2019 version to a 2023 version. We can separately debug the 1.16.1 issue / report it to them / file a PR /etc.


rmuir:
I think i understand the 1.16.1 issue. Looks like there is a bit of jar/maven-hell going on among the commons dependencies. I hit it when I tried to upgrade commons-compress...

Bumping both of them to the latest version is not the solution, it seems to then just cause "module-hell" and still fails the build. But for now we can get them on less ancient versions.

rmuir:
I will open a separate issue for the commons-compress as there is more trouble there...

msfroh:
I figured out what broke in commons-codec 1.16.1 and reported an issue over there: https://issues.apache.org/jira/projects/CODEC/issues/CODEC-324

garydgregory:
Note that the current version of Commons Codec is 1.17.1.

garydgregory:
Note that I pushed a new version of Commons Codec a couple of days ago.

rmuir:
outdated PR: we're now on commons-codec 1.18.0 with update PRs assisted with dependabot

PR #13270, opened 2024-04-04
update commons-compress from 1.19 to 1.21

rmuir:
Another pre-covid dependency version...

I tried to upgrade to the latest version (both with and without the entangled commons-codec upgrade: https://github.com/apache/lucene/pull/13269) and there is trouble everywhere.

Newer versions of this library just drag in more and more dependencies and have more jar/module hell and larger surface area. For example they are dragging in JNI compressors etc!

https://commons.apache.org/proper/commons-compress/dependencies.html

For now, I'd like to just upgrade these two minor versions to get us "less ancient", but going forwards I think we need to evaluate if we should really have a dependency mess just to support `bzip2` format in `lucene/benchmark`.

Alternatively, benchmark could use the gzip support built into the JDK, and for bzip2 files, it could just support running `bzip2 -dc` command and reading from its stdout? It would avoid the entire mess.


rmuir:
1.22+ starts dragging in more stuff such as `commons-lang3` and `commons-io` which is why I stopped at 1.21

dweiss:
If it's used by the benchmark module then I don't think it's that problematic? Similar to test framework dependencies? Your suggestion to use a pipe works too but will hurt people living in darker times (on Windows).

rmuir:
@dweiss I think the problem is more keeping up, eg I'm unable to get things working with their latest versions.

Weve got dependencies on jars like nekohtml unmaintained for over a decade too. And build hacks for them...

rmuir:
anyway, if there's no objection, i'd like to bump this to 1.21 which is an easy win and doesnt modify the dependency graph.

dweiss (approved):

rmuir:
outdated PR: we're now on commons-compress 1.27.1 with updates assisted by dependabot

PR #13276, opened 2024-04-05
Add 'passageSortComparator' option in FieldHighlighter

Seunghan-Jung:
### Description

`FieldHighlighter` always sorts the final selected passages based on startOffset, but this may not align with the user's intentions.

For example, in the case of Solr's multiValue fields, the order might just be a simple listing of items without any significance. In such cases, the order in which passages appear is not important. It might be more effective to maintain sorting by score in these situations.

Therefore, it seems necessary to allow for the final sorting criteria to be determined in accordance with the user's intentions.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -298,6 +298,8 @@ Improvements
 * GITHUB#13385: Add Intervals.noIntervals() method to produce an empty IntervalsSource.
   (Aniketh Jain, Uwe Schindler, Alan Woodward))
 
+* GITHUB#13276: UnifiedHighlighter: new 'passageSortComparator' option to allow sorting other than offset order. (Seunghan Jung)
+
 Optimizations
 ---------------------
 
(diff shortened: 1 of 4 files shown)


dsmiley:
I like this change; very nice!  I shall merge it.  Please share how I shall list you in CHANGES.txt.

Proposed text under improvements to 9.11:
> GITHUB#13276: UnifiedHighlighter: new 'passageSortComparator' option to not do offset order (YOUR NAME)

lucene/CHANGES.txt
```diff
@@ -298,6 +298,8 @@ Improvements
 * GITHUB#13385: Add Intervals.noIntervals() method to produce an empty IntervalsSource.
   (Aniketh Jain, Uwe Schindler, Alan Woodward))
 
+* GITHUB#13276: UnifiedHighlighter: new 'passageSortComparator' option to allow sorting other than offset order. (Seunghan Jung)
```
Seunghan-Jung, line 301: Thank you, @dsmiley 
I have added it to the changes.txt myself. Could you please check it?
During this process, a rebase on main was necessary, so a force-push was performed.

dsmiley: Looks good.
Please don't force-push to a PR.  Instead, simply merge from main (and resolve conflicts).

dsmiley (approved):

lucene/CHANGES.txt
Seunghan-Jung: Oh, I will keep that in mind.

Issue #7049, opened 2014-10-04
Make indexwriter a mere mortal when exceptions strike [LUCENE-5987]


asfimport:
IndexWriter's exception handling is overly complicated. Every method in general reads like this:

```Java
try {
  try {
    try { 
     ...
     // lock order: COMPLICATED
     synchronized(this or that) {
     }
     ...
   } finally {
     if (!success5) {
       deleter.deleteThisFileOrThat();
     }
    ...
  }
}
```

Part of the problem is it acts like its an invincible superhero, e.g. can take a disk full on merge or flush to the face and just keep on trucking, and you can somehow fix the root cause and then just go about making commits on the same instance.

But we have a hard enough time ensuring exceptions dont do the wrong thing (e.g. cause corruption), and I don't think we really test this crazy behavior anywhere: e.g. making commits AFTER hitting disk full and so on.

It would probably be simpler if when such things happen, IW just considered them "tragic" just like OOM and rolled itself back, instead of doing all kinds of really scary stuff to try to "keep itself healthy" (like the little dance it plays with IFD in mergeMiddle manually deleting CFS files).

Besides, without something like a WAL, Indexwriter isn't really fit to be a superhero anyway: it can't prevent you from losing data in such situations. It just doesn't have the right tools for the job.

edit: just to be clear I am referring to abort (low level exception during flush) and exceptions during merge. For simple non-aborting cases like analyzer errors, of course we can deal with this. We already made great progress on turning a lot of BS exceptions that would cause aborts into non-aborting ones recently.

---
Migrated from [LUCENE-5987](https://issues.apache.org/jira/browse/LUCENE-5987) by Robert Muir (@rmuir), updated May 09 2016
Attachments: [LUCENE-5987.patch](https://apache.github.io/lucene-jira-archive/attachments/LUCENE-5987/LUCENE-5987.patch) (versions: 3)

PR #13277, opened 2024-04-06
IndexWriter: Treat java.lang.Error as tragedy

### LINKED ISSUE #13275: Strange Segment State after encoutering JDK22 bug
### Description

There is a nasty bug in JDK 22:

Elasticsearch issue: https://github.com/elastic/elasticsearch/issues/106987
JDK issue: https://bugs.openjdk.org/browse/JDK-8329528

This JDK bug is not present in JDK 21.0.x: https://github.com/elastic/elasticsearch/issues/106987#issuecomment-2039591571 ( it appears to be a regression starting with JDK 22 )

We noticed in places where this bug reared its ugly head, segment state ended up in a strange place. 

```
Unexpected file read error while reading index. (resource=BufferedChecksumIndexInput(NIOFSIndexInput(path="<DATA_PATH>/segments_qt")))
Caused by: java.nio.file.NoSuchFileException: <DATAPATH>/_9xo.si
```

The resulting segment state did create the `segments_N` file, and had a CFS file. But, was missing `.si` and `.cfe`. 

This smells like we have an over-eager cleanup code somewhere. As the rename from `pending_*` seems to have occurred, and the CFS file created, but we deleted the `.si` and `.cfe` files. 

JDK 22

Lucene 9.10

rmuir:
Background:

Historically IndexWriter treated OutOfMemoryError special, for defensive reasons. It was expanded to VirtualMachineError, to try to play it safe in similar disastrous circumstances.

We should treat any Error as a tragedy, as it isn't an Exception, and it isn't something a "reasonable" application should catch. IndexWriter should be reasonable. See #7049 for some of the reasoning.

We can't pretend this will detect any possible scenario that might cause harm, e.g. a jvm bug might simply miscompile some code and cause silent corruption. But we should try harder by playing by the rules.

Closes #13275


uschindler:
We should also check that we always rethrow `Error` and `RuntimeException` in those catch-all `catch (Throwable)` blocks. Otherwise they are silently converted to

I would like to grep on `catch (Throwable` through all Lucene code and check that all is sane with rethrowing errors and runtimes unmodified. This is important especially with MethodHandles.

rmuir:
> We should also check that we always rethrow `Error` and `RuntimeException` in those catch-all `catch (Throwable)` blocks. Otherwise they are silently converted to something different and code here won't see this as tragedy.
> 
> I would like to grep on `catch (Throwable` through all Lucene code and check that all is sane with rethrowing errors and runtimes unmodified. This is important especially with MethodHandles.

This is unchanged from today though, if there are problems, they already affect us with `VirtualMachineError`. This just broadens the behavior check to `Error`.

To be clear, I agree with the cleanup, but it should not be mixed inside this issue and needs to be broken up. There are a couple obvious cleanups in not-so-critical code here, but then there's important stuff, too (e.g. MMap is broke here, and definitely IndexWriter classes). 

Should not be mixed in this, it is just an unrelated cleanup.

uschindler:
All fine, maybe I was not clear! I just wanted to mention the "catch Throwable" issue and that we should additionally deal with it.

rmuir:
Yes, there is real work to do there. I dug into it, I think we will need multiple issues for that one...

ChrisHegarty (approved):

rmuir:
I've been mulling over this PR for a while, not the type of one you want to rush in. But the more I think about it, the more I like it.

* javadocs for Error describe them as `abnormal conditions that should never occur`


* these are Errors IndexWriter is hitting from lucene apis, not application code.
* user can plug into low-level lucene apis such as `Directory`, `Codec`, etc, but the same rules apply
* the fact none of our test suite needed changes is telling.
* All subclasses of LinkageError

uschindler:
> * all subclasses of LinkageError read like "code is broken" to me.

Yes, this is why the code around MMapDir and MethodHandles in general is correct and needs no change. errorprone also complains if you rethrow linking errors from MethodHandles as RuntimeException.

mikemccand:
> AssertionError is ok too I think, if a user wants to run with -ea, let's make it worth their while.

Cool!

mikemccand (approved):
Thanks @rmuir -- this looks great.

I wonder why #7049 is still open after all this time :)  Maybe we can also close it when you merge this.

uschindler (approved):

PR #13279, opened 2024-04-08
Remove unnecessary assert and fix comment in SegmentTermsEnumFrame

vsop-479:
Discussed in https://github.com/apache/lucene/pull/11888.


vsop-479:
@mikemccand 

mikemccand (approved):
Thanks @vsop-479!

PR #13282, opened 2024-04-08
Schedule compileJava after the internal task if it affects source files

### LINKED ISSUE #13240: gradlew regenerate fails with current gradle version
### Description

While working on #13239, I did find another pre-existing condition with `gradlew regenerate` besides the groovy version not being able to read java 21 .class files. 

I think it is new pickiness of the gradle version we are on, that we've just yet to notice until now. it does not like something about how our tasks depend on each other:
```
$ ./gradlew regenerate --rerun-tasks
...
> Task :lucene:core:generateForUtilInternal FAILED

* What went wrong:
A problem was found with the configuration of task ':lucene:core:generateForUtilInternal' (type 'DefaultTask').
  - Gradle detected a problem with the following location: '/home/rmuir/workspace/lucene/lucene/core/src/java/org/apache/lucene/codecs/lucene99/ForUtil.java'.

    Reason: Task ':lucene:core:compileJava' uses this output of task ':lucene:core:generateForUtilInternal' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.

    Possible solutions:
      1. Declare task ':lucene:core:generateForUtilInternal' as an input of ':lucene:core:compileJava'.
      2. Declare an explicit dependency on ':lucene:core:generateForUtilInternal' from ':lucene:core:compileJava' using Task#dependsOn.
      3. Declare an explicit dependency on ':lucene:core:generateForUtilInternal' from ':lucene:core:compileJava' using Task#mustRunAfter.

    For more information, please refer to https://docs.gradle.org/8.4/userguide/validation_problems.html#implicit_dependency in the Gradle documentation.
```

It isn't just this one task nor is it just lucene-core. If you try excluding troublesome tasks such as FOR generation, then you'll just hit it in StandardTokenizer, exclude that, and you'll just hit the same issue in lucene-analyzers with more tokenizers there.

**NOTE**: There's a workaround though for now, I moved past the issue like this:
```console
$ ./gradlew compileJava && ./gradlew regenerate -x compileJava --rerun-tasks
```

dweiss:
Fixes #13240, hopefully.

I've added a manual ordering of compileJava to be scheduled after the internal regeneration tasks, where applicable (which is in most cases). I ran most regeneration without any issues after this patch:

gradlew regenerate -x generateUAX29URLEmailTokenizerInternal --rerun-tasks

That task-wrapping workaround is ugly like hell but I still don't know how to achieve the goals of this in an easier way (checksums, tasks skipping, etc.).

### DIFF
--- a/gradle/generation/antlr.gradle
+++ b/gradle/generation/antlr.gradle
@@ -82,7 +82,10 @@ configure(project(":lucene:expressions")) {
   }
 
   def generateAntlr = wrapWithPersistentChecksums(
-      generateAntlrInternal, [ andThenTasks: ["spotlessJava", "spotlessJavaApply"] ])
+      generateAntlrInternal, [
+          andThenTasks: ["spotlessJava", "spotlessJavaApply"],
+          mustRunBefore: [ "compileJava" ]
+  ])
 
   regenerate.dependsOn generateAntlr
 }
--- a/gradle/generation/forUtil.gradle
+++ b/gradle/generation/forUtil.gradle
@@ -39,7 +39,10 @@ configure(project(":lucene:core")) {
     }
   }
 
-  regenerate.dependsOn wrapWithPersistentChecksums(generateForUtilInternal, [ andThenTasks: ["spotlessJava", "spotlessJavaApply"] ])
+  regenerate.dependsOn wrapWithPersistentChecksums(generateForUtilInternal, [
+          andThenTasks: ["spotlessJava", "spotlessJavaApply"],
+          mustRunBefore: [ "compileJava" ]
+  ])
 }
 
 configure(project(":lucene:backward-codecs")) {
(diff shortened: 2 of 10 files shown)


rmuir:
running a test of this "real quick", thank you. The DFA isn't as hugely painful as it was, I think we made it faster a while back, even on my machine.

rmuir (approved):
It works:
```
BUILD SUCCESSFUL in 9m 55s
127 actionable tasks: 127 executed
```

dweiss:
Ah, forgot to revert that help commit - thank you for remembering about this!

rmuir:
thank you again for wrestling this down with build magic

dweiss:
No problem at all. I really wish this could be done simpler.

PR #13283, opened 2024-04-08
Fix failing BaseKnnVectorQueryTestCase#testTimeout

### LINKED ISSUE #13272: TestKnnByteVectorQuery.testTimeout fails
### Description

Periodic failures for testTimeout, its not locally repeatable. 

```
./gradlew :lucene:core:test --tests "org.apache.lucene.search.TestKnnByteVectorQuery.testTimeout" -Ptests.jvms=2 -Ptests.jvmargs= -Ptests.seed=7F16007D6F6A386B -Ptests.gui=true -Ptests.file.encoding=ISO-8859-1 -Ptests.vectorsize=256
```

kaivalnp:
### Description

Fixes #13272 

The failure happens because of [this assumption](https://github.com/apache/lucene/blob/df0384c34fde8118513e7bc7b79dbde923070c8b/lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java#L788-L791) of the HNSW graph having just 1 level, but we get more than 1 for some seed values -- so the counting timeout of 1 node is exhausted while trying to find the best entry point for the last level, and we consequently get no results, failing [this assertion](https://github.com/apache/lucene/blob/df0384c34fde8118513e7bc7b79dbde923070c8b/lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java#L791)

The assertion wants to check that when an approximate search is performed (some docs are fed to the collector), we return partial results even after a timeout -- and we can check this explicitly in a separate test for `TimeLimitingKnnCollectorManager`

With that test in place, and since we cannot easily control the number of levels in the underlying graph, can we relax the test to check for "at most 1 result" instead of "exactly 1 result"?

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
+++ b/lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
@@ -43,6 +43,8 @@ import org.apache.lucene.index.QueryTimeout;
 import org.apache.lucene.index.StoredFields;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.index.VectorSimilarityFunction;
+import org.apache.lucene.search.knn.KnnCollectorManager;
+import org.apache.lucene.search.knn.TopKnnCollectorManager;
 import org.apache.lucene.store.Directory;
 import org.apache.lucene.tests.analysis.MockAnalyzer;
 import org.apache.lucene.tests.codecs.asserting.AssertingCodec;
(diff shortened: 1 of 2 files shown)


kaivalnp:
Side note: the original command to reproduce from https://github.com/apache/lucene/pull/13202#issuecomment-2037932211:

```
gradlew :lucene:core:test --tests "org.apache.lucene.search.TestKnnByteVectorQuery.testTimeout" -Ptests.jvms=2 -Ptests.jvmargs= -Ptests.seed=7F16007D6F6A386B -Ptests.gui=true -Ptests.file.encoding=ISO-8859-1 -Ptests.vectorsize=256
```

..does not work until we remove the `--tests` parameter like:

```
gradlew :lucene:core:test -Ptests.jvms=2 -Ptests.jvmargs= -Ptests.seed=7F16007D6F6A386B -Ptests.gui=true -Ptests.file.encoding=ISO-8859-1 -Ptests.vectorsize=256
```

Moreover, it deterministically fails with a minimal command of (stripping extra params):

```
gradlew :lucene:core:test -Ptests.jvms=2 -Ptests.seed=7F16007D6F6A386B
```

..which does not reproduce without `tests.jvms`, I'm not entirely sure why..

I tried running the minimal command above using all 3 seed values from the linked comment, and the test fails deterministically before this PR and passes afterwards

vigyasharma (approved):
Good find, I agree with the fix. The current test was fragile. `CountingQueryTimeout` can only count the no. of times `queryTimeout.shouldExit()` was invoked. The test fails if it gets called before a `collect()` call, which is happening here when we find the best entry point.

Since `QueryTimeout` doesn't have access to no. of results collected, I don't see of a reliable way of timing out only after `x` results are picked. We can only provide an upper bound on the assert, not a lower bound.

vigyasharma:
Let us also [unmute](https://github.com/apache/lucene/pull/13273) the broken tests as part of this change?

kaivalnp:
Ah thanks! I fixed and unmuted [one](https://github.com/apache/lucene/blob/26fe2c4b9ca13e7930b818033a04094a5f44d574/lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java#L770) of those earlier, but missed the [other](https://github.com/apache/lucene/blob/26fe2c4b9ca13e7930b818033a04094a5f44d574/lucene/join/src/test/org/apache/lucene/search/join/ParentBlockJoinKnnVectorQueryTestCase.java#L288) -- addressed both now..

benwtrent (approved):
looks good :)

vigyasharma (approved):

vigyasharma:
Backported to `branch_9x`

PR #13284, opened 2024-04-09
Move per-field vector and doc-values readers from TreeMap to HashMap.

jpountz:
Our per-field vector and doc-values readers use `TreeMap`s but don't rely on the iteration order, so these `TreeMap`s can be replaced with more CPU/RAM-efficient `HashMap`s.

The per-field postings reader stays on a `TreeMap` since it relies on the iteration order.


martijnvg (approved):
LGTM

tteofili (approved):

PR #13285, opened 2024-04-09
Add timeout support to AbstractVectorSimilarityQuery

kaivalnp:
### Description

Along similar lines of #13202, adding timeout support for [`AbstractVectorSimilarityQuery`](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java) which performs similarity-based vector searches

While the graph search happens inside `#scorer`, it may go over the configured `QueryTimeout` and we can early terminate it to return whatever partial results are found..

One inherent benefit we have for exact search is that we return a [lazy-loading iterator](https://github.com/apache/lucene/blob/afe982b3ef559f79fceaee307a95245a2c57e9cf/lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java#L147-L148) over all vectors, so this is inherently covered by the `TimeLimitingBulkScorer` (as opposed to [exact search](https://github.com/apache/lucene/blob/afe982b3ef559f79fceaee307a95245a2c57e9cf/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L191-L192) of `AbstractKnnVectorQuery` which manually goes over all vectors to retain the topK during `#rewrite`)


benwtrent:
This seems sane to me.

@vigyasharma what do you think?

benwtrent:
@kaivalnp could you update CHANGES as well?

kaivalnp:
Thanks @benwtrent! Added an entry now..

kaivalnp:
Saw some merge conflicts after a recent commit and resolved those..

kaivalnp:
Hi @benwtrent @vigyasharma could you help review this? Thanks!

lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java
```diff
@@ -105,13 +116,16 @@ public Scorer scorer(LeafReaderContext context) throws IOException {
         LeafReader leafReader = context.reader();
         Bits liveDocs = leafReader.getLiveDocs();
 
+        QueryTimeout queryTimeout = searcher.getTimeout();
```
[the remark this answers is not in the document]
kaivalnp: In this case, the `TimeLimitingKnnCollectorManager` [returns an unwrapped `KnnCollector`](https://github.com/apache/lucene/blob/fc12cc1847e37e588535391a0feed7b6adb75461/lucene/core/src/java/org/apache/lucene/search/TimeLimitingKnnCollectorManager.java#L42-L43) which does not add overhead of time checking (even `null` checks) during graph search (also visible in [benchmarks](https://github.com/apache/lucene/pull/13202#issuecomment-2016087745))

lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java
```diff
@@ -144,22 +158,22 @@ protected boolean match(int doc) {
...
+        } else {
           // Return a lazy-loading iterator
           return VectorSimilarityScorer.fromAcceptDocs(
               this,
               boost,
               createVectorScorer(context),
               new BitSetIterator(acceptDocs, cardinality),
               resultSimilarity);
-        } else if (results.scoreDocs.length == 0) {
-          return null;
```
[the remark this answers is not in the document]
kaivalnp: Yes, it was common in a couple of places so I moved it there to reduce repetition

kaivalnp:
Summary of latest changes:
1. Resolved merge conflicts
2. Moved `CHANGES.txt` entry from 9.11 -> 9.12 since the prior is now released
3. `#Scorer` is now `final` and not overrideable, changed `VectorSimilarityScorer` -> `VectorSimilarityScorerSupplier`

lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java
```diff
@@ -103,16 +114,18 @@ public Explanation explain(LeafReaderContext context, int doc) throws IOExceptio
       public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOException {
         LeafReader leafReader = context.reader();
         Bits liveDocs = leafReader.getLiveDocs();
-        final Scorer vectorSimilarityScorer;
+
+        QueryTimeout queryTimeout = searcher.getTimeout();
+        TimeLimitingKnnCollectorManager timeLimitingKnnCollectorManager =
```
dungba88, line 119: Can we share this variable for all segments? Such as creating it at top-level variable in `createWeight`?

lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java
```diff
@@ -143,27 +156,23 @@ protected boolean match(int doc) {
...
+          if (results.totalHits.relation == TotalHits.Relation.EQUAL_TO
+              // Return partial results only when timeout is met
+              || (queryTimeout != null && queryTimeout.shouldExit())) {
             // Return an iterator over the collected results
-            vectorSimilarityScorer =
-                VectorSimilarityScorer.fromScoreDocs(this, boost, results.scoreDocs);
+            return VectorSimilarityScorerSupplier.fromScoreDocs(boost, results.scoreDocs);
+          } else {
+            // Return a lazy-loading iterator
+            return VectorSimilarityScorerSupplier.fromAcceptDocs(
```
dungba88, line 169: It seems to be a waste that we can't reuse the results from the approximate search (I also saw similar behavior in top-k KnnVectorQuery).

Maybe we can pass the partial results to this method, and we don't need to compute score for those?

dungba88 (commented):
Thank you for this PR. I just left some minor comments/questions

lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java
kaivalnp: We tried to explore this in #12820, but the cost seemed to outweigh the benefit

lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java
kaivalnp: Nice catch, we can reduce unnecessary object creation. I'll update in the next commit

lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java
dungba88: Thanks! It's interesting that we have tried that already.

lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java
```diff
@@ -143,27 +156,23 @@ protected boolean match(int doc) {
...
-            return null;
-          } else {
+          if (results.totalHits.relation == TotalHits.Relation.EQUAL_TO
+              // Return partial results only when timeout is met
+              || (queryTimeout != null && queryTimeout.shouldExit())) {
             // Return an iterator over the collected results
-            vectorSimilarityScorer =
-                VectorSimilarityScorer.fromScoreDocs(this, boost, results.scoreDocs);
+            return VectorSimilarityScorerSupplier.fromScoreDocs(boost, results.scoreDocs);
+          } else {
```
vigyasharma, line 167: Do we also have a test for this case, where we exhaust the filter before hitting timeout? I guess [testFilterWithNoMatches()](https://github.com/apache/lucene/blob/main/lucene/core/src/test/org/apache/lucene/search/BaseVectorSimilarityQueryTestCase.java#L193) tests it but only for `null` QueryTimeout values? Do we need one for non-null timeouts as well?

vigyasharma (approved):
Sorry I somehow missed this PR. Changes look good @kaivalnp , thanks for extending timeout functionality to `*VectorSimilarityQuery`.

Looks like you're planning another iteration addressing [this](https://github.com/apache/lucene/pull/13285/files#r1689584585) comment. We can merge after your changes.

lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java
kaivalnp: Done

lucene/core/src/java/org/apache/lucene/search/AbstractVectorSimilarityQuery.java
kaivalnp: Makes sense, I've added tests to check for a filter + non-null timeout

kaivalnp:
There was a conflict in `CHANGES.txt` after a recent commit, merged from `main` and resolved that
@vigyasharma I've tried to address all open comments, please let me know if something is missing

lucene/core/src/test/org/apache/lucene/search/BaseVectorSimilarityQueryTestCase.java
```diff
@@ -475,6 +479,62 @@ public void testApproximate() throws IOException {
...
+
+      searcher.setTimeout(() -> false); // Set a timeout which is never met
+      assertEquals(numFiltered, searcher.count(filteredQuery));
+
+      searcher.setTimeout(
+          new CountingQueryTimeout(numFiltered - 1)); // Timeout before scoring all filtered docs
+      int filteredCount = searcher.count(filteredQuery);
+      assertTrue(
+          "0 < filteredCount=" + filteredCount + " < numFiltered=" + numFiltered,
+          filteredCount > 0 && filteredCount < numFiltered); // Expect partial results
```
vigyasharma, line 534: So this tests for cases where we timeout before exhausting the filter, nice!

vigyasharma (approved):

kaivalnp:
Thank you @vigyasharma!

PR #13286, opened 2024-04-10
Remove deprecated code

### LINKED ISSUE #13262: Remove deprecated code in `main`
### Description

We have some code that has been deprecated for a long time, see e.g. `IntField` constructors. Let's remove this deprecated code in the `main` branch ahead of 10.0.

Note: we may have code that is only marked as deprecated in `main`. This code shouldn't be removed as users will get the deprecation warning for the first time when they upgrade to 10.0.

iamsanjay:
### Description

closes #13262 

Additionally I removed deprecated code from two more classes: VH_BE_DOUBLE (static member in BitUtil) and variant of BlockJoinSelector.wrap method. Both classes are tagged with @lucene.internal and I searched in Solr and elasticSearch. Let me know If any potential issues arise from its removal, I'll reinstate it.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -98,7 +98,7 @@ API Changes
 
 * GITHUB#13241: Remove Accountable interface on KnnVectorsReader. (Pulkit Gupta)
 
-* GITHUB#13262: Remove deprecated code IntField.IntField(name, value), VH_BE_DOUBLE (static memeber in BitUtil) and
+* GITHUB#13262: Remove deprecated code IntField.IntField(name, value), VH_BE_DOUBLE (static member in BitUtil) and
   BlockJoinSelector.wrap (Sanjay Dutt)
 
 
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -98,6 +98,9 @@ API Changes
 
 * GITHUB#13241: Remove Accountable interface on KnnVectorsReader. (Pulkit Gupta)
 
+* GITHUB#13262: Remove deprecated code IntField.IntField(name, value), VH_BE_DOUBLE (static memeber in BitUtil) and
+  BlockJoinSelector.wrap
+
 
 New Features
 ---------------------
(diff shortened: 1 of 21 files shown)


jpountz:
Thanks for looking into this. I think I'd rather keep `VH_BE_DOUBLE` for now for consistency, since we're also keeping the INT, LONG and FLOAT BE variants.

 - In addition to IntField, can you also remove deprecated constructors from other `*Field` classes such as `DoubleField`?
 - Can you also remove deprecated code in `BooleanQuery` and `ByteBuffersDataInput`?

iamsanjay:
I looked into different Search engines whether they use it or not  -- Many of them still using the deprecated code.
1. `LongPoint.newDistanceFeatureQuery` (ElasticSearch uses it https://github.com/search?q=repo%3Aelastic%2Felasticsearch+LongPoint.newDistanceFeatureQuery&type=code and OpenSearch as well https://github.com/search?q=repo%3Aopensearch-project%2FOpenSearch%20newDistanceFeatureQuery&type=code) 
2.  BooleanQuery.TooManyClauses (Elastic Search https://github.com/search?q=repo%3Aelastic%2Felasticsearch%20TooManyClauses&type=code)
3. BooleanQuery.getMaxClauseCount and BooleanQuery.setMaxClauseCount are still in use.

Should I go ahead and remove it?

jpountz:
@iamsanjay Yes please go ahead. All these problems should be easy to address by these various projects. Thanks for checking!

iamsanjay:
I added MIGRATE docs. Let me know if that's not required.
There is one more class pending `SortField`. I am still looking into some of the test cases that are failing , we may have to remove those test cases along with deprecated code. We can also deal with it in separate pull request.

jpountz (approved):
I'm not sure how to handle the `SortField` deprecations either, but what you have here looks pretty good to me.

lucene/CHANGES.txt
```diff
@@ -98,6 +98,9 @@ API Changes
 
 * GITHUB#13241: Remove Accountable interface on KnnVectorsReader. (Pulkit Gupta)
 
+* GITHUB#13262: Remove deprecated code IntField.IntField(name, value), VH_BE_DOUBLE (static member in BitUtil) and
```
jpountz, line 101: Can you update the CHANGES entry?

jpountz (approved):
This is great.

PR #13289, opened 2024-04-10
Integrate merge-time index reordering with the intra-merge executor.

jpountz:
Index reordering can benefit greatly from parallelism, so it should try to use the intra-merge executor when possible. Because the API only exposes an `Executor` and not a `ForkJoinPool`, parallelism is a bit less good due to some threads potentially waiting for tasks to finish when they could be doing work stealing instead. We should look into whether we can improve this in the future.


jpountz:
I improved the PR by taking advantage of TaskExecutor's work stealing logic. I plan on merging soon if there are no objections.

benwtrent:
@jpountz I think this is a good idea. One concern is that the CMS may not actually use parallelism on tiny merges. Let me separate out the change in this PR to improve our testing around this: https://github.com/apache/lucene/pull/13475/files

PR #13291, opened 2024-04-11
Remove unnecessary calculating for termLen.

vsop-479:
### Description

It seems `tmpLen` is only used after scanning done, We can remove unnecessary calculating for it in `SegmentTermsEnumFrame.scanToTermNonLeaf`'s loop.


vsop-479:
@mikemccand 

mikemccand (approved):
Thanks @vsop-479 -- looks fine.

PR #13296, opened 2024-04-12
Convert FieldEntry to record

iamsanjay:
Task from #13207 

Avoid the usage of compact constructors because the compact constructor can only access the components of record, and most of FieldEntry relies on  IndexInput object which is not a part of record class. Hence, The purposed code is using the static factory method to create the FieldEntry. 

However, there is another way where we can make IndexInput a component of the record class and then access it in compact constructor. While it moves the code inside compact constructor, but the code accessing IndexInput has to be surrounded with try-catch as compact constructors cannot throw checked exception.


lucene/MIGRATE.md
```diff
@@ -180,6 +180,8 @@ access the members using method calls instead of field accesses. Affected classe
 
 - `IOContext`, `MergeInfo`, and `FlushInfo` (GITHUB#13205)
 - `BooleanClause` (GITHUB#13261)
+- `FieldEntry`, a static nested class, from `Lucene90HnswVectorsReader`, `SimpleTextKnnVectorsReader`,
+  `Lucene99FlatVectorsReader`, `Lucene99HnswVectorsReader`, and `Lucene99ScalarQuantizedVectorsReader`
```
jpountz, line 184: No need to mention these classes in the MIGRATE file, they are implementation details that users don't interact with.

jpountz (approved):
The change looks good to me, I'd like @uschindler to have a look since he knows more about records than I do.

lucene/MIGRATE.md
uschindler: You can remove this as it is internal classes.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatVectorsReader.java
```diff
@@ -302,29 +302,50 @@ public void close() throws IOException {
...
+          similarityFunction,
+          vectorEncoding,
+          vectorDataOffset,
+          vectorDataLength,
+          dimension,
+          size,
+          OrdToDocDISIReaderConfiguration.fromStoredMeta(input, size));
+    }
+
+    public FieldEntry(
```
uschindler, line 336: does not matter as record is private, but the "public" here could be removed.

Previously this class only had one ctor, so please merge both.

uschindler (changes requested):
Looks OK. The static create method doing the I/O and invoking default ctor look fine to me.

There are small inconsistencies:
- one get 2 ctors, one public but the original had only one ctor. Mayb merge ll and only have one ctor or create method.
- MIGRATE.md can be removed as this is all private APIs.

Uwe

uschindler:
> The change looks good to me, I'd like @uschindler to have a look since he knows more about records than I do.

They are not too complicated, you just have to think simpler and use the special syntax. Something to keep in mind when you add records:
- The default ctor cannot be hidden, so when you add public records, make sure all consistency checks are done in the default ctor (the one in the source file having no parameters: `public Classname {`). This ensures that all checks are at one place. My recommendation is to override the default ctor. You can add additional ctors for easy usage for common special cases (see IOContext class for examples), calling the default one with this().
- Be sure to remove equals, hashcode, toString: it does not warn you when you define your own

uschindler (approved):
Hi,
I did not closely look into every line if the variables and fields are correctly assigned, but as all record components are final we know that everything works fine. Tests also pass.

The `static create()` methods are a good workaround for the former ctors as they have much more logic than a plain ctor.

lucene/CHANGES.txt
```diff
@@ -102,6 +102,9 @@ API Changes
   Additionally, deprecated methods have been removed from ByteBuffersIndexInput, BooleanQuery and others. Please refer
   to MIGRATE.md for further details. (Sanjay Dutt)
 
+* GITHUB#13207: Converted the FieldEntry, a static nested class, from Lucene90HnswVectorsReader, SimpleTextKnnVectorsReader,
```
uschindler, line 105: small change, please use the PR number here. As this is internal classes also move the change entry to "Other".

uschindler: You can also add you name.

uschindler:
Thanks, will merge soon.

PR #12408, opened 2023-06-30
Initialize facet counting data structures lazily

gsmiller:
This change proposes some faceting optimizations for situations where there are no hits to facet over. While this seems like an odd scenario, at Amazon product search we actually have some situations where this can become common (sparse queries that don't have results in some segments, or have no results altogether).

You could argue that the calling code should handle this optimization and avoid faceting altogether if there are no hits, but that only solves for the case of no results, not no results in some segments. I think it's also nice to move this optimization into the faceting module so that every user doesn't have to do this themselves.

Curious what people think of this idea. Happy to hear feedback, counterarguments, etc. :)

This change covers:
* Taxonomy faceting
  * FastTaxonomyFacetCounts
  * TaxonomyFacetIntAssociations
  * TaxonomyFacetFloatAssociations
* SSDV faceting
  * SortedSetDocValuesFacetCounts
  * ConcurrentSortedSetDocValuesFacetCounts
  * StringValueFacetCounts
* Range faceting:
  * LongRangeFacetCounts
  * DoubleRangeFacetCounts
* Long faceting:
  * LongValueFacetCounts

Left for a future iteration (I'll open a follow up issue if we move forward with this one):
* RangeOnRange faceting
* FacetSet faceting

PR #13300, opened 2024-04-13
Backport to 9x: Initialize facet counting data structures lazily #12408

stefanvodita:
We can bring this optimisation from #12408 to 9x. It would make it easier to backport #12966 afterwards. I think the intention was to backport at first, since [the CHANGES entry is under 9.8 in the commit](https://github.com/apache/lucene/blob/179b45bc23e4496278b7058811577b66ef3af77d/lucene/CHANGES.txt#L164).


stefanvodita:
@gsmiller - I'd like your opinion as the author of the original commit. Am I missing some reason why we shouldn't backport?

stefanvodita:
I went through this change once more to make sure it's safe to push. I would still like a second opinion, just in case I'm missing something, otherwise I'll merge this in a few days.

msokolov (approved):
it seems reasonable, if a bit messy (so many conditionals!) ... but this is what was committed to main branch, right? Did you have to make many changes to pull it in to 9x? Anyway LGTM

stefanvodita:
Thank you for the review Mike! I did have to change a couple things:
1. Added a new constructor to `TaxonomyFacets`, which takes a `FacetsCollector`, since I couldn't change the existing one.
2. Updated `TaxonomyFacetCounts` too, which no longer exists on `main`.

PR #13302, opened 2024-04-14
Align toString methods in geo module

timgrein:
### Description

Some `toString` methods in the `geo` module sometimes use uppercase or the class name. I've aligned them to use always the class name.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/geo/XYPolygon.java
+++ b/lucene/core/src/java/org/apache/lucene/geo/XYPolygon.java
@@ -187,6 +187,7 @@ public final class XYPolygon extends XYGeometry {
   @Override
   public String toString() {
     StringBuilder sb = new StringBuilder();
+    sb.append("XYPolygon");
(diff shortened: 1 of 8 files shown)


lucene/core/src/java/org/apache/lucene/geo/XYPolygon.java
```diff
@@ -187,6 +187,7 @@ public boolean equals(Object obj) {
   @Override
   public String toString() {
     StringBuilder sb = new StringBuilder();
+    sb.append("XYPolygon");
```
slow-J, line 190: While were doing this standardization, can we also add the name to the others that are missing it too? Like Polygon https://github.com/apache/lucene/blob/e84b9229a5f13de545c187cae3c5f4bc2b9fc694/lucene/core/src/java/org/apache/lucene/geo/Polygon.java#L210

slow-J (approved):

tteofili (approved):
LGTM

PR #541, opened 2021-12-14
LUCENE-10315: Speed up BKD leaf block ids codec by a 512 ints ForUtil

gf2121:
Elasticsearch (which based on lucene) can automatically infers types for users with its dynamic mapping feature. When users index some low cardinality fields, such as gender / age / status... they often use some numbers to represent the values, while ES will infer these fields as long, and ES uses BKD as the index of long fields. When the data volume grows, building the result set of low-cardinality fields will make the CPU usage and load very high.

This is a flame graph we obtained from the production environment:
[flame.svg](https://issues.apache.org/jira/secure/attachment/13037560/13037560_addall.svg)

It can be seen that almost all CPU is used in addAll. When we reindex long to keyword, the cluster load and search latency are greatly reduced ( We spent weeks of time to reindex all indices... ). I know that ES recommended to use keyword for term/terms query and long for range query in the document, but there are always some users who didn't realize this and keep their habit of using sql database, or dynamic mapping automatically selects the type for them. All in all, users won't realize that there would be such a big difference in performance between long and keyword fields in low cardinality fields. So from my point of view it will make sense if we can make BKD works better for the low/medium cardinality fields.

As far as i can see, for low cardinality fields, there are two advantages of keyword over long:

1. ForUtil used in keyword postings is much more efficient than BKD's delta VInt, because its batch reading (readLongs) and SIMD decode.
2. When the query term count is less than 16, TermsInSetQuery can lazily materialize of its result set, and when another small result clause intersects with this low cardinality condition, the low cardinality field can avoid reading all docIds into memory.

This approach is targeting to solve the first point. The basic idea is trying to use a 512 ints ForUtil for BKD ids codec. I benchmarked this optimization by mocking some random LongPoint and querying them with PointInSetQuery.

**Benchmark Result**
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/gf/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
[...]

PR #13306, opened 2024-04-15
Performance improvements to use RWLock to access LRUQueryCache

boicehuang:
Elasticsearch (which based on lucene) can automatically infer types for users with its dynamic mapping feature. When users index some low cardinality fields, such as gender / age / status... they often use some numbers to represent the values, while ES will infer these fields as long, and ES uses BKD as the index of long fields. 

Just as #541 said, when the data volume grows, building the result set of low-cardinality fields will make the CPU usage and load very high even if we use a boolean query with filter clauses for low-cardinality fields. 

One reason is that it uses a ReentrantLock to limit accessing LRUQueryCache. QPS and costs of their queries are often high,  which often causes trying locking failures when obtaining the cache, resulting in low concurrency in accessing the cache.

So I replace the ReentrantLock with a ReentrantReadWriteLock. I only use the read lock when I need to get the cache for a query, 

I benchmarked this optimization by mocking some random LongPoint and querying them with one PointInSetQuery with bool filter.

doc count | field cardinality | query terms count | baseline QPS | candidate QPS | diff percentage
-- | -- | -- | -- | -- | --
30000000 | 10 | 1 | 2481 | 5102 | 105.6%
30000000 | 1000000 | 1 | 6396 | 6596.48 |  3.1%

I think this change can help filter queries that need to query low-cardinality fields.


lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -177,7 +180,6 @@ public boolean test(LeafReaderContext context) {
    * @lucene.experimental
    */
   protected void onHit(Object readerCoreKey, Query query) {
-    assert lock.isHeldByCurrentThread();
```
benwtrent, line 180: It seems to me that the accuracy `hitCount` and `missCount` is now broken with this structure.

The volatile parameters can now be updated by more than one read-thread at a time, thus adding race conditions to their increments. 

I do not think we want to add this level of uncertainty to cache statistics. 

What happens to performance if you adjust `missCount` and `hitCount` to be `LongAdder` objects? This way the statistics are only locked when read, but still allows multiple lockless thread updates?

boicehuang:
This optimization also benefits high-cost querys such as terms query with 10000 terms,  by reading cache more frequently instead of searching inverted index

doc count | field cardinality | query terms count | baseline QPS | candidate QPS | diff percentage
-- | -- | -- | -- | -- | --
30000000 | 1000000 | 10000 | 160 | 473 | 191.9%

boicehuang:
This optimization also benefits high-cost querys such as terms query with 10000 terms,  by reading cache more frequently instead of searching inverted index

doc count | field cardinality | query terms count | baseline QPS | candidate QPS | diff percentage
-- | -- | -- | -- | -- | --
30000000 | 1000000 | 10000 | 160 | 473 | 191.9%

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -265,7 +269,6 @@ boolean requiresEviction() {
   }
 
   CacheAndCount get(Query key, IndexReader.CacheHelper cacheHelper) {
-    assert lock.isHeldByCurrentThread();
     assert key instanceof BoostQuery == false;
     assert key instanceof ConstantScoreQuery == false;
     final IndexReader.CacheKey readerKey = cacheHelper.getKey();
```
benwtrent, line 274: As with anything multi-threaded, this is tricky. How we are actually doing the LRU, is via the `uniqueQueries` which is a `LinkedHashMap` constructed by `new LinkedHashMap<>(16, 0.75f, true);`

Which means, a `get` against this will actually do an UPDATE to the LinkedHashMap order.  If multiple readers attempt to do this at the same time, I would think this would then become indeterminate. 

From the java docs to `LinkedHashMap`

```
 * <p><strong>Note that this implementation is not synchronized.</strong>
 * If multiple threads access a linked hash map concurrently, and at least
 * one of the threads modifies the map structurally, it <em>must</em> be
 * synchronized externally.  This is typically accomplished by
 * synchronizing on some object that naturally encapsulates the map.
```

Since we have order dictated at access, not on insertion, we are modifying the map structurally on read.

benwtrent: So, `Collections.synchronizedMap` should be used, or locking manually around this particular `get`. 

And that means now we are locking on read again, which might mean all performance gains go away.

boicehuang: can we replace `LinkedHashMap ` with `ConcurrentHashMap` ?

boicehuang: I adjust `cache` AND `uniqueQueries` to be `ConcurrentHashMap` objects and get the benchmark results as follows.

doc count | field cardinality | query point | baseline QPS | candidate QPS | diff percentage
-- | -- | -- | -- | -- | --
30000000 | 10 | 1 | 2481 | 5102 | 105.6%
30000000 | 10 | 1 | 2481 | 4843 |  95.2% (using `LongAdder `)
30000000 | 10 | 1 | 2481 | 3851 |  55% (using `LongAdder ` And `ConcurrentHashMap`)

boicehuang: In ConcurrentHashMap, although the order of insertion operations is uncertain, it ensures the thread safety of insertion operations. Multiple threads can perform insert operations simultaneously without causing data inconsistency or corruption.

benwtrent: How do you suggest that we achieve the "LRU" aspect of this cache without knowing the queries that were used most recently?

The point of the linked hashmap is that it orders by last access. We need something like this to achieve LRU expiration

boicehuang: I understand your point. I try to use `Collections.synchronizedMap` and get the benchmark results as follows

doc count | field cardinality | query point | baseline QPS | candidate QPS | diff percentage
-- | -- | -- | -- | -- | --
30000000 | 10 | 1 | 2481 | 5102 | 105.6%
30000000 | 10 | 1 | 2481 | 4843 |  95.2% (using `LongAdder `)
30000000 | 10 | 1 | 2481 | 3751 |  51% (using `LongAdder ` And `Collections.synchronizedMap`)

It seems that performance can be improved if the time spent occupying the lock is minimized.

boicehuang: but I don't think it is the best to use `Collections.synchronizedMap`,  we can use `ConcurrentLinkedDeque` to achieve LRU expiration better. Here is an example.
```
public class LRUCache<K, V> {
    private final int capacity;
    private ConcurrentMap<K, V> cache;
    private ConcurrentLinkedDeque<K> keys;

    public LRUCache(int capacity) {
        this.capacity = capacity;
        this.cache = new ConcurrentHashMap<>(capacity);
        this.keys = new ConcurrentLinkedDeque<>();
    }

    public V get(K key) {
        V value = cache.get(key);
        if (value != null) {
            keys.remove(key); // remove first occurrence of key
            keys.addFirst(key); // add key to the head
        }
        return value;
    }

    public void put(K key, V value) {
        if (cache.containsKey(key)) {
            keys.remove(key); // remove first occurrence of key
        }

        while (keys.size() >= capacity) {
            K lastKey = keys.removeLast();
            if (lastKey != null) {
                cache.remove(lastKey);
            }
        }

        cache.put(key, value);
        keys.addFirst(key);
    }
}
```

benwtrent: switching from an LRU to an LFU would require more consideration.

But, your debugging here shows that there is a fair amount of performance being left on the table. I am surprised to see how much of a bottle neck cache access actually is.

boicehuang: Yes, So I want to try best to optimize this neck. I think `uniqueQueries` should be `Collections.synchronizedMap` to achieve LRU expiration, but `cache` can be `ConcurrentHashMap` to improve performance.

benwtrent: I am currently trying to read through all the weird concurrency paths here. It's tricky to get correct.

`IdentityHashMap` is required still for the `Map<IndexReader.CacheKey, LeafCache>`. Changing to any other map will change behavior as now `IndexReader.CacheKey` isn't by object identity only, but object equality. This is a change to caching behavior. 


`uniqueQueries` is used for the LRU portion of the cache, and the class key set is provided in `mostRecentlyUsedQueries`. Consequently, on `get` for `uniqueQueries`, that is a structural modification of the LinkedHashMap.

Then once we get the query, we look in `cache` with just a `get` on that map. This is NOT a structural modification for an IdentityHashMap.

boicehuang: I agree, but  IdentityHashMap is not thread-safe.

benwtrent: > I agree, but IdentityHashMap is not thread-safe.

I may be wrong, but I though if the writing to `IdentityHashMap` was protected (which it is via the write lock, which would block the readers), that means during read it isn't being mutated, and writes are fully synced. Consequently, its perfectly fine to read from multiple threads.

boicehuang: In my optimization, the writing to `cache` object  was still  protected via the write lock,  but reading `cache` object concurrently via making `cache` object to be `ConcurrentHashMap`. This is the key point of my optimization. I 
am also trying to read through all the concurrency paths with a picture.

boicehuang: I drew 2 flow charts and tried to describe all the concurrency paths.
LRUCache 优化结构2

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -628,7 +632,7 @@ private class LeafCache implements Accountable {
 
     LeafCache(Object key) {
       this.key = key;
-      cache = new IdentityHashMap<>();
+      cache = Collections.synchronizedMap(new IdentityHashMap<>());
```
benwtrent, line 635: I don't understand this. Aren't all accesses to `LeafCache` protected by the LRUQueryCache read/write locks?

Since `LeafCache` isn't a static class, it should have access to the enclosing class's lock.

For testing safety, putIfAbsent, remove, onDocIdSetCache, and onDocIdSetEviction should all do a `assert writeLock.isHeldByCurrentThread();`

boicehuang: Sorry, I misunderstood before. Since there is a read lock, concurrent access to IdentityHashMap does not require additional synchronized lock. I optimized the code as per your suggestion.

boicehuang: LRUCache 优化结构3

benwtrent:
@boicehuang what are the new numbers for your benchmarks for the current iteration? Indeed we will be synchronizing more, so I wonder if we will still see improvement.

boicehuang:
> @boicehuang what are the new numbers for your benchmarks for the current iteration? Indeed we will be synchronizing more, so I wonder if we will still see improvement.

doc count | field cardinality | query point | baseline QPS | candidate QPS | diff percentage | diff
-- | -- | -- | -- | -- | -- | --
  30000000 | 10 | 1 | 2481 | 4428 |  78%|  using `LongAdder`. `uniqueQueries` to be `Collections.synchronizedMap`.   `cache` to be `IdentityHashMap`.

jpountz (commented):
Query caching in Lucene is designed to update the content of the cache infrequently (e.g. because we wait until we've seen a query multiple times before caching it), so it makes sense that switching to a RWLock reduced contention.

The change looks correct, I just left minor comments.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -911,15 +919,15 @@ public BulkScorer bulkScorer(LeafReaderContext context) throws IOException {
       }
 
       // If the lock is already busy, prefer using the uncached version than waiting
-      if (lock.tryLock() == false) {
+      if (writeLock.tryLock() == false) {
```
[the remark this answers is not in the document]
boicehuang: Yes, read lock can also be used here

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -96,14 +98,15 @@ public class LRUQueryCache implements QueryCache, Accountable {
...
+  private final ReentrantReadWriteLock.WriteLock writeLock;
   private final float skipCacheFactor;
 
   // these variables are volatile so that we do not need to sync reads
   // but increments need to be performed under the lock
   private volatile long ramBytesUsed;
-  private volatile long hitCount;
-  private volatile long missCount;
+  private volatile LongAdder hitCount;
+  private volatile LongAdder missCount;
```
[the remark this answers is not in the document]
boicehuang: ok

boicehuang:
Current improvement number of the submitted code version.

doc count | field cardinality | query point | baseline QPS | candidate QPS | diff percentage | diff
-- | -- | -- | -- | -- | -- | --
  30000000 | 10 | 1 | 2481 | 4408 |  78%|  using `LongAdder`. `uniqueQueries` to be `Collections.synchronizedMap`.   `cache` to be `IdentityHashMap`.

benwtrent (approved):
I think this is ready for merging. I can do the merging, but won't back port to 9x until we see nightlies. They might catch something we missed. 

@boicehuang could you add a changes entry to 9.11 under optimizations?

boicehuang:
> I think this is ready for merging. I can do the merging, but won't back port to 9x until we see nightlies. They might catch something we missed.
> 
> @boicehuang could you add a changes entry to 9.11 under optimizations?

Do I need to open an issue at https://issues.apache.org/jira/projects/LUCENE/issues ?  To get a Lucene-XXXXX number.

boicehuang:
> I think this is ready for merging. I can do the merging, but won't back port to 9x until we see nightlies. They might catch something we missed.
> 
> @boicehuang could you add a changes entry to 9.11 under optimizations?

Finished

Issue #13281, opened 2024-04-08
Deprecate `COSINE` before Lucene 10 release


benwtrent:
### Description

Over a couple disparate discussions over various PRs/issues, we have tossed around the idea of deprecating `COSINE`. 

To me, this makes sense, we shouldn't have `COSINE` at all IMO. Either users should normalize before indexing, or use max-inner product. 

The question then becomes "What about `byte` vectors"? It seems to me that users should still use `DOT_PRODUCT` and `MAX_INNER_PRODUCT`.

jmazanec15:
@benwtrent Is the main reason to deprecate to stop enabling users to setup non-optimal configurations? Or are there limitations cosine similarity imposes on implementation/optimization for other distance metrics?

benwtrent:
@jmazanec15 

Mainly because cosine has no benefit over normalizing & using `dot_product` and maintaining optimized Cosine similarity functions is an unnecessary burden for Lucene.

jmazanec15:
Thanks @benwtrent that makes sense

wurui90:
One question: the `float` vector has a straightforward definition of "normalized to 1.0", i.e. ||a|| = 1, e.g. (-1.0, 0.0) and (0.0, 1.0) are both normalized to 1. In comparison, do we require `byte` vectors to be normalized to 127, like (-127, 0) and (0, 127) are normalized?

benwtrent:
> do we require byte vectors to be normalized to 127, like (-127, 0) and (0, 127) are normalized?

No, we do not. For regular dot-product, their magnitudes should be the same.

benwtrent:
@wurui90 I need a think a bit more about your question. Admittedly, I was thinking it would be rather simple for users who quantize their vectors to ensure they are quantizing from normalized vectors, thus, magnitude doesn't matter. 

Maybe we cannot deprecate cosine :(. The deprecation currently exists in 10 & 9.12, but can be easily reverted.

I am not sure what to do for users who quantize their own vectors & rely on cosine. 

@msokolov what do you think?

I am wondering if we should, at a minimum, deprecate the cosine functions, and instead store the magnitude, this way the comparisons for the "COSINE" similarity metric are actually the dot-product ones, and the magnitude is stored to adjust the resulting dot-product (thus all that code can go away).

[...]

PR #13308, opened 2024-04-15
Deprecate COSINE VectorSimilarity function

Pulkitg64:
### Description

Tries to solve #13281 by deprecating `COSINE` VectorSimilarity function for Vector Search.


Pulkitg64:
`TestInt8HnswBackwardsCompatibility.java` and `TestBasicBackwardsCompatibility` test cases failing for Lucene versions older than `LUCENE_10_0_0`. For older Lucene version, we try to read Index `directory` from resources folder. For example for Lucene-9.10.0, the directory path is located at `lucene/backward-codecs/build/resources/test/org/apache/lucene/backward_index/int8_hnsw.9.10.0.zip`.

As per my understanding the segment present at above directory already has `KNNVectorField` with MAXIMUM_INNER_PRODUCT as vector similarity function and in above test we try to change similarity function to `DOT_PRODUCT` due to which there is conflict happening when adding more docs to it and hence these test cases failing.

lucene/core/src/java/org/apache/lucene/index/VectorSimilarityFunction.java
```diff
@@ -61,24 +60,6 @@ public float compare(byte[] v1, byte[] v2) {
...
-    @Override
-    public float compare(float[] v1, float[] v2) {
-      return Math.max((1 + cosine(v1, v2)) / 2, 0);
-    }
-
-    @Override
-    public float compare(byte[] v1, byte[] v2) {
-      return (1 + cosine(v1, v2)) / 2;
-    }
-  },
```
benwtrent, line 80: You cannot do this. These enumerations have default ordinals. We need to either:

 - detach these ordinal IDs from the implementation
 - Switch from an enumerator to an `id->implementation` idea. 

Even if deprecated, Lucene needs to be able to read indices from 9x -> 10. So, cosine cannot go away on read.

benwtrent (changes requested):

lucene/core/src/java/org/apache/lucene/index/VectorSimilarityFunction.java
Pulkitg64: Thanks @benwtrent for commenting!

> These enumerations have default ordinals

Yeah, I guess that's why some tests are failing. For Lucene9.x the ordinal is getting mapped to `MAXIMUM_INNER_PRODUCT` which was earlier mapping to `COSINE`.

> We need to either detach these ordinal IDs from the implementation or Switch from an enumerator to an id-implementation idea.

I am not sure how to do this but let me try to give it a shot.

> Even if deprecated, Lucene needs to be able to read indices from 9x -> 10. So, cosine cannot go away on read.

Oh okay, that means we need to remove support for COSINE just from indexing side not from searching side?

jpountz:
> Oh okay, that means we need to remove support for COSINE just from indexing side not from searching side?

Correct. Practically, this means keeping the enum constant but marking it as deprecated, and modifying IndexingChain to fail using COSINE on 10.x Lucene indexes.

It would be nice to also add a new file format under lucene/codecs that does cosine similarity, like we just added for hamming distance. This could help with the migration for some users.

Pulkitg64:
Thanks @jpountz for confirming and suggestions.

lucene/core/src/java/org/apache/lucene/index/VectorSimilarityFunction.java
uschindler: I am a bit confused why this still happens. @ChrisHegarty already submitted a PR to have a separate mapping from file format's internal ordinal numbers to enum constants.
Maybe this was not yet applied, let me check it out.

uschindler: See here: #13119

Maybe the change there was not the whole story. If we have another places where original numbers of enum constants are used, let's use same approach per codec version.

Maybe we need to only change the historical codecs?

uschindler: Anyways, we have to keep the constant in 10.x for reading old indexes.

lucene/core/src/java/org/apache/lucene/codecs/lucene94/Lucene94FieldInfosFormat.java
```diff
@@ -302,7 +301,6 @@ private static VectorSimilarityFunction getDistFunc(IndexInput input, byte b) th
       List.of(
           VectorSimilarityFunction.EUCLIDEAN,
           VectorSimilarityFunction.DOT_PRODUCT,
-          VectorSimilarityFunction.COSINE,
```
uschindler, line 305: Here is exactly the problem. There must be a null entry, so the ordinals in index format stay alive.

lucene/core/src/java/org/apache/lucene/index/VectorSimilarityFunction.java
uschindler: I found the issue. #13119 is fine. The problem was that inside the codec the ordinal to enum constant was changed, which is backwards incompatible.

Basically the mapping in the codec needs to be changed while writing to reject the deprecated constant.

lucene/core/src/java/org/apache/lucene/codecs/lucene94/Lucene94FieldInfosFormat.java
ChrisHegarty: ++ The reason for the list here is exactly to support and allow such evolution and separation from the enum ordinals. A null here trivially need to a change of List implementation type.   And tests need to be carefully updated.

uschindler: of course, null does not work with default lists :-) Also downstream the code needs to check for a null value.

In any case we can later also go back to a Map<Integer,VectorSimilarity> or to an array.

Pulkitg64: Thanks @uschindler @ChrisHegarty  for the pointers.

lucene/core/src/java/org/apache/lucene/index/VectorSimilarityFunction.java
```diff
@@ -61,24 +60,6 @@ public float compare(byte[] v1, byte[] v2) {
     }
   },
 
-  /**
-   * Cosine similarity. NOTE: the preferred way to perform cosine similarity is to normalize all
```
benwtrent, line 65: I think it would be good to deprecate this first, get all the internal paths updated to not use it. 

Then the deprecation can be merged and backported. Then work on removing it from main.

Mostly because the deprecation path and internal code usage will all have to occur anyways :)

lucene/core/src/java/org/apache/lucene/util/quantization/ScalarQuantizedRandomVectorScorer.java
```diff
@@ -30,23 +28,6 @@
 public class ScalarQuantizedRandomVectorScorer
     extends RandomVectorScorer.AbstractRandomVectorScorer<byte[]> {
 
-  public static float quantizeQuery(
```
benwtrent, line 33: Even in Lucene 10, we will need to support quantizing a query sent to a Lucene index that uses Cosine.

lucene/core/src/java/org/apache/lucene/index/VectorSimilarityFunction.java
```diff
@@ -61,24 +60,6 @@ public float compare(byte[] v1, byte[] v2) {
...
-    public float compare(float[] v1, float[] v2) {
-      return Math.max((1 + cosine(v1, v2)) / 2, 0);
-    }
-
-    @Override
-    public float compare(byte[] v1, byte[] v2) {
-      return (1 + cosine(v1, v2)) / 2;
-    }
-  },
-
```
benwtrent, line 81: I think this will get tricky. If you remove this ordinal (I think it would be wise to deprecate first, work through the warnings, etc. get everything passing, etc.), field info & knn vector format readers will need to somehow distinguish between "Max inner product" (whose ordinal decreases to cosine's) and "cosine but in an old index version".

Maybe on write, we adjust the ordinal dynamically, knowing what they are (thus correcting MIP's ordinal on write, allowing for readers to distinguish).

benwtrent (commented):
Left a larger review. 

I think it would be prudent to first attempt to deprecate and handle internal usage, etc. After that is merged and backported, another PR can be made to fully remove it from main only on writes.

At least, that would make reviewing & testing this simpler :)

lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestBasicBackwardsCompatibility.java
```diff
@@ -98,8 +99,14 @@ public class TestBasicBackwardsCompatibility extends BackwardsCompatibilityTestB
   private static final int KNN_VECTOR_MIN_SUPPORTED_VERSION = LUCENE_9_0_0.major;
   private static final String KNN_VECTOR_FIELD = "knn_field";
   private static final FieldType KNN_VECTOR_FIELD_TYPE =
-      KnnFloatVectorField.createFieldType(3, VectorSimilarityFunction.COSINE);
+      KnnFloatVectorField.createFieldType(3, VectorSimilarityFunction.DOT_PRODUCT);
```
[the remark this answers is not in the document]
Pulkitg64: Make sense.

Pulkitg64:
Thanks @benwtrent for all the feedback.
 I have raised another revision just to mark the COSINE function as deprecated. I didn't find any internal usages of this function.
 In a followup PR, I will remove COSINE from writes for LUCENE version >=10.0.0

benwtrent:
@Pulkitg64 could you deprecate also `VectorUtil.cosine`? That is technically a public API.

lucene/CHANGES.txt
```diff
@@ -102,6 +102,8 @@ API Changes
   Additionally, deprecated methods have been removed from ByteBuffersIndexInput, BooleanQuery and others. Please refer
   to MIGRATE.md for further details. (Sanjay Dutt)
 
+* GITHUB#13281: Mark COSINE VectorSimilarityFunction as deprecated. (Pulkit Gupta)
+
```
benwtrent, line 106: This should be under 9.11 as we will backport the deprecation as a way to warn users that it is going away in 10.0

Pulkitg64: Yes, Make sense. Will change in next revision
Thanks for mentioning.

Pulkitg64: Done

ChrisHegarty (approved):
LGTM, though you it would be good get another reviewers approval too.

Pulkitg64:
Hi @benwtrent, could we merge this change, if everything looks good to you?

benwtrent:
@Pulkitg64 sorry for radio silence, I was out of office and swamped with other things. If you could move the change log to 9.12, I can merge and backport.

Pulkitg64:
I somehow messed up with my branch while pulling changes from `main` branch. @benwtrent I have created a new PR for marking the cosine VectorSImilarity function as deprecated.

https://github.com/apache/lucene/pull/13473

Sorry for the mess!

PR #13310, opened 2024-04-16
Narrow MASKS and UTF8Byte.value's type, assign MASKS by left shift in UTF32ToUTF8

vsop-479:
### Description

Change `MASKS` from `int[]` to `byte[]`, and assign it with left shift in `UTF32ToUTF8`.

Edit: 
This change also implemented: 
Only set first byte value for tmp UTF8.
Change `UTF8Byte.value` type from `int` to `byte`.


vsop-479:
Add optimization:
Only set first byte for tmpUTF8.

vsop-479:
@mikemccand 

mikemccand (approved):
Thanks @vsop-479 -- nice to see another `TODO` done!  Silly that this code was using `int` to represent what is really only a `byte`.

PR #13311, opened 2024-04-17
Use jdk11 primitives to allow backport to branch_9x

vigyasharma:
`BaseKnnVectorQueryTestCase#testTimeLimitingKnnCollectorManager` uses some new java primitives that are breaking on backport to branch_9x, which builds with OpenJdk 11. We don't really need the new APIs, this small change allows us to backport without conflicts.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
+++ b/lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
@@ -781,7 +781,7 @@ abstract class BaseKnnVectorQueryTestCase extends LuceneTestCase {
       TimeLimitingKnnCollectorManager noTimeoutManager =
           new TimeLimitingKnnCollectorManager(delegate, null);
       KnnCollector noTimeoutCollector =
-          noTimeoutManager.newCollector(Integer.MAX_VALUE, searcher.leafContexts.getFirst());
+          noTimeoutManager.newCollector(Integer.MAX_VALUE, searcher.leafContexts.get(0));
 
       // Check that a normal collector is created without timeout
       assertTrue(noTimeoutCollector instanceof TopKnnCollector);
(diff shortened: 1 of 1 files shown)


lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
```diff
@@ -781,7 +781,7 @@ public void testTimeLimitingKnnCollectorManager() throws IOException {
       TimeLimitingKnnCollectorManager noTimeoutManager =
           new TimeLimitingKnnCollectorManager(delegate, null);
       KnnCollector noTimeoutCollector =
-          noTimeoutManager.newCollector(Integer.MAX_VALUE, searcher.leafContexts.getFirst());
+          noTimeoutManager.newCollector(Integer.MAX_VALUE, searcher.leafContexts.get(0));
```
benwtrent, line 784: This doesn't really make sense to me :/. Main requires JDK21 and using all the nice things 21 provides is, well, nice. 

Its pretty common for backports to have one or more weird issues due to compilation levels. The biggest one I have seen is `if (klass instanceof ClassFoo classFoo)` or the new `switch` statement syntax.

vigyasharma: What is our preferred approach for handling such compilation issues during backport? I was worried about recurring conflicts if we just change it in branch_9x, is that not the case? Skipping backport for something like this didn't seem right either. I'd like to learn what we consider as good practice for Lucene codebase.

benwtrent: > I was worried about recurring conflicts if we just change it in branch_9x, is that not the case? 

I wouldn't expect the conflicts to be to horrible. The biggest ones I have ran into is the new `switch` statement usages. 

> Skipping backport for something like this didn't seem right either.

No, backporting the test fix is good. Allows them to be unmuted & ensures we are actually testing these paths.

>  I'd like to learn what we consider as good practice for Lucene codebase.

There is no prevalent pattern. Its a "gut" thing, and I would rather there not be any mandate. 

Its fine that you did this, just wanted to indicate that in this particular case, seemed like it wasn't necessary.

PR #13312, opened 2024-04-17
Fix test failure TestKnnByteVectorQuery#testTimeLimitingKnnCollectorManager

easyice:
`TopKnnCollectorManager#newCollector` will return `MultiLeafKnnCollector` instance if the index has multiple segments

```
org.apache.lucene.search.TestKnnByteVectorQuery > testTimeLimitingKnnCollectorManager FAILED
    java.lang.AssertionError
        at __randomizedtesting.SeedInfo.seed([A0DCCB76BFBF940C:8E1E7418F5C62B46]:0)
        at org.junit.Assert.fail(Assert.java:87)
        at org.junit.Assert.assertTrue(Assert.java:53)
        at org.apache.lucene.search.BaseKnnVectorQueryTestCase.testTimeLimitingKnnCollectorManager(BaseKnnVectorQueryTestCase.java:787)
        at org.apache.lucene.search.TestKnnByteVectorQuery.testTimeLimitingKnnCollectorManager(TestKnnByteVectorQuery.java:30)
```

```
gradlew test --tests TestKnnByteVectorQuery.testTimeLimitingKnnCollectorManager -Dtests.seed=A0DCCB76BFBF940C -Dtests.nightly=true
```

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/TimeLimitingKnnCollectorManager.java
+++ b/lucene/core/src/java/org/apache/lucene/search/TimeLimitingKnnCollectorManager.java
@@ -42,13 +42,13 @@ public class TimeLimitingKnnCollectorManager implements KnnCollectorManager {
     if (queryTimeout == null) {
       return collector;
     }
-    return new TimeoutingKnnCollector(collector);
+    return new TimeLimitingKnnCollector(collector);
   }
 
-  class TimeoutingKnnCollector implements KnnCollector {
+  class TimeLimitingKnnCollector implements KnnCollector {
     private final KnnCollector collector;
 
-    TimeoutingKnnCollector(KnnCollector collector) {
+    TimeLimitingKnnCollector(KnnCollector collector) {
       this.collector = collector;
     }
 
--- a/lucene/core/src/java/org/apache/lucene/search/TimeLimitingKnnCollectorManager.java
+++ b/lucene/core/src/java/org/apache/lucene/search/TimeLimitingKnnCollectorManager.java
@@ -42,54 +42,62 @@ public class TimeLimitingKnnCollectorManager implements KnnCollectorManager {
     if (queryTimeout == null) {
       return collector;
     }
-    return new KnnCollector() {
-      @Override
-      public boolean earlyTerminated() {
-        return queryTimeout.shouldExit() || collector.earlyTerminated();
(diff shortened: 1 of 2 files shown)


lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
```diff
@@ -784,7 +785,9 @@ public void testTimeLimitingKnnCollectorManager() throws IOException {
           noTimeoutManager.newCollector(Integer.MAX_VALUE, searcher.leafContexts.getFirst());
 
       // Check that a normal collector is created without timeout
-      assertTrue(noTimeoutCollector instanceof TopKnnCollector);
+      assertTrue(
```
dweiss, line 788: Thanks for trying to fix this. I think this type of test is problematic in that it restricts the implementations that are, in theory, an open set... I would change the implementation returned inside TimeLimitingKnnCollectorManager to a package-private class, then check for that using assertFalse(noTimeoutCollector instanceof TimeLimitingKnnCollectorManager.TimeoutingKnnCollector)?

easyice: Thanks for you suggestions! you are right, I like the idea!

dweiss (approved):

kaivalnp (approved):
Thanks for fixing this!

lucene/core/src/java/org/apache/lucene/search/TimeLimitingKnnCollectorManager.java
```diff
@@ -42,54 +42,62 @@ public KnnCollector newCollector(int visitedLimit, LeafReaderContext context) th
...
-                ? TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO
-                : docs.totalHits.relation;
-
-        return new TopDocs(new TotalHits(docs.totalHits.value, relation), docs.scoreDocs);
-      }
-    };
+    return new TimeoutingKnnCollector(collector);
+  }
+
+  class TimeoutingKnnCollector implements KnnCollector {
```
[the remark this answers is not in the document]
easyice: Thank you @kaivalnp , it looks better, fixed.

PR #13315, opened 2024-04-17
Fix IndexOutOfBoundsException thrown in DefaultPassageFormatter by unordered matches

### LINKED ISSUE #12431: UnifiedHighlighter: DefaultPassageFormatter causes IndexOutOfBoundsException w/ setStoreTermVectorOffsets unless setStoreTermVectorPositions
### Description

Summary of mailing list thread...

https://lists.apache.org/list.html?java-user@lucene.apache.org

* Using `UnifiedHighlighter` w/ `DefaultPassageFormatter`
* Highlighting fields that use `setStoreTermVectors(true)` + `setStoreTermVectorOffsets(true)`
  * but do *NOT* use `setStoreTermVectorPositions(true)`
* `IndexOutOfBoundsException` can occur in `DefaultPassageFormatter` -> `StringBuilder` when query and field includes multiple terms
* Problem caused by `TermVectorOffsetStrategy` producing `Passage` instances where matches are not in order (by start offset)
  * Not clear from `Passage` API if this is allowed
  * `DefaultPassageFormatter` does not expect this (only the possibility that end/start will overlap)
* Problem started happening "by default" in 9.0 due to LUCENE-9431
  * #362
* Known workarounds:
  * Index Time Option: Add `setStoreTermVectorPositions(true)` to fields you wish to highlight that already use `setStoreTermVectors(true)`
  * Alternative Query Time Option: Subclass `UnifiedHighlighter` to override `getFlags(String)` and remove `HighlightFlag.WEIGHT_MATCHES` from the set returned by `super.getFlags(field)`

Test patch demonstrating problem in above linked mailing list thread

_No response_

scampi:
- **test: add unit tests to reproduce the IndexOutOfBoundsException in DefaultPassageFormatter**
- **doc: clarify javadoc of MatchesIterator**
- **fix: sort passages by offset if positions are missing**

Fix #12431

### Description

`DefaultPassageFormatter` may cause an `IndexOutOfBoundsException` in case matches of a passage are out of order.
This happens when term vectors are stored but not the positions.

The tentative solution consists in ordering matches according to offsets in `DisjunctionMatchesIterator` in case positions don't exist.
However, I wonder if it is correct to create such an iterator when positions are not available. Matches with no terms are explicitly removed, and the javadoc of `MATCH_WITH_NO_TERMS` mentions that it indicates a match with _no term positions_. Therefore, from that doc it shouldn't be possible to highlight text without positions. Is that javadoc correct ? Should it be updated ?

https://github.com/apache/lucene/blob/bc678ac67e32c55a27a4e8950c25144cc89cef66/lucene/core/src/java/org/apache/lucene/search/MatchesUtils.java#L67

https://github.com/apache/lucene/blob/bc678ac67e32c55a27a4e8950c25144cc89cef66/lucene/core/src/java/org/apache/lucene/search/MatchesUtils.java#L40-L44

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/DisjunctionMatchesIterator.java
+++ b/lucene/core/src/java/org/apache/lucene/search/DisjunctionMatchesIterator.java
@@ -194,6 +194,15 @@ final class DisjunctionMatchesIterator implements MatchesIterator {
         new PriorityQueue<MatchesIterator>(matches.size()) {
           @Override
           protected boolean lessThan(MatchesIterator a, MatchesIterator b) {
+            if (a.startPosition() == -1 && b.startPosition() == -1) {
+              try {
+                return a.startOffset() < b.startOffset()
+                    || (a.startOffset() == b.startOffset() && a.endOffset() < b.endOffset())
+                    || (a.startOffset() == b.startOffset() && a.endOffset() == b.endOffset());
+              } catch (IOException e) {
+                throw new IllegalArgumentException("Failed to retrieve term offset", e);
+              }
+            }
             return a.startPosition() < b.startPosition()
                 || (a.startPosition() == b.startPosition() && a.endPosition() < b.endPosition())
                 || (a.startPosition() == b.startPosition() && a.endPosition() == b.endPosition());
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -173,6 +173,9 @@ Bug Fixes
 * GITHUB#12878: Fix the declared Exceptions of Expression#evaluate() to match those
   of DoubleValues#doubleValue(). (Uwe Schindler)
 
+* GITHUB#12431: Fix IndexOutOfBoundsException thrown in DefaultPassageFormatter
+  by unordered matches. (Stephane Campinas)
+
 Changes in Runtime Behavior
 ---------------------
 
--- a/lucene/core/src/java/org/apache/lucene/search/MatchesIterator.java
+++ b/lucene/core/src/java/org/apache/lucene/search/MatchesIterator.java
@@ -45,14 +45,14 @@ public interface MatchesIterator {
   boolean next() throws IOException;
 
   /**
-   * The start position of the current match
+   * The start position of the current match, or {@code -1} if positions are not available
    *
    * <p>Should only be called after {@link #next()} has returned {@code true}
    */
   int startPosition();
 
   /**
-   * The end position of the current match
+   * The end position of the current match, or {@code -1} if positions are not available
    *
    * <p>Should only be called after {@link #next()} has returned {@code true}
    */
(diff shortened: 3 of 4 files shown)


lucene/core/src/java/org/apache/lucene/search/DisjunctionMatchesIterator.java
```diff
@@ -194,6 +194,15 @@ private DisjunctionMatchesIterator(List<MatchesIterator> matches) throws IOExcep
...
           protected boolean lessThan(MatchesIterator a, MatchesIterator b) {
+            if (a.startPosition() == -1 && b.startPosition() == -1) {
+              try {
+                return a.startOffset() < b.startOffset()
+                    || (a.startOffset() == b.startOffset() && a.endOffset() < b.endOffset())
+                    || (a.startOffset() == b.startOffset() && a.endOffset() == b.endOffset());
+              } catch (IOException e) {
+                throw new IllegalArgumentException("Failed to retrieve term offset", e);
+              }
+            }
```
mikemccand, line 205: It's weird that `start/endOffset` can throw `IOException` but `start/endPosition` do not :)  Nothing to fix in this PR, just noting the inconsistency ...

mikemccand:
What a fun and tricky corner case -- thank you @scampi for uncovering this, showing the bug with the added unit tests, and the tentative fix.

I think it is actually technically allowed to index term vectors with offsets only (no positions) and then hilite from there: Lucene is recording the character offsets of every occurence of each unique term, per document, and that's all that is needed to hilite.  So I like your fix -- I'll merge it soon unless anyone objects.

Could you maybe add a `lucene/CHANGES.txt` entry?

I think the intent of `MATCH_WITH_NO_TERMS` is different: that is for queries against Lucene index structures that fundamenally do not track positional information (doc values, points).  With term vectors, the positional information indeed exists, it's just that the user may choose to record only offsets.

lucene/core/src/java/org/apache/lucene/search/DisjunctionMatchesIterator.java
romseygeek: It's because the work of loading positions is always done in `next()`, so there's no IO involved when calling `startPosition` or `endPosition`.  But we only load offsets when specifically asked, so IO may be involved when calling those methods.  It does look a bit funky though, I agree!

romseygeek (approved):

scampi:
@mikemccand @romseygeek thanks for the review!

> What a fun and tricky corner case -- thank you @scampi for uncovering this, showing the bug with the added unit tests, and the tentative fix.

I actually just contributed the fix, @hossman found the issue and posted the test.

> Could you maybe add a lucene/CHANGES.txt entry?

I have added a commit for this.

dweiss (approved):

scampi:
@dweiss thanks!

Issue #13180, opened 2024-03-13
Make Weight#scorerSupplier abstract, Weight#scorer final


jpountz:
### Description

A long time ago, we introduced `Weight#scorerSupplier` to enable [query planning for range queries](https://www.elastic.co/blog/better-query-planning-for-range-queries-in-elasticsearch), using either points or doc values to run a range query depending on which on is expected to perform best.

`Weight#scorerSupplier` was initially introduced with a default implementation that would delegace to `Weight#scorer` and we fixed more and more queries to implement `Weight#scorerSupplier` over time. Let's revert this situation now and make `Weight#scorerSupplier` abstract, effectively forcing all queries to implement it properly. `Weight#scorer` can stay around since it's quite convenient in the case when you just want to pull an iterator, but let's make it final and delegate to `scorerSupplier`?

jpountz:
Addressed via #13319.

PR #13319, opened 2024-04-20
Make Weight#scorerSupplier abstract, Weight#scorer final

iamsanjay:
### Description

#13180 

This PR would declare the scoreSupplier method abstract and marked scorer final. The nature of the change itself is not complex, however the large number of files has been modified with this change. 

**Few Observations:**

1. As this change would enforce subclasses of Weight to implement scorerSupplier, the default cost calculation method is used where it was missing

```
@Override
public long cost() {
    return scorer.iterator().cost();
}
```
2. scorer method is removed from SpanQuery.java and therefore all the classes has to implement the scorerSupplier, We can also move the socrerSupplier method to SpanQuery that way the subclasses does not has to provide the implementation, which is same for all subclasses.
3. scorer method removed from AssertingWeight.
4. What would be the implementation for cost where calling scorer throws UnsupportedOpeartionException. Does cost method also throws the same exception, currently it's returning 0. Below are the classes where you will find such code.

```
lucene/core/src/test/org/apache/lucene/search/TestBooleanScorer.java is also such example.
lucene/core/src/test/org/apache/lucene/search/TestConjunctionDISI.java
lucene/core/src/test/org/apache/lucene/search/TestDisiPriorityQueue.java
lucene/core/src/test/org/apache/lucene/search/TestDisjunctionScoreBlockBoundaryPropagator.java
lucene/core/src/test/org/apache/lucene/search/TestMaxScoreBulkScorer.java
lucene/facet/src/java/org/apache/lucene/facet/DrillSidewaysQuery.java
lucene/suggest/src/java/org/apache/lucene/search/suggest/document/CompletionWeight.java

```


iamsanjay:
scorerSupplier has to be null if scorer is null. Therefore scorer was initialized inside the scorerSupplier, instead of scorerSupplier.get method. And if scorer evaluates to null then scorerSupplier would return null.

lucene/core/src/java/org/apache/lucene/search/FilterWeight.java
```diff
@@ -58,11 +58,6 @@ public Explanation explain(LeafReaderContext context, int doc) throws IOExceptio
     return in.explain(context, doc);
   }
 
-  @Override
-  public Scorer scorer(LeafReaderContext context) throws IOException {
-    return in.scorer(context);
-  }
```
jpountz, line 64: This class should delegate scorerSupplier now?

lucene/core/src/java/org/apache/lucene/document/BinaryRangeFieldRangeQuery.java
```diff
@@ -136,7 +136,18 @@ public float matchCost() {
               }
             };
 
-        return new ConstantScoreScorer(this, score(), scoreMode, iterator);
+        final var scorer = new ConstantScoreScorer(this, score(), scoreMode, iterator);
+        return new ScorerSupplier() {
```
jpountz, line 140: Since we seem to be doing this in lots of places, we could introduce a `DefaultScorerSupplier` class that takes a `Scorer` as a ctor argument, alongside `DefaultBulkScorer`.

lucene/core/src/java/org/apache/lucene/document/SortedNumericDocValuesRangeQuery.java
```diff
@@ -101,7 +102,8 @@ public boolean isCacheable(LeafReaderContext ctx) {
       }
 
       @Override
-      public Scorer scorer(LeafReaderContext context) throws IOException {
+      public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOException {
+        final Weight weight = this;
```
jpountz, line 106: Extracting this local variable doesn't seem to help? (maybe it did in a previous version of the change?)

lucene/core/src/java/org/apache/lucene/search/Weight.java
```diff
@@ -122,7 +122,13 @@ public final Query getQuery() {
    * @return a {@link Scorer} which scores documents in/out-of order.
    * @throws IOException if there is a low-level I/O error
    */
-  public abstract Scorer scorer(LeafReaderContext context) throws IOException;
+  public final Scorer scorer(LeafReaderContext context) throws IOException {
```
jpountz, line 125: Can you update javadocs to mention it's just a helper method that delegates to scorerSupplier?

lucene/core/src/java/org/apache/lucene/search/Weight.java
```diff
@@ -131,23 +137,7 @@ public final Query getQuery() {
    *
    * @see #scorer
    */
-  public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOException {
```
jpountz, line 134: Can you update javadocs? E.g. it's not an optional method anymore

lucene/core/src/test/org/apache/lucene/search/TestBoolean2ScorerSupplier.java
```diff
@@ -42,8 +42,18 @@ public Explanation explain(LeafReaderContext context, int doc) throws IOExceptio
...
+        public Scorer get(long leadCost) throws IOException {
+          return null;
+        }
+
+        @Override
+        public long cost() {
+          return 0;
+        }
+      };
     }
```
jpountz, line 57: Let's return a null `ScorerSupplier`? Returning null from `ScorerSupplier#get` is not legal.

lucene/core/src/test/org/apache/lucene/search/TestConjunctionDISI.java
```diff
@@ -98,8 +98,18 @@ public Explanation explain(LeafReaderContext context, int doc) throws IOExceptio
...
+        public Scorer get(long leadCost) throws IOException {
+          return null;
+        }
+
+        @Override
+        public long cost() {
+          return 0;
+        }
+      };
     }
```
jpountz, line 113: Let's return a null `ScorerSupplier`? Returning null from `ScorerSupplier#get` is not legal.

lucene/core/src/test/org/apache/lucene/search/TestDisjunctionScoreBlockBoundaryPropagator.java
```diff
@@ -37,8 +37,18 @@ public Explanation explain(LeafReaderContext context, int doc) throws IOExceptio
...
+        public Scorer get(long leadCost) throws IOException {
+          return null;
+        }
+
+        @Override
+        public long cost() {
+          return 0;
+        }
+      };
     }
```
jpountz, line 52: Let's return a null `ScorerSupplier`?

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -1355,9 +1355,23 @@ protected WeightWrapper(Weight in, AtomicBoolean scorerCalled, AtomicBoolean bul
     }
 
     @Override
-    public Scorer scorer(LeafReaderContext context) throws IOException {
-      scorerCalled.set(true);
-      return in.scorer(context);
+    public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOException {
+      final var scorer = in.scorer(context);
```
jpountz, line 1359: let's delegate to `scorerSupplier()` not `scorer()`?

lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -1709,12 +1719,22 @@ public int hashCode() {
...
+            @Override
+            public Scorer get(long leadCost) throws IOException {
+              scorerCreatedCount.incrementAndGet();
+              return new ConstantScoreScorer(
+                  weight, 1, scoreMode, DocIdSetIterator.all(context.reader().maxDoc()));
+            }
+
+            @Override
+            public long cost() {
+              return 0;
```
jpountz, line 1735: return `context.reader().maxDoc()`?

lucene/core/src/test/org/apache/lucene/search/TestNeedsScores.java
```diff
@@ -131,9 +131,23 @@ public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float bo
       final Weight w = in.createWeight(searcher, scoreMode, boost);
       return new FilterWeight(w) {
         @Override
-        public Scorer scorer(LeafReaderContext context) throws IOException {
-          assertEquals("query=" + in, value, scoreMode);
-          return w.scorer(context);
+        public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOException {
+          final var scorer = w.scorer(context);
```
jpountz, line 135: Can you delegate to `w.scorerSupplier` here as well?

lucene/queries/src/java/org/apache/lucene/queries/spans/SpanWeight.java
```diff
@@ -135,16 +135,6 @@ private Similarity.SimScorer buildSimWeight(
...
-  @Override
-  public SpanScorer scorer(LeafReaderContext context) throws IOException {
-    final Spans spans = getSpans(context, Postings.POSITIONS);
-    if (spans == null) {
-      return null;
-    }
-    final LeafSimScorer docScorer = getSimScorer(context);
-    return new SpanScorer(this, spans, docScorer);
-  }
-
```
jpountz, line 147: I wonder if we need to introduce a `SpanScorerSupplier` class whose `get()` method returns a `SpanScorer` and make `SpanWeight#scorerSupplier` return this new type.

jpountz (commented):
Thanks for looking into this large refactoring! I left some comments but the general direction looks good to me.

To your question about how to handle `cost()` when `get()` throws an exception, I would throw an exception in `cost()` too.

lucene/core/src/java/org/apache/lucene/document/BinaryRangeFieldRangeQuery.java
iamsanjay: +1

lucene/core/src/java/org/apache/lucene/search/FilterWeight.java
iamsanjay: To do that, I had to modified one test case which was failing on overriding the scorerSupplier for `FilterWeight.java`. 
https://github.com/apache/lucene/blob/e6dac3400d7b1118318ff342deea3f2cbc11bb7a/lucene/core/src/test/org/apache/lucene/search/TestFilterWeight.java#L33-L41

**The ticket that introduced that change:**

https://issues.apache.org/jira/browse/LUCENE-7055

https://github.com/apache/lucene/commit/86233cb95de6f24aa2ae7fd016b7d75d535024c7

iamsanjay:
Instead of both get() and cost() throwing the exception, Can make scorerSupplier throw `UnsupportedOperationException()`? 

```
@Override
public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOException {
}
```

lucene/queries/src/java/org/apache/lucene/queries/spans/SpanWeight.java
iamsanjay: Well If we move the implementation of scorerSupplier to SpanWeight then all the subclasses can use the same scorerSupplier and then we can avoid making SpanScorerSupplier, If redundancy was your concern.

Second, after moving the implementation there, we can return the SpanScorer from the `get()` method. However, we still may need to type cast because usually we pass reference to Weight, not to the concrete class.

https://github.com/apache/lucene/blob/e40e1086f5aba195726aa0bc68afc8226e9268db/lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/PhraseHelper.java#L263

Also I had to type cast in SpanWeight as well, because now SpanWeight calling `Weight#scorer()` which return Scorer.

jpountz:
> Instead of both get() and cost() throwing the exception, Can make scorerSupplier throw UnsupportedOperationException()?

In general yes, there may be a few exceptions though like `JustCompileSearch` where we want to keep these methods since it's about detecting API changes.

iamsanjay:
> In general yes, there may be a few exceptions though like `JustCompileSearch` where we want to keep these methods since it's about detecting API changes.

Okay in that case, let's make cost method throw an exception.

iamsanjay:
I also changed scorerSupplier in few classes to delegate it to scorerSupplier instead of scorer.

PR #13322, opened 2024-04-23
Implement Weight#count for vector values in the FieldExistsQuery

bugmakerrrrrr:
### Description

Considering only one vector value can be indexed per doc, we can use the Float/ByteVectorValues#size method to implement Weight#count for vector values in the FieldExistsQuery

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -162,8 +162,6 @@ Optimizations
 
 * GITHUB#12552: Make FSTPostingsFormat load FSTs off-heap. (Tony X)
 
-* GITHUB#13322: Implement Weight#count for vector values in the FieldExistsQuery. (Pan Guixin)
-
 Bug Fixes
 ---------------------
 
@@ -404,6 +402,8 @@ Optimizations
 * GITHUB#13400: Replace Set<Integer> by IntHashSet and Set<Long> by LongHashSet. (Bruno Roustant)
 
 * GITHUB#13406: Replace List<Integer> by IntArrayList and List<Long> by LongArrayList. (Bruno Roustant)
+
+* GITHUB#13322: Implement Weight#count for vector values in the FieldExistsQuery. (Pan Guixin)
 
 Bug Fixes
 ---------------------
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -162,6 +162,8 @@ Optimizations
 
 * GITHUB#12552: Make FSTPostingsFormat load FSTs off-heap. (Tony X)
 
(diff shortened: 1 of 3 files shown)


benwtrent:
I like this. Could you add an entry under "Optimizations" in "CHANGES.txt"?

benwtrent (approved):

jpountz (commented):
I don't remember how vectors handle ghost fields, could this trigger a NPE if a field indexes vectors, then all docs that have vectors get merged away?

This looks like something that could go into 9.11, let's move the CHANGES entry under 9.11?

jpountz (commented):
I'm not comfortable with using `DocIdSetIterator#cost` since it has no accuracy guarantee, could you use `FloatVectorValues#size()` instead?

bugmakerrrrrr:
@jpountz could you please take a look when you get a chance

lucene/CHANGES.txt
```diff
@@ -399,6 +399,8 @@ Optimizations
 
 * GITHUB#13406: Replace List<Integer> by IntArrayList and List<Long> by LongArrayList. (Bruno Roustant)
 
+* GITHUB#13322: Implement Weight#count for vector values in the FieldExistsQuery. (Pan Guixin)
+
```
benwtrent, line 403: Since 9.11 has been cut, this needs moved to 9.12

bugmakerrrrrr: done

benwtrent:
@bugmakerrrrrr Once you handle the conflicts, I can merge and backport. I think this is ready to go unless @jpountz has any outstanding concerns.

lucene/core/src/java/org/apache/lucene/search/FieldExistsQuery.java
```diff
@@ -277,4 +276,20 @@ private String buildErrorMsg(FieldInfo fieldInfo) {
...
         + "' exists and indexes neither of these data structures";
   }
+
+  private int getVectorValuesSize(FieldInfo fi, LeafReader reader) throws IOException {
+    assert fi.name.equals(field);
+    return switch (fi.getVectorEncoding()) {
+      case FLOAT32 -> {
+        FloatVectorValues floatVectorValues = reader.getFloatVectorValues(field);
+        assert floatVectorValues != null : "unexpected null float vector values";
+        yield floatVectorValues == null ? 0 : floatVectorValues.size();
```
jpountz, line 286: Nit: can you remove the `null` check since vector values should never be null? FWIW I added a test case for this here: #13455.

jpountz (approved):

Issue #13234, opened 2024-03-28
Query parser fails to parse a range query string when there are escaped brackets inside the range


marko-bekhta:
### Description

Assume there's a query parser created, e.g.:

```java
Analyzer analyzer = new ClassicAnalyzer();
QueryParser queryParser = new QueryParser( "field", analyzer );
```

trying to parse simple ranges like:

```java
// simple range query with no escapes works fine:
Query query = queryParser.parse( "[ 1 TO 10 ]" );
```
works as expected and a `TermRangeQuery` is created with no exceptions.

But if the range in the string contains some escaped brackets -- it leads to a parsing exception. 
Let's assume one would want to extend the query parser to work with date-time fields and would want to parse something like:

```java
// another range query but now it has some escaping between the range brackets:
query = queryParser.parse( "[ 2024\\-01\\-01T01\\:01\\:01\\+01\\:00\\[Europe\\/Warsaw\\] TO 2025\\-01\\-01T01\\:01\\:01\\+01\\:00\\[Europe\\/Warsaw\\] ]" );
```

where all special characters are escaped, leads to:
```java
Caused by: org.apache.lucene.queryparser.classic.ParseException: Encountered " "]" "] "" at line 1, column 50.
Was expecting:
    "TO" ...
    
	at org.apache.lucene.queryparser.classic.QueryParser.generateParseException(QueryParser.java:1004)
	at org.apache.lucene.queryparser.classic.QueryParser.jj_consume_token(QueryParser.java:867)
	at org.apache.lucene.queryparser.classic.QueryParser.Term(QueryParser.java:532)
	at org.apache.lucene.queryparser.classic.QueryParser.Clause(QueryParser.java:366)
	at org.apache.lucene.queryparser.classic.QueryParser.Query(QueryParser.java:251)
	at org.apache.lucene.queryparser.classic.QueryParser.TopLevelQuery(QueryParser.java:223)
	at org.apache.lucene.queryparser.classic.QueryParserBase.parse(QueryParserBase.java:137)
	... 4 more
```

Note, the idea to do range queries for dates  is to have something along the lines:
```java
QueryParser queryParser = new QueryParser( "field", analyzer ) {
	@Override
	protected Query newRangeQuery(String field, String part1, String part2, boolean startInclusive, boolean endInclusive) {
		var p1 = parseValue(part1);
		var p2 = parseValue(part2);

		return createRangeQueryForDates( field, p1, p2, startInclusive, endInclusive );
	}
};
```
[...]

PR #13323, opened 2024-04-25
Add test for parsing brackets in range queries

benchaplin:
### Description

I've added some unit tests to protect/display the QueryParser's ability to parse brackets in range query terms when in quotes.

[This issue](https://github.com/apache/lucene/issues/13234) raised a question about the QueryParser's ability to handle brackets in a range query's terms, even when escaped, for example:

```java
queryParser.parse( "[ a\\[i\\] TO b\\[i\\] ]" );

/* 
org.apache.lucene.queryparser.classic.ParseException: Cannot parse '[a\[i\] TO b\[i\]]': Encountered " "]" "] "" at line 1, column 6.
Was expecting:
    "TO" ...
 */
```

You can get around this exception by wrapping each range query term in quotes:
```java
queryParser.parse( "[\"a[i]\" TO \"b[i]\"]" );
```

This is because of [this JavaCC grammar rule](https://github.com/apache/lucene/blob/9af3ef8952597ead37a5a9ce5373ed3bbd40238e/lucene/queryparser/src/java/org/apache/lucene/queryparser/classic/QueryParser.jj#L345). `<RANGE_GOOP>` cannot handle a closing bracket, even an escaped one, while `<RANGE_QUOTED>` can.

I've added some unit tests to protect this functionality.

---

Any thoughts on whether `queryParser.parse( "[ a\\[i\\] TO b\\[i\\] ]" );` should work? I did play around with updating the grammar, it should be possible but I need to think more about how it might break backwards compatibility.


dweiss:
> Any thoughts on whether queryParser.parse( "[ a\\[i\\] TO b\\[i\\] ]" ); should work? I did play around with updating the grammar, it should be possible but I need to think more about how it might break backwards compatibility.

Looking at the javacc file it seems range terms are parsed differently compared to normal tokens. This is strange but I bet there was a reason for it - now long forgotten. I think, logically, it should accept the same term notation that is used elsewhere. It will break backward compatibility but maybe consistency is worth this cost (for the major release)?

benchaplin:
I was thinking a fix like this could work:

```java
| <RANGE_GOOP:   (~[ " ", "]", "}" ] | "\\ " | "\\]" | "\\}" )+ >
```

simply allowing the range term to continue parsing through an escaped space or closing bracket.

"Breaking" changes:
- This would permit some queries that previously threw exceptions (like the one described in the original issue)
- I'm trying to think of a query that was permitted and now would break (all existing unit tests pass). My first thought was a range query ending with an escaped bracket, however those are disallowed by the `Term can not end with escape character` rule. Any thoughts?

dweiss:
The joys of escaping... Never easy (hello, Windows command-line users...). You may want to add a randomized test that constructs those terms using a mix of characters and "allowed" escape sequences - a corner case will pop up quickly, I think.

benchaplin:
Thanks for the advice @dweiss, I will play around with some test cases for this fix. In the meantime, I think we can merge this and I will open another PR with the grammar changes if I can get something working.

dweiss:
Apologies for the delay, @benchaplin - I've merged this test into the main branch.

PR #13324, opened 2024-04-28
Fix numDeletesToMerge for unchanged segments

dnhatn:
Currently, we do not trigger numDeletesToMerge for unchanged segments that have not yet entered the reader pool. Consequently, there are cases where force merges overlook segments with soft-deletes, as the merge policy only takes into account hard deletes when determining merge specifications.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -6129,16 +6129,14 @@ public class IndexWriter
     validate(info);
     MergePolicy mergePolicy = config.getMergePolicy();
     final ReadersAndUpdates rld = getPooledInstance(info, true);
-    int numDeletesToMerge;
-    if (rld != null) {
-      numDeletesToMerge = rld.numDeletesToMerge(mergePolicy);
-    } else {
-      // if we don't have a  pooled instance lets just return the hard deletes, this is safe!
-      numDeletesToMerge = info.getDelCount();
+    try {
+      final int numDeletesToMerge = rld.numDeletesToMerge(mergePolicy);
+      assert numDeletesToMerge <= info.info.maxDoc()
+          : "numDeletesToMerge: " + numDeletesToMerge + " > maxDoc: " + info.info.maxDoc();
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -6128,7 +6128,7 @@ public final int numDeletesToMerge(SegmentCommitInfo info) throws IOException {
     ensureOpen(false);
     validate(info);
     MergePolicy mergePolicy = config.getMergePolicy();
-    final ReadersAndUpdates rld = getPooledInstance(info, false);
+    final ReadersAndUpdates rld = getPooledInstance(info, true);
```
jpountz, line 6131: If we pass `create=true` then we don't need to handle the case when `rld` is `null` below? There is a comment there saying that returning the number of hard deletes is safe, which confuses me since you're suggesting it's actually a problem? Maybe @s1monw can help shed a light?

easyice, line 6131: In java doc for `ReaderPool#get`, if `create=true`, the `rld` should be released by `IndexWriter#release` or `ReaderPool#release`

dnhatn: Good catch, thanks @easyice. I pushed https://github.com/apache/lucene/pull/13324/commits/7e970128b573ba91edd048478546db5cdb9923ac

s1monw: I just re-read the change that I made that added this comment and I think it's not allowed to pull a reader here if it's not already visible. I don't recall the side-effects of this but I am sure there are some.  To me the test seems to be artificial since it's really just testing if that works on fresh writer?! I am not against the change I just think we need more coverage for this to ensure it's not triggering the same issues that caused the change in the first place?
I think IW should be very conservative here with pulling memory for segments that are not in memory yet. 

Yet, the comment basically means that we can't return the soft del count since it might overcommit because of it's nature of being dynamic based on the MP.

PR #13327, opened 2024-04-29
Reduce memory usage of field maps in FieldInfos and BlockTree TermsReader.

bruno-roustant:
### Description

Two goals:

1- Reduce the memory usage of field maps when there are many fields.
FieldInfos construtor is refactored to build the byNumber array in a more efficient way, avoiding array growing and copies.
Use a primitive IntObjectHashMap to reduce the memory usage compared to an HashMap in Lucene90BlockTreeTermsReader.

2- Add new IntObjectHashMap in the existing small hppc fork. Leverage this PR to show an example use-case. It hopefully can be reused later for other use-cases.


lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Lucene90BlockTreeTermsReader.java
```diff
@@ -113,7 +114,8 @@ public final class Lucene90BlockTreeTermsReader extends FieldsProducer {
   // produce DocsEnum on demand
   final PostingsReaderBase postingsReader;
 
-  private final Map<String, FieldReader> fieldMap;
+  private final FieldInfos fieldInfos;
+  private final IntObjectHashMap<FieldReader> fieldMap;
```
bruno-roustant, line 118: This PR proposes to leverage the existing field-name -> FieldInfo map in FieldInfos to not repeat the ref to the field name strings here. Instead use the field number (specific to the FieldInfos) as key, so that we can use a compact primitive map.

Then below, in terms(String fieldName), we can use FieldInfos.fieldInfo(String fieldName) as a first mapping to the field number, and then use this compact map to get the Terms.

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
```diff
@@ -81,30 +81,13 @@ public FieldInfos(FieldInfo[] infos) {
...
 
-    int size = 0; // number of elements in byNumberTemp, number of used array slots
-    FieldInfo[] byNumberTemp = new FieldInfo[10]; // initial array capacity of 10
+    byName = new HashMap<>((int) (infos.length / 0.75f) + 1);
     for (FieldInfo info : infos) {
       if (info.number < 0) {
         throw new IllegalArgumentException(
             "illegal field number: " + info.number + " for field " + info.name);
       }
-      size = info.number >= size ? info.number + 1 : size;
```
bruno-roustant, line 91: Instead of always building an array of FieldInfo of the size of the max field number + 1, we have some new code below to create either an ArrayFieldInfoByNumber or a MapFieldInfoByNumber. This way we don't create a sparse array if the max field number is high compared to the actual number of fields in this FieldInfos.

jpountz (commented):
FYI FieldInfos used to switch between sparse and dense in the past, and we changed it to always use a dense encoding 6 years ago: https://issues.apache.org/jira/browse/LUCENE-8033.

It doesn't necessarily mean we cannot merge your change, at least the sparse encoding is used more often than it was in the past. I'm curious if you know why the indexes you're seeing end up having sparse fields?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Lucene90BlockTreeTermsReader.java
jpountz: FWIW this is what Lucene90PointsReader does today as well.

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
```diff
@@ -81,30 +81,13 @@ public FieldInfos(FieldInfo[] infos) {
     String softDeletesField = null;
     String parentField = null;
 
-    int size = 0; // number of elements in byNumberTemp, number of used array slots
-    FieldInfo[] byNumberTemp = new FieldInfo[10]; // initial array capacity of 10
+    byName = new HashMap<>((int) (infos.length / 0.75f) + 1);
```
dsmiley, line 84: see `org.apache.lucene.util.CollectionUtil#newHashMap` instead

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
```diff
@@ -156,15 +139,17 @@ public FieldInfos(FieldInfo[] infos) {
...
+    Arrays.sort(sortedFieldInfos, (fi1, fi2) -> Integer.compare(fi1.number, fi2.number));
+    int maxFieldNumber = infos.length == 0 ? -1 : sortedFieldInfos[infos.length - 1].number;
+    // If there are many fields and the max field number is greater than twice the number
+    // of fields, then a map structure is more compact to store the by-number mapping.
+    byNumber =
+        maxFieldNumber >= 2 * infos.length && maxFieldNumber >= 32
+            ? new MapFieldInfoByNumber(infos)
+            : new ArrayFieldInfoByNumber(infos, maxFieldNumber);
+    // The iteration of FieldInfo is ordered by ascending field number.
+    values = Collections.unmodifiableCollection(Arrays.asList(sortedFieldInfos));
```
dsmiley, line 152: nowadays, do `List.of(sortedFieldInfos)` rather than this double-layer referring to two other classes

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
```diff
@@ -821,4 +803,64 @@ FieldInfos finish() {
       return new FieldInfos(byName.values().toArray(new FieldInfo[byName.size()]));
     }
   }
+
+  private interface FieldInfoByNumber {
```
dsmiley, line 807: or just use `IntFunction` ?

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
bruno-roustant: List.of() makes a copy as it considers the input as an "untrusted array". Here we don't copy, just wrap. Actually we could keep just Arrays.asList(sortedFieldInfos) since we own it privately, so we know we don't modify it (only iterator()).

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Lucene90BlockTreeTermsReader.java
bruno-roustant: Nice. It could benefit from the IntObjectHashMap.

bruno-roustant: Actually there are many usages of Map<Integer, Object>. I could open some PRs when memory (and perf) matters after IntObjectHashMap is in.

bruno-roustant:
So I did some benchmarking.
As I could not reproduce the sparse byNumber mapping, I prefer to remove the code for the primitive map in FieldInfos.

At the same time I could evaluate the time spent by the current FieldInfos code to build the byNumber and values arrays. When we read a segment, the reader provides the FieldInfo[] input always sorted in ascending order for number. So the code that grows byNumberTemp has to grow multiple times when there a many fields. Same for the hashmap initialized with default capacity.
This latest commit changes the way byNumber is built to only allocate one array, or even proposes to use directly the input FieldInfo[] if it is full. This constructor becomes 40% faster than before, even more when the GC is busy, as it allocates less arrays.

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
```diff
@@ -156,15 +148,38 @@ public FieldInfos(FieldInfo[] infos) {
...
-    byNumber = new FieldInfo[size];
-    for (int i = 0; i < size; i++) {
-      byNumber[i] = byNumberTemp[i];
-      if (byNumberTemp[i] != null) {
-        valuesTemp.add(byNumberTemp[i]);
+    if (fieldNumberStrictlyAscending && maxFieldNumber == infos.length - 1) {
+      // The input FieldInfo[] contains all fields numbered from 0 to infos.length - 1 and they are
+      // sorted, use it directly. This is an optimization when reading a segment with all fields
+      // since the FieldInfo[] is sorted.
+      byNumber = infos; // We could copy the input array, but do we need to?
```
bruno-roustant, line 155: Here it seems to me we can use directly the input array. Do you think we should copy it for safety?
This input array is created by the readers and passed as parameter to the constructor.

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
```diff
@@ -139,17 +148,38 @@ public FieldInfos(FieldInfo[] infos) {
...
+        }
+        byNumber[fieldInfo.number] = fieldInfo;
+      }
+      if (maxFieldNumber == infos.length - 1) {
+        // No fields are missing, use byNumber.
+        values = Arrays.asList(byNumber);
+      } else {
+        // The below code is faster than Arrays.stream(byNumber).filter(Objects::nonNull).toList(),
+        // mainly when the input FieldInfo[] is sorted, when reading a segment.
+        FieldInfo[] sortedFieldInfos = ArrayUtil.copyOfSubArray(infos, 0, infos.length);
```
dsmiley, line 178: It looks a bit curious to be calling a method with "sub array" for what is actually the whole array.  Maybe a convenience method copy(infos) should be provided.  Could be deferred of course.

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
```diff
@@ -139,17 +148,38 @@ public FieldInfos(FieldInfo[] infos) {
...
+    if (fieldNumberStrictlyAscending && maxFieldNumber == infos.length - 1) {
+      // The input FieldInfo[] contains all fields numbered from 0 to infos.length - 1 and they are
+      // sorted, use it directly. This is an optimization when reading a segment with all fields
+      // since the FieldInfo[] is sorted.
+      byNumber = infos; // We could copy the input array, but do we need to?
+      values = Arrays.asList(byNumber);
+    } else {
+      byNumber = new FieldInfo[maxFieldNumber + 1];
+      for (FieldInfo fieldInfo : infos) {
+        FieldInfo previous = byNumber[fieldInfo.number];
```
dsmiley, line 160: this is non-obvious.  At a glance, we are retrieving the very same fieldInfo, yet supposedly it's the previous.  Oh, maybe you mean "existing" fieldInfo as opposed to a fieldInfo ordered prior to this one?

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
dsmiley: This is fairly internal stuff labelled lucene.experimental; I say yes but document it in the javadoc.  Obviously look at existing callers to double-check this is fine.

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
bruno-roustant: Yes, existing, to check we have no duplicates. Let's rename it "existing" for clarity.

dsmiley: Thanks.  "previous" wasn't wrong, just ambiguous.  I like "existing"; another possible name is "old".

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
bruno-roustant: I agree. I created another [PR](https://github.com/apache/lucene/pull/13360) to add ArrayUtil.copyOf().

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
```diff
@@ -156,15 +151,38 @@ public FieldInfos(FieldInfo[] infos) {
...
+                  + fieldInfo.number);
+        }
+        byNumber[fieldInfo.number] = fieldInfo;
+      }
+      if (maxFieldNumber == infos.length - 1) {
+        // No fields are missing, use byNumber.
+        values = Arrays.asList(byNumber);
+      } else {
+        // The below code is faster than Arrays.stream(byNumber).filter(Objects::nonNull).toList(),
+        // mainly when the input FieldInfo[] is sorted, when reading a segment.
```
dsmiley, line 180: If fieldNumberStrictlyAscending, and we've given ourselves permission to accept the input array, we can merely wrap it in Arrays.asList and we're done.

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
dsmiley: Don't even need to copy it as we've given ourselves permission to take ownership of the input array; to manipulate it.

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
bruno-roustant: Here maxFieldNumber > infos.length - 1, so some fields are missing in byNumber (byNumber[fieldNumber] == null). We must not return those null FieldInfos in values.iterator().

dsmiley (approved):

PR #13328, opened 2024-04-30
Convert more classes to record classes

shubhamvishu:
### Description

- This PR addresses #13207 to convert more classes on `main` branch to record classes on main (Lucene 10 only).
- It moves a lot of data classes(120 to be precise) to use record classes that were flagged suitable in my IDE and seemed good candidate in general.
- I left a couple of them like `TotalHits` and `SynonymMap` as the PR is already very big and including those were leading to a lot more changes. Maybe we could do those as a separate PR.

&nbsp;
Raising this PR since all the tests are passing(`./gradlew test`) but `renderJavadoc` task is complaining about missing java docs on some record classes(converted in this PR) which I see has the javadocs already eg: `TermStats`, `ReaderSlice`. I'm not sure why its flagging those incorrectly or if maybe I'm missing something.


lucene/core/src/java/org/apache/lucene/codecs/TermStats.java
```diff
@@ -16,24 +16,10 @@
...
-  /** Total number of times this term occurs across all documents in the field. */
-  public final long totalTermFreq;
-
-  /** Sole constructor. */
-  public TermStats(int docFreq, long totalTermFreq) {
-    this.docFreq = docFreq;
-    this.totalTermFreq = totalTermFreq;
-  }
-}
+public record TermStats(int docFreq, long totalTermFreq) {}
```
shubhamvishu, line 25: The `renderJavaDoc` fails for this even though java docs are present. If I see earlier PR #13205 that moved some classes I don't see it failing for such classes. If this is not expected, then maybe we need to fix something in `MissingDoclet.java`?(not sure).

cc - @uschindler

uschindler: What's exactly he message?

uschindler: I could only imagine that MissingDoclet can't correctly handle inner classes!? But on the other hand it works fine with enums and it also operates recursively.

I have no time now to check all this out, so the error message and/or stack trace would be good.

uschindler: oh it is a top-level class. no idea whats wrong!

shubhamvishu: Here is what the failure message looks like :
 
```
> Task :lucene:core:renderSiteJavadoc
/home/runner/work/lucene/lucene/lucene/core/src/java/org/apache/lucene/codecs/TermStats.java:25: error: TermStats.<init> (constructor): javadocs are missing
public record TermStats(int docFreq, long totalTermFreq) {}
       ^
error: TermStats.docFreq (method): javadocs are missing
error: TermStats.totalTermFreq (method): javadocs are missing
/home/runner/work/lucene/lucene/lucene/core/src/java/org/apache/lucene/index/ReaderSlice.java:27: error: ReaderSlice.<init> (constructor): javadocs are missing
public record ReaderSlice(int start, int length, int readerIndex) {
       ^
error: ReaderSlice.start (method): javadocs are missing
error: ReaderSlice.length (method): javadocs are missing
error: ReaderSlice.readerIndex (method): javadocs are missing
```

uschindler: The fix should be a separate PR, as this is too large. Maybe we do some small PR first with only the TermStats class and make it pass the tester. Would this be OK to you? I will help to fix the linter.

uschindler: I think the fix needs to be done here: https://github.com/apache/lucene/blob/9af3ef8952597ead37a5a9ce5373ed3bbd40238e/dev-tools/missing-doclet/src/main/java/org/apache/lucene/missingdoclet/MissingDoclet.java#L260

The problem is: how to know which of the ctors and methods are not synthetic?

uschindler: It's complicated: we need to find all accessor methods and automatically generated ctors and exclude them. You can get the accessor method from the record component. Maybe we should build a set of them and exclude them from the checks: https://docs.oracle.com/en/java/javase/21/docs/api/java.compiler/javax/lang/model/element/RecordComponentElement.html

Let's make a separate issue and remove all problematic records from this PR.

uschindler: Maybe as a quick fix we can exclude analysis of record ctors and methods until we have a fix for that. Let's rename that method to `isSyntheticMethod` (remove "enum" from name):
https://github.com/apache/lucene/blob/9af3ef8952597ead37a5a9ce5373ed3bbd40238e/dev-tools/missing-doclet/src/main/java/org/apache/lucene/missingdoclet/MissingDoclet.java#L274-L282

and modify it:

```java
  private boolean isSyntheticMethod(Element element) {
    if (element.getEnclosingElement().getKind() == ElementKind.RECORD) {
      // TODO: fix this to exclude autogenerated accessors and ctors of records, which is hard
      return true;
    }
    String simpleName = element.getSimpleName().toString();
    if (simpleName.equals("values") || simpleName.equals("valueOf")) {
      if (element.getEnclosingElement().getKind() == ElementKind.ENUM) {
        return true;
      }
    }
    return false;
  }
```

uschindler: I think to fix this we have to check for "mandated" elements: https://docs.oracle.com/en/java/javase/21/docs/api/java.compiler/javax/lang/model/util/Elements.Origin.html#MANDATED

Not sure how to do this, so let's do this in a separate issue.

shubhamvishu: Thank you so much for looking into this @uschindler! I like the idea of having a quick fix and exclude the records ctor and methods for now. Also, I even tried to somehow get rid of the earlier issues with records `TermStats` and `ReaderSlice` but explicitly declaring the ctor and methods but it then started complaining about some other public records we added in this PR from the list of packages you shared above where we always expect javadocs. Disabling the check for records till its fixed in a separate issue makes sense to me.💯

shubhamvishu: I opened #13329 with above quickfix change to temporarily disable the javadoc checks and confirmed the tasks passes for this PR with that change.

uschindler: I fixed the issue and added support for records and also its components (as `@param`) in the missing doclet linter: #13332

Once this is merged we can update this PR.

uschindler:
Hi, you can merge main into your branch and hopefully checks pass. For some packages like the "codecs" one the new code requires you to also document all record components with `@param`.

lucene/core/src/java/org/apache/lucene/util/TermAndVector.java
```diff
@@ -24,23 +24,7 @@
...
-  }
-
-  public BytesRef getTerm() {
-    return this.term;
-  }
-
-  public float[] getVector() {
-    return this.vector;
-  }
+public record TermAndVector(BytesRef term, float[] vector) {
```
uschindler, line 27: this class has the problem that it supplies a method to normalize the inner vector. This is against immutability of records. It is possible to do this, but feels bad.

lucene/core/src/java/org/apache/lucene/search/CollectionStatistics.java
```diff
@@ -116,7 +105,8 @@ public CollectionStatistics(
    *
    * @return field's name, not {@code null}
    */
-  public final String field() {
```
uschindler, line 119: we can remove all accessor methods and add the documentation into `@param` on the record declaration.

lucene/core/src/java/org/apache/lucene/util/TermAndVector.java
shubhamvishu: Makes sense! I changed it back to a class in the new revision

lucene/core/src/java/org/apache/lucene/util/packed/PackedInts.java
```diff
@@ -171,14 +171,7 @@ public final float overheadRatio(int bitsPerValue) {
...
-    public final int bitsPerValue;
-
-    public FormatAndBits(Format format, int bitsPerValue) {
-      this.format = format;
-      this.bitsPerValue = bitsPerValue;
-    }
+  public record FormatAndBits(Format format, int bitsPerValue) {
 
     @Override
     public String toString() {
```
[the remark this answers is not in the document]
shubhamvishu: Removed this now

lucene/core/src/java/org/apache/lucene/search/CollectionStatistics.java
shubhamvishu: Done! I also changed some more records I could find to be as leaner as possible.

shubhamvishu:
I addressed your comments in the new revision and all the checks are passing now after your fix. Thanks!

uschindler:
Thanks, looks fine. I have no time to do another closer review, please give me some time to proceed.

lucene/codecs/src/java/org/apache/lucene/codecs/blocktreeords/OrdsBlockTreeTermsWriter.java
```diff
@@ -139,18 +139,17 @@ public final class OrdsBlockTreeTermsWriter extends FieldsConsumer {
...
-    public final long numTerms;
-    public final long indexStartFP;
-    public final long sumTotalTermFreq;
-    public final long sumDocFreq;
-    public final int docCount;
-    public final BytesRef minTerm;
-    public final BytesRef maxTerm;
-
-    public FieldMetaData(
+  private record FieldMetaData(
```
uschindler, line 142: make this the default ctor (see above and remove the field assignments), just keep asserts or other checks

lucene/core/src/java/org/apache/lucene/document/NearestNeighbor.java
```diff
@@ -34,20 +34,16 @@
...
+  /**
+   * @param distanceSortKey The closest distance from a point in this cell to the query point,
+   *     computed as a sort key through {@link SloppyMath#haversinSortKey}. Note that this is an
+   *     approximation to the closest distance, and there could be a point in the cell that is
+   *     closer.
+   */
+  record Cell(
+      PointTree index, int readerIndex, byte[] minPacked, byte[] maxPacked, double distanceSortKey)
+      implements Comparable<Cell> {
+    Cell(
```
uschindler, line 46: remove that ctor and replace by default ctor and only copy minPacked and maxPacked (see above).

uschindler (changes requested):
I went through the first few files and have some changes:
- there were some hashCodes revoed but equals survied, please remove both
- some toString may be obsolete, too (unless the exact formatting is important)
- some ctors with exact same parameers like default can be rewritten to use the default syntax and remove the field assignments

I will proceed later on the weekend with other classes.

uschindler:
Hi @shubhamvishu,
can you check my comments? The changes here look fine, please fix the remaining issues.
Uwe

shubhamvishu:
Hi @uschindler , sorry I was on vacation until last week, so this PR stalled. I'll take a look at the comments today or tomorrow.

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/GeneratingSuggester.java
```diff
@@ -447,14 +446,8 @@ private static int commonCharacterPositionScore(String s1, String s2) {
...
-
-    Weighted(T word, int score) {
-      this.word = word;
-      this.score = score;
-    }
+  private record Weighted<T extends Comparable<T>>(T word, int score)
+      implements Comparable<Weighted<T>> {
 
     @Override
     public boolean equals(Object o) {
```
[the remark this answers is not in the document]
shubhamvishu: Removed

lucene/analysis/common/src/java/org/apache/lucene/analysis/synonym/word2vec/TermAndBoost.java
```diff
@@ -18,14 +18,13 @@
...
-
+/**
+ * Wraps a term and boost
+ *
+ * @param term the term
+ * @param boost the boost
+ */
+public record TermAndBoost(BytesRef term, float boost) {
   /** Creates a new TermAndBoost */
   public TermAndBoost(BytesRef term, float boost) {
```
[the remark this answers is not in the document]
shubhamvishu: Done!.....I changed others such occurrences that I could find as well

lucene/codecs/src/java/org/apache/lucene/codecs/blocktreeords/OrdsBlockTreeTermsWriter.java
shubhamvishu: addressed in the new revision

lucene/core/src/java/org/apache/lucene/document/NearestNeighbor.java
shubhamvishu: done

shubhamvishu:
Hi @uschindler , thanks for the review. I took another pass at the changes and pushed some commits addressing your comments. Please have a look when you get a chance. Thanks!

shubhamvishu:
Hi @uschindler, Could you please review the current code changes once you get some time and if it looks good maybe we can move this forward?

stefanvodita:
Added the blocker label to draw attention to this PR for the Lucene 10 release.

shubhamvishu:
Hi @uschindler @mikemccand @jpountz , I fixed the recent merge conflicts and looking for reviews as 10.0 release is blocked on this change. This PR has been attracting merge conflicts frequently since its touching 200+ files, so hoping we could close this sooner. Let me know if there is any concerns here(I'm up for quick followups). Thanks!

lucene/core/src/java/org/apache/lucene/search/CollectionStatistics.java
```diff
@@ -17,8 +17,9 @@
 package org.apache.lucene.search;
 
 import java.util.Objects;
-import org.apache.lucene.index.IndexReader; // javadocs
-import org.apache.lucene.index.Terms; // javadocs
+
+// javadocs
+// javadocs
```
jpountz, line 22: you can remove these 2 comments

lucene/misc/src/java/org/apache/lucene/misc/util/fst/UpToTwoPositiveIntOutputs.java
```diff
@@ -42,36 +42,11 @@
...
+  public record TwoLongs(long first, long second) {
+    public TwoLongs {
       assert first >= 0;
       assert second >= 0;
     }
-
-    @Override
-    public String toString() {
-      return "TwoLongs:" + first + "," + second;
-    }
```
jpountz, line 59: we'll get a different toString() but I think it's fine in this case

lucene/monitor/src/java/org/apache/lucene/monitor/HighlightsMatch.java
```diff
@@ -180,47 +180,16 @@ void addHit(String field, int startPos, int endPos, int startOffset, int endOffs
...
-      result = 31 * result + endPosition;
-      result = 31 * result + endOffset;
-      return result;
-    }
+  /**
+   * Represents an individual hit
+   *
+   * @param startPosition The start position
+   * @param startOffset The start offset
+   * @param endPosition The end positions
```
jpountz, line 188:
```suggestion
   * @param endPosition The end position
```

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/CombinedFieldQuery.java
```diff
@@ -163,11 +156,6 @@ public boolean equals(Object o) {
       FieldAndWeight that = (FieldAndWeight) o;
       return Float.compare(that.weight, weight) == 0 && Objects.equals(field, that.field);
     }
```
jpountz, line 158: Did you note remove this equals override because the record would behave differently? (I tried to quickly look up what record does wrt equality for floats and doubles but couldn't find it, will look later)

lucene/test-framework/src/java/org/apache/lucene/tests/codecs/asserting/AssertingLiveDocsFormat.java
```diff
@@ -68,9 +68,7 @@ public String toString() {
     return "Asserting(" + in + ")";
   }
 
-  static class AssertingBits implements Bits {
-    final Bits in;
-
+  record AssertingBits(Bits in) implements Bits {
```
jpountz, line 71: While it's not incorrect, I'm not sure if this class fits my mental model for a record as it's more about behaviors than about packaging other objects tohether.

lucene/test-framework/src/java/org/apache/lucene/tests/index/AssertingLeafReader.java
```diff
@@ -1357,15 +1357,8 @@ public int getDocCount() {
...
-
-    final PointValues pointValues;
-    final PointValues.PointTree in;
-
-    AssertingPointTree(PointValues pointValues, PointValues.PointTree in) {
-      this.pointValues = pointValues;
-      this.in = in;
-    }
+  record AssertingPointTree(PointValues pointValues, PointValues.PointTree in)
+      implements PointValues.PointTree {
```
jpountz, line 1361: likewise: keep it a regular class?

lucene/core/src/java/org/apache/lucene/util/FixedBits.java
```diff
@@ -17,15 +17,7 @@
...
-final class FixedBits implements Bits {
-
-  final long[] bits;
-  final int length;
-
-  FixedBits(long[] bits, int length) {
-    this.bits = bits;
-    this.length = length;
-  }
+record FixedBits(long[] bits, int length) implements Bits {
```
jpountz, line 20: it doesn't really fit my mental model for a record, maybe undo the refactoring on this class?

jpountz (approved):
I left some minor comments but the change looks good to me in general. Can you add an entry to lucene/MIGRATE.txt that explains that TopDocs, CollectionStatistics, TermStatistics and LeafMetadata were converted to records? I'm only considering mentioning these because these are the most user-facing APIs that are migrated to a record from what I can tell (please let me know if you think I'm missing an important one).

lucene/misc/src/java/org/apache/lucene/misc/util/fst/UpToTwoPositiveIntOutputs.java
shubhamvishu: Yes, its fine in this case.

lucene/test-framework/src/java/org/apache/lucene/tests/codecs/asserting/AssertingLiveDocsFormat.java
shubhamvishu: I agree, lets keep it as is.

lucene/core/src/java/org/apache/lucene/util/FixedBits.java
shubhamvishu: Sure!

shubhamvishu:
@jpountz I addressed your comments and added a CHANGES, MIGRATE entry now. Thanks for reviewing!

lucene/CHANGES.txt
```diff
@@ -118,6 +118,8 @@ API Changes
   argument with a `FacetsCollectorManager` and update the return type to include both `TopDocs` results as well as
   facets results. (Luca Cavanna)
 
+* GITHUB#13328: Convert CollectionStatistics, TermStatistics and LeafMetadata etc. to record classes. (Shubham Chaudhary)
```
mikemccand, line 121: Maybe reword to `Convert many basic Lucene classes to record classes, including X, Y, and Z`?  It's not just these three classes, if I'm reading the PR correctly, and the `etc.` should not have so much power ;)

lucene/MIGRATE.md
```diff
@@ -193,6 +193,7 @@ access the members using method calls instead of field accesses. Affected classe
 
 - `IOContext`, `MergeInfo`, and `FlushInfo` (GITHUB#13205)
 - `BooleanClause` (GITHUB#13261)
+- `CollectionStatistics`, `TermStatistics` and `LeafMetadata` (GITHUB#13328)
```
mikemccand, line 196: And  maybe change this wording too (same as `CHANGES.txt`)?  I think @jpountz asked for `TopDocs` to also be mentioned?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/CombinedFieldQuery.java
mikemccand: > (I tried to quickly look up what record does wrt equality for floats and doubles but couldn't find it, will look later)

I found one mention of floating point numbers in the [JEP for records](https://openjdk.org/jeps/395):

```
In addition, for all record classes the implicitly declared equals method is implemented so that it is reflexive and that it behaves consistently with hashCode for record classes that have floating point components. Again, explicitly declared equals and hashCode methods should behave similarly.
```

I'm not sure I understand that paragraph :)

mikemccand (approved):
Thank you for persisting here @shubhamvishu -- this is a massive rote refactoring change.  We should rather bias towards quickly merging such changes since they pick up conflicts so quickly.  I just left some minor feedback about wording in CHANGES and MIGRATE, else +1 to merge.

lucene/CHANGES.txt
shubhamvishu: This looks better indeed. I'll reword it.

lucene/MIGRATE.md
shubhamvishu: Sure, I'll reword. I haven't changed `TopDocs` in this PR and hence I didn't include it. Maybe some confusion(prolly because there are just too many classes changed here)

stefanvodita:
I think we've run into a conflict with #13689. @jpountz - should I fix that or are you already doing it?

jpountz:
@javanna just fixed it, thanks @javanna !

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/CombinedFieldQuery.java
uschindler: Hi,
The equals methods for floats is generated by invokedynamic using a helper. When a record component is a float, the autogenerated impl is generated at runtime as combination of several Method handles. It compares them using this code:

https://github.com/openjdk/jdk/blob/9785e19f3f87306cabc26a862d35b89d41cfef93/src/java.base/share/classes/java/lang/runtime/ObjectMethods.java#L157

This is equivalent to the already existing code. So please remove all custom equals methods.

Unfortunately in the merged PR there are so one more. I will add a not on them. Please remove them in a followup.

lucene/core/src/java/org/apache/lucene/search/MaxScoreAccumulator.java
```diff
@@ -101,11 +93,6 @@ public boolean equals(Object o) {
       return docBase == result.docBase && Float.compare(result.score, score) == 0;
     }
 
-    @Override
```
uschindler, line 104: Please also remove the equals, see this comment: https://github.com/apache/lucene/pull/13328#discussion_r1752842660

uschindler (commented):
There is one as symmetry in removal of equals/hashcode.

Please fix this in a new PR.

lucene/core/src/java/org/apache/lucene/search/MaxScoreAccumulator.java
uschindler: To string is also useless here. Please remove, too

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/CombinedFieldQuery.java
shubhamvishu: Nice! I'll check and followup with removing others I could find.

lucene/core/src/java/org/apache/lucene/search/MaxScoreAccumulator.java
shubhamvishu: Sure, I'll fix it in a followup PR. I could only find this instance in the classes changed to records where we have removable custom equals/toString but I'll take another pass to be sure.

shubhamvishu: I opened a followup PR #13769. Thanks!

PR #13329, opened 2024-04-30
Temporarily exclude record ctors and methods from javadoc checks

shubhamvishu:
### Description

In #13328 we got to know `renderSiteJavadoc` task fails for javadocs checks are not working as expected with records ctor and methods(synthetic ones). As pointed by @uschindler its hard to find which ones are synthetic and fix it. Meanwhile, we could temporarily disable the javadoc checks for record ctors and methods till the main issue is resolved.


uschindler:
I am working on a solution, but also the enum ctors are not even marked as "mandated" elements.

My idea was to replace the whole method (also for the enums by this:

```java
  private boolean isSyntheticEnumMethod(Element element) {
    if (elementUtils.getOrigin(element) != Origin.EXPLICIT) {
      System.err.println(element.getSimpleName().toString());
      return true;
    }
    return false;
  }
```

But the `println` only lists default constructors, which is bad. So I have no idea how to detect those methods.

uschindler:
Maybe it works for records, because they are "newer". I will add a fake record for testing.

uschindler:
With the above code it was able to detect the default constructor of records, but it did not detect the record component accessor methods (same issue like with enums).

shubhamvishu:
Nice....this is great!

 I think this takes care of record ctors along with synthetic methods due to `ExecutableElement.getEnclosingElement().getEnclosedElements()` returning ctor also as enclosed element in `Elements#recordComponentFor`?

uschindler:
> Nice! This is neat. Thank you Uwe!
> 
> I think this takes care of record ctors along with synthetic methods due to `ExecutableElement.getEnclosingElement().getEnclosedElements()` returning ctor also as enclosed element in `Elements#recordComponentFor`?

No, the ctors are excluded because of the first if statement. Actually regarding the comment "UweSays: It should not happen but it happens!" the following if statements should be all not needed (because the elements are not explicit - but mandated - it still reports them.

In short:
- `Elements#recordComponentFor` tests every method if it is the accessor of a record component, if it returns something != null, we can exclude it
- `Elements#getOrigin(element) != Origin.EXPLICIT` finds all ctors which are not declared

uschindler:
I created a new PR #13332 implementing the linter correctly:
- exclude default ctors
- exclude record accessor methods
- check in level=METHOD that all record components have a documentation (using `@param`)

uschindler:
I am closing this in favour of #13332.

PR #13330, opened 2024-04-30
Remove unused "implements Accountable".

jpountz:
A number of file formats no longer implement `Accountable`. This allows removing this interface on a number of internal classes as well.

### DIFF
--- a/lucene/codecs/src/java/org/apache/lucene/codecs/blockterms/VariableGapTermsIndexReader.java
+++ b/lucene/codecs/src/java/org/apache/lucene/codecs/blockterms/VariableGapTermsIndexReader.java
@@ -19,7 +19,6 @@ package org.apache.lucene.codecs.blockterms;
 import static org.apache.lucene.util.fst.FST.readMetadata;
 
 import java.io.IOException;
-import java.util.Collection;
 import java.util.HashMap;
 import org.apache.lucene.codecs.CodecUtil;
 import org.apache.lucene.index.CorruptIndexException;
@@ -27,8 +26,6 @@ import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.IndexFileNames;
 import org.apache.lucene.index.SegmentReadState;
 import org.apache.lucene.store.ChecksumIndexInput;
-import org.apache.lucene.util.Accountable;
-import org.apache.lucene.util.Accountables;
 import org.apache.lucene.util.BytesRef;
 import org.apache.lucene.util.fst.BytesRefFSTEnum;
 import org.apache.lucene.util.fst.FST;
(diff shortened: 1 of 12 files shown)


lucene/codecs/src/java/org/apache/lucene/codecs/blockterms/VariableGapTermsIndexReader.java
```diff
@@ -168,22 +165,6 @@ public FieldIndexEnum getFieldEnum(FieldInfo fieldInfo) {
   @Override
   public void close() throws IOException {}
 
-  @Override
-  public long ramBytesUsed() {
-    long sizeInBytes = 0;
-    for (FST<Long> entry : fields.values()) {
```
benwtrent, line 174: I thought FST's can get pretty large? Or is all the main cost off-heap & no-one really looks at ram-bytes usage for the TermsIndexReader?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
```diff
@@ -121,9 +121,6 @@ void initIndexInput() {
   public Stats computeBlockStats() throws IOException {
 
     Stats stats = new Stats(fr.parent.segment, fr.fieldInfo.name);
-    if (fr.index != null) {
-      stats.indexNumBytes = fr.index.ramBytesUsed();
-    }
```
benwtrent, line 126: This removes all writing to `indexNumBytes`? Do we want to remove this?

lucene/codecs/src/java/org/apache/lucene/codecs/blockterms/VariableGapTermsIndexReader.java
jpountz: These FSTs are on-heap and could get large indeed. I'm removing `Accountable` here to simplify and because I'd expect very few users if any to take advantage of it as it's a non-default postings format (`LuceneVarGapFixedInterval` and `LuceneVarGapDocFreqInterval`) and requires a lot of casting to get access to the Accountable interface: first casting the LeafReader into a CodecReader to be able to pull a PostingsReader, then casting the PostingsReader into an Accountable before PostingsReader no longer implements Accountable since 9.0.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
jpountz: thanks for catching, it's a leftover from a previous change, I'll add it back!

benwtrent (approved):

PR #13331, opened 2024-04-30
Make segment/field attribute updates thread-safe.

jpountz:
Because of concurrent merging (#13124), multiple threads may be updating (different) attributes concurrently, so we need to make reads and writes to attributes thread-safe.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/FieldInfo.java
+++ b/lucene/core/src/java/org/apache/lucene/index/FieldInfo.java
@@ -16,6 +16,8 @@
  */
 package org.apache.lucene.index;
 
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/index/FieldInfo.java
```diff
@@ -628,12 +630,17 @@ public String getAttribute(String key) {
...
    * after merge is undefined.
    */
-  public String putAttribute(String key, String value) {
-    return attributes.put(key, value);
+  public synchronized String putAttribute(String key, String value) {
+    HashMap<String, String> newMap = new HashMap<>(attributes);
+    String oldValue = newMap.put(key, value);
+    // This needs to be thread-safe as multiple threads may be updating (different) attributes
+    // concurrently due to concurrent merging.
+    attributes = Collections.unmodifiableMap(newMap);
```
benwtrent, line 638: I don't understand why we are copying the map, why not use a `ConcurrentHashMap`?

Is this because we don't want this changing at all after something calls `attributes()`?

jpountz: Indeed. Since SegmentInfo's attributes were already using this approach, I thought it made sense to follow the same approach instead of doing something different.

benwtrent (approved):

PR #13332, opened 2024-04-30
Improve MissingDoclet linter to check records correctly

uschindler:
New checks: 
- exclude default ctors
- exclude accessor methods (like with enums)
- on "method" level checking also check that every record component has an @param tag

See #13329 for more information.


uschindler:
We have no tests for the linter, but I validated that the code complains correctly on missing javadocs for components and excludes component accessors.

uschindler:
I will merge this later this evening so we can proceed with #13328

rmuir (approved):
thank you for taking care of this

PR #13336, opened 2024-05-02
Add test case to ensure scalar quantization adheres to known ranges

benwtrent:
Lucene provides int7 & int4 quantization, we should ensure via our tests that the quantized values are within expected ranges.


ChrisHegarty (approved):
LGTM

ChrisHegarty:
Our scorer implementations should be able to take advantage of the ranges being in:
* int7: 0 - 127 (inclusive)
* int4: 0 - 15 ( inclusive)

PR #13339, opened 2024-05-02
Add a MemorySegment Vector scorer - for scoring without copying on-heap

ChrisHegarty:
Add a MemorySegment Vector scorer - for scoring without copying on-heap.

The vector scorer loads values directly from the backing memory segment when available. Otherwise, if the vector data spans across segments, or is a query vector, the scorer copies the vector data on-heap. 

A benchmark shows ~2x performance improvement of this scorer over the default copy-on-heap scorer. The benchmark need a little more scrutiny and evaluation on different platforms. Here's the results on a Max M2:

```
Benchmark                                      (size)   Mode  Cnt  Score   Error   Units
VectorScorerBenchmark.binaryDotProductDefault    1024  thrpt    5  1.391 ± 0.016  ops/us
VectorScorerBenchmark.binaryDotProductMemSeg     1024  thrpt    5  3.013 ± 0.207  ops/us
```

The scorer currently only operates on vectors with an element size of `byte`, since loading vector data from float[] (the fallback), is only supported in JDK 22. We can evaluate if and how to support floats separately. See https://bugs.openjdk.org/browse/JDK-8318678

The vector scorer is implicitly tied to the Panama Vector Utils implementation - you can only have a Memory segment scorer if the Panama vector implementation is present. There is a little room for improvement in how these things are initialised and structured.


lucene/core/src/java/org/apache/lucene/internal/vectorization/VectorizationProvider.java
```diff
@@ -198,6 +201,11 @@ private static void ensureCaller() {
   private static final class Holder {
     private Holder() {}
 
-    static final VectorizationProvider INSTANCE = lookup(false);
+    // TODO: this is not quite right. But we should be able to run tests with Panama Vector
+    static boolean testMode() {
+      return TESTS_VECTOR_SIZE.isPresent() || TESTS_FORCE_INTEGER_VECTORS;
```
ChrisHegarty, line 206: This is not quite right, but it should be possible to run tests with Panama Vector (but our test runner always sets these properties?)

lucene/core/src/java21/org/apache/lucene/internal/vectorization/MemorySegmentByteVectorScorerSupplier.java
```diff
@@ -0,0 +1,129 @@
...
+    input = FilterIndexInput.unwrap(input);
+    if (!(input instanceof MemorySegmentAccess)) {
+      return Optional.empty();
+    }
+    checkInvariants(maxOrd, vectorByteSize, input);
+    return switch (type) {
+      case DOT_PRODUCT -> Optional.of(new DotProduct(dims, maxOrd, vectorByteSize, input, values));
+      case EUCLIDEAN -> Optional.of(new Euclidean(dims, maxOrd, vectorByteSize, input, values));
+      case MAXIMUM_INNER_PRODUCT -> Optional.empty(); // TODO: implement MAXIMUM_INNER_PRODUCT
+      case COSINE -> Optional.empty(); // TODO: implement Cosine
```
ChrisHegarty, line 64: I'll get Cosine and MAX Inner prod added.

lucene/core/src/java21/org/apache/lucene/internal/vectorization/MemorySegmentByteVectorScorerSupplier.java
```diff
@@ -0,0 +1,129 @@
...
+
+  final void checkOrdinal(int ord, int maxOrd) {
+    if (ord < 0 || ord >= maxOrd) {
+      throw new IllegalArgumentException("illegal ordinal: " + ord);
+    }
+  }
+
+  protected final MemorySegment getSegment(int ord, byte[] scratch) throws IOException {
+    checkOrdinal(ord, maxOrd);
+    int byteOffset = ord * vectorByteSize; // TODO: random + meta size
```
ChrisHegarty, line 94: I wanna generalise this so we can use it for scalar quantised too - so a vector byte size + N bytes

lucene/core/src/java/org/apache/lucene/internal/vectorization/VectorizationProvider.java
uschindler: Actually at the moment the easiest is to pass CI=true as environment variable to Gradle.

lucene/core/src/java/org/apache/lucene/codecs/hnsw/FlatVectorScorerProvider.java
```diff
@@ -0,0 +1,63 @@
...
+  /** Returns the default FlatVectorsScorer. */
+  public static FlatVectorsScorer createDefault() {
+    if (isPanamaVectorUtilSupportEnabled()) {
+      // we only enable this scorer if the Panama vector provider is also enabled
+      return lookup();
+    }
+    return new DefaultFlatVectorScorer();
+  }
+
+  public static FlatVectorsScorer lookup() {
```
uschindler, line 40: I don't like the additional code here to lookup the FlatVectorScorerProvider. This should be done in VectorizationProvider.

So we should add a method `newFlatVectorScorer()` to VectorizationProvider. The default one returns the default only, the other one the wrapped one.

This is how the VectorizationProvider interface was designed. It should get method to create "vectorized" instances or return static instances.

So this whole class should be removed and the code should just call VectorizationProvider newDefaultFlatVectorScorer().

uschindler (changes requested):
I don't like the additional Provider interface. There should only be one instantiating all implementations that can be vectorized.

lucene/core/src/java/org/apache/lucene/codecs/hnsw/FlatVectorScorerProvider.java
uschindler: When integrating this code into the VectorizationProvider all those checks below are also obsolete because everything that returns components with vectorization support are at one place.

I may also change this here in your branch if required.

lucene/core/src/java/org/apache/lucene/internal/vectorization/VectorizationProvider.java
uschindler: We must remove this again, because the problem with tests is that they use a special JVM flag to disable tiered compilation, so C2 never gets active.

By the trick `CI=true` we enable this. We should better document this or maybe add another sysprop. The code added here makes no sense its just a workaround to enforce slowness. It works if you have a very fast machine or only run a few tests.

ChrisHegarty: ha! yes. I forgot. Thanks for the reminder. I'll use that for testing.

```
export CI=true; ./gradlew :lucene:core:test
```

lucene/core/src/java/org/apache/lucene/codecs/hnsw/FlatVectorScorerProvider.java
ChrisHegarty: This is much cleaner - done. Maybe you have further improvements, but I think it looks much better now.

lucene/core/src/test/org/apache/lucene/search/TestKnnByteVectorQueryMMap.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.apache.lucene.search;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import org.apache.lucene.store.MMapDirectory;
+import org.apache.lucene.tests.store.BaseDirectoryWrapper;
+import org.apache.lucene.tests.store.MockDirectoryWrapper;
+
+public class TestKnnByteVectorQueryMMap extends TestKnnByteVectorQuery {
```
ChrisHegarty, line 25: I know that we don't strictly need this, but I do find it helpful to have testing over an mmap memory segment index input on every run. And also have it programmatically available, rather than through a system property.

msokolov:
So excited to see this finally come to fruition! No more double-buffering!

uschindler (commented):
Looks much better now. I will have a closer look later, so not yet a +1

uschindler (dismissed):

uschindler:
How can i change the review to "undecided"?

lucene/core/src/java/org/apache/lucene/codecs/hnsw/FlatVectorScorerProvider.java
uschindler: Yes very nice now. The whole construct is still a bit strange but at least in main branch quite clean, because we know that MMapDir always available with memory segments, so the vector provider does not need to differentiate. If we want to backport there are more knobs that could change.

ChrisHegarty:
> How can i change the review to "undecided"?

I re-requested ur review - so there is no official reviewer yet. Take ur time. I have some luceneutil benchmarks to run, etc.

uschindler:
> > How can i change the review to "undecided"?
> 
> I re-requested ur review - so there is no official reviewer yet. Take ur time. I have some luceneutil benchmarks to run, etc.

The benchmark code seems broken after your changes.

lucene/core/src/java21/org/apache/lucene/internal/vectorization/MemorySegmentByteVectorScorerSupplier.java
```diff
@@ -0,0 +1,133 @@
...
+          new DotProductByteVectorScorerSupplier(dims, maxOrd, vectorByteSize, input, values));
+      case EUCLIDEAN -> Optional.of(
+          new EuclideanByteVectorScorerSupplier(dims, maxOrd, vectorByteSize, input, values));
+      case MAXIMUM_INNER_PRODUCT -> Optional.empty(); // TODO: implement MAXIMUM_INNER_PRODUCT
+      case COSINE -> Optional.empty(); // TODO: implement Cosine
+    };
+  }
+
+  MemorySegmentByteVectorScorerSupplier(
+      int dims, int maxOrd, int vectorByteSize, IndexInput input, RandomAccessVectorValues values) {
```
uschindler, line 73: `input` should be typed `MemorySegmentAccess` from beginning, so remove cast below, cast should be in create method when we know what it is.

ChrisHegarty: yeah, there is a clear tension here. I created MemorySegmentAccess so as to avoid making MemorySegmentIndexInput public. We can keep it, and check and cast `input` to both types we need `IndexInput` and `MemorySegmentAccess`. Or we can try to unify the types. I don't have a strong preference, other than we should do one or the other.

ChrisHegarty:
> Dismissing Uwe's review, since he is undecided. Can be explicitly added later, when we convince him ;-)

(Oh, this looks harsh!) I hope that I did this right. If not, I apologise. No offence intended. I just want to reflect @uschindler's comment above about being currently undecided.

lucene/core/src/java21/org/apache/lucene/internal/vectorization/DotProductByteVectorScorerSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+  public float score(int node) throws IOException {
+    // divide by 2 * 2^14 (maximum absolute value of product of 2 signed bytes) * len
+    float raw = PanamaVectorUtilSupport.dotProduct(first, getSegment(node, scratch2));
+    return 0.5f + raw / (float) (dims * (1 << 15));
+  }
+
+  @Override
+  public DotProductByteVectorScorerSupplier copy() throws IOException {
+    return new DotProductByteVectorScorerSupplier(
+        dims, maxOrd, vectorByteSize, input.clone(), values);
```
uschindler, line 44: this was the reason why we needed the original input!

uschindler (commented):
Just a few minor things, still thinking about the whole setup.

Current review state: You got me: The idea is good to have just a minimal interface between memory segments and the scorer so we only need an instanceof check.

It is a bit risky if we unwrap all filter readers (there may be other filters on top of the indexinput except test ones...).

lucene/core/src/java21/org/apache/lucene/internal/vectorization/DotProductByteVectorScorerSupplier.java
uschindler: it's cool that the interface automatically adds the bridge method... thanks javac! :-)

lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/VectorScorerBenchmark.java
```diff
@@ -0,0 +1,114 @@
...
+      out.writeBytes(vec2, 0, vec2.length);
+    }
+    in = dir.openInput("vector.data", IOContext.DEFAULT);
+    vectorValues = vectorValues(size, 2, in);
+    scorer =
+        VectorizationProvider.getInstance()
+            .newFlatVectorScorer()
+            .getRandomVectorScorerSupplier(DOT_PRODUCT, vectorValues);
+
+    // Ensure we're using the right vector scorer
```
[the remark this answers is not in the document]
ChrisHegarty: Was mainly for my own sanity. Can be removed.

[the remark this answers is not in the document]
uschindler: anyways if you have wrong cpu, also all other benchmark will be as fast as non-vectorized one. In this case the assertion just fails, but this is not strongly needed. We print a log message anyways if vectorization is enabled and you see this in the benchmark anyways.

lucene/core/src/java21/org/apache/lucene/internal/vectorization/MemorySegmentByteVectorScorerSupplier.java
```diff
@@ -0,0 +1,134 @@
...
+   * returned.
+   */
+  public static Optional<MemorySegmentByteVectorScorerSupplier> create(
+      int dims,
+      int maxOrd,
+      int vectorByteSize,
+      VectorSimilarityFunction type,
+      IndexInput input,
+      RandomAccessVectorValues values) {
+    input = FilterIndexInput.unwrap(input);
```
uschindler, line 56: this is my only problem: if we unwrap all index inputs we may break some custom wrapper code of users? We should not do this. If we can get rid of that I am fine.

Do we wrap the readers at other places, too!?

uschindler: if we wrap some CFS files, its breaks, too.

ChrisHegarty: Argh!! We don’t have a pattern for unwrapping here??

uschindler: Can we simply remove the unwrapping here?

ChrisHegarty: Hmmm.. yeah maybe. The unwrapping was initially for an instanceof check, which we still do. Don’t we still need to unwrap to do this? Otherwise, it’ll not be executed much (at all?) in tests?

lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/VectorScorerBenchmark.java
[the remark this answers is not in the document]
ChrisHegarty: removed.

lucene/core/src/java21/org/apache/lucene/internal/vectorization/MemorySegmentByteVectorScorerSupplier.java
uschindler: As far as I remember, we don't have any wrappers in tests for IndexInputs. Need to check, I am away from IDE at moment.

In general it would be good to allow us to only remove wrappers added by tests.

ChrisHegarty: ++

ChrisHegarty: The only one I see when running the tests is:
   `org.apache.lucene.tests.store.MockIndexInputWrapper`

ChrisHegarty: I'm not trilled with this!!  653bd26  Open to better ideas.

ChrisHegarty: This is a bit better [fa6db68](https://github.com/apache/lucene/pull/13339/commits/fa6db6856269b6675d673f22f071a589f0750eee)

uschindler:
Still working on reviewing, busy busy busy and on vacation starting from tomorrow!

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java
```diff
@@ -139,7 +139,7 @@ public final class Lucene99HnswVectorsFormat extends KnnVectorsFormat {
 
   /** The format for storing, reading, merging vectors on disk */
   private static final FlatVectorsFormat flatVectorsFormat =
-      new Lucene99FlatVectorsFormat(new DefaultFlatVectorScorer());
+      new Lucene99FlatVectorsFormat(FlatVectorScorerUtil.newFlatVectorScorer());
```
benwtrent, line 142: I would say we also update the `Lucene99ScalarQuantizedVectorsFormat`, we call 

```
    this.flatVectorScorer = new Lucene99ScalarQuantizedVectorScorer(new DefaultFlatVectorScorer());
```

This way folks that send `byte[]` values to that scorer can still get the nice benefits here.

ChrisHegarty:
> Still working on reviewing, busy busy busy and on vacation starting from tomorrow!

I appreciate your ongoing review. Pause your review for now, enjoy your vacation, and I'll not merge before you return and have had time to finish review.

lucene/core/src/java/org/apache/lucene/store/FilterIndexInput.java
```diff
@@ -40,6 +48,19 @@ public static IndexInput unwrap(IndexInput in) {
...
+   */
+  public static IndexInput unwrapOnlyTest(IndexInput in) {
+    while (in instanceof FilterIndexInput && TEST_FILTER_INPUTS.contains(in.getClass())) {
+      in = ((FilterIndexInput) in).in;
+    }
+    return in;
+  }
+
+  // org.apache.lucene.tests.store.MockIndexInputWrapper"
+
```
benwtrent, line 63:
```suggestion

```
Unnecessary?

lucene/core/src/java21/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java
```diff
@@ -307,39 +310,44 @@ private float squareDistanceBody(float[] a, float[] b, int limit) {
 
   @Override
   public int dotProduct(byte[] a, byte[] b) {
+    return dotProduct(MemorySegment.ofArray(a), MemorySegment.ofArray(b));
+  }
+
+  public static int dotProduct(MemorySegment a, MemorySegment b) {
+    assert a.byteSize() == b.byteSize();
```
benwtrent, line 317: 😍

lucene/core/src/test/org/apache/lucene/search/TestKnnByteVectorQueryMMap.java
benwtrent: This is good instead of just having a sysprop or randomness.

lucene/core/src/java21/org/apache/lucene/internal/vectorization/MemorySegmentByteVectorScorerSupplier.java
benwtrent: While I don't foresee us doing this, but it is conceivable that the flat vector storage will change.

I think we have adequate test coverage to catch such a weird behavior, but maybe we need to change the name or add a comment reflecting that it relies on how Lucene95 stored the flat vectors?

benwtrent (commented):
I know next to nothing about the memory segment logic. however, this change is remarkably clean for such a big change!

I left comments about the vector codec & scoring interfaces.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java
ChrisHegarty: Done.

msokolov:
Question / confirmation please -- since this requires Panama, thus JDK 21+ (I think?) it can only target Lucene 10+, correct?

ChrisHegarty:
> Question / confirmation please -- since this requires Panama, thus JDK 21+ (I think?) it can only target Lucene 10+, correct?

Correct that it requires JDK 21+ and Panama Vector, but we can target 9.x also.  And I do plan to backport to 9.x.

The difference between 9.x and 10 is that in 9.x the memory segment mmap provider is optional (`enableMemorySegments`) - so it can be disabled, whereas in 10 it is not. In 9.x we will need an extra check that the memory segment mmap provider is actually enabled, but it is otherwise already hooked off the Panama Vectors provider so it should be fine.

lucene/core/src/java21/org/apache/lucene/internal/vectorization/MemorySegmentByteVectorScorerSupplier.java
```diff
@@ -0,0 +1,237 @@
...
+import org.apache.lucene.util.hnsw.RandomAccessVectorValues;
+import org.apache.lucene.util.hnsw.RandomVectorScorer;
+import org.apache.lucene.util.hnsw.RandomVectorScorerSupplier;
+
+/**
+ * A scorer of vectors whose element size is byte.
+ *
+ * <p>This class is both a scorer supplier and a scorer.
+ */
+public abstract sealed class MemorySegmentByteVectorScorerSupplier
```
benwtrent, line 36: we need to be clear that this is tied to the current codec. Since we are reading offsets ourselves, it relies on the codec storing vectors the way it is. I think we can communicate this by naming `Lucene99MemorySegmentByteVectorScorerSupplier`

benwtrent (approved):
From a vector codec similarity point of view, I like this. Its fairly unobtrusive. 

I leave the memory segment, etc. review for others.

uschindler:
> > Question / confirmation please -- since this requires Panama, thus JDK 21+ (I think?) it can only target Lucene 10+, correct?
> 
> Correct that it requires JDK 21+ and Panama Vector, but we can target 9.x also. And I do plan to backport to 9.x.
> 
> The difference between 9.x and 10 is that in 9.x the memory segment mmap provider is optional (`enableMemorySegments`) - so it can be disabled, whereas in 10 it is not. In 9.x we will need an extra check that the memory segment mmap provider is actually enabled, but it is otherwise already hooked off the Panama Vectors provider lookup so it should be fine.

Still not on vacation... You can easily backport this. Bo extra checks needed. If memory segments are disabled in 9.x, the IndexInput would not implement the new interface and therefore the code never triggers. If everything uses instance of checks all is fine.

uschindler:
Backporting would only require that you may need to duplicate versions for 19, 20, 21+. 19 has no vectorization, but 20 and 21 have identical vector code but differences in memory segments and that may hurt you.

ChrisHegarty:
> Backporting would only require that you may need to duplicate versions for 19, 20, 21+. 19 has no vectorization, but 20 and 21 have identical vector code but differences in memory segments and that may hurt you.

If it hurts too much (to get this on JDK 20) then we'll limit it to JDK 21+. :-)

lucene/core/src/java21/org/apache/lucene/internal/vectorization/MemorySegmentByteVectorScorerSupplier.java
ChrisHegarty: Ok. Done.

lucene/core/src/test/org/apache/lucene/internal/vectorization/TestVectorScorer.java
```diff
@@ -0,0 +1,324 @@
...
+            var scorer4 = MEMSEG_SCORER.getRandomVectorScorer(sim, vectorValues, vectors[idx0]);
+            assertEquals(scorer4.score(idx1), expected, DELTA);
+          }
+        }
+      }
+    }
+  }
+
+  // Tests with a large amount of data (> 2GB), which ensures that data offsets do not overflow
+  @Nightly
```
ChrisHegarty, line 223: This test creates a big file, but it's currently the only way to test that the offset calculations are correctly handled without overflow (if implemented using int)!

uschindler:
Back from vacation. Will look into that till Tuesday!

uschindler (commented):
To me this looks fine. Very clean.
It's not easy to undertstand, but at some point we may change this to be in core classes and we can get rid of all those wrappers.

One thing to think about: I had another idea, which may also help to allow direct use of our vectors from NIOFSDirectory and ByteBuffersDirectory (used by NRT):

How about changing from MemorySegment to ByteBuffer views? They can also be used by the vectorization code.

This would also allow to have the interface we need in main classes and not required to be in Java 21 soureSet. In addition ByteBuffersDirectory and NIOFSDircetory could directly return bytebuffer views, too.

We may add a method like getByteBufferSlice(....).

What do you think?

lucene/core/src/java/org/apache/lucene/internal/vectorization/VectorizationProvider.java
```diff
@@ -91,6 +92,8 @@ public static VectorizationProvider getInstance() {
    */
   public abstract VectorUtilSupport getVectorUtilSupport();
 
+  public abstract FlatVectorsScorer newFlatVectorScorer();
```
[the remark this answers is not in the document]
ChrisHegarty: Ah yes, good call. Done.

lucene/core/src/java21/org/apache/lucene/internal/vectorization/PanamaVectorizationProvider.java
```diff
@@ -73,4 +75,9 @@ private static <T> T doPrivileged(PrivilegedAction<T> action) {
   public VectorUtilSupport getVectorUtilSupport() {
     return vectorUtilSupport;
   }
+
+  @Override
+  public FlatVectorsScorer newFlatVectorScorer() {
```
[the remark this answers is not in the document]
ChrisHegarty: I don't envisage stateful. Replacing _new_ with _get_ is more flexible. Done.

lucene/core/src/java/org/apache/lucene/codecs/hnsw/FlatVectorScorerUtil.java
```diff
@@ -35,6 +35,6 @@ private FlatVectorScorerUtil() {}
    * on certain platforms. Otherwise, a DefaultFlatVectorScorer is returned.
    */
   public static FlatVectorsScorer newFlatVectorScorer() {
```
uschindler, line 37: also rename here

ChrisHegarty: Oops. done.

ChrisHegarty:
> We may add a method like getByteBufferSlice(....).

I experimented locally with similar before, and the performance impact when converting to/from MemorySegment was horrible. I don't disagree that for NIO/BB dir implementations this could be useful, but it would not intersect with the memory segment implementation, since it performs horribly.

uschindler:
> > We may add a method like getByteBufferSlice(....).
> 
> I experimented locally with similar before, and the performance impact when converting to/from MemorySegment was horrible. I don't disagree that for NIO/BB dir implementations this could be useful, but it would not intersect with the memory segment implementation, since it performs horribly (when converting from BB to MS, and vice versa).

OK, this is not good. From my code review it looked like the DirectByteBuffer also only has a addres and therefor the whole thing is identical by performance (except that we have to get a MemorySegment slice first and then call `toByteBuffer()`). So this is interesting, but then I we can't do anything.

Mabe theres some optimizations miissing in ByteBuffer support of vector API.

So looks fine then. Working on a final review.

lucene/core/src/java21/org/apache/lucene/internal/vectorization/MemorySegmentFlatVectorsScorer.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Override
+  public RandomVectorScorerSupplier getRandomVectorScorerSupplier(
+      VectorSimilarityFunction similarityType, RandomAccessVectorValues vectorValues)
+      throws IOException {
+    // currently only supports binary vectors
+    if (vectorValues instanceof RandomAccessVectorValues.Bytes && vectorValues.getSlice() != null) {
+      var scorer =
+          Lucene99MemorySegmentByteVectorScorerSupplier.create(
```
uschindler, line 45: is this ok to return here the Lucene99 variant, although this interface is more generic? How can we migrate to later vector formats?

uschindler (commented):
One more question: How can we migrate to a later index format version? Because we have the `MemorySegmentFlatVectorScorer` (which has no knowledge about version) and this one returns the `Lucene99MemorySegment*` classes.

So in my opinion this is only solvable if we rename also the `MemorySegmentFlatVectorScorer` to `Lucene99MemorySegmentFlatVectorScorer` and also adapt the getter name.

Except this inconsistency it looks fine. A bit hairy but if performance is high prio, I accept this as a step towards getting rid of IndexInput totally.

lucene/core/src/java21/org/apache/lucene/internal/vectorization/MemorySegmentFlatVectorsScorer.java
ChrisHegarty: Oh! And I plan to revisit this once this PR is merged, so as to layer the scalar quantised flat scorer on top. So I'll likely add the additional offset then, as a follow up when it's actually used.

uschindler (approved):

uschindler (approved):
now all fine.

uschindler:
Looks like first Java 22 build also worked fine, so no API incompatibilities in JDK (foreign preview vs final): https://jenkins.thetaphi.de/job/Lucene-main-Linux/48322/consoleText

gautamworah96:
Hey @ChrisHegarty do we have an open issue for extending this to float vectors? 
If not, I can create one

ChrisHegarty:
> Hey @ChrisHegarty do we have an open issue for extending this to float vectors? If not, I can create one
> 
> Edit: nevermind, I see you had pointed this out earlier
> 
> > The scorer currently only operates on vectors with an element size of byte, since loading vector data from float[] (the fallback), is only supported in JDK 22. We can evaluate if and how to support floats separately. See https://bugs.openjdk.org/browse/JDK-8318678
> 
> Perhaps we can implement some logic to leverage fast dot product scoring for float[] if the runtime is >=22.

I filed the following issue to evaluate adding mode off-heap scoring.

uschindler:
> > Hey @ChrisHegarty do we have an open issue for extending this to float vectors? If not, I can create one
> > Edit: nevermind, I see you had pointed this out earlier
> > > The scorer currently only operates on vectors with an element size of byte, since loading vector data from float[] (the fallback), is only supported in JDK 22. We can evaluate if and how to support floats separately. See https://bugs.openjdk.org/browse/JDK-8318678
> > 
> > 
> > Perhaps we can implement some logic to leverage fast dot product scoring for float[] if the runtime is >=22.
> 
> I filed the following issue to evaluate adding mode off-heap scoring.

Hi,
if you map the MemorySegment slice into a ByteBuffer and add a Float view on top of it, it might work.

But I remember that you said using Buffers instead of MemorySegments suffered from bad performance. Because that wozuld have been my favoourite way for off-heap vector operations, because it would also work with old MMapDirectory and ByteBuffersDirectory.

Uwe

Issue #13340, opened 2024-05-02
IndexWriter loses track of parent field when index is empty


msokolov:
### Description

This test fails with ` java.lang.IllegalArgumentException: can't add a parent field to an already existing index without a parent field`. If you index any documents in the index, using the parent field or no, then the test passes.

```  
public void testParentFieldEmptyIndex() throws IOException {
    try (Directory dir = newMockDirectory()) {
      IndexWriterConfig iwc = new IndexWriterConfig(new MockAnalyzer(random()));
      iwc.setParentField("parent");
      try (IndexWriter writer = new IndexWriter(dir, iwc)) {
        writer.commit();
      }
      IndexWriterConfig iwc2 = new IndexWriterConfig(new MockAnalyzer(random()));
      iwc2.setParentField("parent");
      try (IndexWriter writer = new IndexWriter(dir, iwc2)) {
        writer.commit();
      }
    }
  }
```

msokolov:
I hope someone who is familiar with this will quickly see the problem - I'm not sure if we (1) maybe fail to write any FieldInfos when the index is empty, but now we must, or (2) fail to notice that the index we just opened has no FieldInfos and therefore can have a parent field added? It makes more sense to me that we should write a FieldInfos for an empty index.

msokolov:
I tried removing

```
    if (leaves.isEmpty()) {                                                                                                                                                                      
      return FieldInfos.EMPTY;                                                                                                                                                                   
    } else                                                                                                                                                                                       
 ```
from `FieldInfos.getMergedFieldInfos` but this didn't fix it -- although it also might be needed?

msokolov:
OK, now I see that field infos is part of the segment so we would not have written it. I guess in this case we should explicitly recognize and work around the empty case.

PR #13341, opened 2024-05-02
gh-13340: Allow adding a parent field to an index with no fields

msokolov:
Addresses #13340 . It relaxes the previous rule that adding a parent field to an existing index is *always* forbidden, allowing it in the case where no fields have been created. In such a case it will be safe, and this enables the trivial case of opening and closing a new empty index twice to work in a sensible way.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -1122,6 +1122,7 @@ public class IndexWriter
       globalFieldNumberMap = getFieldNumberMap();
       if (create == false
           && conf.getParentField() != null
+          && globalFieldNumberMap.getFieldNames().isEmpty() == false
           && globalFieldNumberMap.getFieldNames().contains(conf.getParentField()) == false) {
         throw new IllegalArgumentException(
             "can't add a parent field to an already existing index without a parent field");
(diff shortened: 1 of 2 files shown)


msokolov:
hm maybe this is not safe? If one creates a doc block filled with empty documents? I'm not sure ...

msokolov:
I'll push since there don't seem to be any concerns raised. If we later want to make the index metadata a first-class file on its own we can always do that.

kkrik-es:
Would it be possible to back-port this fix to v.9.11? I did hit this issue as well while trying to make use of #12829.

msokolov:
I don't think we can backport to 9.11 since it has already been released, but I will backport to 9.x and then it should get released with the next. There was some talk about a 9.11.1, so maybe it would be in that?

kkrik-es:
That would be nice, thanks. @jpountz @benwtrent fyi.

benwtrent:
@msokolov it might be worth doing a 9.11.1 release to fix this particular bug and maybe another other one: https://github.com/apache/lucene/pull/13475

I didn't know about this outstanding bug (parent field with empty index) when I originally did the 9.11 release, I should have investigated.

msokolov:
sigh, I would have backported right away except for some reason I thought the parent-field enforcement was only on 10.x not 9.x

Issue #13188, opened 2024-03-18
Support for building materialized views using Lucene formats


bharath-techie:
### Description

We are exploring the use case of building materialized views for certain fields and dimensions using [Star Tree index](https://github.com/opensearch-project/OpenSearch/issues/12498) while indexing the data. This will be based on the configured fields (dimensions and metrics) during index creation. This is inspired from http://hanj.cs.illinois.edu/pdf/vldb03_starcube.pdf and Apache Pinot’s Star Tree index. Star Tree helps to enforce upper bound on the aggregation queries ensuring predictable latency and resource usage, it is also storage space efficient and configurable.
OpenSearch RFC : https://github.com/opensearch-project/OpenSearch/issues/12498

Creating this issue to discuss approaches to support Star Tree in Lucene and also to get feedback on any other approaches/recommendations from the community.

## Quick overview on Star Tree index creation flow
The Star Tree DocValues fields and Star Tree index are created during the flush / merge flows of indexing

### Flush / merge flow 

1. Create initial set of Star Tree documents based on the configured dimensions and metrics.
2. Sort the Star Tree documents based on dimensions (fields) and aggregate on the metrics (fields).
3. Create Star Tree index.
4. Create Star Tree DocValues fields for each of the Star Tree dimensions and metrics

![star-lucene](https://github.com/apache/lucene/assets/58062316/13c80059-26f5-4c0f-b9e2-7dee8315c277)

### Challenges
Main challenge is that ‘StarTree’ index is a multi-field index compared to other formats in Lucene / OpenSearch. This makes it infeasible to use the PerField extension defined in Lucene today. We explored ‘BinaryDocValues’ to encode dimensions and metrics, but the ‘type’ of dimensions and metrics are different. So we couldn’t find a way to extend it. [Dimensions could be numeric or text or combination].

### Approach 1 - Create a new format to build materialized views
We can create a new dedicated file format (similar to points format, postings format) for materialized views which accepts list of dimensions and metrics and the default implementation for it could be the Star Tree index.

Pros

* This can be the standard format for supporting materialized views in Lucene which developers can use or extend to create custom solutions.
* With this format, users can directly create materialized views during index time without the storage of original documents via DocValues indices

Cons

* This will be a maintenance overhead.

#### Indexing - Extend DocValues to support materialized views
We can extend DocValues format to support a new type of field ‘AGGREGATED’ which will hold the configured list of dimensions and metrics by the user during index creation.
```
AggregatedField {
[...]

PR #13342, opened 2024-05-03
Datacube format changes to support materialized views

bharath-techie:
### Description
Draft PR for supporting extensions for DataCubes that mainly includes 
- New format 
- DataCubesConfig , DataCubeField, DataCubeFieldProvider 
- Changes in SegmentInfo to accomodate above
- Minor changes in Indexingchain flush , SegmentMerger

Related to https://github.com/apache/lucene/issues/13188


lucene/core/src/java/org/apache/lucene/index/SegmentInfo.java
```diff
@@ -124,7 +125,8 @@ public SegmentInfo(
       Map<String, String> diagnostics,
       byte[] id,
       Map<String, String> attributes,
-      Sort indexSort) {
+      Sort indexSort,
+      DataCubesConfig dataCubesConfig) {
```
Bukhtawar, line 129: Should we overload this ctor to avoid changes across the board, unless you think this should be mandatory

bharath-techie: Yeah will revert the second commit which removed the overload. I think overload is good , we can remove it if needed based on feedback.

PR #13346, opened 2024-05-06
Harden BaseDocValuesFormatTestCase

dnhatn:
We hit a [Codec bug](https://github.com/elastic/elasticsearch/pull/108276) in Elasticsearch, but it went unnoticed because our tests extend from BaseDocValuesFormatTestCase, which doesn't attempt to read the doc-values of the same document twice. This change strengthens BaseDocValuesFormatTestCase checks and randomly inserts that access pattern.

### DIFF
--- a/lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseDocValuesFormatTestCase.java
+++ b/lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseDocValuesFormatTestCase.java
@@ -1414,6 +1414,45 @@ public abstract class BaseDocValuesFormatTestCase extends BaseIndexFileFormatTes
     ir.close();
   }
 
+  protected void compareStoredFieldWithSortedNumericsDV(
+      DirectoryReader directoryReader, String storedField, String dvField) throws IOException {
+    for (LeafReaderContext leaf : directoryReader.leaves()) {
+      LeafReader reader = leaf.reader();
+      StoredFields storedFields = reader.storedFields();
+      SortedNumericDocValues docValues = reader.getSortedNumericDocValues(dvField);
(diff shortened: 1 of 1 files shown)


lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseDocValuesFormatTestCase.java
```diff
@@ -1414,6 +1414,45 @@ protected void assertDVIterate(Directory dir) throws IOException {
...
+        for (int r = 0; r < repeats; r++) {
+          if (r > 0 || random().nextBoolean()) {
+            assertTrue(docValues.advanceExact(doc));
+          }
+          for (int v = 0; v < docValues.docValueCount(); v++) {
+            assertEquals(storedValues[v], Long.toString(docValues.nextValue()));
+          }
+        }
+      }
+    }
```
jpountz, line 1453: This loop looks good to me, I'm contemplating doing a second loop using advanceExact every Nth document in case there are bugs lurking in case we ignore some documents?

jpountz (approved):
I left a suggestion but the change looks good to me, thank you!

jpountz (approved):

lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseDocValuesFormatTestCase.java
dnhatn: ++, I pushed https://github.com/apache/lucene/pull/13346/commits/59686c4692b5a6ee63f9f358b68ac923755d44b4

dnhatn:
Thanks @jpountz

Issue #13344, opened 2024-05-04
How to run tests with the Panama Vector implementation


ChrisHegarty:
I previously used to do:

```
export CI=true; ./gradlew :lucene:core:test
```

but this is not enough, in the output we see:
```
WARNING: Vector bitsize and/or integer vectors enforcement; using default vectorization provider outside of testMode
```

because either or both of `tests.vectorsize` or `tests.forceintegervectors` is set.

uschindler:
Could it be the ";" in your command line?

uschindler:
wah you're right. It's not the export.

The reason for that is; During testing we set arbitrary bit sizes, but this will slow you down immense so it disables the panama provider explicit. This was added by Robert at some point.

The current code is fine, but theres no way to run all tests with the default provider.....

Let me think about this. We have to find some way to execute tests "under real conditions".

ChrisHegarty:
Might be that we allow the default sizes/values to be set, so e.g. 128 bit on Mac, etc. without disabling the Panama provider.

uschindler:
I think there should be the option to run this with defaults.

uschindler:
I think we must do some magic to not set those: https://github.com/apache/lucene/blob/40cae087f71a875478309abfc4b1ab1e7b027cab/gradle/testing/randomization.gradle#L108-L113

This is all bad. The problem is that those settings are static and we can't have a per-instance seeting, so it affects both test mode and normal mode. As running all tests would be incredibly slow with wrong settings (non-default), we use default provider for all tests except those with use testMode.

uschindler:
see https://github.com/apache/lucene/pull/12681

uschindler:
> Those settings are really only useful when running the test which compares that the Panama and default implementations emit the same results. If we could somehow just limit setting them to that test!!

That's exactly the problem. According to Robert's testing, the static final constants were better for performance.

uschindler:
No idea yet, still thinking about it!

ChrisHegarty:
I really wanna spawn a new JVM with those options for a small set of tests, rather than have them infect all other running tests.

uschindler:
There are two possibilities how to do this: Spawn the tests that use testMode in a child VM or we add a separate sourceset with tests that run with hardcoded bitsizes.

Maybe for the few tests we can use one of the "child spawner" test examples which do this already.

ChrisHegarty:
I wonder if we can just avoid some of this complexity.

[...]

PR #13351, opened 2024-05-08
Add a separate option to allow running Panama Vectorization for all tests with suitable C2 defaults

ChrisHegarty:
This commit adds a separate option, tests.defaultvectorization, to allow running Panama Vectorization for all tests with suitable C2 defaults.

For example:
```
./gradlew :lucene:core:test -Ptests.defaultvectorization=true
```

A `default` value has also been added to the value in the random picks of vector bit size, which effectively enables the Panama Vector similarities implementation for all tests except TestVectorUtilSupport. This will cover more cases in the general workflow.

When modify or testing the Panama Vector similarities implementation directly, namely TestVectorUtilSupport, the following pattern continues to work:
```
for bits in default 128 256 512; do ./gradlew -p lucene/core test --tests TestVectorUtilSupport -Dtests.vectorsize=$bits; done
```

relates #13344

### DIFF
--- a/gradle/testing/randomization.gradle
+++ b/gradle/testing/randomization.gradle
@@ -112,8 +112,7 @@ allprojects {
           [propName: 'tests.forceintegervectors',
             value: { -> testsDefaultVectorizationRequested() ? false : (randomVectorSize != 'default') },
             description: "Forces use of integer vectors even when slow."],
-          [propName: 'tests.defaultvectorization',
-            value: { -> (randomVectorSize == 'default') },
+          [propName: 'tests.defaultvectorization', value: false,
             description: "Uses defaults for running tests with correct JVM settings to test Panama vectorization (tests.jvmargs, tests.vectorsize, tests.forceintegervectors)."],
       ]
     }
--- a/gradle/testing/randomization.gradle
+++ b/gradle/testing/randomization.gradle
@@ -106,15 +106,14 @@ allprojects {
           [propName: 'tests.bwcdir', value: null, description: "Data for backward-compatibility indexes."],
           // vectorization related
           [propName: 'tests.vectorsize',
-             value: { ->
-               isCIBuild ? "" : RandomPicks.randomFrom(new Random(projectSeedLong), ["", "128", "256", "512"])
-             },
-             description: "Sets preferred vector size in bits."],
+           value: { -> testsDefaultVectorizationRequested() ? 'default' :
+            RandomPicks.randomFrom(new Random(projectSeedLong), ["default", "128", "256", "512"]) },
+           description: "Sets preferred vector size in bits."],
           [propName: 'tests.forceintegervectors',
-             value: { ->
-                 isCIBuild ? false : true
-             },
-             description: "Forces use of integer vectors even when slow."],
+            value: { -> !testsDefaultVectorizationRequested() },
+            description: "Forces use of integer vectors even when slow."],
+          [propName: 'tests.defaultvectorization', value: false,
+            description: "Uses defaults for running tests with correct JVM settings to test Panama vectorization (tests.jvmargs, tests.vectorsize, tests.forceintegervectors)."],
       ]
     }
   }
@@ -176,6 +175,8 @@ allprojects {
         // and not the test JVM itself.
         systemProperties testOptionsResolved
 
+        jvmArgs Commandline.translateCommandline(testOptionsResolved['tests.jvmargs'])
+      
         if (Boolean.parseBoolean(testOptionsResolved['tests.asserts'])) {
           jvmArgs("-ea", "-esa")
         } else {
--- a/gradle/testing/randomization.gradle
+++ b/gradle/testing/randomization.gradle
@@ -63,6 +63,7 @@ allprojects {
 allprojects {
   plugins.withType(JavaPlugin) {
     ext {
+      String randomVectorSize = RandomPicks.randomFrom(new Random(projectSeedLong), ["default", "128", "256", "512"])
       testOptions += [
           // seed, repetition and amplification.
           [propName: 'tests.seed', value: { -> rootSeed }, description: "Sets the master randomization seed."],
@@ -106,11 +107,10 @@ allprojects {
           [propName: 'tests.bwcdir', value: null, description: "Data for backward-compatibility indexes."],
           // vectorization related
           [propName: 'tests.vectorsize',
-           value: { -> testsDefaultVectorizationRequested() ? 'default' :
-            RandomPicks.randomFrom(new Random(projectSeedLong), ["default", "128", "256", "512"]) },
+           value: { -> testsDefaultVectorizationRequested() ? 'default' : randomVectorSize },
            description: "Sets preferred vector size in bits."],
           [propName: 'tests.forceintegervectors',
-            value: { -> !testsDefaultVectorizationRequested() },
+            value: { -> testsDefaultVectorizationRequested() ? false : (randomVectorSize != 'default') },
             description: "Forces use of integer vectors even when slow."],
           [propName: 'tests.defaultvectorization', value: false,
             description: "Uses defaults for running tests with correct JVM settings to test Panama vectorization (tests.jvmargs, tests.vectorsize, tests.forceintegervectors)."],
--- a/gradle/testing/randomization.gradle
+++ b/gradle/testing/randomization.gradle
@@ -112,7 +112,8 @@ allprojects {
           [propName: 'tests.forceintegervectors',
             value: { -> testsDefaultVectorizationRequested() ? false : (randomVectorSize != 'default') },
             description: "Forces use of integer vectors even when slow."],
-          [propName: 'tests.defaultvectorization', value: false,
+          [propName: 'tests.defaultvectorization',
+            value: { -> (randomVectorSize == 'default') },
             description: "Uses defaults for running tests with correct JVM settings to test Panama vectorization (tests.jvmargs, tests.vectorsize, tests.forceintegervectors)."],
       ]
     }
--- a/gradle/testing/randomization.gradle
+++ b/gradle/testing/randomization.gradle
@@ -107,10 +107,14 @@ allprojects {
           // vectorization related
           [propName: 'tests.vectorsize',
              value: { ->
-               RandomPicks.randomFrom(new Random(projectSeedLong), ["128", "256", "512"])
+               isCIBuild ? "" : RandomPicks.randomFrom(new Random(projectSeedLong), ["", "128", "256", "512"])
              },
              description: "Sets preferred vector size in bits."],
-          [propName: 'tests.forceintegervectors', value: "true", description: "Forces use of integer vectors even when slow."],
+          [propName: 'tests.forceintegervectors',
+             value: { ->
+                 isCIBuild ? false : true
+             },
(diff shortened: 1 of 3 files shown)


gradle/testing/randomization.gradle
```diff
@@ -107,10 +107,14 @@ allprojects {
           // vectorization related
           [propName: 'tests.vectorsize',
              value: { ->
-               RandomPicks.randomFrom(new Random(projectSeedLong), ["128", "256", "512"])
+               isCIBuild ? "" : RandomPicks.randomFrom(new Random(projectSeedLong), ["", "128", "256", "512"])
```
dweiss, line 110: I wonder if something like "platform-default" as the constant wouldn't be nicer here. It can be an alias for "" internally, just gives more clarity on what's going on, especially when you run gradlew -p lucene/core testOpts.

dweiss: I don't mind it either way though.

uschindler: we can also set it to `null` which will not pass the property at all. Currently the code in VectorizationProvider looks like this: https://github.com/apache/lucene/blob/06973f5723e97de7b73a84b1ff3c8c35db79d7ec/lucene/core/src/java/org/apache/lucene/internal/vectorization/VectorizationProvider.java#L46-L57

We can change the "isEmpty" line.

uschindler:
I have a little bit of problem with that. Our CI infrastructure passes `CI=true` and basically we only want there that the "make tests fast" JVM option is removed (it removes the setting to disable C2 compiler). The disabled C2 compiler is actually the reason why we disable the vectorization tests for normal developers, because they would run insanely slow without C2.

But actually the CI is there to test all variants and it does not matter how slow it is. So actually the change here is contradictory to what CI should do.

So my idea would be to just add another test property `-Ptests.useProductiveMode=true` (or similar) which then implies:
- disable the "tiered stop at 1" setting from jvmargs.
- use default vectorization and do not enforce integer vectors

So that means: Any developer who wants to test the productive features of Lucene, needs to pass the property. This would slowdown tests a bit and vectorization is always used in the way how it would be in production.

Other ideas how to name this parameter:
- `-Ptests.useProductiveMode=true`
- `-Ptests.defaultVectorization=true`
- `-Ptests.enableIncubatorVectorization=true`

or any else. The changes are quite simple. We just need to make some extra code to override the other ones if one of the above is setup.

gradle/testing/randomization.gradle
uschindler: I changed this to locally:

```java
      vs =
          Stream.ofNullable(System.getProperty("tests.vectorsize"))
              .filter(Predicate.not(Set.of("", "default")::contains))
              .mapToInt(Integer::parseInt)
              .findAny();
```

uschindler:
I am trying to implement this, but actually there is some horrible code duplication in `test-defaults.gradle` and `randomization.gradle`. It resolves all test options 2 times and then it gets inconsistent... @dweiss 

Working on it.

Uwe

uschindler:
My problem is: `randomization.gradle` resolves the test options lazily in `afterEvaluate` and assigns them to `ext.testOptionsResolved`.

In contrast `defaults' test evaluates them non-lazily and assigns them to `resolvedTestOption`, very similar but different.

The jvmArges are then assigned early, but later any changes get lost (because the randomization one) decides to add everthing later.

Why is this?

uschindler:
Hrm, I am hopeless at moment. Need a bit of freetime to walk around. The issue is that jvmArgs are resolved early, while the test properties are done in randomization.gradle. This causes the issue that I cant undo the jvmArgs because they are added early to the config.

Need to rewrite that and add the JVM options late.

P.S.: Let's please merge the default-tests.gradle and randomization.gradle! This code is unreadable! All in one file would be much better.

dweiss:
> P.S.: Let's please merge the default-tests.gradle and randomization.gradle! This code is unreadable! All in one file would be much better.

I'm pretty sure there was some convoluted reason to keep them this way... But I'm not sure - it's been quite a while. I thinkmerging these files is a good idea - they do seem very closely related. I'm also fairly sure the after-evaluation trick was intentional but can't remember why.

uschindler:
I found a solution. It was a bit tricky, but works. Will commit to this branch.

uschindler:
ok, I pushed my changes:
- `CI=true` is only useful for CI and preserves old behaviour
- to enable default vectorization settings and enable Hotspot's C2 use `-Ptests.defaultvectorization=true` (we can change name of this property if you have a better name)
- you can still pass custom JVM options and can still enforce interger vectors if needed. The above settings just changes the efaults

The code is now quite clean, I did not yet cleanup the two source files. I only had to move the `jvmArgs` gradle test task setting to the `afterEvaluate`. I added a separate method to `project.ext` to evaluate the `tests.defaultvectorization`.

We should really merge those 2 files as they are really depending on each other. And maybe decide to resolve all test options late (in afterEvaluate).

uschindler:
Now let's go to vacation and before that have beers on German "Vatertag". Please backport those changes here, too as this bug also affects 9.x

uschindler (approved):

uschindler:
I added a bit of code to make the integer vector enforcement always `false` if the default vectorization settings are used.

lucene/core/src/java/org/apache/lucene/internal/vectorization/VectorizationProvider.java
```diff
@@ -47,7 +47,7 @@ public abstract class VectorizationProvider {
     try {
       vs =
           Stream.ofNullable(System.getProperty("tests.vectorsize"))
-              .filter(Predicate.not(String::isEmpty))
+              .filter(Predicate.not(Set.of("", "default")::contains))
```
ChrisHegarty, line 50: ++

uschindler:
I tested the combination by repeatedly executing: `gradlew :lucene:core:testOpts`, appending several parameters. All combinations seems right:
- if you pass `-Ptests.defaultvectorization=true`, all is default and enforcement is false. In addition the jvmargs are empty!
- if you pass nothing, random values for vectorssize and integer enforcement is enabled. if vectorsize is default, also integer enforcement is off

ChrisHegarty:
There is still one small issue. C2 will still be disabled if `default` is randomly selected, e.g.

```
$ ./gradlew :lucene:core:testOpts | egrep ".*defaultvectorization.*|.*vectorsize.*|.*forceintegervectors.*|.*jvmargs.*"
  tests.defaultvectorization = false    # Uses defaults for running tests with correct JVM settings to test Panama vectorization (tests.jvmargs, tests.vectorsize, tests.forceintegervectors).
C tests.forceintegervectors = false    # (!= default: computed) Forces use of integer vectors even when slow.
C tests.jvmargs            = -XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1 # (!= default: computed) Arguments passed to each forked JVM.
C tests.vectorsize         = default  # (!= default: computed) Sets preferred vector size in bits.
```

ChrisHegarty:
I updated the default setting of `tests.defaultvectorization` to include when `default` is randomly selected. E.g.

```
$ ./gradlew :lucene:core:testOpts | egrep ".*defaultvectorization.*|.*vectorsize.*|.*forceintegervectors.*|.*jvmargs.*"
C tests.defaultvectorization = true     # (!= default: computed) Uses defaults for running tests with correct JVM settings to test Panama vectorization (tests.jvmargs, tests.vectorsize, tests.forceintegervectors).
C tests.forceintegervectors = false    # (!= default: computed) Forces use of integer vectors even when slow.
C tests.jvmargs            =          # (!= default: computed) Arguments passed to each forked JVM.
C tests.vectorsize         = default  # (!= default: computed) Sets preferred vector size in bits.
```

ChrisHegarty:
>Actually that's wanted: Because we want tests by default run fast. For that the CI env var sets empty jvmargs. With your latest change we randomly slowdown the tests.

It's not clear whether tests will run faster when Panama Vector is enabled without C2 or not. I'll check this.

> So please revert your commit.

Happy to revert, if it turns out to be faster without it.

uschindler:
> > Actually that's wanted: Because we want tests by default run fast. For that the CI env var sets empty jvmargs. With your latest change we randomly slowdown the tests.
> 
> It's not clear whether tests will run faster when Panama Vector is enabled without C2 or not. I'll check this.
> 
> > So please revert your commit.
> 
> Happy to revert, if it turns out to be faster without it.

The reason for the jvmargs is to not enable C2 for our short running tests with lots of randomization. This causes dramatic overhead (Robert tested this) on total test runtime. So we use the default C2 JVM settings only on CI, where 50% longer total runtime does not matter.

Without that hack we could revert a lot of code.

uschindler:
See: https://github.com/apache/lucene/issues/10200

ChrisHegarty:
> The reason for the jvmargs is to not enable C2 for our short running tests with lots of randomization. This causes dramatic overhead (Robert tested this) on total test runtime. So we use the default C2 JVM settings only on CI, where 50% longer total runtime does not matter.

Yes, I get this. I was thinking that running vector tests with Panama Vector enabled and C2 disabled will be slow, but it seems to make little difference - I guess we don't have very stressful tests in this area. 

I'll revert my latest change, so that C2 will be disabled regardless of Panama Vector. Then merge and backport.

ChrisHegarty:
There is a problem with `default` - it does not do what you might expect it to do - it does not enable Panama Vector (unless you prevent C2 from being disabled). For example:

```
$ ./gradlew :lucene:core:test --tests TestKnnByteVectorQuery.testVectorEncodingMismatch -Ptests.vectorsize="default" -Ptests.forceintegervectors=false --info
...
org.apache.lucene.search.TestKnnByteVectorQuery > testVectorEncodingMismatch STANDARD_ERROR
    M5 09, 2024 3:26:42 KANG’AMA org.apache.lucene.internal.vectorization.VectorizationProvider lookup
    WARNING: C2 compiler is disabled; Java vector incubator API can't be enabled
...
```

The addition of `default` is opportunistic while we are here, but the fundamental value add of this PR is to add a way to more easily run with the "real world" vector similarities, which we can do with `tests.defaultvectorization`. Let's revert the addition of `default`. It's not adding any real value here.

uschindler:
Hi, good point.

This simplifies logic more. I would also like to move the jvmargs over to randomization.gradle.

uschindler:
We can still keep the default parser in the provider code. It will be used by the default vectorization argument.

uschindler:
If we remove default we should set the default value to null. Then it passes no sysprop at all.

uschindler:
Hm, actually when we keep "default" in the random pick, the CI builds sometimes use real conditions, so we should keep it!

Let's keep the current state.

ChrisHegarty:
> Hm, actually when we keep "default" in the random pick, the CI builds sometimes use real conditions, so we should keep it!

Agreed.  I'm declaring that this PR is done. I will merge and backport.  @uschindler now go on vacation and stop replying! ;-)

uschindler:
My last comment here: It would be good to somehow document this in the file so it is clear why all those combinations discussed here are "correct", although they seem to be wrong. After my vacation I will merge the `randomization.gradle` and `defaults-tests.gradle` files and look again at the lazy evaluation and document this more.

P.S.: Backport?

uschindler:
Thanks!

PR #13355, opened 2024-05-10
Fix default flat vector scorer supplier sharing backing array

ChrisHegarty:
This commit fixes an issue in the default flat vector scorer supplier whereby subsequent scorers created by the supplier can affect previously created scorers.

The issue is that we're sharing the backing array from the vector values, and overwriting it in subsequent scorers. We just need to use the ordinal to protect the scorer instance from mutation.

The test is a little verbose, but duplicates the issue and is extended in another PR to cover more flat vector scorer implementations.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/hnsw/DefaultFlatVectorScorer.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/hnsw/DefaultFlatVectorScorer.java
@@ -19,7 +19,6 @@ package org.apache.lucene.codecs.hnsw;
 
 import java.io.IOException;
 import org.apache.lucene.index.VectorSimilarityFunction;
-import org.apache.lucene.util.ArrayUtil;
 import org.apache.lucene.util.hnsw.RandomAccessVectorValues;
 import org.apache.lucene.util.hnsw.RandomVectorScorer;
 import org.apache.lucene.util.hnsw.RandomVectorScorerSupplier;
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/codecs/hnsw/DefaultFlatVectorScorer.java
```diff
@@ -101,7 +102,9 @@ private ByteScoringSupplier(
 
     @Override
     public RandomVectorScorer scorer(int ord) throws IOException {
-      return new ByteVectorScorer(vectors2, vectors1.vectorValue(ord), similarityFunction);
+      byte[] ba = vectors1.vectorValue(ord);
+      return new ByteVectorScorer(
+          vectors2, ArrayUtil.copyOfSubArray(ba, 0, ba.length), similarityFunction);
```
benwtrent, line 107: My concern is that now for every scorer, we are now creating a new array.

For example, in the HNSW graph builder, now instead of just creating a couple of backing arrays per thread, we create and require GC to collect a new one for every node added. This increases the garbage created by millions of arrays.

ChrisHegarty: Good catch. The prior implementation stored the old rather than the vector bytes. We should do the same here, which will avoid the copy. The scorers (not the supplier), do not suffer a similar issue, since their API doe not allow to score against anything other than the first vector.

benwtrent (approved):

PR #13356, opened 2024-05-10
Ensure negative scores aren not returned from scalar quantization scorer

benwtrent:
Depending on how we quantize and then scale, we can edge down below `0` for dotproduct scores. 

This is exceptionally rare, I have only seen it in extreme circumstances in tests (with random data and low dimensionality).

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -359,7 +359,7 @@ Bug Fixes
 
 * GITHUB#12966: Aggregation facets no longer assume that aggregation values are positive. (Stefan Vodita)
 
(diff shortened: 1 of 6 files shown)


ChrisHegarty (approved):
LGTM

lucene/CHANGES.txt
```diff
@@ -359,6 +359,8 @@ Bug Fixes
 
 * GITHUB#12966: Aggregation facets no longer assume that aggregation values are positive. (Stefan Vodita)
 
+* GITHUB#13356: Ensure negative scores aren not returned from scalar quantization scorer. (Ben Trent)
```
benwtrent, line 362:
```suggestion
* GITHUB#13356: Ensure negative scores are not returned from scalar quantization scorer. (Ben Trent)
```

PR #13358, opened 2024-05-10
Backport to 9x: Reduce duplication in taxonomy facets; always do counts #12966

stefanvodita:
Annoying things I had to do to backport #12966:
1. `TaxonomyFacetCounts` stores the counts twice because it extends `IntTaxonomyFacets`. The correct way would be to extend `TaxonomyFacets`, but I couldn't make that change, like I could for `FastTaxonomyFacets`, which was marked `@lucene.experimental`.
2. Increase visibility of `useHashTable` and `rollup` in `TaxonomyFacets` compared to `main`, from default to protected.
3. Leave `TopOrdAndInt/FloatQueue` as they were, since they are public API, and create new queues for our purposes,  `TopOrdAndInt/FloatNumberQueue`, marked deprecated because they will go away in 10.x.

### DIFF
--- a/gradle/testing/defaults-tests.gradle
+++ b/gradle/testing/defaults-tests.gradle
@@ -51,7 +51,7 @@ allprojects {
            includeInReproLine: false
           ],
           [propName: 'tests.jvmargs',
-           value: { -> propertyOrEnvOrDefault("tests.jvmargs", "TEST_JVM_ARGS", isCIBuild ? "" : "-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1") },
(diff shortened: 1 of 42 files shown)


stefanvodita:
Despite the "annoying" bits in the description, I don't expect this backport to be controversial, but reviews are welcome! I plan to wait over the weekend and then merge.

mikemccand (approved):
Thanks @stefanvodita!

Make sure to update the `main` `lucene/CHANGES.txt` also to move the entry down to 9.11.0 section.

stefanvodita:
Thank you for the review, Mike! I'd already put the CHANGES entries in 9.11 tentatively, now they're correct 😄

PR #13359, opened 2024-05-11
Use `IndexInput#prefetch` for terms dictionary lookups.

jpountz:
This introduces `TermsEnum#prepareSeekExact`, which essentially calls `IndexInput#prefetch` at the right offset for the given term. Then it takes advantage of the fact that `BooleanQuery` already calls `Weight#scorerSupplier` on all clauses, before later calling `ScorerSupplier#get` on all clauses. So `TermQuery` now calls `TermsEnum#prepareSeekExact` on `Weight#scorerSupplier` (if scores are not needed), which in-turn means that the I/O all terms dictionary lookups get parallelized across all term queries of a `BooleanQuery` on a given segment (intra-segment parallelism).


jpountz:
This is a draft as I need to do more work on tests and making sure that this new method cannot corrupt the state of the `SegmentTermsEnum`.

<details>
<summary>But I created a benchmark that starts looking like running a Lucene query that is encouraging. It creates an index with many terms that have very short postings lists, so that running boolean queries on these terms is heavy on terms dictionary lookups rather than reading postings. Then it manually runs a disjunction over 3 terms (some of these terms may not exist in the index as they are created randomly), computing how long it takes to evaluate all hits. To work properly when running a query, we'd need to move `#bulkScorer` from `Weight` to `ScorerSupplier`, which I intend to do as a follow-up.</summary>

```java
import java.io.IOException;
import java.nio.file.Path;
import java.nio.file.Paths;
import java.util.ArrayList;
import java.util.List;
import java.util.Random;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.atomic.AtomicLong;

import org.apache.lucene.document.Field.Store;
import org.apache.lucene.index.IndexWriter;
import org.apache.lucene.index.IndexWriterConfig;
import org.apache.lucene.index.Term;
import org.apache.lucene.index.TieredMergePolicy;
import org.apache.lucene.search.BooleanClause.Occur;
import org.apache.lucene.search.BooleanQuery;
import org.apache.lucene.search.DocIdSetIterator;
import org.apache.lucene.search.IndexSearcher;
import org.apache.lucene.search.ScoreMode;
import org.apache.lucene.search.Scorer;
import org.apache.lucene.search.ScorerSupplier;
import org.apache.lucene.search.TermQuery;
import org.apache.lucene.search.Weight;

public class TermsEnumPrefetchBench {

  public static int DUMMY;
  
    Path dirPath = Paths.get(args[0]);
    Directory dir = FSDirectory.open(dirPath);
    if (DirectoryReader.indexExists(dir) == false) {
      TieredMergePolicy mp = new TieredMergePolicy();
      mp.setSegmentsPerTier(100);
      mp.setMaxMergeAtOnce(100);
      mp.setMaxMergedSegmentMB(1024);
      try (IndexWriter w = new IndexWriter(dir, new IndexWriterConfig()
          .setMergePolicy(mp)
          .setRAMBufferSizeMB(1024))) {
        ExecutorService executor = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
        AtomicLong indexed = new AtomicLong(0);
        for (int task = 0; task < 1000; ++task) {
          executor.execute(() -> {
            for (int i = 0; i < 1_000; ++i) {
              for (int j = 0; j < 10_000; ++j) {
                doc.add(new StringField("f", Long.toString(r.nextLong(20_000_000_000L)), Store.NO));
              }
              try {
                w.addDocument(doc);
              } catch (IOException e) {
              }
              final long actualIndexed = indexed.incrementAndGet(); 
              if (actualIndexed % 10_000 == 0) {
              }
            }
          });
        }

        executor.shutdown();
        executor.awaitTermination(1, TimeUnit.DAYS);
        w.commit();
        System.out.println("Start force merging");
        w.forceMerge(1);
        System.out.println("Done force merging");
        w.commit();
      }
    }
    List<Long> latencies = new ArrayList<>();
    try (IndexReader reader = DirectoryReader.open(dir)) {
      IndexSearcher searcher = new IndexSearcher(reader);

      for (int i = 0; i < 10_000; ++i) {
        BooleanQuery.Builder query = new BooleanQuery.Builder();
        for (int t = 0; t < NUM_TERMS; ++t) {
          query.add(new TermQuery(new Term("f", Long.toString(r.nextLong(20_000_000_000L)))), Occur.SHOULD);
        }
        Weight weight = searcher.createWeight(searcher.rewrite(query.build()), ScoreMode.COMPLETE_NO_SCORES, 1f);
        ScorerSupplier ss = weight.scorerSupplier(reader.leaves().get(0));
        if (ss != null) {
          Scorer scorer = ss.get(Long.MAX_VALUE);
          DocIdSetIterator iter = scorer.iterator();
          for (int d = iter.nextDoc(); d != DocIdSetIterator.NO_MORE_DOCS; d = iter.nextDoc()) {
            DUMMY++;
          }
          long end = System.nanoTime();
          latencies.add((end - start) / 1000);
        }
      }
    }
    latencies.sort(null);
    System.out.println("P50: " + latencies.get(latencies.size() / 2));
    System.out.println("P90: " + latencies.get(latencies.size() * 9 / 10));
    System.out.println("P99: " + latencies.get(latencies.size() * 99 / 100));
  }

}
```
</details>

Without the change:
P50: 286
P90: 403
P99: 532

With the change:
P50: 148
P90: 246
P99: 368

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
```diff
@@ -307,6 +309,31 @@ private boolean setEOF() {
...
+    final long code =
+        fr.readVLongOutput(
+            new ByteArrayDataInput(
+                output.output.bytes, output.output.offset, output.output.length));
+    final long fpSeek = code >>> Lucene90BlockTreeTermsReader.OUTPUT_FLAGS_NUM_BITS;
+    initIndexInput();
+    final long fp = in.getFilePointer();
+    in.seek(fpSeek);
+    in.prefetch(1); // TODO: could we know the length of the block?
+    in.seek(fp); // TODO: do we actually need to do this?
```
rmuir, line 334: I really don't like these calls to `seek()` just to prefetch data. Since it is just prefetching, I'd prefer if this "dance" was an impl detail, if needed. 
It would make the code simpler to just pass parameter to prefetch rather than do this.

Then it is clear that the default implementation won't cause harm (unnecessary io) for any directory subclasses 

So I think prefetch should take location as argument? It is just a hint and not real i/o by the thread. It's intentionally not sequential and sequential API for it only hurts.

jpountz: I opened https://github.com/apache/lucene/pull/13363.

lucene/core/src/java/org/apache/lucene/index/TermsEnum.java
```diff
@@ -61,6 +62,15 @@ public enum SeekStatus {
...
   public abstract boolean seekExact(BytesRef text) throws IOException;
 
+  /**
+   * Prepare a future call to {@link #seekExact}. This typically calls {@link IndexInput#prefetch}
+   * on the right range of bytes under the hood so that the next call to {@link #seekExact} is
+   * faster. This can be used to parallelize I/O across multiple terms by calling {@link
+   * #prepareSeekExact} on multiple terms enums before calling {@link #seekExact(BytesRef)} on the
+   * same {@link TermsEnum}s.
+   */
+  public void prepareSeekExact(BytesRef text) throws IOException {}
```
rmuir, line 72: Can we look into subclasses such as `FilterLeafReader.FilterTermsEnum` to make sure this new method behaves correctly?

rmuir (changes requested):
whew terms stuff is always heavy, i left one concern

mikemccand:
> But I created a benchmark that starts looking like running a Lucene query that is encouraging

Was this with a forced-cold index?

jpountz:
It creates a 50GB terms dictionary while my machine only has ~28GB of RAM for the page cache, so many terms dictionary lookups result in page faults.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
```diff
@@ -307,6 +309,30 @@ private boolean setEOF() {
...
+    }
+
+    if (fr.size() == 0 || target.compareTo(fr.getMin()) < 0 || target.compareTo(fr.getMax()) > 0) {
+      return;
+    }
+
+    // TODO: should we try to reuse the current state of this terms enum when applicable?
+    BytesRefFSTEnum<BytesRef> indexEnum = new BytesRefFSTEnum<>(fr.index);
+    InputOutput<BytesRef> output = indexEnum.seekFloor(target);
+    if (output != null) { // should never be null since we already checked against fr.getMin()?
```
mikemccand, line 325: Indeed -- maybe change to `assert`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
```diff
@@ -307,6 +309,30 @@ private boolean setEOF() {
...
+      final long code =
+          fr.readVLongOutput(
+              new ByteArrayDataInput(
+                  output.output.bytes, output.output.offset, output.output.length));
+      final long fpSeek = code >>> Lucene90BlockTreeTermsReader.OUTPUT_FLAGS_NUM_BITS;
+      initIndexInput();
+      in.prefetch(fpSeek, 1); // TODO: could we know the length of the block?
```
mikemccand, line 332: I *think* you can do `indexEnum.next()` and if that is non-null (it will be null if you are on the very last block -- we could handle that case as well maybe) then get the fp for that next block and subtract the two?

jpountz: I tried this but this doesn't work as expected, it sometimes gives me blocks that have prior offsets and I'm not intimate enough with this terms dictionary to understand why, maybe you do?

mikemccand: Egads, that is really weird -- I would not expect those `fp` to go backwards on `.next()` -- I thought the FST index was a depth-first traversal of all on-disk (leaf) blocks.  I will need to mull some more about this :)

mikemccand: Maybe @vsop-479 has some insight?

mikemccand: Anyway, I don't think we need to hold up this nice PR for this -- we can try to improve this later.  PnP!

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -3754,13 +3754,17 @@ public static Status.TermVectorStatus testTermVectors(
...
+                int seekExactCounter = 0;
                 BytesRef term;
                 while ((term = termsEnum.next()) != null) {
 
                   // This is the term vectors:
                   postings = termsEnum.postings(postings, PostingsEnum.ALL);
                   assert postings != null;
 
+                  if ((seekExactCounter++ & 0xFF) == 0) {
+                    postingsTermsEnum.prepareSeekExact(term);
```
mikemccand, line 3766: Nice -- this is to make sure we are exercising the API?

lucene/core/src/java/org/apache/lucene/index/TermsEnum.java
```diff
@@ -61,6 +62,21 @@ public enum SeekStatus {
...
   public abstract boolean seekExact(BytesRef text) throws IOException;
 
+  /**
+   * Prepare a future call to {@link #seekExact}. This typically calls {@link IndexInput#prefetch}
+   * on the right range of bytes under the hood so that the next call to {@link #seekExact} is
+   * faster. This can be used to parallelize I/O across multiple terms by calling {@link
+   * #prepareSeekExact} on multiple terms enums before calling {@link #seekExact(BytesRef)} on the
+   * same {@link TermsEnum}s.
+   *
+   * <p><b>NOTE</b>: The terms enum is unpositioned after calling this method.
```
mikemccand, line 72: Hmm is it really that it is unpositioned, or, that this method does not alter the `TermsEnum`'s positioned state?

I.e. if I position it to some term, then call this method, won't it still be positioned on that same (prior) term?

Or are we trying to reserve the future right in the API to break the positioning, even though this first impl preserves it?

lucene/core/src/java/org/apache/lucene/index/TermsEnum.java
```diff
@@ -61,6 +62,21 @@ public enum SeekStatus {
...
+   *
+   * <p><b>NOTE</b>: It is not necessary to call this method before calling {@link
+   * #seekExact(BytesRef, TermState)}. {@link TermsEnum} implementations are expected to implement
+   * this method in an I/O-free fashion.
```
mikemccand, line 76: `this method` -> `that method`?  Since I think you mean `seekExact(BytesRef, TermState)` when you say `this method` here (but the previous `this method` two lines up is referring to `prepareSeekExact`?  Pronouns are hard!

lucene/core/src/java/org/apache/lucene/index/TermStates.java
```diff
@@ -172,22 +155,59 @@ public void accumulateStatistics(final int docFreq, final long totalTermFreq) {
...
+   * retrieve {@link Supplier}s across all required terms first before calling {@link Supplier#get}
+   * on all {@link Supplier}s so that the I/O for these terms can be performed in parallel.
    *
    * @param ctx the {@link LeafReaderContext} to get the {@link TermState} for.
-   * @return the {@link TermState} for the given readers ord or <code>null</code> if no {@link
-   *     TermState} for the reader was registered
+   * @return a Supplier for a TermState.
    */
-  public TermState get(LeafReaderContext ctx) throws IOException {
+  public Supplier<TermState> get(LeafReaderContext ctx) throws IOException {
```
mikemccand, line 169: This sort of reminds me of two-phase commit, except at read-time not write-time: we now break up these IO heavy read APIs into two phases, now, where step 1 is the intention to get X soon (allowing prefetch to happen, especially concurrently not just in the background of the calling thread, but, across the N different Xs we want to retrieve).  Step 2 is to then go and block on the IO to retrieve each of the N Xs.  Two phased reads!

lucene/core/src/java/org/apache/lucene/search/BlendedTermQuery.java
```diff
@@ -19,6 +19,7 @@
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.List;
+import java.util.function.Supplier;
```
mikemccand, line 22: Should we open a spinoff issue to maybe add prefetch to `TermInSetQuery` too?

lucene/core/src/java/org/apache/lucene/search/TermQuery.java
```diff
@@ -150,7 +170,12 @@ public Scorer get(long leadCost) throws IOException {
 
         @Override
         public long cost() {
-          return docFreq;
+          try {
+            TermsEnum te = getTermsEnum();
```
mikemccand, line 174: Hmm this getter got more costly.  It's too bad `TermState` is so opaque -- under the hood it (`BlockTermState`) is already storing `docFreq`.

mikemccand (commented):
Thanks @jpountz -- this is a nice change -- two phased reads!

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
jpountz: Correct.

lucene/core/src/java/org/apache/lucene/index/TermsEnum.java
jpountz: Indeed, I'm trying to reserve the right to update the state of the terms enum through this API in the future. See also `AssertingTermsEnum#prepareSeekExact`.

lucene/core/src/java/org/apache/lucene/index/TermsEnum.java
jpountz: English as a whole is hard. :) I'll fix.

lucene/core/src/java/org/apache/lucene/search/BlendedTermQuery.java
jpountz: Well, only `TermQuery` and `SynonymQuery` are handled so we could open an issue for every other query, I'm not sure we should open these issues? But indeed, let's think of how to take advantage of prefetching in `PointRangeQuery`, `TermInSetQuery`, `FeatureQuery`, etc.

lucene/core/src/java/org/apache/lucene/index/TermsEnum.java
jpountz: Hmm actually I mean `prepareSeekExact` when I say `this method`. I'll replace `this method` with `prepareSeekExact` to avoid ambiguity

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
vsop-479: > it sometimes gives me blocks that have prior offsets

> I would not expect those fp to go backwards on .next()

I haven't look into the whole context. 
But this will happens when we finished a `Block` and go back to read its parent block's `term`. 

e.g. 
We have terms like: "regular", "request1", "request2", "rest". Set minTermBlockSize to 2, maxTermBlockSize to 3. We will get blocks: b1: \["re"](root), b2: ["gular", "quest", "st"], b3: ["1", "2"].
Then as we call `next`, we will get the(`term`: `fp`) like: 
"regular": b2,
"request1": b3,
"request2": b3,
"rest": b2.

mikemccand: > But this will happens when we finished a `Block` and go back to read its parent block's `term`.

Yeah I agree that the `rest` term will be back in `b2` block, but, the `FSTEnum` we are talking about is the in-memory terms index that holds the file pointer offset to the start of these blocks ... I would have expected the FST to have e.g. in your example:

  * `reg -> b2`
  * `req -> b3`
  * `rez -> b4`

Or so, with `b4 > b3 > b2` block file pointers ...

mikemccand: OK!  This was nagging at me so I dug into it, printing the `FSTEnum` iteration output on a nightly Lucene benchy index ... I now understand why the pointers indeed go backwards.  I think this was the point you were making above @vsop-479 -- sorry I misunderstood at first ;)

It's because when writing the blocks we write "bottoms up" on depth first traversal through the terms, and only write a node when it is finished / returned from.  Leaf blocks will be written immediately / in order since they are started, terms come out, finished.  But for a non-leaf blocks, first all leaf blocks under them are written (in order), and THEN the non-leaf block is written only when we are done with all those recursions and writing any straggler terms that live in the non-leaf block.

But the prefixes are added to the index FST in the correct (term sorted) order.  So this means the file pointer can indeed go backwards when iterating the terms index.  I'll mull some more about whether we could (efficiently) know the term block length ...

jpountz:
<details>
<summary>
Now that #13408 has been merged, I could update the benchmark to simply call IndexSearcher#search.
</summary>

```java
import java.io.IOException;
import java.nio.file.Path;
import java.nio.file.Paths;
import java.util.ArrayList;
import java.util.List;
import java.util.Random;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.atomic.AtomicLong;

import org.apache.lucene.document.Field.Store;
import org.apache.lucene.index.IndexWriter;
import org.apache.lucene.index.IndexWriterConfig;
import org.apache.lucene.index.Term;
import org.apache.lucene.index.TieredMergePolicy;
import org.apache.lucene.search.BooleanClause.Occur;
import org.apache.lucene.search.BooleanQuery;
import org.apache.lucene.search.IndexSearcher;
import org.apache.lucene.search.Sort;
import org.apache.lucene.search.TermQuery;

public class TermsEnumPrefetchBench {

  public static int DUMMY;
  
    Path dirPath = Paths.get(args[0]);
    Directory dir = FSDirectory.open(dirPath);
    if (DirectoryReader.indexExists(dir) == false) {
      TieredMergePolicy mp = new TieredMergePolicy();
      mp.setSegmentsPerTier(100);
      mp.setMaxMergeAtOnce(100);
      mp.setMaxMergedSegmentMB(1024);
      try (IndexWriter w = new IndexWriter(dir, new IndexWriterConfig()
          .setMergePolicy(mp)
          .setRAMBufferSizeMB(1024))) {
        ExecutorService executor = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors());
        AtomicLong indexed = new AtomicLong(0);
        for (int task = 0; task < 1000; ++task) {
          executor.execute(() -> {
            for (int i = 0; i < 1_000; ++i) {
              for (int j = 0; j < 10_000; ++j) {
                doc.add(new StringField("f", Long.toString(r.nextLong(20_000_000_000L)), Store.NO));
              }
              try {
                w.addDocument(doc);
              } catch (IOException e) {
              }
              final long actualIndexed = indexed.incrementAndGet(); 
              if (actualIndexed % 10_000 == 0) {
              }
            }
          });
        }

        executor.shutdown();
        executor.awaitTermination(1, TimeUnit.DAYS);
        w.commit();
        System.out.println("Start force merging");
        w.forceMerge(1);
        System.out.println("Done force merging");
        w.commit();
      }
    }
    List<Long> latencies = new ArrayList<>();
    try (IndexReader reader = DirectoryReader.open(dir)) {
      IndexSearcher searcher = new IndexSearcher(reader);

      for (int i = 0; i < 10_000; ++i) {
        BooleanQuery.Builder query = new BooleanQuery.Builder();
        for (int t = 0; t < NUM_TERMS; ++t) {
          query.add(new TermQuery(new Term("f", Long.toString(r.nextLong(20_000_000_000L)))), Occur.SHOULD);
        }
        DUMMY += searcher.search(query.build(), 1, Sort.INDEXORDER).totalHits.value;
        long end = System.nanoTime();
        latencies.add((end - start) / 1000);
      }
    }
    latencies.sort(null);
    System.out.println("P50: " + latencies.get(latencies.size() / 2));
    System.out.println("P90: " + latencies.get(latencies.size() * 9 / 10));
    System.out.println("P99: " + latencies.get(latencies.size() * 99 / 100));
  }

}

```

</details>

Results still look good.

Before the change:
P50: 282
P90: 387
P99: 537

After the change:
P50: 161
P90: 253
P99: 379

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
mikemccand: > > But for a non-leaf blocks, first all leaf blocks under them are written (in order), and THEN the non-leaf block is written only when we are done with all those recursions and writing any straggler terms that live in the non-leaf block.
> 
> This means if we subtract the `fp` of a non-leaf block and its next, we will get its sub blocks' total length?

It's tricky.  I think if you do that, you'll get the total length of the next's sub blocks total length?  Because each non-leaf block is written at the end of the recursive (depth first) visit of all of its sub blocks.

I'm still not sure how to cleanly/efficiently get the total bytes length of a leaf block by looking solely at the FST terms index.  So we should proceed with the hint as is (pre-fetch 1 byte from position X) -- "typically" the terms block will fit into a single IO page (512 or 4096 bytes) and any further readahead the IO system does.

jpountz:
I pushed a new approach. Instead of `prepareSeekExact` returning `void`, it now returns a `Supplier` and forbids calling any other method on `TermsEnum` until the `Supplier` has been consumed. There are two benefits:
 - Implementations can early exit when they can figure out that the term doesn't exist without doing any I/O by returning `null`. In turn, this saves creating scorer on other required clauses of the same query.
 - It is easier to split the work between the first and second phases without duplicating work.
However, there is one downside, which is that you can only do one background I/O request per `TermsEnum` instance, so you need multiple `TermsEnum` instances if you want more I/O concurrency than that. This looks like a better trade-off to me, it only makes things like `PKLookup` a bit more sophisticated if they want to do I/O concurrency.

The benchmark still reports similar numbers:

Without the change
```
P50: 307
P90: 423
P99: 585
```

With the change
```
P50: 162
P90: 258
P99: 405
```

jpountz:
I will merge soon if there are no objections.

PR #13362, opened 2024-05-12
Add sub query explanations in DisjunctionMaxQuery#explain on no-match

### LINKED ISSUE #13357: Enhance DisjunctionMaxQuery explanation to include details in case there was no match
### Description

Currently, if the DisjunctionMaxQuery is used and the document did not match, the explanation from DisjunctionMaxWeight#explain only says `No matching clause` with no additional details (and the explanations from the subqueries are discarded).

In some use cases, it may be helpful to include individual "no-match" explanations from subqueries as they may provide more specific reason for why the document did not match.

timgrein:
### Description

https://github.com/apache/lucene/issues/13357 states that it's useful to have the explanations of the sub queries of `DisjuncationMaxQuery` present, even if the document didn't match. Considering that other queries like `CoveringQuery` also include [explanations, if the document didn't match](https://github.com/apache/lucene/blob/25f1efd8eb34c2eb135a3c0850723ca5b03deec7/lucene/sandbox/src/java/org/apache/lucene/sandbox/search/CoveringQuery.java#L218-L223) I've adjusted the logic according to the issue's proposal.

Also added tests for `explain` (match and no match case). I've also adjusted `CheckHits#verifyExplanation` to only check sub explanations, if the overall explanation returned a hit and scores need to be checked.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -293,6 +293,8 @@ Improvements
 * GITHUB#12966: Move most of the responsibility from TaxonomyFacets implementations to TaxonomyFacets itself.
   This reduces code duplication and enables future development. (Stefan Vodita)
 
+* GITHUB#13362: Add sub query explanations to DisjuncationMaxQuery, if the overall query didn't match. (Tim Grein)
+
 Optimizations
 ---------------------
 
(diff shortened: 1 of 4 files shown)


jpountz:
This could make explanations harder to read for large queries, e.g. queries produced through rewriting. I wonder about doing something in-between such as only including the non-matching subs if the query doesn't match, what do you think?

tteofili (approved):

jpountz (approved):
LGTM. Can you add a CHANGES entry under 9.11 / Improvements?

PR #13363, opened 2024-05-13
Make `IndexInput#prefetch` take an offset.

jpountz:
This makes `IndexInput#prefetch` take an offset instead of being relative to the current position. This avoids requiring callers to seek only to call `prefetch()`.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/IndexInput.java
+++ b/lucene/core/src/java/org/apache/lucene/store/IndexInput.java
@@ -199,7 +199,8 @@ public abstract class IndexInput extends DataInput implements Closeable {
    *
    * <p>The default implementation is a no-op.
    *
+   * @param offset start offset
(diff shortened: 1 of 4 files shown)


rmuir:
I like this much better from api perspective: it closer maps to `madvise()` and to me is more straightforward. Especially as the current PRs out there to use it (terms and postings) are using absolute FPs and seeking back-n-forth to workaround the current API.

rmuir (approved):

ChrisHegarty:
++ this absolute addressing is much nicer. We could even add `prefetch` to the RandomAccessInput interface, though I'm not sure how much that is used.

uschindler (commented):
Quick review after coming back from vacation: looks fine and seems to work.

I only don't like the we now have absolute APIs in a stream-like interface. So my choice for this API would have been in RandomAccessInput. But others seem to want to have it like that.

I don't care!

PR #13364, opened 2024-05-13
Use IndexInput#prefetch for postings, skip data and impacts

jpountz:
This uses the `IndexInput#prefetch` API for postings. This relies on heuristics, as we don't know ahead of time what data we will need from a postings list:
 - Postings lists are prefetched entirely when they are short (< 16kB).
 - Impacts enums also prefetch the first page of skip data.
 - Postings enums prefetc skip data on the first call to advance().

Positions, offsets and payloads are never prefetched.

Putting the `IndexInput#prefetch` call in `TermsEnum#postings` and `TermsEnum#impacts` works well because `BooleanQuery` will first create postings/impacts enums for all clauses before it starts unioning/intersecting them. This allows the prefetching logic to run in parallel across all clauses of the same query on the same segment.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99PostingsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99PostingsReader.java
@@ -917,7 +917,7 @@ public final class Lucene99PostingsReader extends PostingsReaderBase {
           payloadByteUpto = skipper.getPayloadByteUpto();
         }
         nextSkipDoc = skipper.getNextSkipDoc();
-      } else if (skipOffset >= 0) {
+      } else {
         // We don't need skip data yet, but we have evidence that advance() is used, so prefetch it
         // in the background.
         if (prefetchedSkipData == false) {
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99PostingsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99PostingsReader.java
@@ -2087,15 +2087,15 @@ public final class Lucene99PostingsReader extends PostingsReaderBase {
       if (state.skipOffset < 0) {
         // This postings list is very short as it doesn't have skip data, prefetch the page that
         // holds the first byte of the postings list.
-        docIn.prefetch(1);
+        docIn.prefetch(state.docStartFP, 1);
       } else if (state.skipOffset <= MAX_POSTINGS_SIZE_FOR_FULL_PREFETCH) {
         // This postings list is short as it fits on a few pages, prefetch it all, plus one byte to
         // make sure to include some skip data.
-        docIn.prefetch(state.skipOffset + 1);
+        docIn.prefetch(state.docStartFP, state.skipOffset + 1);
       } else {
         // Default case: prefetch the page that holds the first byte of postings. We'll prefetch
         // skip data when we have evidence that it is used.
-        docIn.prefetch(1);
+        docIn.prefetch(state.docStartFP, 1);
       }
     }
     // Note: we don't prefetch positions or offsets, which are less likely to be needed.
@@ -2106,10 +2106,7 @@ public final class Lucene99PostingsReader extends PostingsReaderBase {
     if (skipOffset > MAX_POSTINGS_SIZE_FOR_FULL_PREFETCH) {
       // If skipOffset is less than this value, skip data was already prefetched when doing
       // #seekAndPrefetchPostings
-      long fp = docIn.getFilePointer();
-      docIn.seek(docStartFP + skipOffset);
-      docIn.prefetch(1);
-      docIn.seek(fp);
+      docIn.prefetch(docStartFP + skipOffset, 1);
     }
   }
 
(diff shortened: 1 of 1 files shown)


rmuir:
Can we do better than blindly prefetching skipdata? Currently, skipdata is not used until we advance() past doc in the first block: 

https://github.com/apache/lucene/blob/83db8dfba30447dc154efbf98a46048ebe5743ef/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99PostingsReader.java#L409

Should we instead try to preload skipdata the first time advance() is actually called for the enum? This still happens before skipper might get used, because we might fullfill a few calls to advance() from the first block's docs, but I like that we have evidence now that skipper will most likely be needed. e.g. a one time invocation before this "if": https://github.com/apache/lucene/blob/83db8dfba30447dc154efbf98a46048ebe5743ef/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99PostingsReader.java#L502-L504

mikemccand:
We might (eventually, later) consider an API change when pulling postings that expresses that the caller intends to advance?  Or, maybe simpler would be expressing that caller will not use advance, not sure.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99PostingsReader.java
```diff
@@ -902,6 +917,12 @@ public int advance(int target) throws IOException {
           payloadByteUpto = skipper.getPayloadByteUpto();
         }
         nextSkipDoc = skipper.getNextSkipDoc();
+      } else {
+        // We don't need skip data yet, but we have evidence that advance() is used, so prefetch it
+        // in the background.
+        if (prefetchedSkipData == false) {
+          prefetchSkipData(docIn, docTermStartFP, skipOffset);
+        }
```
mikemccand, line 925: Should you set `prefectedSkipData = true` in here?

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99PostingsReader.java
```diff
@@ -1097,7 +1118,9 @@ public BlockImpactsDocsEnum(FieldInfo fieldInfo, IntBlockTermState termState)
       this.docIn = Lucene99PostingsReader.this.docIn.clone();
 
       docFreq = termState.docFreq;
-      docIn.seek(termState.docStartFP);
+      seekAndPrefetchPostings(docIn, termState);
+      // Impacts almost certainly need skip data
+      prefetchSkipData(docIn, termState.docStartFP, termState.skipOffset);
```
mikemccand, line 1123: Hmm also here?  Or maybe fix `prefetchSkipData` to set `prefetchedSkipData = true`?

mikemccand:
This is cool!  In the hot case, do we expect `prefetch` to be a no-op?  So we are hoping for "first do no harm" in that case?  (I haven't looked at `MMapDirectory`'s impl yet).  But in the cold case, this could cause multiple prefetch calls within one segment if the query has multiple terms.  And if cross-segment concurrency is enabled, multiple such calls across slices (thread work units) too.

But in the cold case this may be a big win?

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99PostingsReader.java
jpountz: Ah yes! This is what happens when you try to do something too quickly before signing off!

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99PostingsReader.java
jpountz: No need here since we're calling it in `reset()` which is only called once per postings enum. We only track `prefetchedSkipData` in `PostingsEnum` impls where the prefetching is done as part of `advance()`.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99PostingsReader.java
```diff
@@ -2049,6 +2074,44 @@ public long cost() {
...
+      // docs is a good start already.
+      docIn.seek(state.docStartFP);
+      if (state.skipOffset < 0) {
+        // This postings list is very short as it doesn't have skip data, prefetch the page that
+        // holds the first byte of the postings list.
+        docIn.prefetch(1);
+      } else if (state.skipOffset <= MAX_POSTINGS_SIZE_FOR_FULL_PREFETCH) {
+        // This postings list is short as it fits on a few pages, prefetch it all, plus one byte to
+        // make sure to include some skip data.
+        docIn.prefetch(state.skipOffset + 1);
```
rmuir, line 2093: I'm still trying to wrap my head around this `<= MAX_POSTINGS_SIZE_FOR_FULL_PREFETCH` case.

if the postings are short enough that we are willing to fault them all in at once, why do we even index skip data at all?

jpountz: This is trying to address your concern about the number of system calls by doing a single system call when the postings list is short instead of independently doing a madvise call for postings and skip data. Especially as short postings are less likely to amortize the cost of system calls as iterating all docs is fast CPU-wise?

When there are multiple clauses in the same query, usually the dense clauses would consume a small percentage of their number of docs while the sparser clauses would consume a large percentage of their number of docs, likely hitting all pages. So faulting all pages made sense to me but I don't have a strong feeling either and I'm happy to look at different approaches.

> if the postings are short enough that we are willing to fault them all in at once, why do we even index skip data at all?

You still get the CPU savings in the case when data fits in the page cache. Plus skip data also records impacts, and there are many cases when having impacts on the sparser clauses is important to be able to skip more hits on the denser clauses.

rmuir (approved):

PR #13366, opened 2024-05-13
Protect against nan & inf values in quantizer and test with tiny vectors

benwtrent:
In testing, it was noticed that tiny vectors (<= 2 dims) could cause weird behavior in the quantizer.

Additionally, we should fail any quantization that ends up with NaN or Inf values.


ChrisHegarty (approved):
Thanks @benwtrent LGTM

PR #13368, opened 2024-05-14
Replace Map<Integer, Object> by primitive IntObjectHashMap.

bruno-roustant:
Also replace some Map<Integer, Integer> by IntIntHashMap, if they don't rely on null value.

The goal is to gain globally some memory, maybe some perf on some spots that call the map intensively, with a replacement that does not seem to bring complexity. Most of the time it consists in changing a field type to the primitive map, and the call to the map constructor.
If some areas shouldn't be modified, we can exclude them from the replacement.

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/SuggestibleEntryCache.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/SuggestibleEntryCache.java
@@ -16,12 +16,11 @@
  */
 package org.apache.lucene.analysis.hunspell;
 
-import java.util.HashMap;
-import java.util.Map;
 import java.util.function.Consumer;
 import org.apache.lucene.util.ArrayUtil;
 import org.apache.lucene.util.CharsRef;
 import org.apache.lucene.util.IntsRef;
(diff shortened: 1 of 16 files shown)


lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/SuggestibleEntryCache.java
```diff
@@ -48,22 +45,33 @@ private SuggestibleEntryCache(Map<Integer, SectionBuilder> builders) {
...
             CharsRef root = entry.root();
             if (root.length > Short.MAX_VALUE) {
               throw new UnsupportedOperationException(
                   "Too long dictionary entry, please report this to dev@lucene.apache.org");
+            } else if (root.length > maxLength) {
+              maxLength = root.length;
             }
 
-            builders.computeIfAbsent(root.length, __ -> new SectionBuilder()).add(entry);
+            SectionBuilder builder;
```
bruno-roustant, line 61: This is the only spot where the replacement brings some complexity. It does not seem too complex to me, so I kept it.

dweiss (approved):

PR #13369, opened 2024-05-14
Fix weird NRT bug #13353

### LINKED ISSUE #13353: NRT failure due to FieldInfo & File mismatch
### Description

There has been a nasty test failure in ES for awhile: https://github.com/elastic/elasticsearch/issues/105122

The test simulates a document indexing failure. It turns out, that this test failure is caused by a series of strange conditions in Lucene. If we fail on indexing a field, but have points value field that comes AFTER the field that is indexing, things will blow up when opening a reader if the writer has soft-deletes enabled. 

The failure description is as follows:

 - First, we have an IndexWriter configured with soft-deletes & no commits on closing
 - Index a document with fields as follows ["field that will throw, "nice point field"]
 - We update the FieldInfos eagerly here: https://github.com/apache/lucene/blob/0aa88910ca9a1032d288996d14203eac4953f2de/lucene/core/src/java/org/apache/lucene/index/IndexingChain.java#L592-L603
 - FieldInfos now indicate we have a point field
 - The field that will throw is handled, document indexing fails. Since this is a regular text field, it does not automatically close the indexer
 - A NRT reader is opened on the writer and attempts to flush, but the field info is incorrect given the fields that are there with soft-delete (e.g. don't delete the segment)

<details>

<summary> Test that replicates the failure </summary>

```java

  public void testExceptionJustBeforeFlushWithPointValues() throws Exception {
    Analyzer analyzer =
        new Analyzer(Analyzer.PER_FIELD_REUSE_STRATEGY) {
          @Override
          public TokenStreamComponents createComponents(String fieldName) {
            MockTokenizer tokenizer = new MockTokenizer(MockTokenizer.WHITESPACE, false);
            tokenizer.setEnableChecks(
                false); // disable workflow checking as we forcefully close() in exceptional cases.
            TokenStream stream = new CrashingFilter(fieldName, tokenizer);
            return new TokenStreamComponents(tokenizer, stream);
          }
        };
    DirectoryReader r = null;
    IndexWriterConfig iwc =
        newIndexWriterConfig(analyzer).setCommitOnClose(false).setMaxBufferedDocs(3);
    MergePolicy mp = iwc.getMergePolicy();
    iwc.setMergePolicy(
        new SoftDeletesRetentionMergePolicy("soft_delete", MatchAllDocsQuery::new, mp));
    IndexWriter w = RandomIndexWriter.mockIndexWriter(dir, iwc, random());
    Document newdoc = new Document();
    newdoc.add(newTextField("crash", "do it on token 4", Field.Store.NO));
    newdoc.add(new IntPoint("int", 17));
    expectThrows(IOException.class, () -> w.addDocument(newdoc));
    try {
      r = w.getReader(false, false);
    } catch (AlreadyClosedException ace) {
      // expected
    }
    dir.close();
  }
```

</details>

The exception thrown is:

```
        Caused by:
        java.io.FileNotFoundException: No sub-file with id .kdi found in compound file "_0.cfs" (fileName=_0.kdi files: [_Lucene99_0.tip, .nvm, .fnm, .tvd, _Lucene99_0.doc, _Lucene99_0.tim, _Lucene99_0.pos, .tvm, _Lucene99_0.tmd, .fdm, .nvd, .fdx, .tvx, .fdt])
            at org.apache.lucene.codecs.lucene90.Lucene90CompoundReader.openInput(Lucene90CompoundReader.java:170)
            at org.apache.lucene.codecs.lucene90.Lucene90PointsReader.<init>(Lucene90PointsReader.java:63)
            at org.apache.lucene.codecs.lucene90.Lucene90PointsFormat.fieldsReader(Lucene90PointsFormat.java:74)
            at org.apache.lucene.index.SegmentCoreReaders.<init>(SegmentCoreReaders.java:152)
            ... 55 more
```

_No response_

benwtrent:
The issue outlines the problem. When we have point value dimensions, segment core readers assume that there will be point files. 

However, when allowing soft deletes and a document fails indexing failed before a point field could be written, this assumption fails. Consequently, the NRT fails to open. 

I tried many different ways of fixing this issue.

 - Delaying on setting the field info until document writes are successful. But this causes weirdness with global field numbers and other assumptions and checks later in the document life-cycle
 - Reverting the field info if a all the docs for a field fail to index, but this causes other weirdness and edge cases that seemed too sharp to mess with.

So, I settled on always flushing a point file if the field info says there are point fields, even if there aren't any docs in the buffer.

I am happy to consider other options.

closes https://github.com/apache/lucene/issues/13353

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -365,6 +365,9 @@ Bug Fixes
 
 * GITHUB#13366: Disallow NaN and Inf values in scalar quantization and better handle extreme cases. (Ben Trent)
 
+* GITHUB#13369: Fix NRT opening failure when soft deletes are enabled and the document fails to index before a point
+  field is written (Ben Trent)
+
 Build
 ---------------------
 
--- a/lucene/core/src/java/org/apache/lucene/index/IndexingChain.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexingChain.java
@@ -375,7 +375,7 @@ final class IndexingChain implements Accountable {
         while (perField != null) {
           if (perField.pointValuesWriter != null) {
             // We could have initialized pointValuesWriter, but failed to write even a single doc
-            if (perField.pointValuesWriter.getNumDocs() > 0) {
+            if (perField.fieldInfo.getPointDimensionCount() > 0) {
               if (pointsWriter == null) {
                 // lazy init
                 PointsFormat fmt = state.segmentInfo.getCodec().pointsFormat();
--- a/lucene/core/src/test/org/apache/lucene/index/TestIndexWriterExceptions.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestIndexWriterExceptions.java
@@ -47,6 +47,7 @@ import org.apache.lucene.document.StringField;
 import org.apache.lucene.document.TextField;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.MatchAllDocsQuery;
 import org.apache.lucene.search.PhraseQuery;
 import org.apache.lucene.store.AlreadyClosedException;
 import org.apache.lucene.store.ByteBuffersDirectory;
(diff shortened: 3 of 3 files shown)


benwtrent:
Pinging some folks on review. This is a weird place in the codebase I haven't messed with before. This deserves careful review. Checks are green, but I may have missed some corners.

jpountz (approved):
Your approach makes sense to me, we should use field infos to decide whether to flush, not whether we buffered docs. This is consistent with what happens when all values get merged away, and with what happens when we flush postings, norms or doc values.

PR #13370, opened 2024-05-15
Add a double addressing vector scorer

ChrisHegarty:
This commit adds a method to RandomVectorScorerSupplier that allows to score two vectors based their ordinals.

The existing model of this API first creates a scorer, that effectively binds the ordinal of the first vector, to then score the ordinal of a second vector agains the first. This results in a RandomVectorScorer instance being created each time we want to score against a different vector in the first position. Allowing to score against two given ordinals avoids the creation of a RandomVectorScorer instance, which is likely expensive during graph building.

The new API could seen as a convenience for `scorer(ord).score(node)`, or vice versa, but in fact there can be very different implementation characteristics of these - most notably the avoidance of the creation of a RandomVectorScorer instance.

This PR just updates a few places in the graph building to use the new API. Further analysis can determine where else this API would be beneficial.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/hnsw/DefaultFlatVectorScorer.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/hnsw/DefaultFlatVectorScorer.java
@@ -110,6 +110,12 @@ public class DefaultFlatVectorScorer implements FlatVectorsScorer {
     }
 
     @Override
+    public float score(int firstOrd, int secondOrd) throws IOException {
+      return similarityFunction.compare(
+          vectors1.vectorValue(firstOrd), vectors2.vectorValue(secondOrd));
+    }
+
+    @Override
     public RandomVectorScorerSupplier copy() throws IOException {
       return new ByteScoringSupplier(vectors, similarityFunction);
     }
@@ -139,6 +145,12 @@ public class DefaultFlatVectorScorer implements FlatVectorsScorer {
           return similarityFunction.compare(vectors1.vectorValue(ord), vectors2.vectorValue(node));
         }
       };
+    }
+
+    @Override
+    public float score(int firstOrd, int secondOrd) throws IOException {
+      return similarityFunction.compare(
+          vectors1.vectorValue(firstOrd), vectors2.vectorValue(secondOrd));
     }
 
     @Override
--- a/lucene/core/src/java/org/apache/lucene/codecs/hnsw/ScalarQuantizedVectorScorer.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/hnsw/ScalarQuantizedVectorScorer.java
@@ -125,28 +125,32 @@ public class ScalarQuantizedVectorScorer implements FlatVectorsScorer {
       implements RandomVectorScorerSupplier {
 
     private final RandomAccessQuantizedByteVectorValues values;
+    private final RandomAccessQuantizedByteVectorValues values1;
     private final ScalarQuantizedVectorSimilarity similarity;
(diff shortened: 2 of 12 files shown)


lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99ScalarQuantizedVectorScorer.java
```diff
@@ -291,6 +291,11 @@ public RandomVectorScorer scorer(int ord) throws IOException {
           values2);
     }
 
+    @Override
+    public float score(int firstOrd, int secondOrd) throws IOException {
+      return scorer(firstOrd).score(secondOrd);
```
ChrisHegarty, line 296: we can probably do better here, but this is good enough for now.

benwtrent (commented):
One thing the Scorer object gave us is caching of the single vector that is used many times.

The underlying Offheap vector objects cache the vector on heap and prevents multiple reads.

Anything that calls the same random vectors twice with different ordinals, and those random vectors are read on heap will suffer significant performance issues as that on heap cache is thrashed on every comparison.

I didn't review fully, but wanted to ensure this performance was a consideration.

lucene/core/src/java/org/apache/lucene/codecs/hnsw/ScalarQuantizedVectorScorer.java
```diff
@@ -165,6 +169,15 @@ public float score(int node) throws IOException {
       };
     }
 
+    @Override
+    public float score(int firstOrd, int secondOrd) throws IOException {
+      return similarity.score(
+          values.vectorValue(firstOrd),
```
jimczi, line 175: Same here, this instance should be reserved since it is used by the `scorer(int ord)` case?

lucene/core/src/java/org/apache/lucene/codecs/hnsw/DefaultFlatVectorScorer.java
```diff
@@ -109,6 +109,12 @@ public float score(int node) throws IOException {
       };
     }
 
+    @Override
+    public float score(int firstOrd, int secondOrd) throws IOException {
+      return similarityFunction.compare(
+          vectors1.vectorValue(firstOrd), vectors2.vectorValue(secondOrd));
```
jimczi, line 115: vectors1 instance is already used by the `scorer(int ord)` case. If we want to allow this I believe we need a third copy.

ChrisHegarty: ha!! this is what I originally thought too, until I added a test case for this recently, see https://github.com/apache/lucene/blob/b1d3c086194f247b7d77352b1f105e4ae829d2e8/lucene/core/src/test/org/apache/lucene/codecs/hnsw/TestFlatVectorScorer.java#L80

Since neither scoreSuppliers or scorers are thread safe, then their operation has to be considered in a single-threaded model. For example: if one where to do this:
```
ss.scorer(ord1).score(ord2)
ss.scorer(ord3).score(ord4)
```

Then vectors 1 is used to retrieve the value of, first ord1, then ord3. No issue (other than the internal cache-of-one will be invalidated when ord3 is retrieved). This is already the case.   What this PR proposes is to support a model similar to:
```
ss.score(ord1, ord2)
ss.score(ord3, ord4)
```
, which simply avoids the creation of two scorer instances.  Make sense, or maybe I've missed your point.

Additionally, I added a few clarifying javadoc comments to help clarify the usage.

PR #13374, opened 2024-05-15
Fix bug in SQ when just a single vector present in a segment

ChrisHegarty:
This commit fixes a corner case in the ScalarQuantizer when just a single vector is present. I ran into this when updating a test that previously passed successfully with Lucene 9.10 but fails in 9.x.

The score error correction is calculated to be NaN, as there are no score docs or variance.


tteofili (approved):

mayya-sharipova (approved):
@ChrisHegarty Nice catch!

PR #13379, opened 2024-05-16
lucene-monitor: make TermFilteredPresearcher.ANYTOKEN[_FIELD] public

cpoerschke:
This would allow applications -- e.g. @kotman12's https://github.com/apache/solr/pull/2382 -- to schematically validate the field w.r.t. existence and field type properties.

A more complex alternative solution would be to make things optionally configurable i.e. then applications 'know' what was configured and thus can validate it.


romseygeek (approved):
LGTM

Issue #13375, opened 2024-05-16
Improve LongRange execution for ranges that have min value as Long.MIN_VALUE and max value as Long.MAX_VALUE or min as 0, and max as 2^64-1


gautamworah96:
### Description

At Amazon Search, we came across a lot of client queries that were specifying the min value as 0, and the max value as Long.MAX_VALUE.

We don't use the associated ValueSourceQuery query internally, but use a custom logic for filtering.
An improvement here would be to rewrite ValueSourceQuery to a MatchAllDocsQuery if we detect that the min and max value are the same as the min and max value of Long.

I think we will have to check for something like (`minVal==0` and `maxVal==2^64-1`) or (`minVal==Long.MIN_VALUE` and `maxVal==Long.MAX_VALUE`) since we will have to cover both signed and unsigned cases.

We can implement similar logic for DoubleRange as well.

timgrein:
That sounds interesting, I can take a look at `DoubleRange` at first and do so some benchmarking, shouldn't be too hard to expand it then to the other data types.

gautamworah96:
>  I can take a look at DoubleRange first and do so some benchmarking

I suspect that existing benchmarks won't show needle moving results if they don't have enough queries that can be optimized through this opto. The change is still a positive one tho. We may need to benchmark it on a query set that specifically has a lot of these "full range from 0->Long.MAX_VALUE" clauses.

timgrein:
> I suspect that existing benchmarks won't show needle moving results if they don't have enough queries that can be optimized through this opto.
Yes, I'm gathering some specific numbers for this exact case. I'll also suspect that this rewriting should've a purely positive effect, especially considering that you've a max of `4` dimensions for `FloatRange`, `DoubleRange` etc., which should make detecting a query containing such a query very fast.

msokolov:
Is it valid to rewrite to match all docs though? Some docs may lack a value, and I didn't think we would match in that case.

timgrein:
I took a first pass at this and also noticed that `MatchAllDocsQuery` is semantically wrong as some docs might not have the field (also caught by the existing test cases). What I think could work is to check, whether the `valueSource` is of type `FieldValuesSource` and rewrite the query to a `FieldExistsQuery` if we encounter the maximum possible range (`-Long.MAX_VALUE` and `Long.MAX_VALUE`) at the same time.

WDYT @msokolov ?

msokolov:
+1, that should work.

PR #13383, opened 2024-05-19
Rewrite LongRange.ValueSourceQuery/MultiValueSourceQuery to FieldExistsQuery on max range

timgrein:
Related to https://github.com/apache/lucene/issues/13375

### Description

With this change `ValueSourceQuery` and `MultiValueSourceQuery` will be rewritten to a `FieldExistsQuery`, if the query range uses the maximum range possible (`[Long.MIN_VALUE, Long.MAX_VALUE]`) as the exact values of the queried field won't matter anymore as every value lies within the maximum possible range. If the query range uses exclusive boundaries the query won't be rewritten as a field could contain a value being `Long.MIN_VALUE` or `Long.MAX_VALUE` which wouldn't match anymore.

I've made `LongValuesSource.FieldValuesSource` and `MultiLongValuesSource.FieldMultiValueSource` public and added a corresponding `getField` accessor which will be needed to check, whether the value source is actually relying on the specific field.

I'll plan to open a separate PR for `DoubleRange` performing a similar optimization.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -341,6 +341,9 @@ Optimizations
 
 * GITHUB#13327: Reduce memory usage of field maps in FieldInfos and BlockTree TermsReader. (Bruno Roustant, David Smiley)
 
+* GITHUB#13383: Rewrite LongRange.ValueSourceQuery and LongRange.MultiValueSourceQuery
+  to FieldExistsQuery on a max range query. (Tim Grein)
+
 Bug Fixes
 ---------------------
 
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -341,8 +341,8 @@ Optimizations
 
 * GITHUB#13327: Reduce memory usage of field maps in FieldInfos and BlockTree TermsReader. (Bruno Roustant, David Smiley)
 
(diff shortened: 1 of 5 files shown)


msokolov:
I don't really like that we have to expose these internal classes purely to do class casts. Also, this only covers the `LongValuesSource` case but wouldn't we want to cover other numeric types as well? I wonder if we could add some kind of check `ValuesSource.minValue/maxValue` or `ValuesSource.isRangeComplete` or so that could be used more generically?

msokolov:
But TBH this is probably not worth doing if it requires any major changes since users can very easily handle this in their query-creation logic

timgrein:
> I don't really like that we have to expose these internal classes purely to do class casts.

I kinda oriented myself at `LongValueSource.ConstantLongValuesSource`, which is also `public` and used in a similar way [here](https://github.com/apache/lucene/blob/7db9c8c9bda7452232c329db2ad6df90bdcd4a37/lucene/sandbox/src/java/org/apache/lucene/sandbox/search/CoveringQuery.java#L129), but I'll check, if there's a better way without making the classes public.

> I wonder if we could add some kind of check ValuesSource.minValue/maxValue or ValuesSource.isRangeComplete or so that could be used more generically

I think `ValueSource` is extended by `LongFieldSource`, which seems to have similar semantics as `FieldValueSource` inside `LongValueSource`, but it's not used inside `LongRange` ... 🤔

PR #13389, opened 2024-05-20
Add Intervals.noIntervals() method

### LINKED ISSUE #13388: Add method to `Intervals#noIntervals(String reason)` to `Intervals` class
### Description

Followup of #13385:

While reviewing #13385 I stumbled on the same issue I had several times while building a intervals query parser for patent search (a Solr plugin for a customer). There is no way to get an IntervalsSource that matches no intervals, so it will make one clause of a complex IntervalQuery consisting of serveral clauses be a non-match. One example where I hit this: We expand wildcards and filter some terms from expanding (like adding a limit on number of terms or similar), but when trying to build an empty clause it's impossible.

The easy "workaround" in Lucene/Solr 9.x is to use this class, which is unfortunately pkg private:
https://github.com/apache/lucene/blob/1ee4f8a1115d1de623f242014681032d87ed2c1e/lucene/queries/src/java/org/apache/lucene/queries/intervals/NoMatchIntervalsSource.java

To get an instance there's a workaround, but its undocumented: `public static IntervalsSource Intervals#atLeast(int minShouldMatch, IntervalsSource... sources)`. You can misuse it by passing 1 and an empty array: `Intervals.atLeast(1)`

I wonder how this differs from `IntervalBuilder#NO_INTERVALS`, which was made public in #13385. The implementations of both classes differ: one returns an iterator, which is empty, the other one returns `null`.

I'd suggest to merge those implementations and instead of exposing a constant to end user add another static method like `Intervals#noIntervals("reason")`.

We should revert #13385 and prefer a method instead of a constant.

romseygeek:


lucene/queries/src/java/org/apache/lucene/queries/intervals/IntervalBuilder.java
```diff
@@ -67,15 +67,15 @@
...
+          throws IOException {
     assert stream != null;
 
     TermToBytesRefAttribute termAtt = stream.getAttribute(TermToBytesRefAttribute.class);
     PositionIncrementAttribute posIncAtt = stream.addAttribute(PositionIncrementAttribute.class);
     PositionLengthAttribute posLenAtt = stream.addAttribute(PositionLengthAttribute.class);
 
     if (termAtt == null) {
-      return NO_INTERVALS;
+      return Intervals.noIntervals("No terms in analyzed text");
```
expani, line 78: nit: `No terms in analyzed text` can be a constant

uschindler (approved):
Looks fine to me.

Do we have enough tests to make sure the null iterator is correctly interpreted in all cases? E.g. the one this previous bug was fixing which caused similar messages like Opensearch and previously Elasticsearch has seen?

I am not sure which name is better. No match or no intervals?

lucene/queries/src/java/org/apache/lucene/queries/intervals/IntervalBuilder.java
uschindler: We can still have a constant for the whole instance.

uschindler:
Just some tidy missing. 😜

expani:
- Lot of [interval functions](https://github.com/apache/lucene/tree/main/lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/standard/nodes/intervalfn) ( around 21 usage ) indirectly can end up using the null iterator which expected `NO_INTERVAL` before. There are other places as well which don't seem to extensively covered by UTs. 

- Existing users of `Intervals.atLeast` expect a null iterator to be returned for some cases. 


To support both behaviours, we can create a new constructor for`NoMatchIntervalsSource` which returns `NO_INTERVAL` style iterator instead of null to keep this change safe.

The new API can use this constructor instead of existing one.

romseygeek:
Returning `null` is always OK here - if you look at ConjunctionIntervalsSource or DisjunctionIntervalsSource, the sub-sources can always return `null` to indicate that there are no intervals within a given segment.

uschindler:
> * Lot of [interval functions](https://github.com/apache/lucene/tree/main/lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/standard/nodes/intervalfn) ( around 21 usage ) indirectly can end up using the null iterator which expected `NO_INTERVAL` before. There are other places as well which don't seem to extensively covered by UTs.
> * Existing users of `Intervals.atLeast` expect a null iterator to be returned for some cases.
> 
> To support both behaviours, we can create a new constructor for`NoMatchIntervalsSource` which returns `NO_INTERVAL` style iterator instead of null to keep this change safe.
> 
> The new API can use this constructor instead of existing one.

No need for separate variants. The interval query parser just constructs queries, it never executes them.

We only have to check if the interval code works correct with null iterator.

Users of Lucene are common to null iterators. They are used everywhere (scorer, doc id sets,...). Returning null is much easier than crazy logic with -1 and NO_MORE_DOCS.

expani:
My main concern was the public API `Intervals#analyzedText`  

Clients of lucene were getting a non-null iterator till now and will receive a null iterator after this change. 

Maybe not backport to 9.x as there is already an alternative of `Intervals.atLeast(1)` ?

uschindler:
Why should anybody call that method. It's internal and not for public use?

Those type of changes are common in Lucene. We change behavior of methods unless it's documented to behave exactly as described.

expani:
>It's internal and not for public use?

It's [marked public](https://github.com/apache/lucene/blob/main/lucene/queries/src/java/org/apache/lucene/queries/intervals/Intervals.java#L518) as of today and accessible the same way as `Intervals#atLeast`

uschindler:
> > It's internal and not for public use?
> 
> It's [marked public](https://github.com/apache/lucene/blob/main/lucene/queries/src/java/org/apache/lucene/queries/intervals/Intervals.java#L518) as of today and accessible the same way as `Intervals#atLeast`

This method does not return null.

That the implementation of the iterator method of IntervalSource returns null is not an issue (see comments above). The method is only intended to be called by Lucene code.

expani:
>The method is only intended to be called by Lucene code.

Then it should not be a public API.

uschindler:
Thanks!

PR #13392, opened 2024-05-20
Replace Map<Long, Object> by primitive LongObjectHashMap.

bruno-roustant:
No functional changes, only replacements by primitve maps.

Adds LongObjectHashMap and LongIntHashMap to the org.apache.lucene.util.hppc package, with some refactoring.

Adds a dependency to com.carrotsearch.hppc to the join and spatial modules. This dependency is already present in the facet module. This way these packages can use primitive hash map with float or double values.

The changes in the JoinUtil class are a good example of usage of primitive structures. This highly reduces the auto-boxing around primitives.

### DIFF
--- a/lucene/join/src/java/module-info.java
+++ b/lucene/join/src/java/module-info.java
@@ -16,8 +16,10 @@
  */
 
 /** Index-time and Query-time joins for normalized content */
+@SuppressWarnings({"requires-automatic"})
 module org.apache.lucene.join {
   requires org.apache.lucene.core;
+  requires com.carrotsearch.hppc;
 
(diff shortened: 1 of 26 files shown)


lucene/join/src/java/module-info.java
```diff
@@ -16,8 +16,10 @@
  */
 
 /** Index-time and Query-time joins for normalized content */
+@SuppressWarnings({"requires-automatic"})
```
dweiss, line 19: I should make hppc a proper module, really... I'll try to do this soon.

dweiss (approved):
LGTM.

uschindler (approved):

PR #13398, opened 2024-05-21
Break the loop when segment is fully deleted by prior delTerms or delQueries

vsop-479:
### Description

When a segment is already fully deleted by prior `delTerms` or `delQueries`, in `FrozenBufferedUpdates.applyQueryDeletes` and `FrozenBufferedUpdates.applyTermDeletes`.
We can break the loop to avoid execute remaining queries.


vsop-479:
@mikemccand 

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -58,28 +58,9 @@
...
-import org.apache.lucene.document.FieldType;
-import org.apache.lucene.document.LongPoint;
-import org.apache.lucene.document.NumericDocValuesField;
-import org.apache.lucene.document.SortedDocValuesField;
-import org.apache.lucene.document.SortedNumericDocValuesField;
-import org.apache.lucene.document.SortedSetDocValuesField;
-import org.apache.lucene.document.StoredField;
-import org.apache.lucene.document.StringField;
-import org.apache.lucene.document.TextField;
+import org.apache.lucene.document.*;
```
mikemccand, line 61: We try to avoid wildcard imports -- I'm surprised our static checker isn't angry about this -- could you put back all the direct imports?  THanks.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -202,6 +183,89 @@ public boolean keepFullyDeletedSegment(
...
+    }
+    writer.flush();
+
+    for (int i = 20; i < 30; i++) {
+      doc = new Document();
+      doc.add(new LongField("content", i, Field.Store.NO));
+      writer.addDocument(doc);
+    }
+    writer.flush();
+    writer.deleteDocuments(
```
mikemccand, line 215: Thank you for the added test cases!  But, is this code really executing the new early break?  Shouldn't you also delete by a redundant query matching all docs in the 2nd segment to tickle that new code?

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -202,6 +183,89 @@ public boolean keepFullyDeletedSegment(
...
+        new IndexWriter(
+            dir,
+            new IndexWriterConfig(new MockAnalyzer(random()))
+                .setMergePolicy(NoMergePolicy.INSTANCE));
+
+    Document doc;
+    doc = new Document();
+    doc.add(new TextField("f", "foo bar tea", Field.Store.NO));
+    writer.addDocument(doc);
+
```
mikemccand, line 242: Maybe flush after each added document so you get three segments and then the delTerms will tickle the new early-break code?

mikemccand (commented):
Thanks @vsop-479 -- this change looks great -- I left some minor comments.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
vsop-479: >  is this code really executing the new early break?

Yes, i added two queries in `deleteDocuments`. First query: [10, 19] will match all docs in 2nd segment, and result in 2nd segment `FullyDeleted`. Second query: [12, 15] will make early break happen.

vsop-479: By the way, maybe we can sort the queries(seperate PR?), to let the big range query executed first, so the segment can `FullyDeleted` early.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
vsop-479: In current version, `delTerm` `foo` will result in the segment `FullyDeleted`, and `delTerm` `tea` will make early break happen.

vsop-479:
@mikemccand 
Do you think we can make `FrozenBufferedUpdates.applyQueryDeletes` and `FrozenBufferedUpdates.applyTermDeletes` 's segment task parallel executed?

vsop-479:
@jpountz 

vsop-479:
Hi @mikemccand ,
I think I addressed all your comments, Please take a look when you get a chance.

vsop-479:
@mikemccand @jpountz 

PR #13401, opened 2024-05-21
Use SPI instead of Enum for VectorSimilarityFunctions

Pulkitg64:
### Description

This PR is to get feedback on the idea and any major changes required in the commit.

In this commit we are using Java SPI instead of ENUM to define VectorSimilarityFunction used for vector search. Current implementation in Lucene tightly couples ENUM to Lucene index i.e. enum values are directly used in field info which are stored in index and later used for reading purpose. This makes adding or removing any similarity function to the ENUM very difficult as removing any function from ENUM can change the ordinal value of the functions and cause mismatch of field while reading fields from the Lucene index. On the other hand Java SPI makes life easier by providing pluggable implementation of the similarity function. With SPI, we can extend or remove similarity function easily without the need for changing things at indexing and searching sides.

For backward compatibility, I have kept ordinals with the similarity functions which can be directly used for writing/reading fields to/from the index. For Lucene version >=10 we can avoid using ordinal values and directly use function name for reading/writing the index.


Pulkitg64:
@benwtrent @uschindler @ChrisHegarty
Could you please take a look, if you get a chance?

navneet1v:
+1 on the feature and functionality. I would like to recommend one thing here:

Can we add the reloading the SPIs functionality for VectorSimilarityFunctions just like we have for codecs. Ref: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/Codec.java#L126-L137

this will ensure that vector similarity functions which are not inside Lucene can be created if anyone want to put a different similarity functions like l1, linf etc for doing similarity search.

Pulkitg64:
Makes sense. Thanks @navneet1v for the suggestion.

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene90/Lucene90FieldInfosFormat.java
```diff
@@ -103,16 +104,27 @@
...
+ *         <li>2: COSINE similarity. ({@link
+ *             org.apache.lucene.index.CosineVectorSimilarityFunction})
  *       </ul>
  * </ul>
  *
  * @lucene.experimental
  */
 public final class Lucene90FieldInfosFormat extends FieldInfosFormat {
 
+  private static final Map<Integer, String> SIMILARITY_FUNCTIONS_MAP = new HashMap<>();
```
uschindler, line 120: Use Java 9+ `Map.of()` here

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene90/Lucene90FieldInfosFormat.java
```diff
@@ -257,11 +269,12 @@ private static DocValuesType getDocValuesType(IndexInput input, byte b) throws I
     }
   }
 
-  private static VectorSimilarityFunction getDistFunc(IndexInput input, byte b) throws IOException {
-    if (b < 0 || b >= VectorSimilarityFunction.values().length) {
-      throw new CorruptIndexException("invalid distance function: " + b, input);
+  /** Returns VectorSimilarityFunction from index input and ordinal value */
+  public static VectorSimilarityFunction getDistFunc(IndexInput input, byte b) throws IOException {
+    if ((int) b < 0 || (int) b >= SIMILARITY_FUNCTIONS_MAP.size()) {
```
uschindler, line 274: this check is not correct if we have a sparse set of IDs.

It is better to just use `SIMILARITY_FUNCTIONS_MAP.contains(Integer.valueOf(b))`

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene90/Lucene90FieldInfosFormat.java
```diff
@@ -257,11 +269,12 @@ private static DocValuesType getDocValuesType(IndexInput input, byte b) throws I
...
+      throw new CorruptIndexException("invalid distance function: " + (int) b, input);
     }
-    return VectorSimilarityFunction.values()[b];
+    return VectorSimilarityFunction.forName(SIMILARITY_FUNCTIONS_MAP.get((int) b));
```
uschindler, line 277: use 'Integer.valueOf(b)' so we have type safety, as we should not accidentally use wrong type when looking up in map.

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsReader.java
```diff
@@ -19,18 +19,35 @@
...
 import org.apache.lucene.util.Bits;
 
 /** Reads vectors from an index. */
 public abstract class KnnVectorsReader implements Closeable {
 
+  /**
+   * SIMILAIRTY_FUNCTION_MAP containing hardcoded mapping for ordinal to vectorSimilarityFunction
+   * name
+   */
+  public static final Map<Integer, String> SIMILARITY_FUNCTIONS_MAP = new HashMap<>();
```
uschindler, line 42: same here, use `Map.of()`

lucene/core/src/java/org/apache/lucene/index/VectorSimilarityFunction.java
```diff
@@ -16,104 +16,73 @@
...
-   * poor search results. Floating point vectors must be normalized to be of unit length, while byte
-   * vectors should simply all have the same norm.
-   */
-  DOT_PRODUCT {
-    @Override
-    public float compare(float[] v1, float[] v2) {
-      return Math.max((1 + dotProduct(v1, v2)) / 2, 0);
+    static NamedSPILoader<VectorSimilarityFunction> getLoader() {
+      if (LOADER == null) {
+        throw new IllegalStateException();
```
uschindler, line 33: add useful message here like in the other SPI holder classes (Postingsformats, codec, docvalues)

lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseFieldInfoFormatTestCase.java
```diff
@@ -328,6 +332,17 @@ private int getVectorsMaxDimensions(String fieldName) {
     return Codec.getDefault().knnVectorsFormat().getMaxDimensions(fieldName);
   }
 
+  private VectorSimilarityFunction randomSimilarity() {
```
uschindler, line 335: this should be rewritten to not use ServiceLoader directly and instead use the NamedSPILoader in the provider. If it is needed for tests, you can add a function to retrieve all registered similarities in VectorSimilarityFunction class.

We have something similar for analyzers.

lucene/luke/src/java/org/apache/lucene/luke/app/desktop/components/DocumentsPanelProvider.java
```diff
@@ -1235,17 +1235,17 @@ private static String flags(org.apache.lucene.luke.models.documents.DocumentFiel
         sb.append("K");
         sb.append(String.format(Locale.ENGLISH, "%04d", f.getVectorDimension()));
         sb.append("/");
-        switch (f.getVectorSimilarity()) {
-          case COSINE:
+        switch (f.getVectorSimilarity().getName()) {
```
uschindler, line 1238: i would change this to just return the name and remove switch completely.

uschindler (changes requested):
I have some comments, but this is not a final review. Just things that I stumbled upon on first walkthrough.

I will have no time to do a closer review soon, so please give me some time.

lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseFieldInfoFormatTestCase.java
uschindler: Implement it like this for codecs: https://github.com/apache/lucene/blob/05f04aa08a4fcd828b92eeeb61f55d8ba18cab53/lucene/core/src/java/org/apache/lucene/codecs/Codec.java#L121-L124

lucene/test-framework/src/java/module-info.java
```diff
@@ -19,6 +19,7 @@
 @SuppressWarnings({"module", "requires-automatic", "requires-transitive-automatic"})
 module org.apache.lucene.test_framework {
   uses org.apache.lucene.codecs.KnnVectorsFormat;
+  uses org.apache.lucene.index.VectorSimilarityFunction;
```
uschindler, line 22: this can be removed if you impelemt it correctly (see below)

uschindler: if you are on it, please also remove the ServiceLoader usage in LuceneTestCase that introduces the KnnVectorsFormat, too.

In general both should have a method to list all available formats/siliarities.

uschindler: This will also speed up tests using the randomness as the heavy lookup is prevented. This should not be done over and over.

benwtrent:
I did kind of change before, and the added complexity and backwards compatibility concerns just didn't seem warranted. This is why the decision to do the scorer pluggability was added to the codecs. 

How do you communicate to codecs the similarity kind?

For example, Vector quantization needs to know if the similarity is cosine. Do we do a "cosine".equals(similarity.name())? That is very fragile. What if some one does an "optimized cosine"? Name conflict or they just cant use vector quantization.

What if someone adds an SPI similarity that should be normalized before being quantized?

I don't see a way of doing this without leaking this internal dependency (normalization being required) into the similarity SPI, which is weird to me.

I just don't think this is feasible.

benwtrent:
My old PR: https://github.com/apache/lucene/pull/13200

lucene/core/src/java/org/apache/lucene/index/VectorSimilarityFunction.java
Pulkitg64: Done

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsReader.java
Pulkitg64: Done

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene90/Lucene90FieldInfosFormat.java
Pulkitg64: Done

Pulkitg64:
> I have some comments, but this is not a final review. Just things that I stumbled upon on first walkthrough.
> 
> I will have no time to do a closer review soon, so please give me some time.

Thank you @uschindler for all the comments, I have tried to fix them in the latest commit. Looking forward for more.

Pulkitg64:
> I did kind of change before, and the added complexity and backwards compatibility concerns just didn't seem warranted. This is why the decision to do the scorer pluggability was added to the codecs.
> 
> How do you communicate to codecs the similarity kind?
> 
> For example, Vector quantization needs to know if the similarity is cosine. Do we do a "cosine".equals(similarity.name())? That is very fragile. What if some one does an "optimized cosine"? Name conflict or they just cant use vector quantization.
> 
> What if someone adds an SPI similarity that should be normalized before being quantized?
> 
> I don't see a way of doing this without leaking this internal dependency (normalization being required) into the similarity SPI, which is weird to me.
> 
> I just don't think this is feasible.

Thank you @benwtrent for the feedback and sharing your PR. I didn't know, this attempt was already made before. The main motivation behind this change was to get rid of ENUM implementation which is tightly coupled to field-info. This has caused inconvenience in deprecating the COSINE function from the list. Not sure what's the best approach then. Will take a look at your PR and comments to understand more about this.

lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseFieldInfoFormatTestCase.java
Pulkitg64: Done

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene90/Lucene90FieldInfosFormat.java
Pulkitg64: Done

lucene/test-framework/src/java/module-info.java
Pulkitg64: Done! Thanks for the feedback.

ChrisHegarty:
> The main motivation behind this change was to get rid of ENUM implementation which is tightly coupled to field-info. This has caused inconvenience in deprecating the COSINE function from the list. Not sure what's the best approach then. Will take a look at your PR and comments to understand more about this.

This is not a very compelling reason supporting the proposed change. I agree with @benwtrent, there are several challenges that would need to be ironed out before we could consider moving this forward.    

The recent addition of FlatVectorsScorer has certainly improved the extensibility in this area. For now at least, it appears to offer what is needed to plugin in new implementations.

lucene/test-framework/src/java/org/apache/lucene/tests/util/LuceneTestCase.java
```diff
@@ -3216,10 +3215,11 @@ public static BytesRef newBytesRef(byte[] bytesIn, int offset, int length) {
   }
 
   protected KnnVectorsFormat randomVectorFormat(VectorEncoding vectorEncoding) {
```
uschindler, line 3217: I fixed this already in https://github.com/apache/lucene/pull/13428

lucene/core/src/java/org/apache/lucene/index/VectorSimilarityFunction.java
```diff
@@ -16,104 +16,88 @@
...
+   * <p><em>This method is expensive and should only be called for discovery of new codecs on the
+   * given classpath/classloader!</em>
    */
-  public abstract float compare(byte[] v1, byte[] v2);
+  public static void reloadVectorSimilarityFunction(ClassLoader classloader) {
+    Holder.getLoader().reload(classloader);
+  }
+
+  /** Return list of all VectorSimilarity functions name */
+  public static List<String> getAvailableVectorSimilarityFunction() {
```
uschindler, line 95: please do this in the same way like in the other formats and return a set

lucene/core/src/java/org/apache/lucene/index/VectorSimilarityFunction.java
```diff
@@ -16,104 +16,88 @@
...
+    return Holder.getLoader().availableServices().stream().toList();
+  }
+
+  /** Returns Iterator to VectorSimilarityFunctions */
+  public static Iterator<VectorSimilarityFunction> getIterator() {
```
uschindler, line 100: remove this, not needed. The iterator is highly internal and should not be used.

uschindler (changes requested):

lucene/core/src/java/org/apache/lucene/index/VectorSimilarityFunction.java
uschindler: the function name is wrong: needs to be plural

benwtrent:
Wanted to touch base on this PR as it seems to have been stalled, mainly by me. 

The only format that would support pluggable similarities would be `Lucene99HnswVectorsFormat`. Any of the quantized codecs would have to throw an exception on an unknown similarity name. 

This now complicates a user's mental model support matrix. Having to consider not only codecs, but similarities, all of which are pluggable.

This inherit complexity of making things pluggable is why I think the implication of the "pluggable similarities" is "just make your own format".

However, I am not against moving away from `enum` and moving towards a nominal/id set of core interfaces. Enums are notoriously painful for BWC as removing one adjusts its "id" and now various edge's have to be smoothed out all over the place.


All this talk of a pluggable SPI for vector similarities spawned out of the complexities of adding fully BWC similarity functions and the difficulty of deprecating and moving on.

So, I propose:

 - We deprecate cosine (as we already have) and remove it from being writable in v10
 - Move away from enums to an id/nominal system for the similarities (what this PR could do)

PR #13402, opened 2024-05-21
[9.x] Add a MemorySegment Vector scorer - for scoring without copying on-heap

ChrisHegarty:
Backport of 
 * #13339


uschindler:
This is not so easy to do. I think we have to clone the whole vector code to Java 21, but without the memorysegment shortcuts.

I'd suggest:
- keep the Java 20 code as is
- clone all existing code to java 21
- patch the java 21 code
- add vector parts to Java 21 apijar
I can help with that, but I am travelling at moment.

uschindler:
But I tend to not put this into Lucene 9.x. IMHO, it's too risky.

lucene/core/src/java21/org/apache/lucene/util/VectorUtilPanamaProvider.txt
```diff
@@ -1,2 +0,0 @@
-The version of VectorUtilPanamaProvider for Java 21 is identical to that of Java 20.
```
uschindler, line 1: hahaha, I wanted to remove this file already. I noticed that at some point.

uschindler (approved):
Thanks, exactly as I had in mind :-)

PR #13404, opened 2024-05-23
gradlew: no "--source 11"

dsmiley:
* avoid WrapperDownloader if have the JAR
* don't specify --source More specific than needed, and some JDK/configs may complain about an incompatibility with --release.

From https://github.com/apache/solr/pull/2419


dweiss:
LGTM, although "some JDK/configs may complain about an incompatibility with --release" is intriguing - what are these JDK distributions, exactly?

dweiss (approved):

uschindler (approved):

PR #13405, opened 2024-05-23
Remove unchecked Scorable -> Scorer cast in lucene/monitor.

jpountz:
While doing an unrelated refactoring, I got hit by this unchecked cast, which is incorrect when the presearcher query produces some specialized `BulkScorer`.

### DIFF
--- a/lucene/monitor/src/java/org/apache/lucene/monitor/Monitor.java
+++ b/lucene/monitor/src/java/org/apache/lucene/monitor/Monitor.java
@@ -34,8 +34,6 @@ import org.apache.lucene.search.Matches;
 import org.apache.lucene.search.MatchesIterator;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.ScoreMode;
-import org.apache.lucene.search.Scorer;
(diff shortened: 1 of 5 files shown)


romseygeek (approved):
LGTM

lucene/monitor/src/java/org/apache/lucene/monitor/Monitor.java
```diff
@@ -377,9 +375,7 @@ public ScoreMode scoreMode() {
     @Override
     public void matchQuery(final String id, QueryCacheEntry query, QueryIndex.DataValues dataValues)
         throws IOException {
-      Scorer scorer = ((Scorer) dataValues.scorer);
-      Weight w = scorer.getWeight();
-      Matches matches = w.matches(dataValues.ctx, scorer.docID());
+      Matches matches = dataValues.weight.matches(dataValues.ctx, dataValues.docID);
```
cpoerschke, line 378: `dataValues.scorer` becomes unused now then it seems, #13412 proposes to remove it.

PR #13406, opened 2024-05-23
Replace List<Integer> by IntArrayList and List<Long> by LongArrayList.

bruno-roustant:
Add IntArrayList and LongArrayList to the HPPC fork.
Use them to replace usages of List of Integer or Long.

No functional changes, only optimization for less auto-boxing and more compact structures.
Also, no public API changes.

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/core/FlattenGraphFilter.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/core/FlattenGraphFilter.java
@@ -350,7 +350,7 @@ public final class FlattenGraphFilter extends TokenFilter {
              * The last node in the alt path didn't arrive to remove this reference.
              */
             assert inputNodes.get(inputFrom).tokens.isEmpty() : "about to remove non empty edge";
-            outSrc.inputNodes.remove(inputFrom);
+            outSrc.inputNodes.removeElement(inputFrom);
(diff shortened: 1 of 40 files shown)


lucene/analysis/common/src/java/org/apache/lucene/analysis/core/FlattenGraphFilter.java
```diff
@@ -349,7 +350,7 @@ public boolean incrementToken() throws IOException {
              * The last node in the alt path didn't arrive to remove this reference.
              */
             assert inputNodes.get(inputFrom).tokens.isEmpty() : "about to remove non empty edge";
-            outSrc.inputNodes.remove(Integer.valueOf(inputFrom));
+            outSrc.inputNodes.remove(inputFrom);
```
dweiss, line 353: I wonder if it makes sense to rename this early to "removeElement", in the spirit of changes coming to HPPC?

dweiss (approved):
LGTM.

lucene/analysis/common/src/java/org/apache/lucene/analysis/core/FlattenGraphFilter.java
bruno-roustant: Good catch!
We discussed that name in HPPC after I wrote this line and I forgot to change here. Thanks.

PR #13408, opened 2024-05-24
Move bulkScorer() from Weight to ScorerSupplier

jpountz:
This relates to #13359: we want to take advantage of the `Weight#scorerSupplier` call to start scheduling some I/O in the background in parallel across clauses. For this to work properly with top-level disjunctions, we need to move `#bulkScorer()` from `Weight` to `ScorerSupplier` as well, so that the disjunctive `BooleanQuery` first performs a call to `Weight#scorerSupplier()` on all inner clauses, and then `ScorerSupplier#bulkScorer` on all inner clauses.

`ScorerSupplier#get` and `ScorerSupplier#bulkScorer` only support being called once. This forced me to fix some inefficiencies in `bulkScorer()` implementations when we would pull scorers and then throw it away when realizing that the strategy we were planning on using was not optimal. This is why e.g. `ReqExclBulkScorer` now also supports prohibited clauses that produce a two-phase iterator.


jpountz:
Note: `Boolean2ScorerSupplier` was renamed to `BooleanScorerSupplier` since it now handles both `BS1` and `BS2`, not only `BS2`.

mikemccand (approved):
Thanks @jpountz

PR #13411, opened 2024-05-24
Add prefetching for doc values and norms.

jpountz:
This follows a similar approach as postings and only prefetches the first page of data.

I verified that it works well for collectors such as `TopFieldCollector`, as `IndexSearcher` first pulls a `LeafCollector`, then a `BulkScorer` and only then starts feeding the `BulkScorer` into the `LeafCollector`. So the background I/O for the `LeafCollector` which will prefetch the first page of doc values and the background I/O for the `BulkScorer` will run in parallel.


easyice (approved):
LGTM

PR #13412, opened 2024-05-24
lucene-monitor: remove now-unused Scorable in QueryIndex.DataValues

cpoerschke:
Became unused via #13405 I think.


jpountz:
The monitor doesn't seem to care about scores, so this makes sense to me but I'm curious of @romseygeek 's opinion?

romseygeek (approved):
LGTM.  This is only used for debugging presearcher issues, and that works via matches and doesn't care about scores.

PR #13413, opened 2024-05-24
lucene-monitor: replace wildcard imports

cpoerschke:
It's a minor thing and subjective but in #13412 I was momentarily confused w.r.t. ReadonlyQueryIndex.java not needing an import cleanup when WritableQueryIndex.java did, due to wildcard import use in the former.

Also local style consistency helps when working with code, this now then removes the three remaining wildcard imports in the lucene-monitor code area.


romseygeek (approved):
LGTM.  Does this mean that we can enable some code formatting on this module?

cpoerschke:
> ... Does this mean that we can enable some code formatting on this module?

Good question, I don't know. I was surprised to find there wasn't any logic already disallowing wildcard imports but yeah in general big fan of automatic code formatting.

PR #13417, opened 2024-05-24
Improve Test Coverage: added tests for SlowLog

matthewsah:
This pull request adds tests for the SlowLog class which can be found in org/apache/lucene/monitor. There were no previously written test cases that cover SlowLog, so these tests will improve the code coverage of the project.


lucene/monitor/src/test/org/apache/lucene/monitor/TestSlowLog.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import org.apache.lucene.tests.util.LuceneTestCase;
+
+public class TestSlowLog extends LuceneTestCase {
+
+  public void testAddQuery() {
+    SlowLog slowLog = new SlowLog();
+    long time1 = 1;
```
stefanvodita, line 29: Nit: If we're doing the validation in a loop, why not do the set-up as a loop? Same for the other test cases.

stefanvodita (approved):

PR #13418, opened 2024-05-24
Improve Test Coverage: added tests for IntRange

harenlin:
### Description
In the ```document/TestIntRange.java```, the test case only tests for the valid input. I added the test for the invalid input, which should catch the exception message. By doing so, we can improve the line coverage and branch coverage from 62% (43/69) to 63% (44/69) and 56% (17/30) to 60% (18/30) respectively.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/document/TestIntRange.java
+++ b/lucene/core/src/test/org/apache/lucene/document/TestIntRange.java
@@ -23,4 +23,16 @@ public class TestIntRange extends LuceneTestCase {
     IntRange range = new IntRange("foo", new int[] {1, 11, 21, 31}, new int[] {2, 12, 22, 32});
     assertEquals("IntRange <foo: [1 : 2] [11 : 12] [21 : 22] [31 : 32]>", range.toString());
   }
+
+  public void testInvalidRangeInput() {
+    String actualMessage = "";
+    try {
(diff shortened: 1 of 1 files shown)


lucene/core/src/test/org/apache/lucene/document/TestIntRange.java
```diff
@@ -23,4 +23,16 @@ public void testToString() {
     IntRange range = new IntRange("foo", new int[] {1, 11, 21, 31}, new int[] {2, 12, 22, 32});
     assertEquals("IntRange <foo: [1 : 2] [11 : 12] [21 : 22] [31 : 32]>", range.toString());
   }
+
+  public void testInvalidRangeInput() {
+    String actualMessage = "";
+    try {
+      IntRange range = new IntRange("foo", new int[] {-1, 11, 21, 31}, new int[] {-3, 12, 22, 32});
+      range.toString();
```
stefanvodita, line 31: This doesn't run, right? The exception is thrown from the constructor call.

lucene/core/src/test/org/apache/lucene/document/TestIntRange.java
```diff
@@ -23,4 +23,16 @@ public void testToString() {
...
+    } catch (IllegalArgumentException exception) {
+      actualMessage = exception.getMessage();
+    }
+    String expectedMessage = "min value (-1) is greater than max value (-3)";
```
stefanvodita, line 35: There are several different `InvalidArgumentException`s. Why are we targeting this specific one?

PR #13419, opened 2024-05-24
Add new test case "testGetLines" for lucene/core/analysis/WordlistLoader

hack4chang:
## Description
- Add new test case to test the ```getLines``` in ```WordlistLoader``` functions normally.
I create a string testcase that contains comment lines and blank lines to test ```getLines``` method. In order to pass a InputStream, I converted string into byteArray and then into byteArrayInputStream. This testcase proved that the function works well. By the definition of the function itself, the comment lines would have '#' at the start of the line, but I'm not sure if a line start with some spaces and then the '#' sign appears after the indentation count as a comment line as well. TBD

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/analysis/TestWordlistLoader.java
+++ b/lucene/core/src/test/org/apache/lucene/analysis/TestWordlistLoader.java
@@ -89,6 +89,7 @@ public class TestWordlistLoader extends LuceneTestCase {
     byte[] sByteArr = s.getBytes(charset);
     InputStream sInputStream = new ByteArrayInputStream(sByteArr);
     List<String> lines = WordlistLoader.getLines(sInputStream, charset);
+    assertEquals(3, lines.size());
     assertEquals("One", lines.get(0));
     assertEquals("Two", lines.get(1));
     assertEquals("Three", lines.get(2));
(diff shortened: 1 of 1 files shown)


lucene/core/src/test/org/apache/lucene/analysis/TestWordlistLoader.java
```diff
@@ -77,4 +82,17 @@ public void testSnowballListLoading() throws IOException {
...
   }
+
+  public void testGetLines() throws IOException {
+    String s = "One \n#Comment \n \n Two \n  Three  \n";
+    Charset charset = StandardCharsets.UTF_8;
+    byte[] sByteArr = s.getBytes(charset);
+    InputStream sInputStream = new ByteArrayInputStream(sByteArr);
+    List<String> lines = WordlistLoader.getLines(sInputStream, charset);
+    assertEquals(3, lines.size());
+    assertFalse(lines.contains("#Comment"));
```
stefanvodita, line 93: Nit: The other asserts, taken together, already make a stronger guarantee about this comment than this assert, making it a bit redundant.

stefanvodita (approved):

lucene/core/src/test/org/apache/lucene/analysis/TestWordlistLoader.java
hack4chang: Hi, yeah sounds resonable, I will make a change.

stefanvodita: Maybe I wasn't clear, but I only referred to the assertion about an item missing from the list, while the new commit also removed the assertion about the length of the list. If we assert length 3 and then assert on 3 elements of the list, that fully characterises the list, right?

hack4chang: Ah, got it. Yeah you're right. I'll add the length assertion back.

stefanvodita (approved):
Thank you for persevering with my nitpicky comments!

PR #13420, opened 2024-05-25
Replace Map<Character> by CharObjectHashMap and Set<Character> by CharHashSet.

bruno-roustant:
This is my last round of adding primitive structures to the HPPC fork, and using them to gain memory on various locations.
No functional change, only optimization.

In JapaneseKatakanaUppercaseFilter, in addition to replacing the generic structure, this PR also optimizes the character replacement in the CharTermAttribute.

### DIFF
--- a/lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
+++ b/lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
@@ -16,7 +16,10 @@
  */
 package org.apache.lucene.analysis.ja;
 
+import static org.apache.lucene.analysis.ja.JapaneseFilterUtil.createCharMap;
+
 import java.io.IOException;
+import java.util.Map;
 import org.apache.lucene.analysis.TokenFilter;
 import org.apache.lucene.analysis.TokenStream;
 import org.apache.lucene.analysis.tokenattributes.CharTermAttribute;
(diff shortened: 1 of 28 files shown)


lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseHiraganaUppercaseFilter.java
```diff
@@ -30,25 +30,15 @@
...
-            Map.entry('ぇ', 'え'),
-            Map.entry('ぉ', 'お'),
-            Map.entry('っ', 'つ'),
-            Map.entry('ゃ', 'や'),
-            Map.entry('ゅ', 'ゆ'),
-            Map.entry('ょ', 'よ'),
-            Map.entry('ゎ', 'わ'),
-            Map.entry('ゕ', 'か'),
-            Map.entry('ゖ', 'け'));
+        CharObjectHashMap.from(
```
dweiss, line 39: I think it would be better to construct this from pairs of Map.entry - parallel arrays here are much more difficult to read than char pairs, I think.

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/JapaneseKatakanaUppercaseFilter.java
```diff
@@ -30,41 +30,21 @@
...
-            Map.entry('ュ', 'ユ'),
-            Map.entry('ョ', 'ヨ'),
-            Map.entry('ㇻ', 'ラ'),
-            Map.entry('ㇼ', 'リ'),
-            Map.entry('ㇽ', 'ル'),
-            Map.entry('ㇾ', 'レ'),
-            Map.entry('ㇿ', 'ロ'),
-            Map.entry('ヮ', 'ワ'));
+        CharObjectHashMap.from(
+            new char[] {
```
dweiss, line 40: Same here. An alternative is to initialize from char[][] { {'a', b'}, {'c', 'd'} }, etc.

dweiss (approved):
Just minor comments about code readability here and there.

PR #13423, opened 2024-05-27
MemorySegment scorer should ensure that the values is of the correct type

ChrisHegarty:
This commit updates the MemorySegment scorer so that it ensures the values is of the correct type.  

The offset calculations for vectors in RandomAccessQuantizedByteVectorValues will be different than that of non-quantized. We can generalise the implementation for quantized vectors later, but for now, passing a quantised values indicated bug in wrapping or delegation.


uschindler (approved):
It's an assertion only, but yes we should work on this to ensure this cannot happen.

PR #13424, opened 2024-05-27
Add prefetching support to stored fields.

jpountz:
This adds `StoredFields#prefetch(int)`, which mostly delegates to `IndexInput#prefetch`. Callers can take advantage of this API to parallelize I/O across multiple stored documents by first calling `StoredFields#prefetch` on all doc IDs before calling `StoredFields#document` on all doc IDs.

I added a cache of recently prefetched blocks to the default codec, in order to avoid prefetching the same block multiple times in a short period of time. This felt sensible given that doc ID reordering via recursive graph bisection or index sorting are likely to result in search results being clustered.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene90/compressing/Lucene90CompressingStoredFieldsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene90/compressing/Lucene90CompressingStoredFieldsReader.java
@@ -40,6 +40,7 @@ import static org.apache.lucene.codecs.lucene90.compressing.Lucene90CompressingS
 
 import java.io.EOFException;
 import java.io.IOException;
+import java.util.Arrays;
 import org.apache.lucene.codecs.CodecUtil;
 import org.apache.lucene.codecs.StoredFieldsReader;
 import org.apache.lucene.codecs.compressing.CompressionMode;
@@ -72,6 +73,9 @@ import org.apache.lucene.util.LongsRef;
  */
 public final class Lucene90CompressingStoredFieldsReader extends StoredFieldsReader {
 
+  private static final int PREFETCH_CACHE_SIZE = 1 << 4;
+  private static final int PREFETCH_CACHE_MASK = PREFETCH_CACHE_SIZE - 1;
+
   private final int version;
   private final FieldInfos fieldInfos;
   private final FieldsIndex indexReader;
@@ -86,6 +90,11 @@ public final class Lucene90CompressingStoredFieldsReader extends StoredFieldsRea
   private final long numChunks; // number of written blocks
   private final long numDirtyChunks; // number of incomplete compressed blocks written
   private final long numDirtyDocs; // cumulative number of docs in incomplete chunks
+  // Cache of recently prefetched block IDs. This helps reduce chances of prefetching the same block
+  // multiple times, which is otherwise likely due to index sorting or recursive graph bisection
+  // clustering similar documents together. NOTE: this cache must be small since it's fully scanned.
+  private final long[] prefetchedBlockIDCache;
+  private int prefetchedBlockIDCacheIndex;
   private boolean closed;
 
   // used by clone
@@ -103,6 +112,8 @@ public final class Lucene90CompressingStoredFieldsReader extends StoredFieldsRea
     this.numChunks = reader.numChunks;
     this.numDirtyChunks = reader.numDirtyChunks;
     this.numDirtyDocs = reader.numDirtyDocs;
+    this.prefetchedBlockIDCache = new long[PREFETCH_CACHE_SIZE];
+    Arrays.fill(prefetchedBlockIDCache, -1);
     this.merging = merging;
     this.state = new BlockState();
     this.closed = false;
@@ -150,6 +161,8 @@ public final class Lucene90CompressingStoredFieldsReader extends StoredFieldsRea
       chunkSize = metaIn.readVInt();
 
       decompressor = compressionMode.newDecompressor();
+      this.prefetchedBlockIDCache = new long[PREFETCH_CACHE_SIZE];
+      Arrays.fill(prefetchedBlockIDCache, -1);
       this.merging = false;
       this.state = new BlockState();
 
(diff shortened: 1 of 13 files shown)


jpountz:
Like for previous changes, I wrote a synthetic benchmark to make sure that this new API actually helps.

<details>
<summary>This benchmark simulates fetching 20 random stored documents in parallel. The index it creates is 39GB while my page cache only has a capacity of 25GB.</summary>

```java
import java.io.IOException;
import java.nio.file.Path;
import java.nio.file.Paths;
import java.util.ArrayList;
import java.util.List;
import java.util.Random;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.atomic.AtomicLong;

import org.apache.lucene.document.StoredField;
import org.apache.lucene.index.FilterMergePolicy;
import org.apache.lucene.index.IndexWriter;
import org.apache.lucene.index.IndexWriterConfig;
import org.apache.lucene.index.MergePolicy;
import org.apache.lucene.index.SegmentCommitInfo;
import org.apache.lucene.index.SegmentInfos;
import org.apache.lucene.index.StoredFields;
import org.apache.lucene.index.TieredMergePolicy;

public class StoredFieldsPrefetchBench {

  public static int DUMMY;

    Path dirPath = Paths.get(args[0]);
    Directory dir = FSDirectory.open(dirPath);
    if (DirectoryReader.indexExists(dir) == false) {
      MergePolicy mergePolicy = new FilterMergePolicy(new TieredMergePolicy()) {
        @Override
        public boolean useCompoundFile(SegmentInfos infos, SegmentCommitInfo mergedInfo, MergeContext mergeContext)
            throws IOException {
          return false;
        }
      };
      try (IndexWriter w = new IndexWriter(dir, new IndexWriterConfig().setUseCompoundFile(false).setMergePolicy(mergePolicy))) {
        ExecutorService executor = Executors.newFixedThreadPool(4);
        AtomicLong indexed = new AtomicLong(0);
        for (int task = 0; task < 1000; ++task) {
          executor.execute(() -> {
            for (int i = 0; i < 40_000; ++i) {
              byte[] bytes = new byte[1024];
              r.nextBytes(bytes);
              doc.add(new StoredField("content", bytes));
              try {
                w.addDocument(doc);
              } catch (IOException e) {
              }
              final long actualIndexed = indexed.incrementAndGet();
              if (actualIndexed % 1_000_000 == 0) {
                try {
                  DirectoryReader.open(w).close();
                } catch (IOException e) {
                }
              }
            }
          });
        }

        executor.shutdown();
        executor.awaitTermination(1, TimeUnit.DAYS);
        w.commit();
        System.out.println("Done indexing");
      }
    }
    List<Long> latencies = new ArrayList<>();
    try (IndexReader reader = DirectoryReader.open(dir)) {

      for (int i = 0; i < 10_000; ++i) {
        StoredFields storedFields = reader.storedFields();

        int[] ids = new int[20];
        for (int j = 0; j < ids.length; ++j) {
          ids[j] = r.nextInt(reader.maxDoc());
        }
        for (int doc : ids) {
          storedFields.prefetch(doc);
        }
        for (int doc : ids) {
          DUMMY += storedFields.document(doc).getBinaryValue("content").hashCode();
        }

        long end = System.nanoTime();
        latencies.add((end - start) / 1000);
      }
    }
    latencies.sort(null);
    System.out.println("P50: " + latencies.get(latencies.size() / 2));
    System.out.println("P90: " + latencies.get(latencies.size() * 9 / 10));
    System.out.println("P99: " + latencies.get(latencies.size() * 99 / 100));
  }

}
```
</details>

Before the change:
P50: 2942
P90: 3900
P99: 4726

After the change:
P50: 650
P90: 801
P99: 970

lucene/core/src/java/org/apache/lucene/codecs/lucene90/compressing/Lucene90CompressingStoredFieldsReader.java
```diff
@@ -609,6 +622,23 @@ public void skipBytes(long numBytes) throws IOException {
...
 
+  @Override
+  public void prefetch(int docID) throws IOException {
+    final long blockID = indexReader.getBlockID(docID);
+
+    for (long prefetchedBlockID : prefetchedBlockIDCache) {
+      if (prefetchedBlockID == blockID) {
+        return;
+      }
+    }
```
rmuir, line 633: can we use Arrays.binarysearch instead of scanning?

rmuir: or maybe even simpler, is it enough to just cache the last blockid?

jpountz: The cache is currently an unsorted rolling buffer, this is why I'm linearly scanning all entries, and also why I'm keeping it small (16 entries).

I hesitated with only caching the last blockID, which would have been slightly simpler. To have a good hit ratio though, callers would need to fetch stored documents in doc ID order. I'm not sure how common it is today. I could switch to only caching the last block ID and add javadocs suggesting that applications fetch stored documents in doc ID order, what do you think? Maybe even add a helper method that takes a ScoreDoc[] array and returns the associated stored documents doing the right thing, ie. sorting doc IDs and doing prefetching so that I/O is parallelized?

rmuir: Lets separate the concerns. This is fine for now, I'm just concerned that it stays small.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/compressing/Lucene90CompressingStoredFieldsReader.java
```diff
@@ -609,6 +622,23 @@ public void skipBytes(long numBytes) throws IOException {
...
+
+    final long blockStartPointer = indexReader.getBlockStartPointer(blockID);
+    final long blockLength = indexReader.getBlockLength(blockID);
+    fieldsStream.prefetch(blockStartPointer, blockLength);
+
+    prefetchedBlockIDCache[prefetchedBlockIDCacheIndex++ & PREFETCH_CACHE_MASK] = blockID;
```
gf2121, line 639: It seems `prefetchedBlockIDCacheIndex` may overflow if a instance has too many `prefetch` calls. Let's keep it small ?

gf2121: Got it, thanks for the explanation!

PR #13425, opened 2024-05-27
Rewrite newSlowRangeQuery to MatchNoDocsQuery when upper > lower

ioanatia:
We already have a check in place that rewrites `SortedNumericDocValuesRangeQuery` to a `FieldExistsQuery` when the lookup range interval is `[ Long.MIN_VALUE, Long.MAX_VALUE]`.

This PR adds another check to see whether the upper bound is smaller than the lower bound.
If it is not, we can confidently rewrite `SortedNumericDocValuesRangeQuery` to a `MatchNoDocsQuery`.


jpountz:
This makes sense to me, can you add a CHANGES entry under 9.11?

ioanatia:
thank you @jpountz
I added the changelog so this is ready for another review

jpountz:
@ioanatia Would you mind bumping this change to 9.12 since the 9.11 branch has been bumped in the meantime? Sorry for the inconvenience? The change looks good to me, I'll merge once this is done.

ioanatia:
thank you for the merge - I was out for a few days and did not had a chance to bump the version.

PR #13428, opened 2024-05-27
Remove incorrect/expensive use of ServiceLoader for choosing random format

uschindler:
This cleans up the code:
- Remove expensive usage of ServiceLoader which might be inconsistent with the NamedSPI code
- Add a better readable version with comment about the special case regarding bit vectors


uschindler:
This also make the method static to be in conformance with other LuceneTestCase methods

ChrisHegarty (approved):
LGTM

uschindler:
Merged. I also removed the useless extra "uses" clause in test-frameworks module-info.java. Missed to do this here.

PR #13429, opened 2024-05-27
hunspell: speed up "compress"; minimize the number of the generated entries; don't even consider "forbidden" entries anymore

donnerpeter:


dweiss (approved):
I admit I've no idea what improvement I'm looking at here but I trust you know!

donnerpeter:
> I admit I've no idea what improvement I'm looking at here but I trust you know!

I can add some comments with explanations :)

dweiss:
> I can add some comments with explanations :)

No need. I glanced through the code and it doesn't seem to be a backdoor. :) You know better what works here anyway.

uschindler:
@rmuir can you have a look?

uschindler:
> > I can add some comments with explanations :)
> 
> No need. I glanced through the code and it doesn't seem to be a backdoor. :) You know better what works here anyway.

Hahaha. I know donnerpeter and have seen code by him already. But like the xz/liblzma desaster he could possibly be working for long time on taking over the world! We should check if Gradle code was changed!

dweiss (approved):
I looked again and I think I get the gist. No Gradle executable hooks are included. 
LGTM.

PR #13430, opened 2024-05-27
Add target search concurrency to TieredMergePolicy

### LINKED ISSUE #12877: Add a MergePolicy wrapper that preserves search concurrency?
### Description

We have an issue about decoupling search concurrency from index geometry (#9721), but this comes with trade-offs as the per-segment bit of search is hard to parallelize. Maybe we should also introduce a merge policy wrapper that tries to preserve a search concurrency of `N` by preventing the creation of segments of more than `maxDoc/N` docs?

carlosdelest:
Closes https://github.com/apache/lucene/issues/12877

Adds a minimum number of segments to `TieredMergePolicy`. This allows to ensure that a minimum search concurrency parallelism is achievable using inter-segment parallelism.


jpountz (commented):
Thanks for looking into it!

Currently, it seems to support this new option the same way that it handles the max merged segment size, ie. by greedily merging segments together if the merged segment can be just as large as the target max merged segment size?

Intuitively, this doesn't feel right to me as this boils down to merging more aggressively when `minNumSegments` is configured, while I'd expect the opposite: less aggressive merging in order to preserve search concurrency. I'm curious to get thoughts on a different approach that would consist of:
 - Update the calculation of `allowedSegCount` to take this new option into account, by allowing at least `minNumSegments` if there is a single tier, or `minNumSegments-1` segments on the highest tier otherwise (since the previous tier would have as many docs as a segment on the highest tier).
 - Update `doFindMerge` to simply ignore candidate merges that have more than `allowedDocCount` docs.

I'm curious to get @mikemccand 's thoughts.

Separately, I don't think the `minNumSegments` is good. This suggests that the merge policy will ensure the index has at least this number of segments, which is not what it is doing. Maybe call in `targetSearchConcurrenty` or `targetNumSlices` instead to better reflect the connection with search?

carlosdelest:
Hi @jpountz , thanks for your thoughts!

Let me check whether I got your suggestion right:
- There are at most 2 segment tiers. The first has `minNumSegments` at most, the second `minNumSegments-1` to ensure we maintain `minNumSegments` total segments at the most.
- The max number of docs allowed in a segment is (maxDocs - numDeletes) / minNumSegments. That avoids creating larger segments than necessary so we ensure there's `minNumSegments` segments at a minimum.

Does it align with your thoughts?

Some questions I have:
- `segsPerTier` is not taken into account if `minNumSegments` is specified, correct?
- In case segments reach  `maxMergedSegmentBytes`, should we fall back to the previous `segsPerTier` behaviour?

I see that `minNumSegments` is not a good name with this approach - I think `targetSearchConcurrency` makes sense to be explicit.

Thanks!

jpountz:
> There are at most 2 segment tiers.

Well, there can be more tiers, but since tiers have exponential sizes (e.g. if you merge factor is 10, each tier has segments that are 10x bigger than the previous tier) it's almost certainly fine to ignore segments on tiers beyond the 2nd higher tier, they would account to very few documents compared with the first 2 tiers.

> segsPerTier is not taken into account if minNumSegments is specified, correct?

If `minNumSegments` is bigger than `segsPerTier`, then the highest tier should indeed allow for up to `minNumSegments` (or maybe `minNumSegments-1` if there are multiple tiers, though it may introduce complexity for little value, maybe it's better/easier to keep things simple and allow for `minNumSegments` on the highest tier all the time). But lower tiers should still aim for `segsPerTier` segments.

> In case segments reach maxMergedSegmentBytes, should we fall back to the previous segsPerTier behaviour?

I can understood this question in different ways, so I'll try to clarify how I think these two parameters should interact:
 - Even if `minNumSegments` is configured, we should still honor `maxMergedSegmentBytes` and not create segments that are bigger than that.
 - If your merge policy has segsPerTier = 10, maxMergedSegmentMB = 5GB, floorSegmentBytes = 100MB, minNumSegments = 12, and the total index size is 15GB, it should aim for allowedSegCount=23 segments (10 100MB segments ~=1GB, 14 1GB segments). So even though we could have created a segment of the maximum merged size by merging 10 1GB segments, we did not do it because then we would have ended up with a segment that had more than maxDoc/12 documents (assuming all docs contribute equally to the index size).

mikemccand:
Thank you for tackling this @carlosdelest!  What a hairy challenge ... TMP really is its own little baby chess engine, with many things it is trying to optimize towards (what the ML world now calls "multi-objective"), playing against a worthy opponent (the indexing application that keeps indexing different sized docs, doing some deletions/updates, refresh/commit, etc., producing all kinds of exotic segments).  And of course in addition to the app exotically birthing new segments, TMP itself is causing new big segments to appear when each merge completes!  This really is a game theory problem, heh.

+1 for the name `targetSearchConcurrency` -- it makes it clear this is about optimizing the "inter-segment concurrent search latency" in the resulting index, and, I think it also makes it clear that `maxMergedSegmentBytes` still wins.  I think simply aiming for the largest tier to allow `targetSearchConcurrency` segment count when that exceeds `segsPerTier` is a good approach.  After all it is those biggest segments that you really need for concurrency since searching them takes the most CPU / latency -- the lower tiers (1/10th, 1/100th, ... the size of the top tier) will be super fast by comparison and not the long pole in each query's latency.

carlosdelest:
Thanks @mikemccand and @jpountz for your explanations! This is much clearer now.

I've updated the PR, LMKWYT about the approach!

I'm looking into effective ways of testing in terms of ensuring that the last tier contains the target number of segments for search concurrency. Any suggestions on how can I check that the last tier eventually contains the expected number of segments?

lucene/core/src/test/org/apache/lucene/index/TestTieredMergePolicy.java
```diff
@@ -431,6 +447,128 @@ public void testForcedMergesRespectSegSize() throws Exception {
     dir.close();
   }
 
+  public void testForcedMergesRespectsTargetSearchConcurrency() throws Exception {
```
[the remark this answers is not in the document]
carlosdelest: I've given it a try - LMKWYT!

lucene/core/src/java/org/apache/lucene/index/TieredMergePolicy.java
```diff
@@ -522,21 +550,28 @@ private MergeSpecification doFindMerges(
         final List<SegmentCommitInfo> candidate = new ArrayList<>();
```
carlosdelest, line 550: I see some test failures, failing to merge segments that are eligible both from a bytes and doc perspective.

It seems related to [this line here](https://github.com/apache/lucene/pull/13430/files#diff-da7b621934abf1e7bc74d6bf15f17ddd95cffa4f374776125d8de5cdc0054b2dL507):

```java
      if (mergeType == MERGE_TYPE.NATURAL
          && sortedEligible.size() <= allowedSegCount
          && remainingDelCount <= allowedDelCount) {
        return spec;
      } 
```

Here we're bailing out if we don't have enough eligible segments as the allowed seg count.

When adding more allowed segment counts with the target search concurrency, this optimization makes it impossible to merge segments that are available for merging, and thus get over budget and fail the tests.

I don't get this optimization - why shouldn't we try to merge if we have less sorted eligibles? What am I missing to change this so we can take into account the bigger number of `allowedSegCount` that comes with adding target search concurrency?

lucene/core/src/java/org/apache/lucene/index/TieredMergePolicy.java
```diff
@@ -522,21 +550,28 @@ private MergeSpecification doFindMerges(
...
+                && bytesThisMerge < maxMergedSegmentBytes
+                && docCountThisMerge < allowedDocCount;
             idx++) {
           final SegmentSizeAndDocs segSizeDocs = sortedEligible.get(idx);
           final long segBytes = segSizeDocs.sizeInBytes;
-
+          int segDocCount = segSizeDocs.maxDoc - segSizeDocs.delCount;
+          if (docCountThisMerge + segDocCount > allowedDocCount) {
+            // We don't want to merge segments that will produce more documents than allowedDocCount
+            continue;
```
jpountz, line 565: I think we should `break` here, not `continue`. `continue` allows producing merges of segments whose sizes are not adjacent, I don't think we should allow this for the doc count condition, as this potentially makes merging run in quadratic time.

lucene/core/src/java/org/apache/lucene/index/TieredMergePolicy.java
jpountz: Merging is only necessary if there are more segments than necessary or more deletes than necessary. So this condition just stops looking for more merges when none of these two conditions are met?

lucene/core/src/java/org/apache/lucene/index/TieredMergePolicy.java
jpountz: We probably also should produce a singleton merge here in case `candidate.isEmpty()` (see what we're doing for too large segments below). Most likely this suggested merge will not be selected because it will have a low score, though it could end up selected if it reclaims lots of deletes and we don't have better merges.

lucene/core/src/java/org/apache/lucene/index/TieredMergePolicy.java
```diff
@@ -916,14 +951,13 @@ public MergeSpecification findForcedDeletesMerges(SegmentInfos infos, MergeConte
...
     for (SegmentCommitInfo info : infos) {
       int delCount = mergeContext.numDeletesToMerge(info);
       assert assertDelCount(delCount, info);
+      totalDelCount += delCount;
       double pctDeletes = 100. * ((double) delCount) / info.info.maxDoc();
-      if (pctDeletes > forceMergeDeletesPctAllowed && !merging.contains(info)) {
-        haveWork = true;
-        break;
-      }
+      haveWork = haveWork || (pctDeletes > forceMergeDeletesPctAllowed && !merging.contains(info));
```
jpountz, line 960: Why don't we exit the loop anymore when `haveWork` is true?

lucene/core/src/test/org/apache/lucene/index/TestTieredMergePolicy.java
```diff
@@ -77,20 +77,23 @@ protected void assertSegmentInfos(MergePolicy policy, SegmentInfos infos) throws
...
     double allowedSegCount = 0;
+    final int maxNumSegmentsOnHighestTier =
+        (int) Math.max(tmp.getSegmentsPerTier(), tmp.getTargetSearchConcurrency());
     // below we make the assumption that segments that reached the max segment
     // size divided by 2 don't need merging anymore
     int mergeFactor = (int) Math.min(tmp.getSegmentsPerTier(), tmp.getMaxMergeAtOnce());
     while (true) {
       final double segCountLevel = bytesLeft / (double) levelSizeBytes;
-      if (segCountLevel < tmp.getSegmentsPerTier() || levelSizeBytes >= maxMergedSegmentBytes / 2) {
+      if (segCountLevel < maxNumSegmentsOnHighestTier
```
jpountz, line 87: Should it allow `maxNumSegmentsOnHighestTier` for consistency with the code under `main`? Though I don't expect it to make a big difference since `segCountLevel` is a double.

```suggestion
      if (segCountLevel <= maxNumSegmentsOnHighestTier
```

lucene/core/src/test/org/apache/lucene/index/TestTieredMergePolicy.java
```diff
@@ -111,11 +114,29 @@ protected void assertSegmentInfos(MergePolicy policy, SegmentInfos infos) throws
       }
     }
 
+    // There can be more segments if we can't merge docs - they are balanced between segments
+    int maxDocsPerSegment = tmp.getMaxAllowedDocs(infos.totalMaxDoc());
```
jpountz, line 118: should we subtract deletions to totalMaxDoc?

lucene/core/src/test/org/apache/lucene/index/TestTieredMergePolicy.java
```diff
@@ -111,11 +114,29 @@ protected void assertSegmentInfos(MergePolicy policy, SegmentInfos infos) throws
...
+    int numEligible = 0;
+    int currentSum = 0;
+    for (int i = 0; i < segmentDocs.size(); i++) {
+      currentSum += segmentDocs.get(i);
+      if (currentSum > maxDocsPerSegment) {
+        break;
+      }
+      numEligible++;
+    }
+    boolean eligibleDocsMerge = numEligible > 1;
```
jpountz, line 133: I wonder if the condition can be made simpler. Would it be ok to check the sum of the doc counts of the two smallest segments? If it's greater than the allowed doc count, then no merging is allowed, otherwise this 2-segments merge would have been a legal merge?

lucene/core/src/java/org/apache/lucene/index/TieredMergePolicy.java
carlosdelest: @jpountz , what is your take on this? The problem with this condition as it stands now is that, in case we only have 10 eligible segments for merging, and we have a max allowed segments of 15, we won't try to merge segments (even though we may have 17 total segments, and 7 are not eligible).

Should we take into account the total number of current segments vs the ones that are eligible? Would it make sense to try and do some merges to reduce the number of segments to the maximum allowed?

jpountz: Sorry for the lag, I think I understand what you mean now. We may need to decrement the maximum number of segments we allow on the highest tier for every segment that is too large that we exclude from the set of segments eligible for merging?

jpountz:
I spent some time digging into the new behavior introduced by this PR, as I'd like to get it over the finish line. :) I am getting disappointing results for the write amplification. This seems to be due to the fact that we compute the segment count as if segments could magically be close to the doc count limit, but in practice this is not the case, and it forces `TieredMergePolicy` to perform sub-optimal merging.

I tried to play with other approaches, and one that seems to perform closer to what I would expect consists of removing the `targetConcurrency` biggest segments from the list of eligible segments, in a similar way to how we remove too big segments. And then let the merge policy do merging on the long tail of smaller segments.

jpountz:
@carlosdelest I took the freedom of pushing some changes to your fork, I hope you don't mind!

jpountz:
FWIW I have been playing with `BaseMergePolicyTestCase`'s `doTestSimulateAppendOnly` and `doTestSimulateUpdates`. With this change, I'm observing a number of segments in the order of `targetSearchConcurrency` on the highest tier while  increasing the segment count by a bit less `targetSearchConcurrency` on average and without affecting the write amplification significantly.

carlosdelest:
Thanks for taking a look into this @jpountz !

Your approach makes sense to me, and addresses some of the problems I was having with bigger segments not being merged because of the number of documents, and smaller segments being discarded.

The test checks makes sense as well. I've added setting target search concurrency for `testSimulateUpdates()`

Is there any other work / checks we should be doing for this?

jpountz (approved):
I just reviewed the change, it looks good to me. I also ran more simulations for write amplification, and the results look good to me:

Scenario 1: 555 flushes of exactly 60065-document segments (like nightlies) where each doc contributes 5kB to the index. No updates/deletions.

| Merge policy | Write amplification | Average number of segments | Max number of segments |
| --- | --- | --- | --- |
| TieredMergePolicy defaults | 1.99 | 33.62 | 65 |
| TieredMergePolicy targetSearchConcurrency=4 | 2.05 | 33.53 | 65 |
| TieredMergePolicy targetSearchConcurrency=8 | 2.19 | 34.31 | 65 |
| TieredMergePolicy targetSearchConcurrency=16 | 2.47 | 38.56 | 67 |

Scenario 2: 1000 flushes of random(1, 50k)-document segments where each doc contributes 5kB to the index. No updates/deletions.

| Merge policy | Write amplification | Average number of segments | Max number of segments |
| --- | --- | --- | --- |
| TieredMergePolicy defaults | 2.86 | 34.6 | 54 |
| TieredMergePolicy targetSearchConcurrency=4 | 2.93 | 36.0 | 56 |
| TieredMergePolicy targetSearchConcurrency=8 | 2.98 | 38.8 | 58 |
| TieredMergePolicy targetSearchConcurrency=16 | 3.24 | 43.6 | 61 |

Scenario 3: 1000 flushes of random(1, 50k)-document segments where each doc contributes 5kB to the index. 50% of the documents in every flush are updates, ie. each flush contributes N/2 new documents and N/2 updated documents.

| Merge policy | Write amplification | Average number of segments | Max number of segments |
| --- | --- | --- | --- |
| TieredMergePolicy defaults | 3.13 | 28.8 | 46 |
| TieredMergePolicy targetSearchConcurrency=4 | 3.12 | 28.9 | 46 |
| TieredMergePolicy targetSearchConcurrency=8 | 3.31 | 30.7 | 46 |
| TieredMergePolicy targetSearchConcurrency=16 | 3.66 | 33.5 | 51 |

One thing worth noting is that it becomes easier to enforce the target search concurrency as more segments get added to the index

@mikemccand would you like to take a look before I merge?

As a follow-up, I think we should enable this feature by default? Maybe only in Lucene 10? With a value such as 8?

I just pushed a minor improvement that ignores the limit on the doc count if we're still under the floor segment size, as it felt silly to apply this constraint in this case. 

> Is there any other work / checks we should be doing for this?

I would suggest improving `LuceneTestCase#newTieredMergePolicy(Random)` should randomly set the target concurrency on the merge policy. And remove the call to the setter to the `BaseMergePolicyTestCase#testSimulate*` test cases?

carlosdelest:
> I would suggest improving LuceneTestCase#newTieredMergePolicy(Random) should randomly set the target concurrency on the merge policy. And remove the call to the setter to the BaseMergePolicyTestCase#testSimulate* test cases?

Done in 2d8031581ad1df603c59f5f02d9c0c9afed8ca28

lucene/test-framework/src/java/org/apache/lucene/tests/util/LuceneTestCase.java
```diff
@@ -1095,6 +1095,12 @@ public static TieredMergePolicy newTieredMergePolicy(Random r) {
     } else {
       tmp.setSegmentsPerTier(TestUtil.nextInt(r, 10, 50));
     }
+    if (rarely(r)) {
+      tmp.setTargetSearchConcurrency(TestUtil.nextInt(r, 2, 20));
+    } else {
+      tmp.setTargetSearchConcurrency(TestUtil.nextInt(r, 10, 50));
```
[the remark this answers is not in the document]
carlosdelest: I see - I was mimicking the segments per tier approach here, but I guess that makes sense. Changing it

PR #13431, opened 2024-05-28
Optimize Japanese UserDictionary.

bruno-roustant:
Replace TreeMap by a List of Match in the lookup method. There was a todo on the TreeMap which caught my attention.
I ran some benchmarks and I see 5% performance improvement, but the gain is more on the memory side where we allocate less intermediate structures and less auto-boxing during the lookup.

Use compiled Pattern when reading the dictionary.

### DIFF
--- a/lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/dict/UserDictionary.java
+++ b/lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/dict/UserDictionary.java
@@ -24,7 +24,6 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.regex.Pattern;
-
 import org.apache.lucene.analysis.morph.Dictionary;
 import org.apache.lucene.analysis.util.CSVUtil;
 import org.apache.lucene.util.IntsRefBuilder;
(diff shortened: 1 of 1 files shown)


lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/dict/UserDictionary.java
```diff
@@ -168,33 +171,53 @@ public UserMorphData getMorphAttributes() {
    * @return array of {wordId, position, length}
    */
   public int[][] lookup(char[] chars, int off, int len) throws IOException {
-    // TODO: can we avoid this treemap/toIndexArray?
-    TreeMap<Integer, int[]> result = new TreeMap<>(); // index, [length, length...]
```
bruno-roustant, line 172: The TreeMap sorting was not required because we add segmentations (wordIfAndLength) in ascending order of the position (startOffset - off). I think it allowed to keep the iteration order and also allocated less memory than a LinkedHashMap.
It is now replaced by a List of Match records (position, wordIfAndLength).

lucene/analysis/kuromoji/src/java/org/apache/lucene/analysis/ja/dict/UserDictionary.java
```diff
@@ -168,33 +171,53 @@ public UserMorphData getMorphAttributes() {
...
+        matches.add(new Match(startOffset - off, wordIdAndLength));
+        numResults += wordIdAndLength.length - 1;
+      }
     }
-
-    return found ? toIndexArray(result) : EMPTY_RESULT;
+    if (numResults == 0) {
+      return EMPTY_RESULT;
+    }
+    int[][] result = new int[numResults][];
```
bruno-roustant, line 205: Interestingly, I also tried to build the result array directly inside the first loop on startOffset. While it reduced the code, it slowed down the execution, probably because this second loop disturbed the optimization of the first loop.

PR #13436, opened 2024-05-29
Fix TestIndexWriterOnError.testIOError failure.

### LINKED ISSUE #13434: TestIndexWriterOnVMError.testUnknownError fails with unclosed file handles
### Description

TestIndexWriterOnVMError.testUnknownError fails reliably with unclosed file handles. It fails in 9x and on main.

On main, the teset name is instead `TestIndexWriterOnError.testUnknownError`

```
./gradlew test --tests TestIndexWriterOnVMError.testUnknownError -Dtests.seed=8701E0F810C16F88 -Dtests.nightly=true -Dtests.locale=fr-GA -Dtests.timezone=America/Tortola -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```

jpountz:
Pull request #13406 inadvertly broke Lucene's handling of tragic exceptions by stopping after the first `DocValuesProducer` whose `close()` calls throws an exception, instead of keeping calling `close()` on further producers in the list.

This moves back to the previous behavior.

Closes #13434

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/SegmentDocValues.java
+++ b/lucene/core/src/java/org/apache/lucene/index/SegmentDocValues.java
@@ -77,7 +77,7 @@ final class SegmentDocValues {
   /** Decrement the reference count of the given {@link DocValuesProducer} generations. */
   synchronized void decRef(LongArrayList dvProducersGens) throws IOException {
     IOUtils.applyToAll(
-        dvProducersGens.stream().mapToObj(Long::valueOf),
+        dvProducersGens.stream().<Long>mapToObj(Long::valueOf).toList(),
         gen -> {
           RefCount<DocValuesProducer> dvp = genDVProducers.get(gen);
           assert dvp != null : "gen=" + gen;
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/index/SegmentDocValues.java
```diff
@@ -76,10 +76,12 @@ synchronized DocValuesProducer getDocValuesProducer(
...
-      assert dvp != null : "gen=" + gen;
-      dvp.decRef();
-    }
+    IOUtils.applyToAll(
+        dvProducersGens.stream().mapToObj(Long::valueOf).toList(),
+        gen -> {
+          RefCount<DocValuesProducer> dvp = genDVProducers.get(gen);
+          assert dvp != null : "gen=" + gen;
+          dvp.decRef();
+        });
```
benwtrent, line 85: My only minor concern is that this reverts the minor improvement of using primitive values. But, it fixes a much bigger issue. I am fine with reverting only this part of the optimization that caused the bug.

benwtrent (approved):

lucene/core/src/java/org/apache/lucene/index/SegmentDocValues.java
bruno-roustant: Yes, good catch. This fix is required to continue using IOUtils.
I don't know why I didn't read enough the method javadoc.

bruno-roustant (approved):
Thank you for the fix

PR #13445, opened 2024-05-31
Add new dynamic confidence interval configuration to scalar quantized format

benwtrent:
When int4 scalar quantization was merged, it added a new way to dynamically calculate quantiles. 

However, when that was merged, I inadvertently changed the default behavior, where a `null` confidenceInterval would actually calculate the dynamic quantiles instead of doing the previous auto-setting to `1 - 1/(dim + 1)`.

This commit formalizes the dynamic quantile calculate through setting the confidenceInterval to `0`, and preserves the previous behavior for `null` confidenceIntervals so that users upgrading will not see different quantiles than they would expect.


benwtrent:
I am planning on putting this in 9.11 as it is also a fix for breaking BWC that 9.11 currently does when using the same format configuration between 9.10 vs. 9.11. In 9.10, supplying a `null` confidence interval indicated a static interval of `1 - 1/(dim + 1)`, without this commit, we break this.

ChrisHegarty (approved):
I agree with reverting to the previous behaviour, and allowing to opt-into the new dynamic confidence intervals.

PR #13448, opened 2024-06-03
mention KnnVectorsFormat in o.a.l.codecs package javadocs

msokolov:
I found a dusty corner we seem to have missed


msokolov:
Note: I stumbled on this while trying to figure out how to index a vector field with no HNSW index ... I'm still not clear on how to do that ... we might need some more javadoc updates?

benwtrent:
@msokolov to my knowledge, there is no flat format provided by Lucene at this time.

msokolov:
I'm super confused - I thought that's what you had been working on! But .. I guess it was really more of an internal refactoring? OK ... well, it would sure be nice to expose a flat vector field so we can expose the nice scalar quantization tools.

benwtrent:
> OK ... well, it would sure be nice to expose a flat vector field so we can share the nice scalar quantization tools.

Yep, a new format should be fairly simple.

Issue #11432, opened 2022-01-31
Add support for sparse indexing [LUCENE-10396]


asfimport:
On Elasticsearch we're more and more leveraging index sorting not as a way to be able to early terminate sorted queries, but as a way to cluster doc IDs that share similar properties so that queries can take advantage of it. For instance imagine you're maintaining a catalog of cars for sale, by sorting by car type, then fuel type then price. Then all cars with the same type, fuel type and similar prices will be stored in a contiguous range of doc IDs. Without index sorting, conjunctions across these 3 fields would be almost a worst-case scenario as every clause might match lots of documents while their intersection might not. With index sorting enabled however, there's only a very small number of calls to advance() that would lead to doc IDs that do not match, because these advance() calls that do not lead to a match would always jump over a large number of doc IDs. I had created that example for ApacheCon last year that demonstrates the benefits of index sorting on conjunctions. In both cases, the index is storing the same data, it just gets different doc ID ordering thanks to index sorting:
![sorted_conjunction.png](https://apache.github.io/lucene-jira-archive/attachments/LUCENE-10396/sorted_conjunction.png)

While index sorting can help improve query efficiency out-of-the-box, there is a lot more we can do by taking advantage of the index sort explicitly. For instance `IndexSortSortedNumericDocValuesRangeQuery` can speed up range queries on fields that are primary sort fields by performing a binary search to identify the first and last documents that match the range query. I would like to introduce [sparse indexes](https://en.wikipedia.org/wiki/Database_index#Sparse_index) for fields that are used for index sorting, with the goal of improving the runtime of `IndexSortSortedNumericDocValuesRangeQuery` by making it less I/O intensive and making it easier and more efficient to leverage index sorting to filter on subsequent sort fields. A simple form of a sparse index could consist of storing every N-th values of the fields that are used for index sorting.

 - [x] https://github.com/apache/lucene/pull/13449
 - [x] Add multiple levels to `DocValuesSkipper` to be able to skip many docs at once: https://github.com/apache/lucene/pull/13563
 - [x] Align block boundaries with value changes if possible to have as many constant blocks as possible https://github.com/apache/lucene/pull/13597
 - [x] Take advantage of `DocValuesSkipper` in `SortedSetDocValuesRangeQuery` https://github.com/apache/lucene/pull/13592
 - [ ] #13838

---
[...]

PR #13449, opened 2024-06-04
Sparse index: optional skip list on top of doc values

iverase:
Speaking to Adrien about how a sparse index would look like in lucene, he suggested that the sparse indexing does not need to be a new format bit an additional responsibility if `DocValuesFormat`. 

The idea is to add an option to add a [skip list](https://en.wikipedia.org/wiki/Skip_list) on top of doc values and to expose it via the `DocValuesSkipper` abstraction, which has an API that is similar to what we're doing for impacts today. This provides a very simple index which can be very efficient when the index is sorted and the field belongs to the index sorting.

In order to implement it, we added a new flag in `FieldType.java` that configures whether to create a "skip index" for doc values. This flag is only allowed to be set on doc values of type NUMERIC, SORTED_NUMERIC, SORTED and SORTED_SET. Attempting to index other type of doc values with the flag set results on an exception.

This flag needs to be persisted on the `FieldInfosFormat`. This does not require a format change as we have some unused bit flags  in `Lucene94FieldInfosFormat` that we can use.

We have changed the `DocValuesFormat` to generate the "skip index" whenever the flag is set. For this first implementation we went to the most basic implementation which consist in a skip list with just one level. In this level we collect the documents statistics every 4096 documents and we write them into the index. This basic structure already provides interesting numbers. I discussed with Adrien that as a follow up we should introduce more levels to the skip list and optimise the index for low cardinality fields.

In order to index a field with a skip list, we added static methods to the doc values field, for example NumericDocValuesField#indexedField(String name, long value) which will generated the right FieldType. In order to query it, you can use the existing  NumericDocValuesField#newSlowRangeQuery(String field, long lowerValue, long upperValue). The generated query will use the skip index if exists by using the `DocValuesRangeIterator`. 

Finally, here are some number I got using the geonames data set from lucene util.  

The first test index the field called `modified` and adds the field as the primary sort of the index. 

```
 Index LongField query LongField#newRangeQuery

  INDEX TIME: 42.604 sec
  INDEX DOCS: 12347608 documents
  INDEX SIZE: 12.402745246887207 MB
  QUERY TIME: 1157.7 ms
  QUERY DOCS: 6243379080 documents

Index LongField query IndexSortSortedNumericDocValuesRangeQuery

  INDEX TIME: 42.562 sec
  INDEX DOCS: 12347608 documents
  INDEX SIZE: 12.402745246887207 MB
  QUERY TIME: 662.6 ms
  QUERY DOCS: 6243379080 documents

Index Doc values skipping query SortedNumericDocValuesField#newSlowRangeQuery

  INDEX TIME: 38.927 sec
  INDEX DOCS: 12347608 documents
  INDEX SIZE: 11.800291061401367 MB
  QUERY TIME: 1072.5 ms
  QUERY DOCS: 6243379080 documents
```

This basic implementation is already faster that querying using the bkd tree. The IndexSortSortedNumericDocValuesRangeQuery is faster as it contains many optimisations but my expectation is that we can make this index as fast if not faster than this implementation.

The second test, we are indexing two fields and sorting the index using them; the countryCode as primary sort and the modified field as secondary sort. Then we execute the range queries on the modified field:

```
Index KeywordField, LongField query LongField#newRangeQuery

  INDEX TIME: 50.486 sec
  INDEX DOCS: 12347608 documents
  INDEX SIZE: 24.378992080688477 MB
  QUERY TIME: 1273.0 ms
  QUERY DOCS: 6243379080 documents

Index KeywordField, LongField query SortedNumericDocValuesField#newSlowRangeQuery

  INDEX TIME: 50.486 sec
  INDEX DOCS: 12347608 documents
  INDEX SIZE: 24.378992080688477 MB
  QUERY TIME: 13392.6 ms
  QUERY DOCS: 6243379080 documents

 Index Doc values skipping for both, query  SortedNumericDocValuesField#newSlowRangeQuery
  INDEX TIME: 44.127 sec
  INDEX DOCS: 12347608 documents
  INDEX SIZE: 16.09447193145752 MB
  QUERY TIME: 2975.0 ms
  QUERY DOCS: 6243379080 documents
```
In this case the query is slower than the BKD tree but still much faster than the brute approach. The advantage of the new query is that it does not need to build the big bitset that we might need to build with the BKD tree.


jpountz (approved):
I just did a full pass on changes again, and it looks good to me (disclaimed: I contributed to this branch). Would be good to have someone else take a look as well.

ChrisHegarty (commented):
This looks great. Just a few comments

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
```diff
@@ -0,0 +1,95 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.index;
+
+import java.io.IOException;
+import org.apache.lucene.search.DocIdSetIterator;
+
+/** Skipper for {@link DocValues}. */
+public abstract class DocValuesSkipper {
```
ChrisHegarty, line 23: I struggled with this interface initially, so I'll propose a few javadoc comments which I think should help.

```
 * <p>A skipper has a position that can only be {@link #advance(int) advanced}. A skippers position,
 * along with a {@code level}, determines the interval at which the skipper is currently positioned.
 ```

File: lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java [hunk 2]
```diff
@@ -0,0 +1,95 @@
...
+   * <p><b>NOTE</b>: {@code minDocID(0)} may return a doc ID that is greater than {@code target} if
+   * the target document doesn't have a value.
+   */
+  public abstract void advance(int target) throws IOException;
+
+  /** Return the number of levels. This number may change when moving to a different interval. */
+  public abstract int numLevels();
+
+  /**
+   * Return the minimum doc ID on the given level, inclusive. This returns {@code -1} if {@link
```
ChrisHegarty, line 40:
```
Return the minimum doc ID of the interval at the given level, inclusive.
```

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
```diff
@@ -0,0 +1,95 @@
...
+   * #advance(int)} has not been called yet and {@link DocIdSetIterator#NO_MORE_DOCS} if the
+   * iterator is exhausted. This method is non-increasing when {@code level} increases. Said
+   * otherwise {@code minDocID(level+1) <= minDocId(level)}.
+   */
+  public abstract int minDocID(int level);
+
+  /**
+   * Return the maximum doc ID on the given level, inclusive. This returns {@code -1} if {@link
```
ChrisHegarty, line 48:
```
  Return the maximum doc ID of the interval on the given level, inclusive.
```

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
```diff
@@ -0,0 +1,95 @@
...
+   * #advance(int)} has not been called yet and {@link DocIdSetIterator#NO_MORE_DOCS} if the
+   * iterator is exhausted. This method is non-decreasing when {@code level} decreases. Said
+   * otherwise {@code maxDocID(level+1) >= maxDocId(level)}.
+   */
+  public abstract int maxDocID(int level);
+
+  /**
+   * Return the minimum value at the given level, inclusive.
```
ChrisHegarty, line 56: same - ".. of the internal .. "

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
```diff
@@ -0,0 +1,95 @@
...
+   *
+   * <p><b>NOTE</b>: It is only guaranteed that values in this interval are greater than or equal
+   * the returned value. There is no guarantee that one document actually has this value.
+   */
+  public abstract long minValue(int level);
+
+  /**
+   * Return the maximum value at the given level, inclusive.
```
ChrisHegarty, line 64: same - ".. of the internal .. "

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
iverase: I added as well a comment that the skipper next advance position should be bigger than #maxDocID(0).

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
```diff
@@ -1690,4 +1722,78 @@ long getLongValue(long index) throws IOException {
...
+          minDocID = DocIdSetIterator.NO_MORE_DOCS;
+          maxDocID = DocIdSetIterator.NO_MORE_DOCS;
+        } else {
+          do {
+            minDocID = input.readInt();
+            maxDocID = input.readInt();
+            minValue = input.readLong();
+            maxValue = input.readLong();
+            docCount = input.readInt();
+          } while (target > maxDocID);
```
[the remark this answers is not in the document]
iverase: This does not going against the contract:
```
   * <p><b>NOTE</b>: The behavior is undefined if {@code target} is less than or equal to {@code
   * maxDocID(0)}.
```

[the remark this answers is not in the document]
jpountz: FWIW `DocIdSetIterator#advance` has a similar contract and some implementations also have side-effects when called twice on the same doc (which is why javadocs mention it's illegal to advance on a doc that is less than or equal to the current doc).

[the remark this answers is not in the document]
ChrisHegarty: Ah. I missed that in the javadoc. Sorry for the noise, no issue here.

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
```diff
@@ -0,0 +1,101 @@
...
+   *
+   * <p><b>NOTE</b>: It is only guaranteed that values are greater than or equal the returned value.
+   * There is no guarantee that one document actually has this value.
+   */
+  public abstract long minValue();
+
+  /**
+   * Return the global maximum value.
+   *
+   * <p><b>NOTE</b>: It is only guaranteed that values are greater than or equal the returned value.
```
ChrisHegarty, line 94: typo "greater" -> "less"

ChrisHegarty (approved):
LGTM

gf2121 (approved):

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
iverase: good catch

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
```diff
@@ -197,6 +215,17 @@ private NumericEntry readNumeric(IndexInput meta) throws IOException {
...
 
+  private DocValuesSkipperEntry readDocValueSkipperMeta(IndexInput meta) throws IOException {
+    long offset = meta.readLong();
+    long length = meta.readLong();
+    long minValue = meta.readLong();
+    long maxValue = meta.readLong();
+    int docCount = meta.readInt();
+    int maxDocID = meta.readInt();
+
+    return new DocValuesSkipperEntry(offset, length, minValue, maxValue, docCount, maxDocID);
```
benwtrent, line 226: Could we validate the skipper entry? My understanding is that `length == 24*docCount` right?

benwtrent (commented):
I have minor comments. 

The bulk of the implementation makes sense to me as a first step. 

One weird thing is how we handle deletes. I am wondering about behavior when:
 - All docs in a skipper are deleted. We skip the skipper-entry due to its `maxDocID` actually being deleted as well.
 - Some of the docs deleted in a skipper. Our "min/max" at that point will not be accurate. We will still iterate the doc values in the skipper, which is OK. This just comes with the territory with deletes.

Are my ideas consistent with how we handle deletes?

I am pretty amazed that such a big change is only about 2k loc, many of which are just interface updates being populated.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
```diff
@@ -1749,4 +1781,88 @@ long getLongValue(long index) throws IOException {
...
+          while (true) {
+            maxDocID = input.readInt();
+            if (maxDocID >= target) {
+              minDocID = input.readInt();
+              maxValue = input.readLong();
+              minValue = input.readLong();
+              docCount = input.readInt();
+              break;
+            } else {
+              input.skipBytes(24);
```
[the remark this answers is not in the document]
iverase: The block size is actually 28. We read the first 4 bytes to compute the maxDocID and we skip the rest if it is not competitive.

I am hesitant to add a constant at the moment as this might change if we introduce levels.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
iverase: The validation would look like: 
```
 assert length == 28 * (1 + ((docCount - 1) / SKIP_INDEX_INTERVAL_SIZE));
```
As before I am hesitant to add the validation at the moment as this might change if we introduce levels.

lucene/core/src/java/org/apache/lucene/index/FieldInfo.java
```diff
@@ -289,6 +302,24 @@ static void verifySameDocValuesType(
     }
   }
 
+  /**
+   * Verify that the provided docValues type are the same
+   *
+   * @throws IllegalArgumentException if they are not the same
+   */
+  static void verifySameDocValuesSkipIndex(
```
easyice, line 310: Do we need to do the same check in `FieldInfos#verifySameSchema`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
```diff
@@ -129,16 +131,17 @@ public void addNumericField(FieldInfo field, DocValuesProducer valuesProducer)
...
         new EmptyDocValuesProducer() {
           @Override
           public SortedNumericDocValues getSortedNumeric(FieldInfo field) throws IOException {
             return DocValues.singleton(valuesProducer.getNumeric(field));
           }
-        },
-        false);
+        };
+    if (field.hasDocValuesSkipIndex()) {
+      writeSkipIndex(field, producer);
```
easyice, line 142: Do we need to check the skip index in `CheckIndex`?

lucene/core/src/java/org/apache/lucene/index/FieldInfo.java
iverase: I think you are right but I wonder if we should optimize first FieldInfo.FieldNumbers. I open https://github.com/apache/lucene/pull/13460 for consideration,

iverase: Added support on https://github.com/apache/lucene/pull/13449/commits/73813646f0b7a46aa5d2bf8b6a0b4e0a1ab1a8e8

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
iverase: Added inital aupport on https://github.com/apache/lucene/pull/13449/commits/e782feffc4bd35be3207df0dd46d5dbd7998301c

benwtrent (approved):
This is an exciting first step!

🚀 🚀 🚀

easyice (approved):
Great work!

iverase:
If there are no more comments, I will be merging this PR tomorrow.

jpountz:
FWIW I'm trying to use #11432 as a meta issue for sparse indexing and started listing tasks that I think we (ideally) need to complete to be in a good state for 9.0.

PR #13450, opened 2024-06-04
Set `ReadAdvice#NORMAL` on files that have a forward-only access pattern.

jpountz:
This applies to files where performing readahead could help:
 - Doc values data (`.dvd`)
 - Norms data (`.nvd`)
 - Docs and freqs in postings lists (`.doc`)
 - Points data (`.kdd`)

Other files (KNN vectors, stored fields, term vectors) keep using a `RANDOM`
advice.

### DIFF
--- a/lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
+++ b/lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
@@ -371,16 +371,6 @@ abstract class MemorySegmentIndexInput extends IndexInput
   }
 
   @Override
-  public void readAhead(long offset, long length) throws IOException {
-    // Start loading the first bytes in the background
-    if (length != 0) {
-      prefetch(offset, 1);
-    }
-    // TODO: Is there a hint we can give to the OS to let it optimize for our forward-only access
-    // pattern in the given range?
-  }
-
-  @Override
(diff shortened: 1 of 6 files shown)


lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
```diff
@@ -370,6 +370,16 @@ public void prefetch(long offset, long length) throws IOException {
...
 
+  @Override
+  public void readAhead(long offset, long length) throws IOException {
+    // Start loading the first bytes in the background
+    if (length != 0) {
+      prefetch(offset, 1);
+    }
+    // TODO: Is there a hint we can give to the OS to let it optimize for our forward-only access
+    // pattern in the given range?
+  }
```
rmuir, line 381: `MADV_SEQUENTIAL`

rmuir: or even better, `MADV_NORMAL` for the range, which is less aggressive.

jpountz: I thought you had concerns about `MADV_NORMAL` but I like it better too, I updated the PR to use `ReadAdvice.NORMAL` instead.

rmuir (approved):

PR #13453, opened 2024-06-05
Update Gradle wrapper to 8.8 - supports Java 22

ChrisHegarty:
This commit updates the Gradle wrapper to 8.8, which has support for Java 22.


gradle/testing/alternative-jdk-support.gradle
```diff
@@ -88,6 +88,6 @@ if (jvmGradle != jvmCurrent) {
 // Set up root project's properties.
 rootProject.ext.runtimeJavaExecutable = jvmCurrent.javaExecutable
 rootProject.ext.runtimeJavaHome = jvmCurrent.javaHome
-rootProject.ext.runtimeJavaVersion = jvmDetector.getMetadata(new InstallationLocation(jvmCurrent.javaHome, "specific path")).getLanguageVersion()
+rootProject.ext.runtimeJavaVersion = jvmDetector.getMetadata(InstallationLocation.userDefined(jvmCurrent.javaHome, "specific path")).getLanguageVersion()
```
ChrisHegarty, line 91: this is a needed as the internal Gradle API has changed between 8.4 and 8.8

ChrisHegarty:
While not strictly necessary to be able to run tests with JDK 22, upgrading the gradle wrapper will simplify usage. Since one can just set JAVA_HOME, rather than both JAVA_HOME and RUNTIME_JAVA_HOME.

ChrisHegarty:
@uschindler I was motivated to do this upgrade since stumbling into issues trying to verify the 9.11 RC1 build with `--test-alternative-java <path_to_jdk_22>`, and kept running into silly configuration issues. I think that my local env does not have the correct variables set or something, but it'll be more straightforward if we can just set JAVA_HOME to 22. 

e.g. failure I was seeing locally
```
...
Caused by: org.gradle.api.GradleException: Failed to create Jar file /Users/chegar/.gradle/caches/jars-9/2296546e3b60d825e5f54acd3fba42c8/buildSrc.jar.
        at org.gradle.internal.classpath.ClasspathBuilder.jar(ClasspathBuilder.java:57)
        at org.gradle.internal.classpath.InstrumentingClasspathFileTransformer$BaseTransformation.transform(InstrumentingClasspathFileTransformer.java:228)
        at org.gradle.internal.classpath.InstrumentingClasspathFileTransformer.transform(InstrumentingClasspathFileTransformer.java:194)
        at org.gradle.internal.classpath.InstrumentingClasspathFileTransformer.transform(InstrumentingClasspathFileTransformer.java:161)
        at org.gradle.internal.classpath.DefaultCachedClasspathTransformer.transformFile(DefaultCachedClasspathTransformer.java:265)
        at org.gradle.internal.classpath.DefaultCachedClasspathTransformer.lambda$cachedFile$10(DefaultCachedClasspathTransformer.java:258)
        at org.gradle.internal.UncheckedException.unchecked(UncheckedException.java:74)
        at org.gradle.internal.classpath.DefaultCachedClasspathTransformer$ParallelTransformExecutor.lambda$transformAll$1(DefaultCachedClasspathTransformer.java:316)
        ... 2 more
Caused by: java.io.IOException: Failed to process the entry 'org/apache/lucene/gradle/ProfileResults.class' from '/Users/chegar/git/lucene_msscorer/lucene/buildSrc/build/libs/buildSrc.jar'
        at org.gradle.internal.classpath.InstrumentingClasspathFileTransformer$BaseTransformation.visitEntry(InstrumentingClasspathFileTransformer.java:255)
        at org.gradle.internal.classpath.InstrumentingClasspathFileTransformer$BaseTransformation.lambda$visitEntries$1(InstrumentingClasspathFileTransformer.java:240)
        at org.gradle.internal.classpath.ClasspathWalker.visitJarContents(ClasspathWalker.java:91)
        at org.gradle.internal.classpath.ClasspathWalker.visit(ClasspathWalker.java:55)
        at org.gradle.internal.classpath.InstrumentingClasspathFileTransformer$BaseTransformation.visitEntries(InstrumentingClasspathFileTransformer.java:239)
        at org.gradle.internal.classpath.InstrumentingClasspathFileTransformer$BaseTransformation.lambda$transform$0(InstrumentingClasspathFileTransformer.java:230)
        at org.gradle.internal.classpath.ClasspathBuilder.buildJar(ClasspathBuilder.java:68)
        at org.gradle.internal.classpath.ClasspathBuilder.jar(ClasspathBuilder.java:55)
        ... 9 more
Caused by: java.lang.IllegalArgumentException: Unsupported class file major version 66
        at org.objectweb.asm.ClassReader.<init>(ClassReader.java:199)
        at org.objectweb.asm.ClassReader.<init>(ClassReader.java:180)
        at org.objectweb.asm.ClassReader.<init>(ClassReader.java:166)
        at org.gradle.internal.classpath.InstrumentingClasspathFileTransformer$BaseTransformation.processClassFile(InstrumentingClasspathFileTransformer.java:268)
        at org.gradle.internal.classpath.InstrumentingClasspathFileTransformer$TransformationForAgent.processClassFile(InstrumentingClasspathFileTransformer.java:430)
        at org.gradle.internal.classpath.InstrumentingClasspathFileTransformer$BaseTransformation.visitEntry(InstrumentingClasspathFileTransformer.java:248)
        ... 16 more
```

dweiss (approved):

ChrisHegarty:
I'm planning on backporting this to the 9.x branch, unless there are concerns.

uschindler:
> @uschindler I was motivated to do this upgrade since stumbling into issues trying to verify the 9.11 RC1 build with `--test-alternative-java <path_to_jdk_22>`, and kept running into silly configuration issues. I think that my local env does not have the correct variables set or something, but it'll be more straightforward if we can just set JAVA_HOME to 22.

Yes, the smoker runs the whole testsuite including gradle with all the alternativive JDKs. This was the reason why I did not test Java 22 in my release checks. I ran the tests separately.

+1, all fine here. But I have not tested it, I trust that all gradle paths were executed.

uschindler:
The problem with gradle updates is always: theres is always a small piece of build logic not tested that breaks. Its hard to test the whole thing unless you also run regenerate or do a release.

uschindler (approved):

ChrisHegarty:
I ran regenerate - it found another small issue with InstallationLocation, now fixed.

uschindler:
> I ran regenerate - it found another small issue with InstallationLocation, now fixed.

Yeah, I think the biggest problem is to find one task which executes everything in the whole Gradle Build... We would need some Gradle Coverage Report!

dweiss:
> Yeah, I think the biggest problem is to find one task which executes everything in the whole Gradle Build... We would need some Gradle Coverage Report!

I don't think we can find one particular task to do that, but we can try to include common stuff - see
https://github.com/apache/lucene/pull/13456

jpountz:
I'm getting errors since upgrading to Gradle 8.8 that I can't figure out how to fix. When running with `--stacktrace`, I see the following error that seems to be the root cause. I was not getting it before 8.8. I'll keep digging, but though I'd mention it in case somebody else runs into the same problem or knows how to work around it.

```
Caused by: java.nio.file.FileSystemException: /home/jpountz/.gradle/.tmp/tmp-dsl17753815434809694067.tmp/publications_maven_4o6zjie60n1r0kcieidajnrm8$_run_closure1$_closure2$_closure9$_closure12$_closure13$_closure14$_closure15$_closure16$_closure17.class: File name too long
        at org.gradle.internal.classpath.InPlaceClasspathBuilder$DirectoryEntryBuilder.put(InPlaceClasspathBuilder.java:165)
        at org.gradle.internal.classpath.transforms.BaseClasspathElementTransform.processClassFile(BaseClasspathElementTransform.java:120)
        at org.gradle.internal.classpath.transforms.BaseClasspathElementTransform.visitEntry(BaseClasspathElementTransform.java:94)
        ... 20 more

```

uschindler:
Hi,
on my windows pc (which also has limitations on length of path names), I have seen not issues. On my new laptop I havent perl installed at moment, so building docs failed, but everything else worked.
Maybe you remove all caches in ~/.gradle ? It is always good to clean that up because that directory also collects old gradle versions and so on.

uschindler:
i still see some problems during tests regarding vector module. I have to check what's going on. Possibly some test system properties issue. It looks like the vector module is not always readable due to some strange behaviour with module names.

Let's find that issue. Working on it:

```
> Task :lucene:expressions:test
WARNING: A command line option has enabled the Security Manager
WARNING: The Security Manager is deprecated and will be removed in a future release
WARNING: A command line option has enabled the Security Manager
WARNING: The Security Manager is deprecated and will be removed in a future release
WARNING: A command line option has enabled the Security Manager
WARNING: The Security Manager is deprecated and will be removed in a future release
WARNING: A command line option has enabled the Security Manager
WARNING: The Security Manager is deprecated and will be removed in a future release

Juni 07, 2024 12:07:56 PM org.apache.lucene.internal.vectorization.VectorizationProvider lookup
WARNUNG: Java vector incubator module is not readable. For optimal vector performance, pass '--add-modules jdk.incubator.vector' to enable Vector API.
```

And this one in codecs:test , not sure why this happens:

```
WARNING: Unknown module: org.apache.lucene.core specified to --enable-native-access
```

All a bit strange.

uschindler:
Not sure what's the problem with both projects; in the test configuration all looks fine. Maybe we have not seen those messages before as Gradle did not print them. New version seems to pass logger event to console:

```groovy
      // Enable the vector incubator module on supported Java versions:
      if (rootProject.vectorIncubatorJavaVersions.contains(rootProject.runtimeJavaVersion)) {
        jvmArgs '--add-modules', 'jdk.incubator.vector'
      }
      
      jvmArgs '--enable-native-access=' + (project.path == ':lucene:core' ? 'ALL-UNNAMED' : 'org.apache.lucene.core')
```

I will check this out, it could be that the classfification module has different dependency to lucene core. It uses "moduleApi" instead of "moduleImplementation". Maybe that's the issue. @dweiss ?

uschindler:
If the "moduleApi" dependency is wanted then it makes sense to me. But then we should only add the jvmArg if the module is linked at all. Not sure how to test this.

So basically I am at moment only usure why it complain that vector incubator is not enabled for some Gradle projects.

uschindler:
This seems to be a message out of context. When running expressions tests it has the vector incubator enabled.

The message comes from somewhere about the not readable module comes from somewhere else....

> ```
> > Task :lucene:expressions:test
> WARNING: Using incubator modules: jdk.incubator.vector
> WARNING: A command line option has enabled the Security Manager
> WARNING: The Security Manager is deprecated and will be removed in a future release
> WARNING: Using incubator modules: jdk.incubator.vector
> WARNING: A command line option has enabled the Security Manager
> WARNING: The Security Manager is deprecated and will be removed in a future release
> WARNING: Using incubator modules: jdk.incubator.vector
> WARNING: A command line option has enabled the Security Manager
> WARNING: The Security Manager is deprecated and will be removed in a future release
> WARNING: Using incubator modules: jdk.incubator.vector
> WARNING: A command line option has enabled the Security Manager
> WARNING: The Security Manager is deprecated and will be removed in a future release
> 
> Juni 07, 2024 12:07:56 PM org.apache.lucene.internal.vectorization.VectorizationProvider lookup
> WARNUNG: Java vector incubator module is not readable. For optimal vector performance, pass '--add-modules jdk.incubator.vector' to enable Vector API.
> ```
> All a bit strange.

Mega strange....

uschindler:
When checking jenkisn builds, the message...

```
WARNING: Unknown module: org.apache.lucene.core specified to --enable-native-access
```

...appears on random gradle projects (different ones on each run). So it has nothing to do with `moduleImplementation` vs. `moduleApi` dependency. It looks like Gradle sometimes forgets some command line parameters when executing tests. If this is a real bug with the test runner, we should revert and wait for a bugfix.

uschindler:
I think I know the problem..... It looks like Gradle test runner loads those classes for some checks outside of tests. This is why the message appear randomly in the output.

In the past we had that issue when Gradle was configured to "detect" tests  automatically from bytecode instead of filename. Looks like this is a similar issue here. Will dig into changes of Gradle Test runner and disable loading any test classes outside of test runner.

uschindler:
Most error messages regarding the modules are also seen on Gradle 8.6, so its not a new issue. Sorry for false alarm.

I will still investigate this. It seems to have indeed something to do with moduleApi vs moduleImplementation

jpountz:
FWIW moving my `~/.gradle` directory to a filesystem that is not encrypted worked around the problem.

dweiss:
the api vs implementation difference is about how configurations are inherited and then exported. See [1]. I'm not sure what's happening with regard to the warning - let me know if you'd like me to look into it so we're not duplicating the effort.

[1] https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_configurations_graph

dweiss:
What's the command/ jdk that produces these warnings for you, Uwe?

uschindler:
> What's the command/ jdk that produces these warnings for you, Uwe?

Java 21 with "Gradlew test". You see it also on Policeman Jenkins on every Lucene main build.

I can reproduce it also when running tests locally for a specific project

dweiss:
These messages are sys-outed from each forked test runner and echoed as 'system.err' by the main gradle process. You can see them in the debug log if you run gradle with -debug:
```
2024-06-07T21:11:59.792+0200 [ERROR] [system.err] WARNING: Using incubator modules: jdk.incubator.vector
2024-06-07T21:11:59.831+0200 [ERROR] [system.err] WARNING: A command line option has enabled the Security Manager
2024-06-07T21:11:59.831+0200 [ERROR] [system.err] WARNING: The Security Manager is deprecated and will be removed in a future release
```
and when you launch the full runner executable (again - provided in debug logs), you'll see the same thing:
```
WARNING: A command line option has enabled the Security Manager
WARNING: The Security Manager is deprecated and will be removed in a future release
```
Perhaps gradle finally improved the runner to echo jvm messages emitted before the tests commence (which should help with security log audits...)?

I don't think there's anything we can do about these messages - they're really emitted by the jvm that runs tests.

dweiss:
And looking at JDK sources, I don't think there is a way to dodge those warnings.
```
> java --add-modules jdk.incubator.vector -version
openjdk version "21" 2023-09-19 LTS
OpenJDK Runtime Environment Temurin-21+35 (build 21+35-LTS)
OpenJDK 64-Bit Server VM Temurin-21+35 (build 21+35-LTS, mixed mode, sharing)
```

dweiss:
I also see these warnings, which you've mentioned:
```
2024-06-07T14:17:29.1508657Z WARNING: Java vector incubator module is not readable. For optimal vector performance, pass '--add-modules jdk.incubator.vector' to enable Vector API.
```
Looking whether I can figure out why they're there.

uschindler:
> These messages are sys-outed from each forked test runner and echoed as 'system.err' by the main gradle process. You can see them in the debug log if you run gradle with -debug:
> 
> ```
> 2024-06-07T21:11:59.792+0200 [ERROR] [system.err] WARNING: Using incubator modules: jdk.incubator.vector
> 2024-06-07T21:11:59.831+0200 [ERROR] [system.err] WARNING: A command line option has enabled the Security Manager
> 2024-06-07T21:11:59.831+0200 [ERROR] [system.err] WARNING: The Security Manager is deprecated and will be removed in a future release
> ```
> 
> and when you launch the full runner executable (again - provided in debug logs), you'll see the same thing:
> 
> ```
> WARNING: Using incubator modules: jdk.incubator.vector
> WARNING: A command line option has enabled the Security Manager
> WARNING: The Security Manager is deprecated and will be removed in a future release
> ```
> 
> Perhaps gradle finally improved the runner to echo jvm messages emitted before the tests commence (which should help with security log audits...)?
> 
> I don't think there's anything we can do about these messages - they're really emitted by the jvm that runs tests.

I don't care about those warnings they are expected. See the warnings I posted above. They make no sense, because it says that incubator is enabled but later it says not readable. But this appears in random m modules. I think this comes not from running our tests, but Gradle loading some class in it's own JVM and triggering the message on clinit.

The other one with the warning about the `--add-modules` is the only one that has to do with moduleAPI vs moduleImplementation. It looks like the lucene.core java module is not part of the startup parameters but it wants to enable native access for it.

So please only look at those 2 messages:
```
Juni 07, 2024 12:07:56 PM org.apache.lucene.internal.vectorization.VectorizationProvider lookup
WARNUNG: Java vector incubator module is not readable. For optimal vector performance, pass '--add-modules jdk.incubator.vector' to enable Vector API.

WARNING: Unknown module: org.apache.lucene.core specified to --enable-native-access
```

dweiss:
It appears random because there are multiple runners and messages are proxied back to the console interleaved with everything else. I managed to debug this by running with ``--max-workers 1`` and in debug mode. Then I can see that:
```
2024-06-07T21:46:31.079+0200 [DEBUG] [org.gradle.api.internal.tasks.testing.junit.AbstractJUnitTestClassProcessor] Executing test class org.apache.lucene.codecs.TestCodecLoadingDeadlock
2024-06-07T21:46:31.293+0200 [ERROR] [system.err] Jun 07, 2024 9:46:31 PM org.apache.lucene.internal.vectorization.VectorizationProvider lookup
2024-06-07T21:46:31.293+0200 [ERROR] [system.err] WARNING: Java vector incubator module is not readable. For optimal vector performance, pass '--add-modules jdk.incubator.vector' to enable Vector API.
```
TestCodecLoadingDeadlock forks a subprocess, which inherits IO - this bypasses System.out/err, which I presume is redirected for tests, and logs directly to the worker's system error descriptor, which in turn shows up on the console.

The other message is from :lucene:codecs:test. The way tests are executed for that project is that it runs without modules - see modules.gradle:
```
      // TODO: the tests of these projects currently don't compile or work in
      // module-path mode. Make the modular paths extension use class path only.
      if (sourceSet.name == SourceSet.TEST_SOURCE_SET_NAME && project.path in [
        // Circular dependency between artifacts or source set outputs,
        // causing package split issues at runtime.
        ":lucene:core",
        ":lucene:codecs",
        ":lucene:test-framework",
      ]) {
        modularPaths.mode = ModularPathsExtension.Mode.CLASSPATH_ONLY
      }
```

Yet, the defaults-tests.gradle appends enable-native-access and the JVM complains about it (rightfully).

So it seems like everything is actually explainable. It is really annoying that forked test JVMs cram their syserrs directly to gradle console - makes debugging very difficult. But at least they show up, which is an improvement (?).

I think a follow-up issue should be to correct the test to not inherit IO but redirect somewhere else and to either enable modular mode for the codecs module or add proper exclusion to enable-native-access.

dweiss:
I filed https://github.com/apache/lucene/pull/13471 to quiet down those two warnings.

dweiss:
Feel free to push to that PR, Uwe. I think I'm done for the day.

PR #12156, opened 2023-02-19
Remove custom TermInSetQuery implementation in favor of extending MultiTermQuery

gsmiller:
## Description

`TermInSetQuery`'s implementation is more-or-less an exactly copy of `MultiTermQuery` +  `MultiTermQueryConstantScoreWrapper`. This PR removes the custom `TermInSetQuery` implementation in favor of extending `MultiTermQuery` with `MultiTermQueryConstantScoreWrapper` as the default rewrite behavior.

One nice benefit of this (beyond code cleanup) is that different rewrite methods can be provided for different behavior. Specifically, we can leverage `DocValuesRewriteMethod` to completely replace `SortedSetDocValuesSetQuery` (I would propose doing that in a follow up PR once we're happy with this change).

A couple notes about the change:
1. I needed to give `MultiTermQueryConstantScoreWrapper` a `ScoreSupplier` implementation so `TermInSetQuery` can still be used efficiently with `IndexOrDocValuesQuery`. This also required a new (optional) public API in `MultiTermQuery` where queries can expose their number of terms (where applicable).
2. ~~Retaining the existing `TermInSetQuery#rewrite` behavior was slightly non-trivial. I don't really like how I've done this in the PR, but I also can't think of a better solution. I'm not actually convinced we need to retain this, but I'm not sure if client code might rely on the existing behavior in places. Even if that's the case, it might not be a good enough argument to keep it, but I did find a breaking unit test if I didn't keep it. (`TestPresearcherMatchCollector.testMatchCollectorShowMatches` assumes the query will get rewritten to a BQ). I'm not familiar enough with the `monitor` package to assess if it would be reasonable to just change the unit test, or if there's something more important going on here. If someone is more familiar and/or has opinions on the need to retain the existing "rewrite" behavior, I'd love some feedback.~~

## Performance
When this has been discussed in the past, there's been an open question around performance since the term intersection happens a bit differently (`seekCeil` vs. `seekExact`). I ran some benchmarks using a one-off tool (similar to #12151) and found no noticeable regressions/issues. Here's the output of my tool (which you can check out here: 
[TiSBench.java.txt](https://github.com/apache/lucene/files/10775443/TiSBench.java.txt) )

### All Country Code Filter Terms
| Approach     | Large Lead Terms | Medium Lead Terms | Small Lead Terms | No Lead Terms |
|--------------|------------------|-------------------|------------------|---------------|
| TiS Original | 206.12           | 203.00            | 201.95           | 119.97        |
[...]

PR #13454, opened 2024-06-05
MultiTermQuery return null for ScoreSupplier

mayya-sharipova:
MultiTermQuery return null for ScoreSupplier if there are no terms in an index that
 match query terms.

With the introduction of PR #12156 we saw degradation in performance of bool queries where one of the mandatory clauses is a TermInSetQuery with query terms not present in the field. Before for such cases TermsInSetQuery returned null for ScoreSupplier which would shortcut the whole bool query.

This PR adds ability for MultiTermQuery to return null for ScoreSupplier if a field doesn't contain any query terms.

Relates to PR #12156


mayya-sharipova:
@gsmiller @jpountz  I don't know all the implications for this, so your feedback is welcome.

lucene/core/src/java/org/apache/lucene/search/AbstractMultiTermQueryConstantScoreWrapper.java
```diff
@@ -153,39 +153,28 @@ protected abstract WeightOrDocIdSetIterator rewriteInner(
         List<TermAndState> collectedTerms)
         throws IOException;
 
-    private WeightOrDocIdSetIterator rewrite(LeafReaderContext context, Terms terms)
```
jpountz, line 156: I wonder if you could better encapsulate the change you're after by making this method return a `IOSupplier<WeightOrDocIdSetIterator>`, where `collectTerms` is called when retrieving the IOSupplier, but `rewriteInner` is called when calling `IOSupplier#get`.

jpountz (commented):
This is an interesting change. Instead of pushing the full cost of rewriting the query to `ScorerSupplier#get`, you are collecting the first 16 terms when calling `Weight#scorerSupplier` and the rest (potentially thousands of terms) when calling `ScorerSupplier#get`. This in-turn allows making better decisions for only a small amount of work pushed to `Weight#scorerSupplier`.

I left some style comments but I like the approach.

lucene/core/src/java/org/apache/lucene/search/AbstractMultiTermQueryConstantScoreWrapper.java
mayya-sharipova: @jpountz Great suggestion! Addressed!

mayya-sharipova:
@jpountz Thanks for your feedback, I addressed it, and the code looks much better with your suggestions!

jpountz (approved):

lucene/core/src/java/org/apache/lucene/search/AbstractMultiTermQueryConstantScoreWrapper.java
```diff
@@ -162,21 +163,29 @@ private WeightOrDocIdSetIterator rewrite(LeafReaderContext context, Terms terms)
...
-        for (TermAndState t : collectedTerms) {
-          final TermStates termStates = new TermStates(searcher.getTopReaderContext());
-          termStates.register(t.state, context.ord, t.docFreq, t.totalTermFreq);
-          bq.add(new TermQuery(new Term(q.field, t.term), termStates), BooleanClause.Occur.SHOULD);
-        }
-        Query q = new ConstantScoreQuery(bq.build());
-        final Weight weight = searcher.rewrite(q).createWeight(searcher, scoreMode, score());
-        return new WeightOrDocIdSetIterator(weight);
+      boolean collectResult = collectTerms(fieldDocCount, termsEnum, collectedTerms);
+      if (collectResult && collectedTerms.size() == 0) {
```
gsmiller, line 167: super minor: I tend to prefer `collectedTerms.isEmpty()` instead of `.size() == 0` as a slightly more idiomatic way to express this, but not a strong opinion :) Feel free to ignore.

lucene/CHANGES.txt
```diff
@@ -249,7 +249,7 @@ Improvements
 
 Optimizations
 ---------------------
-(No changes)
+* GITHUB#13454: MultiTermQuery return null for ScoreSupplier (Mayya Sharipova)
```
gsmiller, line 252: Maybe `MultiTermQuery returns null ScoreSupplier in cases where no query terms are present in the index segment` to avoid confusion? I initially read this as MultiTermQuery always returning null :)

gsmiller (approved):
Seems like a reasonable change to me as well. Thanks!

lucene/CHANGES.txt
mayya-sharipova: Thanks for the suggestion, addressed

lucene/core/src/java/org/apache/lucene/search/AbstractMultiTermQueryConstantScoreWrapper.java
mayya-sharipova: Thanks for the suggestion, addressed.

PR #13456, opened 2024-06-05
Add a github workflow that checks common (and less common) gradle tasks when gradle version is changed

dweiss:
This is mostly an automated sanity check that common (and less common) tasks work after an upgrade. This is what the output looks like:

https://github.com/dweiss/lucene/actions/runs/9389691271


dweiss:
This is a follow up to #13453 - perhaps it'll make life a bit easier.

ChrisHegarty:
I think that this is good. Would it run on all CI jobs, or you somehow opt-in?   

Running with `RUNTIME_JAVA_HOME` does have a slightly different code path in the gradle build files. Not sure if that could be easily covered.

I just merged the gradle wrapper update to _main_ and _branch_9x_, so maybe sync up this PR branch to ensure that all is still good. :-)

.github/workflows/run-checks-gradle-upgrade.yml
```diff
@@ -0,0 +1,88 @@
...
+jobs:
+  gradleSanityCheck:
+    name: "Run tasks (java: ${{ matrix.java-version }}, alt-java: ${{ matrix.uses-alt-java }})"
+    timeout-minutes: 30
+
+    strategy:
+      matrix:
+        os: [ ubuntu-latest ]
+        java-version: [ '22' ]
+        uses-alt-java: [ true, false ]
```
dweiss, line 34: this is the configuration matrix switch that is used to control whether the build runs with a separate RUNTIME_JAVA_HOME or without it.

.github/workflows/run-checks-gradle-upgrade.yml
```diff
@@ -0,0 +1,88 @@
...
+      ALT_JAVA_DIR: /tmp/alt-java
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/prepare-for-build
+        with:
+          java-version: ${{ matrix.java-version }}
+
+      - name: Set up RUNTIME_JAVA_HOME variable
+        if: ${{ matrix.uses-alt-java }}
```
dweiss, line 48: If we're running with the "uses-alt-java" switch, we set up an "alternative" java distribution under /tmp and use it to execute the same set of gradle tasks.

dweiss:
I've added a matrix configuration so that one job runs with RUNTIME_JAVA_HOME and the other without it. Seems to be working and increases test coverage.

image

dweiss:
I think this is ready and works reasonably well - it caught #13458 already.

ChrisHegarty (approved):
Thanks for this @dweiss  LGTM (and thanks for catching and fixing the issue I caused by updating the wrapper)

PR #13460, opened 2024-06-06
Merge related HashMaps in FieldInfos#FieldNumbers into one map

### LINKED ISSUE #13459: Should FieldInfo hold one map with index properties instead of a map for each property?
We currently have in FieldInfos the following maps:

```
     this.indexOptions = new HashMap<>();
      this.docValuesType = new HashMap<>();
      this.dimensions = new HashMap<>();
      this.vectorProps = new HashMap<>();
```

This maps are updated and read together so it feels wasteful to have separate maps for each one. They can get pretty big too when there is lost of fields. I wonder if we can simplified them into one map.

iverase:
This commit merges indexOptions, DocValuesType, FieldDimensions and FieldVectorProperties into one map by introducing the record FieldProperties.

In addition it moves to a record the internal classes FieldDimensions and FieldProperties.

fixes https://github.com/apache/lucene/issues/13459

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
+++ b/lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
@@ -643,12 +643,10 @@ public class FieldInfos implements Iterable<FieldInfo> {
      *     {@code dvType} returns a new FieldInfo based based on the options in global field numbers
      */
     FieldInfo constructFieldInfo(String fieldName, DocValuesType dvType, int newFieldNumber) {
-      Integer fieldNumber;
+      FieldProperties fieldProperties;
       synchronized (this) {
-        fieldNumber = fieldProperties.get(fieldName).number;
+        fieldProperties = this.fieldProperties.get(fieldName);
       }
-      if (fieldNumber == null) return null;
-      FieldProperties fieldProperties = this.fieldProperties.get(fieldName);
       if (fieldProperties == null) return null;
       DocValuesType dvType0 = fieldProperties.docValuesType;
       if (dvType != dvType0) return null;
--- a/lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
+++ b/lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
@@ -354,11 +354,15 @@ public class FieldInfos implements Iterable<FieldInfo> {
       VectorEncoding vectorEncoding,
       VectorSimilarityFunction similarityFunction) {}
 
+  private record IndexOptionsProperties(boolean storeTermVectors, boolean omitNorms) {}
+
   // We use this to enforce that a given field never
   // changes DV type, even across segments / IndexWriter
   // sessions:
   private record FieldProperties(
+      int number,
       IndexOptions indexOptions,
+      IndexOptionsProperties indexOptionsProperties,
       DocValuesType docValuesType,
       FieldDimensions fieldDimensions,
       FieldVectorProperties fieldVectorProperties) {}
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
```diff
@@ -403,11 +385,8 @@ static final class FieldNumbers {
     FieldNumbers(String softDeletesFieldName, String parentFieldName) {
       this.nameToNumber = new HashMap<>();
       this.numberToName = new IntObjectHashMap<>();
-      this.indexOptions = new HashMap<>();
-      this.docValuesType = new HashMap<>();
-      this.dimensions = new HashMap<>();
-      this.vectorProps = new HashMap<>();
       this.omitNorms = new HashMap<>();
+      this.fieldProperties = new HashMap<>();
```
iverase, line 389: I wonder if we should add omitNorms and storeTermVectors too? they are only added when they exist.

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
```diff
@@ -701,9 +684,7 @@ synchronized Set<String> getFieldNames() {
     synchronized void clear() {
       numberToName.clear();
       nameToNumber.clear();
-      indexOptions.clear();
-      docValuesType.clear();
-      dimensions.clear();
+      fieldProperties.clear();
```
iverase, line 687: There is maybe a bug that we don't clear vectorProps hashmap?

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
iverase: We can add nameToNumber too?

jpountz: Yes, it looks like we can add all immutable attributes there indeed.

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
jpountz: This is a bug indeed. So if you call deleteAll, then you won't be able to index vectors with different options. I don't think it's a common use-case, which is likely why it's never been reported.

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
iverase: I added those attributes to FieldProperties. I introduced a `IndexOptionsProperties` record to capture omitNorms and  storeTermVectors.

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
```diff
@@ -664,14 +643,13 @@ synchronized void verifyOrCreateDvOnlyField(
      *     {@code dvType} returns a new FieldInfo based based on the options in global field numbers
      */
     FieldInfo constructFieldInfo(String fieldName, DocValuesType dvType, int newFieldNumber) {
-      Integer fieldNumber;
+      FieldProperties fieldProperties;
       synchronized (this) {
-        fieldNumber = nameToNumber.get(fieldName);
+        fieldProperties = this.fieldProperties.get(fieldName);
```
iverase, line 648: This feels like we might change the behaviour here? Not sure why we are getting the number here on a synchronise block.

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
jpountz: Should `IndexOptionsProperties` also capture `IndexOptions`?

jpountz: Oh I see, no it should not, because we set it to `null` when `IndexOptions` is `NONE`?

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
jpountz: This entire class is shared by the IndexWriter so access needs to be protected. The fact that access to `docValuesType` was not protected is more surprising to me, it seems to rely on subtle happens-before guarantees? Anyway, I like that your change is simplifying this.

jpountz (approved):

easyice (approved):

iverase:
This class seems to have change quite a bit and some methods did not have a clear path forward for backportso I set the milestone for this change to 10.0.0

PR #13464, opened 2024-06-06
Reduce the heap use of BKDReader instances

original-brownbear:
We consume a lot of memory for the `indexIn` slices. If `indexIn` is of type `MemorySegmentIndexInput` the overhead of keeping loads of slices around just for cloning is far higher than the extra 12b per reader this adds (the slice description alone often costs a lot). In a number of Elasticsearch example uses with high segment counts I investigated, this change would save up to O(GB) of heap.


iverase (approved):
makes sense to me

jpountz (approved):
Agreed that the intermediate slice is not helping here. The new code is also consistent with e.g. doc values, which only create slices when pulling doc-values instances.

PR #13467, opened 2024-06-07
Honor read advice on compound files.

jpountz:
This includes the following changes:
 - New `IndexInput#slice(String, long, long, ReadAdvice)` API that allows creating slices with different advices.
 - ~~New `CompoundFormat` that aligns start offsets on multiples of 4,096 to prevent multiple sub files from being stored on the same page. This in-turn allows `madvise` to give different advices to different files. This is what causes the large diff due to the dance required when creating a new codec.~~
 - `PosixNativeAccess` now explicitly sets `MADV_NORMAL` when called with `ReadAdvice.NORMAL`. This is required to be able to override a `RANDOM` advice of a compound file with a `NORMAL` advice of a sub file of this compound file.
 - `PosixNativeAccess` now only ignores the first page if a range of bytes starts before the `MemorySegment` instead of the whole range.


jpountz:
@uschindler I'm curious to get your opinion on this approach.

uschindler:
I had no time to review. But why do we need a new format? The alignment used during writing is not part of for at spec. So if we switch to a different alignment for subfiles, it's still readable as offsets in directorsly are absolute.

jpountz:
You are right that the format is still compatible, I'm doing this mostly out of precaution and to keep testing old codecs that had different alignment easy.

jpountz:
I pushed a simpler approach that should probably be good enough. It simply ignores the first page from the madvise call in case the start offset of a page is before the start of the memory segment.

uschindler (approved):

PR #13471, opened 2024-06-07
Silence odd test runner warnings after gradle upgrade

dweiss:
Gradle started to proxy syserr messages from test runner JVMs. We can see some odd messages like:
```
2024-06-07T21:46:31.293+0200 [ERROR] [system.err] WARNING: Java vector incubator module is not readable. For optimal vector performance, pass '--add-modules jdk.incubator.vector' to enable Vector API.
...
WARNING: Unknown module: org.apache.lucene.core specified to --enable-native-access
```

This PR tries to silence both. A proper fix would be to correct modularity for the codec subproject, but it's a different beast.

### DIFF
--- a/gradle/testing/defaults-tests.gradle
+++ b/gradle/testing/defaults-tests.gradle
@@ -135,7 +135,9 @@ allprojects {
       
       jvmArgs '--enable-native-access=' + (project.path in [
               ':lucene:core',
-              ':lucene:codecs'
+              ':lucene:codecs',
+              ":lucene:distribution.tests",
+              ":lucene:test-framework"
       ] ? 'ALL-UNNAMED' : 'org.apache.lucene.core')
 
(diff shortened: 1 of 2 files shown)


dweiss:
Verify with:
```
gradlew :lucene:codecs:test --max-workers 1 --tests TestSimpleTextDocValuesFormat
gradlew :lucene:core:test --max-workers 1 --tests TestCodecLoadingDeadlock
```

uschindler:
Will check tomorrow.

dweiss:
No worries, thank you.

uschindler (approved):
It's not nice with those manual exclusions, but works.

Maybe we should better set a ext property on all projects which do not use module system during tests and check for that property.

But this works for now.

dweiss:
Thanks. I've backported it to 9x as well.

Issue #13478, opened 2024-06-11
Strange ConcurrentMergeScheduler behavior with intra-merge threads


benwtrent:
### Description

It was noticed that the CMS intra-merge behavior was not fully tested. In an effort to do this, a change to override when the intra-merge scheduler is used has been drafted. https://github.com/apache/lucene/pull/13475

This PR has exposed many existing assertions that may all be weird testing failures. But a couple are more worrisome. 

In particular: 

```
./gradlew test --tests TestCodecHoldsOpenFiles.test -Dtests.seed=AA3C2FCDA0773874 -Dtests.locale=ar-PS -Dtests.timezone=Asia/Magadan -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```

Will periodically fail with an NPE on the `sort`:

```
        org.apache.lucene.index.MergePolicy$MergeException: java.lang.NullPointerException: Cannot invoke "org.apache.lucene.search.Sort.getSort()" because the return value of "org.apache.lucene.index.LeafMetaData.getSort()" is null
            at __randomizedtesting.SeedInfo.seed([AA3C2FCDA0773874]:0)
            at app//org.apache.lucene.index.ConcurrentMergeScheduler.handleMergeException(ConcurrentMergeScheduler.java:764)
            at app//org.apache.lucene.index.ConcurrentMergeScheduler$MergeThread.run(ConcurrentMergeScheduler.java:756)

            Caused by:
            java.lang.NullPointerException: Cannot invoke "org.apache.lucene.search.Sort.getSort()" because the return value of "org.apache.lucene.index.LeafMetaData.getSort()" is null
                at org.apache.lucene.index.SortingCodecReader.assertCreatedOnlyOnce(SortingCodecReader.java:739)
                at org.apache.lucene.index.SortingCodecReader.getOrCreate(SortingCodecReader.java:721)
                at org.apache.lucene.index.SortingCodecReader.getOrCreateDV(SortingCodecReader.java:759)
                at org.apache.lucene.index.SortingCodecReader$7.getNumeric(SortingCodecReader.java:571)
                at org.apache.lucene.codecs.DocValuesConsumer$1.getNumeric(DocValuesConsumer.java:201)
                at org.apache.lucene.codecs.lucene90.Lucene90DocValuesConsumer$1.getSortedNumeric(Lucene90DocValuesConsumer.java:138)
[...]

PR #13479, opened 2024-06-11
Adjust assertion check to not throw an NPE

benwtrent:
It is possible when checking this assertion, that we could throw an NPE as `metaData.getSort()` can be `null`.

Let's actually allow the assertions do their checks instead of throwing a scary NPE here.

related: https://github.com/apache/lucene/issues/13478


iverase (approved):
Make sense to me, this code is only run on assert so let's fail on the defined asserts.

PR #13481, opened 2024-06-12
Update document by docID

vsop-479:
### Description
Update or delete document by doc id.


vsop-479:
Close since segment merge.

PR #13486, opened 2024-06-14
Only apply deletion one time for unique term update in FrozenBufferedUpdates.applyTermDeletes

vsop-479:
### Description

For unique field like ES’ `uid`, there is only one live doc in all flushed segments. 

We can early terminate `FrozenBufferedUpdates.applyTermDeletes` after this live doc deleted.


lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -1838,6 +1838,20 @@ public long updateDocument(Term term, Iterable<? extends IndexableField> doc) th
...
 
+  /**
+   * Similar to {@link #updateDocuments(Term, Iterable)}, but only apply deletion once for all
+   * flushed segments. This is useful for unique filed like ES's _id.
+   *
+   * @lucene.experimental
+   */
+  // TODO: If it is unnecessary to validate unique constraint, we can add a isUnique setting to
+  // Term.
+  public long updateDocument(Term term, boolean isUnique, Iterable<? extends IndexableField> doc)
```
mikemccand, line 1849: Hmm this makes me nervous -- we are relying on the application to properly claim `isUnique` but the application may get it wrong.  Though I suppose worst case if the application gets it wrong, documents fail to get deleted (just the first occurrence will), not for example index corruption.

Could Lucene maybe track that a field is actually unique internally and then apply this optimization automatically / always correctly?  We may have to tighten the opto to "isUnique and isNonNull (every doc has a value for the field)", which OpenSearch/Elasaticsearch `id` field would meet?

If not for deletes/updates we could compare number of unique terms in the field == `totalTermFreq`.  Or, maybe we'd instead  track that "this field always has exactly one value" so that neither freqs nor positions would have to be indexed for this opto to apply?  And then if number of unique terms is >= `numDocs`, and every doc has one term in this field, then it is a "primary key"?

vsop-479: Thanks for your suggestion @mikemccand .

> maybe we'd instead track that "this field always has exactly one value" 

I will try to track that.
But I have concerns that we may be able to track that a field is unique in a single segment, maybe it will be more complex to track that a field is unique in all flushed segments?

If we could track that, should we just track fields which specified by user (maybe add a `isUnique` setting to `Field`) ? or, track all fields?

vsop-479: > And then if number of unique terms is >= numDocs, and every doc has one term in this field.

This approach does not suitable for deletes/updates too? e.g. term `a`, docs: `0, 1.` term `b`, docs: `2`, we may delete term `b`, but we still write term `b` and its postings. So, in this case, the number of unique terms  is 2, and `numDocs` is 2 too, but it is not `unique`. 

BTW, since we `applyDeletes` before `writeTerms`, should we skip deleted docs when writing `postings`(maybe skip `storedFields` too).


> "this field always has exactly one value" so that neither freqs nor positions would have to be indexed

Since `KeywordField`(primary key main type? ) has exactly one value, and does not index freqs nor positions, can we judge that by filed type?

vsop-479:
It is more complex to apply this optimization with `soft delete`, since we should not skip `soft deleted` docs if we want apply `hard delete` on them.

vsop-479:
> Could Lucene maybe track that a field is actually unique internally and then apply this optimization automatically / always correctly? 

@jpountz 
Do you have any idea about this?

vsop-479:
Since we already have a `PerThreadPKLookup`, and implemented `lookup` in it, maybe implement delete one time in it is an option?

vsop-479:
> BTW, since we applyDeletes before writeTerms, should we skip deleted docs when writing postings(maybe skip storedFields too).

I see this has been discussed in https://issues.apache.org/jira/browse/LUCENE-5693, and  `IDVersionPostingsWriter#startDoc` implemented it (although it was markd as `Won't Fix`).

PR #13487, opened 2024-06-14
Fix backward codec test after introducing the doc values skip index

iverase:
The introduction of the doc values skip index in https://github.com/apache/lucene/pull/13449 broke the backward codec test as those codecs do not support it. This was missed because the test only run in nightly mode.

In order to fix it we move the new tests into a new base class called `BaseDocValuesFormatWithSkipIndexTestCase` that extends `BaseDocValuesFormatTestCase`.


jpountz (approved):
LGTM. Maybe rename `BaseDocValuesFormatWithSkipIndexTestCase` to `BaseDocValuesFormatTestCase` and `BaseDocValuesFormatTestCase` to `LegacyBaseDocValuesFormatTestCase` or something like that to better convey which base test case a "modern" doc-values format is expected to extend?

PR #13488, opened 2024-06-14
GITHUB#13175: Stop double-checking priority queue inserts in some FacetCount classes

### LINKED ISSUE #13175: Stop double-checking priority queue inserts
### Description

There are several occurrences in our code where we check that an incoming value can be inserted in a priority queue and then call `insertWithOverflow`, which re-does the check. Let's stop checking before calling `insertWithOverflow`.

Credit to @epotyom for [noticing this in `AbstractSortedSetDocValueFacetCounts`](https://github.com/apache/lucene/pull/12966#discussion_r1457584536).

slow-J:
Removing 2 cases of bottomX optimizations where insertWithOverflow already handles the check.

Closes #13175 

The diff looks funny because of unnesting, but, per file, I am just removing `bottomOrd` and `bottomCount` and removing an if statement.


lucene/CHANGES.txt
```diff
@@ -273,6 +273,8 @@ Optimizations
 
 * GITHUB#13431: Replace TreeMap and use compiled Patterns in Japanese UserDictionary. (Bruno Roustant)
 
+* GITHUB#13175: Stop double-checking priority queue inserts. (Jakub Slowinski)
```
mikemccand, line 276: Maybe clarify that this just applies to some of our facet counting impls (not to the many other PQs Lucene has/uses)?

mikemccand (approved):

lucene/CHANGES.txt
slow-J: Good point Mike, will do.

slow-J:
Hi @mikemccand can we merge or are we waiting for more reviewers.

mikemccand:
Thanks @slow-J -- I just backported to 9.12 as well.  I had to resolve a few conflicts, maybe have a peek and see if I did it correctly?

slow-J:
Thanks @mikemccand I think the diff for 9.12 looks good.

Issue #13482, opened 2024-06-12
English grammar error in `Field that stores a per-document <code>long</code> values for scoring` 


wurui90:
### Description

I don't think `a` and `values` can be used together

cpoerschke:
Thanks @wurui90 for spotting and reporting this! Would you like to open a pull request to edit the wording?

sakshichourasia4:
@cpoerschke @wurui90 I have raised the pull request to fix this issue, please have a look.
https://github.com/apache/lucene/pull/13490

wurui90:
Thanks for the attention and the PR! Sorry I'm late here. I added a comment in your PR.

PR #13490, opened 2024-06-15
[Issue #13482] Fix english grammar error in Field that stores a per-document <code>long</code> values for scoring 

ghost:
### Description

Fixing English grammar errors. 

Changed line: 
from 
`Field that stores a per-document <code>long</code> values for scoring` 
to 
`Field that stores per-document <code>long</code> values for scoring`


lucene/core/src/java/org/apache/lucene/document/SortedNumericDocValuesField.java
```diff
@@ -21,8 +21,8 @@
 import org.apache.lucene.search.Query;
 
 /**
- * Field that stores a per-document <code>long</code> values for scoring, sorting or value
- * retrieval. Here's an example usage:
+ * Field that stores a per-document <code>long</code> value for scoring, sorting or value retrieval.
```
wurui90, line 24: I think we should remove the `a` here? It's also mentioned in your PR description:

```
from
Field that stores a per-document <code>long</code> values for scoring
to
Field that stores per-document <code>long</code> values for scoring
```

PR #13491, opened 2024-06-17
Remove unused segNo calculation in IndexWriter.doFlush

vsop-479:

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -4271,12 +4271,7 @@ public class IndexWriter
         boolean flushSuccess = false;
         try {
           long seqNo = docWriter.flushAllThreads();
-          if (seqNo < 0) {
-            seqNo = -seqNo;
-            anyChanges = true;
-          } else {
-            anyChanges = false;
-          }
+          anyChanges = seqNo < 0;
           if (!anyChanges) {
             // flushCount is incremented in flushAllThreads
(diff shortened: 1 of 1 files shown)


lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -4271,12 +4271,7 @@ private boolean doFlush(boolean applyAllDeletes) throws IOException {
         boolean flushSuccess = false;
         try {
           long seqNo = docWriter.flushAllThreads();
```
cpoerschke, line 4273:
```suggestion
          final long seqNo = docWriter.flushAllThreads();
```

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -4271,12 +4271,7 @@ private boolean doFlush(boolean applyAllDeletes) throws IOException {
...
-          if (seqNo < 0) {
-            seqNo = -seqNo;
-            anyChanges = true;
-          } else {
-            anyChanges = false;
-          }
+          anyChanges = seqNo < 0;
```
cpoerschke, line 4274: Or perhaps not have `seqNo` even i.e. inline?
```suggestion
          anyChanges = (docWriter.flushAllThreads() < 0);
```

vsop-479: It makes sense, thanks @cpoerschke . I fixed it.

cpoerschke (approved):

PR #13492, opened 2024-06-17
Fix Method declared 'final' in 'final' class in Term.

vsop-479:


benwtrent (approved):

vsop-479:
Thanks for your review @benwtrent .

Would you like to merge this change, if there is no problem with it?

cpoerschke (approved):

PR #13496, opened 2024-06-17
Expose FlatVectorsFormat (#13469)

msokolov:
PR for testing only - changes already pushed to main


msokolov:
I ... did not mean to push that to 9x so precipitously, was trying to wait for tests to complete, sigh. I think it will be OK; promised to follow up if there are test fails though

PR #13498, opened 2024-06-18
Avoid performance regression by constructing lazily the PointTree in NumericComparator

iverase:
We notice a performance regression in Elasticsearch which we could trace back to the changes in https://github.com/apache/lucene/pull/13199. This change makes constructing NumericComparators more expensive die to the construction of a PointTree eagerly in the constructor. This is particular noticeable when the PointTree is not used later and  your constructing several of them.

This. commit proposes to build the PointTree lazily. The original PR proposed it but it was changed due to simplicity.

We are only targeting the 9.x line as main has already diverged and doing it in a different way.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -39,7 +39,8 @@ Optimizations
 
 Bug Fixes
 ---------------------
-(No changes)
+
+* GITHUB#13498: Avoid performance regression by constructing lazily the PointTree in NumericComparator, (Ignacio Vera)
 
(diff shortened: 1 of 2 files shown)


gf2121 (approved):
Thanks @iverase !

jpountz (approved):
LGTM, sorry for making the suggestion of pulling it eagerly! Can you add a comment explaining why it's important for it to be lazily constructed, so that we don't change it back in the future?

PR #13500, opened 2024-06-18
Removing usage of TopScoreDocCollector + TopFieldCollector deprecated methods (#create, #createSharedManager)

### LINKED ISSUE #13499: Remove the @Deprecated methods from TopScoreDocCollector and TopFieldCollector
### Description

These methods were deprecated in https://github.com/apache/lucene/pull/240 which is part of Lucene 10.0.

Since they are not marked for deprecation in Lucene 9.x, they will not be removed with Lucene 10.0 release (unless we want to deprecate them in 9.x in another backport commit?), but I still think we should remove the internal usage so it will be quicker to remove the methods when the time comes.

I'll submit a PR for a 1:1 change to remove the usages.

slow-J:
Removing usage of TopScoreDocCollector + TopFieldCollector deprecated methods (#create, #createSharedManager)

Addresses #13499

Copying description from issue:

These methods were deprecated in https://github.com/apache/lucene/pull/240 which is part of Lucene 10.0.

Since they are not marked for deprecation in Lucene 9.x, they will not be removed with Lucene 10.0 release (unless we want to deprecate them in 9.x in another backport commit?), but I still think we should remove the internal usage so it will be quicker to remove the methods when the time comes.

This is a 1:1 change to remove the usages.

Test command:
`./gradlew clean; ./gradlew tidy; ./gradlew check`


javanna:
heya @slow-J thanks for the PR, and sorry for the delay in reviewing it. I was cleaning things up in this area as well, and ended up making changes to `TestTopDocsMerge` that are causing merge conflicts in your PR. Besides that, I think that this is a good step forward and I am happy to review it and merge it. Could you perhaps remove the changes to TestTopDocsMerge that are no longer necessary?

slow-J:
Thanks @javanna! I fixed the conflicts in `TestTopDocsMerge`, removing the changes that are no longer needed.

javanna (approved):
LGTM

javanna:
Thanks @slow-J !

slow-J:
Thanks @javanna for the review!

PR #13504, opened 2024-06-18
gh-13340: Allow adding a parent field to an index with no fields (#13341) (#13483)

benwtrent:
This backports @msokolov 's change to 9.11.1 and adds a CHANGES entry for the bugfix.


dnhatn (approved):
LGTM.

PR #13506, opened 2024-06-19
Only check liveDocs is null one time in FreqProxTermsWriter.applyDeletes

vsop-479:
### Description

We check if `state.liveDocs` is null, every doc in postings for every deleteTerm in `FreqProxTermsWriter.applyDeletes`.

This change only check it one time.


vsop-479:
@mikemccand 

PR #13508, opened 2024-06-19
expand TestSegmentToThreadMapping coverage w.r.t. (excess) documents per slice

cpoerschke:
https://github.com/apache/lucene/blob/releases/lucene/9.11.0/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java#L333-L376

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/index/TestSegmentToThreadMapping.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestSegmentToThreadMapping.java
@@ -172,7 +172,7 @@ public class TestSegmentToThreadMapping extends LuceneTestCase {
     leafReaderContexts.add(new LeafReaderContext(secondMediumSegmentReader));
     leafReaderContexts.add(new LeafReaderContext(thirdMediumSegmentReader));
 
-    IndexSearcher.LeafSlice[] resultSlices = IndexSearcher.slices(leafReaderContexts, 250_000, 5);
+    IndexSearcher.LeafSlice[] resultSlices = computeSlices(leafReaderContexts);
 
     assertTrue(resultSlices.length == 1);
 
@@ -201,7 +201,7 @@ public class TestSegmentToThreadMapping extends LuceneTestCase {
     leafReaderContexts.add(new LeafReaderContext(seventhLargeSegmentReader));
     leafReaderContexts.add(new LeafReaderContext(eigthLargeSegmentReader));
 
-    IndexSearcher.LeafSlice[] resultSlices = IndexSearcher.slices(leafReaderContexts, 250_000, 5);
+    IndexSearcher.LeafSlice[] resultSlices = computeSlices(leafReaderContexts);
 
     assertTrue(resultSlices.length == 3);
 
@@ -226,7 +226,7 @@ public class TestSegmentToThreadMapping extends LuceneTestCase {
     leafReaderContexts.add(new LeafReaderContext(secondMediumSegmentReader));
     leafReaderContexts.add(new LeafReaderContext(thirdMediumSegmentReader));
 
-    IndexSearcher.LeafSlice[] resultSlices = IndexSearcher.slices(leafReaderContexts, 250_000, 5);
+    IndexSearcher.LeafSlice[] resultSlices = computeSlices(leafReaderContexts);
 
     assertTrue(resultSlices.length == 3);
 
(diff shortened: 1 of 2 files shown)


cpoerschke:
Added potential reviewers based on https://issues.apache.org/jira/browse/LUCENE-8757 history, thank you.

lucene/core/src/test/org/apache/lucene/index/TestSegmentToThreadMapping.java
```diff
@@ -292,8 +292,43 @@ public void testRandom() {
...
     }
 
-    IndexSearcher.LeafSlice[] resultSlices = IndexSearcher.slices(leafReaderContexts, 250_000, 5);
+    IndexSearcher.LeafSlice[] resultSlices = computeSlices(leafReaderContexts);
 
     assertTrue(resultSlices.length > 0);
   }
+
+  public void testTwoPreciseSlices() {
+    LeafReader firstSegmentReader = dummyIndexReader(250_000);
```
mikemccand, line 301: Should we reference `IndexSearcher.MAX_DOCS_PER_SLICE` instead of hardwiring the `250_000` here?

lucene/core/src/test/org/apache/lucene/index/TestSegmentToThreadMapping.java
```diff
@@ -292,8 +292,43 @@ public void testRandom() {
...
+    leafReaderContexts.add(new LeafReaderContext(secondSegmentReader));
+
+    IndexSearcher.LeafSlice[] resultSlices = computeSlices(leafReaderContexts);
+
+    assertTrue(resultSlices.length == 1);
+    assertTrue(resultSlices[0].leaves.length == 2);
+  }
+
+  private static IndexSearcher.LeafSlice[] computeSlices(List<LeafReaderContext> leaves) {
+    final int maxDocsPerSlice = 250_000;
```
mikemccand, line 315: Also here?

lucene/core/src/test/org/apache/lucene/index/TestSegmentToThreadMapping.java
```diff
@@ -292,8 +292,43 @@ public void testRandom() {
...
+    final int maxSegmentsPerSlice = 5;
```
mikemccand, line 316: And use `IndexSearcher.MAX_SEGMENTS_PER_SLICE` here?

mikemccand (approved):
Thanks @cpoerschke -- looks great!  I left minor refactoring comments...

cpoerschke:
Thanks @mikemccand for reviewing!

Referencing `IndexSearcher.MAX_*_PER_SLICE` would require the constants to be made public, but that might be a good thing actually e.g. for deriving classes. (I've tried to move the test from `index` to `search` package which would allow package instead of public visibility for the constants but then access to one of the other classes used in the test was not available.)

Using `IndexSearcher.MAX_*_PER_SLICE` values specifically in the tests, hmm, not sure -- it could be used where it's an exact match but for the other values then would we do `MAX_... + 1` or `MAX_... / 25` to indicate intentions of values above or significantly below the maximum? As a sort of hybrid, I've added a "constants unchanged" test -- if that were to ever break then hopefully that would also trigger a revisiting of the values used in the test.

lucene/core/src/test/org/apache/lucene/index/TestSegmentToThreadMapping.java
```diff
@@ -292,8 +292,48 @@ public void testRandom() {
...
+    assertEquals(5, IndexSearcher.MAX_SEGMENTS_PER_SLICE);
+  }
+
+  public void testTwoPreciseSlices() {
+    LeafReader firstSegmentReader = dummyIndexReader(250_000);
+    LeafReader secondSegmentReader = dummyIndexReader(250_000);
+    List<LeafReaderContext> leafReaderContexts = new ArrayList<>();
+
+    leafReaderContexts.add(new LeafReaderContext(firstSegmentReader));
+    leafReaderContexts.add(new LeafReaderContext(secondSegmentReader));
```
cpoerschke, line 311: > ... (I've tried to move the test from `index` to `search` package which would allow package instead of public visibility for the constants but then access to one of the other classes used in the test was not available.) ...

The access unavailable was https://github.com/apache/lucene/blob/releases/lucene/9.11.1/lucene/core/src/java/org/apache/lucene/index/LeafReaderContext.java#L48 i.e.

```
reason: LeafReaderContext(LeafReader) is not public in LeafReaderContext; cannot be accessed from outside package
```

PR #528, opened 2021-12-08
LUCENE-10296: Stop minimizing regepx

rmuir:
In current trunk, we let caller (e.g. `RegExpQuery`) try to "reduce" the expression. The parser nor the low-level executors don't implicitly call exponential-time algorithms anymore.

But now that we have cleaned this up, we can see, what is happening is even worse than just calling `determinize()`. We still call `minimize()` which is much crazier and much more.

We stopped doing this for all other `AutomatonQuery` subclasses a long time ago, as we determined that it didn't help performance. Additionally, minimization vs. determinization is even less important than early days where we found trouble: the representation got a lot better. Today when you `finishState()` we do a lot of practical sorting/coalescing on-the-fly. The practical parts of minimization for runtime perf. Also we added our fancy UTF32-to-UTF8 automata convertor, that makes the worst-case-space-per-state significantly lower than with UTF-16 representation? So why `minimize()` ?

Let's just replace `minimize()` calls with `determinize()` calls? I've already swapped them out for all of `src/test`, to get jenkins looking for issues ahead of time.

This change moves Hopcroft minimization (`MinimizeOperations`) to `src/test` for now. I'd like to explore nuking it from there as a next step, any tests that truly need minimization should be fine with brzozowski's algorithm: that's a 2-liner.

PR #513, opened 2021-12-04
LUCENE-10010: don't determinize/minimize in RegExp

rmuir:
Today, RegExp calls minimize() at every parsing step. There is little point to making an NFA execution when it is doing this.

Moreover, some minimize() calls are missing, and in fact in rare cases it can already return an NFA today (for certain syntax)

Instead, RegExp parsing should do none of this, instead it may return a DFA or NFA. NOTE: many simple regexps happen to be still returned as DFA, just because of the algorithms in use.

Callers can decide whether to determinize or minimize. RegExp parsing should not run in exponential time.

All src/java callsites were modified to call minimize(), to prevent any performance problems. minimize() seems unnecessary, but let's approach removing minimization as a separate PR. src/test was fixed to just use determinize() in preparation for this.

This PR is very similar to #485 but just for the regexp parsing. 

One annoying change is we can't support true complement operator (~), which is an obscure non-default operator (must be enabled with flag). negated character classes (e.g. `[^a-z]`) still work as they are contained. But complementing arbitrary automata would require exponential time!

PR #485, opened 2021-11-29
LUCENE-10010: don't determinize in CompiledAutomaton/RunAutomaton

rmuir:
Instead, require that incoming automata is determinized by the caller, throwing an exception if it isn't.

This paves the way for NFA execution in the future: if you pass an NFA to AutomatonQuery, we should use the NFA algorithm on it. No need for lots of booleans or enums.

The idea is that we clean this one up and fold this into the main LUCENE-10010 PR, to keep the APIs simple. But we could also merge it independently first.

PR #13511, opened 2024-06-20
Reduce memory use of MinimizationOperations#minimize

original-brownbear:
It is relatively easy to consume a massive amount of memory for the minimize operation, with its lists of boxed Integer (even though these are mostly cached, it's still more than 4b per instance to store them instead of plain storage) and neverending duplicate+empty StateList instances.
The boxed integer situation we can fix and probably speedup by using the hppc primitive collections. To fix the duplicate/empty StateList instances, we can use a constant. This requires some hacky forking on the write path but that's about it.
This is partly motivated by ES users at times creating broken, very long prefix queries that can then eat up GBs of heap. With this change, the examples I've been looking at become about 6x cheaper heap wise, making it less likely that kind of mistakes impacts stability.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/util/automaton/MinimizationOperations.java
+++ b/lucene/core/src/test/org/apache/lucene/util/automaton/MinimizationOperations.java
@@ -29,10 +29,11 @@
 
 package org.apache.lucene.util.automaton;
 
-import java.util.ArrayList;
 import java.util.BitSet;
-import java.util.HashSet;
 import java.util.LinkedList;
+import org.apache.lucene.internal.hppc.IntArrayList;
+import org.apache.lucene.internal.hppc.IntCursor;
+import org.apache.lucene.internal.hppc.IntHashSet;
 
 /**
  * Operations for minimizing automata.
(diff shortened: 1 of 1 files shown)


rmuir:
This method is only used by tests (hence changes here only to src/test), maybe it should just be removed and replaced by AutomatonTestUtil.minimizeSimple()?

rmuir:
btw this is just referring to `main` branch. only concern would be some tests that might go very slow or have bad complexity with the Brzozowski algorithm, which is why we kept the hopcroft one here.

for 9.x this change could be useful, as queries etc are still minimizing in 9.x code. Another alternative for 9.x would be to backport changes such as #528, #11813, #513, #485

these changes were never backported to 9.x due to "playing it safe" but a lot of them have been baking for years...

rmuir (approved):
changes look correct to me. the tests for this thing should be pretty good, so I  think it'd be safe for 9.x and probably give some memory/speed relief.

but see comment above, alternatively we could probably just stop minimizing at all, since we haven't been doing it in 10.x for years... pick your poison :)

jpountz:
No longer minimizing feels like a big enough change in runtime behavior that it's a better fit for a major version. If there are no concerns, I plan on merging+backporting this change and keep the idea of no longer minimizing automata for 10.0?

rmuir:
+1, sounds good to me

PR #13512, opened 2024-06-21
fixing a wrong check of RollingCharBuffer

tohidemyname:
### Description

The method code is as follows:

'  public int get(int pos) throws IOException {
...
      final int readCount = reader.read(buffer, nextWrite, toRead);
      if (readCount == -1) {
        end = true;
        return -1;
      }
'
When readCount fails, it can return 0. This value should aslo be checked.


benwtrent:
Do you have a test that replicates the bug to verify its fixed?

PR #13524, opened 2024-06-26
Reduce overhead for FSTs in FieldReader

original-brownbear:
We don't need to clone the index input we hold on to in OffHeapFSTStore since we only use it for slicing from known coordinates anyway. -> remove the cloning and add the infrastructure to initialize OffHeapFSTStore without seeking the input to the starting offset.

For Elasticsearch this can save hundreds of MB in `Indexinput` instances.

Essentially the same optimization we have elsewhere already, e.g. #13464


jpountz (commented):
This per-field per-segment clone is useless indeed.

I'm not completely happy with the `OffHeapFSTStore` API, which has now two ways of being initialized (constructor vs. init() method). Could we have only one? FWIW it's an internal class so it's ok to break the API.

original-brownbear:
Sure thing @jpountz how about https://github.com/apache/lucene/pull/13524/commits/75b5693ed0b8742dd7d64d6b22f02940e4975da1 ... just do away with the `FSTStore` API then? If either is ok, the constructor and immutable objects look much nicer to me and actually save a little heap for the on heap version. WDYT?

jpountz (approved):
Thanks, it looks much cleaner this way.

For a follow-up, I wonder if we could add a general test for this. Something like creating an index with 100 fields, and making sure that opening a segment creates less than 100 clones and/or slices?

original-brownbear:
>  Something like creating an index with 100 fields, and making sure that opening a segment creates less than 100 clones and/or slices?

Happy to :) I was meaning to look into testing this stuff and was way overthinking it :D just counting clones+slices seems like a really cool idea here :) I'll see what I can do!

jpountz:
A failure when backporting suggests that the logic in the lucene40 block tree is not good, as we now load bytes starting at `indexStartFP` and no longer add bytes for the FST metadata for version < 8.6. I'm looking into it, it's not good that the main branch didn't catch this issue.

jpountz:
I opened #13527.

jpountz:
> Happy to :) I was meaning to look into testing this stuff and was way overthinking it :D just counting clones+slices seems like a really cool idea here :) I'll see what I can do!

@original-brownbear I'm curious if you had any success with this approach?

Issue #12313, opened 2023-05-19
Multi-value Support for KnnVectorField


alessandrobenedetti:
### Description

It would be nice to support multiple values in a Knn vector field.
This must be compatible with both the Exact and Approximate Nearest Neighbor search.

There are two sides to the coin:

1) Index time support - allowing to add in the indexing data structures multiple vectors for the same field and docID
2) Query time support - how to retrieve a topK list of documents, where each document my have multiple neighbors to the query

The problem is more complicated than it seems.

An initial tentative design and draft implementation is attached

msokolov:
I see a lot of good work on the implementation in the attached PR, great! What I'm lacking though is any understanding of what the use cases for this might be. Do we have some? I think it's important to have at least some envisioned so we can know how to go with the implementation since there will undoubtedly be tradeoffs.

benwtrent:
There are also late-interaction-models that do embeddings per token. While the current HNSW codec wouldn't be best for that, it is another use case for multiple embeddings per document.

uschindler:
I have a customer using Solr to do kNN for trademark images. Each trademark has several images and they want to find te trademark with closest imae match (cosine distance). They currently use some outdated homemade plugin using BinaryDocvalues and  scan through them, but to speed up search kNN looks like right Choice. But to do facetting they want hits counted on the trademarks and not on the images.

benwtrent:
So, I have been thinking of the current implementation and was wondering if we could instead move towards using the `join` functionality?

Just to make sure I am not absolutely crazy.

 - `join` already requires children and parent documents to be indexed next to each other (parent docs as the last doc in the child&parent block).
 - When searching the graph, a separate kind of `NeighborQueue` that requires topK parent documents (ParentNeighboQueue?). This queue would require a `BitSet` of the parent doc ids. Then when a child doc is visited via the graph, we can check its score and determine the parent block via `BitSet#nextSetBit`. Tracking how many parents we have visited and their scores wouldn't be too bad from that avenue.
 - Top scoring documents would be the scores according to the parent docs. I guess this COULD be flexible allowing `mean`, `min`, `max` discovered child during the graph explore.

uschindler:
I would still prefer to have multiple values per document. From the point of view of implementation this does not look crazy to me, but using blockjoins adds too many limitations and often people don't want to use it for other reasons

[...]

PR #13525, opened 2024-06-26
[WIP] Multi-Vector support for HNSW search

vigyasharma:
Adds support for multi-valued vectors to Lucene. 

In addition to max-similarity aggregations like parent-block joins, this change supports ColBERT style distance functions that compute interaction across all query and document vector values. Documents can have a variable number of vector values, but to support distance function computations, we require all values to have the same dimension.

> This is a big change and I still need to work on tests (existing and new), backward compatibility, benchmarks and some code refactor/cleanup. Raising this early version to get feedback on the overall approach. I marked the PR with `no commit` tags.

Addresses #12313 .

.

### Approach
We define a new "Tensor" field that comprises multiple vector values, and a new `TensorSimilarityFunction` to compute distance across multiple vectors (_uses SumMax() currently_). Node ordinal is assigned to the tensor value, giving us one ordinal per document. All vector values of a tensor field are processed together during writing, reading and scoring. They are passed around as a packed `float[]` or `byte[]` array with all vector values concatenated. Consumers (like the `TensorSimilarityFunction`) slice this array by dimension to get individual vector values.

Tensors are stored using a new _FlatVectorStorage_ that supports writing/reading _variable length_ values per field (allowing us to have a different number of vectors per tensor). We reuse the existing HNSW readers and writers. Each graph node is a tensor and maps to a single document.  I also added a new codec tensor format, to allow both tensors and vectors to coexist. I'm not yet sure how to integrate with the quantization changes (separate later date change) and didn't want to force everything into a single format. Tensors continue to work with `KnnVectorWriter/Reader` and extend the `FlatVectorWriter/Reader` classes.

Finally, I named the field and format _"Tensors"_ though technically these are only rank-2 tensors. The thought was that we might extend this field and format if we ever went for higher rank tensors support. I'm open to renaming based on community feedback.

.

### Major Changes
The PR has a lot of files which is not practical to review. Here are the files with key changes. If we align on the approach, I'm happy to reraise separate split PRs with different changes.

1. **New fields and similarity function for tensors.**
	1. lucene/core/src/java/org/apache/lucene/document/FieldType.java
	2. lucene/core/src/java/org/apache/lucene/document/KnnByteTensorField.java
	3. lucene/core/src/java/org/apache/lucene/util/ByteTensorValue.java
	4. lucene/core/src/java/org/apache/lucene/document/KnnFloatTensorField.java
	5. lucene/core/src/java/org/apache/lucene/util/FloatTensorValue.java
	6. lucene/core/src/java/org/apache/lucene/index/TensorSimilarityFunction.java
	7. lucene/core/src/java/org/apache/lucene/index/FieldInfo.java
	8. lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
2. **Indexing chain changes**
	1. lucene/core/src/java/org/apache/lucene/index/IndexingChain.java
	2. lucene/core/src/java/org/apache/lucene/index/VectorValuesConsumer.java
3. **Reader side changes to return a tensor reader for tensor fields**
	1. lucene/core/src/java/org/apache/lucene/index/SegmentCoreReaders.java
	2. lucene/core/src/java/org/apache/lucene/index/SegmentReader.java
	3. 
4. **A new tensor format in the codec**
	1. lucene/core/src/java/org/apache/lucene/codecs/KnnTensorsFormat.java
	2. lucene/core/src/java/org/apache/lucene/index/CodecReader.java
5. **A new tensor scorer to work with multiple vector values**
	1. lucene/core/src/java/org/apache/lucene/codecs/hnsw/FlatTensorsScorer.java
	2. lucene/core/src/java/org/apache/lucene/codecs/hnsw/DefaultFlatTensorScorer.java
6. **A `Lucene99FlatTensorsWriter` for writing in the new flat tensor format** - lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatTensorsWriter.java
7. **A `Lucene99FlatTensorsReader` for reading the flat tensor format** - lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatTensorsReader.java
8. **An `HnswTensorFormat` that uses FlatTensorFormat to initialize the flat storage readers/writers underlying HNSW reader/writer.**
	1. lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswTensorsFormat.java
9. **Hnsw reader and writer changes to support tensor fields and similarity function**
	1. lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java
	2. lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
10. **Off Heap Byte and FloatTensorValues for use by scorers**
	1. lucene/core/src/java/org/apache/lucene/codecs/lucene99/OffHeapByteTensorValues.java
	2. lucene/core/src/java/org/apache/lucene/codecs/lucene99/OffHeapFloatTensorValues.java
11. **Setup to read and write tensor data value offsets to support variable vector count per tensor.** This uses a DirectMonotonicReader/Writer.
	1. lucene/core/src/java/org/apache/lucene/codecs/lucene99/TensorDataOffsetsReaderConfiguration.java
12. **Syntax sugar for tensor queries**
	1. lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
	2. lucene/core/src/java/org/apache/lucene/search/KnnByteTensorQuery.java
	3. lucene/core/src/java/org/apache/lucene/search/KnnFloatTensorQuery.java

.

### Open Questions
1. I like the type safety of a separate field and similarityFn. It avoids user traps around passing a single value VectorSimilarity for tensors. But it does add a bunch of extra code, and ctor changes across a lot of files. Some options to simplify could be:
   - Reuse `vectorEncoding` and `vectorDimension` attributes in FieldInfo instead of a separate tensor encoding and dimension
   - Also reuse VectorSimilarityFunction, but create a separate "tensor aggregator" that corresponds to SumMax.


benwtrent:
Thank you @vigyasharma ! Before digging into API design, etc. (I definitely have thoughts there, but will shelve them for now).

I am wondering about the performance, and if recall significantly suffers in the "passage vector" use case.

Cohere's wikipedia embeddings all indicate their parent page. So, I wonder how this would work on finding the nearest page given the `maxsim(passage)` vs. using the Lucene join logic and putting the passages directly in the graph. 

If we can recover recall by simply gathering more docs or by increasing graph connections, it makes this idea very attractive.

vigyasharma:
Thanks for looking into this @benwtrent! I did some cleanup on FieldType and similarity function interfaces, hope it doesn't cause too much churn to what you've gone through so far. 

We now use existing vector encoding, dimension and similarity; and add only the new info needed to FieldType - the aggregation fn, and a flag to differentiate vector/tensor - I like this abstraction better.

Benchmarking and writing tests are next up on my list. The Cohere "passage vector" use case seems like a valuable exploration, thanks for pointing that out. The baseline comparison with Lucene joins for maxSim is a good first test, although, it's also worth noting that this change offers additional functionality over joins. It enables ColBERT-style late interactions, where both queries and documents can have multiple vector values.

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene87/Lucene87Codec.java
```diff
@@ -27,19 +27,7 @@
...
-import org.apache.lucene.codecs.FilterCodec;
-import org.apache.lucene.codecs.KnnVectorsFormat;
-import org.apache.lucene.codecs.LiveDocsFormat;
-import org.apache.lucene.codecs.NormsFormat;
-import org.apache.lucene.codecs.PointsFormat;
-import org.apache.lucene.codecs.PostingsFormat;
-import org.apache.lucene.codecs.SegmentInfoFormat;
-import org.apache.lucene.codecs.StoredFieldsFormat;
-import org.apache.lucene.codecs.TermVectorsFormat;
+import org.apache.lucene.codecs.*;
```
benwtrent, line 30: it will help with churn, etc. and we prefer not having `*` imports.

benwtrent:
> The baseline comparison with Lucene joins for maxSim is a good first test, although, it's also worth noting that this change offers additional functionality over joins. It enables ColBERT-style late interactions, where both queries and documents can have multiple vector values.

I recognize that. I am just thinking of the flexibility Lucene provides. 

As an aside, at first blush, this seems way too complicated. I honestly would have expected a 

 - some format changes to the EXISTING to allow for more than one vector & an option on how to handle more than one
 - a change to [Float|Byte]VectorValues to allow iterating individual vectors (maybe a "multi" vector values interface)

I am against adding a entirely new format type. It seems like we should extend the existing one.

vigyasharma:
> at first blush, this seems way too complicated.

Happy to simplify it, that's indeed the idea behind this early draft. For my understanding, what are your reservations with a new format type? I have a sense but would like to know your perspective..

By extending the existing format, I assume you mean not having a separate `KnnTensorsFormat` in the codec. It gave me the convenience of initializing FlatVectorsFormat with a TensorScorer. To extend `KnnVectorsFormat`, I could probably modify the FlatVectorScorer to include tensor fns (they need the aggregation fn). On the storage side, tensors have variable length vector arrays and need to read/write their start offsets. I could update the existing FlatVectorsFormat and write these data offsets only for when the field is a tensor. 

How does that sound? I am thinking of keeping it similar to the ScalarQuantization change with separate `HnswReader|Writer` and `FlatVectorReader|Writer` classes that modify the existing format but are not yet the codec defaults. (As an aside, I haven't gone through how Scalar Quantization works yet, so I'll make the change without it, and we can subsequently iterate to include it).

While we're at it, I feel that the name "tensor" is less fitting without the separate format. I'm considering renaming to "multi-vectors", thoughts?

vigyasharma:
> a change to [Float|Byte]VectorValues to allow iterating individual vectors (maybe a "multi" vector values interface)

The work to handle multiple vector values gets taken care of in the similarity function. So we don't really have to change the [Float|Byte]VectorValues, and RandomAccessVectorValues interfaces. We still pass around a T[] vector value, just that it can be of variable length (depending on no. of vectors). The similarity fn slices it by dimension before comparing. 

I feel this fits conveniently with a lot of our existing interfaces. Do you see a specific need for [Float|Byte]VectorValues to iterate on individual vector values instead?

benwtrent:
>  I could update the existing FlatVectorsFormat and write these data offsets only for when the field is a tensor.

I was thinking something like this. We should dynamically handle if more than one vector is provided. Having to configure up front "Hey, more than one vector is incoming" is weird. Why would anybody ever configure the "single vector case" as the multi-vector case would also just handle the single vector one. Seems like we don't need specialized formats but should instead update the current flat vector formats. 



> I feel this fits conveniently with a lot of our existing interfaces. Do you see a specific need for [Float|Byte]VectorValues to iterate on individual vector values instead?

Yes, we need to be able to iterate vectors via doc Ids and gather each individual vector for a given document. 

Three concerns immediately come to mind:

 - rescoring docs that are gathered via some quantized methodology
 - Determining the true nearest vector for a given document.
 - Ability to iterate vectors when quantizing them. We need randomly sample across all vectors. We don't want to sample via docs ids, this will likely add bias and hurt the quantization quality.


As for adding new information to the FieldInfo, another valid option is making it configurable directly on the format and not update fieldinfo. I am not sure its valuable to have it in fieldinfo. I wouldn't expect the useages for how to resolve the multi-vector scoring to be as broad as our similarity functions or vector dimensions.

lucene/core/src/java/org/apache/lucene/codecs/hnsw/FlatVectorsReader.java
```diff
@@ -43,9 +43,18 @@ public abstract class FlatVectorsReader extends KnnVectorsReader implements Acco
   /** Scorer for flat vectors */
   protected final FlatVectorsScorer vectorScorer;
 
+  /** Scorer for flat tensors */
+  protected final FlatTensorsScorer tensorScorer;
+
   /** Sole constructor */
```
cpoerschke, line 49:
```suggestion
```

lucene/core/src/java/org/apache/lucene/document/FieldType.java
```diff
@@ -400,6 +405,42 @@ public VectorSimilarityFunction vectorSimilarityFunction() {
...
+      boolean isTensor,
+      int dimension,
+      VectorEncoding encoding,
+      VectorSimilarityFunction similarityFunction,
+      Aggregation aggregation) {
+    checkIfFrozen();
+    this.isTensor = isTensor;
+    this.vectorDimension = dimension;
+    this.vectorEncoding = Objects.requireNonNull(encoding);
+    this.vectorSimilarityFunction = Objects.requireNonNull(similarityFunction);
```
cpoerschke, line 430: Wondering if we might call `setVectorAttributes` here:
```suggestion
    setVectorAttributes(dimension, encoding, similarityFunction);
```

cpoerschke:
Merged in latest `origin/main` to resolve the merge conflict that arose from #13529 - hope that's okay.

vigyasharma:
> we need to be able to iterate vectors via doc Ids and gather each individual vector for a given document.

Could the consumers of `T[] vectorValue()` unpack the multiple vector values for a doc when they need to? We could add an iterator method for this in `Float|ByteVectorValues` for convenience.. do you have an interface in mind? 

I imagine the DISI would still advance to the next document in `nextDoc()`, and this would be a separate method, like `nextValue()` which iterates on the packed vector value, or optionally calls `nextDoc()` when needed?

> As for adding new information to the FieldInfo, another valid option is making it configurable directly on the format and not update fieldinfo.

I like the flexibility of switching from single to multi-values. Was thinking of adding it in a subsequent iteration. Re: not keeping state in `FieldInfo`, my worry is with the aggregation function. Will we be able to ensure users don't change the aggregation fn midway through indexing their corpus, if this state is only within the format and FieldReader/Writers... I need to think about it in more detail.

vigyasharma:
> Merged in latest `origin/main` to resolve the merge conflict that arose from #13529 - hope that's okay.

That's perfect @cpoerschke , and thank you for looking at this change!

benwtrent:
> I imagine the DISI would still advance to the next document in nextDoc(), and this would be a separate method, like nextValue() which iterates on the packed vector value, or optionally calls nextDoc() when needed?

Yes, something like that exactly.

>  Will we be able to ensure users don't change the aggregation fn midway through indexing their corpus, if this state is only within the format and FieldReader/Writers... I need to think about it in more detail.

I am not sure of a fool-proof limitation other than ensuring everything is the same on merge. But even then, I am not 100% sure that is even necessary. If the score aggregation changes on merge, we just change it. I am not sure its an issue.

lucene/core/src/java/org/apache/lucene/codecs/KnnTensorsFormat.java
```diff
@@ -0,0 +1,145 @@
...
+
+  @Override
+  public String getName() {
+    return name;
+  }
+
+  /** looks up a format by name */
+  public static KnnTensorsFormat forName(String name) {
+    return Holder.getLoader().lookup(name);
+  }
```
cpoerschke, line 77: minor: noting that https://github.com/apache/lucene/blob/releases/lucene/9.11.1/lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsFormat.java#L72-L93 also has `reloadKnnVectorsFormat` and `availableKnnVectorsFormats` but perhaps they aren't applicable here as yet?

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java
```diff
@@ -143,14 +143,22 @@ public int nextDoc() throws IOException {
   public static final class MergedVectorValues {
     private MergedVectorValues() {}
 
+    private static void validateFieldEncoding(FieldInfo fieldInfo, VectorEncoding expected) {
+      assert fieldInfo != null && (fieldInfo.hasVectorValues() || fieldInfo.hasTensorValues());
+      if (fieldInfo.hasTensorValues() && fieldInfo.hasVectorValues()) {
```
cpoerschke, line 148:
```suggestion
      if (fieldInfo.hasVectorValues() && fieldInfo.hasTensorValues()) {
```

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java
```diff
@@ -209,17 +210,17 @@ public static VectorEncoding readVectorEncoding(DataInput input) throws IOExcept
...
           "Inconsistent vector similarity function for field=\""
               + info.name
               + "\"; "
-              + similarityFunction
+              + vectorSimilarityFunction
               + " != "
               + info.getVectorSimilarityFunction());
     }
-    return FieldEntry.create(input, vectorEncoding, info.getVectorSimilarityFunction());
+    return FieldEntry.create(info, input, vectorEncoding, info.getVectorSimilarityFunction());
```
cpoerschke, line 223: Having both `info` and `info.getVectorSimilarityFunction()` jumps out here.
```suggestion
    return FieldEntry.create(info, input, vectorEncoding);
```

or

```suggestion
    return FieldEntry.create(info, input, vectorEncoding, vectorSimilarityFunction);
```

lucene/core/src/java/org/apache/lucene/index/FieldInfo.java
```diff
@@ -92,6 +97,8 @@ public FieldInfo(
       int vectorDimension,
       VectorEncoding vectorEncoding,
       VectorSimilarityFunction vectorSimilarityFunction,
+      boolean isTensor,
+      TensorSimilarityFunction.Aggregation tensorAggregate,
```
cpoerschke, line 101: naive question: is `isTensor` possible/meaningful without `tensorAggregate` and vice versa? and/or could a special `TensorSimilarityFunction.Aggregation.NONE` mean no-tensors?
```suggestion
      TensorSimilarityFunction.Aggregation tensorAggregate,
```

cpoerschke (commented):
51 / 75 files viewed

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java
```diff
@@ -143,14 +143,22 @@ public int nextDoc() throws IOException {
   public static final class MergedVectorValues {
     private MergedVectorValues() {}
 
+    private static void validateFieldEncoding(FieldInfo fieldInfo, VectorEncoding expected) {
```
cpoerschke, line 146: Seeing this method being factored out here inspired me to try and see if more duplication could be removed: pr-link-to-follow

vigyasharma:
Did some refactor to reuse the same format. Also renamed "tensor" to "multi-vector"..

vigyasharma:
Diff is down to 45 files now (from 75, yikes!), and most touch points are still due to FieldInfo changes. These are the files with key changes:
```ruby
lucene/core/src/java/org/apache/lucene/codecs/hnsw/DefaultFlatMultiVectorScorer.java
lucene/core/src/java/org/apache/lucene/codecs/hnsw/FlatVectorsScorer.java

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatMultiVectorsFormat.java
lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatMultiVectorsReader.java
lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatMultiVectorsWriter.java
lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswMultiVectorsFormat.java
lucene/core/src/java/org/apache/lucene/codecs/lucene95/OrdToDocDISIReaderConfiguration.java
lucene/core/src/java/org/apache/lucene/codecs/lucene99/MultiVectorDataOffsetsReaderConfiguration.java

lucene/core/src/java/org/apache/lucene/codecs/lucene99/OffHeapByteMultiVectorValues.java
lucene/core/src/java/org/apache/lucene/codecs/lucene99/OffHeapFloatMultiVectorValues.java

lucene/core/src/java/org/apache/lucene/index/FieldInfo.java
lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
lucene/core/src/java/org/apache/lucene/index/IndexableFieldType.java
lucene/core/src/java/org/apache/lucene/document/FieldType.java
lucene/core/src/java/org/apache/lucene/util/ByteMultiVectorValue.java
lucene/core/src/java/org/apache/lucene/util/FloatMultiVectorValue.java
lucene/core/src/java/org/apache/lucene/document/KnnByteMultiVectorField.java
lucene/core/src/java/org/apache/lucene/document/KnnFloatMultiVectorField.java

lucene/core/src/java/org/apache/lucene/index/IndexingChain.java
lucene/core/src/java/org/apache/lucene/index/MultiVectorSimilarity.java
lucene/core/src/java/org/apache/lucene/index/MultiVectorSimilarityFunction.java
```

vigyasharma:
> As for adding new information to the FieldInfo, another valid option is making it configurable directly on the format and not update fieldinfo.

We create multi-vector FieldWriters differently for flat format as well as hnsw graph builds. In its current form, we'll have to change these objects midway during indexing if users switch from single to multi-vectors. Maybe we can defer multi-vector specific functionality to later stages, like right before flush or merge.. but I need to think on this more deeply before we can move away from keeping multi-vector attributes in FieldInfo.

For now, I've left it as is while I shift focus to benchmarking and testing. We can come back to this later.

lucene/core/src/java/org/apache/lucene/index/FieldInfo.java
vigyasharma: Good point, we could probably skip the isTensor flag by using `TensorSimilarityFunction.Aggregation.NONE`.

cpoerschke: added https://github.com/apache/lucene/pull/13525/commits/b2b95cdb31c7deb73c8ab5a70e0077772bef4414 to give that a go

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatMultiVectorsWriter.java
```diff
@@ -0,0 +1,824 @@
...
+      };
+    }
+
+    FieldWriter(FieldInfo fieldInfo, KnnFieldVectorsWriter<T> indexWriter) {
+      super(indexWriter);
+      this.fieldInfo = fieldInfo;
+      this.dim = fieldInfo.getVectorDimension();
+      this.docsWithField = new DocsWithFieldSet();
+      this.vectors = new ArrayList<>();
+    }
```
cpoerschke, line 734: added https://github.com/apache/lucene/pull/13525/commits/ffda63f1dbc8bf8633fa867fc1e18f8eed4c0354 to remove the `this.isMultiVector = fieldInfo.hasMultiVectorValues();` from here

lucene/core/src/java/org/apache/lucene/index/IndexingChain.java
```diff
@@ -1527,15 +1549,20 @@ void setPoints(int dimensionCount, int indexDimensionCount, int numBytes) {
     }
 
     void setVectors(
-        VectorEncoding encoding, VectorSimilarityFunction similarityFunction, int dimension) {
+        VectorEncoding encoding,
+        VectorSimilarityFunction similarityFunction,
+        int dimension,
+        MultiVectorSimilarityFunction.Aggregation multiVectorAggregate) {
```
cpoerschke, line 1555: added https://github.com/apache/lucene/pull/13525/commits/eacc63cb86f12a4087996c310a68ed3882d6f0f6 to fold `setMultiVectors` into `setVectors` here

lucene/core/src/java/org/apache/lucene/index/FieldInfos.java
```diff
@@ -452,7 +465,8 @@ synchronized int addOrGet(FieldInfo fi) {
                 new FieldVectorProperties(
                     fi.getVectorDimension(),
                     fi.getVectorEncoding(),
-                    fi.getVectorSimilarityFunction()));
+                    fi.getVectorSimilarityFunction()),
+                new FieldMultiVectorProperties(fi.getMultiVectorSimilarityFunction()));
```
cpoerschke, line 469: Wondering if a separate `FieldMultiVectorProperties` is needed or if we could tack onto the `FieldVectorProperties` that is already there?
```suggestion
                    fi.getVectorSimilarityFunction(),
                    fi.getMultiVectorSimilarityFunction()));
```

navneet1v:
@vigyasharma is there a reason to adding the multi vector field support and not use the parent child relationship of the documents to fulfill this use case?

benwtrent:
@navneet1v

The pattern doesn't work well with ColBERT esque models.

vigyasharma:
> The pattern doesn't work well with ColBERT esque models.

+1.. Good question, @navneet1v. I had the same doubts before starting this effort. There is some discussion in [12313](https://github.com/apache/lucene/issues/12313#issuecomment-2044076948). 

Basically, Multi-vectors allow for ColBERT style late interaction models, where queries and documents can both be represented by multiple vector values, and you do a similarity measure using all those values (e.g. sum(max(similarity))). They gather contextual interaction across different parts of a query and a document, without incurring the query-time computational overheads of cross-encoder models.

Nested block joins on the other hand compute distance with a single vector value and then aggregate. That doesn't work well here, esp. with both query and doc multi-vectors.

vigyasharma:
> Cohere's wikipedia embeddings all indicate their parent page. So, I wonder how this would work on finding the nearest page given the `maxsim(passage)` vs. using the Lucene join logic and putting the passages directly in the graph.

@benwtrent : do we have any existing benchmarks for ParentJoin queries in knn? 

I didn't find it in `KnnGraphTester` and the benchmarking code I've looked at so far. I can add support for ParentJoin benchmarks but I see you ran something similar [here](https://github.com/apache/lucene/pull/12434#issuecomment-1750978708), so thought I'd check if we already have this available.

benwtrent:
@vigyasharma 

> do we have any existing benchmarks for ParentJoin queries in knn?

No, we do not. I ended up writing a bunch of throw away code to benchmark latency and recall :(.

Those particular benchmarks I did with elastic/rally

vigyasharma:
I started adding support for ParentJoin benchmarks ([issue](https://github.com/mikemccand/luceneutil/issues/284)). Will raise it in multiple small PRs, here's the [first one](https://github.com/mikemccand/luceneutil/pull/283).

mikemccand:
> I started adding support for ParentJoin benchmarks ([issue](https://github.com/mikemccand/luceneutil/issues/284)). Will raise it in multiple small PRs, here's the [first one](https://github.com/mikemccand/luceneutil/pull/283).

Thank you for improving our benchy tooling @vigyasharma!

mikemccand:
I think it's awesome to invest in our benchmarking tooling to be able to test different approaches for multi-valued vectors, but, I don't think that should be a blocker to merging this?  It adds functionality over the block-join approach, and it's nice to have options for how to model multi-valued things.

Block joins are powerful if you also have other attributes you'd like to query on at the child level, e.g. "KNN(query-vector) AND price < $50" joined up to parent documents ... but are maybe overkill if you really just want to have multiple vectors without such added per-vector constraints.

benwtrent:
> I think it's awesome to invest in our benchmarking tooling to be able to test different approaches for multi-valued vectors, but, I don't think that should be a blocker to merging this?

I don't think its a blocker for code to exist in Lucene Util. But we should be aware on how this actually behaves from a recall stand point. Right now, we don't have a clue.

At a minimum, if the expressed requirement for this is to add ColBERT style things, we need to know how this behaves with that. 

But, it would also be good to see how this would work with passage search and how recall behaves.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatMultiVectorsFormat.java
```diff
@@ -0,0 +1,109 @@
...
+ *   <li><b>[byte]</b> set to 1 if field has multi-vector values, 0 otherwise
+ *   <li><b>[int]</b> multi-vector aggregation function ordinal, present only for multi-vectors
+ *   <li>DataOffsets for variable length multi-vector values. Encoded by {@link
+ *       org.apache.lucene.util.packed.DirectMonotonicWriter}, present only for multi-vectors
+ * </ul>
+ *
+ * @lucene.experimental
+ */
+// noCommit - pending tests
+public final class Lucene99FlatMultiVectorsFormat extends FlatVectorsFormat {
```
benwtrent, line 75: Why would I ever chose the single vector format? 

It seems like the multi-vector format would be optimized when there is only one vector per doc. Since we will support a dynamic number of vectors (of the same dimension) per doc, its conceivable that a segment, or an entire field, will just have one vector per doc. 

Why aren't we updating our existing formats to add support for multi-value vectors?

vigyasharma:
> This PR has not had activity in the past 2 weeks, labeling it as stale...

Just to update on some activity here: 
I'm working on parent block join benchmarks in `luceneutil`, which I'll to use to compare against the passage search use-case for multi-vectors. Had to put it on hold due to other priorities, but I should be able to get back to it next week.

vigyasharma:
Changes to support parentJoin benchmarks in luceneutil -- https://github.com/mikemccand/luceneutil/pull/296

Benchmarking results for the passage search use-case using Cohere embeddings for wikipedia (this is for the default parentJoin implementation, without using multi-vectors):

```ruby
# parent join with quantization
recall  latency (ms)    nDoc    fanout  maxConn beamWidth       quantized       visited index ms        selectivity     filterType
0.158    3.96   253804  20      32      50      4 bits  100     25213   1.00    post-filter
0.162    4.00   253804  20      32      50      7 bits  100     24277   1.00    post-filter
0.015   13.42   253804  20      32      50      8 bits  100     167694  1.00    post-filter

# parentJoin without quantization
recall  latency (ms)    nDoc    fanout  maxConn beamWidth       quantized       visited index ms        selectivity     filterType
0.167    6.22   253804  20      32      50      no      100     27412   1.00    post-filter

# default run (no parentJoin)
recall  latency (ms)    nDoc    fanout  maxConn beamWidth       quantized       visited index ms        selectivity     filterType
0.565    1.89   250000  20      32      50      4 bits  4610    43426   1.00    post-filter
0.820    1.62   250000  20      32      50      7 bits  4198    43593   1.00    post-filter
0.004    3.49   250000  20      32      50      8 bits  9531    86058   1.00    post-filter
```

benwtrent:
Is "default run" from this PR?

lucene/core/src/java/org/apache/lucene/index/MultiVectorSimilarityFunction.java
```diff
@@ -0,0 +1,203 @@
...
+        }
+        // TODO: can we avoid making vector copies?
+        List<float[]> outerList = new ArrayList<>();
+        List<float[]> innerList = new ArrayList<>();
+        for (int i = 0; i <= outer.length; i += dimension) {
+          outerList.add(ArrayUtil.copyOfSubArray(outer, i, dimension));
+        }
+        for (int i = 0; i <= inner.length; i += dimension) {
+          innerList.add(ArrayUtil.copyOfSubArray(inner, i, dimension));
+        }
```
benwtrent, line 77: Agreed, to do this, we will need to update our similarity functions to allow offsets & lengths. But it should be doable. Then, performance wise and logic wise `SUM_MAX` over two single vectors would be the same as going directly to the similarity function.

benwtrent: ^ this optimization can happen after this PR.

lucene/core/src/java/org/apache/lucene/search/KnnFloatVectorQuery.java
```diff
@@ -77,6 +79,35 @@ public KnnFloatVectorQuery(String field, float[] target, int k, Query filter) {
...
+   * @throws IllegalArgumentException if <code>k</code> is less than 1
+   * @return {@link KnnFloatVectorQuery}
+   */
+  public static KnnFloatVectorQuery create(
+      String field, List<float[]> target, int k, Query filter) {
+    if (target == null || target.isEmpty()) {
+      throw new IllegalArgumentException("empty target");
+    }
+    int targetDim = target.get(0).length;
+    float[] packedTarget = new float[targetDim * target.size()];
```
benwtrent, line 98: since we do this huge allocation, we should have some sane upper limit on the number of vectors allowed (maybe 10k? or 100k?).

benwtrent (commented):
I like where this PR has ended up. Sorry I haven't really been able to look deeply at it.

I like the flattened API for vector values to keep our interfaces clean. 

One thing that strikes me is that as I read through this code, it seems like `single vector` is a special form of `multi-vector`. For example, `SUM_MAX` & just vector similarity comparisons would be mathematically (and once we can fix the APIs performance wise) equivalent. 

With this in mind, it seems like this should just be "the way" the new formats are and single vector storage/scoring could be considered optimization (e.g. we don't need to worry about offsets as the number of vectors per doc is the same, its just 1).


Additionally, I remember mentioning "Hey we should be able to iterate the individual vectors, etc." in our FloatVectorValue things...I am not sure about that any more. If the user knew per doc how many vectors where their, we can simply say "Hey, this vector value is a flatten representation", and we can make our other APIs (vector similarity, etc.) take offsets & lengths.

benwtrent:
@msokolov I saw recently you were working on a major refactor where we just make every vector access random access. I think this might make the changes in this PR simpler as we won't have two access methodologies but just a single one. 

Though I read that this might not be done until sometime after the Lucene 10 release.

vigyasharma:
> Is "default run" from this PR?

No. "default run" is knn search where each embedding is a separate document with no relationship between them. I'm still wiring things up to see benchmark results for this PR.

lucene/core/src/java/org/apache/lucene/index/MultiVectorSimilarityFunction.java
vigyasharma: yes, that's what I had in mind. It'll be slower right now with all the extra allocation but we can change it later by making our sim. fn impl (probably in the vectorization provider?) work with offsets and lengths.

benwtrent:
Hey @vigyasharma there is a lot of good work here. 

I am going to shift my focus and see about how I can help here more fully. What are the next steps?

I am guessing handling all the merging from main, I can take care of that sometime next week. 

Just wondering where I can help.

vigyasharma:
Thanks @benwtrent. I've been working on getting a multi-vector benchmark running to wire this end to end. Found some pesky bugs and oversights. I'm planning to split this feature into multiple smaller PRs. This PR was mainly to get inputs on the approach. It's too big to test and review. I'll share a plan of the split PRs soon.

re: the multi-vector benchmark for passage search use-case, I've been stuck on a bug where after I run into an `EOFException` on reading the last multi-vector document through `DenseOffHeapMultiVectorValues`. I could definitely use some help here. If you plan to take a look, you can use the code in this PR (i'll push my fixes) and multi-vector benchmark code from [here](https://github.com/vigyasharma/luceneutil/tree/multivec).

```java
Exception in thread "main" java.lang.RuntimeException: java.io.EOFException: read past EOF: MemorySegmentIndexInput(path="/Users/vigyas/forks/bench/util/knnIndices/cohere-wikipedia-docs-768d.vec-32-50-multiVector.index/_0_Lucene99HnswMultiVectorsFormat_0.vecmv") [slice=multi-vector-data]
        at knn.KnnGraphTester$ComputeBaselineNNFloatTask.call(KnnGraphTester.java:1115)
        at knn.KnnGraphTester.computeNN(KnnGraphTester.java:967)
        at knn.KnnGraphTester.getNN(KnnGraphTester.java:812)
        at knn.KnnGraphTester.run(KnnGraphTester.java:438)
        at knn.KnnGraphTester.runWithCleanUp(KnnGraphTester.java:177)
        at knn.KnnGraphTester.main(KnnGraphTester.java:172)
Caused by: java.io.EOFException: read past EOF: MemorySegmentIndexInput(path="/Users/vigyas/forks/bench/util/knnIndices/cohere-wikipedia-docs-768d.vec-32-50-multiVector.index/_0_Lucene99HnswMultiVectorsFormat_0.vecmv") [slice=multi-vector-data]
        at org.apache.lucene.store.MemorySegmentIndexInput.readByte(MemorySegmentIndexInput.java:146)
        at org.apache.lucene.store.DataInput.readInt(DataInput.java:95)
        at org.apache.lucene.store.MemorySegmentIndexInput.readInt(MemorySegmentIndexInput.java:261)
        at org.apache.lucene.store.DataInput.readFloats(DataInput.java:202)
        at org.apache.lucene.store.MemorySegmentIndexInput.readFloats(MemorySegmentIndexInput.java:231)
        at org.apache.lucene.codecs.lucene99.OffHeapFloatMultiVectorValues.vectorValue(OffHeapFloatMultiVectorValues.java:111)
        at org.apache.lucene.codecs.lucene99.OffHeapFloatMultiVectorValues.vectorValue(OffHeapFloatMultiVectorValues.java:130)
        at org.apache.lucene.codecs.hnsw.DefaultFlatMultiVectorScorer$FloatMultiVectorScorer.score(DefaultFlatMultiVectorScorer.java:185)
        at org.apache.lucene.codecs.lucene99.OffHeapFloatMultiVectorValues$DenseOffHeapMultiVectorValues$1.score(OffHeapFloatMultiVectorValues.java:248)
        at org.apache.lucene.search.AbstractKnnVectorQuery.exactSearch(AbstractKnnVectorQuery.java:220)
        at knn.KnnFloatVectorBenchmarkQuery.exactSearch(KnnFloatVectorBenchmarkQuery.java:33)
        at knn.KnnFloatVectorBenchmarkQuery.runExactSearch(KnnFloatVectorBenchmarkQuery.java:50)
        at knn.KnnGraphTester$ComputeBaselineNNFloatTask.call(KnnGraphTester.java:1111)
        ... 5 more
```

vigyasharma:
>  it seems like single vector is a special form of multi-vector

re: single v/s multi-vectors, I think it makes sense to not force users to chose multi-valued fields upfront. There's value in being able to go from single to multi-values when the need arises (and treating single-vectors like a storage optimization). 

However, I do think that we should not support changing the aggregation function once it has been set. Allowing different aggregate functions per segment will make merging and general debugging overly complicated. 

As such, I'm thinking of keeping the `Aggregation` as part of FieldInfo. There won't be a separate multi-vector field. When we create a vector without specifying the aggregation, the default gets set to `NONE` and only single values are allowed. Lucene validation will support going from `NONE` to a different value, at which point, the field is treated as multi-vector. However, changing the `Aggregation` once it is set to anything other than `NONE` will not be supported. 
How does this sound?

jimczi:
> it seems like single vector is a special form of multi-vector

The solution really depends on the semantics. In its current form, the way multi-vectors are incorporated in this PR doesn’t quite extend the single-vector case. With max similarity, we assume that each similarity score results from a full comparison, which works well when the operations are limited (such as in re-ranking scenarios). However, for ColBERT, where the average number of vectors per document is large (in the hundreds or thousands), using HNSW with max similarity layered on top may not be the optimal approach. This is likely why other vector libraries don’t expose this setup.

If our aim is to introduce max similarity in Lucene, we might need to explore a more effective strategy. Although nested vectors could be promising, they’re currently constrained by the 2B vector limit, which isn’t ideal for ColBERT, given that each input token is represented as a dense vector. The primary limitation with HNSW and the knn codec today seems to be this 2B cap on vectors.

Given these factors, we may want to reconsider HNSW for this purpose. A scalable solution would likely involve running multiple queries (one per query vector) rather than relying on an aggregation strategy. Maybe the first goal should be to incorporate max sim for re-ranking use cases first using a flat format?

vigyasharma:
> Maybe the first goal should be to incorporate max sim for re-ranking use cases first using a flat format

This could be setup using 1) a single-vector field for hnsw matching, and 2) a separate field with multi-vector values to directly access the flat format for the subset of matched readers. Basically a `RandomAccessVectorValues` on the format. I think this PR will allow us to support such a setup, may need a small change to turn off hnsw graph creation for specific fields (via `PerFieldKnnVectorsFormat`)?

vigyasharma:
As mentioned earlier, here is my rough plan for splitting this change into smaller PRs. Some of these steps could be merged if the impl. warrants it:

1. Multi-Vector similarity and aggregation classes.
2. FieldInfo changes to add a new attribute for "aggregation". This will be set to `NONE` for single-valued vectors, by default, and in formats prior to this change. 
3. Multi-vector support to flat vectors writer.
4. Random access vector values for multi-vectors.
5. Multi-vector support to flat vectors reader.
6. Hnsw writer/reader changes to work with multi-vectors if configured.
7. Support to index and query multi-vector values (may need to add this with the flat writer/reader PRs).

jimczi:
The more I think about it, the less I feel like the knn codec is the best choice for this feature (assuming that this issue is focused on late interaction models).

> It is possible that HNSW is not the ideal data structure to expose multi-vector ANN. We don't really change much in hnsw impl, except using multi-vector similarity for comparisons (graph build and search). Users can use the PerFieldKnnVectorsFormat to wire different data structures on top of the flat multi-vector format. We can also provide something off the box in a subsequent change. I think the aggregation fn. interface is also flexible for different types of similarity implementations?

Using the knn codec to handle multi-vectors seems limiting, especially since it treats multi-vectors as a single unit for scoring. This works well for late interaction models, where we’re dealing with a collection of embeddings, but it’s restrictive if we want to index each vector separately.
Using the original max similarity for HNSW is just not practical, it doesn’t scale, and I don’t think it’s something we’d actually want to support. 

It could be helpful to explore other options instead of relying on the knn codec alone. Along those lines, I created a quick draft of a [`LateInteractionField` using binary doc values](https://github.com/apache/lucene/compare/main...jimczi:lucene:late_interaction_field?expand=1), which keeps things simple and avoids major changes to the knn codec. I don’t think the flat vector format really offers any advantages over using binary doc values. In both cases, we’re able to store plain dense vectors as bytes, so there doesn’t seem to be a clear benefit to using the flat format here.

What do you think of this approach? It feels like we could skip the full knn framework if our main goal is just to score a bag of embeddings. This would keep things simpler and allow us to focus specifically on max similarity scoring without the added weight of the full knn codec.

My main worry is that adding multi-vectors to the knn codec as a late interaction model might add complexity later. It’s really two different approaches, and it seems valuable to keep the option for indexing each vector separately. We could expose this flexibility through the aggregation function, but that might complicate things across all codecs, as they’d need to handle both aggregate and independent cases efficiently.

vigyasharma:
One use-case for multi-vectors is indexing product aspects as separate embeddings for e-commerce search. At Amazon Product Search (where I work), we'd like to experiment with separate embeddings to represent product attributes, user product opinions, and product images. Such e-commerce use-cases would have a limited set of embeddings, but leverage similarity computations across all of them.

I see your point about scaling challenges with very high cardinality multi-vectors like token level ColBERT embeddings. Keeping them in a `BinaryDocValues` field is a good idea for scoring only applications. I like the `LateInteractionField` wrapper you shared, we should bring it into Lucene for such usecases.

However, I do think there is space for both solutions. It's not obvious to me how the knn codec gets polluted for future complexity. We would still support single vectors as is. My mental model is: if you want to use multi-vectors in nearest neighbor search (hnsw or newer algos later), index them in the knn field. Otherwise, index them separately as doc-values used only for re-ranking top results.

benwtrent:
> One use-case for multi-vectors is indexing product aspects as separate embeddings for e-commerce search. At Amazon Product Search (where I work), we'd like to experiment with separate embeddings to represent product attributes, user product opinions, and product images. Such e-commerce use-cases would have a limited set of embeddings, but leverage similarity computations across all of them.


This seems like just more than one knn field, or the nested field support. 


But, I understand the desire to add a multi-vector support to the flat codecs. I am honestly torn around whats the best path forward for the majority of users in Lucene.

krickert:
I would love to see a single knn field that supports multiple vectors.   Right now I feel like doing the embedded docs or a child collection to handle these use cases feel a little too overly complex when I only care about the tensors I calculated to perform a similarity search.

vigyasharma:
I tried to find some blogs and benchmarks on other library implementations. Astra Db, Vespa, faiss and nmslib, all seem to support multi-vectors in some form. 

From what I can tell, Astra Db and Vespa have ColBERT style multi-vector support in ANN [[1]](https://docs.datastax.com/en/ragstack/examples/colbert.html) [[2]](https://blog.vespa.ai/semantic-search-with-multi-vector-indexing/). Benchmarks indicate ColBERT outperforms other techniques in quality, but full ColBERT on ANN has higher latency [[3]](https://thenewstack.io/overcoming-the-limits-of-rag-with-colbert/). For large scale applications, users seem to overquery on ANN with single vector representations, and rerank them with ColBERT token vectors [[4]](https://blog.vespa.ai/announcing-long-context-colbert-in-vespa/). However, there's also ongoing work/research on reducing the no. of embeddings in ColBERT, like PLAID which replaces a bunch of vectors with their centroids [[5]](https://arxiv.org/abs/2205.09707). 

...

> I am honestly torn around whats the best path forward for the majority of users in Lucene.

I hear you! And I don't want to add complexity only because we have some body of work in this PR. Thanks for raising the concern Jim, it led me to some interesting reading.

...

My current thinking is that this is a rapidly evolving field, and it's early to lean one way or another. Adding this support unlocks experimentation. We might add different, scalable, ANN algos going forward, and our flat storage format should work with most of them. Meanwhile, there's research on different ways to run late interaction with multiple but fewer vectors. This change will help users experiment with what works at their scale, for their cost/performance/quality requirements.

I'm happy to change my perspective, and would like to hear more opinions. One reason to not add this would be if it makes the single vector setup hard to evolve. I'd like to understand if (and how) this is happening, and think on how we can address those concerns.
...

1: https://docs.datastax.com/en/ragstack/examples/colbert.html
2: https://blog.vespa.ai/semantic-search-with-multi-vector-indexing/
3: https://thenewstack.io/overcoming-the-limits-of-rag-with-colbert/
4: https://blog.vespa.ai/announcing-long-context-colbert-in-vespa/
5: PLAID - https://arxiv.org/abs/2205.09707

krickert:
> My current thinking is that this is a rapidly evolving field, and it's early to lean one way or another. Adding this support unlocks experimentation.

Amen! 

This ends up being so domain specific.  Multi-embeddings become key when you deal with domain voids in the LLMs used to create the embeddings.  That's most big corpuses.   So at least being able to experiment would get you far more feedback.

I would be ok with writing some tests if that helps.

vigyasharma:
> My concern is that this proposal doesn’t truly add support for independent multi-vectors.

That's a valid concern. I've been thinking about a more comprehensive multi-vector solution. Sharing some raw thoughts below, would love to get feedback.

We support a default aggregation value of `NONE`, which builds the graph with independent multi-vectors. Each node will be a separate vector value. As a first change, we can just support this without creating an aggregation enum. (Adding a plan for indexing this in a follow-up comment).

Once this is in place, we can add support for "dependent" multi-vector values like ColBERT. They'll take an aggregation function. Each graph node will represent all vectors for a document and use aggregated similarity (like in this PR). This will let us experiment with full ANN on ColBERT style multi-vectors.

krickert:
> And we can use getAllVectorValues() for scoring with max or avg of all vectors in the doc at query time.

Your proposal to implement `getAllVectorValues()` for scoring documents by aggregating their vectors (using methods like max or average) at query time has a lot of use cases and think it's a great idea. However, in my domain-specific data, this approach hasn't enhanced search results.  However, providing a default implementation, as you suggested, with the option for customization, could be beneficial.

(sidenote: if you are doing max/average, you can do that during index time though, right?)

I'm currently conducting A/B tests on three methods to retrieve and rank documents with multiple vectors:

1. **Aggregate Scoring:** Computing a single relevance score per document by aggregating all its vectors. Flexibility in the aggregation method would help me a lot.
2. **Chunk-Based Highlighting:** Treating each vector as a distinct document chunk to facilitate highlighting. This involves returning the top N documents so K would be more dynamic based on aggregate scores, with each document potentially containing multiple relevant sections - which would make K a bit more dynamic because we want the top N documents and K would represent the chunks that represent those documents. Implementing thresholds per-doc can help manage performance.
3. **Custom Aggregation with Embedding Tags:** Associating vectors with specific tags, such as user access levels or n-gram embeddings, to enable dynamic aggregation strategies. This allows for personalized and context-sensitive relevance scoring and would require the ability to override/customize.

The third approach is particularly promising for domain-specific applications, where standard aggregation methods may not suffice. For instance, embedding tags could be linked to user access controls, unlocking certain vectors at query time, or to specific n-grams, activating them based on query content.

Incorporating a mechanism to override the default aggregation method would facilitate experimentation with these strategies.

vigyasharma:
Thank you for sharing these use-cases @krickert !

1. **Aggregate Scoring** – I think we can do this today by joining the child doc hits with their parents and calculating score on all children in the `ToParentBlockJoinQuery`. The `getAllVectorValues()` api should make this easier by avoiding the two phase query. We can also use the aggregate query scores during approximate search graph traversal itself (use aggregate query similarity with all vector values for the doc)?

2. **Chunk-Based Highlighting** – Interesting. With `getAllVectorValues()`, we can find all vector values with similarity above a separate sim-threshold for highlights? 

3. **Custom Aggregation with Embedding Tags** – I think this one plays better with a separate child doc per vector value. We can store these tags and access related data as separate fields in child docs and filter on them during search.

Honestly, I think the existing parent-block join can achieve most use-cases for independent multi-vectors (the passage vector use case). But the approach above might make it easier to use? We also need it for dependent multi-vectors like ColBERT, though it's a separate question on whether ANN is even viable for ColBERT (v/s only for reranking).

I'd like to know what issues or limitations do people face with the existing parent-child support for multiple vector values, so we can address them here.

krickert:
> Chunk-Based Highlighting – Interesting. With getAllVectorValues(), we can find all vector values with similarity above a separate sim-threshold for highlights?

Not sure.  But it is frustrating for me: we only calculate K chunks and not N documents.  I want to return N documents all the time, and keep running K until N is reached.  Since it runs K on the chunks, I'd rather it return all thee chunks that it can until it reaches N amount of documents.  Then we can return the chunks that match which can be used by highlighting.

>  I think this one plays better with a separate child doc per vector value. We can store these tags and access related data as separate fields in child docs and filter on them during search.

Indexing the child docs requires making more docs.  We just care about the resulting embedding, so why not treat it like a tensor instead of an entire document?  It's frustrating to always make a child doc for multiple vectors when I can just do a keyword-value style instead.  Also, there's def some limitations with how you can use it with scoring and the query ends up looking like a mess.  If we can simplify the query syntax that would help a lot.

If you can get a unit test going for your PR, I'd be glad to expand on it and play with it a bit.

vigyasharma:
I pivoted to an approach that handles independent multi-vectors within flat storage, instead of requiring index time parent-block joins. Have raised a draft PR here – #14173

heemin32:
Certainly! Here's a clearer and more polished rephrasing of your message:

---

I believe this proposal to add a new field for multi-vector support is facing significant challenges primarily because we aim to support HNSW-based search on it. However, if our goal were limited to enabling MAXSIM scoring between two multi-vectors using only exact search (i.e., without KNN indexing for faster retrieval), many of these challenges could be avoided. This exact search approach would still be valuable for re-scoring documents after an initial KNN query retrieves a small candidate set.

If we want to represent multi-vectors as an array of vectors within a single document—rather than relying on a parent-child document structure—it might make sense to integrate this into the existing field type. However, such an approach should clearly demonstrate a performance advantage to justify the change.

heemin32:
I believe this proposal to add a new field for multi-vector support is facing significant challenges primarily because we aim to support HNSW-based search on it. However, if our goal were limited to enabling MAXSIM scoring between two multi-vectors using only exact search (i.e., without KNN indexing for faster retrieval), many of these challenges could be avoided. This exact search approach would still be valuable for re-scoring documents after an initial KNN query retrieves a small candidate set.

If we want to represent multi-vectors as an array of vectors within a single document—rather than relying on a parent-child document structure—it might make sense to integrate this into the existing field type. However, such an approach should clearly demonstrate a performance advantage to justify the change.

heemin32:
On a second thought, if user just diable indexing, implementing multi vector in the same field might be same as having a separate field without indexing capability.

vigyasharma:
> if our goal were limited to enabling MAXSIM scoring between two multi-vectors using only exact search (i.e., without KNN indexing for faster retrieval), many of these challenges could be avoided.

one of the challenges, and perhaps the main concern with this PR, has been practical scaling challenges of working with multi-vectors on the entire corpus. This would apply to exact search as well. A viable middle ground would be to over-collect documents via regular single-vector knn search in the first pass, and later rerank them using late-interaction multi-vectors in the second pass. This idea was shared by jimczi@ earlier in [this thread](https://github.com/apache/lucene/pull/13525#issuecomment-2445295372).

I'm working on a PR that combines his approach with a FunctionScore query for reranking. Should be able to raise it for review this week.

Issue #13519, opened 2024-06-26
Significant drop in recall for 8 bit Scalar Quantizer


naveentatikonda:
### Description

Based on some of the benchmarking tests that I ran from OpenSearch, there is a significant drop in recall ( appx. 0.03) for 8 bits irrespective of space type, confidence interval or few other parameters. For the same configuration, the recall for 7 bits is atleast greater than 0.85.

### Root Cause
As part of [quantization](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/quantization/ScalarQuantizer.java#L126-L147), after normalizing each dimension of the vector into [0 to 2^bits - 1] range, we are casting it into byte to bring it into byte range of [-128 to 127]. For 7 bits, we are normalizing each value into 0 to 127 which is already in the byte range. So, there is no rotation or shifting of data. But, for 8 bits any vector dimension which is within 128 to 255 range after normalization, the sign and magnitude changes when it is type casted into byte which leads to a non-uniform shifting or distribution of data. As per my understanding, this is the potential root cause for this huge drop in recall.
```
// pseudo code for existing implementation in Lucene SQ for 8 bits
    
    float dx = v - minQuantile; // v is float vector value in a vector at destIndex
    float dxc = Max(minQuantile, Min(maxQuantile, v)) - minQuantile;
    
    // Normalize it into 0 to 255 range for 8 bits
    // scale = 255/(maxQuantile - minQuantile)
    float dxs = scale * dxc;

    if (dest != null) {
      dest[destIndex] = (byte) Math.round(dxs); // type casting into byte after normalization
    }
```

To validate this, I have updated the quantization code and tested it against L2 space type by linearly shifting (subtracting 128) each dimension after normalizing it into 0 to 255 range such that each dimension is uniformly distributed and within the byte range(finally round it and clip it to handle edge cases) of -128 to 127. With these changes, we get a min. recall of 0.86 for the same configuration.

**Note** - The below pseudo code is not a fix and it is a different quantization technique used to validate the root cause. This works only for L2 spacetype because L2 is shift invariant but other spacetypes like cosinesimil and inner product are not shift invariant.
```
// pseudo code for custom changes in Lucene SQ for 8 bits and L2 space type
    
    float dx = v - minQuantile; // v is float vector value in a vector at destIndex
    float dxc = Max(minQuantile, Min(maxQuantile, v)) - minQuantile;
    
    // Normalize it into 0 to 255 range for 8 bits
[...]

PR #13526, opened 2024-06-27
Update quantization technique to fix drop in recall for 8 bits 

naveentatikonda:
### Description
Update quantization technique to fix drop in recall for 8 bits.

### Issues Resolved
https://github.com/apache/lucene/issues/13519

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/quantization/ScalarQuantizer.java
+++ b/lucene/core/src/java/org/apache/lucene/util/quantization/ScalarQuantizer.java
@@ -115,7 +115,7 @@ public class ScalarQuantizer {
     assert src.length == dest.length;
     float correction = 0;
     for (int i = 0; i < src.length; i++) {
-      correction += quantizeFloat(src[i], dest, i);
+      correction += quantizeFloat(src[i], dest, i, similarityFunction);
     }
     if (similarityFunction.equals(VectorSimilarityFunction.EUCLIDEAN)) {
       return 0;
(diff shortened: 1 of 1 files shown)


lucene/core/src/java/org/apache/lucene/util/quantization/ScalarQuantizer.java
```diff
@@ -136,6 +137,14 @@ private float quantizeFloat(float v, byte[] dest, int destIndex) {
     // We multiply by `alpha` here to get the quantized value back into the original range
     // to aid in calculating the corrective offset
     float dxq = Math.round(dxs) * alpha;
+
+    if (similarityFunction.equals(VectorSimilarityFunction.EUCLIDEAN) && bits == 8) {
```
mikemccand, line 141: This is probably a silly question ... why would the distance metric impact how we quantize?  Shouldn't `int8` quantization map float space down to signed byte space as best it can (minimizing the resulting quantization noise for this particular dataset (segment's vectors))?

Or, are we expecting/requiring/insisting the incoming vectors to sometimes be non-negative in all dimensions?  I thought the incoming vectors are allowed to be all over the map (positive, negative, not on the unit sphere, etc.)?

benwtrent: The similarity is required as the correction applied to Euclidean is different than dot product.

This is derived via some algebraic expansions explained in the class docs.

PR #13527, opened 2024-06-27
Add bw tests for block-tree with inlined metadata.

jpountz:
The backport of #13524 found a hole in the testing of `Lucene40BlockTreeTerms` for versions before we moved metadata to its own file. This PR adds explicit bw testing for this version. Adding the correct if/else statements made the code extremely complicated so I opted for restoring the file as it was at the time when we bumped the version.

This also fixes the bug that we introduced in #13524.


lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestAncientIndicesCompatibility.java
```diff
@@ -196,6 +196,7 @@ public void testUnsupportedOldIndexes() throws Exception {
       ByteArrayOutputStream bos = new ByteArrayOutputStream(1024);
       CheckIndex checker = new CheckIndex(dir);
       checker.setInfoStream(new PrintStream(bos, false, UTF_8));
+      checker.setLevel(CheckIndex.Level.MIN_LEVEL_FOR_INTEGRITY_CHECKS);
```
jpountz, line 199: I also added this, as it looks like we decreased coverage for ancient indices when we decreased the default level for CheckIndex. Unfortunately, this would still not have been enough to find this bug as the bw indices are too small to have meaningful data in their terms index that would have trigger the bug.

PR #13528, opened 2024-06-28
[Abstract]Knn[Byte|Float]VectorQuery tweaks: reduce duplicate method calls

cpoerschke:
From code reading only i.e. not based on profiling or benchmarking or something like that.


benwtrent (approved):

mayya-sharipova (approved):
Thanks for refactoring

PR #13529, opened 2024-06-28
Lucene99HnswVectorsReader.search float-vs-byte variants: reduce code duplication

cpoerschke:


lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java
```diff
@@ -248,45 +248,53 @@ public ByteVectorValues getByteVectorValues(String field) throws IOException {
...
+  private abstract static class RandomVectorScorerSupplier {
+    abstract RandomVectorScorer get() throws IOException;
+  }
+
+  private void search(
+      FieldEntry fieldEntry,
+      KnnCollector knnCollector,
+      Bits acceptDocs,
+      VectorEncoding vectorEncoding,
+      RandomVectorScorerSupplier scorerSupplier)
```
benwtrent, line 289: Why not use a `IOSupplier<RandomVectorScorer>` ? The added abstract class doesn't seem worth it. 

Then the calling code can be simplified to `() -> flatVectorsReader.getRandomVectorScorer(field, target)`

benwtrent (commented):
I think it could be simplified a bit more with `IOSupplier` and lambdas

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java
cpoerschke: Ah, thanks! I didn't know about `IOSupplier`'s existence: https://github.com/apache/lucene/blob/releases/lucene/9.11.1/lucene/core/src/java/org/apache/lucene/util/IOSupplier.java

Will make the change to use it.

benwtrent (approved):
So CLEAN!

tteofili (approved):

PR #13530, opened 2024-06-28
Fix more vector similarity query tests

benwtrent:
Related to https://github.com/apache/lucene/pull/13260

The test failure is in 9x, but we need to add the fix here.

@msokolov I am planning to backport your code & then add this test fix. Do you agree?


msokolov (approved):

msokolov:
yes, please go ahead and backport!

Issue #13325, opened 2024-04-29
Examine performance of individual data accessor methods of MemorySegmentIndexInput


ChrisHegarty:
Examine performance of individual data accessor methods of MemorySegmentIndexInput, as compared to the ByteBufferIndexInput. Specifically getByte, getShort, and getBytes.

This issue has been filed to raise awareness of a seemingly perf degradation of MemorySegmentIndexInput when used heavily in decompression, see https://github.com/dacapobench/dacapobench/issues/264. I've not yet investigated the low-level segment guards, but maybe there is something here (or in the VarHandles) that is not optimising?

mikemccand:
I've also opened https://github.com/mikemccand/luceneutil/issues/267 to understand why our nightly benchmarks didn't notice this.  @uschindler maybe you have an idea!

uschindler:
The issue is analyzed already: The benchmark is wrong. It closes and opens indexes all the time. Due to the additional safety applied on closing indexes (to not crush the JVM with SIGSEGV), the JVM signals to all other threads that they need to go into safepoint to do the thread-local handshake.

The benchmakr should be rewritten to use a single IndexReader instance with many threads and not open a separate one for short time all the time in search threads. This causes thread-local-handshake havoc.

uschindler:
So this issue is not in Lucene. The slowdown comes from incorrect use of Lucene. I am waiting for the confirmation by them.

ChrisHegarty:
Thanks for your quick reply @uschindler. Your analysis makes sense to me.

uschindler:
> Thanks for your quick reply @uschindler. Your analysis makes sense to me.

If they don't respond, I will rewrite their benchmark to only open the index once and then hammer it with different threads. If this is then identical speed the issue seen is confirmed to be caused by the thread-local handshakes.

Although this is incorrect use of Lucene, we should still try to investigate with Hotspot people why the handshakes/safepoints always lead to some kind of deoptimization in all threads accessing MemorySegments. This is not necessarily bad in normal workloads, but could lead to slowdowns in NRT use cases with high index refresh frequencies.

We could also add an option to MMapDirectory so you can make it ready for purely single-threaded use. I am not sure if this is useful in practise, but we could easily allow this. If another thread then reads from an index file it will trigger IllegalStateException.

uschindler:
We have the confimation: https://github.com/dacapobench/dacapobench/issues/264#issuecomment-2084907716

I think we can close this issue. 

[...]

PR #13535, opened 2024-07-02
Use a confined Arena for IOContext.READONCE

ChrisHegarty:
Use a confined Arena for IOContext.READONCE.

This change will require inputs opened with READONCE to be consumed and closed on the creating thread. Further testing and assertions will need to be added.

relates #13325

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/PendingSoftDeletes.java
+++ b/lucene/core/src/java/org/apache/lucene/index/PendingSoftDeletes.java
@@ -232,7 +232,7 @@ final class PendingSoftDeletes extends PendingDeletes {
                 segInfo
                     .getCodec()
                     .compoundFormat()
-                    .getCompoundReader(segInfo.dir, segInfo, IOContext.DEFAULT);
+                    .getCompoundReader(segInfo.dir, segInfo, IOContext.READONCE);
       } else {
         toClose = null;
         dir = segInfo.dir;
--- a/lucene/core/src/java/org/apache/lucene/index/PendingSoftDeletes.java
+++ b/lucene/core/src/java/org/apache/lucene/index/PendingSoftDeletes.java
@@ -232,7 +232,7 @@ final class PendingSoftDeletes extends PendingDeletes {
                 segInfo
                     .getCodec()
                     .compoundFormat()
-                    .getCompoundReader(segInfo.dir, segInfo, IOContext.READONCE);
+                    .getCompoundReader(segInfo.dir, segInfo, IOContext.DEFAULT);
       } else {
         toClose = null;
         dir = segInfo.dir;
--- a/lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
+++ b/lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
@@ -578,8 +578,7 @@ final class ReadersAndUpdates {
       // IndexWriter.commitMergedDeletes).
       final SegmentReader reader;
       if (this.reader == null) {
-        IOContext context = info.info.getUseCompoundFile() ? IOContext.DEFAULT : IOContext.READONCE;
-        reader = new SegmentReader(info, indexCreatedVersionMajor, context);
+        reader = new SegmentReader(info, indexCreatedVersionMajor, IOContext.READONCE);
         pendingDeletes.onNewReader(reader, info);
       } else {
         reader = this.reader;
--- a/lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
+++ b/lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
@@ -578,7 +578,8 @@ final class ReadersAndUpdates {
       // IndexWriter.commitMergedDeletes).
       final SegmentReader reader;
       if (this.reader == null) {
-        reader = new SegmentReader(info, indexCreatedVersionMajor, IOContext.READONCE);
+        IOContext context = info.info.getUseCompoundFile() ? IOContext.DEFAULT : IOContext.READONCE;
+        reader = new SegmentReader(info, indexCreatedVersionMajor, context);
         pendingDeletes.onNewReader(reader, info);
       } else {
         reader = this.reader;
--- a/lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
+++ b/lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
@@ -103,12 +103,6 @@ abstract class MemorySegmentIndexInput extends IndexInput
     }
   }
 
-  void ensureNotConfined() {
-    if (confined) {
-      throw new IllegalStateException("confined");
-    }
-  }
-
   void ensureAccessible() {
     if (confined && curSegment.isAccessibleBy(Thread.currentThread()) == false) {
       throw new IllegalStateException("confined");
@@ -536,7 +530,6 @@ abstract class MemorySegmentIndexInput extends IndexInput
 
   @Override
   public final MemorySegmentIndexInput clone() {
-    ensureNotConfined();
     final MemorySegmentIndexInput clone = buildSlice((String) null, 0L, this.length);
     try {
       clone.seek(getFilePointer());
--- a/lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
+++ b/lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
@@ -109,6 +109,12 @@ abstract class MemorySegmentIndexInput extends IndexInput
     }
   }
 
+  void ensureAccessible() {
+    if (confined && curSegment.isAccessibleBy(Thread.currentThread()) == false) {
+      throw new IllegalStateException("confined");
+    }
+  }
+
   // the unused parameter is just to silence javac about unused variables
   RuntimeException handlePositionalIOOBE(RuntimeException unused, String action, long pos)
       throws IOException {
@@ -530,6 +536,7 @@ abstract class MemorySegmentIndexInput extends IndexInput
 
   @Override
   public final MemorySegmentIndexInput clone() {
+    ensureNotConfined();
     final MemorySegmentIndexInput clone = buildSlice((String) null, 0L, this.length);
     try {
       clone.seek(getFilePointer());
@@ -582,7 +589,7 @@ abstract class MemorySegmentIndexInput extends IndexInput
   /** Builds the actual sliced IndexInput (may apply extra offset in subclasses). * */
   MemorySegmentIndexInput buildSlice(String sliceDescription, long offset, long length) {
     ensureOpen();
-    ensureNotConfined();
+    ensureAccessible();
 
     final long sliceEnd = offset + length;
     final int startIndex = (int) (offset >>> chunkSizePower);
--- a/lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
+++ b/lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
@@ -53,6 +53,7 @@ abstract class MemorySegmentIndexInput extends IndexInput
   final long length;
   final long chunkSizeMask;
   final int chunkSizePower;
+  final boolean confined;
   final Arena arena;
   final MemorySegment[] segments;
 
@@ -67,12 +68,15 @@ abstract class MemorySegmentIndexInput extends IndexInput
       Arena arena,
       MemorySegment[] segments,
       long length,
-      int chunkSizePower) {
+      int chunkSizePower,
+      boolean confined) {
     assert Arrays.stream(segments).map(MemorySegment::scope).allMatch(arena.scope()::equals);
     if (segments.length == 1) {
-      return new SingleSegmentImpl(resourceDescription, arena, segments[0], length, chunkSizePower);
+      return new SingleSegmentImpl(
+          resourceDescription, arena, segments[0], length, chunkSizePower, confined);
     } else {
-      return new MultiSegmentImpl(resourceDescription, arena, segments, 0, length, chunkSizePower);
+      return new MultiSegmentImpl(
+          resourceDescription, arena, segments, 0, length, chunkSizePower, confined);
     }
   }
 
@@ -81,12 +85,14 @@ abstract class MemorySegmentIndexInput extends IndexInput
       Arena arena,
       MemorySegment[] segments,
       long length,
-      int chunkSizePower) {
+      int chunkSizePower,
+      boolean confined) {
     super(resourceDescription);
     this.arena = arena;
     this.segments = segments;
     this.length = length;
     this.chunkSizePower = chunkSizePower;
+    this.confined = confined;
     this.chunkSizeMask = (1L << chunkSizePower) - 1L;
     this.curSegment = segments[0];
   }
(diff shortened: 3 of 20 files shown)


ChrisHegarty:
Loopy testing runs without issue for 100+ times ( I just stopped it at that point ).

```
$ export x=; while ./gradlew :lucene:core:test ; do echo $x | wc -c; export x=x$x; done
```

uschindler (approved):
see my previous comment, but that looks simple.

We can work on improving the error messages. Possibly we should add some test that verifies that we get a valid 

When backporting to 9.x we may need to change the logic a bit as we have renamed the IOContext constants and have getter there to check for read-once status.

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
```diff
@@ -45,7 +45,12 @@ public IndexInput openInput(Path path, IOContext context, int chunkSizePower, bo
     path = Unwrappable.unwrapAll(path);
 
     boolean success = false;
-    final Arena arena = Arena.ofShared();
+    Arena arena;
```
uschindler, line 48: you can keep it final and use a ternary operator

uschindler:
When backporting we need to apply same changes for java 19 and 20.

ChrisHegarty:
> Cool. Seems useful to achieve the goal.
> 
> As written in the original issue maybe we should disallow clones, random access and slices of IndexInput on top of that. This may allow us to find more bugs.

Yeah,  we can either pass in a boolean, or something based on `MemorySegment::isAccessibleBy`. The former seems more straightforward, while the latter seems more flexible to allow slices, etc, but just for the calling thread. But that seems likely incorrect usage and not something we need to add complexity for.

> In general we should document thois somehow on the methods in the base class Directory that the IOContext has implications on usage from multiple threads.

++ I'll see if I can find, at least, some minimal wording that we can add here.

uschindler:
I also grepped through source code: https://github.com/search?q=repo%3Aapache%2Flucene%20READONCE&type=code

It looks like `IOContext.READONCE` is used at all places we are interested in. ChecksumIndexInput is not compatible to multithreading anyways (you can only read sequentially) and the other ones like reading lifedocs also look correct.

There are some occurrences with index sorting. We should make sure that there is no multithreading (optionally) used.

uschindler:
Maybe run all tests with `-Ptests.directory=MMapDirectory`; this raises chance that we hit a bug.

ChrisHegarty:
loopy testing with `-Ptests.directory=MMapDirectory` all successfully after several dozen runs.

uschindler:
Looks ok to me. Maybe ask @jpountz for his opinion; maybe he has more ideas where we only work single threaded.

uschindler:
Not sure about this: We could possibly also modify the general Exception handler which catches IllegalStateException and rethrow it as IOException.  We do this already for closed inputs to throw EOFException. Not sure about pros and cons, but when we add more tests and an "official documentation" for the READONCE case, this might be useful.

jpountz (commented):
I believe that this could be a problem for cross-data-structure merging concurrency (which we just disabled, but would like to re-enable soon-ish) since merging uses `READONCE`. In this case, I'd expect a `SegmentReader` to be opened in one thread, and then `XXXProducer#getMergeInstance()` (a `clone()` optimized for merging) to be called on the thread where the actual merging happens. So while the clone would be consumed in the thread where it's been created, the arena was previously initialized in a different thread?

uschindler:
> I believe that this could be a problem for cross-data-structure merging concurrency (which we just disabled, but would like to re-enable soon-ish) since merging uses `READONCE`. In this case, I'd expect a `SegmentReader` to be opened in one thread, and then `XXXProducer#getMergeInstance()` (a `clone()` optimized for merging) to be called on the thread where the actual merging happens. So while the clone would be consumed in the thread where it's been created, the arena was previously initialized in a different thread?

Merging uses sequential read ahead but from my review it never passes the constant IOContext.READONCE (see the link with references to READONCE above).

Merging uses an IOContext with a MergeInfo attached. This uses same ReadAdvice SEQUENTIAL, but that different to READONCE. Or do I miss something?

ChrisHegarty:
>Merging uses an IOContext with a MergeInfo attached. This uses same ReadAdvice SEQUENTIAL, but that is different to the Singleton READONCE IOContext. Or do I miss something?

I see same as you @uschindler, however ... 

An alternative to modifying the semantics of the existing `IOContext.READONCE` constant is to add a new constant that indicates both the read-once, sequential, and same-thread semantics. Then opt-in use sites as appropriate, for example, ChecksumIndexInput, etc.

uschindler:
P.S.: Because of this I proposed that READONCE IndexInputs should throw exceptions on clone/slice.

uschindler:
bq. I'm contemplating making MockDirectoryWrapper also fail if IndexInputs get used in a different thread from the one where they were created.

You mean this for READONCE ones, correct?

ChrisHegarty:
I updated clone and slice. It finds some issues. These look like incorrect usage of READONCE.

ChrisHegarty:
... there's still some intermittent failures because of this. I'll track them down so that we can discuss.

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
```diff
@@ -45,7 +45,8 @@ public IndexInput openInput(Path path, IOContext context, int chunkSizePower, bo
     path = Unwrappable.unwrapAll(path);
 
     boolean success = false;
-    final Arena arena = Arena.ofShared();
+    final boolean confined = context == IOContext.READONCE;
+    final Arena arena = context == IOContext.READONCE ? Arena.ofConfined() : Arena.ofShared();
```
uschindler, line 49: use the confined boolean here, too.

uschindler:
What intermittent failures did you see?

uschindler:
To me those problems where a CFS file was opened with READONCE look like a bug. It is also unlikely that the CFS file is just opened once and then closed again.

If we want to keep the CFS file support, we should at least allow slices of IndexInput. Whenever you create clones its a bug., because those are used in different threads, but slices may be acceptable (e.g. for CFS files).

uschindler:
> To me those problems where a CFS file was opened with READONCE look like a bug. It is also unlikely that the CFS file is just opened once and then closed again.
> 
> If we want to keep the CFS file support, we should at least allow slices of IndexInput. Whenever you create clones its a bug., because those are used in different threads, but slices may be acceptable (e.g. for CFS files).

Maybe the pending deletes stuff really only wants to open the CFS file to read deleted docs, but this still looks strange to me.

uschindler:
Thanks, looks fine now. I think we can revert the changes to the SegmentReader and look into this in another issue. It still looks strage not me, but I had not time to look closely into the code of the parts which used READONCE on CFS files, that you now changed to DEFAULT.

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
```diff
@@ -578,7 +578,8 @@ public synchronized boolean writeFieldUpdates(
       // IndexWriter.commitMergedDeletes).
       final SegmentReader reader;
       if (this.reader == null) {
-        reader = new SegmentReader(info, indexCreatedVersionMajor, IOContext.READONCE);
+        IOContext context = info.info.getUseCompoundFile() ? IOContext.DEFAULT : IOContext.READONCE;
```
uschindler, line 581: This looks correct with READONCE, as we close the segment reader down there, so it is only open for short time and we never look into it by different threads.

So I think we can revert that change and use READONCE. It is still a bit expensive to mmap a huge CFS file  just to read a few bytes.

lucene/core/src/java/org/apache/lucene/index/PendingSoftDeletes.java
```diff
@@ -232,7 +232,7 @@ private FieldInfos readFieldInfos() throws IOException {
                 segInfo
                     .getCodec()
                     .compoundFormat()
-                    .getCompoundReader(segInfo.dir, segInfo, IOContext.READONCE);
+                    .getCompoundReader(segInfo.dir, segInfo, IOContext.DEFAULT);
```
uschindler, line 235: READONCE is also fine here, because we only open the CFS file for very short time and close it down there. So as said before, this is a bit of overhead to mmap a large CFS file just to read fieldinfos.

We should think about this a bit in a separate issue. It is good you catched it.

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
```diff
@@ -53,6 +53,7 @@ abstract class MemorySegmentIndexInput extends IndexInput
   final long length;
   final long chunkSizeMask;
   final int chunkSizePower;
+  final boolean confined;
```
uschindler, line 56: We could also pass down the IOContext for more flexibility in the future.

For now a single boolean is fine!

uschindler (changes requested):

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
```diff
@@ -591,15 +611,17 @@ MemorySegmentIndexInput buildSlice(String sliceDescription, long offset, long le
           null, // clones don't have an Arena, as they can't close)
           slices[0].asSlice(offset, length),
           length,
-          chunkSizePower);
+          chunkSizePower,
+          confined);
```
uschindler, line 615: passing here confined for the slice looks correct, because this would disallow clones of possible slices.

+1

ChrisHegarty: I added a test for this.

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
ChrisHegarty: Reverting this results in the following failure:

```
    java.lang.IllegalStateException: confined
        at __randomizedtesting.SeedInfo.seed([E35616674660C0C9:EB8233FC2A66214D]:0)
        at org.apache.lucene.store.MemorySegmentIndexInput.ensureNotConfined(MemorySegmentIndexInput.java:108)
        at org.apache.lucene.store.MemorySegmentIndexInput.clone(MemorySegmentIndexInput.java:539)
        at org.apache.lucene.store.MemorySegmentIndexInput$SingleSegmentImpl.clone(MemorySegmentIndexInput.java:661)
        at org.apache.lucene.tests.store.MockIndexInputWrapper.clone(MockIndexInputWrapper.java:94)
        at org.apache.lucene.tests.store.MockIndexInputWrapper.clone(MockIndexInputWrapper.java:30)
        at org.apache.lucene.codecs.CodecUtil.checksumEntireFile(CodecUtil.java:607)
        at org.apache.lucene.codecs.blocktreeords.OrdsBlockTreeTermsReader.<init>(OrdsBlockTreeTermsReader.java:100)
        at org.apache.lucene.tests.codecs.mockrandom.MockRandomPostingsFormat.fieldsProducer(MockRandomPostingsFormat.java:379)
        at org.apache.lucene.codecs.perfield.PerFieldPostingsFormat$FieldsReader.<init>(PerFieldPostingsFormat.java:328)
        at org.apache.lucene.codecs.perfield.PerFieldPostingsFormat.fieldsProducer(PerFieldPostingsFormat.java:390)
        at org.apache.lucene.index.SegmentCoreReaders.<init>(SegmentCoreReaders.java:98)
        at org.apache.lucene.index.SegmentReader.<init>(SegmentReader.java:95)
        at org.apache.lucene.index.ReadersAndUpdates.writeFieldUpdates(ReadersAndUpdates.java:581)
        at org.apache.lucene.index.ReaderPool.commit(ReaderPool.java:353)
        at org.apache.lucene.index.IndexWriter.writeReaderPool(IndexWriter.java:3990)
        at org.apache.lucene.index.IndexWriter.prepareCommitInternal(IndexWriter.java:3692)
        at org.apache.lucene.index.IndexWriter.commitInternal(IndexWriter.java:4122)
        at org.apache.lucene.index.IndexWriter.shutdown(IndexWriter.java:1331)
        at org.apache.lucene.index.IndexWriter.close(IndexWriter.java:1369)
        at org.apache.lucene.util.IOUtils.close(IOUtils.java:85)
        at org.apache.lucene.util.IOUtils.close(IOUtils.java:72)
        at org.apache.lucene.index.TestNumericDocValuesUpdates.testMultipleUpdatesSameDoc(TestNumericDocValuesUpdates.java:116)
...
```

`CodecUtil.checksumEntireFile` is cloning. It could just slice the whole input.

uschindler: Hm damn.

At least the other one only reading field Infos looks fine.

But this should also fail if it is not a compound file...

uschindler:
I am not really sure how we should proceed. The `new SegmentReader()` which is closed a few lines below looks like a valid use case for READ_ONLY, but SegmentReader creates clones. - damn!

Not sure how we should handle that. Maybe let's remove disallowing of clones and slices, but check that the threads are correct. @jpountz ?

About the CFS files we should really look if we can't get a "better way" to only request a single file from a huge CFS. But this is just an observation we have seen here and has nothing to do with this isse: It seems very bad to mmap a whole huge CFS file just to extract one child file from it in an readonce context. This is an idea for another issue to refactor the CFS Directory to better work with an underlying index input and maybe allow to mmap only slices from begining. This is more complicated to implement but might be worth to look into.

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
jpountz: > This looks correct with READONCE, as we close the segment reader down there, so it is only open for short time and we never look into it by different threads.

I keep being confused by `READONCE` because the name suggests that data is read "once" but often it's used to mean "mostly sequentially" but not necessarily "once", such as here: we'd pass the doc values reader from this `SegmentReader` to a doc-values consumer, which typically performs multiple passes on the data: a first one to collect stastistics, and a second one to actually write the data. So data is not actually read once, but multiple times. (We do have files that get read once, e.g. all metadata files: `.tmd`, `.fdm`, `.kdm`, etc.)

I'm unsure if we should rename `READONCE` or introduce a new context, but it looks like there is something worth fixing here.

Separately, `OrdsBlockTreeTermsReader` is bad: it stores index data and metadata next to one another in the same file, and needs to seek towards the end of the file at open time to read the offset where metadata starts. We should fix it by storing data and metadata in separate files.

jpountz:
>> I'm contemplating making MockDirectoryWrapper also fail if IndexInputs get used in a different thread from the one where they were created.
>
> You mean this for READONCE ones, correct?

I meant all `IndexInput`s actually, they should all be used in the thread where they were created (either via `Directory#openInput` or `IndexInput#clone`)? But I'm happy to look into it in a separate issue.

> Not sure how we should handle that. Maybe let's remove disallowing of clones and slices, but check that the threads are correct. @jpountz ?

This sounds like ok progress over perferction to me. We can look into getting stricter in a followup.

ChrisHegarty:
Thanks for the comments so far. I updated the PR to only check same-thread semantics for MSII clone and slice. And also added some basic thread checks to MockIndexInputWrapper.   I think this is a reasonable improvement and a good place to stop in this PR. As suggested, further improvements can be done as a follow up.

uschindler (approved):
Looks ok.

uschindler (approved):

PR #13536, opened 2024-07-03
Only search soft deleted in SoftDeletesRetentionMergePolicy.applyRetentionQuery

vsop-479:
### Description

Instead of `only search deleted`, we can `only search soft deleted` with `hardLiveDocs` in `SoftDeletesRetentionMergePolicy.applyRetentionQuery`.

But it seems we never use `wrapLiveDocs` to filter docs?

Can we extract `softLiveDocs` from `liveDocs` and `hardLiveDocs`, and use it directly, even instead of FieldExistsQuery?


vsop-479:
> Can we extract softLiveDocs from liveDocs and hardLiveDocs, and use it directly, even instead of FieldExistsQuery?

Maybe it won't get any speed up.

vsop-479:
I think we can get `softLiveDocs` by from liveDocs and hardLiveDocs by using XNOR (https://github.com/apache/lucene/pull/13540), and use it replace `wrapLiveDocs`. 

Or, we can remove `wrapLiveDocs`, if it is useless.

PR #13538, opened 2024-07-03
Reduce heap usage for knn index writers

benwtrent:
This slightly reduces heap utilization for KnnIndex writers by:

 - Only constructing one `DocsWithFieldSet` instead of 3 for Scalar quantized HNSW formats
 - Only having one `List<T>` instead of 3 for Scalar quantized HNSW formats
 

This does change an experimental API (FlatVectorFormats) slightly to allow reuse.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -280,6 +280,8 @@ Optimizations
 
 * GITHUB#13175: Stop double-checking priority queue inserts in some FacetCount classes. (Jakub Slowinski)
 
+* GITHUB#13538: Slightly reduce heap usage for HNSW and scalar quantized vector writers. (Ben Trent)
+
 Changes in runtime behavior
 ---------------------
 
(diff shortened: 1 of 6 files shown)


gautamworah96:
Thanks for the quick turnaround on this Ben. I am not too familiar with this code and am taking some time off this week. I'll try to review the code by Friday. 

In the meantime, I am also trying to understand some core IndexWriter related concepts. It seems that tuning the `indexWriterBufferSizeMB` to a lower value seems to be helping (which makes sense because the sum of length of the float[] array across segments being flushed is small, I am trying to see if I can get larger segments by keeping merges enabled, in the email thread I had mentioned that my merges had been disabled).

gautamworah96 (approved):

benwtrent:
Related: https://github.com/apache/lucene/pull/13553

tteofili (approved):

PR #13543, opened 2024-07-04
Override single byte writes to OutputStreamIndexOutput to remove locking

original-brownbear:
Single byte writes to BufferedOutputStream show up pretty hot in indexing benchmarks. We can save the locking overhead introduced by JEP374 by overriding and providing a no-lock fast path.

Didn't benchmark this with Lucene benchmarks but there's this https://gist.github.com/shipilev/71fd881f7cbf3f87028bf87385e84889 and the issue is somewhat obvious looking at the profiling for the nightlies I believe.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/OutputStreamIndexOutput.java
+++ b/lucene/core/src/java/org/apache/lucene/store/OutputStreamIndexOutput.java
@@ -135,5 +135,19 @@ public class OutputStreamIndexOutput extends IndexOutput {
       BitUtil.VH_LE_LONG.set(buf, count, i);
       count += Long.BYTES;
     }
+
+    @Override
+    public void write(int b) throws IOException {
+      // override single byte write to avoid synchronization overhead now that JEP374 removed biased
+      // locking
+      byte[] buffer = buf;
+      int count = this.count;
+      if (count >= buffer.length) {
+        super.write(b);
+      } else {
+        buffer[count] = (byte) b;
+        this.count = count + 1;
+      }
+    }
   }
 }


lucene/core/src/java/org/apache/lucene/store/OutputStreamIndexOutput.java
```diff
@@ -135,5 +135,19 @@ void writeLong(long i) throws IOException {
...
+      // locking
+      byte[] buffer = buf;
+      int count = this.count;
+      if (count >= buffer.length) {
+        super.write(b);
+      } else {
+        buffer[count] = (byte) b;
+        this.count = count + 1;
+      }
+    }
```
jpountz, line 151: Should we implement it using the same pattern as `writeShort`/`writeInt`/`writeLong`?

original-brownbear: I think this version is a bit faster (and maybe more than just a bit even depending on the calling code) by avoiding the (most of the time unnecessary) method call to `flushIfNeeded` (which itself does a bunch of field lookups). For single-byte operations this matters somewhat I believe. It also saves potentially unnecessary flush calls to the underlying stream, though those might be irrelevant since we should only be hitting a single stream (from file output stream) with a noop flush implementation here as far as I can tell.

msfroh:
I was just looking at a flame graph for an indexing performance regression with JDK21 and noticed a lot of time spent in `BufferedOutputStream#growIfNeeded`. Without virtual threads, it looks like `growIfNeeded` (uselessly) does an addition and two comparisons, which doesn't sound like much, but may add up when `BufferedOutputStream#write(int)` is called a bajillion times.

This change should also help with that by cutting the number of calls to `BufferedOutputStream#write(int)` by a factor of 8192, which cuts the number of calls to `growIfNeeded` by the same factor.

Of course, if it followed the same pattern as `writeShort`/`writeInt`/`writeLong`, we could eliminate calls to `growIfNeeded` altogether.

msokolov:
> This change should also help with that by cutting the number of calls to BufferedOutputStream#write(int) by a factor of 8192, which cuts the number of calls to growIfNeeded by the same factor.

this change was merged: did you observe any change in the flame graphs?

msfroh:
> this change was merged: did you observe any change in the flame graphs?

Yes! 

We were benchmarking on Lucene 9.11 and saw the time spent in `growIfNeeded` on JDK21 (and about a 7% slowdown in indexing versus JDK17). I noticed this change in my local clone of main, cherry-picked it onto the 9.11 branch, and subsequent JDK21 runs caught up with JDK17. It helped a lot -- not for the biased locking deprecation (which went into JDK15), but because it bypassed these extra comparisons that were added to support virtual threads.

PR #13544, opened 2024-07-04
Optimize MaxScoreBulkScorer

original-brownbear:
Don't use Comparator.comparingDouble(...) in a hotish loop here, it causes allocations that escape analysis is not able to remove. => lets just manually inline this to get predictable behavior and save up to 0.5% of all allocations in some benchmark runs.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/MaxScoreBulkScorer.java
+++ b/lucene/core/src/java/org/apache/lucene/search/MaxScoreBulkScorer.java
@@ -332,12 +332,13 @@ final class MaxScoreBulkScorer extends BulkScorer {
     // make a difference when using custom scores (like FuzzyQuery), high query-time boosts, or
     // scoring based on wacky weights.
     System.arraycopy(allScorers, 0, scratch, 0, allScorers.length);
+    // Do not use Comparator#comparingDouble below, it might cause unnecessary allocations
     Arrays.sort(
         scratch,
-        (c1, c2) -> {
+        (scorer1, scorer2) -> {
           return Double.compare(
-              (double) c1.maxWindowScore / Math.max(1L, c1.cost),
-              (double) c2.maxWindowScore / Math.max(1L, c2.cost));
+              (double) scorer1.maxWindowScore / Math.max(1L, scorer1.cost),
+              (double) scorer2.maxWindowScore / Math.max(1L, scorer2.cost));
         });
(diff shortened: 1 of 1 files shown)


lucene/core/src/java/org/apache/lucene/search/MaxScoreBulkScorer.java
```diff
@@ -335,8 +334,11 @@ boolean partitionScorers() {
     System.arraycopy(allScorers, 0, scratch, 0, allScorers.length);
     Arrays.sort(
         scratch,
-        Comparator.comparingDouble(
-            scorer -> (double) scorer.maxWindowScore / Math.max(1L, scorer.cost)));
+        (c1, c2) -> {
```
jpountz, line 337: Nit: can you add a comment about not moving back to `Comparator#comparingDouble`, and maybe rename `c1`/`c2` to `scorer1`/`scorer2`?

jpountz (approved):
It's fine with me, do you have a benchmark / profiler output that shows the improvement?

lucene/core/src/java/org/apache/lucene/search/MaxScoreBulkScorer.java
original-brownbear: Done :)

original-brownbear:
Thanks Adrien!

> do you have a benchmark / profiler output that shows the improvement?

The nightlies show this:

```
0.49%         4090M         java.lang.invoke.DirectMethodHandle#allocateInstance()
                              at java.lang.invoke.DirectMethodHandle$Holder#newInvokeSpecial()
                              at java.lang.invoke.Invokers$Holder#linkToTargetMethod()
                              at java.util.Comparator#comparingDouble()
                              at org.apache.lucene.search.MaxScoreBulkScorer#partitionScorers()
                              at org.apache.lucene.search.MaxScoreBulkScorer#score()
                              at org.apache.lucene.search.BulkScorer#score()
                              at org.apache.lucene.search.IndexSearcher#search()
```

This change makes it go away :) So it's essentially saving 0.5% of all allocations here. I couldn't reproduce a big/signifcant QPS change as a result since we're not CPU bound when running the benchmark (this might change with the `AtomicLong` change in the other PR ;)), but I think saving those GBs from getting allocated in the first place is enough reason to fix this kind of spot :)

PR #13545, opened 2024-07-05
Improve VectorUtil::xorBitCount perf on ARM

ChrisHegarty:
This commit improves the performance of `VectorUtil::xorBitCount` on ARM by ~4x.

This change is effectively a workaround for the lack of vectorization of `Long::bitCount` on ARM, see https://github.com/ChrisHegarty/hammingBench/. I'll get an issue filed against Hotspot for this.  ( JDK bug tracking this issue: https://bugs.openjdk.org/browse/JDK-8336000 )

On x64 there is no issue, the long variant of xorBitCount outperforms the int variant by ~15%.
 
Before (measures throughput in seconds, so bigger numbers are better)
```
Benchmark                             (dims)     (nb)   Mode  Cnt   Score   Error  Units
HammingDistanceBenchmark.xorBitCount    1024  1000000  thrpt    5  29.128 ± 5.697  ops/s
```
 After
  ```
Benchmark                             (dims)     (nb)   Mode  Cnt   Score   Error  Units
HammingDistanceBenchmark.xorBitCount    1024  1000000  thrpt    5  115.430 ± 3.086  ops/s
```

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/VectorUtil.java
+++ b/lucene/core/src/java/org/apache/lucene/util/VectorUtil.java
@@ -213,6 +213,14 @@ public final class VectorUtil {
   }
 
   /**
+   * For xorBitCount we stride over the values as either 64-bits (long) or 32-bits (int) at a time.
+   * On ARM Long::bitCount is not vectorized, and therefore produces less than optimal code, when
+   * compared to Integer::bitCount. While Long::bitCount is optimal on x64. TODO: include the
+   * OpenJDK JIRA url
+   */
+  static final boolean XOR_BIT_COUNT_STRIDE_AS_INT = Constants.OS_ARCH.equals("aarch64");
+
+  /**
    * XOR bit count computed over signed bytes.
    *
    * @param a bytes containing a vector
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by jimczi — APPROVED** [review]

**Comment by uschindler** [comment]
Hi,
in the backport to 9.x the benchmark file was wrongly merged. It landed in the test directory. In 9.x we have no benchmark-jmh module in Gradle, so the file should have been left out while cherry-picking.
Can you remove the file in a followup commit?

File: lucene/core/src/java/org/apache/lucene/util/VectorUtil.java [hunk 1]
```diff
@@ -212,6 +212,14 @@ public static int int4DotProductPacked(byte[] unpacked, byte[] packed) {
     return IMPL.int4DotProduct(unpacked, false, packed, true);
   }
 
+  /**
+   * For xorBitCount we stride over the values as either 64-bits (long) or 32-bits (int) at a time.
+   * On ARM Long::bitCount is not vectorized, and therefore produces less than optimal code, when
+   * compared to Integer::bitCount. While Long::bitCount is optimal on x64. TODO: include the
+   * OpenJDK JIRA url
```
uschindler, line 219: Do you have the JIRA issue number already?

uschindler:
I reverted the addition of the file to 9.x branch: 86d080a4e0b4e53e0c9a3f2e2b120bff204c7276

ChrisHegarty:
@uschindler Apologies, I didn't notice this when cherrypicking. Thanks for reverting (while I was sleeping ;-) )

lucene/core/src/java/org/apache/lucene/util/VectorUtil.java
ChrisHegarty: Oops. I just added it in https://github.com/apache/lucene/commit/4baaedaa67046adeed19cdc28a346c1ebc9e8b5d

https://bugs.openjdk.org/browse/JDK-8336000

PR #13548, opened 2024-07-07
Refactor and javadoc update for KNN vector writer classes

zhaih:
### Description

* Refactor logic of translating old ord to new ord
* Javadoc for Lucene99HnswVectorsFormat is a bit outdated after we separate FlatVectorsFormat out, updated

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java
@@ -142,6 +142,14 @@ public abstract class KnnVectorsWriter implements Accountable, Closeable {
     }
   }
 
+  /**
+   * Given old doc ids and an id mapping, maps old ordinal to new ordinal. Note: this method return
+   * nothing and output are written to parameters
+   *
+   * @param old2NewOrd int[] maps from old ord to new ord
+   * @param new2OldOrd int[] maps from new ord to old ord
+   * @param newDocsWithField set of new doc ids which has the value
+   */
   public static void mapOldOrdToNewOrd(
       DocsWithFieldSet oldDocIds,
       Sorter.DocMap sortMap,
@@ -149,6 +157,8 @@ public abstract class KnnVectorsWriter implements Accountable, Closeable {
       int[] new2OldOrd,
       DocsWithFieldSet newDocsWithField)
       throws IOException {
+    // TODO: a similar function exists in IncrementalHnswGraphMerger#getNewOrdMapping
+    //       maybe we can do a further refactoring
     assert (old2NewOrd != null || new2OldOrd != null || newDocsWithField != null);
     IntIntHashMap newIdToOldOrd = new IntIntHashMap();
     DocIdSetIterator iterator = oldDocIds.iterator();
--- a/lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java
@@ -20,14 +20,17 @@ package org.apache.lucene.codecs;
 import java.io.Closeable;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import org.apache.lucene.index.ByteVectorValues;
 import org.apache.lucene.index.DocIDMerger;
+import org.apache.lucene.index.DocsWithFieldSet;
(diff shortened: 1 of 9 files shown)


lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatVectorsFormat.java
```diff
@@ -56,8 +56,8 @@
  *   <li><b>[vlong]</b> length of this field's vectors, in bytes
  *   <li><b>[vint]</b> dimension of this field's vectors
  *   <li><b>[int]</b> the number of documents having values for this field
- *   <li><b>[int8]</b> if equals to -1, dense – all documents have values for a field. If equals to
- *       0, sparse – some documents missing values.
+ *   <li><b>[int8]</b> if equals to -2, empty - no vectory values. If equals to -1, dense – all
```
benwtrent, line 59:
```suggestion
 *   <li><b>[int8]</b> if equals to -2, empty - no vector values. If equals to -1, dense – all
```

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java
```diff
@@ -139,6 +142,54 @@ public int nextDoc() throws IOException {
     }
   }
 
+  /**
+   * Given old doc ids and an id mapping, maps old ordinal to new ordinal. Note: this method return
+   * nothing and output are written to parameters
+   *
```
benwtrent, line 148:
```suggestion
   * @param oldDocIds the old or current document ordinals. Must not be null.
   * @param sortMap, the document sorting map for how to make the new ordinals. Must not be null.
```

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java
```diff
@@ -139,6 +142,54 @@ public int nextDoc() throws IOException {
...
+  public static void mapOldOrdToNewOrd(
+      DocsWithFieldSet oldDocIds,
+      Sorter.DocMap sortMap,
+      int[] old2NewOrd,
+      int[] new2OldOrd,
+      DocsWithFieldSet newDocsWithField)
+      throws IOException {
+    // TODO: a similar function exists in IncrementalHnswGraphMerger#getNewOrdMapping
+    //       maybe we can do a further refactoring
+    assert (old2NewOrd != null || new2OldOrd != null || newDocsWithField != null);
```
benwtrent, line 162:
```suggestion
    Objects.requireNonNull(oldDocIds);
    Objects.requireNonNull(sortMap);
    assert (old2NewOrd != null || new2OldOrd != null || newDocsWithField != null);
```

lucene/core/src/java/org/apache/lucene/index/FieldUpdatesBuffer.java
```diff
@@ -356,7 +356,7 @@ BufferedUpdate next() throws IOException {
       }
     }
 
-    BytesRef nextTerm() throws IOException {
+    private BytesRef nextTerm() throws IOException {
```
benwtrent, line 359: seems like an extra thing out of nowhere? Its probably ok in this commit, just struck me as a strange inclusion :)

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java
```diff
@@ -139,6 +142,54 @@ public int nextDoc() throws IOException {
...
+    IntIntHashMap newIdToOldOrd = new IntIntHashMap();
```
benwtrent, line 163: Since this method isn't in charge of creating `old2NewOrd` or `new2OldOrd` can we assert those array sizes they are provided?

benwtrent (commented):
Overall looks good. Some minor concerns to ensure the new method isn't abused.

lucene/core/src/java/org/apache/lucene/index/FieldUpdatesBuffer.java
zhaih: Yeah I just edited it when I was going over the code, thought it should be fine leaving it here :)

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java
zhaih: good idea

benwtrent (approved):

PR #13054, opened 2024-01-30
Move synonym map off-heap for SynonymGraphFilter

msfroh:
### Description
This stores the synonym map's FST and word lookup off-heap in a separate, configurable directory. 

The initial implementation is rough, but the unit tests pass with this change randomly enabled.

Obvious things that need work are:
1. I tried to do something like a codec, but not really a codec for the synonym map files. For a solution that could evolve over time, we should probably at least write something to the metadata file saying what format was used.
2. Right now it makes no effort to detect changes to the synonym files. I would suggest that SynonymGraphFilterFactory rebuild the directory if a checksum of the input files doesn't match a value recorded in the metadata file.
3. I don't think I like the random seeks in `OffHeapBytesRefHashLike`, but I don't see an alternative (besides moving it on-heap). Given that the original issue was only about moving the FST off-heap, maybe we can keep the word dictionary on-heap.

Fixes #13005

PR #13549, opened 2024-07-08
Refactor FST.saveMetadata() to FSTMetadata.save()

dungba88:
### Description

A precursor change for both https://github.com/apache/lucene/pull/12985 and https://github.com/apache/lucene/pull/13054. These are cases where we only want to save the FST metadata without the need to construct a FST object. Move the save method out of FST enable these use cases.


mikemccand (approved):
Thanks @dungba88 -- this looks like a simple refactoring, and makes it easier for other uses (the two PRs you linked) to separately save the `FSTMetadata`.

dungba88:
Thank you for merging @mikemccand !

PR #13550, opened 2024-07-08
Introduce TestLucene90DocValuesFormatVariableSkipInterval for testing docvalues skipper index

iverase:
In https://github.com/apache/lucene/pull/13449 we introduced a sparse index on top of doc values. One of the challenges was to properly test it as the default interval size is 4096 documents so we need a fair amount of documents to have a good number of intervals.

Therefore in this commit we propose to be able to configure dynamically the interval size for testing, and add a new test that changes the interval size randomly.


lucene/test-framework/src/java/org/apache/lucene/tests/codecs/skipper/SkipperCodec.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.lucene.codecs.DocValuesFormat;
+import org.apache.lucene.codecs.FilterCodec;
+import org.apache.lucene.codecs.lucene90.Lucene90DocValuesFormat;
+import org.apache.lucene.tests.util.TestUtil;
+
+/**
+ * A codec that uses {@link Lucene90DocValuesFormat} with a random skip index size for its doc
+ * values and delegates to the default codec for everything else.
+ */
+public final class SkipperCodec extends FilterCodec {
```
jpountz, line 29: Maybe rename to `Skipper90Codec` to reflect that it maps to `Lucene90DocValuesFormat`? I assume we'd create a new one when we release a new DV format?

Also, does it need to be a new Codec, or could it extend `Lucene911Codec` and override `getDocValuesFormatForField`?

lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseDocValuesFormatTestCase.java
```diff
@@ -157,6 +158,7 @@ public void testNumberMergeAwayAllValuesWithSkipper() throws IOException {
     Analyzer analyzer = new MockAnalyzer(random());
     IndexWriterConfig iwconfig = newIndexWriterConfig(analyzer);
     iwconfig.setMergePolicy(newLogMergePolicy());
+    iwconfig.setCodec(SkipperCodec.randomInstance(random()));
```
jpountz, line 161: This doesn't look right, as this overrides the codec that is configured on the test case, e.g. `SimpleTextDocValuesFormat` would no longer be exercised? Should we create a new `TestLucene90DocValuesFormatVariableSkipInterval` class or something along these lines instead?

iverase: sure

lucene/test-framework/src/java/org/apache/lucene/tests/codecs/skipper/SkipperCodec.java
iverase: right, I was too complicated

iverase:
@jpountz I added a new test and change the title and description of the issue as we don't need to add a new Codec.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
```diff
@@ -96,6 +97,7 @@ public Lucene90DocValuesConsumer(
           state.segmentInfo.getId(),
           state.segmentSuffix);
       maxDoc = state.segmentInfo.maxDoc();
+      this.skipIndexIntervalSize = skipIndexIntervalSize;
```
jpountz, line 100: Do we make any assumption anywhere that this needs to be a power of two? If so we should validate it?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesFormat.java
```diff
@@ -138,15 +138,23 @@
...
 public final class Lucene90DocValuesFormat extends DocValuesFormat {
 
+  private final int skipIndexIntervalSize;
+
   /** Default constructor. */
   public Lucene90DocValuesFormat() {
+    this(DEFAULT_SKIP_INDEX_INTERVAL_SIZE);
+  }
+
+  /** Default constructor. */
```
jpountz, line 148: Nit: this is no longer the default ctor.

lucene/core/src/test/org/apache/lucene/codecs/lucene90/TestLucene90DocValuesFormatVariableSkipInterval.java
```diff
@@ -0,0 +1,36 @@
...
+public class TestLucene90DocValuesFormatVariableSkipInterval extends BaseDocValuesFormatTestCase {
+
+  @Override
+  protected Codec getCodec() {
+    return new Lucene99Codec() {
+      @Override
+      public DocValuesFormat getDocValuesFormatForField(String name) {
+        return new Lucene90DocValuesFormat(random().nextInt(1 << 3, 1 << 10));
+      }
+    };
```
jpountz, line 34: Let's do `TestUtil.alwaysDocValuesFormat(new Lucene90DocValuesFormat(random().nextInt(1 << 3, 1 << 10)))`? This will add some assertions to the mix, and will make it easier when we bump the codec version.

jpountz (approved):

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
iverase: No , there is no assumption. On the other hand I added in the codec a check so skipIndexIntervalSize must be 1. In addtion I replace 1 << x by an integer to remove the suggestion that this is required.

lucene/core/src/test/org/apache/lucene/codecs/lucene90/TestLucene90DocValuesFormatVariableSkipInterval.java
iverase: sure

jpountz (approved):

PR #13553, opened 2024-07-09
Fix quantized vector writer ram estimates

benwtrent:
I still need to write a test, but wanted to open this PR early.

Scalar Quantized vector writer ram usage estimates completely ignores the raw float vectors. Meaning, if you have flush based on ram usage configured, you can easily overshoot that estimate and cause an OOM.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
@@ -171,6 +171,7 @@ public final class Lucene99HnswVectorsWriter extends KnnVectorsWriter {
   @Override
   public long ramBytesUsed() {
     long total = SHALLOW_RAM_BYTES_USED;
+    // The vector delegate will also account for this writer's KnnFieldVectorsWriter objects
     total += flatVectorWriter.ramBytesUsed();
     return total;
   }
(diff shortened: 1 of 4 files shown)


lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -172,9 +172,6 @@ public void finish() throws IOException {
   public long ramBytesUsed() {
     long total = SHALLOW_RAM_BYTES_USED;
     total += flatVectorWriter.ramBytesUsed();
-    for (FieldWriter<?> field : fields) {
-      total += field.ramBytesUsed();
-    }
```
benwtrent, line 177: the delegate writer will take into account these higher level indexing writers.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99ScalarQuantizedVectorsWriter.java
```diff
@@ -299,9 +299,7 @@ public void finish() throws IOException {
   @Override
   public long ramBytesUsed() {
     long total = SHALLOW_RAM_BYTES_USED;
-    for (FieldWriter field : fields) {
-      total += field.ramBytesUsed();
-    }
+    total += rawVectorDelegate.ramBytesUsed();
```
benwtrent, line 302: rawVectorDelegate takes into account its field writer estimates which then take the scalar quantized writer estimates into account.

benwtrent:
@gautamworah96 @msokolov this might be part of the reason for the OOMs, the estimates were completely ignoring the float[] vector sizes for fieldwriters 🤦 . I plan on iterating on this fix first, then proceed with https://github.com/apache/lucene/pull/13538 

The other PR helps reduce heap usage slightly, but not by much. Honestly, the issue is that we just weren't tracking the usage in the estimate correctly.

msokolov (approved):
Thanks for digging! It makes sense that this would have led to OOM

gautamworah96:
Interesting. I didn't know the ram usage estimator is what is actually used for flushing based on ram size. I used to think it was only for external monitoring/alarming, and some other IndexWriter based store would actually do the ram usage calculation. 
Thanks for the fix.

PR #13555, opened 2024-07-09
Group memory arenas by segment to reduce costly `Arena.close()`

magibney:


magibney:
Follows an approach analogous to the ["custom arenas"](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/foreign/Arena.html#custom-arenas) case described in Arena javadocs.

lucene/core/src/java21/org/apache/lucene/store/GroupedArena.java
```diff
@@ -0,0 +1,212 @@
...
+  static Arena get(Path p, ConcurrentHashMap<String, GroupedArena> arenas) {
+    String filename = p.getFileName().toString();
+    String segmentName = IndexFileNames.parseSegmentName(filename);
+    if (filename.length() == segmentName.length()) {
+      // no segment found; return a 1-off Arena
+      return Arena.ofShared();
+    }
+    String scopeId = p.getParent().resolve(segmentName).toString();
+    Arena ret;
+    do {
```
uschindler, line 49: this code is a bit strange, also with the boolean[] as modifiable holder. I'd rewrite this.

Basically the code should better check the refcount of the GroupedArena and if it is 0 return itself, incrementing by 1. For all other cases it shoiuld just return a clone with incremented refcount.

Ret should never be 0, please remove the code that could make it null in the coloneIfActive(). While loop shouldn't be needed.

lucene/core/src/java21/org/apache/lucene/store/GroupedArena.java
```diff
@@ -0,0 +1,212 @@
...
+        //  above, they'll simply create a new Arena, and we're no worse off than we
+        //  would have been if every Arena was created as a one-off.
+        backing.close();
+      }
+    } else {
+      assert ct > 0 : "refCt should never be negative; found " + ct;
+    }
+  }
+
+  @Override
```
uschindler, line 108: I am not so happy that we need to implement all those methods. Let's keep the default ones.
Maybe let the required ones throw UOE, because we never use the arena to allocate memory.

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
```diff
@@ -24,12 +24,15 @@
...
 import java.nio.file.StandardOpenOption;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 import org.apache.lucene.util.Constants;
 import org.apache.lucene.util.Unwrappable;
 
 @SuppressWarnings("preview")
 final class MemorySegmentIndexInputProvider implements MMapDirectory.MMapIndexInputProvider {
 
+  private static final ConcurrentHashMap<String, GroupedArena> ARENAS = new ConcurrentHashMap<>();
```
uschindler, line 34: I don't like that it is a static global arena. It should be private per MMapDirectory instance. This should be doable, as we can pass the map to `GroupedArena.get(path, arena)`. So it can also be a private per-instance one.

If you do this, the file name is the only key needed. The current code with adding full path is only needed because it is global.

The ConcurrentHashMap expeczted should be typed `ConcurrentHashMap<String, ?>` as `Arena` should never escape the private Java 21 code.

I am afraid of memory leaks caused by this.

uschindler (changes requested):
I like the idea, but I have a few things to think about:
- Make the ConcurrentHashMap a per-MMapDirectory instance field (see comments about type issues). The value should mabye a wildcard type param `?`
- I am not sure if we need to clone the arena for each new clone. We can reuse the same instance and just modify it's close to call backing close when refcount reaches 0.
- I don't like to implement all SegmentAllocator methods. Maybe we can change the required ones to throw UOE. Anyways, we don't need to override default impls.

Otherwise, nice idea. We should benchmark it!.

lucene/core/src/java21/org/apache/lucene/store/GroupedArena.java
ChrisHegarty: ++

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
ChrisHegarty: yes, let's make this an instance field on MMapDirectory.

uschindler: Maybe we can make it `CHM<String,Closeable>`.

lucene/core/src/java21/org/apache/lucene/store/GroupedArena.java
magibney: Ah right; I hadn't bothered looking at how Arena is actually used. Should we not throw UOE for _all_ of them? All of these have default implementations, so we end up only having to override `scope()` and `close()`, which are the only required to implement (and also the only ones we actually call).

lucene/core/src/java21/org/apache/lucene/store/GroupedArena.java
magibney: 60ff62d9cdd9dec584b8e6ce63e4a7af13b828a2 rewrites this (without the need to clone, and without the `boolean[]` modifiable holder), but I still think we may need the while loop; I tried to better explain my reasoning in the comment:
```java
    // We loop below to protect against the possibility that we get an entry that is
    // in the process of being closed. It is unlikely (perhaps impossible?) that this
    // would happen in practice, based on how Lucene opens IndexInputs. But if we
    // don't check here it's theoretically possible that our return value would be
    // backed by an Arena that is closed (and would throw IllegalStateException).
    // NOTE also: if we admit the possibility that a `Directory` can be used for
    // purposes other than a Lucene index, then we don't have the implicit protection
    // conferred by expected usage patterns, and it becomes all the more important to
    // loop to protect against the possibility of concurrent open/close.
```

magibney:
Thanks for the suggestions! Most of these are now addressed (with a couple of questions: 1. maybe still need chm.computeIfAbsent() loop, 2. which if any Arena methods to throw UOE).

I also wanted to call attention to another concern that occurred to me, described in a comment:
```java
      // TODO: what if the associated Directory is used for some purpose other than Lucene
      //  index files, and some of the filenames happen to be patterned so that a "segment"
      //  can be parsed from them? That could result in files being spuriously grouped
      //  together in the same arena. This would only practically be a problem if such a
      //  "segment" was never closed (refCt never 0), _and_ enough files associated with that
      //  "segment" were opened and closed (without ever closing the associated Arena)
      //  to exhaust virtual memory space.
```

ChrisHegarty:
Hi,   I don't have access to commit to this branch, so (sorry) I just created an alternative PR to sketch out an idea around a simplification of ref counting the arena. See #13570

magibney:
Closing in favor of #13570

PR #13556, opened 2024-07-10
Lookup next when current doc is deleted in PerThreadPKLookup.lookup

vsop-479:
### Description

In current implementation, we won't get the live doc, if we have deletes on unFlushed segment.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/index/TestTermsEnum.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestTermsEnum.java
@@ -30,6 +30,7 @@ import java.util.Set;
 import java.util.TreeSet;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field;
+import org.apache.lucene.document.KeywordField;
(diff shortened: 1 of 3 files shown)


lucene/core/src/test/org/apache/lucene/index/TestTermsEnum.java
```diff
@@ -998,6 +999,43 @@ public void testCommonPrefixTerms() throws Exception {
     d.close();
   }
 
+  public void testPKLookupWithUpdate() throws Exception {
```
jpountz, line 1002: Can you create a new `TestPerThreadPKLookup` under lucene/test-framework/src/test` and move this test there?

jpountz (commented):
Thanks for fixing this. I left a minor comment, otherwise LGTM.

lucene/core/src/test/org/apache/lucene/index/TestTermsEnum.java
vsop-479: Thanks @jpountz , I moved the test case.

PR #13560, opened 2024-07-10
[9.x] Use a confined Arena for IOContext.READONCE (#13535)

ChrisHegarty:
Backport of:
 * #13535

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/store/TestMMapDirectory.java
+++ b/lucene/core/src/test/org/apache/lucene/store/TestMMapDirectory.java
@@ -20,9 +20,14 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Objects;
 import java.util.Random;
+import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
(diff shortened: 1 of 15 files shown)


lucene/core/src/test/org/apache/lucene/store/TestMMapDirectory.java
```diff
@@ -141,4 +146,57 @@ public void testWithRandom() throws Exception {
       }
     }
   }
+
+  // Opens the input with IOContext.READONCE to ensure slice and clone are appropriately confined
+  public void testConfined() throws Exception {
+    assumeTrue("Only testable with memory segments", Runtime.version().feature() >= 19);
```
uschindler, line 152: Use `isMemorySegmentImpl()`, see https://github.com/apache/lucene/blob/9e5d278cd6e6f17f09dea8d05295008c855c971a/lucene/core/src/test/org/apache/lucene/store/TestMMapDirectory.java#L67

uschindler (changes requested):

lucene/core/src/test/org/apache/lucene/store/TestMMapDirectory.java
ChrisHegarty: ah ha! that is what I was looking for. Thanks!

uschindler (approved):

Issue #12627, opened 2023-10-06
HnwsGraph creates disconnected components


nitirajrathore:
### Description

I work for Amazon Retail Product search and we are using Lucene KNN for
semantic search of products. 
We recently noticed that the hnsw graphs generated are not always strongly
connected and in worst case scenario some products may be undiscoverable.
Connectedness of Hierarchical graph can be complicated, so below I am
mentioning my experiment details.

### Experiment:
I took the Lucene indexes from our production servers and for each segment
(hnsw graph) I did following test.
At each level graph I took the same entry point, the entry point of HNSW
graph, checked how many nodes are reachable from this entrypoint. Note that
connectedness at each level was checked independently of other levels.
Sample code attached. My observations are as below.

- Observation :
1. Of all the graphs across all the segments, across 100s of indexes that I
considered, one graph for each "level" of HNSW, almost 18% of the graphs
had some disconnectedness.
2. Disconnectedness was observed at all the levels of HNSW graphs. I saw we have
at most 3 to 4 levels in HNSW graphs.
3. percentage disconnectedness ranged from small fractions 0.000386% (1
disconnected out of 259342)  to 3.7% (eg. 87 disconnected out of 2308).
In some extreme case the entry-point(EP) in zeroth level graph was disconnected
from rest of the graph making the %age disconnected as high as 99.9% (only 65
reachable nodes from EP out of 252275). But this does not necessarily mean
that the 99.9% of nodes were not discoverable, it just means that if
unluckily we end up on EP in the 0th level graph for a query, there can at
max be 65 nodes that can be reached. But had we diverted our path from EP
to some other node in the upper level graphs then may be more nodes be
discoverable via that node.

### Reproduciability
   Although it is fairly easy to be reproduced with Amazon's internal data, but I found it really hard to reproduce it with random vectors. I will attached a simple main class of my test and the parameters that I tried. I only got some disconnectedness as listed in below table. For higher number of vectors, the execution becomes painfully slow because of hnsw graph creation. I will continue to work to find some  parameters for reproducibilty and also explore *open source datasets* for the same.

``DIS-L0 means: %age of disconnected nodes in graph at level-0 of HNSW graph.``

| randomness seed       | dimension | no of doc vectors      | Max-conn   | Beam-width | dis-l0 | dis-l1                      | dis-l2 | dis-l3 | dis-l4 |
| ---------- | --- | --------- | --- | --------- | ------- | ---------------------------- | ------- | ------- | ------- |
[...]

PR #13561, opened 2024-07-10
Add HnswGraphBuilder.getCompletedGraph() and record completed state

msokolov:
See https://github.com/apache/lucene/issues/12627 for context

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswConcurrentMergeBuilder.java
+++ b/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswConcurrentMergeBuilder.java
@@ -41,6 +41,7 @@ public class HnswConcurrentMergeBuilder implements HnswBuilder {
   private final TaskExecutor taskExecutor;
   private final ConcurrentMergeWorker[] workers;
   private InfoStream infoStream = InfoStream.getDefault();
+  private boolean frozen;
 
   public HnswConcurrentMergeBuilder(
       TaskExecutor taskExecutor,
@@ -69,6 +70,9 @@ public class HnswConcurrentMergeBuilder implements HnswBuilder {
 
   @Override
   public OnHeapHnswGraph build(int maxOrd) throws IOException {
+    if (frozen) {
+      throw new IllegalStateException("graph has already been built");
+    }
     if (infoStream.isEnabled(HNSW_COMPONENT)) {
       infoStream.message(
           HNSW_COMPONENT,
@@ -84,7 +88,8 @@ public class HnswConcurrentMergeBuilder implements HnswBuilder {
           });
     }
     taskExecutor.invokeAll(futures);
-    return workers[0].getGraph();
+    frozen = true;
+    return workers[0].getCompletedGraph();
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


lucene/core/src/java/org/apache/lucene/util/hnsw/HnswConcurrentMergeBuilder.java
```diff
@@ -84,7 +88,8 @@ public OnHeapHnswGraph build(int maxOrd) throws IOException {
           });
     }
     taskExecutor.invokeAll(futures);
-    return workers[0].getGraph();
+    frozen = true;
+    return workers[0].getCompletedGraph();
```
benwtrent, line 92: Why are we freezing here instead of within `getCompletedGraph` like its done for the single threaded builder?

msokolov: Yeah ... I guess this allows a little more strictness that we can't have in the other case because this builder *only* allows building via `build()` method wheras the other one can accept individual nodes.  But perhaps consistency is better and we should move this `frozen = true` into the `getCompletedGraph`

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
```diff
@@ -156,14 +157,20 @@ public OnHeapHnswGraph build(int maxOrd) throws IOException {
       infoStream.message(HNSW_COMPONENT, "build graph from " + maxOrd + " vectors");
     }
     addVectors(maxOrd);
-    return hnsw;
```
benwtrent, line 159: I think checking for `frozen` at the start of build is wise and prevents info stream writing when we are already frozen.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
```diff
@@ -156,14 +157,20 @@ public OnHeapHnswGraph build(int maxOrd) throws IOException {
...
 
+  @Override
+  public OnHeapHnswGraph getCompletedGraph() {
+    frozen = true;
+    return getGraph();
+  }
+
   @Override
   public OnHeapHnswGraph getGraph() {
     return hnsw;
```
benwtrent, line 176: We should check for frozen directly in `addVectors` to prevent infoStream writing if there is some sub-class calling this method erroneously

benwtrent (approved):
When to free the concurrent merge builder or not isn't a big issue with me. But if we are going to have different behaviors, they should be commented and called out so future readers know why.

I have some concerns to ensure frozen is checked before we write out infoStream messages. But nothing major.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswConcurrentMergeBuilder.java
msokolov: I added `frozen=true` to `getCompletedGraph` so it has the same semantics as `HnswGraphBuilder`

PR #13563, opened 2024-07-11
Add levels to DocValues skipper index

iverase:
Currently the DocValues skipper index collects the stats every 4096 documents that allow implementors to used them to decide if they want to process those documents or they can be skipped. The idea of adding levels is to be able to skip several of those block (called intervals in the code) in one step by collecting the stats of the blocks and adding them to the index.

This implementation collects the stats of 8 of those intervals.  For the first level, the first interval gets a new level with the stats from the next 8 intervals, then the 9th interval gets another level with the stats from the following 8 intervals and so on. For the second level, the first interval gets a new level with the stats from the next 8 level1  intervals, which is the same as the stats from the first 64 level 0 intervals and so on. 

I run some  basic experiments and I must say I did not see much change on the performance, still it feels the right thing to do, therefore I opened this PR.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
@@ -211,8 +211,8 @@ final class Lucene90DocValuesConsumer extends DocValuesConsumer {
     }
 
     void accumulate(SkipAccumulator other) {
-      minDocID = Math.min(minDocID, other.minDocID);
-      maxDocID = Math.max(maxDocID, other.maxDocID);
+      assert minDocID <= other.minDocID && maxDocID < other.maxDocID;
+      maxDocID = other.maxDocID;
       minValue = Math.min(minValue, other.minValue);
       maxValue = Math.max(maxValue, other.maxValue);
       docCount += other.docCount;
@@ -241,16 +241,13 @@ final class Lucene90DocValuesConsumer extends DocValuesConsumer {
     long globalMinValue = Long.MAX_VALUE;
     int globalDocCount = 0;
     int maxDocId = -1;
-    final List<List<SkipAccumulator>> accumulators = new ArrayList<>(SKIP_INDEX_MAX_LEVEL);
-    for (int i = 0; i < SKIP_INDEX_MAX_LEVEL; i++) {
-      accumulators.add(new ArrayList<>());
-    }
+    final List<SkipAccumulator> accumulators = new ArrayList<>();
     SkipAccumulator accumulator = null;
     final int maxAccumulators = 1 << (SKIP_INDEX_LEVEL_SHIFT * (SKIP_INDEX_MAX_LEVEL - 1));
     for (int doc = values.nextDoc(); doc != DocIdSetIterator.NO_MORE_DOCS; doc = values.nextDoc()) {
       if (accumulator == null) {
         accumulator = new SkipAccumulator(doc);
-        accumulators.get(0).add(accumulator);
+        accumulators.add(accumulator);
       }
       accumulator.nextDoc(doc);
       for (int i = 0, end = values.docValueCount(); i < end; ++i) {
@@ -264,9 +261,7 @@ final class Lucene90DocValuesConsumer extends DocValuesConsumer {
         accumulator = null;
         if (accumulators.size() == maxAccumulators) {
           writeLevels(accumulators);
-          for (List<SkipAccumulator> accumulatorList : accumulators) {
-            accumulatorList.clear();
-          }
+          accumulators.clear();
         }
       }
     }
@@ -290,40 +285,39 @@ final class Lucene90DocValuesConsumer extends DocValuesConsumer {
     meta.writeInt(maxDocId);
   }
 
-  private void writeLevels(List<List<SkipAccumulator>> accumulators) throws IOException {
-    for (int i = 1; i < accumulators.size(); i++) {
-      buildLevel(accumulators.get(i), accumulators.get(i - 1));
+  private void writeLevels(List<SkipAccumulator> accumulators) throws IOException {
+    final List<List<SkipAccumulator>> accumulatorsLevels = new ArrayList<>(SKIP_INDEX_MAX_LEVEL);
+    accumulatorsLevels.add(accumulators);
+    for (int i = 0; i < SKIP_INDEX_MAX_LEVEL - 1; i++) {
+      accumulatorsLevels.add(buildLevel(accumulatorsLevels.get(i)));
     }
-    int totalAccumulators = accumulators.get(0).size();
+    int totalAccumulators = accumulators.size();
     for (int index = 0; index < totalAccumulators; index++) {
       // compute how many levels we need to write for the current accumulator
       final int levels = getLevels(index, totalAccumulators);
-      // build the levels
-      final SkipAccumulator[] accLevels = new SkipAccumulator[levels];
-      for (int level = 0; level < levels; level++) {
-        accLevels[level] =
-            accumulators.get(level).get(index / (1 << (SKIP_INDEX_LEVEL_SHIFT * level)));
-      }
       // write the number of levels
       data.writeByte((byte) levels);
       // write intervals in reverse order. This is done so we don't
       // need to read all of them in case of slipping
       for (int level = levels - 1; level >= 0; level--) {
-        data.writeInt(accLevels[level].maxDocID);
-        data.writeInt(accLevels[level].minDocID);
-        data.writeLong(accLevels[level].maxValue);
-        data.writeLong(accLevels[level].minValue);
-        data.writeInt(accLevels[level].docCount);
+        final SkipAccumulator accumulator =
+            accumulatorsLevels.get(level).get(index >> (SKIP_INDEX_LEVEL_SHIFT * level));
+        data.writeInt(accumulator.maxDocID);
+        data.writeInt(accumulator.minDocID);
+        data.writeLong(accumulator.maxValue);
+        data.writeLong(accumulator.minValue);
+        data.writeInt(accumulator.docCount);
       }
     }
   }
 
-  private static void buildLevel(
-      List<SkipAccumulator> collector, List<SkipAccumulator> accumulators) {
-    int levelSize = 1 << SKIP_INDEX_LEVEL_SHIFT;
+  private static List<SkipAccumulator> buildLevel(List<SkipAccumulator> accumulators) {
+    final int levelSize = 1 << SKIP_INDEX_LEVEL_SHIFT;
+    final List<SkipAccumulator> collector = new ArrayList<>();
     for (int i = 0; i < accumulators.size() - levelSize + 1; i += levelSize) {
       collector.add(SkipAccumulator.merge(accumulators, i, levelSize));
     }
+    return collector;
   }
 
   private int getLevels(int index, int size) {
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
@@ -240,13 +240,16 @@ final class Lucene90DocValuesConsumer extends DocValuesConsumer {
     long globalMinValue = Long.MAX_VALUE;
     int globalDocCount = 0;
     int maxDocId = -1;
-    List<SkipAccumulator> accumulators = new ArrayList<>();
+    final List<List<SkipAccumulator>> accumulators = new ArrayList<>(SKIP_INDEX_MAX_LEVEL);
+    for (int i = 0; i < SKIP_INDEX_MAX_LEVEL; i++) {
+      accumulators.add(new ArrayList<>());
+    }
     SkipAccumulator accumulator = null;
     final int maxAccumulators = 1 << (SKIP_INDEX_LEVEL_SHIFT * (SKIP_INDEX_MAX_LEVEL - 1));
     for (int doc = values.nextDoc(); doc != DocIdSetIterator.NO_MORE_DOCS; doc = values.nextDoc()) {
       if (accumulator == null) {
         accumulator = new SkipAccumulator(doc);
-        accumulators.add(accumulator);
+        accumulators.get(0).add(accumulator);
       }
       accumulator.nextDoc(doc);
       for (int i = 0, end = values.docValueCount(); i < end; ++i) {
@@ -260,7 +263,9 @@ final class Lucene90DocValuesConsumer extends DocValuesConsumer {
         accumulator = null;
         if (accumulators.size() == maxAccumulators) {
           writeLevels(accumulators);
-          accumulators.clear();
+          for (List<SkipAccumulator> accumulatorList : accumulators) {
+            accumulatorList.clear();
+          }
         }
       }
     }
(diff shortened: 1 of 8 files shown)


lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
```diff
@@ -1792,61 +1794,88 @@ public DocValuesSkipper getSkipper(FieldInfo field) throws IOException {
...
+          // find next interval
+          assert target > maxDocID[0] : "target must be bigger that current interval";
           while (true) {
-            maxDocID = input.readInt();
-            if (maxDocID >= target) {
-              minDocID = input.readInt();
-              maxValue = input.readLong();
-              minValue = input.readLong();
-              docCount = input.readInt();
+            levels = input.readByte();
```
jpountz, line 1824: I'm a bit confused by this, because this `levels` variable feels like the number of lower levels that are getting updated, while some highel levels may still be valid? But then it would not be correct to return `levels` in `numLevels()` as we'd be missing these higher levels that are still valid?

E.g. if there are 8 intervals, there are two levels. But when we read the second interval, then `levels` is 1 because only the lower level needs updating, but there are still 2 valid levels?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
```diff
@@ -207,65 +210,120 @@ void accumulate(long value) {
...
+  private void writeLevels(List<SkipAccumulator> accumulators) throws IOException {
+    for (int index = 0; index < accumulators.size(); index++) {
+      // compute how many levels we need to write for the current accumulator
+      final int levels = getLevels(index, accumulators.size());
+      // build the levels
+      final SkipAccumulator[] accLevels = new SkipAccumulator[levels];
+      for (int level = 0; level < levels; level++) {
+        accLevels[level] =
+            SkipAccumulator.merge(
+                accumulators, index, SKIP_INDEX_NUMBER_INTERVALS_PER_LEVEL[level]);
```
jpountz, line 296: It would be a bit more efficient to compute the accumulators on a level based on the accumulator of the previous level rather than using accumulators at level 0 all the time?

jpountz (commented):
It makes sense to me overall.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
iverase: This is done in purpose as I am thinking in this data structure as a skip list, no a tree, so any given interval at any level should only be evaluated once and therefore only appear once while reading the index.

The idea is that users only see new levels the first time they get updated. At that point, user can decide if they want to skip it or not. If they don't skip it, then they go inside that level but they should not have to evaluate it again and therefore it should not appear in lower levels. 

I think otherwise, we will be evaluating the same interval many times?

jpountz: I see. I need to think more about it. It makes sense to me for top-level queries which would visit the full doc ID range anyway. But if the query is part of a conjunction, then the leading clause of the conjunction could advance this clause to an arbitrary doc in the doc ID space, and I wonder if we're losing potential efficiency by not making the higher levels visible.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
iverase: I was hoping this approach would be a bit more heap friendly as we will only have to create those accumulators for a short period of time and not having a list per level. 

But yes, it might make sense to build all the levels upfront using previous levels.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
iverase: mmm, advancing doc ids is different to exposing levels through the API. It is related to the implementation of `#advance(int docID)`. We are currently reading the next interval when we advance which can be inefficient on advancing to an arbitrary doc ID. 

We can probably do better there

jpountz: FWIW the call site I have in mind is this one: https://github.com/apache/lucene/blob/5e52b8094aca4dab2898c7dd17084fc7cf7fd016/lucene/core/src/java/org/apache/lucene/document/DocValuesRangeIterator.java#L89. How could it know that it should advance to an earlier doc ID if it wanted to have access to the higher levels?

iverase: I modified the implementation so we will have access to higher levels at any point. Let me know what you think.

jpountz (commented):
Thanks, it does indeed make more sense to me to expose all levels. I left a few comments.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
```diff
@@ -207,65 +210,133 @@ void accumulate(long value) {
...
-      if (++counter == skipIndexIntervalSize) {
+      if (accumulator.docCount == skipIndexIntervalSize) {
         globalMaxValue = Math.max(globalMaxValue, accumulator.maxValue);
         globalMinValue = Math.min(globalMinValue, accumulator.minValue);
         globalDocCount += accumulator.docCount;
         maxDocId = accumulator.maxDocID;
-        accumulator.writeTo(data);
-        counter = 0;
+        accumulator = null;
+        if (accumulators.size() == maxAccumulators) {
```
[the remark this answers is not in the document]
iverase: I removed the List of List so this is not an issue any more

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
```diff
@@ -207,65 +210,133 @@ void accumulate(long value) {
...
-    // TODO: This disk compression once we introduce levels
-    long start = data.getFilePointer();
-    SortedNumericDocValues values = valuesProducer.getSortedNumeric(field);
+    final long start = data.getFilePointer();
+    final SortedNumericDocValues values = valuesProducer.getSortedNumeric(field);
     long globalMaxValue = Long.MIN_VALUE;
     long globalMinValue = Long.MAX_VALUE;
     int globalDocCount = 0;
     int maxDocId = -1;
+    final List<List<SkipAccumulator>> accumulators = new ArrayList<>(SKIP_INDEX_MAX_LEVEL);
```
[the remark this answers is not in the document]
iverase: yes, this list of list is awful. I was thinking in reusing ti but it creates more issues that it helps.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
```diff
@@ -207,65 +210,133 @@ void accumulate(long value) {
...
+    }
+    int totalAccumulators = accumulators.get(0).size();
+    for (int index = 0; index < totalAccumulators; index++) {
+      // compute how many levels we need to write for the current accumulator
+      final int levels = getLevels(index, totalAccumulators);
+      // build the levels
+      final SkipAccumulator[] accLevels = new SkipAccumulator[levels];
+      for (int level = 0; level < levels; level++) {
+        accLevels[level] =
+            accumulators.get(level).get(index / (1 << (SKIP_INDEX_LEVEL_SHIFT * level)));
```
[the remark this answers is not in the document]
iverase: done

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
```diff
@@ -207,65 +210,133 @@ void accumulate(long value) {
...
+  private void writeLevels(List<List<SkipAccumulator>> accumulators) throws IOException {
+    for (int i = 1; i < accumulators.size(); i++) {
+      buildLevel(accumulators.get(i), accumulators.get(i - 1));
+    }
+    int totalAccumulators = accumulators.get(0).size();
+    for (int index = 0; index < totalAccumulators; index++) {
+      // compute how many levels we need to write for the current accumulator
+      final int levels = getLevels(index, totalAccumulators);
+      // build the levels
+      final SkipAccumulator[] accLevels = new SkipAccumulator[levels];
```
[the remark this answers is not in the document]
iverase: yes, I did not realize when I modified the way we were writing the levels

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
```diff
@@ -1792,61 +1794,91 @@ public DocValuesSkipper getSkipper(FieldInfo field) throws IOException {
...
+              }
+              minDocID[level] = input.readInt();
+              maxValue[level] = input.readLong();
+              minValue[level] = input.readLong();
+              docCount[level] = input.readInt();
+            }
+            if (competitive) {
+              // adjust levels
+              while (levels < SKIP_INDEX_MAX_LEVEL) {
+                if (maxDocID[levels] == -1 || maxDocID[levels] < target) {
```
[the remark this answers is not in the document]
iverase: done

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -3301,17 +3301,17 @@ private static void checkDocValueSkipper(FieldInfo fi, DocValuesSkipper skipper)
...
       for (int level = 0; level < levels; level++) {
-        if (skipper.minDocID(level) < doc) {
-          throw new CheckIndexException(
-              "skipper dv iterator for field: "
-                  + fieldName
-                  + " reports wrong minDocID, got "
-                  + skipper.minDocID(level)
-                  + " < "
-                  + doc);
-        }
```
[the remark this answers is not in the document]
iverase: The current condition is `skipper.minDocID(level) < doc` where the doc is the provided doc on advance. So the provided doc needs to be lower than or equal to the minDocID not to trigger the error.  I

It checks we have advance to the closer interval forward. This check is strange as it is only valid for implementations with exact bounds and it is clearly not valid for levels above 0.

lucene/test-framework/src/java/org/apache/lucene/tests/index/AssertingLeafReader.java
```diff
@@ -1194,24 +1194,27 @@ public int numLevels() {
...
-      Objects.checkIndex(level, numLevels());
       int minDocID = in.minDocID(level);
       assert minDocID <= in.maxDocID(level);
-      if (level > 0) {
-        assert minDocID <= in.minDocID(level - 1);
+      if (minDocID != -1 && minDocID != DocIdSetIterator.NO_MORE_DOCS) {
+        Objects.checkIndex(level, numLevels());
+      }
+      for (int i = 0; i < level; i++) {
+        assert minDocID >= in.minDocID(i);
```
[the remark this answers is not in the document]
jpountz: Could we make `numLevels` return 1 to avoid this issue?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
```diff
@@ -207,65 +210,127 @@ void accumulate(long value) {
...
+
+  private int getLevels(int index, int size) {
+    final int left = size - index;
+    for (int level = SKIP_INDEX_MAX_LEVEL - 1; level > 0; level--) {
+      final int numberIntervals = 1 << (SKIP_INDEX_LEVEL_SHIFT * level);
+      if (left >= numberIntervals && index % numberIntervals == 0) {
+        return level + 1;
+      }
+    }
+    return 1;
```
jpountz, line 331: There must be a way to do this in constant time rather than linearly with SKIP_INDEX_MAX_LEVEL?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
```diff
@@ -1792,61 +1794,91 @@ public DocValuesSkipper getSkipper(FieldInfo field) throws IOException {
     if (input.length() > 0) {
       input.prefetch(0, 1);
     }
+    // TODO: should we write to disk the actual max level for this segment?
```
jpountz, line 1797: This, or we can compute it based on maxDoc?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
```diff
@@ -1792,61 +1794,91 @@ public DocValuesSkipper getSkipper(FieldInfo field) throws IOException {
...
-            if (maxDocID >= target) {
-              minDocID = input.readInt();
-              maxValue = input.readLong();
-              minValue = input.readLong();
-              docCount = input.readInt();
+            levels = input.readByte();
+            assert levels <= SKIP_INDEX_MAX_LEVEL && levels > 0
+                : "level out of range [" + levels + "]";
+            boolean competitive = true;
+            // check if current interval is competitive or we can jump to the next position
```
jpountz, line 1828: I got a bit confused at first by the use of `competitive`. Maybe `valid`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
```diff
@@ -1792,61 +1794,91 @@ public DocValuesSkipper getSkipper(FieldInfo field) throws IOException {
...
+              docCount[level] = input.readInt();
+            }
+            if (competitive) {
+              // adjust levels
+              while (levels < SKIP_INDEX_MAX_LEVEL) {
+                if (maxDocID[levels] < target) {
+                  break;
+                }
+                levels++;
+              }
```
jpountz, line 1847: Let's combine both conditions under the while condition?

```java
while (levels < SKIP_INDEX_MAX_LEVEL && maxDocIDS[levels] < target) {
  levels++;
}
```

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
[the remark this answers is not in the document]
jpountz: This makes sense.

jpountz (approved):

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
iverase: of course, the condition looks more like: 
```java
           while (levels < SKIP_INDEX_MAX_LEVEL && maxDocID[levels] >= target) {
                levels++;
              }
```

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
iverase: I have plans to allow some intervals to have different number of documents depending on the unique values in the interval so we would not be able to compute the number of levels by looking at the number of docs.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
```diff
@@ -207,65 +210,127 @@ void accumulate(long value) {
...
-      if (++counter == skipIndexIntervalSize) {
+      if (accumulator.docCount == skipIndexIntervalSize) {
         globalMaxValue = Math.max(globalMaxValue, accumulator.maxValue);
         globalMinValue = Math.min(globalMinValue, accumulator.minValue);
         globalDocCount += accumulator.docCount;
         maxDocId = accumulator.maxDocID;
-        accumulator.writeTo(data);
-        counter = 0;
+        accumulator = null;
+        if (accumulators.size() == maxAccumulators) {
```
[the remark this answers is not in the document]
iverase: yes, but probably not frequent enough so I change the range interval for the test to use smaller values but still keeping a reasonable performance on the test.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesConsumer.java
iverase: I haven't been able to find a more elegant solution yet!

PR #13566, opened 2024-07-11
gh-12627: HnswGraphBuilder connects disconnected HNSW graph components

msokolov:
This looks pretty good to me, as far as it goes, but I wanted to post early so people can poke holes / suggest improvements. Remaining TODO's I'm tracking are:
* testing recall (might improve a bit)
* also apply to concurrent graph build
* verify we see fewer disconnected graphs when testing with real data
* think if there is a way to *guarantee* connectedness - this is best-effort only because of max-connections limitation


msokolov:
... oh another TODO is whether or not to worry about graph levels > 0. I kind of think it's not needed since no matter which graph component you traverse on the higher level, you eventually end up in level0 and if that is fully connected you should be able to find your way to the right place? OTOH there could be some speed/recall benefit to ensuring full-connection for all the levels

msokolov:
note: addresses https://github.com/apache/lucene/issues/12627

msokolov:
I tested using KnnGraphTester and in the process changed this to handle multiple levels (it just seemed to make sense form a consistency perspective). I also found there were a couple of bugs:
* the `M`/max-connections check was off-by-one
* added a call to `finish()` so it would actually do something!

Testing w/1M 256-dim documents I got results that are within noise for recall, latency, and index time:

|test|recall|latency|index time|
|-----|-----|-----|-----|
| baseline| 0.927 | 2.32 | 164755 |
| baseline| 0.926 | 2.26 | 167982 |
| candidate| 0.924 | 2.16 | 159109 |
| candidate| 0.927 | 2.25 | 169839 |

I printed out the graph connectivity and observed that there were many singleton nodes produced (nodes not connected to any other node) in the baseline case and these were largely eliminated in the candidate, although I did still see some on level 1 on one of my tests in some segments.

I also printed out the time taken to do the connection - it was about 3s in a 3m indexing run. Overall I think this is worth doing in order to improve sanity.

It would still be better if we could *guarantee* that the graphs are fully connected. It is challenging though since we assume nodes will also comply with max-connection limit and in theory every node in a component could be maximally connected which would not allow for any further connection, so I think I will put this off for some future and opt for P not P. 

Still need to address the concurrent graph builder.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
```diff
@@ -408,7 +410,23 @@ private void finish() throws IOException {
   }
 
   private void connectComponents() throws IOException {
-    List<Component> components = HnswUtil.components(hnsw);
+    long start = System.nanoTime();
+    for (int level = 0; level < hnsw.numLevels(); level++) {
+      if (connectComponents(level) == false) {
```
benwtrent, line 415: I like the logging, we should do this for sure. But I also think this should be an `assert` as it should never happen in tests.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswUtil.java
```diff
@@ -30,46 +30,83 @@
...
+   * Returns the sizes of the distinct graph components on the given level. If the graph is
+   * fully-connected there will only be a single component. If the graph is empty, the returned list
+   * will be empty.
+   */
+  public static List<Integer> componentSizes(HnswGraph hnsw, int level) throws IOException {
+    return components(hnsw, level).stream().map(Component::size).toList();
+  }
+
+  // Finds all connected components of the graph
+  static List<Component> components(HnswGraph hnsw, int level) throws IOException {
```
benwtrent, line 68: Do you mind providing a bit more color to this java doc. It took me a minute to understand that a component is really "A selection of nodes that are traversable to each other".

benwtrent (commented):
I will benchmark as well with some other data sets & settings. 

Getting this ready to go for the concurrent merger would be nice as I have noticed that graphs can get differently connected with it vs. the incremental merger.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
msokolov: I don't think we can make this an `assert` without some more major change for the reason (mentioned above) that one of the components may already be maximally connected (ie its nodes all have the max number of neighbors).

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswUtil.java
msokolov: sure - I'll explain the terminology -- it comes from graph theory https://en.wikipedia.org/wiki/Component_(graph_theory)

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
msokolov: A degenerate case is `M=1` in this case every component will have 2 nodes connected to each other and we cannot connect them further, but this will arise for M>1 too. We can try to use pruning again, but then we run the risk of orphaning a new component and would have to iterate and as far as I can tell there is no guarantee we would converge even for higher M. I think in order to guarantee connectedness we have to relax the number of connections limit, but we assume elsewhere (not only in NeighborArray) that this limit is obeyed (eg when searching)

benwtrent: Yeah, ok, the degenerate case is indeed problematic. I think it is OK to make this best effort and log. 

If there are still many disconnected components, the better way is for the number of connections to be bumped higher.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswUtil.java
msokolov: After reading that I realized that this whole notion of components really is only well-defined for undirected graphs. Ours are directed, so this isn't really the same thing after all. Actually it raises a question about this whole business because we can make it so every node is reachable from node 0, but that still doesn't guarantee that every node is reachable from every other node. I still think this is progress, but we might want to think about how to measure this more global property?

msokolov: https://en.wikipedia.org/wiki/Directed_graph#Directed_graph_connectivity defines the "strongly connected" property, which I think is what we want

msokolov: OK, I have an idea how we can maybe compute strongly-connected efficiently b/c our graphs are "mostly" undirected -- I think we can check back-links while walking the graph? I'll noodle on it and see if I can come up with a stronger test

msokolov:
This version also applies the connectedness-checking and patching to concurrent graph build. It was a trivial addition. I tested and it seems to maybe help a bit:

|test       | recall | latency | index time |
|-----------|--------|---------|------------|
| baseline  | 0.925  | 2.29    | 172225     |
| baseline  | 0.925  | 2.39    | 168809     |
| candidate | 0.928  | 2.27    | 163188     |
| candidate | 0.928  | 2.25    | 168005     |

I still plan to try applying the strongly-connected graph check, so might want to hold off further testing until we have that. In theory it could suggest additional connections to add to the graph, although I suspect in practice its impact will be limited since what we seem to see is one big component (probably strongly connected?) plus a bunch of single disconnected nodes, but we'll see.

benwtrent:
I have benchmarked 2 data sets with 2 scenarios (int4 and int7 quantization) and have found no significant difference in runtime between this branch and the main branch. 

|                | e5Small build | e5Small recall | e5small vectors visited | CohereV2 Build | CohereV2 recall | CohereV2 vectors visited |
|----------------|---------------|----------------|-------------------------|----------------|-----------------|--------------------------|
| candidate int7 | 582387        | 0.974          | 3090                    | 1360523        | 0.825           | 4219                     |
| baseline int7  | 574839        | 0.974          | 3090                    | 1360658        | 0.825           | 4218                     |
| candidate int4 | 571491        | 0.866          | 3118                    | 1367738        | 0.516           | 4513                     |
| baseline int4  | 586543        | 0.866          | 3118                    | 1381992        | 0.516           | 4512                     |


We should test a previously bad case to see how much longer indexing takes & how much recall is improved.

msokolov:
To make things look bad, I think we need to reduce M and/or beamWidth? In my test I saw some impact with M=16 and beamWidth=50

msokolov:
I tried implementing a "strongly-connected" version of this thing. My implementation is super slow (makes indexing take way longer) and the result in terms of search metrics was about the same. I might see about using another implementation, maybe this one? https://www.geeksforgeeks.org/tarjan-algorithm-find-strongly-connected-components/ but it may not be worth it given this whole thing is best-effort anyway

msokolov:
I plan to  revisit this with a modified approach to address some gaps here:

1. Instead of computing the components rooted at node 0 and others, or trying to compute strongly-connected components, we should compute components rooted at the graph entry points (which are the nodes of the next higher level). In the end we want to ensure every node is reachable by searching: it should be enough to ensure that every node is reachable from at least one entry point.
2. While traversing the graph we can keep track of nodes that are not full and then when connecting isolated nodes, we can limit the search to those nodes. This should help ensure we always make connections when possible.

msokolov:
I posted an updated patch that uses an approach that is more aligned with how we search. The idea is we don't need to care about strongly-connected graphs; we just want to make sure there is a path to every document via search. This means every document can be reached from the entry point, which is a single node on the top level. So this patch tries to ensure that every node on each level is reachable from one of the entry points of that level (which are just the nodes on the next higher level).  This is easy to compute and seems to correspond to the property we would like to guarantee: no orphaned nodes.

The patch also has a stronger approach to linking up disconnected components. Previously we would check the closest N nodes and if none of them could be linked (ie they were all full), we would give up. This patch tracks all the non-full nodes and uses that bitset as a filter on the search, so if there are any, it will find them. It also does not insist on bidirectional linking, which isn't necessary to get the property of reachability.

The results do show that this approach does a good (better than before) job of linking about disconnected components. You have to look hard to see a measurable impact on recall. Latency and indexing time don't seem to be affected 

In this test I measured with low values of M in order to get disconnected graphs. This test used  1M 256-dim vectors from an AMZ internal data set. I forced-merge but did not use concurrent merging (an oversight, it should work). A bunch of other tests showed improvements in connectedness, but not any significant change in other metrics.  Connectedness reported here is what KnnGraphTester prints - it's basically % of nodes reachable from a single node (the entryPoint node) on each level. Not exactly the same as the criteria being optimized, but it is closely correlated.

||     test || connectedness          || recall || latency || index time || M || beam ||
candidate   |  1 0.98 0.98 0.97 0.96 1 |  0.860 |     0.49 |    215111   |  8 |    50 |
candidate   |  1 1    0.98 0.97 0 .96  1 |  0.862 |     0.49 |    205430   |  8 |    50 |
baseline    |  1 1 0.97 0.93 0.88 0.97  |  0.858 |     0.49 |    203067   |  8 |    50 |
baseline    |  1 1 0.97 0.93 0.88 0.97  |  0.854 |     0.47 |    205423   |  8 |    50 |

benwtrent:
>  So this patch tries to ensure that every node on each level is reachable from one of the entry points of that level (which are just the nodes on the next higher level). This is easy to compute and seems to correspond to the property we would like to guarantee: no orphaned nodes.

This is a very nice exploitation of HNSW. One concern I have is local minima. Since we only ever use a single entry point, might be that we hit a region that is actually disconnected from the true nearest neighbors. So, no matter what we do (no matter how much we increase efSearch), we will never get the true nearest neighbors.

What do we think of increasing the number of entry points to the lower level?

> The results do show that this approach does a good (better than before) job of linking about disconnected components. You have to look hard to see a measurable impact on recall. Latency and indexing time don't seem to be affected

I would imagine recall is helped given abusive edge cases.

msokolov:
> What do we think of increasing the number of entry points to the lower level?

My first attempt searched for 2 instead of 1, and reconnected both of them, but doing this led to complications because nodes that were previously notFullyConnected may become fully connected while making the new connections. Although I suppose we could simply check that for that and skip them if it's the case.  I'll try that, it should be simple enough.

msokolov:
>>    What do we think of increasing the number of entry points to the lower level?
> My first attempt searched for 2 instead of 1, and reconnected both of them ...

Oh wait I misunderstood. You are talking about verifying that every node is reachable from multiple entry points? I was mentioning that we could easily search for more potential connections, which is different.  I think what you're asking (if I understood it) is trickier.  I guess we'd need to track a count per doc rather than a bit per doc? Or some other algorithm I haven't thought of

msokolov:
> What do we think of increasing the number of entry points to the lower level?

or maybe you're suggesting we change the way we search, not specifically something about this patch?  That does kind of make sense to me - I'm not sure why the HNSW paper settled on a single node at each level

benwtrent:
> or maybe you're suggesting we change the way we search, not specifically something about this patch? That does kind of make sense to me - I'm not sure why the HNSW paper settled on a single node at each level

^ This is what I am thinking. maybe select a logarithmically scaled number entry points when compared to the number of nearest neighbors that are being searched? 

This might mean its usually just 1, but maybe this is OK?

msokolov:
I ran one more test using Cohere (768 dim) vectors.  Looks OK to me, I'll merge this; at least it does no harm.

test case | recall | t    | M | beam | index time | connectivity
----------|--------|------|---|------|------------|-----------------------------
baseline  | 0.589  | 0.91 | 8 | 50   | 404738     | 1/1/0.99/0.98/0.93/0.87/0.96
baseline  | 0.589  | 0.96 | 8 | 50   | 416311     | 1/1/1/0.98/0.93/0.87/0.96
candidate | 0.591  | 0.92 | 8 | 50   | 411824     | 1/1/1/0.99/0.97/0.94/1
candidate | 0.588  | 0.93 | 8 | 50   | 406238     | 1/1/1/1/0.96/0.93/1

PR #13570, opened 2024-07-14
Aggregate files from the same segment into a single Arena

ChrisHegarty:
This commit adds a ref counted shared arena to support aggregating segment files into a single Arena.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexFileNames.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexFileNames.java
@@ -16,6 +16,7 @@
  */
 package org.apache.lucene.index;
 
+import java.util.Optional;
 import java.util.regex.Pattern;
 import org.apache.lucene.codecs.Codec;
 
@@ -140,6 +141,26 @@ public final class IndexFileNames {
       filename = filename.substring(idx);
     }
     return filename;
+  }
+
+  /**
+   * Parses the segment identifier out of the given segment name.
+   *
+   * @return an optional containing the segment identifier, or an empty optional if the segment name
+   *     does not start with '_' and parse to a valid segment identifier
+   */
+  @SuppressWarnings("unused")
+  public static Optional<String> parseSegmentIdentifier(String segmentName) {
+    if (segmentName.length() > 1 && segmentName.charAt(0) == '_') {
+      String segmentId = segmentName.substring(1);
+      try {
+        Long.parseLong(segmentId, Character.MAX_RADIX);
+        return Optional.of(segmentId);
+      } catch (NumberFormatException unused) {
+        // unable to parse as a segment identifier
+      }
+    }
+    return Optional.empty();
   }
 
   /** Returns the generation from this file name, or 0 if there is no generation. */
--- a/lucene/core/src/java/org/apache/lucene/index/IndexFileNames.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexFileNames.java
@@ -16,7 +16,6 @@
  */
 package org.apache.lucene.index;
 
-import java.util.Optional;
 import java.util.regex.Pattern;
 import org.apache.lucene.codecs.Codec;
 
@@ -141,26 +140,6 @@ public final class IndexFileNames {
       filename = filename.substring(idx);
     }
     return filename;
-  }
-
-  /**
-   * Parses the segment identifier out of the given segment name.
-   *
-   * @return an optional containing the segment identifier, or an empty optional if the segment name
-   *     does not start with '_' and parse to a valid segment identifier
-   */
-  @SuppressWarnings("unused")
-  public static Optional<String> parseSegmentIdentifier(String segmentName) {
-    if (segmentName.length() > 1 && segmentName.charAt(0) == '_') {
-      String segmentId = segmentName.substring(1);
-      try {
-        Long.parseLong(segmentId, Character.MAX_RADIX);
-        return Optional.of(segmentId);
-      } catch (NumberFormatException unused) {
-        // unable to parse as a segment identifier
-      }
-    }
-    return Optional.empty();
   }
 
   /** Returns the generation from this file name, or 0 if there is no generation. */
--- a/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
@@ -22,6 +22,7 @@ import java.lang.invoke.MethodType;
 import java.nio.channels.ClosedChannelException; // javadoc @link
 import java.nio.file.Path;
 import java.util.Locale;
+import java.util.Optional;
 import java.util.concurrent.Future;
 import java.util.function.BiPredicate;
 import org.apache.lucene.util.Constants;
@@ -101,6 +102,9 @@ public class MMapDirectory extends FSDirectory {
    * </ul>
    */
   public static final long DEFAULT_MAX_CHUNK_SIZE;
+
+  /** A provider specific context object or null, that will be passed to openInput. */
+  private final Object attachment = PROVIDER.attachment().orElse(null);
 
   final int chunkSizePower;
 
@@ -199,19 +203,26 @@ public class MMapDirectory extends FSDirectory {
     ensureOpen();
     ensureCanRead(name);
     Path path = directory.resolve(name);
-    return PROVIDER.openInput(path, context, chunkSizePower, preload.test(name, context));
+    return PROVIDER.openInput(
+        path, context, chunkSizePower, preload.test(name, context), attachment);
   }
 
   // visible for tests:
-  static final MMapIndexInputProvider PROVIDER;
+  static final MMapIndexInputProvider<Object> PROVIDER;
 
-  interface MMapIndexInputProvider {
-    IndexInput openInput(Path path, IOContext context, int chunkSizePower, boolean preload)
+  interface MMapIndexInputProvider<A> {
+    IndexInput openInput(
+        Path path, IOContext context, int chunkSizePower, boolean preload, A attachment)
         throws IOException;
 
     long getDefaultMaxChunkSize();
 
     boolean supportsMadvise();
+
+    /** An optional attachment of the provider, that will be passed to openInput. */
+    default Optional<A> attachment() {
+      return Optional.empty();
+    }
 
     default IOException convertMapFailedIOException(
         IOException ioe, String resourceDescription, long bufSize) {
@@ -256,7 +267,7 @@ public class MMapDirectory extends FSDirectory {
     }
   }
 
-  private static MMapIndexInputProvider lookupProvider() {
+  private static <A> MMapIndexInputProvider<A> lookupProvider() {
     final var lookup = MethodHandles.lookup();
     try {
       final var cls = lookup.findClass("org.apache.lucene.store.MemorySegmentIndexInputProvider");
@@ -264,7 +275,7 @@ public class MMapDirectory extends FSDirectory {
       // access through the lookup:
       final var constr = lookup.findConstructor(cls, MethodType.methodType(void.class));
       try {
-        return (MMapIndexInputProvider) constr.invoke();
+        return (MMapIndexInputProvider<A>) constr.invoke();
       } catch (RuntimeException | Error e) {
         throw e;
       } catch (Throwable th) {
--- a/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
@@ -22,8 +22,11 @@ import java.lang.invoke.MethodType;
 import java.nio.channels.ClosedChannelException; // javadoc @link
 import java.nio.file.Path;
 import java.util.Locale;
+import java.util.Optional;
 import java.util.concurrent.Future;
 import java.util.function.BiPredicate;
+import java.util.function.Function;
+import org.apache.lucene.index.IndexFileNames;
 import org.apache.lucene.util.Constants;
 
 /**
@@ -83,6 +86,19 @@ public class MMapDirectory extends FSDirectory {
    */
   public static final BiPredicate<String, IOContext> NO_FILES = (filename, context) -> false;
 
+  /** Argument for {@link #setGroupingFunction(Function)} that configures no grouping. */
+  public static final Function<String, Optional<String>> NO_GROUPING = filename -> Optional.empty();
+
+  /** Argument for {@link #setGroupingFunction(Function)} that configures grouping by segment. */
+  public static final Function<String, Optional<String>> GROUP_BY_SEGMENT =
+      filename -> {
+        String segmentName = IndexFileNames.parseSegmentName(filename);
+        if (filename.length() == segmentName.length()) {
+          return Optional.empty();
+        }
+        return Optional.of(segmentName);
+      };
+
   /**
    * Argument for {@link #setPreload(BiPredicate)} that configures files to be preloaded upon
    * opening them if they use the {@link ReadAdvice#RANDOM_PRELOAD} advice.
@@ -104,6 +120,8 @@ public class MMapDirectory extends FSDirectory {
 
   /** A provider specific context object or null, that will be passed to openInput. */
   final Object attachment = PROVIDER.attachment();
+
+  Function<String, Optional<String>> groupingFunction;
 
   final int chunkSizePower;
 
@@ -170,6 +188,7 @@ public class MMapDirectory extends FSDirectory {
       throw new IllegalArgumentException("Maximum chunk size for mmap must be >0");
     }
     this.chunkSizePower = Long.SIZE - 1 - Long.numberOfLeadingZeros(maxChunkSize);
+    this.groupingFunction = GROUP_BY_SEGMENT;
     assert (1L << chunkSizePower) <= maxChunkSize;
     assert (1L << chunkSizePower) > (maxChunkSize / 2);
   }
@@ -188,6 +207,21 @@ public class MMapDirectory extends FSDirectory {
   }
 
   /**
+   * Configures a grouping function for files that are part of the same logical group. The gathering
+   * of files into a logical group is a hint that allows for better handling of resources.
+   *
+   * <p>By default, grouping is {@link #GROUP_BY_SEGMENT}. To disable, invoke this method with
+   * {@link #NO_GROUPING}.
+   *
+   * @param groupingFunction a function that accepts a file name and returns an optional group key.
+   *     If the optional is present, then its value is the logical group to which the file
+   *     belongs. Otherwise, the file name if not associated with any logical group.
+   */
+  public void setGroupingFunction(Function<String, Optional<String>> groupingFunction) {
+    this.groupingFunction = groupingFunction;
+  }
+
+  /**
    * Returns the current mmap chunk size.
    *
    * @see #MMapDirectory(Path, LockFactory, long)
@@ -203,7 +237,12 @@ public class MMapDirectory extends FSDirectory {
     ensureCanRead(name);
     Path path = directory.resolve(name);
     return PROVIDER.openInput(
-        path, context, chunkSizePower, preload.test(name, context), attachment);
+        path,
+        context,
+        chunkSizePower,
+        preload.test(name, context),
+        groupingFunction.apply(name),
+        attachment);
   }
 
   // visible for tests:
@@ -211,7 +250,12 @@ public class MMapDirectory extends FSDirectory {
 
   interface MMapIndexInputProvider<A> {
     IndexInput openInput(
-        Path path, IOContext context, int chunkSizePower, boolean preload, A attachment)
+        Path path,
+        IOContext context,
+        int chunkSizePower,
+        boolean preload,
+        Optional<String> group,
+        A attachment)
         throws IOException;
 
     long getDefaultMaxChunkSize();
--- a/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
@@ -16,6 +16,8 @@
  */
 package org.apache.lucene.store;
 
+import static org.apache.lucene.index.IndexFileNames.CODEC_FILE_PATTERN;
+
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
@@ -97,11 +99,12 @@ public class MMapDirectory extends FSDirectory {
   /** Argument for {@link #setGroupingFunction(Function)} that configures grouping by segment. */
   public static final Function<String, Optional<String>> GROUP_BY_SEGMENT =
       filename -> {
-        String segmentName = IndexFileNames.parseSegmentName(filename);
-        if (filename.length() == segmentName.length()) {
+        if (!CODEC_FILE_PATTERN.matcher(filename).matches()) {
           return Optional.empty();
         }
-        return IndexFileNames.parseSegmentIdentifier(segmentName);
+        String segmentName = IndexFileNames.parseSegmentName(filename).substring(1);
+        long gen = IndexFileNames.parseGeneration(filename);
+        return Optional.of(segmentName + "-" + gen);
       };
 
   /**
--- a/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
@@ -103,7 +103,7 @@ public class MMapDirectory extends FSDirectory {
    *
    * @lucene.internal
    */
-  static final String SHARED_ARENA_MAX_PERMITS_SYSPROP =
+  public static final String SHARED_ARENA_MAX_PERMITS_SYSPROP =
       "org.apache.lucene.store.MMapDirectory.sharedArenaMaxPermits";
 
   /** Argument for {@link #setGroupingFunction(Function)} that configures no grouping. */
@@ -117,7 +117,10 @@ public class MMapDirectory extends FSDirectory {
         }
         String groupKey = IndexFileNames.parseSegmentName(filename).substring(1);
         try {
-          groupKey += IndexFileNames.parseGeneration(filename) == 0 ? "" : "-g";
+          // keep the original generation (=0) in base group, later generations in extra group
+          if (IndexFileNames.parseGeneration(filename) > 0) {
+            groupKey += "-g";
+          }
         } catch (
             @SuppressWarnings("unused")
             NumberFormatException unused) {
@@ -336,18 +339,21 @@ public class MMapDirectory extends FSDirectory {
     }
   }
 
-  private static Optional<Integer> getSharedArenaMaxPermitsSysprop() {
+  private static int getSharedArenaMaxPermitsSysprop() {
+    int ret = 1024; // default value
     try {
-      return Optional.ofNullable(System.getProperty(SHARED_ARENA_MAX_PERMITS_SYSPROP))
-          .map(Integer::valueOf);
+      String str = System.getProperty(SHARED_ARENA_MAX_PERMITS_SYSPROP);
+      if (str != null) {
+        ret = Integer.parseInt(str);
+      }
     } catch (@SuppressWarnings("unused") NumberFormatException | SecurityException ignored) {
       Logger.getLogger(MMapDirectory.class.getName())
           .warning(
               "Cannot read sysprop "
                   + SHARED_ARENA_MAX_PERMITS_SYSPROP
                   + ", so the default value will be used.");
-      return Optional.empty();
     }
+    return ret;
   }
 
   private static <A> MMapIndexInputProvider<A> lookupProvider() {
@@ -357,8 +363,7 @@ public class MMapDirectory extends FSDirectory {
       final var cls = lookup.findClass("org.apache.lucene.store.MemorySegmentIndexInputProvider");
       // we use method handles, so we do not need to deal with setAccessible as we have private
       // access through the lookup:
-      final var constr =
-          lookup.findConstructor(cls, MethodType.methodType(void.class, Optional.class));
+      final var constr = lookup.findConstructor(cls, MethodType.methodType(void.class, int.class));
       try {
         return (MMapIndexInputProvider<A>) constr.invoke(maxPermits);
       } catch (RuntimeException | Error e) {
--- a/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
@@ -102,9 +102,13 @@ public class MMapDirectory extends FSDirectory {
         if (!CODEC_FILE_PATTERN.matcher(filename).matches()) {
           return Optional.empty();
         }
-        String segmentName = IndexFileNames.parseSegmentName(filename).substring(1);
-        long gen = IndexFileNames.parseGeneration(filename);
-        return Optional.of(segmentName + "-" + gen);
+        String groupKey = IndexFileNames.parseSegmentName(filename).substring(1);
+        try {
+          groupKey += "-" + IndexFileNames.parseGeneration(filename);
+        } catch (NumberFormatException unused) {
+          // does not confirm to the generation syntax, or trash
+        }
+        return Optional.of(groupKey);
       };
 
   /**
--- a/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
@@ -28,6 +28,7 @@ import java.util.Optional;
 import java.util.concurrent.Future;
 import java.util.function.BiPredicate;
 import java.util.function.Function;
+import java.util.logging.Logger;
 import org.apache.lucene.index.IndexFileNames;
 import org.apache.lucene.util.Constants;
 
@@ -93,6 +94,18 @@ public class MMapDirectory extends FSDirectory {
    */
   public static final BiPredicate<String, IOContext> NO_FILES = (filename, context) -> false;
 
+  /**
+   * This sysprop allows to control the total maximum number of mmapped files that can be associated
+   * with a single shared {@link java.lang.foreign.Arena foreign Arena}. For example, to set the max
+   * number of permits to 256, pass the following on the command line pass {@code
+   * -Dorg.apache.lucene.store.MMapDirectory.sharedArenaMaxPermits=256}. Setting a value of 1
+   * associates one file to one shared arena.
+   *
+   * @lucene.internal
+   */
+  static final String SHARED_ARENA_MAX_PERMITS_SYSPROP =
+      "org.apache.lucene.store.MMapDirectory.sharedArenaMaxPermits";
+
   /** Argument for {@link #setGroupingFunction(Function)} that configures no grouping. */
   public static final Function<String, Optional<String>> NO_GROUPING = filename -> Optional.empty();
 
@@ -104,7 +117,7 @@ public class MMapDirectory extends FSDirectory {
         }
         String groupKey = IndexFileNames.parseSegmentName(filename).substring(1);
         try {
-          groupKey += "-" + IndexFileNames.parseGeneration(filename);
+          groupKey += IndexFileNames.parseGeneration(filename) == 0 ? "" : "-g";
         } catch (
             @SuppressWarnings("unused")
             NumberFormatException unused) {
@@ -323,15 +336,31 @@ public class MMapDirectory extends FSDirectory {
     }
   }
 
+  private static Optional<Integer> getSharedArenaMaxPermitsSysprop() {
+    try {
+      return Optional.ofNullable(System.getProperty(SHARED_ARENA_MAX_PERMITS_SYSPROP))
+          .map(Integer::valueOf);
+    } catch (@SuppressWarnings("unused") NumberFormatException | SecurityException ignored) {
+      Logger.getLogger(MMapDirectory.class.getName())
+          .warning(
+              "Cannot read sysprop "
+                  + SHARED_ARENA_MAX_PERMITS_SYSPROP
+                  + ", so the default value will be used.");
+      return Optional.empty();
+    }
+  }
+
   private static <A> MMapIndexInputProvider<A> lookupProvider() {
+    final var maxPermits = getSharedArenaMaxPermitsSysprop();
     final var lookup = MethodHandles.lookup();
     try {
       final var cls = lookup.findClass("org.apache.lucene.store.MemorySegmentIndexInputProvider");
       // we use method handles, so we do not need to deal with setAccessible as we have private
       // access through the lookup:
-      final var constr = lookup.findConstructor(cls, MethodType.methodType(void.class));
+      final var constr =
+          lookup.findConstructor(cls, MethodType.methodType(void.class, Optional.class));
       try {
-        return (MMapIndexInputProvider<A>) constr.invoke();
+        return (MMapIndexInputProvider<A>) constr.invoke(maxPermits);
       } catch (RuntimeException | Error e) {
         throw e;
       } catch (Throwable th) {
--- a/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
@@ -22,7 +22,6 @@ import java.lang.invoke.MethodType;
 import java.nio.channels.ClosedChannelException; // javadoc @link
 import java.nio.file.Path;
 import java.util.Locale;
-import java.util.Optional;
 import java.util.concurrent.Future;
 import java.util.function.BiPredicate;
 import org.apache.lucene.util.Constants;
@@ -104,7 +103,7 @@ public class MMapDirectory extends FSDirectory {
   public static final long DEFAULT_MAX_CHUNK_SIZE;
 
   /** A provider specific context object or null, that will be passed to openInput. */
-  private final Object attachment = PROVIDER.attachment().orElse(null);
+  private final Object attachment = PROVIDER.attachment();
 
   final int chunkSizePower;
 
@@ -220,8 +219,8 @@ public class MMapDirectory extends FSDirectory {
     boolean supportsMadvise();
 
     /** An optional attachment of the provider, that will be passed to openInput. */
-    default Optional<A> attachment() {
-      return Optional.empty();
+    default A attachment() {
+      return null;
     }
 
     default IOException convertMapFailedIOException(
--- a/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
@@ -105,7 +105,9 @@ public class MMapDirectory extends FSDirectory {
         String groupKey = IndexFileNames.parseSegmentName(filename).substring(1);
         try {
           groupKey += "-" + IndexFileNames.parseGeneration(filename);
-        } catch (NumberFormatException unused) {
+        } catch (
+            @SuppressWarnings("unused")
+            NumberFormatException unused) {
           // does not confirm to the generation syntax, or trash
         }
         return Optional.of(groupKey);
--- a/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
@@ -103,7 +103,7 @@ public class MMapDirectory extends FSDirectory {
   public static final long DEFAULT_MAX_CHUNK_SIZE;
 
   /** A provider specific context object or null, that will be passed to openInput. */
-  private final Object attachment = PROVIDER.attachment();
+  final Object attachment = PROVIDER.attachment();
 
   final int chunkSizePower;
 
--- a/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
@@ -101,7 +101,7 @@ public class MMapDirectory extends FSDirectory {
         if (filename.length() == segmentName.length()) {
           return Optional.empty();
         }
-        return Optional.of(segmentName);
+        return IndexFileNames.parseSegmentIdentifier(segmentName);
       };
 
   /**
--- a/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
@@ -45,6 +45,11 @@ import org.apache.lucene.util.Constants;
  * performance of searches on a cold page cache at the expense of slowing down opening an index. See
  * {@link #setPreload(BiPredicate)} for more details.
  *
+ * <p>This class supports grouping of files that are part of the same logical group. This is a hint
+ * that allows for better handling of resources. For example, individual files that are part of the
+ * same segment can be considered part of the same logical group. See {@link
+ * #setGroupingFunction(Function)} for more details.
+ *
  * <p>This class will use the modern {@link java.lang.foreign.MemorySegment} API available since
  * Java 21 which allows to safely unmap previously mmapped files after closing the {@link
  * IndexInput}s. There is no need to enable the "preview feature" of your Java version; it works out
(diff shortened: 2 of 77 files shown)


lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -199,20 +203,27 @@ public IndexInput openInput(String name, IOContext context) throws IOException {
...
+    IndexInput openInput(
+        Path path, IOContext context, int chunkSizePower, boolean preload, A attachment)
         throws IOException;
 
     long getDefaultMaxChunkSize();
 
     boolean supportsMadvise();
 
+    /** An optional attachment of the provider, that will be passed to openInput. */
+    default Optional<A> attachment() {
```
uschindler, line 223: I think this can just return null by default and let's remove the Optional. Optional would only make sense if we also pass the optional on openInput().

uschindler: We could also use `default <A> Optional<A> attachment()` as signature here so we don't need to make the whole class typed.

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -256,15 +267,15 @@ default IOException convertMapFailedIOException(
     }
   }
 
-  private static MMapIndexInputProvider lookupProvider() {
+  private static <A> MMapIndexInputProvider<A> lookupProvider() {
```
uschindler, line 270: like here, move this to `attachment()`

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
```diff
@@ -0,0 +1,112 @@
...
+
+@SuppressWarnings("preview")
+final class RefCountedSharedArena implements Arena {
+
+  static final int OPEN = 0;
+  static final int CLOSED = -1;
+
+  static final VarHandle STATE;
+
+  static {
```
uschindler, line 32: You want to use the VarHandle approach to spare instances, because AtomicReference would do the same?

uschindler (changes requested):
I like the code much more:
- the attachment is a good idea for the meantime to clearly separate compilation units
- the refocunting looks much better here - thanks. That was my main problem with the previous approach by @magibney

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
uschindler: or without optional: `default <A> A attachment()`

ChrisHegarty: The use of generics here is a little less than ideal. Defining the type parameter at the class-level allows the attachment type to be the same across the two instance methods, `openInput` and `attachment`. Whereas a type param on each of the individual methods means that the type could be different.

ChrisHegarty: The alternative, of course, is to just drop the use of generics here and use `Object`.

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
ChrisHegarty: There are certainly different approaches here. Conceptually, I think of ref counting as a simple `int` counter, then how to operate on that in a thread-safe way - that's how I got to what is in the PR.   Of course, now that I look at it, it could be rewritten with a `AtomicInteger`, since the CAS operations we need are there. And we don't operate on the state outside of that.

I updated to use AtomicInteger, since the code is a little more readably, and I don't think that the instance creation will be a problem.  I'm on the fence on this one, happy to keep or revert it.

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  void release() {
+    int value;
+    while (true) {
+      value = state.get();
+      if (value <= OPEN) {
+        throw new IllegalStateException("already closed");
+      }
+      if (state.compareAndSet(value, value - 1)) {
```
magibney, line 67: Outer CAS could be replaced by `decrementAndGet()`, removing the need for a `while` loop?

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
```diff
@@ -0,0 +1,102 @@
...
+final class RefCountedSharedArena implements Arena {
+
+  static final int OPEN = 0;
+  static final int CLOSED = -1;
+
+  private final String segmentName;
+  private final Runnable removeFromMap;
+  private final Arena arena;
+
+  AtomicInteger state;
```
magibney, line 33: any reason for `state` to not be `final`?

magibney (commented):
This looks good (sorry about the lack of write access to the original PR -- institutional repo, not an option).

`attachment()` is quite nice.

Unless I'm mistaken, this PR has ended up with refcounting that's nearly identical to the current state of #13555 (though it fixes a bug in the other PR's acquire, where CAS must be used in place of `getAndIncrement()`)? It's a small point, but I think CAS is _not_ necessary for `release()` here.

My only other concern applies here equally as to my earlier PR:
```java
      // TODO: what if the associated Directory is used for some purpose other than Lucene
      //  index files, and some of the filenames happen to be patterned so that a "segment"
      //  can be parsed from them? That could result in files being spuriously grouped
      //  together in the same arena. This would only practically be a problem if such a
      //  "segment" was never closed (refCt never 0), _and_ enough files associated with that
      //  "segment" were opened and closed (without ever closing the associated Arena)
      //  to exhaust virtual memory space.

File: lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java (hunk 5 above)
Comment (uschindler) [reply lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java:33]: +1

I think that's a relic from the previous plain integer with VarHandle.

**Review by uschindler — APPROVED** [review]
OK from my side, only the missing final (maybe it should be private, too) should be fixed.

The Generics are in fact a bit strange, but due to the MethodHandle and casting after ctor it works woth `Object` as type. I would have wished to get rid of the generic type. Maybe the field in MMapDirectory should use "?" as type, not "Object"? Just an idea.

Let me think ab bit about it, but it is a workaround for something that could be fixed later when the whole provider gets obsolete (Java 24 or 25 LTS).

File: lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java (hunk 5 above)
Comment (uschindler) [reply lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java:33]: Could be private, too!

**Comment by uschindler** [comment]
Hi,
we also get Updates on the JDK fixes: https://github.com/openjdk/jdk/pull/20158 (see issue about more details and some improvements Solr and/or Elasticsearch/Opensearch should do).

File: lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java (hunk 5 above)
Comment (ChrisHegarty) [reply lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java]: Done.

**Comment by magibney** [comment]
To expand a bit on the concern I raised above:

IIUC, in order for this to work properly (guaranteed to not potentially leak virtual memory address space) it depends on segment filename usage patterns and segment lifecycle. I think we may either need to:
1. provide a way to configure specific MMapDirectory instances to bypass this Arena pooling, or
2. disallow use of MMapDirectory for anything other than Lucene index files (because of the potential for collision of filename patterns that could be parsed as "segments", potentially allowing an Arena to live forever, accumulating an unlimited number of associated files.

Am I missing something -- is this not a concern?

**Comment by uschindler** [comment]
> To expand a bit on the concern I raised above:
> 
> IIUC, in order for this to work properly (guaranteed to not potentially leak virtual memory address space) it depends on segment filename usage patterns and segment lifecycle. I think we may either need to:
> 
> 1. provide a way to configure specific MMapDirectory instances to bypass this Arena pooling, or
> 2. disallow use of MMapDirectory for anything other than Lucene index files (because of the potential for collision of filename patterns that could be parsed as "segments", potentially allowing an Arena to live forever, accumulating an unlimited number of associated files.
> 
> Am I missing something -- is this not a concern?

In addition we could add a separate ctor option to disable grouping.

In addition I would go for option #2: Like for the segment main file it should also not use grouped segments for any other file that is not a correct segment number (it could validate the filename using the base32 hash starting with "_" or whatever this is).

Maybe add a protected method to MMapDirectory that checks if a filename is a segement file) and pass this one as `Predicate<String>` to the provider interface.

**Review by uschindler — APPROVED** [review]

File: lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java (hunk 4 above)
Comment (ChrisHegarty) [reply lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java]: That's simpler, thanks. Done.

**Comment by uschindler** [comment]
We may need some other logic: We have some files that can change without the segment name changing? How do we handle *.del od docValues update files? If they correspond to the same segment, they would only be closed when actually the whole segment is closed after full refresh.

This should be easily testable on windows: Files that have mmapped areas can't be deleted, so we may cause issues when closing deletion files.

But on the other hand: Aren't ".del" files not READONCE!?

**Review by uschindler — CHANGES_REQUESTED** [review]
I changed to request changes, because we should confirm the deletion/softdeletes and docvalues updates.

Maybe the file pattern should really be a Predicate to separate the implementation. Maybe add a method to return `protected Optional<String> getArenaGroupingKey(String filename)` (or as `Function<String,Optional<String>>`), which returns empty optional if no grouping allowed or the segment number otherwise as key.

**Comment by uschindler** [comment]
> I changed to request changes, because we should confirm the deletion/softdeletes and docvalues updates.
> 
> Maybe the file pattern should really be a Predicate to separate the implementation. Maybe add a method to return `protected Optional<String> getArenaGroupingKey(String filename)` (or as `Function<String,Optional<String>>`), which returns empty optional if no grouping allowed or the segment number otherwise as key.

This could be similarily implemented like the `setPreload(BiPredicate<String,IOContext>)` method.

**Comment by magibney** [comment]
heh, was just writing to suggest something similar (`Function<String,Optional<String>>`). SGTM.

**Comment by uschindler** [comment]
It could even be simple by passing the grouping key to the provider method directly in addition to filename.

File: lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java [hunk 6]
```diff
@@ -199,20 +202,27 @@ public IndexInput openInput(String name, IOContext context) throws IOException {
...
+  interface MMapIndexInputProvider<A> {
+    IndexInput openInput(
+        Path path, IOContext context, int chunkSizePower, boolean preload, A attachment)
         throws IOException;
 
     long getDefaultMaxChunkSize();
 
     boolean supportsMadvise();
 
+    /** An optional attachment of the provider, that will be passed to openInput. */
```
dsmiley, line 221: Without looking at the actual implementation, it wasn't clear what "attachment" is, or what's expected of this method.  Apparently this should return a *new* "attachment", and it's called on MMapDirectory's constructor (thus once, not per openInput).  Renaming "createXXXXX" would clarify.  

Are we being overly generic?  We can call it what it is -- arenaBySegment.

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+@SuppressWarnings("preview")
+final class RefCountedSharedArena implements Arena {
+
+  static final int OPEN = 0;
+  static final int CLOSED = -1;
+
+  private final String segmentName;
+  private final Runnable removeFromMap;
```
dsmiley, line 30: I would call this onClose

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
```diff
@@ -125,4 +134,31 @@ private final MemorySegment[] map(
     }
     return segments;
   }
+
+  public ConcurrentHashMap<String, RefCountedSharedArena> attachment() {
```
dsmiley, line 138: hm; I'd expect an Override annotation

dsmiley (commented):
Coincidentally I'm working on a similar ref-counting in a ConcurrentHashMap thing in Solr for something else (not for Arena) so it's nice to review this; bit of deja-vu.

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
```diff
@@ -125,4 +134,31 @@ private final MemorySegment[] map(
...
+   * Gets an arena for the give path, potentially aggregating files from the same segment into a
+   * single ref counted shared arena. A ref counted shared arena, if created will be added to the
+   * given arenas map.
+   */
+  static Arena getSharedArena(Path p, ConcurrentHashMap<String, RefCountedSharedArena> arenas) {
+    String filename = p.getFileName().toString();
+    String segmentName = IndexFileNames.parseSegmentName(filename);
+    if (filename.length() == segmentName.length()) {
+      // no segment found; just use a shared segment
+      return Arena.ofShared();
```
dsmiley, line 152: A global shared arena seems dangerous.  Shouldn't we _at least_ have an Arena instance for the Index (i.e. for the Directory instance?).  You could use the empty string as a key into the map, accomplishing this trivially.

uschindler: This is a new Arena so it's private to that file. It's basically a fall ack to current behaviour.

uschindler: I think we should make sure to read the segments file with READONCE. The other code here is fine. All unknown index files are opened with a separate shared Arena, like before.

So no need to change anything here. Just the one thing: Make sure we open the segments file with READONCE.

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
uschindler: Yes that's missing.

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
uschindler: I like it more generic. Its an implementation detail of the private provider. In Lucene 9.x we have another implementation which is a noop, but we don't have the term Arena there. Also in JDK 19 it is called "Session".

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
ChrisHegarty: That's better. Done.

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
ChrisHegarty: Oops. Added.

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
ChrisHegarty: Exactly. Additionally, we avoid the prefix `newXX`, or `createXX`, which prescribes that the implementation _must_ return a new instance. This may not be the case. In fact, an implementation could even choose to return a Supplier of XX if it wanted to. The point is that it can be almost anything.

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
ChrisHegarty: I opened the following PR to use READONCE in more places and ensure that segment files are opened with it, #13574

dsmiley: If Arena.ofShared returns a new Arena, then I really don't understand what this PR is accomplishing; what the value is.

uschindler: The case you are looking here at is when there is no relationship to a specific segment of an index. If it is a file which does not belong to a segment it returns a new shared arena (fallback to default).

uschindler: P.S.: For known files like `segments_*` that are definitely READONCE we added a check in https://github.com/apache/lucene/pull/13574. Those are opened with READONCE semantics and MMapDir will use a confined segment.

So unless we have some other files which are not belonging to segments, this case here should happen seldom (only for custom index files added by extra codecs).

dsmiley (commented):
Can someone take a stab at summarizing this for CHANGES.txt, thus avoiding details a reader is unlikely to know about?  Like me :-). How will a user of Lucene benefit?

ChrisHegarty:
> Can someone take a stab at summarizing this for CHANGES.txt, thus avoiding details a reader is unlikely to know about? Like me :-). How will a user of Lucene benefit?

```
* GITHUB#13570, GITHUB#13574, GITHUB#13535: Avoid performance degradation with closing shared Arenas. Index
files are mmapped one-to-one with the JDK's foreign shared Arena. The JVM deoptimizes the top few frames of
all threads when closing a shared Arena (see JDK-8335480). We mitigate this situation by 1) using a confined
Arena where appropriate, and 2) grouping files from the same segment to a single shared Arena.
```

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
ChrisHegarty: Ok, https://github.com/apache/lucene/pull/13574 is merged, and we have enforcement in the tests to disallow opening `segment_*` with a context other than `READONCE`. Unless I'm mistaken, then this concern has been addressed.

magibney:
This is looking good. I think making it possible for MMapDirectory to customize/bypass grouping is still important -- something along the lines of what Uwe was thinking in [these](#issuecomment-2228827755) [comments](#issuecomment-2228845198):
>>Maybe the file pattern should really be a Predicate to separate the implementation. Maybe add a method to return protected Optional<String> getArenaGroupingKey(String filename) (or as Function<String,Optional<String>>), which returns empty optional if no grouping allowed or the segment number otherwise as key.
>
>This could be similarily implemented like the setPreload(BiPredicate<String,IOContext>) method.

>It could even be simpler by passing the grouping key to the provider method directly in addition to filename. Then the code that figures out which group is used can reside in MMapDirectory only.

uschindler:
I am fine with everything - BUT: we should still check how the grouping works when for an existing segment an additional `*.del` file is updated (same for softdeletes and docvalues updates). If it is treated as same segment, at least for del files it would not cause too big issues (because those are quite small), but if we have in-place atomic updates, we could have files already closed and deleted, but arena still there.

I will do some checks on Windows which is very sensitive with deleting still open files.

uschindler:
On Windows I got the following test error - kindly what I expected (Windows is not allowed to delete files which are still open or mmapped):

```
TestIndexWriter > testDeleteUnusedFiles FAILED
    java.lang.AssertionError
        at __randomizedtesting.SeedInfo.seed([6C3A14A35BBB8424:D30C462C33E329CB]:0)
        at org.junit.Assert.fail(Assert.java:87)
        at org.junit.Assert.assertTrue(Assert.java:53)
        at org.apache.lucene.index.TestIndexWriter.testDeleteUnusedFiles(TestIndexWriter.java:1303)
```

ChrisHegarty:
> I am fine with everything - BUT: we should still check how the grouping works when for an existing segment an additional `*.del` file is updated (same for softdeletes and docvalues updates). If it is treated as same segment, at least for del files it would not cause too big issues (because those are quite small), but if we have in-place atomic updates, we could have files already closed and deleted, but arena still there.
> 
> I will do some checks on Windows which is very sensitive with deleting still open files.

Same, I checking too.  Additionally, I'm still sketching out the few options around API support for a `groupingFunction`, etc, to allow for an opt-out.

uschindler:
It fails reproducible. Let me look into it :-) Not sure what the issue is (if it is deletes, softdeletes or docvalues updates).

ChrisHegarty:
> On Windows I got the following test error - actually what I expected (Windows is not allowed to delete files which are still open or mmapped

Ha! you're too fast. I'm still spinning up a VM to test Windows.

> It fails reproducible. Let me look into it :-) Not sure what the issue is (if it is deletes, softdeletes or docvalues updates).

Thanks. It's with you. Appreciate it.

uschindler:
> It fails reproducible. Let me look into it :-) Not sure what the issue is (if it is deletes, softdeletes or docvalues updates).

This test failure is "interesting". This test enforces MMapDirectory and it actually tests that the file can't be deleted. It is a CFS file....

```
      // NOTE: here we rely on "Windows" behavior, i.e. even though IW wanted to delete _0.cfs
      // since it was merged away, because we have a reader open against this file,
      // it should still be here:
      assertTrue(Files.exists(indexPath.resolve("_0.cfs")));
```

uschindler:
Test also fails in main branch, so this is not new. Must have been to do with our previous changes. Interestingly it does not fail on [Policeman Jenkins](https://jenkins.thetaphi.de/job/Lucene-main-Windows/). I need more beers.....

uschindler:
So somebody what has more knowledge about how the filenames of deletion and/or softdeletes/docvalues instant update files are generated should help us.

I think deleteion files should not be a problem, as those are READONCE, but for the docvalues updates we may possibly need to have them on the exclusion list for grouping.

I have no time to look into this now, I will check how to fix the test with Windows 11 and possibly setup a Windows 11 node on Policeman Jenkins.

ChrisHegarty:
I added a grouping function to the API - see [4591f7f](https://github.com/apache/lucene/pull/13570/commits/4591f7f904f0f76d4ef3fa5b4accee184183ddf1)

ChrisHegarty:
> So somebody what has more knowledge about how the filenames of deletion and/or softdeletes/docvalues instant update files are generated should help us.
> 
> I think deleteion files should not be a problem, as those are READONCE, but for the docvalues updates we may possibly need to have them on the exclusion list for grouping.

I'll try to look into this, and also get someone to help, if needed.

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -83,6 +86,19 @@ public class MMapDirectory extends FSDirectory {
...
+
+  /** Argument for {@link #setGroupingFunction(Function)} that configures grouping by segment. */
+  public static final Function<String, Optional<String>> GROUP_BY_SEGMENT =
+      filename -> {
+        String segmentName = IndexFileNames.parseSegmentName(filename);
+        if (filename.length() == segmentName.length()) {
+          return Optional.empty();
+        }
+        return Optional.of(segmentName);
+      };
```
magibney, line 100: I think Uwe had floated the idea of doing extra validation that we're actually getting a segment name -- e.g. something like:
```java
String segmentPrefix = IndexFileNames.parseSegmentName(filename);
int segmentPrefixLength = segmentPrefix.length();
if (segmentPrefixLength > 1 && segmentPrefixLength != filename.length() && segmentName.charAt(0) == '_') {
  String segmentName = segmentPrefix.substring(1);
  try {
    // try to parse a proper segment identifier
    Long.parseLong(segmentName, Character.MAX_RADIX);
    return Optional.of(segmentName);
  } catch (NumberFormatException ex) {
    // unable to parse as a segment identifier
  }
}
return Optional.empty();
```

If we can account for all explicit exceptions (where we don't expect to be able to parse a segment name) we could even throw `IllegalArgumentException` here instead of returning `Optional.empty()`? (this would prevent grouping from being used in cases where it should not be, by failing hard -- and could be a good complement to ensuring `READ_ONCE` is used where appropriate).

magibney (commented):
> I added a grouping function to the API - see [4591f7f](https://github.com/apache/lucene/pull/13570/commits/4591f7f904f0f76d4ef3fa5b4accee184183ddf1)

Grouping function looks good! Left one comment about possibly adding extra validation in the `GROUP_BY_SEGMENT` function.

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
uschindler: Do we also have a function that parses the Long without reimplementing it here?

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -102,6 +118,11 @@ public class MMapDirectory extends FSDirectory {
    */
   public static final long DEFAULT_MAX_CHUNK_SIZE;
 
+  /** A provider specific context object or null, that will be passed to openInput. */
+  final Object attachment = PROVIDER.attachment();
+
+  Function<String, Optional<String>> groupingFunction;
```
uschindler, line 124: Maybe initialize it here (like for preload). Can also be private.

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -167,6 +188,7 @@ public MMapDirectory(Path path, LockFactory lockFactory, long maxChunkSize) thro
       throw new IllegalArgumentException("Maximum chunk size for mmap must be >0");
     }
     this.chunkSizePower = Long.SIZE - 1 - Long.numberOfLeadingZeros(maxChunkSize);
+    this.groupingFunction = GROUP_BY_SEGMENT;
```
uschindler, line 191: See above.

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -184,6 +206,21 @@ public void setPreload(BiPredicate<String, IOContext> preload) {
     this.preload = preload;
   }
 
+  /**
```
uschindler, line 209: Maybe add some short description also to class definition (like for preload).

uschindler (changes requested):
Looks fine. Some small changes and I agree with @magibney about the segment name parsing.

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
uschindler: Maybe add a method to validate and extract a segment name to `IndexFileNames` class. We possibly also have one in test framework that could be moved...

ChrisHegarty: I added `IndexFileNames::parseSegmentIdentifier` (from segmentName), which cleans things up nicely.  It still returns an empty optional for non parsable names, but we can provide a test-specific version of the groupingFunction to enforce certain restrictions of our code ( as suggested by @magibney  ). It's probably best to do this in a MockXX test class rather than impose this for all consumers.

lucene/core/src/java/org/apache/lucene/index/IndexFileNames.java
```diff
@@ -142,6 +143,26 @@ public static String stripSegmentName(String filename) {
     return filename;
   }
 
+  /**
```
uschindler, line 146: Hi,

I think the parseLong is too special, there is also a pattern to validate codec-generated filenames:
https://github.com/apache/lucene/blob/29fe87f08716bd8820a0a43f458d5a797a954f0c/lucene/core/src/java/org/apache/lucene/index/IndexFileNames.java#L220

We should use this one, if it does not match the pattern, we should return a shared segment.

In addition about deletions and in-place updates: There is the ability to extract the "generation" of a file. If the generation is not 0, the file is an update to an already existing segment (like new deletions and docvalues updates). As those should not be stay open forever, those should be a separate key. Or alternatively handle all files with generation!=0 as a separate shared arena.

so ideally we should create the key from the sgement name and the generation. :-)

ChrisHegarty: I need to spend a little more time thinking about this, but I pushed an initial change based on your suggestion.

uschindler: It broke because the parsing of Generation names somehow broke for older indexes. I think if that fails the fallback should be to don't use the generation name.

magibney: > In addition about deletions and in-place updates: There is the ability to extract the "generation" of a file. If the generation is not 0, the file is an update to an already existing segment (like new deletions and docvalues updates). As those should not be stay open forever, those should be a separate key. Or alternatively handle all files with generation!=0 as a separate shared arena.

We could have 2 Arenas per segment: 1 for segment files _without_ generation, 1 for segment files _with_ generation -- and the Arena for segment files _with_ generation could have some arbitrary built-in limit to the number of times (e.g., 1k?) the shared Arena can be used before `acquire()` returns false. This would cause Arenas to get rotated out periodically, ensuring that virtual memory would always be freed eventually.

In fact, I wonder if we might want some arbitrary limit to the number of mappings allowed for _each_ `RefCountedSharedArena` (not just the "with-generation" ones) ... I'm not sure -- if you map the same file with the same Arena multiple times, does the Arena itself do any deduplication? If one were to open a single Arena, and then use that Arena to repeatedly map the same 2G file, would the system eventually run out of virtual memory space?

ChrisHegarty: >  I'm not sure -- if you map the same file with the same Arena multiple times, does the Arena itself do any deduplication? If one were to open a single Arena, and then use that Arena to repeatedly map the same 2G file, would the system eventually run out of virtual memory space?

Arena as an API can do quite a few of things, but the way we're using it in Lucene is quite straightforward. In our usage it controls the lifetime of the mmapped files, allowing for timely deallocation. We're not using it to do any allocations directly, just associating the mappings to allow explicit release.

magibney: Right, I understand that part; my question was: if it's the act of mapping the file that allocates virtual memory, and there's no deduping, then in theory even with a limited number of files per segment, and even if segment generations don't change at all, if you re-map the files enough times on the same Arena without closing/de-allocating, you could still exhaust resources.

uschindler: Why should we remap already open files multiple times? Lucene opens only new segments, but old ones are kept open on index refreshes. We do not close or reopen already open files. Of course, we could track file names, too, but why should we do this?

We only have generations for updated files and those are now opened in separate arenas (as the generation for the original files is "0").

uschindler: In fact we could add some logic to place all files with a generation number!=0 (or none at all) in a single arena. Of course it would cause that files in that arena are never unmapped unless the last file is closed. This happens of course if the index segment is closed, but it might indeed be a good idea to limit arena reuse. If acquire returns false we could instruct the code to generate a new arena.

magibney: > Why should we remap already open files multiple times? Lucene opens only new segments, but old ones are kept open on index refreshes. We do not close or reopen already open files.

I understand that this is how it should be, and why. I was asking to be certain this wasn't a concern (esp. because the related expanded use of `IOContext.READ_ONCE` and `Arena.ofConfined()` was addressing this exact situation -- repeatedly re-opening files that had already been mapped). If this is not a concern, that's great.

And separate Arenas per generation seems good too, I just wanted to raise the possibility of multiple generations using the same arbitrarily-limited Arena. Shouldn't be _necessary_, but I thought it might strike a useful balance of further reducing `Arena.close()` while still providing "timely-enough" de-allocation.

ChrisHegarty: Cool!  I added an arbitrary limit to the ref counting, after which the `RefCountedShareArena` is removed from the map, and a new instance is created to handle the next set of group files. 👍 

I chose a limit of 256. This is quite arbitrary, but we can amend it once we're happy with the mechanism. We can even make this a property of the provider, and make it configurable through a system property.

I left the grouping as it was. That is:
 1. for segment files, one group per generation, and one group for all other files without a generation.
 2. No grouping for files are not valid codec segment files

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
```diff
@@ -47,9 +48,12 @@ boolean acquire() {
     int value;
     while (true) {
       value = state.get();
-      if (value < OPEN) {
+      if (value >= LIMIT) {
```
ChrisHegarty, line 51: I considered remodelling the state here (separating the count and state), in order to disallow future acquire, even after releases - so hitting the limit would be a terminal state. But it seemed not worth it, given our usage.

magibney: Hmm, but I think that unless we _do_ separate "total lifetime count" from state, then this is unlikely to ever hit in practice. The situation we're trying to protect against is repeated map and un-map, which, for a single segment or segment/gen, could easily stay below an outstanding limit of 256, while allocating an unlimited amount of VM space. (Accordingly, assuming we go with separate "total lifetime count" here, a higher limit would probably be appropriate).

ChrisHegarty: Yeah, I agree. It does help with a pathological test case, but that is not representative of the case we're concerned with. Lifetime max it is then.

uschindler: Yes. 👍👍👍👍

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
```diff
@@ -157,6 +157,8 @@ static Arena getSharedArena(
           arenas.computeIfAbsent(key, s -> new RefCountedSharedArena(s, () -> arenas.remove(s)));
       if (refCountedArena.acquire()) {
         return refCountedArena;
+      } else {
+        arenas.remove(key);
```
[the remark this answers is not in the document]
uschindler: This is cool. Yes let's atomically replace the arena in the map.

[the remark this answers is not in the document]
ChrisHegarty: Thanks, this is nice. Done.

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
ChrisHegarty: I pushed an implementation of this that allows a total of 1024 acquires per an instance.

ChrisHegarty:
Subject to review comment, I think the only thing that remains is whether we want to allow the default number of total ref counting to be configurable, say by a system property or something.

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
```diff
@@ -0,0 +1,138 @@
...
+    int value;
+    while (true) {
+      value = state.get();
+      if (value == CLOSED) {
+        throw new IllegalStateException("closed");
+      }
+      final int count = value & 0xFFFF;
+      if (count == 0) {
+        throw new IllegalStateException("nothing to release");
+      }
```
magibney, line 101: these 2 ISEs are effectively redundant, could be combined as:
```java
      final int count = value & 0xFFFF;
      if (count == 0) {
        throw new IllegalStateException(value == CLOSED ? "closed" : "nothing to release");
      }
```

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
```diff
@@ -0,0 +1,138 @@
...
+    int value;
+    while (true) {
+      value = state.get();
+      if (value == CLOSED) {
+        throw new IllegalStateException("closed");
+      }
+      final int remaining = value >>> 16;
+      if (remaining == 0) {
+        return false;
+      }
```
magibney, line 82: I think we must remove the `if (value == CLOSED)` ISE here, and simply return `false` below if `remaining == 0`. Unlike in `release()`, where we expect to hold a permit that would have prevented the entry from being closed, there's no such guarantee here. We can instead simply stick with `if (remaining == 0) return false`.

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -256,15 +323,15 @@ default IOException convertMapFailedIOException(
     }
   }
 
-  private static MMapIndexInputProvider lookupProvider() {
+  private static <A> MMapIndexInputProvider<A> lookupProvider() {
```
magibney, line 326: This is a minor point, but I wonder if at some point it'd be worth constraining the "attachment" type param to `interface MMapIndexInputProvider<A extends AutoCloseable>`. If attachment extends `AutoCloseable` that just feels cleaner from a lifecycle perspective, since the MMapDirectory that owns each attachment can then report back to the `MMapIndexInputProvider` when it's finished with the attachment (closing `attachment` in the `Directory.close()` method).

Then again, given that the current implementations don't actually require this functionality, if there'd be no barrier to modifying the API if/when such functionality is actually required, there's no reason to accommodate it provisionally (can leave it as-is).

(this thought occurred to me while considering handling all `Arena.close()` in a single-threaded executor within the MMapIndexInputProvider. If this is worth considering at all, it would be in a separate follow-up PR; I mention this here only as context for why I was thinking about making attachment `AutoCloseable`).

magibney (changes requested):
This looks good; I left a few comments, but the only one that really has to happen I think is the removal of the ISE in `acquire()`.

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
```diff
@@ -0,0 +1,138 @@
...
+      int newValue = ((remaining - 1) << 16) | ((value & 0xFFFF) + 1);
```
dsmiley, line 83: deserves a comment!

magibney: 👍 ... how about [this version](#discussion_r1686746629) of the same code? lol

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
ChrisHegarty: Yeah, your suggestion is better. Which is closer to the original intent or the semantics of acquire.

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
ChrisHegarty: Thanks, that's better.

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
```diff
@@ -0,0 +1,138 @@
...
+      if (this.state.compareAndSet(value, newValue)) {
```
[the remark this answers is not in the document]
ChrisHegarty: I like it. Thanks.

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
```diff
@@ -0,0 +1,138 @@
...
+ * class is 1024. When the total number of acquires is exhausted, then not more acquires are
+ * permitted and {@link #acquire()} returns false. This is independent of the actual number of the
+ * ref count.
+ */
+@SuppressWarnings("preview")
+final class RefCountedSharedArena implements Arena {
+
+  static final int CLOSED = 0;
+  // initial state of 0x400 (1024) maximum permits, and a ref count of 0
+  static final int INITIAL = 0x04000000;
```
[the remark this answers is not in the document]
ChrisHegarty: I added a sys property and limited it to 0x7FFF. 

`-Dorg.apache.lucene.store.MemorySegmentIndexInputProvider.sharedArenaMaxPermits=1` is equivalent to disabling sharing.

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -83,6 +93,26 @@ public class MMapDirectory extends FSDirectory {
...
+  public static final Function<String, Optional<String>> NO_GROUPING = filename -> Optional.empty();
+
+  /** Argument for {@link #setGroupingFunction(Function)} that configures grouping by segment. */
+  public static final Function<String, Optional<String>> GROUP_BY_SEGMENT =
+      filename -> {
+        if (!CODEC_FILE_PATTERN.matcher(filename).matches()) {
+          return Optional.empty();
+        }
+        String groupKey = IndexFileNames.parseSegmentName(filename).substring(1);
+        try {
```
uschindler, line 106: Now that we have a maximum number of increments per segment, we could easily handle all generations!=0 as one shared arena.
So when new deletions are added and the segment reloaded, it would create a new arena for all changes to segment up to the total number of updates.

uschindler: So I think we should simply use two slots: "generation=0/invalid" vs. "generation>0"

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
```diff
@@ -125,4 +135,77 @@ private final MemorySegment[] map(
...
+  @Override
+  public ConcurrentHashMap<String, RefCountedSharedArena> attachment() {
+    return new ConcurrentHashMap<>();
+  }
+
+  /**
+   * This sysprop allows to control the max number of permits that a RefCountedSharedArena will
+   * support for its lifetime. For example, to set the max number of permits to 256, pass the
+   * following on the command line pass {@code
+   * -Dorg.apache.lucene.store.MemorySegmentIndexInputProvider.sharedArenaMaxPermits=256}.
```
uschindler, line 148: I'd remove the internal class name and use MMapDirectory (see the example from 9.x: https://github.com/apache/lucene/blob/c23f6c09f7de569d48803712da1540d007746be6/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java#L159-L160) and the code reading it.

uschindler (approved):

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
uschindler: The goog thing is you could make this constant into MMapDirectory and add it to documentation there (like done for the above in 9.x).

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
ChrisHegarty: Yes, that's good. For the group key, we now append `-g` for all generation other than 0, and omit any suffix for invalid/0 generation.

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java
ChrisHegarty: I moved this. Rather than add additional methods to the interface, I passed the max permits as an arg to the constructor. I think this is fine.

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -83,6 +94,38 @@ public class MMapDirectory extends FSDirectory {
...
+          groupKey += IndexFileNames.parseGeneration(filename) == 0 ? "" : "-g";
```
uschindler, line 120: its easier to read that way:

```java
// keep the original generation (=0) in base group, later generations in extra group
if (IndexFileNames.parseGeneration(filename) > 0) {
  groupKey += "-g";
}
```

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -83,6 +94,38 @@ public class MMapDirectory extends FSDirectory {
...
+  /**
+   * This sysprop allows to control the total maximum number of mmapped files that can be associated
+   * with a single shared {@link java.lang.foreign.Arena foreign Arena}. For example, to set the max
+   * number of permits to 256, pass the following on the command line pass {@code
+   * -Dorg.apache.lucene.store.MMapDirectory.sharedArenaMaxPermits=256}. Setting a value of 1
+   * associates one file to one shared arena.
+   *
+   * @lucene.internal
+   */
+  static final String SHARED_ARENA_MAX_PERMITS_SYSPROP =
```
uschindler, line 106: should be public as it has Javadocs (to be conform to other sysprops in 9.x).

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -256,15 +336,31 @@ default IOException convertMapFailedIOException(
     }
   }
 
-  private static MMapIndexInputProvider lookupProvider() {
+  private static Optional<Integer> getSharedArenaMaxPermitsSysprop() {
```
uschindler, line 339: OptionalInt?

IMHO, this should be an integer from beginning and initialized to default in this class.

uschindler (commented):
Looks fine, some suggestions.

To make the javadocs visible the static sysprop name constant should be public.

uschindler (approved):
Looks fine now.

ChrisHegarty:
This looks like it's in good shape. @magibney Any final comments? Otherwise, I plan to merge tomorrow. And then figure out how to backport!

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
```diff
@@ -0,0 +1,146 @@
...
+ * limitations under the License.
+ */
+package org.apache.lucene.store;
+
+import java.lang.foreign.Arena;
+import java.lang.foreign.MemorySegment;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * A reference counted share Arena.
```
magibney, line 24: typo: share -> shared

magibney (approved):
Looks good, thanks! (aside from a javadoc typo that I happened to notice)

lucene/core/src/java21/org/apache/lucene/store/RefCountedSharedArena.java
ChrisHegarty: Fixed, thanks

uschindler:
> Otherwise, I plan to merge tomorrow. And then figure out how to backport!

Code duplication with Arena vs. Session hell!

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -83,6 +94,41 @@ public class MMapDirectory extends FSDirectory {
...
+        String groupKey = IndexFileNames.parseSegmentName(filename).substring(1);
+        try {
+          // keep the original generation (=0) in base group, later generations in extra group
+          if (IndexFileNames.parseGeneration(filename) > 0) {
+            groupKey += "-g";
+          }
+        } catch (
+            @SuppressWarnings("unused")
+            NumberFormatException unused) {
+          // does not confirm to the generation syntax, or trash
```
dsmiley, line 127: I would use an assert here

uschindler: This fails with backwards indexes, so we can't assert here.

dsmiley: interesting; thanks for checking.

ChrisHegarty: Right. After we noticed this, we added a unit test to cover the back compat scenario too. 👍

uschindler:
It's finally done. Thanks to all and help while implementing my idea!

uschindler:
Do we have a backport PR? Should I work on it?

uschindler:
@ChrisHegarty the PR ist listed in the 9.x section of changes. I could work on a backport (possibly tomorrow). I just want to make sure you haven't started.

I am not yet sure how to handle java 20 and java 19. I may possibly leave those as they are.

ChrisHegarty:
> @ChrisHegarty the PR ist listed in the 9.x section of changes. I could work on a backport (possibly tomorrow). I just want to make sure you haven't started.
> 
> I am not yet sure how to handle java 20 and java 19. I may possibly leave those as they are.

Yeah, I had the same thought too - to just do this for JDK 21+. 

I’ve not started, but it is on my todo list. I’ll raise a PR for it tomorrow, and if you have time maybe you could help review.

uschindler:
We may need to change the changes entry to tell that to enduser.

PR #13574, opened 2024-07-16
Ensure to use IOContext.READONCE when reading segment files

ChrisHegarty:
This commit uses IOContext.READONCE in more places where the index input  is clearly being read once by the thread opening it. We can then enforce that segment files are only opened with READONCE, in the test specific Mock directory wrapper.

Much of the changes in this PR update individual test usage, but there is one non-test change to Directory::copyFrom.

relates #13570

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/Directory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/Directory.java
@@ -172,7 +172,7 @@ public abstract class Directory implements Closeable {
 
   /**
    * Copies an existing {@code src} file from directory {@code from} to a non-existent file {@code
-   * dest} in this directory.
+   * dest} in this directory. The given IOContext is only used for opening the destination file.
    */
   public void copyFrom(Directory from, String src, String dest, IOContext context)
       throws IOException {
--- a/lucene/core/src/java/org/apache/lucene/store/Directory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/Directory.java
@@ -172,7 +172,7 @@ public abstract class Directory implements Closeable {
 
   /**
    * Copies an existing {@code src} file from directory {@code from} to a non-existent file {@code
-   * dest} in this directory. The source file will be opened with {@link IOContext#READONCE}
+   * dest} in this directory.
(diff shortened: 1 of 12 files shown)


lucene/core/src/java/org/apache/lucene/store/Directory.java
```diff
@@ -172,12 +172,12 @@ public ChecksumIndexInput openChecksumInput(String name) throws IOException {
 
   /**
    * Copies an existing {@code src} file from directory {@code from} to a non-existent file {@code
-   * dest} in this directory.
+   * dest} in this directory. The source file will be opened with {@link IOContext#READONCE}
```
ChrisHegarty, line 175: Rethinking this, Maybe we can drop the javadoc change, it's an implementation detail.

lucene/test-framework/src/java/org/apache/lucene/tests/store/MockDirectoryWrapper.java
```diff
@@ -814,6 +814,14 @@ public synchronized IndexInput openInput(String name, IOContext context) throws
 
     context = LuceneTestCase.newIOContext(randomState, context);
     final boolean confined = context == IOContext.READONCE;
+    if (name.startsWith(IndexFileNames.SEGMENTS) && confined == false) {
```
uschindler, line 817: Great!

lucene/core/src/java/org/apache/lucene/store/Directory.java
```diff
@@ -177,7 +177,7 @@ public ChecksumIndexInput openChecksumInput(String name) throws IOException {
   public void copyFrom(Directory from, String src, String dest, IOContext context)
```
uschindler, line 177: I think the change to Javadocs was mostly fine, maybe it should only be an @implNote.

Maybe it should not be explicit and only say: The given IOContext is only used for opening the destination file.

uschindler (approved):

uschindler:
If you find other use cases for READONCE, be free to add them.

I really like the MockDirectory check!

PR #13576, opened 2024-07-16
Reduce memory usage of SkipListWriter

bugmakerrrrrr:
### Description

Currently, we always use the size `Lucene99PostingsFormat.MAX_SKIP_LEVELS` to initialize the array fields of SkipListWriter. However, we have already calculated the maximum possible skip levels upfront in `MultiLevelSkipListWriter`.

### DIFF
--- a/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene90/Lucene90SkipWriter.java
+++ b/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene90/Lucene90SkipWriter.java
@@ -46,8 +46,8 @@ import org.apache.lucene.store.IndexOutput;
  * uptos(position, payload). 4. start offset.
  */
 final class Lucene90SkipWriter extends MultiLevelSkipListWriter {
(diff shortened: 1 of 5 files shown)


bugmakerrrrrr:
@jpountz @mikemccand could you take a look when you get a chance

jpountz (approved):
This looks correct to me.

mikemccand:
This looks correct to me too -- we know the max number of skip levels for this particular segment and should size the arrays based on that, not based on the global worst case max ever for the worst case segment.

But, we are now moving away from the multi-level skip lists to inlined double level skip lists (https://github.com/apache/lucene/pull/13585)...

bugmakerrrrrr:
@mikemccand  ok, I will close this PR. Btw, I think when we can do the same optimization for the skip reader, and we usually need to read old indices.  I want to know if I should optimize the old skip reader, or if I just need to close #13623?"

jpountz:
Closing this PR as discussed above. We could fix heap usage on the reader side, though I haven't seen any complaints about it.

PR #13577, opened 2024-07-16
WIP do not merge

msokolov:
### Description
This is a follow-on to  https://github.com/apache/lucene/pull/13566 that adds some unit testing for HnswUtil plus an implementation that identifies strongly-connected components of a directed graph. However this implementation is terrible - adding it into the HNSW merge step makde the whole indexing take 4-5x longer!  I think it's correct but does lots of extra work. So the main thing of value here is probably the tests...


msokolov:
This strongly-connected test is hard to make efficient and it's actually more than we need, given the way we search the graphs hierarchically. I'll follow up with a different approach

Issue #13580, opened 2024-07-17
Are we properly accounting for `NeighborArray.rwlock`?


msokolov:
### Description

Really three issues:
1. `OnHeapHnswGraph.ramBytesUsed` has a complicated job - it's hard to tell whether it takes into account the probably quite significant RAM usage of the `ReentrantReadWriteLock` that gets created for every node in the graph.
2. We create these locks even in the case where we are flushing (via `HnswGraphBuilder`) and don't ever use them. We only need locks available when merging
3. We could hash the node ids and share a smaller pool of locks among them and thus save a bunch of allocations

msokolov:
I also found we are acquiring these locks for writing in `HnswGraphBuilder.addDiverseNeighbors` even in the single-threaded case where it is not needed.

benwtrent:
@msokolov related: https://github.com/apache/lucene/issues/12732

msokolov:
ooh thanks @benwtrent I had forgotten about that. I'm working up a new version based on (3) above that I hope will reduce this usage and only require it for concurrent mergers.

PR #13581, opened 2024-07-17
HnswLock: access locks via hash and only use for concurrent indexing

msokolov:
Addresses https://github.com/apache/lucene/issues/13580 by adding a locking wrapper for OnHeapHnswGraph's NeighborArrays, and supplying this when running concurrent merges. With this:

1. We no longer guard access to rows with locks when graph build is done in a single thread.
2. We only create a fixed number of locks rather than a lock per node

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
+++ b/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
@@ -63,6 +63,7 @@ public class HnswGraphBuilder implements HnswBuilder {
       beamCandidates; // for levels of graph where we add the node
 
   protected final OnHeapHnswGraph hnsw;
+  protected final HnswLock hnswLock;
 
   private InfoStream infoStream = InfoStream.getDefault();
   private boolean frozen;
@@ -110,6 +111,7 @@ public class HnswGraphBuilder implements HnswBuilder {
         beamWidth,
         seed,
         hnsw,
+        null,
         new HnswGraphSearcher(new NeighborQueue(beamWidth, true), new FixedBitSet(hnsw.size())));
   }
 
@@ -131,6 +133,7 @@ public class HnswGraphBuilder implements HnswBuilder {
       int beamWidth,
       long seed,
       OnHeapHnswGraph hnsw,
+      HnswLock hnswLock,
       HnswGraphSearcher graphSearcher)
       throws IOException {
     if (M <= 0) {
@@ -146,6 +149,7 @@ public class HnswGraphBuilder implements HnswBuilder {
     this.ml = M == 1 ? 1 : 1 / Math.log(1.0 * M);
     this.random = new SplittableRandom(seed);
     this.hnsw = hnsw;
+    this.hnswLock = hnswLock;
     this.graphSearcher = graphSearcher;
     entryCandidates = new GraphBuilderKnnCollector(1);
     beamCandidates = new GraphBuilderKnnCollector(beamWidth);
@@ -327,12 +331,14 @@ public class HnswGraphBuilder implements HnswBuilder {
         continue;
       }
       int nbr = candidates.nodes()[i];
-      NeighborArray nbrsOfNbr = hnsw.getNeighbors(level, nbr);
-      nbrsOfNbr.rwlock.writeLock().lock();
-      try {
+      if (hnswLock != null) {
+        try (HnswLock.LockedRow rowLock = hnswLock.write(level, nbr)) {
+          NeighborArray nbrsOfNbr = rowLock.row;
+          nbrsOfNbr.addAndEnsureDiversity(node, candidates.scores()[i], nbr, scorerSupplier);
+        }
+      } else {
+        NeighborArray nbrsOfNbr = hnsw.getNeighbors(level, nbr);
         nbrsOfNbr.addAndEnsureDiversity(node, candidates.scores()[i], nbr, scorerSupplier);
-      } finally {
-        nbrsOfNbr.rwlock.writeLock().unlock();
       }
     }
   }
--- a/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswConcurrentMergeBuilder.java
+++ b/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswConcurrentMergeBuilder.java
@@ -40,6 +40,7 @@ public class HnswConcurrentMergeBuilder implements HnswBuilder {
 
   private final TaskExecutor taskExecutor;
   private final ConcurrentMergeWorker[] workers;
+  private final HnswLock hnswLock;
   private InfoStream infoStream = InfoStream.getDefault();
   private boolean frozen;
 
@@ -55,6 +56,7 @@ public class HnswConcurrentMergeBuilder implements HnswBuilder {
     this.taskExecutor = taskExecutor;
     AtomicInteger workProgress = new AtomicInteger(0);
     workers = new ConcurrentMergeWorker[numWorker];
+    hnswLock = new HnswLock(hnsw);
     for (int i = 0; i < numWorker; i++) {
       workers[i] =
           new ConcurrentMergeWorker(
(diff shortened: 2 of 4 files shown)


msokolov:
Tested with 1M 256-dim docs ... at least it doesn't seem to make anything worse?

# M=16,width=50
Tests with 1M 256-d vectors, M=16, beam-width-index=50
| condition | merge threads | recall | latency | index time |
|-----------|---------------|--------|---------|------------|
| candidate |  12           | 0.927  | 2.37    | 177206     |
| candidate |  12           | 0.928  | 2.57    | 174283     |
| baseline  |  12           | 0.927  | 2.49    | 178528     |
| baseline  |  12           | 0.927  | 2.40    | 180748     |
| candidate |   1           | 0.925  | 2.37    | 176221     |
| candidate |   1           | 0.925  | 2.36    | 186181     |
| baseline  |   1           | 0.926  | 2.46    | 176939     |
| baseline  |   1           | 0.926  | 2.55    | 185635     |

# M=32,width=100
Tests with 1M 256-d vectors, M=32, beam-width-index=100
| condition | merge threads | recall | latency | index time |
|-----------|---------------|--------|---------|------------|
| candidate |  12           | 0.969  | 2.54    | 390794     |
| candidate |  12           | 0.969  | 2.51    | 434838     |
| baseline  |  12           | 0.969  | 2.55    | 393600     |
| baseline  |  12           | 0.967  | 2.82    | 441376     |
| candidate |   1           | 0.969  | 2.48    | 428682     |
| candidate |   1           | 0.970  | 2.49    | 392494     |
| baseline  |   1           | 0.970  | 2.72    | 403489     |
| baseline  |   1           | 0.968  | 2.44    | 409693     |

benwtrent:
It's weird that the merge threads have no impact on index build time. 

Did you create a bunch of segments and then force merge to exercise this code path?

Or was the buffer size too large to even allow merges?

msokolov:
I agree it's weird we saw no impact -- I'll retry with -forceMerge -- probably there was not enough merge activity?

benwtrent:
@msokolov whenever I had to benchmark the original parallel merge change, the way to isolate was reduce the KnnIndexer buffer size dramatically to create multiple segments, then measure how long it takes to force merge to a single segment. 

The merge activity during indexing was generally not enough to detect meaningful change.

benwtrent:
OK, I ran this over 500k vectors, merging with 8 threads, flushing every 10MB. 

Main
```
Force merge done in: 123535 ms
0.966	 1.94	500000	100	16	250	3919	262525	1.00	post-filter
```

this branch:
```
Force merge index in wiki1024en.train-16-250.index
Force merge done in: 130107 ms
0.966	 1.76	500000	100	16	250	3908	268437	1.00	post-filter
```

So, maybe 5% slower?

@zhaih could you test as well? This change does seem to clean things up a bit. It would be worth it if it doesn't significantly slow things down.

msokolov:
I've been running a bunch of tests -- ram buffer was set to 64MB (I think this is what's checked in on luceneutil main rn). I believe to get the most accurate assessment we should sum the indexing time (since significant merging does happen there) and the force merge time. I found I can get it to go faster by adding more locks - continuing to explore in order to get a full picture, and will share once I have a few more runs complete. I also experimented with not allocating these LockedRow objects but it doesn't seem to help - I think they are well-handled by escape analysis?

Edarke:
@msokolov Sorry to interject, but the hash function `Objects.hash(level, node) % NUM_LOCKS;` seems like it is prone to lock contention if workers have power of 2 offsets. Would it be worthwhile to avoid the `Integer[]` allocation and smear the hash instead?

```java
private static int hash(int level, int node) {
   int h = level * 31 + node; 
   h ^= (h >>> 20) ^ (h >>> 12);
   return h ^ (h >>> 7) ^ (h >>> 4);
 }

```

zhaih:
I have run the benchmark and got:
```
baseline:
reindex takes 416602ms
Force merge done in: 275695 ms

candidate:
reindex takes 410387 ms
Force merge done in: 278062 ms
```
To me feels like no obvious impact on performance, using 8 merge thread and 12 workers

msokolov:
I'm planning to merge - would just like to get a measure of the impact on heap usage. Looking at ReentrantReadWriteLock I see that it contains three objects: a ReadLock, WriteLock, and a Sync (and pointers to the shared Sync in the r/w locks). The Sync itself looks quite complex, including a doubly-linked list of Nodes, which I think represent Threads that have acquired the lock. I gave up trying to calculate the heap usage by inspection since it seems to be allocating dynamically based on lock acquisition? So maybe some empirical measurement will help

msokolov:
> @msokolov Sorry to interject, but the hash function Objects.hash(level, node) % NUM_LOCKS; seems like it is prone to lock contention if workers have power of 2 offsets. Would it be worthwhile to avoid the Integer[] allocation and smear the hash instead?

Thanks for interjection! I'm sorry, but I don't really understand. Which Integer[] allocation did you mean?

msokolov:
Thank you for the measurement! 125 bytes/lock seems like quite a bit, comparing it to M=32 we would expect 128 bytes per node so this would almost double the RAM usage of OnHeapHnswGraph. Given that, and the fact it is not being accounted for, and that the performance change is negligible, I think we can go ahead.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
```diff
@@ -327,12 +331,14 @@ private void addDiverseNeighbors(int level, int node, NeighborArray candidates)
         continue;
       }
       int nbr = candidates.nodes()[i];
-      NeighborArray nbrsOfNbr = hnsw.getNeighbors(level, nbr);
-      nbrsOfNbr.rwlock.writeLock().lock();
-      try {
+      if (hnswLock != null) {
+        try (HnswLock.LockedRow rowLock = hnswLock.write(level, nbr)) {
+          NeighborArray nbrsOfNbr = rowLock.row;
```
zhaih, line 336: This way we're not forcing people to use the lock if it presents, e.g. someone can still carelessly just use `hnsw.getNeighbors` without getting any errors. I wonder whether we could put the lock in `OnHeapHnswGraph` and expose a `getNeighbors(int level, int node, Consumer<NeighborArray> operation, bool read)` method instead? Also we can throw exception in original method in case the lock exists?

msokolov: I like the idea of more safety in the API to prevent accidentally circumventing the lock, but I'm not sure if I followed the idea completely. What would the Consumer do ?

zhaih: Sorry it's better to name it `operateOnNeighbors`.
The consumer would pass the actual operation into the method, since I feel it's better to not let the caller deal with the lock at all, so in `HnswGraphBuilder` we should always call
```
operateOnNeighbors(...)
```
and within `OnHeapHnswGraph`
```
operateOnNeighbors(int level, int node, Consumer<NeighborArray> operation, bool read) {
    if (this.lock != null) {
        // use RowLock and call operation.accept(nbrArray)
    } else {
        operation.accept(getNeighbor(leve, node));
   }
}
```

msokolov:
I'm going to merge as-is and we can follow up with the additional safety measure

msokolov:
Also backported to 9x

Edarke:
> > @msokolov Sorry to interject, but the hash function Objects.hash(level, node) % NUM_LOCKS; seems like it is prone to lock contention if workers have power of 2 offsets. Would it be worthwhile to avoid the Integer[] allocation and smear the hash instead?
> 
> Thanks for interjection! I'm sorry, but I don't really understand. Which Integer[] allocation did you mean?

Ah sorry, I was referring to the varargs invocation of `Objects.hash(Object...)` implicitly boxing the primitive ints. Not sure if that's a concern, but `Objects.hash` produces low entropy in the lower bits, which is ordinarily fine because collections like `HashMap` will smear hashCodes before computing array indexes.

```java
int lockid = Objects.hash(level, node) % NUM_LOCKS;
```

PR #13585, opened 2024-07-18
Inline skip data into postings lists

jpountz:
This updates the postings format in order to inline skip data into postings. This format is generally similar to the current `Lucene99PostingsFormat`, e.g. it shares the same block encoding logic, but it has a few differences:
 - Skip data is inlined into postings to make the access pattern more sequential.
 - There are only 2 levels of skip data: on every block (128 docs) and every 32 blocks (4096 docs).

In general, I found that the fact that skip data is inlined may slow down a bit queries that don't need skip data at all (e.g. `CountOrXXX` tasks that never advance of consult impacts) and speed up a bit queries that advance by small intervals. The fact that the greatest level only allows skipping 4096 docs at once means that we're slower at advancing by large intervals, but data suggests that it doesn't significantly hurt performance. Phrase queries and term queries sorted by field are slower for reasons that I haven't understood yet.

These results were produced on wikibigall without inter-segment concurrency.

```
               HighTermTitleSort      152.82      (1.3%)      105.67      (0.9%)  -30.9% ( -32% -  -29%) 0.000
                          Phrase       11.67      (5.2%)       10.13      (4.2%)  -13.2% ( -21% -   -4%) 0.000
                 CountOrHighHigh       56.79     (33.3%)       49.41     (21.1%)  -13.0% ( -50% -   62%) 0.141
               HighTermMonthSort     3598.70      (3.2%)     3372.04      (2.7%)   -6.3% ( -11% -    0%) 0.000
                  CountOrHighMed      104.44     (21.2%)       99.90     (18.1%)   -4.3% ( -36% -   44%) 0.486
                        Wildcard       54.26      (3.0%)       52.23      (2.6%)   -3.7% (  -9% -    1%) 0.000
                      TermDTSort      349.67      (6.0%)      339.57      (4.3%)   -2.9% ( -12% -    7%) 0.081
                          IntNRQ      113.09     (21.2%)      110.12     (21.6%)   -2.6% ( -37% -   51%) 0.699
                       CountTerm     9104.21      (4.1%)     8870.31      (6.0%)   -2.6% ( -12% -    7%) 0.115
                         Prefix3      296.80      (1.9%)      290.04      (2.0%)   -2.3% (  -6% -    1%) 0.000
                        HighTerm      383.13      (5.2%)      377.50      (7.5%)   -1.5% ( -13% -   11%) 0.472
                        PKLookup      286.07      (1.5%)      281.91      (2.1%)   -1.5% (  -4% -    2%) 0.012
           HighTermDayOfYearSort      758.57      (2.6%)      748.44      (2.9%)   -1.3% (  -6% -    4%) 0.121
            HighTermTitleBDVSort       13.27      (4.9%)       13.13      (6.2%)   -1.1% ( -11% -   10%) 0.546
                          Fuzzy1       98.52      (1.7%)       97.67      (2.1%)   -0.9% (  -4% -    3%) 0.154
                     AndHighHigh       62.93      (1.9%)       62.46      (1.5%)   -0.7% (  -4% -    2%) 0.164
                          Fuzzy2       62.42      (1.5%)       61.96      (1.9%)   -0.7% (  -4% -    2%) 0.184
                         Respell       49.68      (1.3%)       49.39      (1.5%)   -0.6% (  -3% -    2%) 0.171
              Or2Terms2StopWords      157.28      (1.7%)      157.04      (1.7%)   -0.2% (  -3% -    3%) 0.777
                      OrHighHigh       72.02      (1.7%)       72.21      (1.8%)    0.3% (  -3% -    3%) 0.642
                    AndStopWords       29.81      (2.2%)       29.94      (1.7%)    0.4% (  -3% -    4%) 0.495
             And2Terms2StopWords      151.81      (1.5%)      152.86      (1.8%)    0.7% (  -2% -    4%) 0.183
                    OrHighNotLow      384.08      (5.0%)      388.68      (6.9%)    1.2% ( -10% -   13%) 0.531
                   OrHighNotHigh      210.18      (6.1%)      213.18      (7.3%)    1.4% ( -11% -   15%) 0.502
                    OrHighNotMed      324.28      (5.3%)      329.41      (6.8%)    1.6% (  -9% -   14%) 0.413
                         MedTerm      567.00      (5.4%)      578.90      (8.1%)    2.1% ( -10% -   16%) 0.333
                     CountPhrase        3.24     (10.3%)        3.31     (13.2%)    2.2% ( -19% -   28%) 0.551
                         LowTerm      854.03      (4.9%)      873.32      (7.2%)    2.3% (  -9% -   15%) 0.248
                      AndHighMed      197.59      (1.5%)      203.05      (2.2%)    2.8% (   0% -    6%) 0.000
                   OrNotHighHigh      178.76      (6.5%)      184.38      (7.5%)    3.1% ( -10% -   18%) 0.156
                     OrStopWords       32.36      (2.8%)       33.56      (1.7%)    3.7% (   0% -    8%) 0.000
                        Or3Terms      158.54      (1.6%)      164.51      (2.1%)    3.8% (   0% -    7%) 0.000
                       OrHighMed      231.23      (1.8%)      241.40      (2.9%)    4.4% (   0% -    9%) 0.000
                       And3Terms      157.12      (1.3%)      164.32      (1.5%)    4.6% (   1% -    7%) 0.000
                       OrHighLow      732.71      (1.6%)      786.67      (3.1%)    7.4% (   2% -   12%) 0.000
                    OrNotHighMed      282.64      (6.5%)      306.83      (8.5%)    8.6% (  -6% -   25%) 0.000
                      OrHighRare      237.87      (7.8%)      259.37      (4.6%)    9.0% (  -3% -   23%) 0.000
                    OrNotHighLow      833.05      (2.4%)      946.10      (3.8%)   13.6% (   7% -   20%) 0.000
                CountAndHighHigh       41.24      (2.0%)       46.91      (2.7%)   13.8% (   8% -   18%) 0.000
                      AndHighLow      748.77      (1.7%)      870.25      (3.1%)   16.2% (  11% -   21%) 0.000
                 CountAndHighMed      120.32      (2.0%)      140.26      (3.5%)   16.6% (  10% -   22%) 0.000
```


mikemccand:
SO EXCITING, thank you for tackling this @jpountz -- I'll have a closer look soon.

#4036 is the long-ago issue where we started some discussion and linked to a cool paper about this.

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsWriter.java
```diff
@@ -0,0 +1,563 @@
...
+  private final ForDeltaUtil forDeltaUtil;
+
+  private boolean fieldHasNorms;
+  private NumericDocValues norms;
+  private final CompetitiveImpactAccumulator competitiveFreqNormAccumulator =
+      new CompetitiveImpactAccumulator();
+  private final CompetitiveImpactAccumulator skipCompetitiveFreqNormAccumulator =
+      new CompetitiveImpactAccumulator();
+
+  private final ByteBuffersDataOutput spareOutput = ByteBuffersDataOutput.newResettableInstance();
```
mikemccand, line 99: Maybe add some comments about how these are used?

It seems like `spareOutput` is to buffer skip bytes pointing to start of each block, and it spills into `skipOutput` which buffers those bytes until `SKIP_FACTOR` blocks are buffered, at which point it's all written (inlined) into the actual `docOut` file?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsWriter.java
```diff
@@ -0,0 +1,563 @@
...
+    }
+
+    blockOutput.copyTo(skipOutput);
+    blockOutput.reset();
+    lastBlockDocID = docID;
+    skipCompetitiveFreqNormAccumulator.addAll(competitiveFreqNormAccumulator);
+    competitiveFreqNormAccumulator.clear();
+
+    if ((docCount & SKIP_MASK) == 0) {
+      docOut.writeVInt(docID - lastSkipDocID);
```
mikemccand, line 353: Maybe add comment that this happens every `SKIP_FACTOR * BLOCK_SIZE` (4096) docs?

mikemccand (commented):
I LOVE how much simpler this is than the existing `MultiLevelSkipReader/Writer`!  It's so simple that the skipping logic itself is now inlined into the postings reading/writing, nicely mirroring how the skip data itself is now also inlined into the `.doc` file.

I'll make more time for another review (just scratched surface this time), but I couldn't see where we also wrote the file pointer into pos/pay files into the `spareOutput`/`skipOutput`?

It looks like every 4096 (`SKIP_FACTOR * BLOCK_SIZE`) docs we will insert a blob holding the skip data (bi-level skip list) for the next 4096 docs?

Maybe you could add a quick description into the `Lucene912PostingFormat.java` about how the bi-level skip list is encoded.  It's so simple, I love it.

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsWriter.java
```diff
@@ -0,0 +1,563 @@
...
+          lastSkipPayFP = payOut.getFilePointer();
+        }
+      }
+      spareOutput.copyTo(docOut);
+      spareOutput.reset();
+      skipOutput.copyTo(docOut);
+      skipOutput.reset();
+      lastSkipDocID = docID;
+      skipCompetitiveFreqNormAccumulator.clear();
+    } else if (end) {
```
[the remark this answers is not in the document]
mikemccand: OK I think I see why: when `end` is true, we still have buffered a bunch of blocks (< `SKIP_FACTOR` blocks) which will be written after this pending skip data is written.  So this is correct.

jpountz:
> I couldn't see where we also wrote the file pointer into pos/pay files

Indeed level 0 doesn't write pointers into pos/pay, it only records the total term freq of the block to know how many positions to skip when skipping a block. This may be the reason for the slowdown with phrase queries, I will look into recording pointers instead. Level 1 does record pointers though.

> It looks like every 4096 (SKIP_FACTOR * BLOCK_SIZE) docs we will insert a blob holding the skip data (bi-level skip list) for the next 4096 docs?

This is correct.

> Maybe you could add a quick description into the Lucene912PostingFormat.java about how the bi-level skip list is encoded. It's so simple, I love it.

I'll do it. As you can imagine, it's changed many (many many) times as I was figuring out how to make skipping fast and skip data space-efficient, but now that I'm starting to be happy with it, better file format description would indeed be helpful!

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsWriter.java
```diff
@@ -0,0 +1,563 @@
...
+    docCount++;
+
+    lastDocID = docID;
+  }
+
+  private void flushDocBlock(boolean end) throws IOException {
+    assert docBufferUpto != 0;
+
+    if (docBufferUpto < BLOCK_SIZE) {
+      PostingsUtil.writeVIntBlock(blockOutput, docDeltaBuffer, freqBuffer, docBufferUpto, writeFreqs);
```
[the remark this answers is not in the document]
jpountz: Indeed.

lucene/core/src/java/org/apache/lucene/codecs/lucene912/ForUtil.java
```diff
@@ -0,0 +1,1148 @@
...
+      out.writeLong(tmp[i]);
+    }
+  }
+
+  /** Number of bytes required to encode 128 integers of {@code bitsPerValue} bits per value. */
+  int numBytes(int bitsPerValue) {
+    return bitsPerValue << (BLOCK_SIZE_LOG2 - 3);
+  }
+
+  private static void decodeSlow(int bitsPerValue, DataInput in, long[] tmp, long[] longs)
```
mikemccand, line 303: It's too bad we have to fork all this code every time we make a new codec.  I wonder how to easily see the diffs of these supporting classes against the old postings format (`Lucene99PostingsFormat`)?  I could just download and do the diff myself ... but maybe GitHub has a quick way for me to diff this file against one that I specify?

jpountz: I don't know if you can do it. Maybe we should check in a clone of the existing format first, and then open a PR with changes so that the diff would be easier to read. I can do that if you'd like. FWIW I plan in updating the block format in a follow up, but right now `ForUtil` is exactly the same as `Lucene99`.

jpountz:
Skip data at level 0 now stores pointers into pos/pay files instead of incrementing posPendingCount by the total term freq of the block. This seems to slow down term queries marginally and improve phrase queries a bit. Also I noticed we would sometimes decode the same block of positions multiple times when it's shared by two doc blocks (because when moving to the next doc block we reset the position FP to the start of the pos block and decode them again, while they were already decoded, it looks like it's an existing issue in Lucene99 too), but fixing it only yielded a minor speedup.

luceneutil now gives this on wikibigall:

```
                 CountOrHighHigh       54.55     (28.7%)       48.60     (18.8%)  -10.9% ( -45% -   51%) 0.155
               HighTermMonthSort     3730.94      (3.2%)     3388.64      (2.2%)   -9.2% ( -14% -   -3%) 0.000
                         Prefix3       35.43      (5.2%)       34.45      (3.3%)   -2.8% ( -10% -    6%) 0.044
                  CountOrHighMed      102.26     (20.1%)       99.71     (16.3%)   -2.5% ( -32% -   42%) 0.666
                        Wildcard      106.00      (4.9%)      103.94      (3.7%)   -1.9% (  -9% -    6%) 0.154
                    OrHighNotLow      345.58      (4.7%)      339.60      (6.2%)   -1.7% ( -12% -    9%) 0.321
               HighTermTitleSort      132.39      (6.1%)      130.53      (2.2%)   -1.4% (  -9% -    7%) 0.329
                     AndHighHigh       62.75      (1.7%)       62.14      (1.8%)   -1.0% (  -4% -    2%) 0.078
                       CountTerm     9407.38      (4.8%)     9317.61      (3.1%)   -1.0% (  -8% -    7%) 0.454
                      TermDTSort      342.50      (3.8%)      339.94      (4.1%)   -0.7% (  -8% -    7%) 0.548
                        PKLookup      289.43      (1.8%)      287.33      (2.8%)   -0.7% (  -5% -    3%) 0.330
                         Respell       50.56      (1.8%)       50.26      (2.1%)   -0.6% (  -4% -    3%) 0.339
                          Fuzzy2       70.03      (1.5%)       69.67      (1.5%)   -0.5% (  -3% -    2%) 0.271
           HighTermDayOfYearSort      801.40      (2.3%)      797.96      (2.5%)   -0.4% (  -5% -    4%) 0.570
                        HighTerm      442.74      (5.8%)      441.35      (6.1%)   -0.3% ( -11% -   12%) 0.867
                          Fuzzy1       74.12      (1.6%)       74.28      (1.6%)    0.2% (  -2% -    3%) 0.665
            HighTermTitleBDVSort       14.58      (7.7%)       14.61      (7.2%)    0.3% ( -13% -   16%) 0.914
                    AndStopWords       29.90      (2.0%)       29.99      (1.8%)    0.3% (  -3% -    4%) 0.626
              Or2Terms2StopWords      159.18      (1.4%)      160.08      (1.2%)    0.6% (  -1% -    3%) 0.159
                      OrHighHigh       66.89      (1.7%)       67.38      (1.6%)    0.7% (  -2% -    4%) 0.155
             And2Terms2StopWords      152.71      (1.5%)      154.79      (1.3%)    1.4% (  -1% -    4%) 0.002
                    OrHighNotMed      322.27      (5.3%)      327.40      (7.3%)    1.6% ( -10% -   15%) 0.432
                         MedTerm      560.59      (6.6%)      571.19      (7.5%)    1.9% ( -11% -   17%) 0.399
                   OrHighNotHigh      233.42      (6.3%)      239.76      (7.7%)    2.7% ( -10% -   17%) 0.223
                          IntNRQ      140.44     (18.3%)      144.57     (18.8%)    2.9% ( -28% -   49%) 0.616
                      AndHighMed      151.63      (1.5%)      156.15      (1.5%)    3.0% (   0% -    6%) 0.000
                     OrStopWords       32.73      (1.9%)       33.84      (1.6%)    3.4% (   0% -    7%) 0.000
                         LowTerm      972.10      (5.9%)     1005.36      (6.5%)    3.4% (  -8% -   16%) 0.081
                          Phrase       11.68      (4.5%)       12.08      (4.1%)    3.4% (  -4% -   12%) 0.012
                       OrHighMed      199.56      (1.8%)      207.12      (2.1%)    3.8% (   0% -    7%) 0.000
                   OrNotHighHigh      214.79      (6.5%)      223.81      (8.2%)    4.2% (  -9% -   20%) 0.073
                        Or3Terms      159.96      (1.5%)      167.15      (1.4%)    4.5% (   1% -    7%) 0.000
                       And3Terms      157.48      (1.6%)      165.21      (1.5%)    4.9% (   1% -    8%) 0.000
                     CountPhrase        3.33     (11.8%)        3.56     (13.9%)    6.7% ( -16% -   36%) 0.100
                       OrHighLow      695.89      (2.2%)      743.44      (2.7%)    6.8% (   1% -   12%) 0.000
                      OrHighRare      242.91      (4.0%)      263.09      (4.2%)    8.3% (   0% -   17%) 0.000
                    OrNotHighMed      258.59      (6.9%)      285.71      (9.6%)   10.5% (  -5% -   28%) 0.000
                CountAndHighHigh       41.68      (2.3%)       47.42      (2.9%)   13.8% (   8% -   19%) 0.000
                      AndHighLow      913.56      (2.2%)     1063.18      (2.3%)   16.4% (  11% -   21%) 0.000
                    OrNotHighLow      843.74      (1.7%)      982.67      (3.5%)   16.5% (  11% -   22%) 0.000
                 CountAndHighMed      121.18      (2.0%)      142.55      (3.1%)   17.6% (  12% -   23%) 0.000
```

mikemccand:
> Also I noticed we would sometimes decode the same block of positions multiple times when it's shared by two doc blocks (because when moving to the next doc block we reset the position FP to the start of the pos block and decode them again, while they were already decoded, it looks like it's an existing issue in Lucene99 too), but fixing it only yielded a minor speedup.

Egads!  That must be a longstanding performance bug.

mikemccand:
Do you have any measure of how many bytes in a big posting is spent on skip data vs doc/freq blocks?

The gains on the last benchy look awesome!  It's surprising `CountOrHighHigh` is so affected: it never uses skip data right?  It just skips the skip data (heh)?  Hmm or is it doing that cool opto of `CountOrXY = Count(X) + Count(Y) - CountAndXY`?  Was that ever merged or just idea phase?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/ForUtil.java
mikemccand: No need to separate the commits like that.  Maybe next time!  We need a "make a new Codec format best practices guide".

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsWriter.java
```diff
@@ -0,0 +1,597 @@
...
+   * Output for a single block. This is useful to be able to prepend skip data before each block,
+   * which can only be computed once the block is encoded. The content is then typically copied to
+   * {@link #skipOutput}.
+   */
+  private final ByteBuffersDataOutput blockOutput = ByteBuffersDataOutput.newResettableInstance();
+
+  /**
+   * Output for groups of 32 blocks. This is useful to prepend skip data for these 32 blocks, which
+   * can only be done once we have encoded these 32 blocks. The content is then typically copied to
+   * {@link #docCount}.
```
mikemccand, line 113: `docCount` -> `docOut`?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -0,0 +1,1998 @@
...
+        }
+
+        nextSkipDoc += docIn.readVInt();
+        long impactLength;
+        if (indexHasFreq) {
+          impactLength = docIn.readVLong();
+        } else {
+          impactLength = 0L;
+        }
+        long skipLength = docIn.readVLong();
```
mikemccand, line 481: You could maybe make decoding faster for cases where skip data isn't needed/used if you first wrote `totalNumBytes` for the entire skip blob, so such use cases would just read that one vLong then skip over that many bytes?  After/within that you could separately express impact vs skip bytes, `nextSkipDoc` delta, etc.

jpountz: This is already what BlockDocsEnum does when `nextDoc()` is used (unless I misunderstood your recommendation), see line 531 in `BlockDocsEnum#moveToNextBlock`. I cannot do it for level 1 because I do not know at the time of decoding skip data whether the caller will later call advance(), but the overhead should be low in any case since we're only decoding level 1 skip data every 32 blocks.

mikemccand: OK thinking more ... I think it is not so simple.  This `PostingsEnum` does not know up front whether skipping will be used right?  The `TermsEnum#postings` flags don't include "I will not skip" flag.  So we must decode the skip data as we encounter it on the possibility that `.advance` is suddenly called.

jpountz: Right, this is why I always fully read skip data on level 1.

On level 0, we can only skip it because the only piece of information we need from skip data for BlockDocsEnum is the last docID in the block. But we can also know this by looking up the value at the last index in the decoded doc block. So BlockDocsEnum skips the skip data and reads the last doc in the block that it would usually know from skip data from decoded doc IDs instead.

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsWriter.java
```diff
@@ -0,0 +1,597 @@
...
+    blockOutput.copyTo(skipOutput);
+    blockOutput.reset();
+    lastBlockDocID = docID;
+    if (writeFreqs) {
+      skipCompetitiveFreqNormAccumulator.addAll(competitiveFreqNormAccumulator);
+      competitiveFreqNormAccumulator.clear();
+    }
+
+    if ((docCount & SKIP_MASK) == 0) { // true every 32 blocks (4,096 docs)
+      docOut.writeVInt(docID - lastSkipDocID);
```
mikemccand, line 384: Maybe rename `lastSkipDocID` to `lastLevel1DocID`?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsWriter.java
```diff
@@ -0,0 +1,597 @@
...
+      spareOutput.writeVLong(blockOutput.size());
+      numSkipBytes += spareOutput.size();
+      skipOutput.writeVLong(numSkipBytes);
+      spareOutput.copyTo(skipOutput);
+      spareOutput.reset();
+    }
+
+    blockOutput.copyTo(skipOutput);
+    blockOutput.reset();
+    lastBlockDocID = docID;
```
mikemccand, line 377: Maybe rename `lastBlockDocID` to `lastLevel0DocID`?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsWriter.java
```diff
@@ -0,0 +1,597 @@
...
+      long numSkipBytes = blockOutput.size();
+      forDeltaUtil.encodeDeltas(docDeltaBuffer, blockOutput);
+      if (writeFreqs) {
+        pforUtil.encode(freqBuffer, blockOutput);
+      }
+
+      spareOutput.writeVInt(docID - lastBlockDocID);
+      spareOutput.writeVLong(blockOutput.size());
+      numSkipBytes += spareOutput.size();
+      skipOutput.writeVLong(numSkipBytes);
```
mikemccand, line 370: Maybe note to self: I get that `numSkipBytes` is how much skip data we are inserting before writing the block data (docs/freqs), but how come we don't also write `numBlockBytes` here?  It seems like at read time to quickly skip over this block we would need to know total bytes in the block, not just the size of the skip data?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -0,0 +1,2028 @@
...
+    final boolean indexHasOffsetsOrPayloads;
+
+    private int docFreq; // number of docs in this posting list
+    private long totalTermFreq; // sum of freqBuffer in this posting list (or docFreq when omitted)
+    private int blockUpto; // number of docs in or before the current block
+    private int doc; // doc we last read
+    private long accum; // accumulator for doc deltas
+
+    // level 0 skip data
+    private int lastDocInBlock;
```
mikemccand, line 316: Maybe `nextLastLevel0SkipDoc` or so?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -0,0 +1,2028 @@
...
+    // level 1 skip data
+    private int nextSkipDoc;
```
mikemccand, line 318: Maybe `nextLastLevel1SkipDoc` or so?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -0,0 +1,2028 @@
...
+        }
+      }
+    }
+
+    private void skipLevel0To(int target) throws IOException {
+      while (true) {
+        accum = lastDocInBlock;
+        if (docFreq - blockUpto >= BLOCK_SIZE) {
+          final long skip0Len = docIn.readVLong(); // skip len
+          final long skip0End = docIn.getFilePointer() + skip0Len;
```
mikemccand, line 505: No need for these `final`s?  `javac` infers this these days?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -0,0 +1,2028 @@
...
+          break;
+        }
+      }
+    }
+
+    private void skipLevel0To(int target) throws IOException {
+      while (true) {
+        accum = lastDocInBlock;
+        if (docFreq - blockUpto >= BLOCK_SIZE) {
+          final long skip0Len = docIn.readVLong(); // skip len
```
mikemccand, line 504: Maybe `skip0NumBytes`?

Note to self: I still remain confused because we seem to be using the `skip0Len` here as if it includes all the bytes in the block (including the encoded doc-delta/freq), but at write time we seem to only include the level 0 skip byte count.

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -0,0 +1,2028 @@
...
+          final long skip0End = docIn.getFilePointer() + skip0Len;
+          int docDelta = docIn.readVInt();
+          lastDocInBlock += docDelta;
+
+          if (target <= lastDocInBlock) {
+            docIn.seek(skip0End);
```
mikemccand, line 510: OK this one makes sense to me, since `skip0End` is end of skip data and start of real block data.

At this point, that `docIn.seek` is just skipping over one vLong (the number of bytes for the true block data (doc deltas, freqs))?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -0,0 +1,2028 @@
...
+            break;
+          }
+
+          // skip block
+          docIn.skipBytes(docIn.readVLong());
+          blockUpto += BLOCK_SIZE;
```
mikemccand, line 516: Maybe rename `blockUpto` to `docCountUpto`?  It's not really counting blocks but rather docs.

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -0,0 +1,2028 @@
...
+        lastDocInBlock = nextSkipDoc;
+        docIn.seek(nextSkipDocFP);
+        nextBlockPosFP = nextSkipPosFP;
+        nextBlockPosUpto = nextSkipPosUpto;
+        if (indexHasOffsetsOrPayloads) {
+          nextBlockPayFP = nextSkipPayFP;
+          nextBlockPayUpto = nextSkipPayUpto;
+        }
+        blockUpto = nextSkipBlockUpto;
+        nextSkipBlockUpto += SKIP_TOTAL_SIZE;
```
mikemccand, line 812: Maybe rename `SKIP_TOTAL_SIZE` to `SKIP_TOTAL_NUM_DOCS`?

And `nextSkipBlockUpto` to `nextSkipDocCount`?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -0,0 +1,2028 @@
...
+        long skipLength = docIn.readVLong();
+        nextSkipPosFP += docIn.readVLong();
+        nextSkipPosUpto = docIn.readVInt();
+        if (indexHasOffsetsOrPayloads) {
+          nextSkipPayFP += docIn.readVLong();
+          nextSkipPayUpto = docIn.readVInt();
+        }
+        nextSkipDocFP = docIn.getFilePointer() + impactLength + skipLength;
+
+        if (nextSkipDoc >= target) {
```
mikemccand, line 830: The equal case here means the target is `==` the very last doc in the L1 skip block right?  So we must stop our L1 skipping, switch to doing 31 L0 skips, then scan all the way to the last (127th) doc in that last block?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsFormat.java
```diff
@@ -0,0 +1,473 @@
...
+  public static final String PAY_EXTENSION = "pay";
+
+  /** Size of blocks. */
+  public static final int BLOCK_SIZE = ForUtil.BLOCK_SIZE;
+
+  public static final int BLOCK_SIZE_LOG2 = ForUtil.BLOCK_SIZE_LOG2;
+  public static final int BLOCK_MASK = BLOCK_SIZE - 1;
+
+  /** We insert skip data on every block and every SKIP_FACTOR=32 blocks. */
+  public static final int SKIP_FACTOR = 32;
```
mikemccand, line 346: Maybe rename to `LEVEL1_SKIP_FACTOR`?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsWriter.java
mikemccand, line 384: Maybe factor out the body of this `if` into `writeLevel1SkipData` or so?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsWriter.java
```diff
@@ -0,0 +1,597 @@
...
+
+    if ((docCount & SKIP_MASK) == 0) { // true every 32 blocks (4,096 docs)
+      docOut.writeVInt(docID - lastSkipDocID);
+      if (writeFreqs) {
+        writeImpacts(skipCompetitiveFreqNormAccumulator.getCompetitiveFreqNormPairs(), spareOutput);
+        docOut.writeVLong(spareOutput.size());
+      }
+      docOut.writeVLong(skipOutput.size());
+      if (writePositions) {
+        docOut.writeVLong(posOut.getFilePointer() - lastSkipPosFP);
```
mikemccand, line 391: Maybe rename `lastSkipX` to `lastLevel1SkipX` or so?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsWriter.java
mikemccand: Maybe rename `spareOutput` to `scratchOutput`?  I think of `scratch` as more strongly explaining that it is used solely in the context of one method to hold bytes and then move them elsewhere and reset right away.

Whereas the other two buffers here accumulate bytes over multiple method invocations...

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -0,0 +1,2028 @@
...
+        } else {
+          lastDocInBlock = DocIdSetIterator.NO_MORE_DOCS;
+          break;
+        }
+      }
+    }
+
+    private void moveToNextBlock() throws IOException {
+      if (doc >= nextSkipDoc) { // advance skip data on level 1
+        skipLevel1To(doc + 1);
```
mikemccand, line 526: I guess this is needed for when we `nextDoc()` over a level 1 skip data blob?

Would `doc` ever really be `> nextSkipDoc`?

You might eek a little performance if you specialized this case to `readNextLevel1SkipBlob` or so, but likely not worth the code dup micro fork.  I.e. in this situation we are not advancing to a future doc, just reading/decoding the next level 1 blob.

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -0,0 +1,2028 @@
...
+      } else {
+        lastDocInBlock = DocIdSetIterator.NO_MORE_DOCS;
+        refillRemainder();
+      }
+    }
+
+    @Override
+    public int nextDoc() throws IOException {
+      if (doc >= lastDocInBlock) { // advance skip data on level 0
+        moveToNextBlock();
```
mikemccand, line 543: Would `doc` ever really be `> lastDocInBlock`?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -0,0 +1,2028 @@
...
+    final boolean indexHasPos;
+    final boolean indexHasOffsets;
+    final boolean indexHasPayloads;
+    final boolean indexHasOffsetsOrPayloads;
+
+    private int docFreq; // number of docs in this posting list
+    private long totalTermFreq; // sum of freqBuffer in this posting list (or docFreq when omitted)
+    private int blockUpto; // number of docs in or before the current block
+    private int doc; // doc we last read
+    private long accum; // accumulator for doc deltas
```
mikemccand, line 1526: Maybe rename to `firstDocIDInBlock` or so?  This is sort of the "base" docID for the whole block?

mikemccand (commented):
Every time I peel the onion a bit more and understand this change in more detail I get more excited!  Heh.

Now I understand that we inline skip data before every block (not just every 32, which would be hard because at read-time it'd require holding state about where the reader is in that big skip data even as it decodes the blocks beyond it, too similar to the current multi-level skip reader).  If the block number is `0 mod SKIP_FACTOR`, we insert a level 1 skip block (and level 0 skip data after that).

Some of my comments are head scratching / talking to myself as I gradually understood things -- please ignore ;)  I also left some naming suggestions -- I don't feel super strongly about any of them, but the names would've helped my fresh (ish) eyes grok faster I think.

Besides updating the `Format` docs to explain this new skip layout, what else is making this still WIP?  It seems like it's close?  Exciting!

Thanks @jpountz.

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsWriter.java
mikemccand: Answer to self is 2 lines above heh ;)  On line 368 we also write the number of bytes taken by the actual block data (doc deltas, freqs), which is used at read-time to fully skip the block, not just the skip data preceding the block.

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
jpountz: It's only the number of bytes for skip data, so there are two cases:
 - If this block is a match, we'll skip only over the rest of skip data (e.g. position deltas)
 - If this block is not a match, we'll skip over not only the rest of skip data but also the block.

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
jpountz: If the index options are DOCS_ONLY then indeed. Otherwise, this would also skip over position skip data (pos delta, pos block upto, pay delta, pay block upto).

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
jpountz: This is correct.

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
jpountz: Indeed. I did not specialize because it only occurs once every 4,096 docs, so it didn't feel worth the code specialization. I'm only doing it on level 0 (`moveToNextBlock()`).

jpountz:
Thanks @mikemccand for taking a look at this large PR! I think I applied all your suggestions. The format docs should be up-to-date wrt how skip data is stored, and I did the codec dance (moving Lucene99PostingsFormat and Lucene99Codec to lucene/backward-codecs, making Lucene99PostingsFormat read-only, adding a test-only Lucene99RWPostingsFormat for testing, moving all our special terms dictionary formats to Lucene912PostingsFormat, moving the completion postings format to Lucene912PostingsFormat).

Note that I removed `TestLazyProxSkipping`, which assumed separate skip data and postings.

I plan on merging soon, as this PR is now in a state where conflicts easily arise and are hard to address. One cannot run luceneutil anymore as luceneutil complains that `Lucene99Codec` is missing, but my last run before doing the codec dance was very similar to the last one shared above.

mikemccand:
> Note that I removed `TestLazyProxSkipping`, which assumed separate skip data and postings.

YAY!

> I plan on merging soon, as this PR is now in a state where conflicts easily arise and are hard to address.

+1 to merge now.  This is an amazing improvement.  Thanks @jpountz!

dsmiley:
I'm looking at the performance results in the description.  The bottom end (performance improvements) look really nice, but the top end (performance regressions) look even worse.  Am I reading this wrong?

jpountz:
Things got a bit better later on (https://github.com/apache/lucene/pull/13585#issuecomment-2246112137), but your reading is correct that some queries get slower. This seems to especially affect queries that rely on performance of `BlockDocsEnum#nextDoc()` such as disjunction counts and queries sorted by field. This is because `BlockDocsEnum#nextDoc()` could previously completely ignore skip data, since it was stored separately, while it now has to skip over it by reading a vInt encoding the number of bytes for skip data and then calling `IndexInput#skipBytes` over this number of bytes. `CountOrHighHigh` runs are particularly noisy on my machine, apparently because my JVM makes different inlining decision on every run, I'm looking forward to nightly benchmarks getting published to get a better estimation of the slowdown.

mikemccand:
Nice pop in the nightly benchmarks from this!  [`OrHighMedium`](https://home.apache.org/~mikemccand/lucenebench/OrHighMed.html) jumped.  Even [`Phrase`](https://home.apache.org/~mikemccand/lucenebench/Phrase.html) jumped a bit.

And the worst case query above ([`CountOrHighHigh`](https://home.apache.org/~mikemccand/lucenebench/CountOrHighHigh.html)) lost only a bit, ~4-5%.

jpountz:
Hmm, [`CombinedHighHigh`](https://home.apache.org/~mikemccand/lucenebench/CombinedHighHigh.html) is angry. I had not benchmarked it while developping, I'll check it out.

Some speedups are spectacular though, e.g. [`CountAndHighMed`](https://home.apache.org/~mikemccand/lucenebench/CountAndHighMed.html). And some hard queries like [`CountPhrase`](https://home.apache.org/~mikemccand/lucenebench/CountPhrase.html) had a good speedup too, though it may have more to do with the fact that we no longer decompress the same block of positions multiple times than with the inlining of postings.

jpountz:
I found the problem with `CombinedHighHigh`, the logic for lazily decoding frequencies was broken and we'd decode the whole block of frequencies on every freq() calls. It's now fixed so benchmarks should look better tomorrow.

mikemccand:
Phew, thanks for catching the performance regression and tracking it down @jpountz.  GO BENCHMARKING!

HoustonPutman:
It looks like grouping queries are really affected by this change. The throughput of each of them were halved:
[100 groups](https://home.apache.org/~mikemccand/lucenebench/TermGroup100.html), [10K groups](https://home.apache.org/~mikemccand/lucenebench/TermGroup10K.html), [1M groups](https://home.apache.org/~mikemccand/lucenebench/TermGroup1M.html), [1M groups (two pass block grouping)](https://home.apache.org/~mikemccand/lucenebench/TermBGroup1M.html), [1M groups (single pass block grouping)](https://home.apache.org/~mikemccand/lucenebench/TermBGroup1M1P.html)

jpountz:
@HoustonPutman This is the same issue as reported above: the logic for lazily decoding blocks of freqs was broken and would decompress whole blocks of freqs on every doc ID. It is now fixed as the benchmarks that you linked confirm (requires zooming in).

HoustonPutman:
Thanks for the correction, sorry for the noise!

PR #13586, opened 2024-07-18
Feature/vector io prefetch

benwtrent:
I am trying out some prefetching for vector search and HNSW.

This right now is a dead-simple version that simply prefetches the next neighbor we will explore.

I will respond with some benchmarking.

Opening as Draft for some initial discussion.

### DIFF
--- a/lucene/codecs/src/java/org/apache/lucene/codecs/bitvectors/FlatBitVectorsScorer.java
+++ b/lucene/codecs/src/java/org/apache/lucene/codecs/bitvectors/FlatBitVectorsScorer.java
@@ -81,6 +81,11 @@ public class FlatBitVectorsScorer implements FlatVectorsScorer {
     }
 
     @Override
+    public void prepareToScore(int node) throws IOException {
+      vectorValues.prefetchOrdinal(node);
+    }
+
+    @Override
     public int maxOrd() {
       return vectorValues.size();
     }
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene95/OffHeapByteVectorValues.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene95/OffHeapByteVectorValues.java
@@ -84,6 +84,11 @@ public abstract class OffHeapByteVectorValues extends ByteVectorValues
   }
 
   @Override
+  public void prefetchOrdinal(int ordinal) throws IOException {
+    slice.prefetch((long) ordinal * byteSize, byteSize);
+  }
+
+  @Override
   public IndexInput getSlice() {
     return slice;
   }
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene95/OffHeapFloatVectorValues.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene95/OffHeapFloatVectorValues.java
@@ -87,6 +87,11 @@ public abstract class OffHeapFloatVectorValues extends FloatVectorValues
     return value;
   }
 
+  @Override
+  public void prefetchOrdinal(int ordinal) throws IOException {
+    slice.prefetch((long) ordinal * byteSize, byteSize);
+  }
+
   public static OffHeapFloatVectorValues load(
       VectorSimilarityFunction vectorSimilarityFunction,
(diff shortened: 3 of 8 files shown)


benwtrent:
🤔 my benchmarking is suspicious. I wonder if I am doing something wrong. 

I have a 4GB index, on a 4GB machine, 1GB set aside for the JVM. So, QPS should be about the same.

However, I am getting far worse numbers and some weird artifacts (this is 5 runs in a row with the exact same parameters). 

baseline:

```
completed 1000 searches in 4776 ms: 209 QPS CPU time=4741ms
completed 1000 searches in 4923 ms: 203 QPS CPU time=4904ms
completed 1000 searches in 4557 ms: 219 QPS CPU time=4533ms
completed 1000 searches in 4667 ms: 214 QPS CPU time=4655ms
completed 1000 searches in 4702 ms: 212 QPS CPU time=4700ms
```

Candidate:
```
completed 1000 searches in 5151 ms: 194 QPS CPU time=5106ms
completed 1000 searches in 5383 ms: 185 QPS CPU time=5371ms
completed 1000 searches in 8719 ms: 114 QPS CPU time=5894ms
completed 1000 searches in 7343 ms: 136 QPS CPU time=5554ms
completed 1000 searches in 5285 ms: 189 QPS CPU time=5272ms
```

I can restrict memory even more to try and force a significant difference, but if this code is so much worse even when everything already fits in memory, something is wrong.

jpountz:
Thanks for looking into this! It's disappointing that this small change degrades performance so much indeed! I'm curious if you are able to run your benchmark under a profiler to confirm if the bottkeneck is the call to madvise or something silly.

> I can restrict memory even more to try and force a significant difference, but if this code is so much worse even when everything already fits in memory, something is wrong.

Right now prefetch() has logic to disable itself if data is consistently already in the page cache. Thinking out loud, if your benchmark shows that prefetch() hurts more than it helps when the data is often but not always in the page cache, we could look into disabling madvise in that case too and only enable it when data is rarely already in the page cache.

benwtrent:
I ran a bunch more benchmarks, the numbers really jump around for prefetch, but are way more consistent for baseline. 

Here are two JFRs, but they aren't very enlightening. 
[Archive.zip](https://github.com/user-attachments/files/16338780/Archive.zip)

jpountz:
Thanks @benwtrent, not very enlightening indeed. I wonder what benchmark you ran in case I can reproduce it and play with it?

benwtrent:
@jpountz I build an index with ~1M CohereV3 floating point vectors (this requires about ~4GB of ram), force merged into a single segment, and benchmarked on `e2-medium` (4GB of ram) with 1GB set aside for the heap. 

To download the vectors, I used a script like the following:
```bash
#!/bin/sh

base_url="https://huggingface.co/api/datasets/Cohere/wikipedia-2023-11-embed-multilingual-v3/parquet/en/train/"

for i in {0..10}
do
    url="${base_url}${i}.parquet"
    output_file="${i}-en.parquet"
    echo "Downloading: $url"
    curl -L "$url" -o "$output_file" &
done
wait
```

Then to concat into a single file that can be used by Lucene Util:

```python
import pyarrow.parquet as pq
import numpy as np

name = "wiki1024en"
tbs = [pq.read_table(f"{x}-en.parquet", columns=['emb']) for x in range(11)]
nps = [tb[0].to_numpy() for tb in tbs]
np_total = np.concatenate(nps)

flat_ds = list()
for vec in np_total:
    flat_ds.append(vec)
np_flat_ds = np.array(flat_ds)
print(f"{np_flat_ds.shape}")
row_count = np_flat_ds.shape[0]
dims = np_flat_ds.shape[1]
query_count = 1_000
training_rows = row_count - query_count
print(f"{name} num rows: {training_rows}")

with open(f"{name}.test", "wb") as out_f:
    np_flat_ds[training_rows:-1].tofile(out_f)

with open(f"{name}.train", "wb") as out_f:
    np_flat_ds[0:training_rows].tofile(out_f)
```

benwtrent:
I think this logic will completely change now that we are doing the bulk scoring. We should pre-fetch the bulk scoring vectors.

PR #13587, opened 2024-07-18
Use Max WAND optimizations with ToParentBlockJoinQuery when using ScoreMode.Max

Mikep86:
Updates `ToParentBlockJoinQuery` to propagate the min competitive score when using `ScoreMode.Max`


lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -101,12 +99,7 @@ public Weight createWeight(
               .rewrite(new ConstantScoreQuery(childQuery))
               .createWeight(searcher, weightScoreMode, 0f);
     } else {
-      // if the score is needed we force the collection mode to COMPLETE because the child query
-      // cannot skip
-      // non-competitive documents.
-      childWeight =
-          childQuery.createWeight(
-              searcher, weightScoreMode.needsScores() ? COMPLETE : weightScoreMode, boost);
+      childWeight = childQuery.createWeight(searcher, weightScoreMode, boost);
```
jpountz, line 102: Should we keep using COMPLETE if the score mode is not max since we wouldn't be propagating the min competitive score?

lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinScorer.java
```diff
@@ -118,4 +110,108 @@ public void testScoreNone() throws IOException {
...
+                new String[] {"A", "B", "C"},
+                new String[] {"B"},
+                new String[] {"B", "C"})) {
+          List<Document> docs = new ArrayList<>();
+
+          Document childDoc = new Document();
+          childDoc.add(newStringField("type", "child", Field.Store.NO));
+          for (String value : values) {
+            childDoc.add(newStringField("value", value, Field.Store.NO));
+          }
```
jpountz, line 139: The test would be more interesting if there were multiple children per parent that produce different scores. Here you are always creating a single child document per parent.

lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinScorer.java
```diff
@@ -118,4 +110,108 @@ public void testScoreNone() throws IOException {
...
+                .setMinimumNumberShouldMatch(2)
+                .build();
+        BitSetProducer parentsFilter =
+            new QueryBitSetProducer(new TermQuery(new Term("type", "parent")));
+        ToParentBlockJoinQuery parentQuery =
+            new ToParentBlockJoinQuery(childQuery, parentsFilter, ScoreMode.Max);
+
+        Weight weight = searcher.createWeight(searcher.rewrite(parentQuery), TOP_SCORES, 1);
+        ScorerSupplier ss = weight.scorerSupplier(searcher.getIndexReader().leaves().get(0));
+        ss.setTopLevelScoringClause();
```
jpountz, line 177: We probably need to implement this method on BlockJoinScorer to propagate to the inner clause when the score mode is max.

jpountz (commented):
Thanks for looking into it, the idea makes sense to me: if the score of the parent query must be at least X and the score of a parent doc is the max of the scores of its children, then it's ok if we filter children whose score is at least X as well.

I left a few inline comments.

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
Mikep86: My logic with this change was that it's simpler to pass `weightScoreMode` directly because the child query can now skip non-competitive documents. I thought about forcing the collection mode to `COMPLETE` when `weightScoreMode.isExhaustive == true`, but that only happens when `weightScoreMode` is either `COMPLETE` or `COMPLETE_NO_SCORES`.

Are there larger side-effects to always passing through `weightScoreMode` directly? We could force the collection mode to `COMPLETE` except when `weightScoreMode.isExhaustive == false` and `scoreMode == ScoreMode.Max`, but I didn't do that initially because I thought that could miss edge cases. WDYT?

jpountz: > Are there larger side-effects to always passing through weightScoreMode directly?

It's complicated, but the short answer is that if you pass TOP_SCORES, then you get a scorer that knows how to skip hits based on scores, but has more overhead at iterating documents than if you had passed COMPLETE. So the idea before your change was that since we don't take advantage of `setMinCompetitiveScore` anyway, then let's use the COMPLETE score mode to be more efficient.

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -101,12 +101,16 @@ public Weight createWeight(
...
-      // cannot skip
-      // non-competitive documents.
+      // if the score is needed and the score mode is not max, we force the collection mode to
+      // COMPLETE because the
+      // child query cannot skip non-competitive documents.
       childWeight =
           childQuery.createWeight(
-              searcher, weightScoreMode.needsScores() ? COMPLETE : weightScoreMode, boost);
+              searcher,
+              weightScoreMode.needsScores() && childScoreMode != ScoreMode.Max
```
Mikep86, line 110: Technically we don't need to check if `weightScoreMode.needsScores()` is true here, it always will be if we get to this point. I left it because it makes the logic clearer, LMKWYT!

jpountz: this sounds good to me, maybe add a comment about it?

lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinScorer.java
```diff
@@ -118,4 +110,114 @@ public void testScoreNone() throws IOException {
...
+                    BooleanClause.Occur.SHOULD)
+                .add(
+                    new BoostQuery(
+                        new ConstantScoreQuery(new TermQuery(new Term("value", "C"))), 3),
+                    BooleanClause.Occur.SHOULD)
+                .add(
+                    new BoostQuery(
+                        new ConstantScoreQuery(new TermQuery(new Term("value", "D"))), 4),
+                    BooleanClause.Occur.SHOULD)
+                .setMinimumNumberShouldMatch(2)
```
jpountz, line 174: I wonder if you initially set it because the produced scorer would otherwise not skip properly? If this was the reason, the fact that you now propagate `setTopLevelScoringClause` should hopefully fix the root cause without requiring a min should match to be configured.

lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinScorer.java
```diff
@@ -16,23 +16,16 @@
...
-import org.apache.lucene.search.IndexSearcher;
-import org.apache.lucene.search.MatchAllDocsQuery;
-import org.apache.lucene.search.Query;
-import org.apache.lucene.search.ScoreMode;
-import org.apache.lucene.search.Scorer;
-import org.apache.lucene.search.ScorerSupplier;
-import org.apache.lucene.search.TermQuery;
-import org.apache.lucene.search.Weight;
+import org.apache.lucene.index.*;
+import org.apache.lucene.search.*;
```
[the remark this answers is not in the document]
Mikep86: Will revert, my IDE did that automatically

lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinScorer.java
Mikep86: This is just a holdover from the test I used as a template when writing this. I'll remove it.

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -101,12 +101,18 @@ public Weight createWeight(
...
               .createWeight(searcher, weightScoreMode, 0f);
     } else {
-      // if the score is needed we force the collection mode to COMPLETE because the child query
-      // cannot skip
-      // non-competitive documents.
+      // if the score is needed and the score mode is not max, we force the collection mode to
+      // COMPLETE because the
+      // child query cannot skip non-competitive documents.
+      // weightScoreMode.needsScores() will always be true here, but keep the check to make the
+      // logic clearer.
```
jpountz, line 108: I know it's a pre-existing issue, but can you fix how lines are broken?

jpountz (approved):
This looks great. Can you add an entry to lucene/CHANGES.txt?

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
Mikep86: I tried fixing this manually, but then the spotless checks fail. How do you suggest I fix this in a way that doesn't break those checks?

jpountz: ususally I remove the unintended line breaks and then run `gradlew tidy`

Mikep86: Got it, thank you :)

PR #13588, opened 2024-07-18
Avoid wrap readers without soft-deletes

dnhatn:
I analyzed a heap dump of Elasticsearch where FixedBitSet uses more than 1GB of memory. Most of these FixedBitSets are used by soft-deletes reader wrappers, even though these segments have no deletes at all. I believe these segments previously had soft-deletes, but these deletes were pruned by merges. The reason we wrap soft-deletes is that the soft-deletes field exists. Since these segments had soft-deletes previously, we carried the field-infos into the new segment. Ideally, we should have ways to check whether the returned docValues iterator is empty or not so that we can avoid allocating FixedBitSet completely, or we should prune fields without values after merges.


jpountz (approved):
It makes sense to me but I'm curious how this came to your attention?

dnhatn:
wrapper

dnhatn:
@jpountz Thanks for reviewing! I am about to write the description and have updated it.

PR #13557, opened 2024-07-10
Check whether liveDoc is null out of loop in Weight.scoreAll

vsop-479:
### Description

I am not sure whether it is worth to do that, or whether `JIT` will do similar optimization.

PR #13589, opened 2024-07-19
Remove useless todo in PerThreadPKLookup.

vsop-479:
### Description

No difference, see https://github.com/apache/lucene/pull/13557.


jpountz:
Thank for closing the loop.

PR #13592, opened 2024-07-19
Take advantage of the doc value skipper when it is primary sort

iverase:
Similar to what we do in `IndexSortSortedNumericDocValuesRangeQuery` for primary sort, we can do the same in `SortedSetDocValuesRangeQuery` and `SortedNumericDocValuesRangeQuery` when it is primary sort and there are not delete documents.
We find the first and last document using the skipper and return the dense iterator.


jpountz (commented):
I wonder if we could simplify this logic by adding helper methods on `DocValuesSkipper`. In my mind, the primary-sort optimization wouldn't need to know if the sort is reverse or not, it would first advance to the first block that intersects with the range, from them iterate documents until it finds the first matching doc, then it would skip all blocks that are fully contained in the range, then it would iterate documents until it finds the first non-matching doc?

iverase:
I tried to find a generic method that could help here but I think the logic relies too much on the fact that the index is sorted. 

For example, your mental model somewhat breaks if the range lies between too blocks so there is no intersecting block. In that case you would need to read the full index just to check there is no intersecting block.

iverase:
I introduced the method DocValuesSkipper#advance(long,long) to advance the iterator to a matching range. wdyt?

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
```diff
@@ -98,4 +98,29 @@ public abstract class DocValuesSkipper {
...
+   * <p><b>NOTE</b>: The behavior is undefined if this method is called and {@link #advance(int)}
+   * has not been called yet.
+   */
+  public final void advance(long minValue, long maxValue) throws IOException {
+    while (true) {
+      if (minDocID(0) == DocIdSetIterator.NO_MORE_DOCS
+          || (minValue(0) <= maxValue && maxValue(0) >= minValue)) {
+        break;
+      } else {
+        int maxDocID = maxDocID(0);
```
gsmiller, line 115: This might just be me, but I got really confused for a while until I realized that `maxDocID` does not necessarily return the docID associated with `maxValue` (in the standard ascending sort case it would, but it's reversed in the reverse sort case... assuming I'm getting this right). I had the same confusing back in the calling code in the reverse-sort case. Would you mind adding a couple comments to this code to make that a little more clear in case someone else suffers from my same confusion?

gsmiller (approved):
This looks good to me. Left one request for a couple code comments for people like me that get confused easily :)

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
```diff
@@ -98,4 +98,29 @@ public abstract class DocValuesSkipper {
...
+
+  /**
+   * Advance this skipper so that all levels intersects the range given by {@code minValue} and
+   * {@code maxValue}. If there are no intersecting levels, the skipper is exhausted.
+   *
+   * <p><b>NOTE</b>: The behavior is undefined if this method is called and {@link #advance(int)}
+   * has not been called yet.
+   */
+  public final void advance(long minValue, long maxValue) throws IOException {
+    while (true) {
```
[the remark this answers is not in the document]
iverase: make sense, I wonder if we should actively advance the iterator if it hasn't been called yet.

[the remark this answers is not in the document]
gsmiller: Good point. I can't think of a good reason why we shouldn't do this? It would make the API a little easier for callers. I suppose it adds a very slight overhead to the call, but seems trivial.

lucene/core/src/test/org/apache/lucene/search/TestDocValuesQueries.java
```diff
@@ -42,34 +45,100 @@
...
+  public void testDuelPointRangeNumericRangeWithSkipperQuery() throws IOException {
+    doTestDuelPointRangeNumericRangeQuery(false, 1, true);
+  }
+
+  public void testDuelPointNumericSortedWithSkipperRangeQuery() throws IOException {
+    Directory dir = newDirectory();
+    IndexWriterConfig config = new IndexWriterConfig();
+    config.setIndexSort(new Sort(new SortField("dv", SortField.Type.LONG, random().nextBoolean())));
+    RandomIndexWriter iw = new RandomIndexWriter(random(), dir, config);
+    config.setCodec(getCodec());
```
jpountz, line 82: can you move the call to `setCodec` before the IndexWriter is created? since the codec is not supposed to be updateable (the setter is only on IndexWriterConfig, not LiveIndexWriterConfig), it would be legal for IndexWriter to cache the codec that it gets at creation time

jpountz (approved):
The logic looks good to me. One thing I'm not clear about is how this should interact with `IndexSortSortedNumericDocValuesRangeQuery`. Users might be tempted to wrap their doc values range queries with this class to take advantage of the index sort, but if they do that they will get the slow binary search on doc values from `IndexSortSortedNumericDocValuesRangeQuery` instead of this new code that takes advantage of the sparse index. Maybe we should look into pulling the optimizations from this class directly into our point/doc values queries to remove this trap (in follow-ups).

lucene/core/src/test/org/apache/lucene/search/TestDocValuesQueries.java
iverase: sure!

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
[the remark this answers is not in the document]
iverase: I have changed the API so we accept unpositioned skippers to call this method.

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
iverase: I added some comments which hopefully makes it more clear.

PR #13594, opened 2024-07-20
[KNN] Add comment and remove duplicate code

dungba88:
### Description

- Add comments to methods for easier to understand their behaviors
- Remove duplicate codes to createBitSet used in both the main top-k KNN query and vector similarity query
- Also remove unused variable `weight`

To discuss: The `createFilterWeight` might also merit to be shared with the vector similarity query. There is a discrepancy between the 2 classes: the top-k KNN query enhance the filter so that only documents having values on the KNN field will be matched, but the vector similarity query doesn't. I think they should share similar behavior as they are both vector search.

Moreover, should we make their behavior more closely so that they can share/extend one another? ~~The top-k KNN query has the main logic inside rewriteQuery while vector-similarity query has the main logic in Weight.~~ (I now understand that for top-k Query we need to get top-k across all segments and thus cannot just do it in createWeights, which only acts as segment-level matcher). Query timeout check is also not currently supported by vector similarity query.


kaivalnp:
+1 to share as much logic as possible (including `createFilterWeight`). The `FieldExistsQuery` proposal (to only collect pre-filtered docs which have vectors) seems promising too

> they can share/extend one another?

IMO we should not tie `AbstractVectorSimilarityQuery` and `AbstractKnnVectorQuery` because they compute results differently (in the scorer v/s rewrite) which has broader implications (eg the similarity query is cacheable, inherent parallelism across segments, etc)

Maybe some common KNN utility classes / functions instead?

dungba88:
I think common utility makes sense. I'll move both createFilterWeights and createBitSet to a utility class.

dungba88:
There were some recent commits I need to rebase first as well.

lucene/core/src/java/org/apache/lucene/search/KnnQueryUtils.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+package org.apache.lucene.search;
+
+import java.io.IOException;
+import org.apache.lucene.util.BitSet;
+import org.apache.lucene.util.BitSetIterator;
+import org.apache.lucene.util.Bits;
+
+/** Common utilities for KNN queries. */
+class KnnQueryUtils {
```
kaivalnp, line 25: Q: Should we name it something like `Ann*` (Approximate Near Neighbor *) to fit with `*VectorSimilarityQuery` too? (`Knn*` sounds specific to `Knn*VectorQuery`)

kaivalnp (approved):
LGTM overall, thanks!

lucene/core/src/java/org/apache/lucene/search/KnnQueryUtils.java
dungba88: That makes sense to me. I'll change it to AnnQueryUtils

dungba88:
Thanks Kaival for reviewing and approving.

Could someone from Lucene committers help review and merge this PR if it looks good?

lucene/core/src/java/org/apache/lucene/search/AnnQueryUtils.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+package org.apache.lucene.search;
+
+import java.io.IOException;
+import org.apache.lucene.util.BitSet;
+import org.apache.lucene.util.BitSetIterator;
+import org.apache.lucene.util.Bits;
+
+/** Common utilities for ANN queries. */
+class AnnQueryUtils {
```
benwtrent, line 25: this needs to be a `final` class

benwtrent: And can we name it `KnnQueryUtils` since all the queries are `Knn` queries?

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -53,8 +55,13 @@ abstract class AbstractKnnVectorQuery extends Query {
 
   private static final TopDocs NO_RESULTS = TopDocsCollector.EMPTY_TOPDOCS;
 
+  /** the KNN vector field to search */
   protected final String field;
+
+  /** the number of documents to find */
   protected final int k;
+
+  /** the filter to executed before KNN search */
```
benwtrent, line 64: While the filter is executed before search to gather the relevant bit set, we should be clear that when the filter is applied is up to the underlying knn index.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -116,6 +115,7 @@ private TopDocs searchLeaf(
     return results;
   }
 
+  // Execute the filter if any and perform KNN search at each segment.
```
benwtrent, line 118: this comment doesn't make sense to me. If we really want to add java docs to private methods, lets fix the wording. Maybe something like

"Perform kNN search for the provided LeafReaderContext applying filterWeight as necessary"

lucene/core/src/java/org/apache/lucene/search/AnnQueryUtils.java
dungba88: I'll change it to a final class.

It's also used by VectorSimilarityQuery which is not exactly KNN (k-nearest) though.

PR #13596, opened 2024-07-21
Add reopen method in PerThreadPKLookup

vsop-479:
### Description

Reuse loaded segments' `termsEnum` and `postingsEnum`.


vsop-479:
@jpountz 

lucene/test-framework/src/java/org/apache/lucene/tests/index/PerThreadPKLookup.java
```diff
@@ -97,5 +111,82 @@ public int lookup(BytesRef id) throws IOException {
...
 
-  // TODO: add reopen method to carry over re-used enums...?
+  /**
+   * Reuse loaded segments' termsEnum and postingsEnum.
+   *
+   * @return null if there are no changes; else, a new DirectoryReader instance which you must
+   *     eventually close.
+   * @throws IOException if there is a low-level IO error.
+   */
+  public DirectoryReader reopen() throws IOException {
```
jpountz, line 121: Can you make it return a new PerThreadPKLookup instance instead of reopening in-place?

vsop-479: I modified it.

vsop-479:
@jpountz 

vsop-479:
I can't figure out  what can we reuse from prior `PostingsEnum`.

If we want reuse prior `PostingsEnum`'s docBuffer( if it still stay in first block), we should not reset `docBufferUpto` to `BLOCK_SIZE` (maybe set to 0?) when prior docStartFP equals current's,  which will result in refillDocs.

For the situation of PK, every PK has different `PostingsEnum`, prior `PostingsEnum` is mostly useless if reuse means reuse same `PostingsEnum` loaded docBuffer.

@jpountz Please correct me if I am wrong.

lucene/test-framework/src/java/org/apache/lucene/tests/index/PerThreadPKLookup.java
```diff
@@ -37,16 +43,21 @@
...
-  protected final boolean hasDeletions;
+  private IndexReader reader;
+  private final String idFieldName;
+  protected TermsEnum[] termsEnums;
+  protected PostingsEnum[] postingsEnums;
+  protected Bits[] liveDocs;
+  protected int[] docBases;
+  protected int numSegs;
+  protected boolean hasDeletions;
+  protected Map<SegmentInfo, Integer> segmentInfoMap = new HashMap<>();
```
jpountz, line 54: Can you make all these fields final again?

lucene/test-framework/src/java/org/apache/lucene/tests/index/PerThreadPKLookup.java
```diff
@@ -97,5 +174,28 @@ public int lookup(BytesRef id) throws IOException {
...
-  // TODO: add reopen method to carry over re-used enums...?
+  /**
+   * Reuse loaded segments' termsEnum and postingsEnum.
+   *
+   * @return null if there are no changes; else, a new PerThreadPKLookup instance which you must
+   *     eventually close its reader by {@link #closeReader}.
+   * @throws IOException if there is a low-level IO error.
+   */
+  public PerThreadPKLookup reopen() throws IOException {
+    DirectoryReader newReader = DirectoryReader.openIfChanged((DirectoryReader) reader);
```
jpountz, line 185: After calling `openIfChanged`, the previous reader should get closed. IMO, the API should take a new reader and leave calling `openIfChanged` and `close` to the caller?

lucene/test-framework/src/java/org/apache/lucene/tests/index/PerThreadPKLookup.java
```diff
@@ -97,5 +174,28 @@ public int lookup(BytesRef id) throws IOException {
...
+    if (newReader == null) {
+      return null;
+    }
+
+    return new PerThreadPKLookup(this, newReader);
```
jpountz, line 190: This creates a new `PerThreadPKLookup` from scratch. The new reader will often have many leaves in common with the old reader (which can be identified through https://lucene.apache.org/core/9_8_0/core/org/apache/lucene/index/LeafReader.html#getCoreCacheHelper()). We should reuse TermsEnums on these common leaves.

lucene/test-framework/src/java/org/apache/lucene/tests/index/PerThreadPKLookup.java
```diff
@@ -97,5 +174,28 @@ public int lookup(BytesRef id) throws IOException {
...
+  }
+
+  /**
+   * Close reader.
+   *
+   * @throws IOException if there is a low-level IO error
+   */
+  public void closeReader() throws IOException {
+    reader.close();
+  }
```
jpountz, line 200: I would remove this method and make it a responsibility of the caller.

lucene/test-framework/src/java/org/apache/lucene/tests/index/PerThreadPKLookup.java
vsop-479: I added a `Map<SegmentInfo, Integer> segmentInfoMap` to record each segment's index in `termsEnums`, and apply reuse common TermsEnums by getting SegmentInfo‘s index when create a new `PerThreadPKLookup` by new reader, in current version.

Do you mean use `LeafReader#getCoreCacheHelper` to instead of `SegmentInfo` as the key?

jpountz: Argh sorry, for some reason I overlooked this logic. Indeed it would be better to use the result of `getCoreCacheHelper().getKey()` than the SegmentInfo object.

jpountz: It would be nice if we could factor the two constructors together, e.g. by having a private constructor take a `Map<CacheKey, PerSegmentCorePKLookup>` cache, where PerSegmentCorePKLookup would be a record around the various things we want to reuse across reopens (TermEnum and PostingsEnum), and that would be empty when called from the non-reusing constructor.

vsop-479: Thanks, I will try it.

vsop-479: We still need to record enums in an array, right? Since we sorted them by docs num, and visit them in desc order.

vsop-479: > we could factor the two constructors together, e.g. by having a private constructor take a Map<CacheKey, PerSegmentCorePKLookup> cache

I merged the two constructors into one, but didn't add a split `PerSegmentCorePKLookup` for cache. Since I only cached the index(Integer) of enums array, like this: `Map<IndexReader.CacheKey, Integer> enumIndexes`.

jpountz (approved):
I wanted to look at your PR locally to understand what it did. It looked good to me but I saw a few simplifications I wanted to make, such as not caching `null` TermsEnum, since returning `null` Terms is not expensive on a reader. I felt free to push to your branch, I hope you don't mind. The change looks good to me.

vsop-479:
This change looks clearer, Thanks @jpountz .

PR #13597, opened 2024-07-22
Align doc value skipper interval boundaries when an interval contains a constant value

iverase:
Doc values skipper currently creates a new interval every 4096 documents. This can be inefficient in the case that the field has low cardinality and values are sorted. In that case, we might want to align the boundaries for intervals that contains just a single value.

Therefore this change proposes to keep adding documents in an interval if the following conditions apply:

- The next document is single valued.
-  The interval only contains a single value and next value is the same value
- The interval is dense and the next document keeps the density.

In this situation we will keep adding documents into the interval until one of the conditions break it.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/codecs/lucene90/TestLucene90DocValuesFormatVariableSkipInterval.java
+++ b/lucene/core/src/test/org/apache/lucene/codecs/lucene90/TestLucene90DocValuesFormatVariableSkipInterval.java
@@ -16,11 +16,24 @@
  */
 package org.apache.lucene.codecs.lucene90;
 
+import java.io.IOException;
+import java.util.Arrays;
 import org.apache.lucene.codecs.Codec;
+import org.apache.lucene.document.Document;
+import org.apache.lucene.document.NumericDocValuesField;
+import org.apache.lucene.document.SortedNumericDocValuesField;
+import org.apache.lucene.index.DocValuesSkipper;
+import org.apache.lucene.index.IndexReader;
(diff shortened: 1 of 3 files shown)


lucene/core/src/test/org/apache/lucene/codecs/lucene90/TestLucene90DocValuesFormatVariableSkipInterval.java
```diff
@@ -36,4 +49,159 @@ public void testSkipIndexIntervalSize() {
             () -> new Lucene90DocValuesFormat(random().nextInt(Integer.MIN_VALUE, 2)));
     assertTrue(ex.getMessage().contains("skipIndexIntervalSize must be > 1"));
   }
+
+  public void testSkipperAllEqualValue() throws IOException {
+    final IndexWriterConfig config = new IndexWriterConfig().setCodec(getCodec());
+    try (Directory directory = newDirectory();
+        RandomIndexWriter writer = new RandomIndexWriter(random(), directory, config)) {
+      int numDocs = random().nextInt(100);
```
jpountz, line 57: did you mean atLeast(100)?

jpountz (approved):
The change looks good, nice testing.

lucene/core/src/test/org/apache/lucene/codecs/lucene90/TestLucene90DocValuesFormatVariableSkipInterval.java
iverase: yes

PR #13602, opened 2024-07-23
Udpate ReadTask to not rely on search(Query, Collector)

javanna:
This commit modifies ReadTask to no longer call the deprecated search(Query, Collector). Instead, it creates a collector manager and calls search(Query, CollectorManager).

The existing protected createCollector method is removed in favour of createCollectorManager that returns now a CollectorManager in place of a Collector.

SearchWithCollectorTask works the same way if "topScoreDoc" is provided as config. Loading of a custom collector will no longer work, and needs to be replaced with loading a collector manager by class name instead.

### DIFF
--- a/lucene/benchmark/conf/collector.alg
+++ b/lucene/benchmark/conf/collector.alg
@@ -17,11 +17,10 @@
 # -------------------------------------------------------------------------------------
 # multi val params are iterated by NewRound's, added to reports, start with column name.
 
-# collector.class can be:
-#    Fully Qualified Class Name of a Collector with a empty constructor
-#    topScoreDocOrdered - Creates a TopScoreDocCollector that requires in order docs
-#    topScoreDocUnordered - Like above, but allows out of order
-collector.class=coll:topScoreDoc
+# collector.manager.class can be:
+#    Fully Qualified Class Name of a CollectorManager with a empty constructor
+#    topScoreDoc - Creates a TopScoreDocCollectorManager
+collector.manager.class=coll:topScoreDoc
 
 analyzer=org.apache.lucene.analysis.core.WhitespaceAnalyzer
 directory=FSDirectory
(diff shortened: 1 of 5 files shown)


lucene/benchmark/src/java/org/apache/lucene/benchmark/byTask/tasks/SearchWithCollectorTask.java
```diff
@@ -46,17 +46,17 @@ public boolean withCollector() {
...
-      collector = Class.forName(clnName).asSubclass(Collector.class).getConstructor().newInstance();
-
+      collectorManager =
+          Class.forName(clnName).asSubclass(CollectorManager.class).getConstructor().newInstance();
     } else {
-      collector = super.createCollector();
+      collectorManager = super.createCollectorManager();
     }
-    return collector;
+    return collectorManager;
```
javanna, line 59: I am not familiar with this utility and I wonder if this type of change is acceptable for a major release, whether there's other ways. If we continue down this route, how do we document the breaking change, and should we deprecate createCollector in the 9x branch?

javanna: I have now added a changes entry, and replaced collector.class with a new config parameter. We throw error once collector.class is provided, so we fail fast and users know what they have to do.

lucene/benchmark/conf/collector.alg
```diff
@@ -17,11 +17,10 @@
...
 
-# collector.class can be:
-#    Fully Qualified Class Name of a Collector with a empty constructor
-#    topScoreDocOrdered - Creates a TopScoreDocCollector that requires in order docs
-#    topScoreDocUnordered - Like above, but allows out of order
-collector.class=coll:topScoreDoc
+# collector.manager.class can be:
+#    Fully Qualified Class Name of a CollectorManager with a empty constructor
+#    topScoreDoc - Creates a TopScoreDocCollectorManager
+collector.manager.class=coll:topScoreDoc
```
javanna, line 23: This looked like it needed updating following my changes, but I could not tell where these files are loaded, and what I should run to verify that my update works.

jpountz (approved):

PR #13604, opened 2024-07-23
KMeans clustering algorithm

mayya-sharipova:
Implement Kmeans clustering algorithm for vectors.

Knn algorithms that further reduce memory usage of vectors (such as 
Product Quantization, RaBitQ etc) require clustering of vectors. 
This implements KMeans clustering algorithm.


mayya-sharipova:
I did [benchmarking](https://github.com/mayya-sharipova/kmeans-test)  on MNIST dataset and compared the accuracy with [KMeans algorithm](https://github.com/mayya-sharipova/kmeans-test/blob/main/kmeans_cluster.py)  from `sklearn.cluster`.

| Clusters | Python accuracy| Java accuracy|
| ---: | --- :| --- :|
| 10  |     0.5712|   0.5855  |
| 16  |     0.6867|    0.6703 |
| 64  |    0.8356 |    0.8194 |
| 256 |    0.9123 |    0.9043 |

mikemccand:
Whoa, cool!  What is (roughly) the run-time of KMeans as a function of number of vectors?  Do you tell it how many clusters to create, or, do you ask it to keep splitting into more clusters until it reaches a target accuracy?

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
```diff
@@ -0,0 +1,344 @@
...
+import org.apache.lucene.util.VectorUtil;
+import org.apache.lucene.util.hnsw.RandomAccessVectorValues;
+
+/** KMeans clustering algorithm for vectors */
+public class KMeans {
+  public static final int DEFAULT_RESTARTS = 5;
+  public static final int DEFAULT_ITRS = 10;
+  public static final int DEFAULT_SAMPLE_SIZE = 100_000;
+
+  private final RandomAccessVectorValues.Floats vectors;
```
benwtrent, line 37: I don't think we should restrict this to only `float` vector values.

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
```diff
@@ -0,0 +1,344 @@
...
+    // Store distances of each point to the nearest centroid
+    float[] minDistances = new float[numVectors];
+    Arrays.fill(minDistances, Float.MAX_VALUE);
+
+    // Step 2 and 3: Select remaining centroids
+    for (int i = 1; i < numCentroids; i++) {
+      // Update distances with the new centroid
+      double totalSum = 0;
+      for (int j = 0; j < numVectors; j++) {
+        float dist = VectorUtil.squareDistance(vectors.vectorValue(j), initialCentroids[i - 1]);
```
benwtrent, line 221: I can understand the simplicity here. But, I know for at least `byte[]` vectors, we should be able to utilize a flat scorer or something so that we don't needlessly read vectors onto heap.

Maybe this is a future optimization. But always reading the sampled vectors onto heap seems unnecessary.

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
```diff
@@ -0,0 +1,344 @@
...
+    RESERVOIR_SAMPLING,
+    PLUS_PLUS
+  }
+
+  /**
+   * Results of KMeans clustering
+   *
+   * @param centroids the produced centroids
+   * @param vectorCentroids for each vector which centroid it belongs to, we use short type, as we
+   *     expect less than 32767 centroids. Can be {@null} if they were not computed.
```
benwtrent, line 341: is this `32767` limit enforced anywhere? Might be good to validate and throw.

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
benwtrent: Maybe thats a follow up, as it may take a bunch more work to figure out. But this should also work with byte vectors.

benwtrent (commented):
looking nice!

benwtrent:
@mikemccand 

The runtime depends on configured number of iterations, restarts, sample size, and cluster count. But, it can be very fast. I will leave mMayya to talk about some specific numbers. 

> Do you tell it how many clusters to create, or, do you ask it to keep splitting into more clusters until it reaches a target accuracy?

You tell it how many clusters you want. 

I could see us eventually doing a "KMeans-tree" type thing where we can split and combine clusters more efficiently, but having KMeans would be step one.

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
mayya-sharipova: @benwtrent Thanks for comment. How will this work? Would I need to create a separate `RandomVectorScorer` class for each centroid?

Also, in case when I need to access `vectorValue()`  outside of scoring as in `runKMeansStep` does it still make sense to use vector scorers>

mayya-sharipova:
@mikemccand Here are some numbers on my mac M3:

Doing 34 clusters with defaults (5 restarts, 10 inters each) on vectors of 1024 dims:

| N docs | Performance in seconds| 
| ---: |  ---: | 
| 131K  |     146|   
| 1M  |     347|  
| 9M  |    5119 |

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
benwtrent: @mayya-sharipova so, yes, we would need access to `vectorValue` directly to initialize the clusters. But, (this might be a future optimization), I think we would want something like `vectorScorer` that is provided via `FloatVectorValues`. 

It might be tricky as we require the similarity for the centroid calculation to be `squareDistance`. But initializing the clusters with the raw floating point values, and then calculating the distances via scalar quantized values would be powerful and utilize way less off-heap memory and might be significantly faster with little impact on quality.

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
```diff
@@ -0,0 +1,346 @@
...
+    double squaredDist = 0;
+    float[][] bestCentroids = null;
+
+    for (int restart = 0; restart < restarts; restart++) {
+      float[][] centroids =
+          switch (initializationMethod) {
+            case FORGY -> initializeForgy();
+            case RESERVOIR_SAMPLING -> initializeReservoirSampling();
+            case PLUS_PLUS -> initializePlusPlus();
+          };
```
tteofili, line 158: would it make sense to have a `default` (e.g. forgy)?

tteofili (approved):
this looks good, thanks @mayya-sharipova !

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
mayya-sharipova: @tteofili FORGY is the simplest and fastest, that's why was chosen by default.  PLUS_PLUS takes much more time, may be we can optimize it in the follow-up. What default would you suggest?

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
mayya-sharipova: Clarified in 44893adc5c6279bdf931482af98a6f7e1516320d

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
tteofili: I think defaulting to _FORGY_ is just fine, thanks Mayya ;-)

lucene/sandbox/src/test/org/apache/lucene/sandbox/codecs/quantization/TestKMeans.java
```diff
@@ -0,0 +1,92 @@
...
+      KMeans.Results results = KMeans.cluster(vectorValues, similarityFunction, nClusters);
+      assertEquals(nClusters, results.centroids().length);
+      assertEquals(nVectors, results.vectorCentroids().length);
+    }
+    // expert case
+    {
+      boolean assignCentroidsToVectors = random().nextBoolean();
+      int randIdx2 = random().nextInt(KMeans.KmeansInitializationMethod.values().length);
+      KMeans.KmeansInitializationMethod initializationMethod =
+          KMeans.KmeansInitializationMethod.values()[randIdx2];
```
john-wagster, line 49: Might be worth testing the other initialization methods as well?

john-wagster (approved):
LGTM!

ChrisHegarty (approved):
LGTM

lucene/sandbox/src/test/org/apache/lucene/sandbox/codecs/quantization/TestKMeans.java
mayya-sharipova: @john-wagster I was thinking since initialization method is chose at random, it will be exercised in other tests as well.

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
mayya-sharipova: @benwtrent Thanks for the comment.  I understood the need to use vector scorers. If you don't mind, we can do that in the follow up.  I've added TODO: items in the code.

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
```diff
@@ -0,0 +1,348 @@
...
+        int c = random.nextInt(numCentroids);
+        initialCentroids[c] = ArrayUtil.copyOfSubArray(vector, 0, vector.length);
+      }
+    }
+    return initialCentroids;
+  }
+
+  /** Initialize centroids using Kmeans++ method */
+  private float[][] initializePlusPlus() throws IOException {
+    float[][] initialCentroids = new float[numCentroids][];
```
tveasey, line 207: Same check as Forgy for paranoia.

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
```diff
@@ -0,0 +1,348 @@
...
+        float dist = VectorUtil.squareDistance(vectors.vectorValue(j), initialCentroids[i - 1]);
+        if (dist < minDistances[j]) {
+          minDistances[j] = dist;
+        }
+        totalSum += minDistances[j];
+      }
+
+      // Randomly select next centroid
+      double r = totalSum * random.nextDouble();
+      double cummulativeSum = 0;
```
tveasey, line 232: nit: typo
```suggestion
      double cumulativeSum = 0;
```

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
```diff
@@ -0,0 +1,348 @@
...
+      int nextCentroidIndex = -1;
+      for (int j = 0; j < numVectors; j++) {
+        cummulativeSum += minDistances[j];
+        if (cummulativeSum >= r) {
```
tveasey, line 236: I think there is an edge case here. `random.nextDouble()` could be 0 and the first vector could have been selected, but `0 >= 0` so this would return the same vector twice. I think you should probably do 
```suggestion
        if (cumulativeSum >= r && minDistances[j] > 0) {
```

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
```diff
@@ -0,0 +1,348 @@
...
+    return cluster(
+        vectors,
+        numClusters,
+        true,
+        42L,
+        KmeansInitializationMethod.PLUS_PLUS,
+        similarityFunction == VectorSimilarityFunction.COSINE,
+        DEFAULT_RESTARTS,
+        DEFAULT_ITRS,
+        DEFAULT_SAMPLE_SIZE);
```
tveasey, line 70: This should probably also account for `MAX_NUM_CENTROIDS`: 3 values vectors per cluster is too low. I would do something like `max(DEFAULT_SAMPLE_SIZE, 100 * numClusters)`. This extra constraint could be applied inside `cluster` though.

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
```diff
@@ -0,0 +1,348 @@
...
+
+    double sumSquaredDist = 0;
+    for (int docID = 0; docID < vectors.size(); docID++) {
+      float[] vector = vectors.vectorValue(docID);
+
+      short bestCentroid = 0;
+      if (numCentroids > 1) {
+        float minSquaredDist = Float.MAX_VALUE;
+        for (short c = 0; c < numCentroids; c++) {
+          // TODO: replace with RandomVectorScorer::score possible on quantized vectors
```
tveasey, line 286: Agreed, but we should make sure that we minimise Euclidean distance though.

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
```diff
@@ -0,0 +1,348 @@
...
+        for (int dim = 0; dim < newCentroids[c].length; dim++) {
+          centroids[c][dim] = newCentroids[c][dim] / newCentroidSize[c];
+        }
+      } else {
+        // this centroid did not get any points, assign a random data point to it
+        int rndIdx = random.nextInt(docCentroids.length);
+        float[] vector = vectors.vectorValue(rndIdx);
+        for (int dim = 0; dim < newCentroids[c].length; dim++) {
+          centroids[c][dim] = vector[dim];
+        }
```
tveasey, line 321: This should happen v.rarely given initialisation I would think. In such a case it would be better to assign outlying points I think: i.e. chose points by descending distance to the current centroid set (a bit like ++ initialisation).

tveasey (commented):
A few small things, but basically looking good!

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
```diff
@@ -0,0 +1,348 @@
...
+      float[][] centroids =
+          switch (initializationMethod) {
+            case FORGY -> initializeForgy();
+            case RESERVOIR_SAMPLING -> initializeReservoirSampling();
+            case PLUS_PLUS -> initializePlusPlus();
+          };
+
+      for (int iter = 0; iter < iters; iter++) {
+        squaredDist =
+            runKMeansStep(vectors, random, centroids, vectorCentroids, false, normalizeCenters);
```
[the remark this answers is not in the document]
mayya-sharipova: Addressed; added a check for `prevSquaredDist` with every iteration.

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
```diff
@@ -0,0 +1,348 @@
...
+    return bestCentroids;
+  }
+
+  /**
+   * Initialize centroids using Forgy method: randomly select numCentroids vectors for initial
+   * centroids
+   */
+  private float[][] initializeForgy() throws IOException {
+    Set<Integer> selection = new HashSet<>();
+    while (selection.size() < numCentroids) {
```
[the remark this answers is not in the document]
mayya-sharipova: Addressed, add adjustment of numClusters based on vectors.size

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
mayya-sharipova: Addressed, add adjustment of numClusters based on vectors.size

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
mayya-sharipova: addressed

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
mayya-sharipova: addressed

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
mayya-sharipova: addessed, added assignment of unassigned centroids based on descending distance to the current centroid set

mayya-sharipova:
@tveasey Thank you for your detailed feedback, it was addressed in the last commit.

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
benwtrent: Yeah, no doubt. Its an optimization, so this is fine.

tveasey (approved):
LGTM!

lucene/sandbox/src/java/org/apache/lucene/sandbox/codecs/quantization/KMeans.java
```diff
@@ -0,0 +1,394 @@
...
+          switch (initializationMethod) {
+            case FORGY -> initializeForgy();
+            case RESERVOIR_SAMPLING -> initializeReservoirSampling();
+            case PLUS_PLUS -> initializePlusPlus();
+          };
+      double prevSquaredDist = Double.MAX_VALUE;
+      for (int iter = 0; iter < iters; iter++) {
+        squaredDist = runKMeansStep(vectors, centroids, vectorCentroids, false, normalizeCenters);
+        // Check for convergence
+        if (prevSquaredDist < squaredDist) {
```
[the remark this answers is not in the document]
mayya-sharipova: Makes sense, @tveasey . I've also added a small epsilon for comparison: https://github.com/apache/lucene/pull/13604/commits/5ccd0ec2f8579ec80ed46dbd1ed67d2786bb1ce3

PR #13608, opened 2024-07-24
Save allocating some zero length byte arrays

original-brownbear:
Something I found in an ES heap dump. For large numbers of `FieldReader` where the minimum term is an empty string, we allocate MBs worth of empty `byte[]` for larger nodes. Worth adding the conditional here to save some heap I think, allocating empty instances is also somewhat common in other scenarios.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Lucene90BlockTreeTermsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Lucene90BlockTreeTermsReader.java
@@ -270,9 +270,8 @@ public final class Lucene90BlockTreeTermsReader extends FieldsProducer {
       throw new CorruptIndexException("invalid bytes length: " + numBytes, in);
     }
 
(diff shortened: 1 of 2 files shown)


uschindler (approved):

uschindler:
> Something I found in an ES heap dump. For large numbers of `FieldReader` where the minimum term is an empty string, we allocate MBs worth of empty `byte[]` for larger nodes. Worth adding the conditional here to save some heap I think, allocating empty instances is also somewhat common in other scenarios.

Normally the allocation of empty arrays is elided by the optimizer (under normal circumstances). So something like 'list.toArray(new array[0])` is optimized away and its basically a bad idea to allocate anything larger to fit list size.

The problem here is that those arrays are long-living because they are packed into a BytesRef and the compiler can't figure out at runtime in an easy way that it is useless to allocate it at all.

So I'd merge this and backport it to 9.x. Thanks.

uschindler:
Sorry I fogot the changes text!

uschindler:
Backported.

original-brownbear:
Thanks Uwe!

PR #13609, opened 2024-07-25
Release memory for cancelled tasks earlier in TaskExecutor

original-brownbear:
We can save some memory in failure scenarios here (and a tiny bit in every case) by moving our started flag to the `FutureTask` and using the callable outright. First of all, we save the wrapper callable, but that also allows us to just `set(null)` on cancellation instead of waiting for the task to run to set the `null`.
In case we have longer running tasks executing and it would take a while to get to the cancelled tasks, this saves some memory and allows us to return from the method earlier.


javanna:
I believe this change is ok, but I am no longer very confident about `TestTaskExecutor#testCancelTasksOnException` testing a real-life scenario, because it was recently changed to no longer rely on executorService. If I restore that, I get some failures here and there (before as well as after this change). It seems to me that the test was simplified by removing concurrency, and that changes entirely what the test ends up testing.

Additionally, could we have a test around memory being released? (does not have to be perfect, but just something that verifies that the result is nullified?

original-brownbear:
> I believe this change is ok, but I am no longer very confident about TestTaskExecutor#testCancelTasksOnException testing a real-life scenario, because it was recently changed to no longer rely on executorService. If I restore that, I get some failures here and there 

I don't think it was ever materially changed. Before it used to run on a single threaded executor and we'd block the main thread so it was effectively single threaded anyway?
If you want to test that a task failure makes all tasks past it in the queue fail, you kinda need a single thread only for things to be deterministic?


> (does not have to be perfect, but just something that verifies that the result is nullified?

Hmm, it's quite hard to create a useful test here with concurrency enabled. But maybe we could assert that all tasks/runnables are either completed or have been started at the end of `setException`, i.e. after the cancelAll?

javanna:
Ok , my hope was that we can somehow test cancellations with a scenario that's closer to reality, but I do see challenges. I also spent quite a bit of time on this and hit different issues that made testing very difficult.

Perhaps we could at least write a test that fails without the change you made, around nulling memory?

original-brownbear:
@javanna 

> Perhaps we could at least write a test that fails without the change you made, around nulling memory?

I'm really sorry. I thought about this quite a bit, but outside of actually testing that the callables go out of scope via some `Cleaner` magic I think there's no way to do this, all of the functionality is internal to the task executor class. There's just not really any functional change here, it's just that I made the `Callable`s get GCed quicker and saved a little bit of indirection in general.
I think this might be one of these things where we have to trust the code?

jpountz:
Presumably, this micro optimization wouldn't be visible in practice because the tasks that we run allocate even bigger objects such as IndexInput clones? That said, it doesn't look more complicated to me, so I'm good with it.

original-brownbear:
> Presumably, this micro optimization wouldn't be visible in practice because the tasks that we run allocate even bigger objects such as IndexInput clones? 

I wouldn't even see this as an optimization. The point is more to avoid potential memory leaks after cancellation (I guess not quite leaks they'd resolve eventually) in case of a long queue on the executor because we'd still hold a reference to the futures list unless cancel resolves the future.

javanna (approved):

original-brownbear:
Thanks Luca!

PR #13610, opened 2024-07-25
Deduplicate bytes for `FieldReader#rootCode`

original-brownbear:
Loooking at how these instances are serialized to disk it appears that the empty output in the FST metadata is always the same as the rootCode bytes and the assertion I put in here suggests as much as well? Maybe we can just do away with the `rootCode` field outright and simply refer to the FST here?
Without changing the serialization we could at least deduplicate here, saving hundreds of MB in some high-segment count use cases I observed in ES.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/FieldReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/FieldReader.java
@@ -101,7 +101,7 @@ public final class FieldReader extends Terms {
     */
     BytesRef emptyOutput = metadata.getEmptyOutput();
     if (rootCode.equals(emptyOutput) == false) {
-      // TODO: this branch is never taken?
+      // TODO: this branch is never taken
       assert false;
       this.rootCode = rootCode;
     } else {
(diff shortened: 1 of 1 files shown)


jpountz:
The write logic computes the `rootCode` as `root.index.getEmptyOutput()` so it looks like you may be right. @mikemccand Do you know a bit more about this?

mikemccand:
Egads, how does this add up to 100s of MB?  Is each `rootCode` really so big?

I think they are indeed one and the same.  I'm not sure why we are duplicating it -- is it really duplicated (not the same `BytesRef` instance, i.e. `==`)?

`+1` to dedup!

original-brownbear:
@mikemccand It's admittedly the fault of design decisions made and each individually isn't that big. We've simply navigated ourselves into a situation where we have far too many segments per node temporarily at least. They're ~ a couple hundred bytes each in the case I looked into, but it adds up quickly if you have loads of segments as well as many fields per segment. 

But yes it's actually duplicated. It's not the same instance unfortunately, these are just read from the stream twice in two separate reads, same way we duplicate the writes.
Not sure what the right strategy is for fixing the serialization side of things, but I assume that's a new version of the format and it makes sense to do a PR like this one first to save memory for existing segments?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/FieldReader.java
```diff
@@ -100,6 +99,14 @@ public final class FieldReader extends Terms {
      w.close();
      }
     */
+    BytesRef emptyOutput = metadata.getEmptyOutput();
+    if (rootCode.equals(emptyOutput) == false) {
+      // TODO: this branch is never taken?
```
jpountz, line 104: Nit: Remove the question mark, which suggests we're not sure, while both the code and Mike suggest that they are indeed always the same?

jpountz (approved):

original-brownbear:
Thanks Adrien!

PR #13612, opened 2024-07-26
hunspell: add Suggester#proceedPastRep to avoid losing relevant suggestions

donnerpeter:
This PR introduces a way to get a copy of the suggester instance that doesn't stop after encountering acceptable words
(after applying REP rules). By default, Hunspell stops when it finds any, but this behavior may
not always be desirable, e.g., if we have "REP i ea", "tims" would be replaced only by "teams" and not "times", which 
could also be meant.

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Suggester.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Suggester.java
@@ -53,16 +53,21 @@ public class Suggester {
   private final Dictionary dictionary;
   private final SuggestibleEntryCache suggestibleCache;
   private final FragmentChecker fragmentChecker;
+  private final boolean proceedPastRep;
 
   public Suggester(Dictionary dictionary) {
-    this(dictionary, null, FragmentChecker.EVERYTHING_POSSIBLE);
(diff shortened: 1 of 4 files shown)


lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Suggester.java
```diff
@@ -71,16 +76,26 @@ private Suggester(
...
    * Returns a copy of this suggester instance with {@link FragmentChecker} hint that can improve
    * the performance of the "Modification" phase performance.
    */
   public Suggester withFragmentChecker(FragmentChecker checker) {
-    return new Suggester(dictionary, suggestibleCache, checker);
+    return new Suggester(dictionary, suggestibleCache, checker, proceedPastRep);
+  }
+
+  /**
+   * Returns a copy of this suggester instance that doesn't stop after encountering acceptable words
```
dweiss, line 92: It would be good to copy-paste this explanation to the PR's description, Peter. Explains what the patch is doing. I'll do this.

dweiss (approved):

lucene/analysis/common/src/java/org/apache/lucene/analysis/hunspell/Suggester.java
donnerpeter: Thank you!

PR #13613, opened 2024-07-26
Revert cosine deprecation

benwtrent:
Opening for more pointed discussion. See latest discussion here: https://github.com/apache/lucene/issues/13281

I was hoping to have a full answer for folks who use `byte` models by Lucene 10, but I just don't have that. 

I still want to remove the internal cosine optimized methods. We can do this if we store magnitudes along side the raw vectors. This way we can remove all the internal optimized cosine code as its complicated.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -249,9 +249,7 @@ Other
 API Changes
 ---------------------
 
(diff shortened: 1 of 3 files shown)


ChrisHegarty (approved):
I agree with the revert of the deprecation. I’m less sure that we need to remove the optimized cosine code, which is admittedly not part of this PR.

jpountz (approved):

PR #13615, opened 2024-07-27
Fix ScalarQuantization when used with COSINE similarity

### LINKED ISSUE #13614: int4 doesn't handle `cosine` similarity correctly
### Description

just with some default settings, `glove-200` does poorly with `int4` HNSW and when using `cosine`. The bug occurs on merge. When recalculating the quantiles, the vectors aren't normalized like they should, so the quantiles will get all out of whack. We can actually see this in some of the below experiments. All these values should indicate a normalized vectory, however some of the results are > 1 or < -1.

Some experiments I have done:

### Dynamic confidence interval:
 - Floats used to build the graph (e.g. when buffer is large enough): `0.693` (I regrettably didn't get the quantiles here...)
 - When `int4` vectors are used `0.189` (minQuantile=-1.2686034, maxQuantile=1.3250866)

### 0.5 confidence interval (locally patched to allow it):
 - `int4` vectors building the graph `0.649` (minQuantile=-0.277436, maxQuantile=0.29298353)

### 0.75 confidence interval (locally patched to allow it): 
 - `int4` vectors building the graph `0.535` (minQuantile=-0.48500386, maxQuantile=0.5028781)

### 0.9 confidence interval:
 - `int4` vectors building the graph `0.407` (minQuantile=-0.71112806, maxQuantile=0.73441404)

benwtrent:
When quantizing vectors in a COSINE vector space, we normalize them. However, there is a bug when building the quantizer quantiles and we didn't always use the normalized vectors. Consequently, we would end up with poorly configured quantiles and recall will drop significantly (especially in sensitive cases like int4).

closes: https://github.com/apache/lucene/issues/13614


john-wagster (approved):
LGTM

lucene/core/src/java/org/apache/lucene/util/VectorUtil.java
```diff
@@ -128,6 +128,11 @@ public static float[] l2normalize(float[] v) {
     return v;
   }
 
+  public static boolean isUnitVector(float[] v) {
+    double l1norm = IMPL.dotProduct(v, v);
+    return Math.abs(l1norm - 1.0d) <= 1e-4;
```
tteofili, line 133: any specific reason for the specific overflow precision?

tteofili (approved):
LGTM

lucene/core/src/java/org/apache/lucene/util/VectorUtil.java
mayya-sharipova: Looks like it is better to make it  `1e-3`,  for [reference](https://github.com/elastic/elasticsearch/pull/110240)

mayya-sharipova: May be also correct `float[] l2normalize` with the same epsilon

mayya-sharipova (approved):
Thanks Ben

PR #13617, opened 2024-07-29
Removing all deprecated TopScoreDocCollector + TopFieldCollector methods (#create, #createSharedManager)

### LINKED ISSUE #13499: Remove the @Deprecated methods from TopScoreDocCollector and TopFieldCollector

slow-J:
Following up on: https://github.com/apache/lucene/pull/13500

These methods already marked for deprecation in 9.x and we previously removed all internal use of these methods in 10.0 (in https://github.com/apache/lucene/pull/13500).

Tested with: `./gradlew tidy && ./gradlew test`

Closes: #13499


javanna (approved):
Looks great, thank you!

javanna:
Thanks again @slow-J . Heads up, I moved the changes entry to the "API changes" section above after merging, I did not catch this while reviewing :)

slow-J:
My bad, sorry for the mistake, thanks for catching it and the review!

javanna:
> My bad, sorry for the mistake, thanks for catching it and the review!

no big deal, just a small oversight. Cheers!

PR #13618, opened 2024-07-30
Deduplicate min and max term in single-term FieldReader

original-brownbear:
I noticed that single-term readers are an edge case but not that uncommon in Elasticsearch heap dumps. It seems quite common to have a constant value for some field across a complete segment (e.g. a version value that is repeated endlessly in logs).
Seems simple enough to deduplicate here to save a couple MB of heap, though it's admittedly an edge case that mostly matters for large segment counts.


jpountz (approved):
It feels like a corner case, but the change is simple enough. LGTM.

original-brownbear:
Thanks Adrien!

PR #13621, opened 2024-07-30
Fix failing unit test - TestTopDocsCollector#testResultsOrder

### LINKED ISSUE #13620: Reproducible unit test failure
### Description

https://jenkins.thetaphi.de/job/Lucene-main-MacOSX/11653/

I can repro with `./gradlew test --tests TestTopDocsCollector.testResultsOrder -Dtests.seed=207A6071B3338CA6`

The test passes if I revert https://github.com/apache/lucene/commit/d491dfe1315c80319547ec183be3d7aa902d2e9e

I believe this bug occurs due to having 2 collectors with the same indexes and the score of `9.17561f` is never inserted into the pq.

I found a fix, pending some tests.

_No response_

slow-J:
Closes #13620

I believe this bug happens due to having 2 collectors with the same indexes and the score of `9.17561f` is never inserted into the pq, failing the assertion.

Another potential solution is having a SingletonMyTopDocsCollectorMananger which always returns the same `MyTopDocsCollector`.

TEST: `./gradlew tidy && ./gradlew test` + `./gradlew test --tests TestTopDocsCollector.testResultsOrder -Dtests.seed=207A6071B3338CA6`

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/TestTopDocsCollector.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestTopDocsCollector.java
@@ -72,9 +72,6 @@ public class TestTopDocsCollector extends LuceneTestCase {
   }
 
   private static final class MyTopDocsCollector extends TopDocsCollector<ScoreDoc> {
-
-    private int idx = 0;
-
     public MyTopDocsCollector(int size) {
       super(new HitQueue(size, false));
     }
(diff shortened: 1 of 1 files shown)


jpountz:
I think that we could do better by tracking `idx` on the LeafCollector instead of the Collector, and then looking up the score by doing `context.docbase + idx`.

slow-J:
Good point, @jpountz , actually I just tested and the change to `scores[context.docBase + idx++]` is enough to fix the failing test, we can still keep the idx on the Collector.

jpountz:
Keeping `idx` on the collector feels fragile to me, it would only work if there is a single segment per slice?

jpountz (approved):

slow-J:
Thanks for approving @jpountz! Do you think we can merge?

jpountz:
Yes. :)

PR #13622, opened 2024-07-31
Wrap Executor in TaskExecutor to never reject

original-brownbear:
Make it so rejected tasks are execute right away on the caller thread. Users of the API shouldn't have to worry about rejections when we don't expose any upper limit to the task count that we put on the executor that would help in sizing a queue for the executor.


jpountz (commented):
Can you add a simple test, e.g. that submitting 10 tasks to an executor that only has capacity for 2 threads and 3 items in the queue still results in all tasks successfully completing?

original-brownbear:
Done @jpountz, just added 1k count-up tasks, that reliably got me rejections :)

jpountz (commented):
Thanks, this looks correct.

One thing I wonder about is if we should try to better preserve the execution order of tasks. Users sometimes organize their indexes in such a way that the more competitive segments are first in the segment list (e.g. segments that have more recent data). With your approach, if the first half of segments is submitted successfully and the second half is rejected, there are chances that the second half gets processed while the first half is still waiting in the queue. I wonder if we should simply ignore rejections and wait for the current thread to pick up rejected tasks later on via the existing work stealing mechanism. Intuitively, this would be more likely to preserve task execution order?

original-brownbear:
>  if the first half of segments is submitted successfully and the second half is rejected, there are chances that the second half gets processed while the first half is still waiting in the queue.

No that is not true I think. Each of the submitted tasks will take the lowest unprocessed index in the futures list (via the `taskId` counter and process it. The tasks that we submit are not specific tasks to be executed yet, it's always the same `Runnable` that will then grab the next task for the futures list.
Execution order should be preserved as much as possible when processing on more than a single thread with the current approach?

jpountz (approved):
Ahhh that makes sense, thanks for explaining.

original-brownbear:
npnp, thanks for taking a look Adrien!

dsmiley:
Looking back at this, might it have been better to instead wrap `TaskExecutor.invokeAll`'s  call of `executor.execute` in a loop to catch `RejectedExecutionException` and then don't bother submitting additional work?  The rationale is that if we see rejection, it's wasteful to continue to submit tasks that will almost certainly also reject.

original-brownbear:
I agree @dsmiley , I actually did continue to work on this on the ES side lately in https://github.com/elastic/elasticsearch/pull/120024. 
What I did there was introduce logarithmic activation into this, making each thread start one other thread and then have the tasks started that way all loop on top of the same set of tasks.
In that scheme the level of parallelism becomes a natural function of what the pool can deal with.
The main issue I had with this approach in Lucene core is the `Executor` API. I can't really apply the approach neatly to it because I don't know if a call to `execute` will fork or not.
Starting to think it might be worthwhile to special case `ThreadPoolExecutor` here like we do in ES to get around this issue in the short-term.

PR #13625, opened 2024-08-01
Remove some BitSet#nextSetBit code duplication

gsmiller:
After merging #13559 I noticed an opportunity to remove some redundant code in the `nextSetBit` implementations.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/SparseFixedBitSet.java
+++ b/lucene/core/src/java/org/apache/lucene/util/SparseFixedBitSet.java
@@ -355,7 +355,7 @@ public class SparseFixedBitSet extends BitSet {
    */
   private int nextSetBitInRange(int start, int upperBound) {
     assert start < length;
-    assert upperBound > start;
+    assert upperBound > start && upperBound <= length : "upperBound=" + upperBound + ", start=" + start + ", length=" + length;
     final int i4096 = start >>> 12;
     final long index = indices[i4096];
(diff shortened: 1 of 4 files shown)


gsmiller:
@epotyom this minor refactoring occurred to me after merging your recent work. WDYT?

lucene/core/src/java/org/apache/lucene/util/SparseFixedBitSet.java
```diff
@@ -337,34 +337,23 @@ private int firstDoc(int i4096, int i4096upper) {
...
+  }
+
+  /**
+   * Returns the next set bit in the specified range, but treats `upperBound` as a best-effort hint
+   * rather than a hard requirement. Note that this may return a result that is >= upperBound in
+   * some cases, so callers must add their own check if `upperBound` is a hard requirement.
+   */
+  private int nextSetBitInRange(int start, int upperBound) {
     assert start < length;
     assert upperBound > start;
```
epotyom, line 358: I know it's preexisting, but maybe we can extend the assert to make potential debugging easier?
```suggestion
    assert upperBound > start && upperBound <= length : "upperBound=" + upperBound + ", start=" + start + ", length=" + length;
```

epotyom (approved):
I like it, thank you for making the change! One minor suggestion:

gsmiller:
Thanks @epotyom for the feedback!

Issue #13628, opened 2024-08-02
These attributes are better for the final state


mrhbj:
### Description

 when Lucene99SkipWriter creating，you set up the maxSkipLevels size to lastSkipDoc,lastSkipDocPointer,curCompetitiveFreqNorms,and their size never change again。
make them final. Is it better?

### Version and environment details

public final class Lucene99SkipWriter extends MultiLevelSkipListWriter {
  private int[] lastSkipDoc;
  private long[] lastSkipDocPointer;
  private long[] lastSkipPosPointer;
  private long[] lastSkipPayPointer;

  private final IndexOutput docOut;
  private final IndexOutput posOut;
  private final IndexOutput payOut;

  private int curDoc;
  private long curDocPointer;
  private long curPosPointer;
  private long curPayPointer;
  private int curPosBufferUpto;
  private int curPayloadByteUpto;
  private CompetitiveImpactAccumulator[] curCompetitiveFreqNorms;
  private boolean fieldHasPositions;
  private boolean fieldHasOffsets;
  private boolean fieldHasPayloads;

  /** Sole constructor. */
  public Lucene99SkipWriter(
      int maxSkipLevels,
      int blockSize,
      int docCount,
      IndexOutput docOut,
      IndexOutput posOut,
      IndexOutput payOut) {
    super(blockSize, 8, maxSkipLevels, docCount);
    this.docOut = docOut;
    this.posOut = posOut;
    this.payOut = payOut;

    lastSkipDoc = new int[maxSkipLevels];
    lastSkipDocPointer = new long[maxSkipLevels];
    if (posOut != null) {
      lastSkipPosPointer = new long[maxSkipLevels];
      if (payOut != null) {
        lastSkipPayPointer = new long[maxSkipLevels];
      }
    }
    curCompetitiveFreqNorms = new CompetitiveImpactAccumulator[maxSkipLevels];
    for (int i = 0; i < maxSkipLevels; ++i) {
      curCompetitiveFreqNorms[i] = new CompetitiveImpactAccumulator();
    }
  }

gsmiller:
Added a small comment in your PR (#13630). Thanks!

gsmiller:
Addressed in #13639

PR #13630, opened 2024-08-02
These attributes are better for the final state(#13628)

mrhbj:
These attributes are better for the final state(#13628)

### DIFF
--- a/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene99/Lucene99SkipWriter.java
+++ b/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene99/Lucene99SkipWriter.java
@@ -46,8 +46,8 @@ import org.apache.lucene.store.IndexOutput;
  * uptos(position, payload). 4. start offset.
  */
 public final class Lucene99SkipWriter extends MultiLevelSkipListWriter {
-  private int[] lastSkipDoc;
-  private long[] lastSkipDocPointer;
+  private final int[] lastSkipDoc;
+  private final long[] lastSkipDocPointer;
   private long[] lastSkipPosPointer;
   private long[] lastSkipPayPointer;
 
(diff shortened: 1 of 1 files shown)


lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene99/Lucene99SkipWriter.java
```diff
@@ -46,8 +46,8 @@
  * uptos(position, payload). 4. start offset.
  */
 public final class Lucene99SkipWriter extends MultiLevelSkipListWriter {
-  private int[] lastSkipDoc;
-  private long[] lastSkipDocPointer;
+  private final int[] lastSkipDoc;
+  private final long[] lastSkipDocPointer;
   private long[] lastSkipPosPointer;
   private long[] lastSkipPayPointer;
```
gsmiller, line 52: We can make these final too if we tidy the ctor a little bit. Let's go ahead and do this if we're doing a pass on marking fields `final`?

gsmiller (commented):
Always nice to mark fields `final` when possible IMO. Thanks! I noted two more we can also do this with if we just tidy the ctor a little bit to explicitly init them to `null` when not needed. Let's add that as well?

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene99/Lucene99SkipWriter.java
gsmiller: @mrhbj were you planning to push an update to this PR that also marks these fields `final`? Thanks!

mrhbj:
ok i will do this



---Original---
From: "Greg ***@***.***&gt;
Date: Fri, Aug 9, 2024 23:12 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [apache/lucene] These attributes are better for the finalstate(#13628) (PR #13630)




 
@gsmiller commented on this pull request.
 
 
In lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene99/Lucene99SkipWriter.java:
 &gt;    private long[] lastSkipPosPointer;    private long[] lastSkipPayPointer;  
@mrhbj were you planning to push an update to this PR that also marks these fields final? Thanks!

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene99/Lucene99SkipWriter.java
mrhbj: @gsmiller I have did that in #13639

gsmiller:
Closing in favor of #13639

PR #13632, opened 2024-08-05
CandidateMatcher public matching functions

bjacobowitz:
### Description

In `CandidateMatcher`, increase visibility of `matchQuery()` (from protected), `finish()` and `reportError()` (from package-private) to public, so that a `CandidateMatcher` can be used effectively from outside the `org.apache.lucene.monitor` package.

### Problem

The current access protections on `CandidateMatcher` make it infeasible for a user of the Lucene Monitor library to use an existing `CandidateMatcher` if they are outside the `org.apache.lucene.monitor` package (e.g. as part of the implementation of their own `CandidateMatcher`).

For example, a user working outside of the `org.apache.lucene.monitor` package could not build their own version of `ParallelMatcher` by making use of the matcher from `QueryMatch.SIMPLE_MATCHER`, because they cannot access `matchQuery`, `reportError` or `finish` on it.

### Solution

This PR takes the simplest solution of increase the visibility of those functions in `CandidateMatcher` to `public`. This also requires modifying some existing `CandidateMatcher` implementations that override `protected matchQuery()` to instead override `public matchQuery()`. 

I've also added some just-compile tests in a subpackage `org.apache.lucene.monitor.otherpackage`, which call the newly-public functions to verify that they are accessible from outside the `org.apache.lucene.monitor` package.

See previous discussion of this approach and alternatives in #13109.

### Merge Request

If this PR gets merged, can you please use my `bjacobowitz1@bloomberg.net` email address for the squash+merge. Thank you.

### DIFF
--- a/lucene/monitor/src/test/org/apache/lucene/monitor/outsidepackage/TestCandidateMatcherVisibility.java
+++ b/lucene/monitor/src/test/org/apache/lucene/monitor/outsidepackage/TestCandidateMatcherVisibility.java
@@ -19,160 +19,21 @@ package org.apache.lucene.monitor.outsidepackage;
 
 import java.io.IOException;
 import java.util.Collections;
-import org.apache.lucene.index.BinaryDocValues;
-import org.apache.lucene.index.ByteVectorValues;
-import org.apache.lucene.index.DocValuesSkipper;
-import org.apache.lucene.index.FieldInfos;
-import org.apache.lucene.index.FloatVectorValues;
-import org.apache.lucene.index.LeafMetaData;
-import org.apache.lucene.index.LeafReader;
-import org.apache.lucene.index.NumericDocValues;
-import org.apache.lucene.index.PointValues;
-import org.apache.lucene.index.SortedDocValues;
-import org.apache.lucene.index.SortedNumericDocValues;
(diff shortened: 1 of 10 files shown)


lucene/monitor/src/test/org/apache/lucene/monitor/outsidepackage/TestCandidateMatcherVisibility.java
```diff
@@ -0,0 +1,204 @@
...
+import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.KnnCollector;
+import org.apache.lucene.search.TermQuery;
+import org.apache.lucene.util.Bits;
+import org.apache.lucene.util.Version;
+import org.junit.Test;
+
+public class TestCandidateMatcherVisibility {
+
+  // Dummy empty IndexReader for use in creating a matcher
```
romseygeek, line 50: Using a MemoryIndex here would probably be simpler?

romseygeek (changes requested):
Thanks @bjacobowitz, this looks great!  Can you also create a CHANGES.txt entry under 'API changes'?

romseygeek (approved):

romseygeek:
Thanks @bjacobowitz!

Issue #13634, opened 2024-08-06
Seeding HNSW Search


seanmacavaney:
### Description

In some vector search cases, users may already know some documents that are likely related to a query. Let's support seeding HNSW's scoring stage with these documents, rather than using HNSW's hierarchical stage.

An example use case is hybrid search, where both a traditional and vector search are performed. The top results from the traditional search are likely reasonable seeds for the vector search. Even when not performing hybrid search, traditional matching can often be faster than traversing the hierarchy, which can be used to speed up the vector search process (up to 2x faster for the same effectiveness), as was demonstrated in [this article](https://arxiv.org/abs/2307.16779) (full disclosure: I'm an author of the article).

This enhancement proposes adding a `seed` query, alongside the existing `filter` query, to the KNN query classes. The results of this query will be fed into `HnswGraphSearcher`, and ultimately replace the graph entry points [here](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java#L103-L106). If the seed query fails (e.g., keywords do not match any documents), the approach will fall back onto the existing hierarchical search process.

Pull request to follow.

benwtrent:
@seanmacavaney I like this idea (I remember reading this paper a while back and getting excited about it). 

A couple of concerns I have are:

 - The API, this is always tricky to get correct
 - The `seed` query, will it be scored? How will we ensure there is a limit (e.g. that somebody doesn't just pass a match all docs query). 
 - What should the behavior be when the seed query matches NO documents? I would assume the correct behavior here is to traverse the graph as normal.

Looking forward to the PR :)

seanmacavaney:
Thanks! I just opened a draft PR (#13635). To answer your questions:

> The API, this is always tricky to get correct

I've struggled a bit with this. The PR has an attempt, and I would totally appreciate feedback on it!

> The seed query, will it be scored? How will we ensure there is a limit (e.g. that somebody doesn't just pass a match all docs query).

Yes, it's scored. The PR sets a limit of 10 seed documents. (In contrast, HNSW uses a single entry point based on the hierarchical search.) This could also be something configurable, though I wouldn't want to complicate the API too much.

> What should the behavior be when the seed query matches NO documents? I would assume the correct behavior here is to traverse the graph as normal.

Yep, I agree that falling back on the default behavior is reasonable. This is what the PR implements.

msokolov:
[...]

PR #13635, opened 2024-08-06
Add AbstractKnnVectorQuery.seed for seeded HNSW

seanmacavaney:
### Description

This PR addresses #13634.

The main changes are in:
 - `AbstractKnnVectorQuery`, which adds a `seed` field. It scores this query if provided, and passes these seeds along to the `HnswGraphSearcher` as a `DocIdSetIterator`.
 - `HnswGraphSearcher`, which bypasses the `findBestEntryPoint` step if seeds are provided.

Unfortunately, passing the seeds down to `HnswGraphSearcher` resulted in changes in several places, e.g., to backward codecs.


lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -156,6 +189,44 @@ private TopDocs getLeafResults(
     }
   }
 
+  private DocIdSetIterator executeSeedQuery(LeafReaderContext ctx, Weight seedWeight)
```
benwtrent, line 192: Interesting indeed. 

I would suppose, that depending on the query, this might actually be slower than doing kNN search. Though, I guess you could have the benefit of requiring less candidates (smaller efSearch). 

How should this weight behave with pre-filtering? Shouldn't we apply the filter with the seed query to ensure the starting entry points are also valid candidates?

benwtrent (commented):
It would be very interesting to see how this behaves in Lucene directly. I could see it being very nice, or behaving poorly depending on the seed query (which, I guess is expected).


The API of adding the `DocIdSetIterator` for the seeding docs is what I would initially think of as well.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
benwtrent: Also, at a minimum, we would need to make sure that the seed docs have the vector field. It is possible (actually rather likely), that the query actually doesn't match any docs that contain vectors, then the down stream translation from docId -> vector ordinal completely fails.

seanmacavaney: Right, yes! I think it should be as simple as applying a similar boolean query, as is done for the filter weight:

https://github.com/apache/lucene/pull/13635/files#diff-0a10fed5fe4af3bba13deab713dea47f98a298d3962361a88d5c2a073c4892a7R86-R90

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java
```diff
@@ -70,6 +72,43 @@ public static void search(
...
+   * @param acceptOrds {@link Bits} that represents the allowed document ordinals to match, or
+   *     {@code null} if they are all allowed to match.
+   * @param entryPointOrds the entry points for search.
+   */
+  public static void search(
+      RandomVectorScorer scorer,
+      KnnCollector knnCollector,
+      HnswGraph graph,
+      Bits acceptOrds,
+      DocIdSetIterator entryPointOrds)
```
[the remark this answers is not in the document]
seanmacavaney: The filter should have needed to address the same mismatch somehow-- I'll see how it's resolved there and follow a similar technique.

seanmacavaney:
> I could see it being very nice, or behaving poorly depending on the seed query (which, I guess is expected).

We could probably predict whether a seed set is good or bad based on how many documents it's needed to score and terminate early if it seems like it's too many. Could be accomplished with the existing `KnnCollector` visited count. I'll look into this.

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsReader.java
```diff
@@ -82,9 +83,16 @@ protected KnnVectorsReader() {}
    * @param knnCollector a KnnResults collector and relevant settings for gathering vector results
    * @param acceptDocs {@link Bits} that represents the allowed documents to match, or {@code null}
    *     if they are all allowed to match.
+   * @param seedDocs {@link Bits} that represents the documents used to seed the search, or {@code
```
cpoerschke, line 86:
```suggestion
   * @param seedDocs {@link DocIdSetIterator} that represents the documents used to seed the search, or {@code
```

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsReader.java
```diff
@@ -110,9 +118,16 @@ public abstract void search(
...
```
cpoerschke, line 121:
```suggestion
   * @param seedDocs {@link DocIdSetIterator} that represents the documents used to seed the search, or {@code
```

lucene/core/src/java/org/apache/lucene/search/KnnByteVectorQuery.java
```diff
@@ -72,14 +72,30 @@ public KnnByteVectorQuery(String field, byte[] target, int k) {
...
   public KnnByteVectorQuery(String field, byte[] target, int k, Query filter) {
-    super(field, k, filter);
+    this(field, target, k, filter, null);
+  }
+
+  /**
+   * Find the <code>k</code> nearest documents to the target vector according to the vectors in the
+   * given field. <code>target</code> vector.
+   *
+   * @param field a field that has been indexed as a {@link KnnFloatVectorField}.
```
cpoerschke, line 82:
```suggestion
   * @param field a field that has been indexed as a {@link KnnByteVectorField}.
```
copy/paste from existing line 68 above -- #13637 for that.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java
```diff
@@ -70,6 +72,43 @@ public static void search(
...
+      throws IOException {
+    ArrayList<Integer> entryPointOrdInts = new ArrayList<Integer>();
+    if (entryPointOrds != null) {
```
cpoerschke, line 95: maybe, to avoid `entryPointOrdInts` object allocation if `entryPointOrds` is null.
```suggestion
    ArrayList<Integer> entryPointOrdInts = null;
    if (entryPointOrds != null) {
      entryPointOrdInts = new ArrayList<Integer>();
```

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java
```diff
@@ -70,6 +72,43 @@ public static void search(
...
+      int entryPointOrdInt;
+      while ((entryPointOrdInt = entryPointOrds.nextDoc()) != NO_MORE_DOCS) {
+        entryPointOrdInts.add(entryPointOrdInt);
+      }
+    }
+    if (entryPointOrdInts.size() == 0) {
```
cpoerschke, line 101:
```suggestion
    if (entryPointOrdInts == null || entryPointOrdInts.isEmpty()) {
```

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -156,6 +196,44 @@ private TopDocs getLeafResults(
...
+      throws IOException {
+    if (seedWeight != null) {
+      // Execute the seed query
+      TopScoreDocCollector seedCollector =
+          new TopScoreDocCollectorManager(k, Integer.MAX_VALUE).newCollector();
```
cpoerschke, line 204: Wondering if we could use a lower `totalHitsThreshold` here since `seedCollector.topDocs().totalHits` seems to be unused?
```suggestion
          new TopScoreDocCollectorManager(k, k).newCollector();
```

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -491,4 +580,44 @@ public int hashCode() {
           classHash(), contextIdentity, Arrays.hashCode(docs), Arrays.hashCode(scores));
     }
   }
+
+  private class TopDocsDISI extends DocIdSetIterator {
+    int idx = -1;
+    List<Integer> sortedDocIdList;
+
+    public TopDocsDISI(TopDocs topDocs) {
+      sortedDocIdList = new ArrayList<Integer>();
```
cpoerschke, line 589:
```suggestion
      sortedDocIdList = new ArrayList<Integer>(topDocs.scoreDocs.length);
```

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java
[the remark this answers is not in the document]
benwtrent: @seanmacavaney looking at the code, I think you can do what you want if you get access to the underlying `SparseOffHeapVectorValues*` implementation. Then you can use the `IndexedDISI` to advance to each matched doc & use `disi.index()` to get the corresponding ordinal.

Maybe `FlatVectorsReader` can have a method called `DocIdSetIterator 
 getVectorOrdinals(String field, DocIdSetIterator docIdIterator)` ?

Or `RandomAccessVectorValues` can have such a method. Both of these interfaces are experimental, so changing them is OK.

seanmacavaney:
Thanks all! I was away for a bit but back on it now :)

benwtrent:
I haven't been able to review fully, but will soon. 

I like this idea. Do you have any scripts & test data where you have tested that this Lucene implementation works and gives as good as results as you had in the original paper?

One thing that still bugs me is attaching yet another parameter to a common method. Especially a parameter that will be rarely used (my gut reaction is choosing the correct seed query is difficult and average users will never use it). I will think more about this. Maybe its just fine and I am overreacting.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -156,6 +196,58 @@ private TopDocs getLeafResults(
...
+            scorer.score(leafCollector, ctx.reader().getLiveDocs());
+          } catch (
+              @SuppressWarnings("unused")
+              CollectionTerminatedException e) {
+            // collection was terminated prematurely
+            // continue with the following leaf
+          }
+        }
+        leafCollector.finish();
+      }
```
cpoerschke, line 233: minor: this looks like it could move into the `try` block

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -491,4 +594,44 @@ public int hashCode() {
           classHash(), contextIdentity, Arrays.hashCode(docs), Arrays.hashCode(scores));
     }
   }
+
+  private class TopDocsDISI extends DocIdSetIterator {
+    int idx = -1;
+    List<Integer> sortedDocIdList;
```
cpoerschke, line 600:
```suggestion
  private static class TopDocsDISI extends DocIdSetIterator {
    private final List<Integer> sortedDocIdList;
    private int idx = -1;
```

lucene/core/src/java/org/apache/lucene/search/KnnByteVectorQuery.java
```diff
@@ -133,4 +149,18 @@ public int hashCode() {
   public byte[] getTargetCopy() {
     return ArrayUtil.copyArray(target);
   }
+
+  /**
+   * Returns a new iterator that maps the provided docIds to the vector ordinals.
+   *
+   * <p>This method assumes that all docIds have corresponding orginals.
```
cpoerschke, line 156:
```suggestion
   * <p>This method assumes that all docIds have corresponding ordinals.
```

lucene/core/src/java/org/apache/lucene/search/KnnFloatVectorQuery.java
```diff
@@ -136,4 +152,18 @@ public int hashCode() {
   public float[] getTargetCopy() {
     return ArrayUtil.copyArray(target);
   }
+
+  /**
+   * Returns a new iterator that maps the provided docIds to the vector ordinals.
+   *
+   * <p>This method assumes that all docIds have corresponding orginals.
```
cpoerschke, line 159:
```suggestion
   * <p>This method assumes that all docIds have corresponding ordinals.
```

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -156,6 +196,58 @@ private TopDocs getLeafResults(
...
+  }
+
+  /**
+   * Returns a new iterator that maps the provided docIds to the vector ordinals.
+   *
+   * @lucene.internal
+   * @lucene.experimental
+   */
+  protected abstract DocIdSetIterator convertDocIdsToVectorOrdinals(
+      LeafReaderContext ctx, DocIdSetIterator docIds) throws IOException;
```
cpoerschke, line 249: Since both implementations do a `ctx.reader()` call, wondering if the signature might be narrower then?
```suggestion
      LeafReader reader, DocIdSetIterator docIds) throws IOException;
```

lucene/core/src/java/org/apache/lucene/index/FloatVectorValues.java
```diff
@@ -87,4 +87,16 @@ public static void checkField(LeafReader in, String field) {
    * @return a {@link VectorScorer} instance or null
    */
   public abstract VectorScorer scorer(float[] query) throws IOException;
+
+  /**
+   * Returns a new iterator that maps the provided docIds to the vector ordinals.
+   *
+   * <p>This method assumes that all docIds have corresponding orginals.
```
cpoerschke, line 94:
```suggestion
   * <p>This method assumes that all docIds have corresponding ordinals.
```
or
```suggestion
```

lucene/core/src/java/org/apache/lucene/index/ByteVectorValues.java
```diff
@@ -86,4 +86,16 @@ public static void checkField(LeafReader in, String field) {
    * @return a {@link VectorScorer} instance or null
    */
   public abstract VectorScorer scorer(byte[] query) throws IOException;
+
+  /**
+   * Returns a new iterator that maps the provided docIds to the vector ordinals.
+   *
+   * <p>This method assumes that all docIds have corresponding orginals.
```
cpoerschke, line 93:
```suggestion
   * <p>This method assumes that all docIds have corresponding ordinals.
```
or
```suggestion
```

lucene/core/src/java/org/apache/lucene/index/LeafReader.java
```diff
@@ -251,7 +252,13 @@ public final PostingsEnum postings(Term term) throws IOException {
    * @lucene.experimental
    */
   public final TopDocs searchNearestVectors(
-      String field, float[] target, int k, Bits acceptDocs, int visitedLimit) throws IOException {
+      String field,
+      float[] target,
+      int k,
+      Bits acceptDocs,
+      DocIdSetIterator seedDocs,
```
cpoerschke, line 259: TODO: javadoc update to add the new param

lucene/core/src/java/org/apache/lucene/index/LeafReader.java
```diff
@@ -295,7 +302,13 @@ public final TopDocs searchNearestVectors(
    * @lucene.experimental
    */
   public final TopDocs searchNearestVectors(
-      String field, byte[] target, int k, Bits acceptDocs, int visitedLimit) throws IOException {
+      String field,
+      byte[] target,
+      int k,
+      Bits acceptDocs,
+      DocIdSetIterator seedDocs,
```
cpoerschke, line 309: TODO: javadoc update to add the new param

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java
```diff
@@ -70,6 +72,44 @@ public static void search(
     search(scorer, knnCollector, graph, graphSearcher, acceptOrds);
   }
 
+  /**
+   * Searches the HNSW graph for for the nerest neighbors of a query vector, starting from the
```
cpoerschke, line 76:
```suggestion
   * Searches the HNSW graph for the nearest neighbors of a query vector, starting from the
```

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java
```diff
@@ -70,6 +72,44 @@ public static void search(
...
+      while ((entryPointOrdInt = entryPointOrds.nextDoc()) != NO_MORE_DOCS) {
+        entryPointOrdInts.add(entryPointOrdInt);
+      }
+    }
+    if (entryPointOrdInts == null || entryPointOrdInts.isEmpty()) {
+      search(scorer, knnCollector, graph, acceptOrds);
+    } else {
+      HnswGraphSearcher graphSearcher =
+          new HnswGraphSearcher(
+              new NeighborQueue(knnCollector.k(), true), new SparseFixedBitSet(graph.size()));
```
cpoerschke, line 107: maybe
```suggestion
              new NeighborQueue(knnCollector.k(), true), new SparseFixedBitSet(getGraphSize(graph)));
```

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java
```diff
@@ -70,6 +72,44 @@ public static void search(
...
+      int[] entryPointOrdIntsArr = entryPointOrdInts.stream().mapToInt(Integer::intValue).toArray();
+      graphSearcher.searchLevel(knnCollector, scorer, 0, entryPointOrdIntsArr, graph, acceptOrds);
```
cpoerschke, line 109: naive questions: looking at https://github.com/apache/lucene/blob/releases/lucene/9.11.1/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java#L103-L106 and https://github.com/apache/lucene/blob/releases/lucene/9.11.1/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java#L133 we 'know' it's the 0-th layer. (how) do we know that it's also the 0-th layer/level here and/or could the entry points from the seed query also be from different layers/levels? would love to see a short comment in the code re: that.
```suggestion
      graphSearcher.searchLevel(knnCollector, scorer, 0 /* level */, entryPointOrdIntsArr, graph, acceptOrds);
```

benwtrent: Node ordinals are the same no matter the layer. Meaning `ord: 42` refers to vector `42` no matter what layer its on.

seanmacavaney:
Thanks for the comments @cpoerschke. That should be them all addressed.

seanmacavaney:
Thanks @benwtrent!

> One thing that still bugs me is attaching yet another parameter to a common method. Especially a parameter that will be rarely used (my gut reaction is choosing the correct seed query is difficult and average users will never use it). I will think more about this. Maybe its just fine and I am overreacting.

That's a fair point. What might be the alternative that doesn't affect the main `LeafReader.searchNearestVectors` methods? Perhaps a private member with a getter/setter?

> I like this idea. Do you have any scripts & test data where you have tested that this Lucene implementation works and gives as good as results as you had in the original paper?

I've tested it on some internal datasets and it's performed favorably, albeit not quite as strong as the original paper. (This is likely due to both differences in the distribution of the data and implementation differences, especially around the efficiency of lexical search.) Automated tests in the suite test that it's functioning as expected.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -156,6 +196,55 @@ private TopDocs getLeafResults(
...
+                DocIdSetIterator.NO_MORE_DOCS /* max */);
+          }
+          leafCollector.finish();
+        }
+      } catch (
+          @SuppressWarnings("unused")
+          CollectionTerminatedException e) {
+        // there is no doc of interest in this reader context
+        // continue with the following leaf
+        leafCollector = null;
```
cpoerschke, line 229:
```suggestion
```

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -156,6 +196,55 @@ private TopDocs getLeafResults(
...
+      TopScoreDocCollector seedCollector =
+          new TopScoreDocCollectorManager(
+                  k /* numHits */,
+                  null /* after */,
+                  Integer.MAX_VALUE /* totalHitsThreshold */,
+                  false /* supportsConcurrency */)
+              .newCollector();
+      LeafCollector leafCollector;
+      try {
+        leafCollector = seedCollector.getLeafCollector(ctx);
```
cpoerschke, line 212:
```suggestion
      try {
        final LeafCollector leafCollector = seedCollector.getLeafCollector(ctx);
```

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -156,6 +196,55 @@ private TopDocs getLeafResults(
...
+      } catch (
+          @SuppressWarnings("unused")
+          CollectionTerminatedException e) {
+        // there is no doc of interest in this reader context
+        // continue with the following leaf
+        leafCollector = null;
+      }
+
+      TopDocs seedTopDocs = seedCollector.topDocs();
+      return convertDocIdsToVectorOrdinals(ctx.reader(), new TopDocsDISI(seedTopDocs));
```
cpoerschke, line 233: minor/subjective: could introduce local variable to avoid repeat `ctx.reader()` call and show that the two parts of the code using the reader are connected
```
final LeafReader leafReader = ctx.reader();
```

lucene/core/src/java/org/apache/lucene/search/KnnByteVectorQuery.java
```diff
@@ -133,4 +149,18 @@ public int hashCode() {
   public byte[] getTargetCopy() {
     return ArrayUtil.copyArray(target);
   }
+
+  /**
+   * Returns a new iterator that maps the provided docIds to the vector ordinals.
+   *
+   * <p>This method assumes that all docIds have corresponding .
```
cpoerschke, line 156:
```suggestion
   * <p>This method assumes that all docIds have corresponding ???.
```

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene94/Lucene94HnswVectorsReader.java
```diff
@@ -283,11 +289,17 @@ public void search(String field, float[] target, KnnCollector knnCollector, Bits
         scorer,
         new OrdinalTranslatedKnnCollector(knnCollector, vectorValues::ordToDoc),
         getGraph(fieldEntry),
-        vectorValues.getAcceptOrds(acceptDocs));
+        vectorValues.getAcceptOrds(acceptDocs),
+        seedDocs);
```
cpoerschke, line 293: This question is perhaps a side effect of the history of this pull request and/or a naive question on my part: when do and don't we change the implementation of classes in the `backwards_codecs` (implementation other than signature) -- I note that here for 94 and 95 there are changes but for 92 there is no change. and 99 has no change (as yet). My speculation (but it is only that) is that perhaps the 94 and 95 changes need to move to 99 and that anything in `backwards_codecs` would get the method signature change only but no implementation change?

lucene/test-framework/src/java/org/apache/lucene/tests/index/MismatchedLeafReader.java
```diff
@@ -68,6 +71,28 @@ public CacheHelper getReaderCacheHelper() {
...
+  public void searchNearestVectors(
+      String field,
+      byte[] target,
+      KnnCollector knnCollector,
+      Bits acceptDocs,
+      DocIdSetIterator seedDocs)
+      throws IOException {
+    in.searchNearestVectors(field, target, knnCollector, acceptDocs, seedDocs);
+  }
+
```
cpoerschke, line 95: Might not be necessary if/since `FilterLeafReader.java` has the implementation.
```suggestion
```

cpoerschke (commented):
50 / 57 files viewed

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene94/Lucene94HnswVectorsReader.java
seanmacavaney: Good points -- I think this is a question for @benwtrent?

lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
```diff
@@ -608,6 +614,94 @@ public void testRandomWithFilter() throws IOException {
     }
   }
 
+  /** Tests with random vectors and a random seed. Uses RandomIndexWriter. */
+  public void testRandomWithSeed() throws IOException {
+    int numDocs = 1000;
+    int dimension = atLeast(5);
+    int numIters = atLeast(10);
+    try (Directory d = newDirectory()) {
```
cpoerschke, line 622:
```suggestion
    try (Directory d = newDirectoryForTest()) {
```

lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
```diff
@@ -608,6 +614,94 @@ public void testRandomWithFilter() throws IOException {
...
+        for (int i = 0; i < numIters; i++) {
+          int k = random().nextInt(80) + 1;
+          int n = random().nextInt(100) + 1;
+
+          // All documents as seeds
+          Query seed1 = new MatchAllDocsQuery();
+          AbstractKnnVectorQuery query =
+              getKnnVectorQuery("field", randomVector(dimension), k, null, seed1);
+          TopDocs results = searcher.search(query, n);
+          int expected = Math.min(Math.min(n, k), reader.numDocs());
```
cpoerschke, line 654: Instead of `reader.numDocs()` should this be the number of docs that have the vector field?

lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
```diff
@@ -608,6 +614,94 @@ public void testRandomWithFilter() throws IOException {
...
+          }
+
+          // No seed documents -- falls back on full approx search
+          Query seed3 = new BooleanQuery.Builder().build();
+          query = getKnnVectorQuery("field", randomVector(dimension), k, null, seed3);
+          results = searcher.search(query, n);
+          expected = Math.min(Math.min(n, k), reader.numDocs());
+          // we may get fewer results than requested if there are deletions, but this test doesn't
+          // test that
+          assert reader.hasDeletions() == false;
```
cpoerschke, line 691: seems duplicated three times, maybe move to start of the loop right after `k` and `n` variables.

lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
```diff
@@ -608,6 +614,94 @@ public void testRandomWithFilter() throws IOException {
...
+          assertTrue(results.totalHits.value() >= results.scoreDocs.length);
+          // verify the results are in descending score order
+          last = Float.MAX_VALUE;
+          for (ScoreDoc scoreDoc : results.scoreDocs) {
+            assertTrue(scoreDoc.score <= last);
+            last = scoreDoc.score;
+          }
+
+          // No seed documents -- falls back on full approx search
+          Query seed3 = new BooleanQuery.Builder().build();
```
cpoerschke, line 685: Wondering if `MatchNoDocsQuery()` might work here too?
```suggestion
          Query seed3 = new MatchNoDocsQuery();
```

lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
```diff
@@ -608,6 +614,94 @@ public void testRandomWithFilter() throws IOException {
...
+      try (IndexReader reader = DirectoryReader.open(d)) {
+        IndexSearcher searcher = newSearcher(reader);
+        for (int i = 0; i < numIters; i++) {
+          int k = random().nextInt(80) + 1;
+          int n = random().nextInt(100) + 1;
+
+          // All documents as seeds
+          Query seed1 = new MatchAllDocsQuery();
+          AbstractKnnVectorQuery query =
+              getKnnVectorQuery("field", randomVector(dimension), k, null, seed1);
```
cpoerschke, line 652: Wondering if randomly `Query filter = random().nextBoolean() ? null : new SomethingQuery()` could be null or non-null and then used in the test, for coverage?
```suggestion
              getKnnVectorQuery("field", randomVector(dimension), k, filter, seed1);
```

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
cpoerschke: sorry, i think i got this suggestion wrong, `leafCollector.finish()` needs to be called even if an exception was caught and as such the declaration needs to be outside the `try-catch` block.

cpoerschke: or if we can establish that `getLeafCollector` will not throw the exception then the `try-catch` could be just around the bulkScore/score calls?

benwtrent:
Thinking more and more, I do not like the idea of adding to the leaf function definition. 

But, this does seem useful. I think we can attach it to kNN collectors.

I don't think the base interface for the knn collectors should have it. But maybe a new interface indicating that it provides a collection of graph entry points and that searches should start there.

Then downstream callers can check if that interface is satisfied and use it. 

This interface should be considered experimental so we can adjust it without backwards compatibility concerns.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -156,6 +196,49 @@ private TopDocs getLeafResults(
...
+              leafReader.getLiveDocs(),
+              0 /* min */,
+              DocIdSetIterator.NO_MORE_DOCS /* max */);
+        }
+        leafCollector.finish();
+      }
+    } catch (
+        @SuppressWarnings("unused")
+        CollectionTerminatedException e) {
+    }
```
cpoerschke, line 227: So that `leafCollector.finish()` is called even when `CollectionTerminatedException` was caught, and because (by inspection) `getLeafCollector` on this code path does not throw that exception.
```suggestion
    final LeafReader leafReader = ctx.reader();
      final LeafCollector leafCollector = seedCollector.getLeafCollector(ctx);
      if (leafCollector != null) {
    try {
        BulkScorer scorer = seedWeight.bulkScorer(ctx);
        if (scorer != null) {
          scorer.score(
              leafCollector,
              leafReader.getLiveDocs(),
              0 /* min */,
              DocIdSetIterator.NO_MORE_DOCS /* max */);
        }
    } catch (
        @SuppressWarnings("unused")
        CollectionTerminatedException e) {
    }
        leafCollector.finish();
      }
```

lucene/core/src/test/org/apache/lucene/document/TestManyKnnDocs.java
```diff
@@ -46,27 +54,139 @@ public void testLargeSegment() throws Exception {
     mp.setMaxMergeAtOnce(256); // avoid intermediate merges (waste of time with HNSW?)
     mp.setSegmentsPerTier(256); // only merge once at the end when we ask
     iwc.setMergePolicy(mp);
-    String fieldName = "field";
-    VectorSimilarityFunction similarityFunction = VectorSimilarityFunction.DOT_PRODUCT;
```
cpoerschke, line 50: Could we retain use of existing similarity function? And if alternative is needed perhaps have that for an additional field?

lucene/core/src/test/org/apache/lucene/document/TestManyKnnDocs.java
```diff
@@ -46,27 +54,139 @@ public void testLargeSegment() throws Exception {
...
+    VectorSimilarityFunction similarityFunction = VectorSimilarityFunction.EUCLIDEAN;
```
cpoerschke, line 57: Could we retain use of existing similarity function? And if alternative is needed perhaps have that for an additional field?
```suggestion
    VectorSimilarityFunction similarityFunction = VectorSimilarityFunction.DOT_PRODUCT;
```

lucene/core/src/test/org/apache/lucene/document/TestManyKnnDocs.java
```diff
@@ -46,27 +54,139 @@ public void testLargeSegment() throws Exception {
...
-      Document doc = new Document();
-      doc.add(new KnnFloatVectorField(fieldName, vector, similarityFunction));
       for (int i = 0; i < numVectors; i++) {
+        float[] vector = new float[128];
+        Document doc = new Document();
         vector[0] = (i % 256);
+        vector[1] = (float) (i / 256.);
+        doc.add(new KnnFloatVectorField("field", vector, similarityFunction));
+        doc.add(new KeywordField("int", "" + i, org.apache.lucene.document.Field.Store.YES));
+        doc.add(new StoredField("intValue", i));
```
cpoerschke, line 70: Could you help me understand this change? `new float[1]` to `new float[128]` sort of jumps out.

lucene/core/src/java/org/apache/lucene/search/SeededKnnCollector.java
```diff
@@ -0,0 +1,73 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.search;
+
+/** A {@link KnnCollector} that provides an initial set of seeds to initialize the search. */
+class SeededKnnCollector implements KnnCollector {
```
cpoerschke, line 20: minor/subjective: could potentially factor out a `FilterKnnCollector` class (similar to `FilterCollector`) either in addition to or instead of `SeededKnnCollector` e.g.

```
class FilterKnnCollector implements KnnCollector
class SeededKnnCollector extends FilterKnnCollector
```

or

```
class FilterKnnCollector implements KnnCollector

if (seedDocs != null) {
  knnCollector = new FilterKnnCollector(knnCollector) {
    @Override
    public DocIdSetIterator getSeedEntryPoints() {
      return seedDocs;
    }
  };
}
```

seanmacavaney:
Thanks all! Here is a summary of the design:

1. Addition of `seed` to `*KnnVectorQuery`. This is the main user-facing API.
2. Addition of methods to convert between docids and ordinals in `*KnnVectorQuery` and `*VectorValues`. A new inner class `MappedDISI` facilitates this conversion process.
3. A new `SeededKnnCollector` with a `getSeedEntryPoints` method to faciliate passing the seed ordinals down to the `HnswGraphSearcher` without affecting the `LeafReader`.

Is this better? If so, I want to go ahead and re-run some benchmarks to make sure that performance didn't degrade with the redesign.

lucene/core/src/java/org/apache/lucene/search/SeededKnnCollector.java
seanmacavaney: I like that, thanks!

lucene/core/src/test/org/apache/lucene/document/TestManyKnnDocs.java
seanmacavaney: It was just for testing before I had a proper benchmark set up. I switched back to [1].

lucene/core/src/java/org/apache/lucene/search/SeededKnnCollector.java
seanmacavaney: Alright, I've added a generic `KnnCollector.Decorator`, which applies in several places. I stopped short of applying filtering via this same mechanism, since that would ultamately change the `LeafCollector`.

lucene/core/src/java/org/apache/lucene/search/KnnCollector.java
```diff
@@ -85,4 +85,95 @@ public interface KnnCollector {
...
+  }
+
+  /**
+   * KnnCollector.Decorator is the base class for decorators of KnnCollector objects, which extend
+   * the object with new behaviors.
+   *
+   * @lucene.experimental
+   */
+  public abstract static class Decorator implements KnnCollector {
+    private KnnCollector collector;
```
cpoerschke, line 109:
```suggestion
    final private KnnCollector collector;
```

lucene/core/src/java/org/apache/lucene/search/KnnCollector.java
```diff
@@ -85,4 +85,95 @@ public interface KnnCollector {
...
+    }
+  }
+
+  /**
+   * KnnCollector.Seeded is a KnnCollector decorator that replaces the seedEntryPoints.
+   *
+   * @lucene.experimental
+   */
+  public static class Seeded extends Decorator {
+    private DocIdSetIterator seedEntryPoints;
```
cpoerschke, line 167:
```suggestion
    final private DocIdSetIterator seedEntryPoints;
```

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -83,14 +94,29 @@ public Query rewrite(IndexSearcher indexSearcher) throws IOException {
...
+      }
+      Query seedRewritten = indexSearcher.rewrite(booleanSeedQueryBuilder.build());
+      seedWeight = indexSearcher.createWeight(seedRewritten, ScoreMode.TOP_SCORES, 1f);
+    } else {
+      seedWeight = null;
+    }
+
     TimeLimitingKnnCollectorManager knnCollectorManager =
         new TimeLimitingKnnCollectorManager(
             getKnnCollectorManager(k, indexSearcher), indexSearcher.getTimeout());
```
benwtrent, line 114: So, I was actually thinking of a new collector manager as well that would:

 - delegate to other managers as necessary
 - Build the seed weight & run the query when it builds the individual collectors (since it has access to the leaf context). 

This helps abstract away all the messing about with the seed stuff into 

`SeededKnnCollectorManager` which will then return `SeededKnnCollectors` that again, delegate as necessary.

We don't need to really update anything directly in any of the queries as all the new logic could live in these two new classes.

This will make the change much more contained, and keep us from having to update the `approximateSearch` API as the query classes are not experimental (and thus subject to bwc).

lucene/core/src/java/org/apache/lucene/index/ByteVectorValues.java
```diff
@@ -64,6 +65,18 @@ public static void checkField(LeafReader in, String field) {
...
+  /**
+   * Returns a new iterator that maps the provided docIds to the vector ordinals.
+   *
+   * <p>This method assumes that all docIds have corresponding ordinals.
+   *
+   * @lucene.internal
+   * @lucene.experimental
+   */
+  public DocIdSetIterator convertDocIdsToVectorOrdinals(DocIdSetIterator docIds) {
+    return docIds;
```
cpoerschke, line 77: Wondering if this could move to the `KnnVectorValues` base class and be abstract and/or if a default implementation here (and in `FloatVectorValues`) should be `throw new UnsupportedOperationException();` i.e. a mapping needs to be provided by the implementations of this abstract class?

benwtrent:
@seanmacavaney I realize I am making some broad API suggestions. If it would help, I can attempt to convey my ideas in a branch from your fork.

I do like the direction now by providing something via the collectors. This allows us to keep the query objects effectively unchanged (which is sensitive to BWC).

lucene/core/src/java/org/apache/lucene/index/ByteVectorValues.java
benwtrent: FWIW, I am against default implementations that throw, just require the interface and thus catch it at compile time.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -156,6 +196,49 @@ private TopDocs getLeafResults(
...
+              leafCollector,
+              leafReader.getLiveDocs(),
+              0 /* min */,
+              DocIdSetIterator.NO_MORE_DOCS /* max */);
+        }
+        leafCollector.finish();
+      } catch (
+          @SuppressWarnings("unused")
+          CollectionTerminatedException e) {
+      }
```
cpoerschke, line 226:
```suggestion
      } catch (
          @SuppressWarnings("unused")
          CollectionTerminatedException e) {
      }
      leafCollector.finish();
```

lucene/core/src/java/org/apache/lucene/search/KnnCollector.java
cpoerschke: or could be `protected` if deriving classes might need access
```suggestion
    final protected KnnCollector collector;
```

benwtrent:
@seanmacavaney here is my idea more fully: https://github.com/apache/lucene/compare/main...benwtrent:lucene:seeds-refactor-idea?expand=1

I didn't create a byte version of the query, nor am I handling IOExceptions correctly, but this should better show what I had in mind. 

All the code should be able to be mostly encapsulated into the collector & collector manager with no changes to either query.

seanmacavaney:
Thanks- I like the design. I'll try to make the remaining changes accordingly.

cpoerschke:
> @seanmacavaney here is my idea more fully: https://github.com/apache/lucene/compare/main...benwtrent:lucene:seeds-refactor-idea?expand=1
> ...

Wondering if the

```
public SeededKnnFloatVectorQuery(String field, float[] target, int k, Query filter, Query seed) {
```

then potentially makes it easier to have a seed count value too i.e. it could be different from `k` value?

```
public SeededKnnFloatVectorQuery(String field, float[] target, int k, Query filter, Query seed, int seed_count) {
```

seanmacavaney:
Good call!

benwtrent:
@seanmacavaney do you still want this contributed to Apache Lucene? Its excellent work and I don't want it dying on the vine.

If its ok with you, I plan on refactoring it (with attribution to you of course), and attempting to get it into Lucene as a new query. Seeding entry points for HNSW exploration (and not just HNSW, but other vector indices as well), is powerful and can provide some unique interactions with vector search.

seanmacavaney:
Hey @benwtrent -- it's been on my todo list to get back to this, but I've gotten bogged down with a bunch of other stuff.

If you're willing, please do go ahead and refactor :D! (I'd also love to see how the refactor looks to better understand Lucene internals.)

seanmacavaney:
closing due to #14084

PR #13639, opened 2024-08-09
These attributes are better for the final state(#13628)（#13630）

mrhbj:


mrhbj:
@gsmiller thanks for your advise. I think you are right. So I do this. Could you please review it？

gsmiller:
@mrhbj merged, thanks!

PR #13641, opened 2024-08-09
Unify how missing field entries are handle in knn formats

### LINKED ISSUE #13626: Lucene99FlatVectorsReader.getFloatVectorValues(): NPE: Cannot read field "vectorEncoding" because "fieldEntry" is null
### Description

One of our internal users hit this error when merging their index after loading all documents which contain some vector fields.  I couldn't reproduce this myself.  Is there a corner case which has triggered this?

```
java.lang.NullPointerException: Cannot read field "vectorEncoding" because "fieldEntry" is null
	at org.apache.lucene.codecs.lucene99.Lucene99FlatVectorsReader.getFloatVectorValues(Lucene99FlatVectorsReader.java:234) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
	at org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsReader.getFloatVectorValues(Lucene99HnswVectorsReader.java:240) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
	at org.apache.lucene.codecs.KnnVectorsWriter$MergedVectorValues.mergeFloatVectorValues(KnnVectorsWriter.java:158) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
	at org.apache.lucene.codecs.lucene99.Lucene99FlatVectorsWriter.mergeOneFieldToIndex(Lucene99FlatVectorsWriter.java:312) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
	at org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsWriter.mergeOneField(Lucene99HnswVectorsWriter.java:349) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
	at org.apache.lucene.codecs.KnnVectorsWriter.merge(KnnVectorsWriter.java:99) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
	at org.apache.lucene.index.SegmentMerger.mergeVectorValues(SegmentMerger.java:257) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
	at org.apache.lucene.index.SegmentMerger.mergeWithLogging(SegmentMerger.java:300) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
	at org.apache.lucene.index.SegmentMerger.merge(SegmentMerger.java:151) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
	at org.apache.lucene.index.IndexWriter.mergeMiddle(IndexWriter.java:5293) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
	at org.apache.lucene.index.IndexWriter.merge(IndexWriter.java:4761) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
	at org.apache.lucene.index.IndexWriter$IndexWriterMergeSource.merge(IndexWriter.java:6582) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
	at org.apache.lucene.index.SerialMergeScheduler.merge(SerialMergeScheduler.java:38) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
	at org.apache.lucene.index.IndexWriter.executeMerge(IndexWriter.java:2326) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
	at org.apache.lucene.index.IndexWriter.maybeMerge(IndexWriter.java:2321) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
	at org.apache.lucene.index.IndexWriter.forceMerge(IndexWriter.java:2162) ~[lucene-core-9.11.1.jar:9.11.1 0c087dfdd10e0f6f3f6faecc6af4415e671a9e69 - 2024-06-23 12:31:02]
```	

This is with Lucene 9.11.1.

benwtrent:
It is possible to inappropriately use the knn formats and attempt to merge segments with mismatched field names.

None of the formats actually check for `null`, they just all assume that the field entry exists. 

I opted to throw an IllegalArgumentException here to keep with the current behavior, but prevent an unexpected NPE.

Additionally, this unifies the scalar quantized formats, these checked for null values and returned null. But we should behave uniformly across all formats. 

closes: https://github.com/apache/lucene/issues/13626


msokolov:
> It is possible to inappropriately use the knn formats and attempt to merge segments with mismatched field names.

I don't understand this - what's inappropriate about it? I guess we have something like segment 1 has field A and segment 2 has field B. Shouldn't the merged segment have both fields?

benwtrent:
@msokolov maybe?

Looking at the code, this has never worked this way. Field entry existence was just assumed to be handled by the perfield codec.

I tried digging into the other formats that have an "per field" version (e.g. docValues,) and when I tested, they also threw a NPE in a scenario like this.

So, I opted to keep what seems like the status quo. 

But, I am happy to change it if I am wrong about intended behavior.

lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseKnnVectorsFormatTestCase.java
```diff
@@ -230,6 +236,120 @@ public void testIllegalDimChangeTwoWriters() throws Exception {
...
+      if (codec.knnVectorsFormat() instanceof PerFieldKnnVectorsFormat perFieldKnnVectorsFormat) {
+        final KnnVectorsFormat format =
+            perFieldKnnVectorsFormat.getKnnVectorsFormatForField("field");
+        iwc.setCodec(
+            new FilterCodec(codec.getName(), codec) {
+              @Override
+              public KnnVectorsFormat knnVectorsFormat() {
+                return format;
+              }
+            });
```
jpountz, line 297: Does it actually work? I would expect the read logic to look up the codec by name and to get the original codec back rather than your FilterCodec.

jpountz (commented):
+1 on removing the leniency, it is illegal to call the KnnVectorsReader on a field that doesn't have vectors enabled, which is why CodecReader performs a FieldInfo check

To @msokolov 's point, we need to fix the merging logic by protecting calls to the vectors reader with checks on the field info.

Separately, it looks like we call the KnnVectorsReader with the merged field info, which is not legal, we should call it with the field info of the segment.

lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseKnnVectorsFormatTestCase.java
benwtrent: it indeed works we run into an IAE as expected on merge. I can validate the exception method for further validation.

jpountz:
> Separately, it looks like we call the KnnVectorsReader with the merged field info, which is not legal, we should call it with the field info of the segment.

Apologies, I was wrong here: KnnVectorsReader takes a String (rather than a FieldInfo) so it's not subject to this problem.

benwtrent:
Unless there are rejections, I am going to move forward with merging this. Throwing an IAE seems way better than throwing an NPE. 

If we want to change this behavior to be more lenient, that is fine, but I would rather fix this first and then we can figure out what the improved behavior should be.

benwtrent:
> Fine with me, I will work on fixing the merging logic then unless you plan on working on it.



Ah dang, yeah, I completely spaced on that since this PR is so old. I will do that right now & we can do that in this PR.

benwtrent:
OK, I pushed up a fix for @msokolov's suggestion and your reminder @jpountz . I added a protection around our access to the KnnVectorReaders during merge. Additionally, I kept my new test, but these protections keep it from throwing (previously it did). This way we unify our access logic on missing (e.g. throw), and protect during merge by not attempting to access vector values that aren't there for a given field.

lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseKnnVectorsFormatTestCase.java
benwtrent: So, this test would have the merge throw on missing fields with this change. Now, we protect against this and this test no longer throws on missing fields.

bugmakerrrrrr:
@benwtrent should we do the same check for `PerFieldKnnVectorsFormat.FieldsReader`? If the target reader is null, throw an IAE and ensure that we do not return null.

benwtrent:
@bugmakerrrrrr I don't want to make that change as the PerFieldDocValuesFormat & PerFieldPostingsFormat field readers will return `null` if the field is missing. I want to at least try to unify the `PerFieldKnnVectorsFormat` with those

However, we are fairly inconsistent all over the place here. For example, in our Postings formats, we will check for `null` and return null, but for doc value formats, those look like they will throw NPEs. 

Doc value exmaple:

```
  @Override
  public NumericDocValues getNumeric(FieldInfo field) throws IOException {
    NumericEntry entry = numerics.get(field.number);
    return getNumeric(entry);
  }
```

Postings:
```
  @Override
  public Terms terms(String field) throws IOException {
    assert field != null;
    FieldInfo fieldInfo = fieldInfos.fieldInfo(field);
    return fieldInfo == null ? null : fieldMap.get(fieldInfo.number);
  }
```

lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseKnnVectorsFormatTestCase.java
```diff
@@ -230,6 +236,108 @@ public void testIllegalDimChangeTwoWriters() throws Exception {
     }
   }
 
+  @SuppressWarnings("unused")
```
jpountz, line 239: why do we need this suppresswarnings annotation?

lucene/codecs/src/test/org/apache/lucene/codecs/simpletext/TestSimpleTextKnnVectorsFormat.java
```diff
@@ -41,4 +41,14 @@ public void testRandomBytes() throws Exception {
...
+
+  @Override
+  public void testMergingWithDifferentKnnFields() {
+    // unimplemented
+  }
+
+  @Override
+  public void testMergingWithDifferentByteKnnFields() {
+    // unimplemented
+  }
```
jpountz, line 53: I'm not clear why SimpleText (and some other codecs) needs to disable these tests?

benwtrent: Simple text and per field are left over from before me fixing vector merging. For the others, only the `byte` version is overridden because those older formats didn't support bytes.

Will push up a commit soon to make sure perfield and simple run this test.

jpountz (approved):

Issue #13640, opened 2024-08-09
testMergeStability failing for Knn formats


benwtrent:
### Description

All KNN formats are periodically failing `testMergeStability`. 

I have verified its due to https://github.com/apache/lucene/pull/13566 

The stability failure is due to a different size in the `vex` (e.g. vector graph connections).

### Gradle command to reproduce

```
./gradlew test --tests TestLucene99HnswScalarQuantizedVectorsFormat.testMergeStability -Dtests.seed=84298DFFA7C134B7 -Dtests.locale=kn -Dtests.timezone=Hongko -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```

benwtrent:
@msokolov ^ I haven't been able to look into fixing it yet. Just now noticed it.

msokolov:
hmm thanks I'll take a look soon

msokolov:
I didn't know about this constraint until now. Basically what happens is during merge we check for disconnected components and attempt to *add* connections to connect them. So it makes sense we might be adding some bytes to the vex file.  Maybe a way to avoid this is to skip recreating the HNSW graph when merging a single segment. Honestly I don't know why we would be doing that.  I'll dig a bit more/

msokolov:
OK, I guess we would need to actually build a graph when merging a single segment in case there are deletions. In any case it would be nice if the graph reconnection were stable. This test exposes some interesting problems! Yay for our tests.

msokolov:
Apparently that patch did not fix all the things; this failure got generated on that patched version:

```
   gradlew test --tests TestPerFieldKnnVectorsFormat.testMergeStability -Dtests.seed=A6835D6A0735B851 -Dtests.multiplier=3 -Dtests.locale=ar-IL -Dtests.timezone=Europe/Berlin -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```

msokolov:
found and fixed a branch_9x-only problem. Hopefully this calms down now

benwtrent:
Not to be the bearer of bad news: 
```
./gradlew test --tests TestLucene95HnswVectorsFormat.testMergeStability -Dtests.seed=3D48DB65BA75CD03 -Dtests.locale=uz-Latn-UZ -Dtests.timezone=Europe/Tirane -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```
Fails on main.

[...]

PR #13642, opened 2024-08-10
Two fixes for recently-added HnswGraphBuilder.connectComponents:

msokolov:
1. properly set frozen flag to avoid re-duplicative work
2. don't try to join a node to itself


msokolov:
addresses #13640 

with this fix the failing test seed on that issue succeeded. I re-ran the test with tests.iters=1000 and didn't get any fails

PR #13648, opened 2024-08-13
gh-13640: Connect hnsw graph components when flushing

msokolov:
It turns out that on branch9x we were not calling connectComponents on flush, only on merge, so the graph would grow when merging violating the stability test


msokolov:
addresses #13640

PR #13649, opened 2024-08-13
Fix eclipse ide settings generation

### LINKED ISSUE #13638: `gradlew eclipse` no longer works
### Description

Hi @dweiss,
since your last cleanup, when trying to run `gradlew eclipse` it complains because the task does not exist. Looks ike the reason is that the root project no longer is a Java project, so the plugin is not applied. When commenting out the `plugins.withType(JavaPlugin)` it breaks with an NPE.

No sure how to fix this. We need the eclipse plugin ONLY on root project because we use it to create a global project.

_No response_

dweiss:
A fix to https://github.com/apache/lucene/issues/13638.

I don't see how the refactoring in https://github.com/apache/lucene/pull/13484 had something to do with the ide setup breaking. There are some things we do in the build that make it run a bit unpredictably - accessing root project properties when they're not yet set, for example. 

I've tested this PR on a mac and the newest eclipse distribution and everything seems to work.

### DIFF
--- a/gradle/validation/dependencies.gradle
+++ b/gradle/validation/dependencies.gradle
@@ -82,8 +82,8 @@ configure(rootProject) {
     ]
   }.configureEach {
     it.doLast {
-      File inFile = it.catalogFile.get().asFile
-      ant.fixcrlf(srcdir: inFile.getParent(), includes: inFile.getName(), eol: "lf", fixlast: "true", encoding: "UTF-8")
+      ant.fixcrlf(file: it.catalogFile.get().asFile,
+        eol: "lf", fixlast: "true", encoding: "UTF-8")
     }
   }
 
--- a/gradle/validation/dependencies.gradle
+++ b/gradle/validation/dependencies.gradle
@@ -83,11 +83,7 @@ configure(rootProject) {
   }.configureEach {
     it.doLast {
       File inFile = it.catalogFile.get().asFile
-      String text = inFile.getText()
-      String cleaned = text.replaceAll("\r\n", "\n")
-      if (!cleaned.equals(text)) {
-        inFile.setText(cleaned, "UTF-8")
-      }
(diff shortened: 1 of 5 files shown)


uschindler:
Hi,
The main problem was that it wasn't able to find the task "eclipse". You hacked around this by converting the root project to a java project.

I have no idea how it was before the change in your refactoring, but it looks like for some reason the root project had the java plugin loaded.

Would it be enough to load the Java-Base plugin in root project, too.

Uwe

dweiss:
> Hi, The main problem was that it wasn't able to find the task "eclipse". You hacked around this by converting the root project to a java project.

Well, I didn't hack around it - it must have been a Java project before too... but I also don't see what was applying the java project to the root project, which is a mistery to me! 

> Would it be enough to load the Java-Base plugin in root project?

I don't think it's going to work - the jdt extension isn't visible then. I may look at it again tomorrow.

uschindler (approved):

gradle/validation/dependencies.gradle
```diff
@@ -75,6 +75,22 @@ configure(rootProject) {
     it.dependsOn(":versionCatalogFormatDeps")
   }
 
+  // correct crlf/ default encoding after version catalog formatting finishes.
```
uschindler, line 78: Small nit: Can't we use the fixcrlf task of Ant like we do when autogenerating files?

dweiss: Sure. I just wasn't sure about the default encoding that ant reads those files with, but I don't think it'll matter - it's plain ascii anyway.

uschindler: https://github.com/apache/lucene/blob/97e30fa487d2133fc43d7180d34c5605e3805926/gradle/generation/jflex.gradle#L188

uschindler: It also uses file property not parent dir with pattern.

uschindler: I changed it. Works as expected, not Git warnings anymore.

uschindler:
Works all fine, merging now. Thanks!

PR #13651, opened 2024-08-13
Add a Better Binary Quantizer format for dense vectors

### LINKED ISSUE #13650: Add higher quantization level for kNN vector search
### Description

There have been multiple discussions around supporting higher vector quantization levels (e.g. more than scalar quantization can provide).

We (elastic) have been doing some research around what the best next step will be for higher levels of quantization in Lucene. 

A couple of months ago we came across an interesting paper: https://arxiv.org/abs/2405.12497 and have been actively experimenting with RaBitQ. 

Its approach to quantization fits well with Lucene:

 - Higher levels of compression (32x for floating point)
 - Very cheap quantization costs
 - Fast vector similarities
 - Lower levels of reranking required

In our investigations we have found:

 - RaBitQ achieves the same or better recall at the same compression ratios as Product Quantization.
 - RaBitQ is 20-30x faster at quantizing
 - RaBitQ (in Java) is more than 2x faster to query than PQ (can likely be made faster, where PQ cannot really...)

Let me share some PoC numbers. All our testing with PQ assumed a 32x reduction. Higher levels of reduction then require more reranking, which then becomes the main bottleneck. Our goal was to achieve good recall at only 5x reranking and without Optimized PQ (which is insanely expensive, the PQ already is very expensive).

30M cohereV3 (on a GCP intel machine avx512):

- RaBitQ:
  - Time to quantize vectors: `998363ms`
  - Avg Brute-force query time for Recall@10:50 `1776ms` with 98% recall
  - Time to build HNSW graph: 40043229ms
     - HNSW recall@10:50: 86.2% at 1641qps
- PQ:
  - Time to calculate codebooks: `472260ms`
  - Time to quantize vectors: `12644293ms`
  - Avg brute-force query time for Recall@10:50 `5790ms` with recall 98% recall.
  - WE didn't bother building the large graph here as, frankly, its PQ is just too slow.

522k Quora encoded with e5Small (noted for being horrible with binary quantization & higher compression ratios in general) (On my macbook, Arm M1-MAx)

- RaBitQ:
  - 1051ms to quantize vectors
  - Avg Brute-force query time for Recall@10:50 `11ms` with 99% recall
- PQ:
  - 28360ms to build the code books
  - 31364ms to encode the docs
  - Avg Brute-force query time for Recall@10:50 `19ms` with 99% recall

Consequently, given all the various trade-offs and benefits, we are working on bringing a Rabitq format into Lucene.

I will open a draft PR soon to make the work more visible.

benwtrent:
# Highlevel design

RaBitQ is basically a better binary quantization, which works across all models we have tested against. Like PQ, it does require coarse grained clustering to be effective at higher vector densities (effective being defined as only requiring 5x or lower oversampling for recall>95%). But in our testing, the number of vectors required per cluster can be exceptionally large (10s to 100s of millions). 

The euclidean vectors as stored in the index:

| quantized vector | distance_to_centroid | vector magnitude|
| - | - |- |
| (vector_dimension/8) bytes | float | float |

For dot-product vectors:

| quantized vector | vector dot-product with binarized self | vector magnitude | centroid dot-product |
| - | - |- | - |
| (vector_dimension/8) bytes | float | float | float |

The vector metadata, in addition to all the regular things (similarity, encoding, sparse vector DISI, etc.).

For indexing into HNSW we actually have a multi-step process. Better binary encodes the query vectors different than the index vectors. Consequently, during segment merge & HNSW building, another temporary file is written containing the query quantized vectors over the configured centroids. One downside is that this temporary file will actually be larger than the regular vector index. This is because we use asymmetric quantization to keep good information around. But once the merge is complete, this file is deleted.

We then read from the query temporary file when adding a vector to the graph and when exploring HNSW, we search the indexed quantized values.

closes: https://github.com/apache/lucene/issues/13650


mayya-sharipova:
> possibly switch to LongValues for storing vectorOrd -> centroidOrd mapping

I was thinking about adding centroids mappings as LongValues at the end of meta file, but this could potentially make meta file quite large (for 100M docs, we would need extra 100Mb). We really try to keep meta files small, so I would prefer either:
- to keep the current approach (add a byte at the end of each vector in the vectors file). Indeed, this will throw off paging size.
- adding an extra file for centroids mapping

What do you think?

benwtrent:
100MB assumes that even when compressed, it's a single byte per centroid. 100M vectors might only have 2 centroids and thus only need two bits two store.

Also, I would expect the centroids to be at the end of the "veb" file, not metadata. Like we do already for the sparse vector ord to doc resolution.

But, either solution needs testing for sure.

lucene/core/src/java21/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java
```diff
@@ -761,4 +763,81 @@ private static int squareDistanceBody128(MemorySegment a, MemorySegment b, int l
     // reduce
     return acc1.add(acc2).reduceLanes(ADD);
   }
+
+  @Override
+  public long ipByteBinByte(byte[] q, byte[] d) {
+    if (VECTOR_BITSIZE == 128) {
```
rmuir, line 769: @ChrisHegarty I'm confused about this, what if machine has avx-256? Also since this uses integers there must be a check for `PanamaVectorConstants.HAS_FAST_INTEGER_VECTORS`. on machines without avx, `VECTOR_BITSIZE` of 128 will be claimed, but any operations will be 30x slower.

lucene/core/src/java21/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java
```diff
@@ -761,4 +763,81 @@ private static int squareDistanceBody128(MemorySegment a, MemorySegment b, int l
...
+      sumShort2 = acc1.reinterpretAsShorts().lanewise(VectorOperators.LSHR, 8);
+      subRet1 += sumShort1.add(sumShort2).reduceLanes(VectorOperators.ADD);
+
+      sumShort1 = acc2.reinterpretAsShorts().and((short) 0xFF);
+      sumShort2 = acc2.reinterpretAsShorts().lanewise(VectorOperators.LSHR, 8);
+      subRet2 += sumShort1.add(sumShort2).reduceLanes(VectorOperators.ADD);
+
+      sumShort1 = acc3.reinterpretAsShorts().and((short) 0xFF);
+      sumShort2 = acc3.reinterpretAsShorts().lanewise(VectorOperators.LSHR, 8);
+      subRet3 += sumShort1.add(sumShort2).reduceLanes(VectorOperators.ADD);
```
rmuir, line 827: is there any sneaky way to prevent having to call typically-costly `reduceLanes` 4 times?

lucene/core/src/java21/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java
ChrisHegarty: Yeah, the implementation only works on ARM at the minute. I'm iterating on an x64 implementation, but haven't found the sweet spot yet. I'll merge something soon and share the jmh benchmarks.

lucene/core/src/java21/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java
ChrisHegarty: reduceLanes is indeed costly, but is only done once per 2048 elements of the dimension of each vector, so is typically not too bad in practice. I'll add some better docs/comments to help describe this.

lucene/core/src/java21/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java
rmuir: toss a floating point addition to a static variable in there? :)

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912BinaryFlatVectorsScorer.java
```diff
@@ -0,0 +1,317 @@
...
+    private final BinaryQueryVector[] queryVectors;
+    private final RandomAccessBinarizedByteVectorValues targetVectors;
+    private final VectorSimilarityFunction similarityFunction;
+
+    private final int discretizedDimensions;
+    private final float sqrtDimensions;
+    private final float maxX1;
+
+    public BinarizedRandomVectorScorer(
+        BinaryQueryVector[] queryVectors,
```
mayya-sharipova, line 162: @benwtrent Why are we accepting an array of `queryVectors` and only using a single, first of them?

benwtrent: Because I didn't add all the logic for multiple clusters

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912BinaryFlatVectorsScorer.java
```diff
@@ -0,0 +1,317 @@
...
+    return nonQuantizedDelegate.getRandomVectorScorerSupplier(similarityFunction, vectorValues);
+  }
+
+  @Override
+  public RandomVectorScorer getRandomVectorScorer(
+      VectorSimilarityFunction similarityFunction,
+      RandomAccessVectorValues vectorValues,
+      float[] target)
+      throws IOException {
+    if (vectorValues instanceof RandomAccessBinarizedByteVectorValues binarizedQueryVectors) {
```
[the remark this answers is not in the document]
benwtrent: Copy paste error

lucene/core/src/java/org/apache/lucene/codecs/lucene912/BinarizedByteVectorValues.java
```diff
@@ -0,0 +1,78 @@
...
+    return clusterId;
+  }
+
+  public abstract float getMagnitude() throws IOException;
+
+  public abstract float getOOQ() throws IOException;
+
+  public abstract float getNormOC() throws IOException;
+
+  public abstract float getODotC() throws IOException;
```
benwtrent, line 52: I would really like if we just did `float[]` and the callers (which are scorers) know how to use it. The downside now is that if something calls `getODotC()` but it wasn't available, this would blow up right?

john-wagster: Was naively following the pattern to get things working.  This had getMagnitude and getDistanceToCentroid in it prior to adding getOOQ, NormOC, and ODotC.  For MIP it only needs OOQ, NormC, and ODotC and essentially ignores (actually overlaps with the other values or 0f).  So I don't think it blows up as is but it is storing an extra unnecessary float for Euclidean now.  But I could be missing something.  

Two thoughts I had and hadn't quite gotten to yet.  I think I agree with your thoughts here that this should deal with arbitrary float (or byte) values.  I think where we had last landed was some way to decode those values on both read and write where the Scorer? or Format? (via reflection? to avoid versioning problems?) would supply to the Reader and Writer how to serialize and deserialize appropriately based on the similarity function a set of bytes which would eliminate some overhead there (1 byte for Euclidean).   I would buy we could update these interfaces to deal with an array of floats in the meantime depending on if quantization of these corrective factors is not a high priority; I was operating right now under the assumption that we do want to try to refactor to store a minimum overhead for each vector (2 bytes / vector in the "Euclidean" and 3 bytes in the "MIP" use-case).

Second thought was I think for all of these values we want to swap them out with 1 byte quantized forms and without having started that work yet it seems pretty straightforward in my head to do so, which is probably why I'm seeing it as part of this first pass.  

Was thinking to do both of things together and refactor this to return a getCorrectiveBytes instead of these individual corrective factors and put the burden on the caller to decode them.

There is a separate place where specifically for query (not indexing) instead of pulling through a set of corrective float[] factors we are pulling through an object QueryFactors now, which may be is worth discussing further (but is a different place in the code)?  I did this because we could optimize one of the floats to a short and with the belief that we could / want to optimize the other values as well (as well as greatly improving readability).  Although subsequent conversations about those with the team suggest that we don't care about quantizing these values (at least not as much) because the query quantization that gets serialized is only serialized to a temporary file.  I digress.

benwtrent:
@tanyaroosta we are still doing larger scale testing, but if you want to test with LuceneUtil, here is the branch I am using: https://github.com/mikemccand/luceneutil/compare/main...benwtrent:luceneutil:bbq?expand=1

2-5x oversampling is required with HNSW to get good recall. 

The original paper does periodic rescoring with the raw vectors which effectively means you need random access with raw vectors, even when NOT using HNSW. This just doesn't scale when trying to reduce memory usage.

Instead, we are opting for a more traditional approach of oversampling and reranking.

ShashwatShivam:
Following up on the above comment by tanyaroosta, the dataset I was using for benchmarking RaBitQ through Luceneutil (main branch) was amazon's ASIN and query embeddings (which are 256 dim float vectors). Here are three sets of results I got:

**Normal HNSW**
```
Results:
recall  latency (ms)     nDoc  topK  fanout  maxConn  beamWidth  quantized  index s  force merge s  num segments  index size (MB)
 0.721         0.128  1439443    10       6       32         40         no    81.51          42.58             1          1455.25
 0.828         0.166  1439443    10       6       64        100         no   202.30          78.03             1          1472.66
 0.886         0.202  1439443    10       6       64        250         no   545.83         242.29             1          1493.95
 0.914         0.227  1439443    10       6       64        500         no  1111.43         452.62             1          1511.73
```

**RaBitQ HNSW (same parameters) - codec Lucene912HnswBinaryQuantizedVectorsFormat**
```
Results:
recall  latency (ms)     nDoc  topK  fanout  maxConn  beamWidth  quantized  index s  force merge s  num segments  index size (MB)
 0.448         0.095  1439443    10       6       32         40         no    86.15          45.45             1          1526.72
 0.522         0.130  1439443    10       6       64        100         no   195.14          89.08             1          1555.16
 0.542         0.167  1439443    10       6       64        250         no   524.03         207.12             1          1589.50
 0.547         0.192  1439443    10       6       64        500         no  1035.65         415.44             1          1618.38
``` 

**Only RaBitQ - codec Lucene912BinaryQuantizedVectorsFormat**
```
Results:
recall  latency (ms)     nDoc  topK  fanout  maxConn  beamWidth  quantized  index s  force merge s  num segments  index size (MB)
 0.000        87.658  1439443    10     100       64         40         no    12.39           8.01             1          1471.69
 0.000        87.830  1439443   100     100       64         40         no    12.53           7.78             1          1471.69
```

The first issue we're facing is the large recall reduction for RaBitQ HNSW when compared to pure HNSW. Is this recall regression expected? Secondly, the pure RaBitQ implementation returns 0 recall, which is definitely suspect. Perhaps there is a bug on my end when trying to benchmark it using luceneutil?

@benwtrent I'll also try working with your branch of luceneutil to see if it changes the results

benwtrent:
@ShashwatShivam so, the flat codec version is sneaky, depending on when you cloned the repo, it might not be doing anything 


Lucene by default will return nothing for approx kann over a flat index (I think this should be fixed).

But I think the latest BBQ branch should have this patched. I can check in a little while

benwtrent:
Here is some more flat index test results. This was to exercise and see how the number of coarse grained centroids changes recall & speed. 

| Lucene912BinaryQuantizedVectorsFormat                               | recall@100 | recall@100\|500 | recall@100\|1000 | index time & force-merge | mean search time (brute-force) |
|-------------------------------------------------------|------------|-----------------|------------------|--------------------------|--------------------------------|
| hotpotqa-E5-small (5233329 vectors, 1 centroid)       | 47         | 81              | 90               | 79618ms                  | 177ms                          |
| hotpotqa-E5-small (5233329 vectors, 255 centroids)    | 52         | 86              | 94               | 267522ms                 | 211ms                          |
| hotpotqa-gte-base (5233329 vectors, 1 centroid)       | 66         | 96              | 99               | 214254ms                 | 227ms                          |
| hotpotqa-gte-base (5233329 vectors, 255 centroids)    | 71         | 98              | 99               | 617887ms                 | 275ms                          |
| dbpedia-entity-arctic (4635922 vectors, 1 centroid)   | 56         | 89              | 95               | 199827ms                 | 205ms                          |
| dbpedia-entity-arctic (4635922 vectors, 255 centroid) | 59         | 91              | 96               | 618797ms                 | 220ms                          |


Models used:

 - https://huggingface.co/thenlper/gte-base
 - https://huggingface.co/intfloat/e5-small
 - https://huggingface.co/Snowflake/snowflake-arctic-embed-m

benwtrent:
I am currently working on moving this to Lucene101 format with the bug fixes we discovered in additional testing.

benwtrent:
I will open a PR against Lucene Util to update it to utilize these formats and show y'all some runs with it soon. But The PR is ready for general review.

john-wagster (approved):
LGTM

benwtrent:
Here is some Lucene Util Benchmarking. Some of these numbers actually contradict some of my previous benchmarking for int4. Which is frustrating, I wonder what I did wrong then or now. Or of float32 got faster between then and now :)

Regardless, this shows that bit quantization is generally as fast as int4 search or faster and you can get good recall with some oversampling. Combining with the 32x reduction in space its pretty nice.

The oversampling rates were `[1, 1.5, 2, 3, 4, 5]`. HNSW params `m=16,efsearch=100`. `Recall@100`.

## Cohere v2 1M

| quantization      | Index Time | Force Merge time | Mem Required |
|-------------------|------------|------------------|--------------|
| 1 bit             | 395.18     | 411.67           | 175.9MB      |
| 4 bit (compress)  | 1877.47    | 491.13           | 439.7MB      |
| 7 bit             | 500.59     | 820.53           | 833.9MB      |
| raw               | 493.44     | 792.04           | 3132.8MB     |

cohere-v2-bit-1M

## Cohere v3 1M

1M Cohere v3 1024

| quantization      | Index Time | Force Merge time | Mem Required |
|-------------------|------------|------------------|--------------|
| 1 bit             | 338.97     | 342.61           | 208MB        |
| 4 bit (compress)  | 1113.06    | 5490.36          | 578MB        |
| 7 bit             | 437.63     | 744.12           | 1094MB       |
| raw               | 408.75     | 798.11           | 4162MB       |


cohere-v3-bit-1M

# e5Small

| quantization      | Index Time | Force Merge time | Mem Required |
|-------------------|------------|------------------|--------------|
| 1 bit             | 161.84     | 42.37            | 57.6MB       |
| 4 bit (compress)  | 665.54     | 660.33           | 123.2MB      |
| 7 bit             | 267.13     | 89.99            | 219.6MB      |
| raw               | 249.26     | 77.81            | 793.5MB      |

e5small-bit-500k

lucene/core/src/java/org/apache/lucene/internal/vectorization/VectorUtilSupport.java
```diff
@@ -44,4 +44,7 @@ public interface VectorUtilSupport {
 
   /** Returns the sum of squared differences of the two byte vectors. */
   int squareDistance(byte[] a, byte[] b);
+
+  /* Returns the ... TODO ... */
```
ChrisHegarty, line 48: Maybe just provide a simple high-level javadoc comment here.

lucene/core/src/java/org/apache/lucene/util/VectorUtil.java
```diff
@@ -307,4 +379,16 @@ public static float[] checkFinite(float[] v) {
     }
     return v;
   }
+
+  // FIXME: allow for a settable B_QUERY of 4 or 2
+  public static final short B_QUERY = 4;
+
+  /** Return the .. TODO .. */
+  public static long ipByteBinByte(byte[] q, byte[] d) {
```
ChrisHegarty, line 387: same, remove todo

tteofili (approved):

ChrisHegarty (approved):
LGTM

lucene/core/src/java/org/apache/lucene/codecs/lucene101/Lucene101BinaryQuantizedVectorsFormat.java
```diff
@@ -0,0 +1,125 @@
...
+ * dimensions, and file offset information.
+ *
+ * <ul>
+ *   <li><b>int</b> the field number
+ *   <li><b>int</b> the vector encoding ordinal
+ *   <li><b>int</b> the vector similarity ordinal
+ *   <li><b>vint</b> the vector dimensions
+ *   <li><b>vlong</b> the offset to the vector data in the .veb file
+ *   <li><b>vlong</b> the length of the vector data in the .veb file
+ *   <li><b>vint</b> the number of vectors
```
mayya-sharipova, line 77: Also:

 <li><b>[float]</b> clusterCenter
 <li><b>int</b> dotProduct of clusterCenter with itself

mayya-sharipova (approved):
Amazing work! Thanks Ben and the team!

benwtrent:
Hey @ShashwatShivam https://github.com/mikemccand/luceneutil/compare/main...benwtrent:luceneutil:bbq

that is the testing script I use. 

But if Lucene has since been updated with a 101 codec, I would need to update this branch.

ShashwatShivam:
@benwtrent thanks for giving the link to the testing script, it works! One question - the index size it reports is larger than the HNSW index size. For e.g. I was working with a Cohere 768 dim dataset with 500k docs and the index sizes were 1488.83 MB and 1544.79 MB for HNSW and RaBitQ (Lucene101HnswBinaryQuantizedVectorsFormat) respectively, which seems incorrect. Could you please tell me why this discrepancy occurs, if you've seen this issue before?

benwtrent:
@ShashwatShivam why do you think the index size (total size of all the files) should be smaller?

We store the binary quantized vectors and the floating point vectors. So, I would expect about a 5% increase in disk size just from vectors only.

I have also noticed that the HNSW graph itself ends up being more densely connected, but this is only a marginal increase in disk space as well.

ShashwatShivam:
@benwtrent makes sense, I wasn't accounting for the fact that the floating vectors are being stored too. I guess I should have instead asked how to reproduce the 'memory required' column, which shows a marked reduction for 1 bit quantization v/s raw?

benwtrent:
@ShashwatShivam I don't think there is a "memory column" provided anywhere. I simply looked at the individual file sizes (veb, vex) and summed their sizes together.

ShashwatShivam:
Hey @benwtrent,
Thank you for all your help so far! I have a question about the oversampling used to increase recall. From what I understand, it scales up the top-k and fanout values by the oversampling factor. In the final match set, do we return only the best top-k documents (not scaled up, but the original value)? I couldn't locate the code where the reranking or selection of the best k results from the expanded match set happens. Could you please help me find that part?
Thanks again!

mikemccand:
> @ShashwatShivam I don't think there is a "memory column" provided anywhere. I simply looked at the individual file sizes (veb, vex) and summed their sizes together.

Once this cool change is merged let's fix `luceneutil`'s KNN benchy tooling (`knnPerfTest.py`, `KnnGraphTester.java`) to compute/report the "memory column" ("hot RAM", "searchable RAM", something)?  Basically everything except the original (`float32` or `byte`) vectors.  I'll open an upstream `luceneutil` issue...

benwtrent:
Quick update, we have been bothered with some of the numbers (for example, models like "gist" perform poorly) and we have some improvements to get done first before flipping back to "ready for review". 


@mikemccand YES! That would be great! "Memory required" would be the quantized file size + hnsw graph file size (if the graph exists).



@ShashwatShivam 

Sorry for the late reply. There are no "out of the box" rescoring actions directly in Lucene. Mainly because the individual tools are (mostly) already available to you. You can ask for more overall vectors with one query, and then rescore the individual documents according to the raw vector comparisons. I admit, this requires some Lucene API know how.


It would be good for a "vector scorer" to indicate if its an estimation or not to allow for smarter actions in the knn doc collector...

ShashwatShivam:
I conducted a benchmark using Cohere's 768-dimensional data. Here are the steps I followed for reproducibility:

1. **Set up** the [luceneutil repository](https://github.com/mikemccand/luceneutil/) following the installation instructions provided.
2. **Switch branches** to [this specific branch](https://github.com/mikemccand/luceneutil/compare/main...benwtrent:luceneutil:bbq) since the latest mainline branch is not compatible with the feature needed for this experiment.
3. **Change the branch** of `lucene_candidate` to [benwtrent:feature/adv-binarization-format](https://github.com/benwtrent/lucene/tree/feature/adv-binarization-format) to incorporate advanced binarization formats.
4. **Run** `knnPerfTest.py` after specifying the document and query file paths to the stored Cohere data files. The runtime parameters were set as follows:
   - `nDoc = 500,000`
   - `topk = 10`
   - `fanout = 100`
   - `maxConn = 32`
   - `beamWidth = 100`
   - `oversample` values tested: `{1, 1.5, 2, 3, 4, 5}`
   
   I used `quantizeBits = 1` for RaBitQ+HNSW and `quantizeBits = 32` for regular HNSW.

A comparison was performed between HNSW and RaBitQ, and I observed the recall-latency tradeoff, which is shown in the attached image:  
output.

tanyaroosta:
FYI, a blog post on RaBitQ:

https://dev.to/gaoj0017/quantization-in-the-counterintuitive-high-dimensional-space-4feg

gaoj0017:
Thanks, Tanya @tanyaroosta , for sharing our blog about RaBitQ in this thread. I am the first author of the [RaBitQ paper](https://arxiv.org/abs/2405.12497). I am glad to know that our RaBitQ method has been discussed in the threads here. Regarding the BBQ (Better Binary Quantization) method mentioned in these threads, my understanding is that it majorly follows the framework of RaBitQ and makes some minor modifications for practical performance consideration. The claimed key features of BBQ as described in a blog from Elastic - [Better Binary Quantization (BBQ) in Lucene and Elasticsearch](https://www.elastic.co/search-labs/blog/better-binary-quantization-lucene-elasticsearch) - e.g., normalization around a centroid, multiple error correction values, asymmetric quantization, bit-wise operations, all originate from our RaBitQ paper. 

We note that it is quite often that the industry customizes some methods from academia to better suit their applications, but the industry rarely gives the variant a new name and claim as a new method. For example, the PQ and HNSW methods are from academia and have been widely adopted in the industry with some modifications, but the industry still respects their original names. We believe the same practice should be followed for RaBitQ.

In addition, we would like to share that we have extended RaBitQ to support quantization beyond 1-bit per dimension (e.g., 2-bit, 3-bit, …). The [paper](https://arxiv.org/abs/2409.09913) of the extended RaBitQ was made available in Sep 2024. It achieves so by constructing a larger codebook than that of RaBitQ and can be equivalently understood as an optimized scalar quantization method. For details, please refer to the paper and also a [blog](https://dev.to/gaoj0017/extended-rabitq-an-optimized-scalar-quantization-method-83m) that we have recently posted.

benwtrent:
Closing this PR in deference to this one: https://github.com/apache/lucene/pull/14078

An evolution of scalar quantization proved more flexible and provided better recall in our experiments.

Issue #13653, opened 2024-08-14
TestHnswFloatVectorGraph.testReadWrite fails on branch 9x


benwtrent:
### Description

TestHnswFloatVectorGraph.testReadWrite fails due to a mismatched number of connections. 

`java.lang.AssertionError: arcs differ for node 3 expected:<[2, 52, 6]> but was:<[2, 52, 6, 13]>`

This is caused by commit: https://github.com/apache/lucene/pull/13648

I have confirmed that this failure is not occurring on main, only on 9x.

### Gradle command to reproduce

```
./gradlew test --tests TestHnswFloatVectorGraph.testReadWrite -Dtests.seed=118030F93FB3695D -Dtests.locale=eu -Dtests.timezone=Indian/Cocos -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```

msokolov:
I'm suspecting this may have something to do with the connectComponents changes - I'll dig

msokolov:
OK I'll mark this fixed now since we seem to have gone ~24h without any more failures from automated builds that I can see

PR #13655, opened 2024-08-14
gh-13653: Call finish() from HnswGraphBuilder.build()

msokolov:
This utility method is no longer used by the indexing chain, only by some unit tests, so we missed tracking the change in behavior when merging. Maybe we ought to change the unit test instead? But it makes sense to me that build() should be consistent with what we do when flushing.


benwtrent (approved):
I think this is sane. Though it is used by incremental merger still, which is still a viable and good way to merge graphs :)

I ran some tests 50+ times and didn't get any weird failures. We shall see what CI says. 

Looks like this is just more backporting of the changes in main already.

msokolov:
> I think this is sane. Though it is used by incremental merger still, which is still a viable and good way to merge graphs :)

Oh! well, I have no excuse then :) Thanks for looking

PR #13658, opened 2024-08-15
Speed up prefix sums when decoding doc IDs.

jpountz:
This updates file formats to compute prefix sums by summing up 8 deltas per long at the same time if the number of bits per value is 4 or less, and 4 deltas per long at the same time if the number of bits per value is between 5 included and 11 included. Otherwise, we keep summing up 2 deltas per long like today.

The `PostingDecodingUtil` was slightly modified due to the fact that more numbers of bits per value now need to apply different shifts to the input data. E.g. now that we store integers that require 5 bits per value as 16-bit integers under the hood rather than 8, we extract the first values by shifting by 16-5=11, 16-2*5=6 and 16-3*5=1 and then decode tail values from the remaining bit per 16-bit integer.

Micro benchmarks suggest a noticeable speedup for prefix sums and bits per value 2, 3, 4 when we now sum up 8 values at once instead of 2, and a minor speedup otherwise. For reference, wikibigall is frequently using 2, 3 or 4 bits per value on blocks of doc deltas of stop words ("the", "a", "1", etc.)

Before (without enabling the vector module):

```
Benchmark                                      (bpv)   Mode  Cnt   Score   Error   Units
PostingIndexInputBenchmark.decode                  2  thrpt   15  55,166 ± 0,480  ops/us
PostingIndexInputBenchmark.decode                  3  thrpt   15  51,257 ± 0,086  ops/us
PostingIndexInputBenchmark.decode                  4  thrpt   15  55,769 ± 0,271  ops/us
PostingIndexInputBenchmark.decode                  5  thrpt   15  50,789 ± 0,062  ops/us
PostingIndexInputBenchmark.decode                  6  thrpt   15  49,804 ± 0,187  ops/us
PostingIndexInputBenchmark.decode                  7  thrpt   15  47,552 ± 0,522  ops/us
PostingIndexInputBenchmark.decode                  8  thrpt   15  61,442 ± 0,330  ops/us
PostingIndexInputBenchmark.decode                  9  thrpt   15  40,030 ± 0,084  ops/us
PostingIndexInputBenchmark.decode                 10  thrpt   15  41,480 ± 0,458  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      2  thrpt   15  20,808 ± 0,063  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      3  thrpt   15  20,181 ± 0,184  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      4  thrpt   15  20,931 ± 0,141  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      5  thrpt   15  20,195 ± 0,118  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      6  thrpt   15  20,304 ± 0,185  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      7  thrpt   15  19,334 ± 0,204  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      8  thrpt   15  21,579 ± 0,071  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      9  thrpt   15  17,726 ± 0,253  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum     10  thrpt   15  18,021 ± 0,267  ops/us
```

After (without enabling the vector module):

```
Benchmark                                      (bpv)   Mode  Cnt   Score   Error   Units
PostingIndexInputBenchmark.decode                  2  thrpt   15  57,723 ± 0,501  ops/us
PostingIndexInputBenchmark.decode                  3  thrpt   15  52,526 ± 0,255  ops/us
PostingIndexInputBenchmark.decode                  4  thrpt   15  57,395 ± 0,424  ops/us
PostingIndexInputBenchmark.decode                  5  thrpt   15  50,513 ± 0,076  ops/us
PostingIndexInputBenchmark.decode                  6  thrpt   15  47,176 ± 0,146  ops/us
PostingIndexInputBenchmark.decode                  7  thrpt   15  44,838 ± 0,138  ops/us
PostingIndexInputBenchmark.decode                  8  thrpt   15  61,604 ± 0,262  ops/us
PostingIndexInputBenchmark.decode                  9  thrpt   15  37,737 ± 0,057  ops/us
PostingIndexInputBenchmark.decode                 10  thrpt   15  37,079 ± 0,364  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      2  thrpt   15  30,823 ± 0,052  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      3  thrpt   15  28,148 ± 0,155  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      4  thrpt   15  30,545 ± 0,059  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      5  thrpt   15  22,332 ± 0,100  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      6  thrpt   15  22,240 ± 0,029  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      7  thrpt   15  21,809 ± 0,038  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      8  thrpt   15  23,279 ± 0,376  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      9  thrpt   15  21,624 ± 0,073  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum     10  thrpt   15  21,348 ± 0,033  ops/us
```

After (with enabling the vector module - Apple M3 - preferredBitSize=128):
```
Benchmark                                      (bpv)   Mode  Cnt   Score   Error   Units
PostingIndexInputBenchmark.decode                  2  thrpt   15  51,391 ± 0,171  ops/us
PostingIndexInputBenchmark.decode                  3  thrpt   15  58,821 ± 1,314  ops/us
PostingIndexInputBenchmark.decode                  4  thrpt   15  65,562 ± 0,412  ops/us
PostingIndexInputBenchmark.decode                  5  thrpt   15  58,495 ± 1,771  ops/us
PostingIndexInputBenchmark.decode                  6  thrpt   15  57,617 ± 0,201  ops/us
PostingIndexInputBenchmark.decode                  7  thrpt   15  52,641 ± 0,330  ops/us
PostingIndexInputBenchmark.decode                  8  thrpt   15  61,280 ± 0,488  ops/us
PostingIndexInputBenchmark.decode                  9  thrpt   15  45,320 ± 0,927  ops/us
PostingIndexInputBenchmark.decode                 10  thrpt   15  46,437 ± 0,131  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      2  thrpt   15  28,179 ± 0,408  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      3  thrpt   15  29,989 ± 0,261  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      4  thrpt   15  31,706 ± 0,279  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      5  thrpt   15  21,305 ± 0,031  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      6  thrpt   15  25,399 ± 0,034  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      7  thrpt   15  25,336 ± 0,081  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      8  thrpt   15  27,047 ± 0,262  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum      9  thrpt   15  24,075 ± 0,044  ops/us
PostingIndexInputBenchmark.decodeAndPrefixSum     10  thrpt   15  24,549 ± 0,273  ops/us
```


lucene/core/src/java/org/apache/lucene/codecs/lucene912/ForDeltaUtil.java
```diff
@@ -62,22 +256,374 @@ void encodeDeltas(long[] longs, DataOutput out) throws IOException {
       assert or != 0;
       final int bitsPerValue = PackedInts.bitsRequired(or);
       out.writeByte((byte) bitsPerValue);
-      forUtil.encode(longs, bitsPerValue, out);
+      encodeDeltas(longs, bitsPerValue, out);
+    }
+  }
+
+  private void encodeDeltas(long[] longs, int bitsPerValue, DataOutput out) throws IOException {
```
gsmiller, line 263: I think this is identical to `ForDelta#encode` right? Can we leverage that instead of duplicating it? I guess the issue is that `ForUtil` relies on its `tmp` buffer so the method isn't static, but we could split out a helper static method that takes the buffer as input that could be common I think.

lucene/core/src/java/org/apache/lucene/codecs/lucene912/ForDeltaUtil.java
```diff
@@ -62,22 +256,374 @@ void encodeDeltas(long[] longs, DataOutput out) throws IOException {
...
+      }
+    }
+
+    for (int i = 0; i < numLongsPerShift; ++i) {
+      out.writeLong(tmp[i]);
+    }
+  }
+
+  /** Number of bytes required to encode 128 integers of {@code bitsPerValue} bits per value. */
+  int numBytes(int bitsPerValue) {
```
gsmiller, line 336: I think this is also identical to what's in `ForUtil`? Could we reference it and avoid the duplication?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/ForDeltaUtil.java
```diff
@@ -62,22 +256,374 @@ void encodeDeltas(long[] longs, DataOutput out) throws IOException {
...
+    return bitsPerValue << (BLOCK_SIZE_LOG2 - 3);
+  }
+
+  private static void decodeSlow(
```
gsmiller, line 340: And one more method that I think is duplicated from `ForUtil`?

lucene/core/src/java/org/apache/lucene/internal/vectorization/DefaultPostingDecodingUtil.java
```diff
@@ -29,13 +29,14 @@ public DefaultPostingDecodingUtil(IndexInput in) {
 
   @Override
   public void splitLongs(
```
gsmiller, line 31: What if we add the "slow" implementation as a static method in `PostingDecodingUtil` (as sketched out below) and then delegate to it here and in the "slow path" in the vectorized implementation? That we can avoid the code duplication.

```
  protected static void splitLongsSlow(IndexInput in, int count, long[] b, int bShift, int dec, long bMask, long[] c, int cIndex, long cMask) throws IOException {
    in.readLongs(c, cIndex, count);
    // The below loop is auto-vectorized
    for (int i = 0; i < count; ++i) {
      for (int j = 0, end = (bShift - 1) / dec; j <= end; ++j) {
        b[count * j + i] = (c[cIndex + i] >>> (bShift - j * dec)) & bMask;
      }
      // can we determine cIndex as `count * ((bShift - 1) / dec + 1)`?
      c[cIndex + i] &= cMask;
    }
  }
```

lucene/core/src/java/org/apache/lucene/codecs/lucene912/PostingIndexInput.java
```diff
@@ -50,6 +53,6 @@ public void decode(int bitsPerValue, long[] longs) throws IOException {
    * and store results into {@code longs}.
    */
   public void decodeAndPrefixSum(int bitsPerValue, long base, long[] longs) throws IOException {
```
gsmiller, line 55: It looks like this method is only around for benchmarking? Could benchmarks directly call `ForDeltaUtil#decodeAndPrefixSum` so we could not need this?

gsmiller: Actually, I guess this whole class is now only used by benchmarks? Do we actually need it or could benchmarks go directly against `ForUtil` / `ForDeltaUtil`?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/ForUtil.java
```diff
@@ -712,14 +465,14 @@ private static void decode7(IndexInput in, PostingDecodingUtil pdu, long[] tmp,
     }
   }
 
-  private static void decode8(IndexInput in, PostingDecodingUtil pdu, long[] tmp, long[] longs)
+  static void decode8(IndexInput in, PostingDecodingUtil pdu, long[] tmp, long[] longs)
```
gsmiller, line 468: If I remember correctly, there was a version of the change where you always read the longs through `PostingDecodingUtil` and didn't need `IndexInput` for these edge-cases. It seems a little easier to understand that way (to me at least). Since `PostingDecodingUtil` is always getting created with an `IndexInput`, would it make sense to have a method in `PostingDecodingUtil` that just does the read delegation? Or an alternative would be to expose `IndexInput` (e.g., it could just be a public field in the abstract class)? It's a minor detail but it seems like it would make the code wiring slightly cleaner to not have to provide both `IndexInput` and `PostingDecodingUtil` when initializing `ForUtil` / `ForDeltaUtil`.

gsmiller (commented):
Looks good! I like that you've specialized bpv1-4 as well now. I left a few minor comments/suggestions. Thanks!

lucene/core/src/java/org/apache/lucene/codecs/lucene912/PostingIndexInput.java
jpountz: I forgot to add a comment about that: `PostingsDecodingUtil` instances can only be retrieved from specific call sites, see `VectorizationProvider#ensureCaller`. So I needed to introduce some abstraction that could be used from the benchmark-jmh module. Before this PR, this didn't look too odd because this abstraction was also used in `Lucene912PostingsReader`, but indeed now it's only used by the benchmark-jmh module. I'll add a comment.

jpountz:
Thanks for taking a look @gsmiller ! I believe that I addressed all your (good) comments.

lucene/core/src/java/org/apache/lucene/internal/vectorization/DefaultPostingDecodingUtil.java
gsmiller: > We could do something like that, though it's often more convenient to start at 0 when the c array is tmp rather than longs.

Ha oops. That was a note to myself to think about it some more. Didn't mean for that to make it into my feedback :)

lucene/core/src/java/org/apache/lucene/codecs/lucene912/PostingIndexInput.java
gsmiller: That makes sense, thanks.

gsmiller (approved):
Looks good to me. Thanks!

jpountz:
Sadly, this seems to have caused a minor regression on nightly benchmarks, e.g. [`AndHighMed`](https://people.apache.org/~mikemccand/lucenebench/AndHighMed.html) lost 2% with a p-value of 0.

I cannot reproduce the slowdown locally. However, one thing I am observing is that `splitLongs` is never inlined into the `decodeX` methods, which is annoying as I would expect it to be important performance-wise to unroll various loops. I am not sure if this is due to insufficient warming or if we could organize the code a bit differently to help the JVM make better inlining decisions.

jpountz:
Thanks for looking @gsmiller ! Are the splitLongs calls also not inlined for you as well? I submitted a PR to make nightly reports tell us about which methods get inlined / compiled / interpreted: https://github.com/apache/lucene/pull/13670 as this sounds like useful information for performance analysis.

I'm currently playing with small refactorings to see if I can get this code inlined without bumping `-XX:InlineSmallCode=2500` but I'm not successful so far.

gsmiller:
Good question @jpountz. If I'm looking at this properly, it seems to not be getting inlined in my benchmark run. (I grabbed your patch from #13670 and loaded one of the jfr files produced from luceneutil):

```
PERCENT       CPU SAMPLES   STACK
[...]
0.05%         5             org.apache.lucene.internal.vectorization.MemorySegmentPostingDecodingUtil#splitLongs() [JIT compiled code]
```

jpountz:
I ran wikibigall on a M3, which is interesting because it does inline the splitLongs call, both before and after the change (presumably because the generated native code is smaller thus still candidate for inlining), and the benchmark doesn't get a slowdown (but neither a speedup or a very small one):

```
                       OrHighLow     1387.75      (4.7%)     1358.50      (6.4%)   -2.1% ( -12% -    9%) 0.236
                    AndStopWords       90.68      (5.1%)       88.85      (6.6%)   -2.0% ( -13% -   10%) 0.278
                       And3Terms      438.98      (4.6%)      433.21      (5.7%)   -1.3% ( -11% -    9%) 0.425
               HighTermTitleSort      167.18      (4.6%)      165.66      (4.3%)   -0.9% (  -9% -    8%) 0.519
            HighTermTitleBDVSort       82.36      (4.3%)       81.75      (5.1%)   -0.7% (  -9% -    9%) 0.618
                   OrHighNotHigh      624.72      (4.5%)      620.27      (5.1%)   -0.7% (  -9% -    9%) 0.642
                     AndHighHigh      222.77      (3.9%)      221.34      (4.1%)   -0.6% (  -8% -    7%) 0.611
              Or2Terms2StopWords      424.69      (4.7%)      422.05      (5.3%)   -0.6% ( -10% -    9%) 0.695
                      AndHighMed      426.20      (4.9%)      423.75      (5.0%)   -0.6% ( -10% -    9%) 0.715
                        Or3Terms      424.04      (4.8%)      421.69      (5.0%)   -0.6% (  -9% -    9%) 0.721
                      OrHighRare      486.01      (5.4%)      483.99      (5.8%)   -0.4% ( -10% -   11%) 0.813
                       OrHighMed      513.21      (4.9%)      511.77      (4.8%)   -0.3% (  -9% -    9%) 0.854
                        PKLookup      409.03      (6.5%)      408.23      (7.2%)   -0.2% ( -12% -   14%) 0.928
               HighTermMonthSort     1781.87      (4.0%)     1780.09      (6.2%)   -0.1% (  -9% -   10%) 0.952
                   OrNotHighHigh      570.80      (5.0%)      570.42      (7.1%)   -0.1% ( -11% -   12%) 0.973
           HighTermDayOfYearSort     1509.01      (5.9%)     1512.74      (6.1%)    0.2% ( -11% -   12%) 0.896
                    OrNotHighMed      686.50      (5.2%)      690.48      (7.7%)    0.6% ( -11% -   14%) 0.781
                    OrHighNotMed      784.76      (5.2%)      789.88      (4.8%)    0.7% (  -8% -   11%) 0.683
                 CountAndHighMed      507.69      (4.4%)      511.31      (5.8%)    0.7% (  -9% -   11%) 0.664
                      OrHighHigh      193.31      (4.8%)      195.18      (4.3%)    1.0% (  -7% -   10%) 0.506
                      TermDTSort      526.16      (3.1%)      531.24      (5.7%)    1.0% (  -7% -   10%) 0.508
             And2Terms2StopWords      449.28      (4.5%)      454.04      (5.5%)    1.1% (  -8% -   11%) 0.506
                      AndHighLow     2221.82      (6.3%)     2255.93      (7.5%)    1.5% ( -11% -   16%) 0.484
                CountAndHighHigh      188.97      (4.4%)      191.99      (4.9%)    1.6% (  -7% -   11%) 0.278
                    OrHighNotLow     1001.86      (6.4%)     1021.97      (6.2%)    2.0% (  -9% -   15%) 0.314
                     OrStopWords      100.73      (4.0%)      102.78      (6.4%)    2.0% (  -8% -   12%) 0.229
                        HighTerm      914.24      (5.6%)      933.01      (6.2%)    2.1% (  -9% -   14%) 0.271
                         MedTerm     1122.89      (5.2%)     1146.58      (7.2%)    2.1% (  -9% -   15%) 0.287
                         LowTerm     1617.38      (7.5%)     1661.67      (6.7%)    2.7% ( -10% -   18%) 0.221
                    OrNotHighLow     2220.24      (5.5%)     2294.40      (8.9%)    3.3% ( -10% -   18%) 0.152
                  CountOrHighMed      461.03      (5.7%)      477.71      (6.4%)    3.6% (  -7% -   16%) 0.058
                 CountOrHighHigh      265.96      (6.2%)      279.21      (7.7%)    5.0% (  -8% -   20%) 0.024
```

I'm debating whether I should revert this change or whether we should keep it because the micro benchmark suggests it's a speedup (both on my Ryzen and my M3) and we might find ways to take advantage of this speedup later on. I'm curious if you have an opinion @gsmiller.

gsmiller:
@jpountz I'm in favor of keeping the change as well (your reasoning here makes sense to me). I'll see if I can get another data point by running our internal Amazon product search benchmarks as well to see if anything interesting shows up there on a different workload (not implying that should be a deciding factor by any means, just trying to think of ways to get additional information). If I'm able to find time for that this week I'll report back.

PR #13668, opened 2024-08-19
Convert BKDConfig to a record

iverase:
BKDConfig holds the configuration of a BKD tree and it can benefit of being represented as a record. The only annoying thing is that we cannot hide the canonical constructor but I don't think that's a big issue.

relates https://github.com/apache/lucene/issues/13207

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/bkd/BKDConfig.java
+++ b/lucene/core/src/java/org/apache/lucene/util/bkd/BKDConfig.java
@@ -102,7 +102,7 @@ public record BKDConfig(
     }
     if (packedIndexBytesLength != numIndexDims * bytesPerDim) {
       throw new IllegalArgumentException(
-          "packedBytesLength must be "
+          "packedIndexBytesLength must be "
               + (numIndexDims * bytesPerDim)
               + " (got: "
               + packedIndexBytesLength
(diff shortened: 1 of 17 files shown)


lucene/core/src/java/org/apache/lucene/util/bkd/BKDConfig.java
```diff
@@ -31,48 +48,23 @@ public final class BKDConfig {
...
+        numIndexDims * bytesPerDim,
+        numDims * bytesPerDim + Integer.BYTES);
   }
 
-  private static void verifyParams(
-      final int numDims,
-      final int numIndexDims,
-      final int bytesPerDim,
-      final int maxPointsInLeafNode) {
+  public BKDConfig {
```
uschindler, line 67: moving all the checks here is fine. Can we avoid that `bytesPerDoc` (and others) gets initialized with wrong values by this generic ctor?

uschindler (approved):

lucene/core/src/java/org/apache/lucene/util/bkd/BKDConfig.java
iverase: I added extra validation to make sure we cannot set invalid values. In addition and add a new test suite to check for invalid values.

uschindler: An alternative to this would be to not define the calculated values as field and just declare them as final fields (this is possible in records, too) and populate them in default ctor.

iverase: I don't think we can have instance fields on a record. I add them as getters as those methods are mainly used on indexing where they are less performance sensitive.

uschindler:
Thanks. All fine now. 👍

PR #13670, opened 2024-08-19
Display frame types when analyzing top frames.

jpountz:
This helps understand if a given frame is interpreted, compiled or inlined for instance.


gsmiller (approved):

jpountz:
Nightly benchmarks picked it up, see e.g. https://home.apache.org/~mikemccand/lucenebench/2024.08.20.18.04.44.html#profiler_searching_4_cpu. cc @mikemccand

PR #13672, opened 2024-08-20
Leverage doc value skip lists in DocValuesRewriteMethod if indexed

gsmiller:
### Description

`DocValuesRewriteMethod` (used under-the-hood in `MultiTermQuery`) is another spot where we can leverage the new doc value skip lists when present to potentially skip over doc blocks that don't contain one of the query terms.


lucene/core/src/java/org/apache/lucene/document/DocValuesRangeIterator.java
```diff
@@ -24,8 +24,10 @@
 /**
  * Wrapper around a {@link TwoPhaseIterator} for a doc-values range query that speeds things up by
  * taking advantage of a {@link DocValuesSkipper}.
+ *
+ * @lucene.experimental
  */
-final class DocValuesRangeIterator extends TwoPhaseIterator {
+public final class DocValuesRangeIterator extends TwoPhaseIterator {
```
rmuir, line 30: move to .search package?

lucene/core/src/test/org/apache/lucene/search/TestDocValuesRewriteMethod.java
```diff
@@ -61,14 +61,19 @@ public void setUp() throws Exception {
...
       int numTerms = random().nextInt(4);
       for (int j = 0; j < numTerms; j++) {
         String s = TestUtil.randomUnicodeString(random());
         doc.add(newStringField(fieldName, s, Field.Store.NO));
-        doc.add(new SortedSetDocValuesField(fieldName, new BytesRef(s)));
+        if (createDocValueSkiplist == false) {
+          doc.add(new SortedSetDocValuesField(fieldName, new BytesRef(s)));
+        } else {
+          doc.add(SortedSetDocValuesField.indexedField(fieldName, new BytesRef(s)));
+        }
```
rmuir, line 76: can we swap order so that there's no negated comparison. it is always easier to read

gsmiller:
It just occurred to me we can’t use DocValuesRangeIterator directly here as it could produce false positive matches. I’ll write a test to capture this and iterate on the PR.

gsmiller:
Cleaned up this PR a bit and added testing. Also addressed Robert's feedback (thanks @rmuir). Should be ready for another review if anyone is interested. Thanks!

lucene/core/src/test/org/apache/lucene/search/TestDocValuesRewriteMethod.java
```diff
@@ -61,14 +61,19 @@ public void setUp() throws Exception {
                 .setMaxBufferedDocs(TestUtil.nextInt(random(), 50, 1000)));
     List<String> terms = new ArrayList<>();
     int num = atLeast(200);
+    boolean createDocValueSkiplist = random().nextBoolean();
```
jpountz, line 64: Maybe split the test so that we test both cases explicitly every time?

jpountz (approved):
Nice to see more usage of doc values skippers!

jpountz (approved):

PR #13678, opened 2024-08-22
Support JDK 23 in Panama Vectorization Provider

ChrisHegarty:
This commit updates the Vectorization Provider to support JDK 23. The API has not changed so the changes minimally bump the major JDK check, and enable the incubating API during testing.


uschindler (approved):
As we are in release candidate phase of JDK 23, I see no problem with this.

uschindler:
Backport looks also fine.

PR #13682, opened 2024-08-23
[9.x] Aggregate files from the same segment into a single Arena

ChrisHegarty:
Backport of 
 * #13570

The backport only supports the JDK 21 version of the code - JDK 19 and 20 specific versions do not aggregate files so therefore remain the same. The MemorySession/SegmentScope/Arena APIs are different in 19 and 20. Seems not worth the effort as things would have to be restructured a little in the ref counting implementation.

I modified the changeslog and also included verbiage about the system property. Once we agree final wording I'll consolidate this in the _main_ branch too.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
@@ -16,6 +16,8 @@
  */
 package org.apache.lucene.store;
 
+import static org.apache.lucene.index.IndexFileNames.CODEC_FILE_PATTERN;
+
 import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
@@ -27,7 +29,9 @@ import java.util.Locale;
 import java.util.Optional;
 import java.util.concurrent.Future;
 import java.util.function.BiPredicate;
+import java.util.function.Function;
 import java.util.logging.Logger;
+import org.apache.lucene.index.IndexFileNames;
 import org.apache.lucene.util.Constants;
 import org.apache.lucene.util.SuppressForbidden;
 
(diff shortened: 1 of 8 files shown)


lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -312,6 +361,21 @@ public boolean getPreload() {
     return preload == ALL_FILES;
   }
 
+  /**
+   * Configures a grouping function for files that are part of the same logical group. The gathering
+   * of files into a logical group is a hint that allows for better handling of resources.
```
ChrisHegarty, line 366: Grouping is a **hint**, and therefore may be ignored by an implementation, e.g. JDK 19 and 20 versions of the MemorySegment implementation. We could make this more clear, but I'm on the fence about whether it's worth it or not.

uschindler: It is also ignored by Java 11 mmap. 😜

uschindler (approved):

lucene/core/src/java/org/apache/lucene/store/MappedByteBufferIndexInputProvider.java
```diff
@@ -62,7 +63,13 @@ public MappedByteBufferIndexInputProvider() {
...
 
   @Override
-  public IndexInput openInput(Path path, IOContext context, int chunkSizePower, boolean preload)
+  public IndexInput openInput(
+      Path path,
+      IOContext context,
+      int chunkSizePower,
+      boolean preload,
+      Optional<String> group,
+      Object attachment)
```
uschindler, line 72: Maybe name it unusedAttachment like at other places.

ChrisHegarty: Done.

uschindler:
🙏🥳

PR #13684, opened 2024-08-25
[WIP] Parallel apply termDeletes and queryDeletes by virtual thread in FrozenBufferedUpdates.

vsop-479:


uschindler:
The problem with virtual threads is that you can't use them easily with file access, as all virtual threads have zero permissions.

vsop-479:
Thanks for reminding this @uschindler, I think this is the reason of the check failure about this commit.

PR #13686, opened 2024-08-26
Replace Map<String,Object> with IntObjectHashMap for DV producer

bugmakerrrrrr:
### Description
Today, the map between field name and corresponding meta entry in DocValues Producer is represented by a `HashMap<String, Object>`. To reduce memory usage, we can replace it with an `IntObjectHashMap` that maps field numbers to meta entries.


jpountz (approved):
This works for me, `Lucene90PointsReader` already does this.

PR #13691, opened 2024-08-27
Fixed exponent value in explain of SigmoidFunction

owaiskazi19:
### Description

Coming from https://github.com/opensearch-project/OpenSearch/issues/14921. This PR fixes the correct value of exponent from pivot in the Sigmoid function.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -403,6 +403,8 @@ Bug Fixes
 
 * GITHUB#13627: Fix race condition on flush for DWPT seqNo generation. (Ben Trent, Ao Li)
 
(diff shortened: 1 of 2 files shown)


mikemccand (approved):
Whoa, thank you for bringing the fix from OpenSearch discovery to Lucene @owaiskazi19!

Could you also add a `lucene/CHANGES.txt` entry explaining the bug & fix?  Thanks!

mikemccand (approved):

mikemccand:
Thank you @owaiskazi19 -- I backported to 9.12 as well.

owaiskazi19:
> Thank you @owaiskazi19 -- I backported to 9.12 as well.

Thanks @mikemccand for the help here.

PR #13693, opened 2024-08-27
Use range optimizations for "slow" MultiTermQueries when terms happen to be contiguous

gsmiller:
### Description

This is a small optimization that treats "slow" multi-term queries (e.g., TermInSet, RegexpQuery, etc.) as ordinal ranges when the query terms create a contiguous range. This is likely a rare case, but it doesn't seem too hard to optimize, and I think it's worth doing so especially now that we have a doc value index/skipper.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/document/SortedSetDocValuesRangeQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/document/SortedSetDocValuesRangeQuery.java
@@ -16,19 +16,11 @@
  */
 package org.apache.lucene.document;
 
-import java.io.IOException;
-import java.util.Objects;
-import java.util.function.LongPredicate;
 import org.apache.lucene.index.DocValues;
 import org.apache.lucene.index.DocValuesSkipper;
-import org.apache.lucene.index.LeafReader;
 import org.apache.lucene.index.LeafReaderContext;
-import org.apache.lucene.index.SortedDocValues;
 import org.apache.lucene.index.SortedSetDocValues;
-import org.apache.lucene.search.ConstantScoreScorer;
 import org.apache.lucene.search.ConstantScoreWeight;
-import org.apache.lucene.search.DocIdSetIterator;
-import org.apache.lucene.search.DocValuesRangeIterator;
 import org.apache.lucene.search.FieldExistsQuery;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.Query;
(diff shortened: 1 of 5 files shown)


lucene/core/src/java/org/apache/lucene/document/SortedSetDocValuesRangeQuery.java
```diff
@@ -161,66 +156,8 @@ public Scorer get(long leadCost) throws IOException {
               return new ConstantScoreScorer(score(), scoreMode, DocIdSetIterator.empty());
             }
 
-            // all terms matched in this segment
```
gsmiller, line 164: This looks like a lot of change but it's mostly just a lift-and-shift into the new `SortedSetDocValuesRangeScorer` class

lucene/core/src/java/org/apache/lucene/document/SortedSetDocValuesRangeQuery.java
```diff
@@ -236,69 +173,4 @@ public boolean isCacheable(LeafReaderContext ctx) {
       }
     };
   }
-
-  private DocIdSetIterator getDocIdSetIteratorOrNullForPrimarySort(
```
gsmiller, line 240: Same comment as above (this is all just code that mostly moved as-is into `SortedSetDocValuesRangeScorer`).

gsmiller:
@iverase if you have a chance to take a look at this, I'd appreciate your thoughts (no rush). I'd be particularly interested what you think about moving the scoring logic of `SortedSetDocValuesRangeQuery` out into `SortedSetDocValuesRangeScorer` so it can be shared with `DocValuesRewriteMethod`. I don't love that part since it means making the scoring logic public to share across sub-packages. (Any other feedback also welcome of course!)

iverase:
I have my doubts in this approach and I am unsure we should expose the scorer that way. On the other hand, I wonder if we can rewrite the query to a range query. I understand that in this case, you need to be able to apply the optimisation to all segments, but that's probably the most common case?

gsmiller:
Thanks so much @iverase for the feedback! I wanted to approach this as a query rewrite initially, but I'm not sure it's all the feasible. The issue is that the ordinal mapping is segment-specific (potentially, at lest), and the rewrite would have to inspect all the segment-level doc values to determine if the range is contiguous. That's the issue I ran into at least. 

I also don't love exposing the scoring implementation publicly. It really is an internal implementation detail. Maybe there's a better way to approach this? I'm open to suggestions. I'm also open to dropping this if we think the exposure isn't work the semi-rare likelihood of the optimization. I just don't like leaving optimizations "on the table" when there's an opportunity to improve.

Maybe I'm overlooking an obvious way to take a query rewrite approach? If so, please let me know. Happy to explore other avenues. Thanks again!

PR #13694, opened 2024-08-28
Add a few toString methods for HNSW scoring classes

mikemccand:
I was trying to debug whether the distance metric I had requested using `luceneutil`'s `knnPerfTest.py` was properly working its way all the way down that stack, and found a few `toString` methods to help in my journey.

I also ran `./gradlew tidy` which brought in an unrelated and whitespace-only change to `build-tools/build-infra/src/main/java/org/apache/lucene/gradle/ProfileResults.java`.

### DIFF
--- a/build-tools/build-infra/src/main/java/org/apache/lucene/gradle/ProfileResults.java
+++ b/build-tools/build-infra/src/main/java/org/apache/lucene/gradle/ProfileResults.java
@@ -156,7 +156,12 @@ public class ProfileResults {
 
   /** Process all the JFR files passed in args and print a merged summary. */
   public static void printReport(
-      List<String> files, String mode, int stacksize, int count, boolean lineNumbers, boolean frameTypes)
+      List<String> files,
+      String mode,
+      int stacksize,
+      int count,
+      boolean lineNumbers,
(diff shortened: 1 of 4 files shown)


tteofili (approved):
LGTM

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/ProfileResults.java
```diff
@@ -156,7 +156,12 @@ private static String pad(String input) {
 
   /** Process all the JFR files passed in args and print a merged summary. */
   public static void printReport(
-      List<String> files, String mode, int stacksize, int count, boolean lineNumbers, boolean frameTypes)
```
msokolov, line 159: curious that this change was allowed or required

msokolov (approved):

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/ProfileResults.java
mikemccand: Yeah I'm also not sure why `tidy` needed to do this.

PR #13697, opened 2024-08-28
Add Bulk Scorer For ToParentBlockJoinQuery

Mikep86:


lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinBulkScorer.java
```diff
@@ -0,0 +1,264 @@
...
+  private static final String VALUE_FIELD_NAME = "value";
+  private static final String PARENT_FILTER_VALUE = "parent";
+  private static final String CHILD_FILTER_VALUE = "child";
+
+  private enum MatchValue {
+    MATCH_A("A", 1),
+    MATCH_B("B", 2),
+    MATCH_C("C", 3);
+
+    private static final List<MatchValue> VALUES = List.of(values());
```
Mikep86, line 43: Can we use syntax beyond Java 8 if we want to backport to Lucene 9.12?

lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinBulkScorer.java
```diff
@@ -0,0 +1,264 @@
...
+        childDoc.add(newStringField(TYPE_FIELD_NAME, CHILD_FILTER_VALUE, Field.Store.NO));
+
+        final int matchCount = random().nextInt(maxChildDocMatches + 1);
+        List<MatchValue> matchValues = new ArrayList<>(matchCount);
+        for (int k = 0; k < matchCount; k++) {
+          // Add a match to the child doc
+          MatchValue matchValue = MatchValue.random();
+          matchValues.add(matchValue);
+          childDoc.add(newStringField(VALUE_FIELD_NAME, matchValue.getText(), Field.Store.NO));
+        }
```
Mikep86, line 102: I plan on changing this to include at most, one of each `MatchValue` value since including more than that when using `ConstantScoreQuery` is kind of pointless

lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinBulkScorer.java
```diff
@@ -0,0 +1,264 @@
...
+          ToParentBlockJoinQuery parentQuery =
+              new ToParentBlockJoinQuery(childQueryBuilder.build(), parentsFilter, joinScoreMode);
+
+          Weight weight = searcher.createWeight(searcher.rewrite(parentQuery), searchScoreMode, 1);
+          ScorerSupplier ss = weight.scorerSupplier(searcher.getIndexReader().leaves().get(0));
+
+          // TODO: Why is score supplier null sometimes?
+          if (ss == null) {
+            continue;
+          }
```
Mikep86, line 238: I'm not sure why the `ScorerSupplier` is null sometimes. Thoughts?

lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinBulkScorer.java
Mikep86, line 1: I plan on adding another test to check that the min competitive score is respected when using the bulk scorer

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -440,6 +500,83 @@ private String formatScoreExplanation(int matches, int start, int end, ScoreMode
...
+
+    // TODO: Need to resolve parent doc IDs in multi-reader space?
+    private BatchAwareLeafCollector wrapCollector(LeafCollector collector) {
+      return new BatchAwareLeafCollector(collector) {
+        private Scorable scorer = null;
+
+        @Override
+        public void setScorer(Scorable scorer) throws IOException {
+          this.scorer = scorer;
+          super.setScorer(scorer != null ? currentParentScore : null);
```
Mikep86, line 548: Are there any scenarios where the passed-in scorer can be null? I didn't find any in the tests, but I'd like to check if that assumption holds in all production code paths as well.

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -440,6 +500,83 @@ private String formatScoreExplanation(int matches, int start, int end, ScoreMode
...
+          }
+
+          if (scorer != null && scoreMode != ScoreMode.None) {
+            currentParentScore.addChildScore(scorer.score());
+          }
+        }
+
+        @Override
+        public void endBatch(int doc) throws IOException {
+          if (currentParent != null && doc > currentParent) {
```
jpountz, line 570: I believe that the second condition (doc > currentParent) would always be true? So we don't even need to pass a doc ID to this method, and we could just make sure to collect the current parent if there is one?

jpountz (commented):
Thanks for looking into it, it's exciting that `ToParentBlockJoinQuery` may soon be able to take advantage of our specialized bulk scorers for disjunctions and conjunctions! I left a few comments but this looks on the right track to me.

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -275,6 +289,51 @@ public float matchCost() {
     }
   }
 
+  private static class Score extends Scorable {
+    private final ScoreMode scoreMode;
+    private Float score;
```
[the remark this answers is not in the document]
Mikep86: Yes, assuming a matching doc can never have a score of 0. I assume this is true?

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
Mikep86: I think we can do this once we update `score` to according to your earlier comment. As currently implemented, if `doc == currentParent`, we cannot emit the parent because it equals `max`, causing assertions to fail.

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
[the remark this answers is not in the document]
jpountz: 0 is a legal score (but negative values are not legal scores), but I'm missing where we would need to assume that a matching doc cannot have a score of zero?

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -440,6 +500,83 @@ private String formatScoreExplanation(int matches, int start, int end, ScoreMode
...
+      this.parents = parents;
+      this.currentParentScore = new Score(scoreMode);
+      this.currentParent = null;
+    }
+
+    @Override
+    public int score(LeafCollector collector, Bits acceptDocs, int min, int max)
+        throws IOException {
+      BatchAwareLeafCollector wrappedCollector = wrapCollector(collector);
+      childBulkScorer.score(wrappedCollector, acceptDocs, min, max);
```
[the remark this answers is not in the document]
Mikep86: Updated to meet this spec, LMKWYT!

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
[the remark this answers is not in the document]
Mikep86: I'm trying to determine what the uninitialized value should be. This matters for when the score mode is `Min`, I think we need to know when the score is uninitialized so we can properly handle the first doc score.

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -440,6 +498,98 @@ private String formatScoreExplanation(int matches, int start, int end, ScoreMode
...
+      // Subtract one because max is exclusive w.r.t. score but inclusive w.r.t prevSetBit
+      int lastParent = parents.prevSetBit(Math.min(parentsLength - 1, max - 1));
+      if (lastParent < min) {
+        // No parent docs in this range. Save the passed-in min so we can score over this range once
+        // we find a parent doc.
+        currentMin = min;
+        return max;
+      } else if (currentMin == -1) {
+        currentMin = min;
+      }
```
jpountz, line 536: I would rather not track `currentMin` and find the previous parent, e.g. something like that:

```java
int prevParent = min == 0 ? -1 : parents.prevSetBit(min - 1);
```

And then score childBulkScorer using `min = prevParent + 1` and `max = lastParent` (`lastParent + 1` is technically correct too, it shouldn't matter since `lastParent + 1` must not match the child query anyway).

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -440,6 +498,101 @@ private String formatScoreExplanation(int matches, int start, int end, ScoreMode
...
+      // Subtract one because max is exclusive w.r.t. score but inclusive w.r.t prevSetBit
+      int lastParent = parents.prevSetBit(Math.min(parentsLength, max) - 1);
+      int prevParent = min == 0 ? -1 : parents.prevSetBit(min - 1);
+      if (lastParent == prevParent) {
+        // No parent docs in this range
+        return max;
+      }
+
+      BatchAwareLeafCollector wrappedCollector = wrapCollector(collector);
+      childBulkScorer.score(wrappedCollector, acceptDocs, prevParent + 1, lastParent + 1);
```
Mikep86, line 534: I set `max` to `lastParent + 1` so that we catch if the query matches `lastParent`

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -275,6 +286,52 @@ public float matchCost() {
...
+          throw new AssertionError();
+      }
+
+      freq++;
+    }
+
+    @Override
+    public float score() {
+      float score = this.score;
+      if (scoreMode == ScoreMode.Avg && freq > 0) {
```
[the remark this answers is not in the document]
Mikep86: Good call, i'll convert that to an assert

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -440,6 +477,99 @@ private String formatScoreExplanation(int matches, int start, int end, ScoreMode
...
+
+    @Override
+    public int score(LeafCollector collector, Bits acceptDocs, int min, int max)
+        throws IOException {
+      // Subtract one because max is exclusive w.r.t. score but inclusive w.r.t prevSetBit
+      int lastParent = parents.prevSetBit(Math.min(parentsLength, max) - 1);
+      int prevParent = min == 0 ? -1 : parents.prevSetBit(min - 1);
+      if (lastParent == prevParent) {
+        // No parent docs in this range
+        return max;
```
jpountz, line 509: Should it return `NO_MORE_DOCS` if `max >= parentsLength` like you're doing in the other return statement?

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -440,6 +477,99 @@ private String formatScoreExplanation(int matches, int start, int end, ScoreMode
...
+
+    @Override
+    public long cost() {
+      return childBulkScorer.cost();
+    }
+
+    // TODO: Need to resolve parent doc IDs in multi-reader space?
+    private BatchAwareLeafCollector wrapCollector(LeafCollector collector) {
+      return new BatchAwareLeafCollector(collector) {
+        private final Score currentParentScore = new Score(scoreMode);
```
jpountz, line 529: You need to extend it to override `setMinCompetitiveScore` and delegate to `scorer` when appropriate, if you want dynamic pruning to work.

jpountz: By the way, can you add a test that dynamic pruning works with your new bulk scorer? See e.g. `TestMaxScoreBulkScorer#testBasicsWithTwoDisjunctionClauses` for an example of a similar test.

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -275,6 +286,53 @@ public float matchCost() {
...
+      this.score = 0;
+      this.freq = 0;
+    }
+
+    public void reset(float firstChildScore) {
+      score = firstChildScore;
+      freq = 1;
+    }
+
+    public void addChildScore(float childScore) {
```
jpountz, line 305: I wonder if it should take a `Scorable` instead of a `float` so that we could encapsulate the logic of not computing the score if `scoreMode == NONE` in this class instead of forcing callers to care about it.

jpountz (commented):
I left some more comments, it's close!

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
Mikep86: Will do!

jpountz (commented):
I believe it's only missing a setMinCompetitiveScore test now, otherwise LGTM. Can you mark this PR as ready for review and add a CHANGES entry under 9.12?

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -440,6 +478,114 @@ private String formatScoreExplanation(int matches, int start, int end, ScoreMode
...
+                  return currentParentScore.score();
+                }
+
+                @Override
+                public void setMinCompetitiveScore(float minScore) throws IOException {
+                  if (scoreMode == ScoreMode.None || scoreMode == ScoreMode.Max) {
+                    scorer.setMinCompetitiveScore(minScore);
+                  }
+                }
+              });
```
Mikep86, line 553: LMKWYT of this approach. I originally added the child scorer to `currentParentScore`'s constructor and created `currentParentScore` in `setScorer`, but that didn't work because `setScorer` is potentially called multiple times.

I then tried creating `currentParentScore` once, on the first call to `setScorer`, and checking that the same child scorer is passed in subsequent calls in an assertion. At least in the tests, this didn't work because the child scorer is wrapped in a different `AssertingScorable` in each call.

This approach seemed like the simplest that avoids the above problems. I could go back to them though if we can assume that subsequent calls to `setScorer` will pass the same child scorer without checking in an assertion.

jpountz: It looks good to me this way.

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -440,6 +478,101 @@ private String formatScoreExplanation(int matches, int start, int end, ScoreMode
...
+      this.scoreMode = scoreMode;
+      this.parents = parents;
+      this.parentsLength = parents.length();
+    }
+
+    @Override
+    public int score(LeafCollector collector, Bits acceptDocs, int min, int max)
+        throws IOException {
+      // Subtract one because max is exclusive w.r.t. score but inclusive w.r.t prevSetBit
+      int lastParent = parents.prevSetBit(Math.min(parentsLength, max) - 1);
```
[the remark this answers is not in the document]
Mikep86: What if we checked for the general `min == max` case? In that case, would it be best to do nothing and return  `max`?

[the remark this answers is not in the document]
jpountz: That would work too. Like with your other return statements, we may need to return `max >= maxDoc ? NO_MORE_DOCS : max`.

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -156,6 +162,11 @@ public Scorer get(long leadCost) throws IOException {
           return new BlockJoinScorer(childScorerSupplier.get(leadCost), parents, scoreMode);
         }
 
+        @Override
+        public BulkScorer bulkScorer() throws IOException {
+          return new BlockJoinBulkScorer(childScorerSupplier.bulkScorer(), scoreMode, parents);
```
gsmiller, line 167: This is super nit-picky, but could you make the order of this ctor's arguments consistent with `BlockJoinScorer` for slightly easier reading (i.e., flip `scoreMode` and `parents`)?

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -352,28 +412,10 @@ private void setScoreAndFreq() throws IOException {
       if (childApproximation.docID() >= parentApproximation.docID()) {
         return;
       }
-      double score = scoreMode == ScoreMode.None ? 0 : childScorer.score();
-      int freq = 1;
+      parentScore.reset(childScorer);
```
gsmiller, line 415: I might be overlooking something here, but wouldn't it be more efficient to check `scoreMode` here and avoid this entire update-loop that iterates through all matching children if score mode is NONE?

gsmiller (commented):
Looks good! Left a couple small comments/questions. Thanks!

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -408,16 +407,25 @@ public void setMinCompetitiveScore(float minScore) throws IOException {
...
+        return parentScore.score();
       }
-      parentScore.reset(childScorer);
-      while (childApproximation.nextDoc() < parentApproximation.docID()) {
-        if (childTwoPhase == null || childTwoPhase.matches()) {
-          parentScore.addChildScore(childScorer);
+
+      float score = 0;
+      if (scoreMode == ScoreMode.None) {
+        childApproximation.advance(parentApproximation.docID());
```
gsmiller, line 417: Do we actually need to advance here? Won't this be handled by the consuming code of the iterator anyway? Probably not a big deal either way, but could save an unnecessary advance I think?

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
gsmiller: Thanks! Agreed that the default just works through `nextDoc` but in practice the implementation is likely overriding it with something a little more performant I expect.

gsmiller (commented):
Thanks!

lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinBulkScorer.java
```diff
@@ -0,0 +1,450 @@
...
+        {
+          Map<Integer, Float> expectedScores =
+              Map.of(
+                  2, 6.0f,
+                  10, 10.0f);
+
+          ScorerSupplier ss = weight.scorerSupplier(searcher.getIndexReader().leaves().get(0));
+          ss.setTopLevelScoringClause();
+          assertScores(ss.bulkScorer(), scoreMode, 6.0f, expectedScores);
+        }
```
Mikep86, line 377: @jpountz This part of the test fails because `MaxScoreBulkScorer` determines that the maximum possible score for doc 13 (B + C) is 6. I traced it as far back as here: https://github.com/apache/lucene/blob/8d4f7a6e99d2da802b7019247b0f8f305d71c024/lucene/core/src/java/org/apache/lucene/search/MaxScoreBulkScorer.java#L300

Thoughts as to what could be going on?

jpountz: I suspect that this is due to the fact that `MaxScoreBulkScorer` hasn't scored A yet, and doesn't know if A is going to match or not. So it needs to compute the max score as if A was matching, ie. score=B+C+A=6. Later it advances A, and notices that it doesn't match. But since it has already computed the full score of the boolean query, it still calls the collector, there is not significant work that can be saved by not calling the collector.

We probably need to relax the test to allow the scorer to visit some docs that have a score that is less than the min competitive score. E.g. could we assert that it scores at least the docs that are in your expected map, and also that it does not exhaustively evaluated all hits (ie. the collector is called a number of times that is less than the number of matching docs)?

jpountz (approved):

lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinBulkScorer.java
```diff
@@ -0,0 +1,454 @@
...
+
+        {
+          // Doc 16 is returned because MaxScoreBulkScorer scores assuming A will match in doc 13,
+          // leading to a potential max score of 6. By the time it determines that A doesn't match,
+          // scoring is complete and thus there is no advantage to not collecting the doc.
+          Map<Integer, Float> expectedScores =
+              Map.of(
+                  2, 6.0f,
+                  10, 10.0f,
+                  16, 5.0f);
```
Mikep86, line 376: @jpountz The simplest way I see to relax this test is to add doc 16 as an expected match. I think this is OK because we are working with a static index and query. Also, we know that `collect` in `assertScores` is not called more times than we expect because each call will result in a new entry in `actualScores`. WDYT?

jpountz: I'm good with that.

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -156,6 +162,14 @@ public Scorer get(long leadCost) throws IOException {
           return new BlockJoinScorer(childScorerSupplier.get(leadCost), parents, scoreMode);
         }
 
+        @Override
+        public BulkScorer bulkScorer() throws IOException {
+          if (scoreMode == ScoreMode.None) {
+            return super.bulkScorer();
```
jpountz, line 168: Can you add a comment? E.g. something like "BlockJoinBulkScorer evaluates all child hits exhaustively, however when the score mode is NONE we only need to evaluate a single child document per parent document".

lucene/CHANGES.txt
```diff
@@ -376,6 +376,8 @@ Optimizations
 
 * GITHUB#13587: Use Max WAND optimizations with ToParentBlockJoinQuery when using ScoreMode.Max (Mike Pellegrini)
 
+* GITHUB#13697: Add Bulk Scorer For ToParentBlockJoinQuery (Mike Pellegrini)
```
jpountz, line 379: Add a bit more color, e.g. "Add a bulk scorer to ToParentBlockJoinQuery, which delegates to the bulk scorer of the child query. This should speed up query evaluation when the child query has a specialized bulk scorer, such as disjunctive queries."

jpountz (approved):

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -348,34 +410,23 @@ public void setMinCompetitiveScore(float minScore) throws IOException {
...
-            case Max:
-              score = Math.max(score, childScore);
-              break;
-            case None:
-              break;
-            default:
-              throw new AssertionError();
+
+      float score = 0;
+      if (scoreMode != ScoreMode.None) {
```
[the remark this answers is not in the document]
gsmiller: I'll have to look at the test case a little closer, but I'd rather we didn't unnecessarily advance to make a test case happy. If there's a much sparser query clause leading a conjunctive iteration, the eager advancing of the child iterator to the next parent could add meaningful overhead. I think anyway?

I'm also not totally sure why we even have that check in the TPBJQ logic. It seems like a slightly odd place to do that check? Maybe there's a different place we could do that check if necessary? Maybe we could do a check in `ParentApproximation#advance` to validate that the child iterator doesn't provide a doc that's present in the parent bitset? Another alternative might be an `assert` that eagerly advances to the next parent (like you had)? With an `assert`, we could bypass the potential performance drain while still getting some validation coverage?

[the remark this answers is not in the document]
gsmiller: Do we have a clear understanding of why it's important to do this error-check in the first place? It's unclear to me at the moment, and I'm a little worried we're bending the production code to a specific test case without understanding whether-or-not this is important (and why). My best guess here is that the check is trying to ensure we don't incorrectly allow parent documents to contribute to the store accumulation if the docs are indexed incorrectly?

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -242,14 +242,25 @@ public int nextDoc() throws IOException {
...
+          target == 0 ? 0 : parentBits.prevSetBit(Math.min(target, parentBits.length()) - 1);
+
       int childDoc = childApproximation.docID();
-      if (childDoc < firstChildTarget) {
-        childDoc = childApproximation.advance(firstChildTarget);
+      if (childDoc < prevParent) {
+        childDoc = childApproximation.advance(prevParent);
+      }
+
+      // TODO: Can we assume that doc 0 is never a real parent that also matches the child query?
```
Mikep86, line 253: Because we use 0 to indicate "no previous parent doc", we cannot detect the case when:

- A real parent doc exists at 0
- That parent doc matches the child query

I don't think this is a problem since any real parent doc at 0 would never be returned by `ToParentBlockJoinQuery` since it cannot have child docs. Thoughts?

lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinScorer.java
```diff
@@ -81,7 +81,7 @@ public void testScoreNone() throws IOException {
         new QueryBitSetProducer(new TermQuery(new Term("docType", "parent")));
     CheckJoinIndex.check(reader, parentsFilter);
 
-    Query childQuery = new MatchAllDocsQuery();
+    Query childQuery = new TermQuery(new Term("docType", "child"));
```
Mikep86, line 84: Needed to modify this test since the child doc  == parent doc check is now performed by the parent doc iterator

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -431,17 +442,6 @@ private float scoreChildDocs() throws IOException {
...
-          && (childTwoPhase == null || childTwoPhase.matches())) {
-        throw new IllegalStateException(
-            "Child query must not match same docs with parent filter. "
-                + "Combine them as must clauses (+) to find a problem doc. "
-                + "docId="
-                + parentApproximation.docID()
-                + ", "
-                + childScorer.getClass());
-      }
-
```
Mikep86: I don't think we need this check in `scoreChildDocs` anymore since it is now done in the parent doc iterator, but let me know if I'm wrong about that

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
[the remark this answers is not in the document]
Mikep86: Based on my understanding, this error check is important to ensure that parent docs do not match child doc queries. If such a thing were to happen in production, I think (at a minimum) there would be negative scoring implications. Perhaps @jpountz could provide some more background on why these checks were added.

Based on the tests that are written to exercise this error check, I do not believe we are bending production code to fit a test case. From what I see, it's the opposite: the tests are written to simulate a scenario that could happen in production.

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -224,18 +242,30 @@ public int nextDoc() throws IOException {
...
+                + ", "
+                + this.getClass());
       }
-      if (childDoc >= parentBits.length() - 1) {
+
+      if (childDoc >= parentBits.length()) {
         return doc = NO_MORE_DOCS;
       }
-      return doc = parentBits.nextSetBit(childDoc + 1);
+      return doc = parentBits.nextSetBit(childDoc);
```
Mikep86, line 268: I updated this to handle the case when there is one parent doc at 0 that also matches the child query

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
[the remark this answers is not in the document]
Mikep86: I did a deep dive into how to avoid advancing `childApproximation` and the approach as of https://github.com/apache/lucene/pull/13697/commits/5a993e3b401a76aeca0c92780f945c4cffc13073 is _so close_ to working, except it doesn't when the child approximation isn't exact and a two-phase iterator is required. I will continue investigating how to apply my solution with a two-phase iterator, but given the deadline with FF I think we should prepare for the outcome where we roll back to advancing `childApproximation` in `scoreChildDocs`.

I think this is a really interesting optimization and if we run out of time to implement it for Lucene 9.12 & 10, we can do it for a future release.

Thoughts?

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
jpountz: The child query is not supposed to match any parent doc, so my understanding is that it's for the case when the child query produces a two-phase iterator and  onlythe approximation matches the parent doc? Can you leave a comment since it's probably not obvious?

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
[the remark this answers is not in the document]
jpountz: I believe that we have these checks because Lucene could otherwise fail with confusing errors if the child query matched parent documents, which would suggest a bug in Lucene rather than a problem in the application code. 

It's ok to remove some of these checks if they cause issues (these checks are best effort anyway), but I'd err on the side of keeping them when they're cheap and e.g. don't impact the set of docs that we're evaluating.

Keeping the removal of this call to advance() works for me, it's somewhat orthogonal to this PR. I agree that removing this wasteful call to advance() would be nice.

jpountz (approved):

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
Mikep86: I'm going to revert this change in this PR anyways, but I will add a comment about why it's necessary in a follow-up PR.

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -348,34 +414,24 @@ public void setMinCompetitiveScore(float minScore) throws IOException {
...
+        while (childApproximation.nextDoc() < parentApproximation.docID()) {
+          if (childTwoPhase == null || childTwoPhase.matches()) {
+            parentScore.addChildScore(childScorer);
           }
         }
+
+        score = parentScore.score();
       }
+
+      // TODO: Fix this error check when score mode is None
```
jpountz, line 434: can you explain a bit more why it's broken, someone coming to this code without context wouldn't understand immediately I believe

jpountz (approved):

jpountz (approved):

lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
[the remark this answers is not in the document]
gsmiller: Thanks @Mikep86 & @jpountz!

jpountz:
FYI @Mikep86 opened a PR at https://github.com/apache/lucene/pull/13785.

PR #13701, opened 2024-08-29
Remove IOException from MultiTermQuery#getTermsCount

cbuescher:
Neither this method nor any of the two overrides can throw an IOException. This change removes the throws clauses from this method in order simplify not have to handle them on the callers side.


ChrisHegarty (approved):
I don't see any reason why the terms count should throw an IOException. LGTM.

PR #13708, opened 2024-09-02
move Operations.sameLanguage/subsetOf to AutomatonTestUtil in test-framework

rmuir:
These methods run in quadratic time and have been traps in the past: they run in quadratic time.

I think originally this was `equals()` but it is so costly, that we factored out into separate `sameLanguage` methods.

Additionally methods are a bit smelly, they will throw exception if the inputs aren't deterministic, and they'll only `assert` if there are transitions to dead states... they are really only suitable for tests code.

We've been making progress moving some of this code to test framework, this is just the next iteration. Would love to factor out more (dead states checks etc) but that's more difficult.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/StatePair.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/StatePair.java
@@ -35,9 +35,14 @@ package org.apache.lucene.util.automaton;
  * @lucene.experimental
  */
 public class StatePair {
+  // only mike knows what it does (do not expose)
   int s;
-  int s1;
-  int s2;
+
+  /** first state */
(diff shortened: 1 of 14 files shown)


mikemccand (approved):
Thanks @rmuir.

lucene/core/src/java/org/apache/lucene/util/automaton/StatePair.java
```diff
@@ -35,9 +35,14 @@
  * @lucene.experimental
  */
 public class StatePair {
+  // only mike knows what it does (do not expose)
```
dweiss, line 38: lol.

dweiss (approved):

PR #13711, opened 2024-09-03
Move anonymous Weight implementation in PointRangeQuery to named class

jainankitk:
### Description

Moves the anonymous `Weight` implementation in `PointRangeQuery#createWeight` to named class for better extensibility and resusability.


jpountz:
I see from the linked issue that you would like to extend `PointRangeQuery`, but in general we don't like to think of our queries as being extensible. I wonder if you could do what you need through composition rather than extension?

jainankitk:
> I see from the linked issue that you would like to extend `PointRangeQuery`, but in general we don't like to think of our queries as being extensible. I wonder if you could do what you need through composition rather than extension?

@jpountz - Thanks for your feedback. While I agree with using composition rather than extension for `PointRangeQuery` itself, `Weight` implementation itself within `PointRangeQuery` will be more reusable (via composition instead of extension) by being a non-anonymous class.

jpountz:
Sorry, I don't think we should make Lucene's `Weight` implementations public.

I looked up the OpenSearch issue, if I understand correctly, the problem you're trying to solve is that it's wasteful for `PointRangeQuery` to evaluate the whole range when it's only asked for the first 10 doc IDs that match the range query. I agree it's wasteful. We have the same problem on nightly benchmarks and the IntNRQ task. I wonder if there are better ways to do what you're after, e.g. adding a `TopDocs Weight#topk(int n, int totalHitsThreshold)` API that would default to collecting hits, and that some classes such as `PointRangeQuery` could override. As I'm writing this, I'm not convinced that it's actually a good idea. Using sparse indexing would likely be a better approach, especially if the index can be sorted, as this would produce good iterators that don't have this huge up-front cost of evaluating the query against the entire segment.

PR #13712, opened 2024-09-04
Dry up TestScorerPerf

javanna:
Some of the test methods were commented out when this test class was added. They got later removed but the removal left unused method behind. I also adjusted visibility of all the internal methods that were public and should have been private, which led me to further clean up: `MatchingHitCollector` was not needed and can be removed.


gsmiller (approved):
Looks good to me. Thanks for tightening this up!

javanna:
Thanks  a lot for the review @gsmiller !

PR #13714, opened 2024-09-04
Make Operations#repeat create simpler automata.

jpountz:
`Operations#repeat` currently creates an automaton that has one more final state, and 2x more transitions for each of the final states. With this change, the returned automaton has a single final state and only 2x more transitions for state 0.

Currently a draft as I'm still looking into whether this new logic is correct.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/Operations.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/Operations.java
@@ -183,40 +183,73 @@ public final class Operations {
       return a;
     }
 
-    if (a.isAccept(0) && a.getAcceptStates().nextSetBit(1) == -1) {
-      // If the only accept state is 0, then this automaton already repeats itself. Automata
-      // returned by this function only accept state 0, so this makes this function idempotent.
+    if (a.isAccept(0) && a.getAcceptStates().cardinality() == 1) {
+      // If state 0 is the only accept state, then this automaton already repeats itself.
       return a;
     }
 
-    Automaton.Builder builder = new Automaton.Builder();
-    // Create the initial state, this is the only accepted state.
-    builder.createState();
-    builder.setAccept(0, true);
-
-    int[] stateMap = new int[a.getNumStates()];
-    for (int state = 0; state < a.getNumStates(); ++state) {
-      if (a.isAccept(state)) {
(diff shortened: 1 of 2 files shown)


jpountz:
A practical side-effect of this change is that `.*` now gets parsed to the below automaton:

automaton

while our main branch currently parses it to the below automaton:

automaton2

jpountz:
I just pushed a commit that avoids creating dead states if the input automaton doesn't have dead states itself. So the automaton created by `.*` would now have a single state.

For context, I started looking into optimizing the single-transition case only (hence the branch name) but as I started looking deeper into it, it looked like improving the general case wouldn't be much harder, as the algorithm boils down to renumbering final states to 0 and other states to `state + 1`.

mikemccand (approved):
I left a couple small comments.  I think this change (optimization) is correct, but it is sure hard to think about!

Do we have good randomized testing of this method?  Maybe we could add a random test that makes a random regexp, build the A from that (minimized or determinized or not or so), calls repeat, then goes back to the original regexp and compiles `(<original-regexp>)*` and asserts `sameLanguage`?

rmuir:
> Do we have good randomized testing of this method? Maybe we could add a random test that makes a random regexp, build the A from that (minimized or determinized or not or so), calls repeat, then goes back to the original regexp and compiles `(<original-regexp>)*` and asserts `sameLanguage`?

When optimizing a scary method, a previous approach we have taken is to take existing implementation unmodified, move it to `AutomatonTestUtil.foobarSimple()`, and then generate random automata and assert that optimized method and unoptimized method accept the same language.

jpountz:
Good ideas, I'll look into it.

rmuir:
Was thinking something similar to this test but for `repeat`: https://github.com/apache/lucene/blob/main/lucene/core/src/test/org/apache/lucene/util/automaton/TestMinimize.java#L39-L45

jpountz:
Thanks for the test idea, it found interesting test failures in the case when accept states had transitions, as these transitions would then become transitions on the initial state, which is not expected.

mikemccand (approved):
Thanks @jpountz!

rmuir (approved):

rmuir:
thanks for adding comments to the new repeat code.

PR #13702, opened 2024-08-30
Remove CollectorOwner class (#13671)

epotyom:
~~Deprecate~~ Remove CollectorOwner class (#13671)

Also remove `drillDownLeafCollector` attr in `DrillSidewaysQuery` and `DrillSidewaysScorer` as its job can be done by main collector

PR #13716, opened 2024-09-04
Backport of #13702

gsmiller:


lucene/facet/src/java/org/apache/lucene/facet/DrillSideways.java
```diff
@@ -217,16 +215,18 @@ public DrillSidewaysResult search(DrillDownQuery query, Collector hitCollector)
 
     Map<String, Integer> drillDownDims = query.getDims();
 
+    FacetsCollector drillDownCollector = createDrillDownFacetsCollector();
+    Collector mainCollector;
+    if (drillDownCollector != null) {
+      mainCollector = MultiCollector.wrap(hitCollector, drillDownCollector);
+    } else {
+      mainCollector = hitCollector;
+    }
```
epotyom, line 224: If I understand correctly, we used to use `createDrillDownFacetsCollector` only when there were no drill-sideways dimensions, and when there were dimensions, we used to get drill-down collectors using `createDrillDownFacetsCollectorManager`. Although this seems like the correct behavior to me, we are slightly changing the way it works. However, this change would only be noticeable to someone who overrides one of these methods to return something different from `FacetsCollector(Manager)`.

To keep current behavior, we can move this block to under `if (drillDownDims.isEmpty())` ; and for the code with dimensions do something like:

```
FacetsCollectorManager drillDownCollectorManager = createDrillDownFacetsCollectorManager();
Collector mainCollector;
 if (drillDownCollectorManager != null) {
      mainCollector = MultiCollector.wrap(hitCollector, drillDownCollectorManager.newCollector());
} else {
      mainCollector = hitCollector;
}
```

What do you think?

epotyom (commented):
Thank you for making the change @gsmiller ! I have one suggestion, otherwise the change looks good to me.

lucene/facet/src/java/org/apache/lucene/facet/DrillSideways.java
gsmiller: Shoot, yeah good catch @epotyom. I think the pre-existing behavior was an unintentional oversight, but I agree that we should retain it, especially since this is likely the last release with this method and it's going away in 10.0. I opened a small PR with your suggestion: #13722

PR #13721, opened 2024-09-05
Early exit from Operations#removeDeadStates when an automaton doesn't have dead states.

jpountz:


rmuir (approved):
nice, thank you!

ChrisHegarty (commented):
Nice!

PR #13722, opened 2024-09-05
Follow-up to GH#13702

gsmiller:
Ensures we retain pre-existing (but strange) inconsistency in DrillSideways#search(DrillDownQuery, Collector). This is a deprecated method so I propose we retain this inconsistency since the method will go away with 10.0.


epotyom (approved):
Looks good to me, thanks for making the change!

PR #13726, opened 2024-09-05
Remove leftover search(Query, Collector) usages in TestTaxonomyFacetAssociations

javanna:


gsmiller (approved):
Thanks for dragging this across the finish line and chipping away at this deprecation task!

javanna:
It seems like I missed something, maybe that's why these methods were not migrated yet :) I got a failure with seed `8DDC50A89D6B6EC9`. Digging.

gsmiller:
@javanna saw you addressed this by disabling concurrency in the test. I wonder if we should instead make the test more resilient to ordering and leverage the work in #13723 as a follow-up. WDYT?

PR #13728, opened 2024-09-05
jgit/ clean status check should ignore any 'untracked folders' 

### LINKED ISSUE #13719: simplify checkWorkingCopyClean to make backporting easier?
### Description

I run into trouble when backporting, the `checkWorkingCopyClean` check always wants to fail on things that are actually in gitignore, from switching branches. IMO, working copy is clean as "git status" has no output.

Should we implement this differently? at my dayjob, I use the following one-liner to implement this check, which will respect .gitignore:
```
check-working-copy:
  @test -z "`git status --porcelain`" || (echo "working copy is dirty" && git status --porcelain && exit 2)
```

dweiss:
Fixes #13719.

We can't easily determine whether an "untracked folder" contains something that is .gitignored and currently we report the status as dirty (contrary to what git --porcelain says). This patch just removes all the "untracked folders" logic.


rmuir (approved):
yes, I think this folder logic is what I keep hitting, thank you

dweiss:
I'll merge this tomorrow and backport to 9x.

uschindler:
Thanks. I wasn't aware of this extra File walk. Who added this? Thanks for removing it.

dweiss:
> Thanks. I wasn't aware of this extra File walk. Who added this? Thanks for removing it.

Can't remember. Could have been me. Clearly not working in all cases though.

PR #13732, opened 2024-09-06
Add migration note about the deprecation of optional RegExp complement syntax

ChrisHegarty:
This commit adds a migration note about the deprecation of optional RegExp complement support.

### DIFF
--- a/lucene/MIGRATE.md
+++ b/lucene/MIGRATE.md
@@ -83,13 +83,14 @@ These classes no longer take a `determinizeWorkLimit` and no longer determinize
 behind the scenes. It is the responsibility of the caller to call
 `Operations.determinize()` for DFA execution.
 
-### RegExp no longer supports the optional complement syntax
+### RegExp optional complement syntax has been deprecated
 
-The `RegExp` class no longer supports the optional complement syntax, previously
-supported by the `~` character. When working with automata retrieved from regexp,
-equivalent functionality can be achieved by using `Operations.complement`. For
-example, `new RegExp("~(foo)").toAutomaton()` that matches input that is not
-"foo", can be rewritten as `Operations.complement(new RegExp("(foo)").toAutomaton(), ...)`.
+Support for the optional complement syntax (`~`) has been deprecated.
+The `COMPLEMENT` syntax flag has been removed and replaced by the
+`DEPRECATED_COMPLEMENT` flag. Users wanting to enable the deprecated
+complement support can do so by explicitly passing a syntax flags that
+has `DEPRECATED_COMPLEMENT` when creating a `RegExp`. For example:
+`new RegExp("~(foo)", RegExp.DEPRECATED_COMPLEMENT)`.
 
 Alternatively, and quite commonly, a more simple _complement bracket expression_,
 `[^...]`, may be a suitable replacement, For example, `[^fo]` matches any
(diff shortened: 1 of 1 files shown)


ChrisHegarty:
Relates to Elasticsearch issue https://github.com/elastic/elasticsearch/issues/112453

ChrisHegarty:
The removal of the optional complement syntax is technically a "breaking" change. It is of course fine to do such things in a major release, no issue there. Just want to ensure that it is intentional, and we provide whatever guidance is necessary in the migration, etc.

mikemccand:
Thanks @ChrisHegarty for catching this!  It's important to document all breaks in `MIGRATE.md`.

I'm trying to understand why/when we removed this operator.  Was it really during [adding NFA support](https://issues.apache.org/jira/browse/LUCENE-10010) (linked from your linked Elasticsearch issue)?  I don't think that was intentional, or (more likely) I at least don't remember it -- I thought that change was purely additive.

mikemccand:
I think it was [this commit](https://github.com/apache/lucene/commit/84e4b85b094ccf2dc248ad90a302307d3618b81e) that removed it ...?

rmuir:
I removed it. I don't wish to revisit the situation. But it was not enabled by default for a reason.

rmuir (approved):
maybe if you want, mention that complement-of-range/label (e.g. `[^a-z]`) is still supported. This is the common use case and does not require exponential time.

rmuir:
maybe this small post with diagram helps explain the situation: I hope it is more clear that there's no way we could support "let caller decide NFA vs DFA" with this operator: https://eugene-eeo.github.io/blog/nfa-complement.html

rmuir:
and the title of this video can't be beat: https://www.youtube.com/watch?v=bbMku8ZAoBk

mikemccand:
Phew, thank you for the historical context @rmuir!!  I forgot about the horrors of complementing NFAs... I love that talk title.

mikemccand (approved):
Thanks @ChrisHegarty!

ChrisHegarty:
> Sorry, i was a bit offended that it might have been accidental :)

Apologies. My wording was a little clumsy. No offence intended.

> maybe if you want, mention that complement-of-range/label (e.g. [^a-z]) is still supported. This is the common use case and does not require exponential time.

Good idea. I'll add something about this.

rmuir:
> Apologies. My wording was a little clumsy. No offence intended.

no need, nobody's fault. I read too far into @mikemccand comment without enough caffeine

rmuir:
also just FYI

Operations.complement() still exists if you need it. And it documents the horrors that it does and requires you to pass work limit since determinization is necessary for correctness.

Hence instead of:
```java
var automaton = new Regexp("~(crazystuff)").toAutomaton();
```

you can do:
```java
var automaton = Operations.complement(new Regexp("(crazystuff").toAutomaton(), MY_DETERMINIZE_WORK_LIMIT);
```

The key to this change is that regexp parsing doesn't implicitly minimize() or determinize() whenever it feels like it. Instead, it is the user's choice (caller responsibility).

ChrisHegarty:
Thank @rmuir. I included that advice in the migration note.

rmuir (approved):

rmuir:
This is the best alternative I could come up with. For the user it just kicks the can down the road, but that's probably years out until the next release so it can make the change easier to cope with? https://github.com/apache/lucene/pull/13739

Meanwhile we can just remove it from `main` as soon as 10.x is branched and proceed normally with improving NFA support.

rmuir (approved):

PR #13734, opened 2024-09-06
remove snake case parameter names in RegexpQuery

rmuir:
Maybe there is a way to prevent this stuff going forwards, but for now just fixing the snake case stuff here since this is java and in the public API


slow-J (approved):

ChrisHegarty (approved):
LGTM

cpoerschke (approved):

PR #13737, opened 2024-09-06
Deprecate FacetsCollector#search utility methods

javanna:
These can now be replaced by the corresponding methods added to FacetsCollectorManager that accept a FacetsCollectorManager as last argument in place of a Collector

This is the backport #13733 , minus the removal of the static methods which get instead deprecated in the 9x branch.

Relates to #11041

### DIFF
--- a/lucene/facet/src/java/org/apache/lucene/facet/FacetsCollectorManager.java
+++ b/lucene/facet/src/java/org/apache/lucene/facet/FacetsCollectorManager.java
@@ -19,7 +19,20 @@ package org.apache.lucene.facet;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.List;
+import org.apache.lucene.search.Collector;
 import org.apache.lucene.search.CollectorManager;
+import org.apache.lucene.search.FieldDoc;
+import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.MultiCollectorManager;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.ScoreDoc;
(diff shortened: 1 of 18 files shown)


lucene/facet/src/java/org/apache/lucene/facet/FacetsCollectorManager.java
```diff
@@ -54,4 +79,167 @@ public ReducedFacetsCollector(final Collection<FacetsCollector> facetsCollectors
           facetsCollector -> matchingDocs.addAll(facetsCollector.getMatchingDocs()));
     }
   }
+
+  /** Utility method, to search and also collect all hits into the provided {@link Collector}. */
```
gsmiller, line 83: Shoot, I think I missed this in the earlier review but looks like we need a small javadoc cleanup here (need to mention `FacetsCollectorManager` instead of `Collector` right?). Same comment applies to all the new methods. Sorry for overlooking that.

I just updated these on `main` with a quick commit (dc47adbbe736ef55f56feb8d66dc768eaba05fff). Could you cherry-pick this down into the back-port as well (and feel free to tweak the language if you think of ways to improve on it). Thanks!

gsmiller (approved):
Looks right to me. I caught one javadoc slip-up from the commit onto main. Left a comment around that. Thanks!

lucene/facet/src/java/org/apache/lucene/facet/FacetsCollectorManager.java
javanna: good catch, I missed that. Will backport your commit!

PR #13739, opened 2024-09-07
Add deprecated complement (~) operator to RegExp

rmuir:
Previously all regexp parsing required determinization and minimization up-front, which can be costly (exponential time).

Lucene 10 removes the determinization and minimization from RegExp and allows the user to choose:

* determinize() the result and get the DFA query execution of previous releases.
* don't determinize() and possibly get a new NFA query that determinizes-as-it-goes.

Complement of arbitrary automata is incompatible with this choice, as it requires determinization for correctness. It was previously a non-default operator that could be enabled with a special flag: `RegExp.COMPLEMENT`, or would be included with `RegExp.ALL`, which turns on all special syntax flags. Lucene 10 removed the operator, as it can't be supported while still giving the user the NFA/DFA choice, and requires exponential time during parsing.

To ease transition: add `RegExp.DEPRECATED_COMPLEMENT` syntax flag and `Kind.DEPRECATED_COMPLEMENT` node:
* syntax flag can be enabled with `RegExp(s, RegExp.DEPRECATED_COMPLEMENT);`
* syntax flag is **NOT** included by `RegExp.ALL`: e.g. you must do `RegExp(s, RegExp.ALL | RegExp.DEPRECATED_COMPLEMENT)` to get ALL flags and also the deprecated complement (~) operator. This enforces a java deprecation reference in the calling code to enable the flag.
* deprecated complement (~) runs with an internal limit: `Operations.DEFAULT_DETERMINIZE_WORK_LIMIT`. It is not configurable. If it is exceeded, you get `TooComplexToDeterminize` exception.
* there is intentionally only a single dead-simple test so that this hack doesn't cause us pain with CI/builds. We don't want random automata testing to only rarely encounter an exponential algorithm!

After lucene 10 is branched, this deprecated support can be removed by reverting this commit.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
@@ -368,7 +368,7 @@ public class RegExp {
     REGEXP_REPEAT_MIN,
     /** An expression that repeats a minimum and maximum number of times */
     REGEXP_REPEAT_MINMAX,
-    /** The complement of an expression */
+    /** The complement of a character class */
     REGEXP_COMPLEMENT,
     /** A Character */
     REGEXP_CHAR,
@@ -387,7 +387,14 @@ public class RegExp {
     /** An Interval expression */
     REGEXP_INTERVAL,
     /** An expression for a pre-defined class e.g. \w */
-    REGEXP_PRE_CLASS
+    REGEXP_PRE_CLASS,
+    /**
+     * The complement of an expression.
+     *
+     * @deprecated Will be removed in Lucene 11
+     */
+    @Deprecated
+    REGEXP_DEPRECATED_COMPLEMENT
   }
 
   // -----  Syntax flags ( <= 0xff )  ------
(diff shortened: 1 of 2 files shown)


ChrisHegarty (approved):
I appreciate this effort, and it would be generally useful to help migration to Lucene 10. The code actually looks quite straightforward.  

That said, we can achieve similar, albeit with a bit more effort on the user side. So no need to merge this "just for me"! I can parse the regex and build the automaton with pieces.

rmuir:
I think I overestimated the obscurity of this operator. In my opinion, with this strategy it won't annoy lucene at all and give users a lot more time to deal with migration. Just seems like an easy win. Maybe other users/projects are affected too, but just not as proactive about discovering this before the release.

ChrisHegarty:
> … with this strategy it won't annoy lucene at all and give users a lot more time to deal with migration. Just seems like an easy win.

++ agreed. If you want to add migration advise in this PR then I’ll close the other #13732. Otherwise, I’ll update that PR to reflect what you’ve done here.

Thanks @rmuir

rmuir:
Lets update your migration PR with upgrade notes. Not many people want to mess with automaton regex parser, so I see some value in keeping the commit revertable? In general, trying to minimize surface area here to dodge any future issues with 10.x back porting too.

dweiss (approved):

mikemccand:
Thanks @rmuir!

Issue #12892, opened 2023-12-07
Remove all deprecated IndexSearcher#search(Query, Collector) usage / methods in the next major release


zacharymorn:
### Description

As a follow-up of https://github.com/apache/lucene/issues/11041, we would like to remove all deprecated `IndexSearcher#search(Query, Collector)` methods in the next major release, preferably `10.0.0`.

javanna:
I have been looking into this, there are unfortunately ~85 leftover usages of this method. Would be great to clean this up for Lucene 10. I opened a few small PRs around this.

I think though that `FacetsCollector` needs extensive rework to leverage collector managers instead of collectors. I think it would be much simpler if the static search methods that `FacetsCollector` exposes took a `FacetsCollector` as last argument instead of a generic `Collector`. That seems to cover all of our internal usages, but I am not sure what the expectations are around users usage for such methods. We should discuss how we want to revisit the facets API as a whole.

javanna:
I edited the description of this issue to include an extensive list of the leftover usages that need fixing.

gsmiller:
Opened #13735 for the `monitor` usage

msfroh:
Opened https://github.com/apache/lucene/pull/13747 for `join`/`join.test`.

PR #13748, opened 2024-09-10
Update QueryUtils to use CollectorManager

harshavamsi:
### Description

Relates to https://github.com/apache/lucene/issues/12892

We introduce a new `SimpleCollectorWithLastReader` class to help keep context of LastReader per thread. This way calling `CollectorManager#reduce` can work with an executor. 

The diff makes it look big, but in reality we just wrap SimpleCollector inside a CollectorManager.

### DIFF
--- a/lucene/test-framework/src/java/org/apache/lucene/tests/search/QueryUtils.java
+++ b/lucene/test-framework/src/java/org/apache/lucene/tests/search/QueryUtils.java
@@ -25,6 +25,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Random;
+import java.util.function.Supplier;
 import org.apache.lucene.index.BinaryDocValues;
 import org.apache.lucene.index.ByteVectorValues;
 import org.apache.lucene.index.DocValuesSkipper;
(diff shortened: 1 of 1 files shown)


lucene/test-framework/src/java/org/apache/lucene/tests/search/QueryUtils.java
```diff
@@ -345,175 +348,199 @@ public static void checkSkipTo(final Query q, final IndexSearcher s) throws IOEx
...
-                  for (int d = iterator.advance(lastDoc[0] + 1);
-                      d != DocIdSetIterator.NO_MORE_DOCS;
-                      d = iterator.nextDoc()) {
-                    if (liveDocs == null || liveDocs.get(d)) {
-                      more = true;
-                      break;
+                    @Override
+                    public LeafReader getLastReader() {
+                      return null;
                     }
```
msfroh, line 516: You'll need to return `lastReader`.

lucene/test-framework/src/java/org/apache/lucene/tests/search/QueryUtils.java
```diff
@@ -540,117 +567,148 @@ public static void checkFirstSkipTo(final Query q, final IndexSearcher s) throws
...
 
-            lastReader[0] = context.reader();
-            lastDoc[0] = -1;
-          }
-        });
+              @Override
+              public List<LeafReader> reduce(Collection<SimpleCollectorWithLastReader> collectors) {
+                List<LeafReader> lastReaders = new ArrayList<>();
+                for (SimpleCollectorWithLastReader collector : collectors) {
+                  lastReaders.add(collector.getLastReader());
```
msfroh, line 695: You may want to check if the value returned from `getLastReader` is `null` before adding it to the array (and in the other reduce method).

msfroh:
I had another thought while walking home. Both of the anonymous CollectorManagers are the same (but produce slightly different collectors). What if you split it out into an inner class that takes a `Supplier<SimpleCollectorWithLeafReader>` as a constructor arg?

That way the two checks can each have their anonymous implementations, but the anonymous nesting won't be as deep.

lucene/test-framework/src/java/org/apache/lucene/tests/search/QueryUtils.java
```diff
@@ -345,175 +348,201 @@ public static void checkSkipTo(final Query q, final IndexSearcher s) throws IOEx
...
-                              + "  "
-                              + q.getClass().getName()
-                              + "\n\t Searcher="
-                              + s
-                              + "\n\t Order="
-                              + sbord
-                              + "\n\t Op="
-                              + (op == skip_op ? " skip()" : " next()"));
+                      lastReader = context.reader();
+                      assert readerContextArray.get(leafPtr).reader() == context.reader();
```
msfroh, line 508: This assert is failing. 

I assume it's because each slice gets a subset of segments, but the old logic assumed that the LeafReaders would be passed to the collector in the same order that they're in the top-level reader.

In a concurrent search world, I don't think this assertion adds value. As long as all the segments get collected, I don't think the order really matters.

javanna: I did not have time to review this in depth, but there is more to it than just the assertion. These collectors are rather hard to follow at first glance. Also, this is a case where we cannot resort to a single-threaded collector manager unfortunately, because the searcher is externally provided and that would require disabling concurrency in all those tests that call `QueryUtils` methods somehow. Calling search(Query, Collector) bypassed concurrency so far despite an executor is provided to the searcher, which is why we want to remove it.

PR #13731, opened 2024-09-06
PayloadScoreQuery javadoc update w.r.t. SpanQuery use

cpoerschke:

PR #13749, opened 2024-09-10
add RawTFSimilarity class

cpoerschke:
Motivation is to use the TF like a payload but without needing to have payloads and positions.

see also
* #13731 `PayloadScoreQuery` javadoc update w.r.t. `SpanQuery` use
* https://github.com/apache/lucene/blob/releases/lucene/9.11.1/lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/DelimitedTermFrequencyTokenFilter.java for indexing


cpoerschke:
> ... use the TF like a payload ...

Initially I thought that this would require a custom `Term[Query|Scorer]` style change but then from a brief chat with @seanmacavaney (thank you!) I learnt that maybe `TFSimilarity` could be a thing here instead and it turns out `TestConjunctions` already had it as an inner class also.

lucene/core/src/java/org/apache/lucene/search/similarities/TFSimilarity.java
```diff
@@ -0,0 +1,41 @@
...
+    return 1; // we dont care
+  }
+
+  @Override
+  public SimScorer scorer(
+      float boost, CollectionStatistics collectionStats, TermStatistics... termStats) {
+    return new SimScorer() {
+      @Override
+      public float score(float freq, long norm) {
+        return freq;
```
seanmacavaney, line 37: Should this be `boost * freq`?

cpoerschke: Good catch, thanks! Will also add test coverage to match, marking as draft in the meantime.
```suggestion
        return boost * freq;
```

lucene/core/src/java/org/apache/lucene/search/similarities/TFSimilarity.java
```diff
@@ -0,0 +1,41 @@
...
+import org.apache.lucene.search.CollectionStatistics;
+import org.apache.lucene.search.TermStatistics;
+
+/** Similarity that returns the TF as score */
+public class TFSimilarity extends Similarity {
+
+  @Override
+  public long computeNorm(FieldInvertState state) {
+    return 1; // we dont care
+  }
```
rmuir, line 29: I'd recommend against this. It would encode the norm differently, making it impossible for users to switch similarities without reindexing.

I would encode it the same as BM25Similarity, TFIDFSimilarity, SimilarityBase, etc.

just ignore the value in scorer() as you are doing already.

if the user wants to not encode norms then they can omit them on the field, it is a separate concern.

lucene/core/src/java/org/apache/lucene/search/similarities/TFSimilarity.java
```diff
@@ -0,0 +1,41 @@
...
+
+  @Override
+  public SimScorer scorer(
+      float boost, CollectionStatistics collectionStats, TermStatistics... termStats) {
+    return new SimScorer() {
+      @Override
+      public float score(float freq, long norm) {
+        return boost * freq;
```
rmuir, line 37: I think i'd prefer a different name such as `RawTFSimilarity`.

Otherwise, user thinks this is just "tf" component of "tf/idf" or something. But it is missing even some of the logic for tf component, e.g. there's no saturation.

jpountz:
Your reference to `DelimitedTermFrequencyTokenFilter` suggests that the freq here is more a feature than an actual frequency of a term in a doc. From an API perspective, this would make me want to expose it via an IndexableField sub class, with a query factory, a bit like `FeatureQuery` but for integer values?

lucene/core/src/java/org/apache/lucene/search/similarities/TFSimilarity.java
rmuir: I'd also support removing this footgun as a separate PR. Similarity is supposed to be an "expert" interface, but this abstract method that impacts the index format is trappy, for the reasons shown here.

Maybe it should have a default implementation (currently copied about 3 or 4 different places in subclasses)? https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/similarities/Similarity.java#L114

I wouldn't mess with any other method, except their javadocs. If we have a default implementation for `computeNorm` then we can suggest in these places a code snippet of how to decode it (e.g. link to SmallFloat.byte4ToInt or whatever) to help users:
* https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/similarities/Similarity.java#L158
* https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/similarities/Similarity.java#L167

It would make the class easier to use. Users might look at BM25Similarity to try to figure out how to do it today, but that one has heavy optimizations such as precomputed length table. A couple javadocs would go a long way on the query-side there.

cpoerschke: Started #13757 to factor out a `Similarity.doComputeNorm` (or similar) method.

lucene/core/src/java/org/apache/lucene/search/similarities/RawTFSimilarity.java
```diff
@@ -0,0 +1,76 @@
...
+  public long computeNorm(FieldInvertState state) {
+    final int numTerms;
+    if (state.getIndexOptions() == IndexOptions.DOCS && state.getIndexCreatedVersionMajor() >= 8) {
+      numTerms = state.getUniqueTermCount();
+    } else if (discountOverlaps) {
+      numTerms = state.getLength() - state.getNumOverlap();
+    } else {
+      numTerms = state.getLength();
+    }
+    return SmallFloat.intToByte4(numTerms);
```
cpoerschke, line 63: Marking as draft here until the #13757 is available and used here.

lucene/core/src/java/org/apache/lucene/search/similarities/TFSimilarity.java
cpoerschke: Renamed in https://github.com/apache/lucene/pull/13749/commits/861034b2316d1df12e1f75899524af10e036d41f commit.

lucene/core/src/java/org/apache/lucene/search/similarities/RawTFSimilarity.java
```diff
@@ -0,0 +1,76 @@
...
+   * <ul>
+   *   <li>{@code discountOverlaps = true}
+   * </ul>
+   */
+  public RawTFSimilarity() {
+    this(true);
+  }
+
+  /**
+   * @param discountOverlaps True if overlap tokens (tokens with a position of increment of zero)
```
cpoerschke, line 41:
```suggestion
  /**
   * RawTFSimilarity
   *
   * @param discountOverlaps True if overlap tokens (tokens with a position of increment of zero)
```

seanmacavaney (approved):
Looks great, thanks @cpoerschke!

cpoerschke:
> Your reference to `DelimitedTermFrequencyTokenFilter` suggests that the freq here is more a feature than an actual frequency of a term in a doc. From an API perspective, this would make me want to expose it via an IndexableField sub class, with a query factory, a bit like `FeatureQuery` but for integer values?

(belatedly) thanks for this mention! yes, the value is a feature of a term in a doc, and actually originally non-integer. (still work in progress) https://github.com/apache/solr/pull/3318 documentation now includes both a `RawTFSimilarity` and a `FeatureQuery` section.

PR #13750, opened 2024-09-10
DrillSideways#search: use bounded wildcard for the list of drill-sideways CollectorManagers

epotyom:
`DrillSideways#search`: use bounded wildcard for the list of drill-sideways CollectorManagers

Lists are invariant, so current API doesn't allow passing something like `List<CollectorManagerImplementation>` where `class CollectorManagerImplementation implements CollectorManager<A, B>`. Invariant makes it so `List<CollectorManagerImplementation>` does not extend `List<CollectorManager<A, B>>`. Using bounded wildcard type allows to overcome that.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -293,7 +293,7 @@ API Changes
 * GITHUB#13568: Add DoubleValuesSource#toSortableLongDoubleValuesSource and
   MultiDoubleValuesSource#toSortableMultiLongValuesSource methods. (Shradha Shankar)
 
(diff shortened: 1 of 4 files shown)


stefanvodita (approved):
Looks good! We should be able to get this in 9.12 since it's only touching code from #13702.

stefanvodita:
@epotyom - could you still add a CHANGES entry to keep track of this change?

epotyom:
Thank you fore reviewing @stefanvodita 

> @epotyom - could you still add a CHANGES entry to keep track of this change?

I've added the PR ID to the existing changelog record as it was not released yet and to avoid spamming CHANGES.txt with similar items. Do you think it's ok?

PR #13751, opened 2024-09-10
Provide a custom hash implementation in HnswLock

ChrisHegarty:
This commit is a micro optimisation to the HNSW Lock implementation, which avoids integer and vararg boxing when determining the hash of the node and level. Trivially we provide our own arity specialised hash, rather than the more generic `java.util.Objects.hash(Object...)`.

I observed the boxing in allocation profiles that can be see in the nightly benchmarks
https://benchmarks.mikemccandless.com/index.html

[Indexing fast ~1 KB docs with vectors (2024.09.08.18.04.35)](https://blunders.io/jfr-demo/indexing-1kb-vectors-2024.09.08.18.04.35/jvm_info)

https://blunders.io/jfr-demo/indexing-1kb-vectors-2024.09.08.18.04.35/allocations-drill-down

Screenshot 2024-09-10 at 11 22 21


lucene/core/src/java/org/apache/lucene/util/hnsw/HnswLock.java
```diff
@@ -67,4 +66,10 @@ public void close() {
       lock.unlock();
     }
   }
+
+  static int hash(int v1, int v2) {
+    int result = 31 + v1;
+    result = 31 * result + v2;
```
dweiss, line 72: Not sure whether it has to be identical to Object.hash... When you look at it unrolled it just shifts by a constant, which doesn't seem to be gaining much when the next thing you do is a modulo.
```
31 * 31 + v1 * 31 + v2
```
If we care about such optimizations, maybe just replace the hashing function completely? BitMixer.mixPhi(node * 31 + level), for example...

dweiss: (or just drop the constant shifting part and leave v1 * 31 + v2...).

ChrisHegarty: I initially just wanted a like for like replacement that avoided the potential allocations. But yeah, I agree, this doesn't have to be the same as Objects:hash.

dweiss (approved):

shubhamvishu (approved):

msokolov:
thank you @ChrisHegarty !

PR #13759, opened 2024-09-11
Rename FieldInfo(s)#hasVectors to hasTermVectors.

jpountz:
This will help remove confusion between dense numeric vectors and term vectors.


mikemccand (approved):
THANK YOU!

How awesome that Lucene has been alive/active for so long that the industry terminology shifts out from under it over the years ...

PR #13763, opened 2024-09-11
Replace Map<String,Object> with IntObjectHashMap for KnnVectorsReader

bugmakerrrrrr:
### Description
Following up on #13686.


lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldKnnVectorsFormat.java
```diff
@@ -239,51 +245,69 @@ public FieldsReader(final SegmentReadState readState) throws IOException {
...
     }
 
     @Override
     public FloatVectorValues getFloatVectorValues(String field) throws IOException {
-      KnnVectorsReader knnVectorsReader = fields.get(field);
-      if (knnVectorsReader == null) {
+      final FieldInfo info = fieldInfos.fieldInfo(field);
+      KnnVectorsReader reader;
+      if (info == null || (reader = fields.get(info.number)) == null) {
         return null;
```
bugmakerrrrrr, line 267: According to the Javadoc of `KnnVectorsReader#getFloatVectorValues`:
> Returns the FloatVectorValues for the given field. The behavior is undefined if the given field doesn't have KNN vectors enabled on its FieldInfo. The return value is never null.

Maybe we should return an empty instance if the target field doesn't exists in the segment or we should correct the Javadoc.

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene90/Lucene90HnswVectorsReader.java
```diff
@@ -223,7 +224,8 @@ public void checkIntegrity() throws IOException {
 
   @Override
   public FloatVectorValues getFloatVectorValues(String field) throws IOException {
-    FieldEntry fieldEntry = fields.get(field);
+    final FieldInfo info = fieldInfos.fieldInfo(field);
```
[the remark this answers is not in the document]
bugmakerrrrrr: Yes, this is also what I want to discuss. You can ref to my comment below, if we all agree to check for `null`, I can fix this.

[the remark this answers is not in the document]
jpountz: In my opinion, the most important thing is to remove the leniency we have in some places, like `if (fieldEntry == null) return EMPTY_VECTOR_VALUES`, which could hide bugs. Checking for null explicitly is not required, though it's obviously nice.

lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldKnnVectorsFormat.java
jpountz: The javadocs are correct. AssertingKnnVectorsReader checks for this.

bugmakerrrrrr: forgive me if i misunderstand something, but the javadoc says that **the return value is never null**, and we do return null in `PerFieldKnnVectorsFormat`. If we check the field info in caller, `AssertingKnnVectorsReader` will always get the non-null value from the delegate.

bugmakerrrrrr:
@benwtrent @jpountz I have merged main branch into this one, can I get a review on this?

benwtrent (approved):

lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldKnnVectorsFormat.java
jpountz: It's probably not worded as well as it could, it wants to say that the return value is never null when the field has vector enabled on its field infos.

jpountz: These checks on info==null and reader == null are not hurting, but they are not needed either since KnnVectorsFormat#getFloatVectorValues should not be called when the field doesn't have vectors enabled on its infos.

ChrisHegarty (changes requested):
Awaiting result of discussion #13805

bugmakerrrrrr:
hi @jpountz ,since we have moved to lucene 10, should we merge this and add back #13686?

jpountz:
Would you like to make this PR up-to-date and open a new one for the other change (a cherry-pick isn't clean due to other changes)?

bugmakerrrrrr:
> Would you like to make this PR up-to-date and open a new one for the other change (a cherry-pick isn't clean due to other changes)?

@jpountz sure, I‘d be happy to do this.

bugmakerrrrrr:
@jpountz I have merged main branch, PTAL:)

jpountz (approved):

PR #13764, opened 2024-09-11
Back-port Add Bulk Scorer For ToParentBlockJoinQuery (#13697) to 9.x

Mikep86:
Back-port changes from #13697 to 9.x

### DIFF
--- a/lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
+++ b/lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
@@ -16,6 +16,7 @@
  */
 package org.apache.lucene.search.join;
 
+import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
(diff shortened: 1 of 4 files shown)


lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
```diff
@@ -180,6 +186,26 @@ public void setTopLevelScoringClause() throws IOException {
...
+        // BlockJoinScorer is optimized to skip child hit evaluation when scoreMode is None.
+        return super.bulkScorer(context);
+      }
+
+      ScorerSupplier scorerSupplier = scorerSupplier(context);
+      if (scorerSupplier == null) {
+        // No docs match
+        return null;
+      }
+      scorerSupplier.setTopLevelScoringClause();
```
Mikep86, line 204: The default `bulkScorer` implementation checks that `scorerSupplier(context)` is non-null and calls `scorerSupplier.setTopLevelScoringClause()`, so I think we need to do that as well here too.

jpountz (approved):

PR #13767, opened 2024-09-11
Remove 8 bit quantization for HNSW/KNN vector indexing

mikemccand:
It is buggy today (see #13519).  4 and 7 bit quantization still work.

It's a bit tricky because 9.11 indices may have 8 bit compressed vectors which are buggy at search time (and users may not realize it, or may not be using them at search time).  But the index is still intact since we keep the original full float precision vectors.  Merges will write new segments (at 4 or 8 bit quantization, or no quantization) and be correct.  So, users can force rewrite all their 9.11 written segments (or reindex those docs), and can change to 4 or 7 bit quantization for newly indexed documents.  The 9.11 index is still usable.

(I added a couple test cases confirming that one can indeed change their mind, indexing a given vector field first with 4 bit quantization, then later (new IndexWriter / Codec) with 7 bit or with no quantization.)

I added MIGRATE.md explanation.

Separately, I also tightned up the `compress` boolean to throw an exception unless bits=4.  Previously (for 7 bit compression) it silently ignored `compress=true` for 7, 8 bit quantization.  And tried to improve its javadocs a bit.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -211,6 +211,19 @@ Bug Fixes
 * GITHUB#12878: Fix the declared Exceptions of Expression#evaluate() to match those
   of DoubleValues#doubleValue(). (Uwe Schindler)
 
+* GITHUB#13519): 8 bit scalar vector quantization is no longer
+  supported: it was buggy starting in 9.11 (GITHUB#13197).  4 and 7
+  bit quantization are still supported.  Existing (9.x) Lucene indices
+  that previously used 8 bit quantization can still be read/searched
+  but the results from `KNN*VectorQuery` are silently buggy.  Further
+  8 bit quantized vector indexing into such (9.11) indices is not
+  permitted, so your path forward if you wish to continue using the
+  same 9.11 index is to index additional vectors into the same field
+  with either 4 or 7 bit quantization (or no quantization), and ensure
+  all older (9.11 written) segments are rewritten either via
+  `IndexWriter.forceMerge` or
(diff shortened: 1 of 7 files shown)


mikemccand:
I also wonder if we still need to give the `compress` choice to users?  I know originaly @benwtrent was concerned about the added cost of decoding the packed/compressed form (two dimensions in one byte), but I think I also saw a change to SIMD-optimize that case?  Maybe the performance penalty is not so bad and we could always do `compress=true` when `bits=4`?  I'll open a spinoff issue, and try to benchmark this using luceneutil's `knnPerfTest.py`.

mikemccand:
Also, we should backport this for 9.12?  I think it's OK to make a hard break there for 8 bit quantization, since it's not working today at search time in 9.11 anyways?

benwtrent:
> I also wonder if we still need to give the compress choice to users?

Yeah, this is indeed something to reconsider. Before I wrote the SIMD focused operations it was compressed 4 bits was pretty rough (slower than 7 bits). But now its faster than 7 bits and sometimes even faster than uncompressed 4 bits (I do not fully understand why that is the case myself...)

mikemccand:
I plan to resolve the conflict (just `MIGRATE.md`) and merge this in the next day or so!

mikemccand:
> Also, we should backport this for 9.12? I think it's OK to make a hard break there for 8 bit quantization, since it's not working today at search time in 9.11 anyways?

Actually, since these quantized `KNNVectorFormat`s are already marked `@lucene.experimental`, I plan on backport to also hard-remove the 8 bit case for 9.12 as well.

PR #13772, opened 2024-09-12
Migrate more classes to records.

jpountz:
The most interesting one in this list is probably TermAndVector since it's public.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/TermAndVector.java
+++ b/lucene/core/src/java/org/apache/lucene/util/TermAndVector.java
@@ -24,37 +24,24 @@ import java.util.Locale;
  *
  * @lucene.experimental
  */
-public class TermAndVector {
-
-  private final BytesRef term;
-  private final float[] vector;
-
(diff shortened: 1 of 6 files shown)


lucene/core/src/java/org/apache/lucene/util/TermAndVector.java
```diff
@@ -24,37 +24,24 @@
...
     float vectorLength = 0;
     for (int i = 0; i < vector.length; i++) {
       vectorLength += vector[i] * vector[i];
     }
     vectorLength = (float) Math.sqrt(vectorLength);
+    float[] newVector = new float[vector.length];
     for (int i = 0; i < vector.length; i++) {
-      vector[i] /= vectorLength;
+      newVector[i] = vector[i] / vectorLength;
     }
```
shubhamvishu, line 43: Could we instead just use the existing `VectorUtil#l2normalize` here?

jpountz: Good call

shubhamvishu (approved):

PR #13773, opened 2024-09-12
Fix TestPrefixRandom failures when intra-segment concurrency is enabled

javanna:
The order in which documents are processed is not a guarantee, hence we may return a different set of documents when early terminating the search. Those are not incorrect results though.

I opted for increasing the number of hits so that we never early terminate the search for this test.

I believe this test should have failed with inter-segment concurrency as well, and intra-segment concurrency makes it more likely to fail (although not so frequently).

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/TestPrefixRandom.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestPrefixRandom.java
@@ -43,6 +43,7 @@ public class TestPrefixRandom extends LuceneTestCase {
   private IndexSearcher searcher;
   private IndexReader reader;
   private Directory dir;
+  private int numDocs;
 
   @Override
   public void setUp() throws Exception {
@@ -59,8 +60,8 @@ public class TestPrefixRandom extends LuceneTestCase {
     Field field = newStringField("field", "", Field.Store.NO);
     doc.add(field);
 
-    int num = atLeast(1000);
-    for (int i = 0; i < num; i++) {
+    numDocs = atLeast(1000);
+    for (int i = 0; i < numDocs; i++) {
       field.setStringValue(TestUtil.randomUnicodeString(random(), 10));
       writer.addDocument(doc);
     }
@@ -134,8 +135,8 @@ public class TestPrefixRandom extends LuceneTestCase {
     PrefixQuery smart = new PrefixQuery(new Term("field", prefix));
     DumbPrefixQuery dumb = new DumbPrefixQuery(new Term("field", prefix));
 
-    TopDocs smartDocs = searcher.search(smart, 25);
-    TopDocs dumbDocs = searcher.search(dumb, 25);
+    TopDocs smartDocs = searcher.search(smart, numDocs);
+    TopDocs dumbDocs = searcher.search(dumb, numDocs);
     CheckHits.checkEqual(smart, smartDocs.scoreDocs, dumbDocs.scoreDocs);
   }
 }


jpountz:
> Those are not incorrect results though.

I think that these are incorrect results. Inter-segment concurrency has a mechanism so that if the top-N-th hit has score S and doc ID D, then it would require a score of at least Math.nextUp(S) for doc IDs > D and S for doc IDs < D (relying on tie-breaking by doc ID). It looks like this mechanism no longer works with intra-segment concurrency?

javanna:
Happy that you commented, thanks!

Can you help me better understand why these are incorrect results?

```
Error Message:
java.lang.AssertionError: Hit 0 docnumbers don't match
Hits length1=25 length2=25
hit=0: doc0=1.0 shardIndex=-1,   doc155=1.0 shardIndex=-1
hit=1: doc1=1.0 shardIndex=-1,   doc156=1.0 shardIndex=-1
hit=2: doc2=1.0 shardIndex=-1,   doc157=1.0 shardIndex=-1
hit=3: doc3=1.0 shardIndex=-1,   doc158=1.0 shardIndex=-1
hit=4: doc4=1.0 shardIndex=-1,   doc159=1.0 shardIndex=-1
hit=5: doc5=1.0 shardIndex=-1,   doc160=1.0 shardIndex=-1
hit=6: doc6=1.0 shardIndex=-1,   doc161=1.0 shardIndex=-1
hit=7: doc7=1.0 shardIndex=-1,   doc162=1.0 shardIndex=-1
hit=8: doc8=1.0 shardIndex=-1,   doc163=1.0 shardIndex=-1
hit=9: doc9=1.0 shardIndex=-1,   doc164=1.0 shardIndex=-1
hit=10: doc10=1.0 shardIndex=-1,         doc165=1.0 shardIndex=-1
hit=11: doc11=1.0 shardIndex=-1,         doc166=1.0 shardIndex=-1
hit=12: doc12=1.0 shardIndex=-1,         doc167=1.0 shardIndex=-1
hit=13: doc13=1.0 shardIndex=-1,         doc168=1.0 shardIndex=-1
hit=14: doc14=1.0 shardIndex=-1,         doc169=1.0 shardIndex=-1
hit=15: doc15=1.0 shardIndex=-1,         doc170=1.0 shardIndex=-1
hit=16: doc16=1.0 shardIndex=-1,         doc171=1.0 shardIndex=-1
hit=17: doc17=1.0 shardIndex=-1,         doc172=1.0 shardIndex=-1
hit=18: doc18=1.0 shardIndex=-1,         doc173=1.0 shardIndex=-1
hit=19: doc19=1.0 shardIndex=-1,         doc174=1.0 shardIndex=-1
hit=20: doc20=1.0 shardIndex=-1,         doc175=1.0 shardIndex=-1
hit=21: doc21=1.0 shardIndex=-1,         doc176=1.0 shardIndex=-1
hit=22: doc22=1.0 shardIndex=-1,         doc177=1.0 shardIndex=-1
hit=23: doc23=1.0 shardIndex=-1,         doc178=1.0 shardIndex=-1
hit=24: doc24=1.0 shardIndex=-1,         doc179=1.0 shardIndex=-1
for query:field:*
```

From the failure, I see all docs have the same score, but doc ids don't match. I assume that the column on the right targeted a leaf partition, which is why doc ids don't start from 0. I think that we are still early terminating though or you mean that we shouldn't have as there were docs with same score but lower doc ids? I do think that I am missing something though, or this test could have failed with inter-segment concurrency too, because it would otherwise rely on the first segment to be the first one being searched at all times?

jpountz:
TopScoreDocCollectorManager is supposed to return the top-N results that have the best score, tie-broken by doc ID. So (docID=0, score=1) compares better than (docID=155, score=1) and should appear first in the resulting top docs.

> I think that we are still early terminating though or you mean that we shouldn't have as there were docs with same score but lower doc ids?

Right.

Looking at the code, it looks like the problem is here: https://github.com/apache/lucene/blob/ff8b81afc59dc09ea4dfdd02a58398f08fcad8a8/lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java#L260. If a segment has two partitions, then its leaf collectors will collect hits with the same "docBase", so the tie-breaking logic will not know that hits from the first partition should compare better than hits from the second partition.

javanna:
Understood, thanks for the explanation, I did not realize doc ids played a role, I thought it was an incorrect assumption and it worked with inter-segment concurrency only by coincidence. I will go ahead and close this PR, and see if I can fix the production code instead of adapting the test. There's other recent failures around the same problem.

javanna:
I opened #13777 . It needs some proper review because I am not entirely sure what side effects it may have. There is currently no direct link between leaf collector and scorers, to pass along the minDocId that we could have used somehow as docBase in the maxScoreAccumulator.

PR #13774, opened 2024-09-12
Address length used to copy array in FacetsCollector to not be out of bounds

javanna:
This has caused a few recent test failures with jdk23 and jdk24. It should get fixed replacing the length with the length of the array that we copy from, rather than using the new length.


stefanvodita (approved):
Thank you for fixing this!

PR #13775, opened 2024-09-12
Lower number of hits in TestLargeNumHitsTopDocsCollector

javanna:
There's two tests where we use 250_000 as number of collected hits, but we only ever index max 2000 docs. That makes use create a priority queue of size 250_000 for each segment partition which causes out of memory errors when the number of partitions is higher than a few.

With this commit I propose that we lower the threshold to 2000 for those tests that need a high number of collected hits. The assumption that a priority queue is not built within the LargeNumHitsTopDocsCollector still holds so this change should not defeat the purpose of the tests.

### DIFF
--- a/lucene/sandbox/src/java/org/apache/lucene/sandbox/search/LargeNumHitsTopDocsCollector.java
+++ b/lucene/sandbox/src/java/org/apache/lucene/sandbox/search/LargeNumHitsTopDocsCollector.java
@@ -137,7 +137,7 @@ public final class LargeNumHitsTopDocsCollector implements Collector {
     }
 
     // Total number of hits collected were less than requestedHitCount
-    assert totalHits < requestedHitCount;
+    assert totalHits <= requestedHitCount;
     Collections.sort(
         hits,
         Comparator.comparing((ScoreDoc scoreDoc) -> scoreDoc.score)
--- a/lucene/sandbox/src/java/org/apache/lucene/sandbox/search/LargeNumHitsTopDocsCollector.java
+++ b/lucene/sandbox/src/java/org/apache/lucene/sandbox/search/LargeNumHitsTopDocsCollector.java
@@ -129,7 +129,7 @@ public final class LargeNumHitsTopDocsCollector implements Collector {
    */
   protected void populateResults(ScoreDoc[] results, int howMany) {
     if (pq != null) {
-      assert totalHits >= requestedHitCount;
+      assert totalHits > requestedHitCount;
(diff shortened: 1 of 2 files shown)


lucene/sandbox/src/java/org/apache/lucene/sandbox/search/LargeNumHitsTopDocsCollector.java
```diff
@@ -137,7 +137,7 @@ protected void populateResults(ScoreDoc[] results, int howMany) {
     }
 
     // Total number of hits collected were less than requestedHitCount
-    assert totalHits < requestedHitCount;
+    assert totalHits <= requestedHitCount;
```
javanna, line 140: this appears to be an existing bug

jpountz:
A property of this collector is that it's supposed to allocate RAM in order of the actual number of collected hits rather than the max number of hits to retrieve. So this change defeats a bit the prupose of the test. I would rather update the test to contain the total number of slices.

javanna:
Thanks for clarifying!

But then is it a feature that it goes out of memory now? Especially given that it's a side by side comparison, and what causes the OOM is the ordinary top docs search, and not the large num hits collector, which suggests that is indeed more optimized in terms of memory allocated. 

I would think that if memory usage is what we want to test, we should probably have a different test? Or maybe it makes sense to lower the threshold only for the top docs collector part, which is the one that causes issues and we are only using to compare final results? Note that the second search does not even use concurrency as it uses the deprecated `search(Query, Collector)`

jpountz:
> Or maybe it makes sense to lower the threshold only for the top docs collector part,

+1 to that, let's pass n=reader.numDocs()?

javanna:
> +1 to that, let's pass n=reader.numDocs()?

Sure, I made that change.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/LargeNumHitsTopDocsCollector.java
jpountz: I believe you're right. And then a few lines above it should be `assert totalHits > requestedHitCount;`?

jpountz (approved):

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/LargeNumHitsTopDocsCollector.java
javanna: good catch!

PR #13777, opened 2024-09-12
Replace docBase with actual docId in MaxScoreAccumulator

javanna:
We have been encoding docBase and the score in MaxScoreAccumulator#accumulate. That makes the assumption that segments are going to be processed in doc order and implements global max score accounting across segments searched concurrently.

With the introduction of intra-segment concurrency, the same segment may be seen multiple times, once per segment partition. Partitions are all going to have the same docBase, hence you may end up with topN results with higher docIds than expected, because the search early terminates before docs with same score and lower doc ids are seen.

This commit encodes the docId in the accumulator in place of the docBase to resolve the described issue.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
+++ b/lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
@@ -232,7 +232,7 @@ public abstract class TopScoreDocCollector extends TopDocsCollector<ScoreDoc> {
       // the next float if the global minimum score is set on a document id that is
       // smaller than the ids in the current leaf
       float score =
-          docBase >= maxMinScore.docBase() ? Math.nextUp(maxMinScore.score()) : maxMinScore.score();
+          docBase >= maxMinScore.docId() ? Math.nextUp(maxMinScore.score()) : maxMinScore.score();
       if (score > minCompetitiveScore) {
         assert hitsThresholdChecker.isThresholdReached();
         scorer.setMinCompetitiveScore(score);
@@ -254,10 +254,9 @@ public abstract class TopScoreDocCollector extends TopDocsCollector<ScoreDoc> {
         totalHitsRelation = TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO;
         minCompetitiveScore = localMinScore;
         if (minScoreAcc != null) {
-          // we don't use the next float but we register the document
-          // id so that other leaves can require it if they are after
-          // the current maximum
-          minScoreAcc.accumulate(docBase, pqTop.score);
+          // we don't use the next float but we register the document id so that other leaves or
+          // leaf partitions can require it if they are after the current maximum
+          minScoreAcc.accumulate(docBase + pqTop.doc, pqTop.score);
         }
       }
     }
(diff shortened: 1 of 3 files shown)


javanna:
I am not getting many test failures, which is surprising, because I think that this is entirely the wrong fix, in that it defeats  the early termination purpose of top score docs collector. I don't have good ideas on how to fix this properly so far, without API changes, in that leaf collector don't have a way to get notified that the current segment is seen as part of a segment partition. It would be enough to get the min doc id of the partition, and use it as doc base in the accumulator, but it's hard to get it there with the current API.

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
```diff
@@ -254,10 +254,9 @@ protected void updateMinCompetitiveScore(Scorable scorer) throws IOException {
         totalHitsRelation = TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO;
         minCompetitiveScore = localMinScore;
         if (minScoreAcc != null) {
-          // we don't use the next float but we register the document
-          // id so that other leaves can require it if they are after
-          // the current maximum
-          minScoreAcc.accumulate(docBase, pqTop.score);
+          // we don't use the next float but we register the document id so that other leaves or
+          // leaf partitions can require it if they are after the current maximum
+          minScoreAcc.accumulate(docBase + pqTop.doc, pqTop.score);
```
jpountz, line 259: pqTop.doc should already be a top-level doc ID, so adding `docBase`shouldn't be needed?

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
```diff
@@ -232,7 +232,7 @@ protected void updateGlobalMinCompetitiveScore(Scorable scorer) throws IOExcepti
       // the next float if the global minimum score is set on a document id that is
       // smaller than the ids in the current leaf
       float score =
-          docBase >= maxMinScore.docBase() ? Math.nextUp(maxMinScore.score()) : maxMinScore.score();
+          docBase >= maxMinScore.docId() ? Math.nextUp(maxMinScore.score()) : maxMinScore.score();
```
jpountz, line 235: In theory, we could replace this docBase with the current doc ID as well (not important for correctness, but could help skip more docs when some segments have multiple partitions). I'm good to make this change in a follow-up PR instead of this one if it helps.

jpountz (commented):
Besides one small comment, this looks like the correct change to me.

jpountz:
> I think that this is entirely the wrong fix

Why is it the wrong fix? This looks correct to me: we want to consider docs with equal or greater scores if the doc ID is less than the top-n-th hit so far, but only docs with a greater doc ID otherwise?

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
javanna: ah good point!

javanna:
> Why is it the wrong fix?

Happy to be proven wrong, especially when I believe that my fix is wrong. ahah

I was under the impression that we may no longer early terminate if we replace the docBase with the docId. A test failure on `TestConstantScoreScorer` seems to show that pretty clearly. I also thought that this is too easy to be true especially if it does not have side effects :)

javanna:
> A test failure on TestConstantScoreScorer seems to show that pretty clearly.

Ok, that was caused by using the docBase when already included in the doc id. Phew. This looks to be right after all! Sorry for the back and forth and thanks for the help.

lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
javanna: I see what you mean , I gave it a try, but then I reverted it, I prefer to think about testing and do it as a follow-up

javanna:
This should be good to go now :)

jpountz (approved):

Issue #13778, opened 2024-09-12
Support random access byte/float vector values as first class part of API 


msokolov:
### Description

While working on implementing BP reordering over vectors, the idea came up that we might be able to simplify our APIs by surfacing random access to vector values. Today it exists and every vector reader implements it, but the API hides it behind an interface (RandomAccessVectorValues) that requires casting to access it. I'm opening this issue to see how that might look and get feedback. This is intended to be a refactor that would make it easier to maintain and enhance these classes and shouldn't introduce functional changes.

PR #13779, opened 2024-09-12
First-class random access API for KnnVectorValues

msokolov:
addresses https://github.com/apache/lucene/issues/13778

Key things in this PR:

1. Introduces abstract `KnnVectorValues` from which `ByteVectorValues` and `FloatVectorValues` derive;
2. Folds `RandomAccessVectorValues` into `KnnVectorValues` thus eliminating some casts.
3. `RandomAccessVectorValues.Floats` becomes `FloatVectorValues` and `RandomAccessVectorValues.Bytes` becomes `ByteVectorValues`. `RandomAccessQuantizedByteVectorValues` folded into `QuantizedByteVectorValues`.
4. `IndexInput getSlice()` moved to a new `HasIndexSlice` interface.
5. Introduces `VectorEncoding KnnVectorValues.getEncoding()` to enable type-specific branches in a few places where we are dealing with abstract `KnnVectorValues` (tests only IIRC). Also used that to provide a default `getVectorByteLength()`.
6. `KnnVectorValues` no longer extends  `DocIdSetIterator`; rather it provides a tightly-coupled `iterator()`. This enables refactoring common iteration patterns that were repeated many times in the code base. This new iterator, `DocIndexIterator` provides an additional method `index()` analogous to `IndexedDISI`.

Some of the methods on `KnnVectorValues` have default impls that throw `UnsupportedOperationException` enabling subclasses to provide partial implementations and relying on testing to catch missing required methods. I'd like feedback on this. Should we provide implementations we never use, just to make these classes complete? That didn't make sense to me. But the previous alternative of attempting to provide strict adherence to declarative contracts was becoming in my view, overly restrictive and leading to hard-to-maintain code. Some of these readers would only ever be used iteratively. Random access is required for search, but not used when  merging the values themselves, and when we merge we do search, but using a temporary file so that searching is always done over a file-based value. Random access also gets used during merging when the index is sorted, again this is provided by specialized readers, so not every reader needs to implement random access. But the API maintenance is greatly simplified if we allow partial implementation. Anyway that is the idea I am trying out here. Can we live with a little less API purity and gain some simplicity and less boilerplate?

Notes for reviewers:

There is a lot of code change here, but much of it is repetitive. I recommend starting with `KnnVectorValues` and checking its `DocIndexIterator` inner class. The rest of the changes are basically consequences of introducing those abstrations in place of the `Random*Values` we removed.


msokolov:
another concern I have is how this would impact ongoing work to enable multiple vectors per doc/field. There would almost certainly be conflicts with that PR on the surface, but I hope this could actually simplify things in that the new `DocIndexIterator` class could be enhanced / extended to provide access to a series of values (maybe a list or array?) instead of (or in addition to?) a single one, possibly centralizing the required changes (since we have many fewer iterator implementations after this change).

benwtrent:
>  but I hope this could actually simplify things 

That is my intuition as well.

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
```diff
@@ -0,0 +1,281 @@
...
+        }
+      };
+    }
+  }
+
+  /**
+   * Creates an iterator for instances where every doc has a value, and the value ordinals are equal
+   * to the docids.
+   */
+  protected DocIndexIterator createDenseIterator() {
```
jpountz, line 167: Could we fix call sites to use `DocIdSetIterator#all(size())` instead?

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
```diff
@@ -0,0 +1,281 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.index;
+
+import java.io.IOException;
+import org.apache.lucene.codecs.lucene90.IndexedDISI;
```
jpountz, line 20: In general, we should avoid having oal.index APIs depend on codec-specific APIs. Can we move the vector values / IndexedDISI conversion to the lucene90 package instead?

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
```diff
@@ -0,0 +1,281 @@
...
+   */
+  public int ordToDoc(int ord) {
+    return ord;
+  }
+
+  /**
+   * Creates a new copy of this {@link KnnVectorValues}. This is helpful when you need to access
+   * different values at once, to avoid overwriting the underlying vector returned.
+   */
+  public abstract KnnVectorValues copy() throws IOException;
```
jpountz, line 60: I wonder if we could make the API a bit nicer by removing this copy() and instead have something like a `FloatVectorDictionary { float[] vectorValue(int ord); }` and a method here that can return a new `FloatVectorDictionary` (a bit like `SortedDocValues` and `TermsEnum`).

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
```diff
@@ -0,0 +1,281 @@
...
+
+  /**
+   * Return the iterator for this instance. If you need multiple iterators, call <code>
+   * this.copy().iterator()</code>.
+   */
+  public DocIndexIterator iterator() {
+    if (iterator == null) {
+      iterator = createIterator();
+    }
+    return iterator;
```
jpountz, line 98: Could we make this return a new iterator every time to make the API a bit nicer? From a quick look, it seems that call sites could easily be adjusted to not rely on this method returning a shared instance?

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
```diff
@@ -0,0 +1,281 @@
...
+        return docsWithField.cost();
+      }
+    };
+  }
+
+  /**
+   * Creates an iterator from this instance's ordinal-to-docid mapping which must be monotonic
+   * (docid increases when ordinal does).
+   */
+  protected DocIndexIterator fromOrdToDoc() {
```
jpountz, line 245: nit: could we make it look a bit more like `DocIdSetIterator#all` by moving it to `DocIndexIterator#all`?

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
```diff
@@ -0,0 +1,281 @@
...
+   */
+  public abstract static class DocIndexIterator extends DocIdSetIterator {
+
+    /** return the value index (aka "ordinal" or "ord") corresponding to the current doc */
+    public abstract int index();
+
+    @Override
+    public int advance(int target) throws IOException {
+      return slowAdvance(target);
+    }
```
jpountz, line 121: This looks like it could be a performance trap, which is why DocIdSetIterator offers this helper method without making it the default impl. Should we leave it without a default impl here too?

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
```diff
@@ -0,0 +1,281 @@
...
+
+    @Override
+    public long cost() {
+      throw new UnsupportedOperationException();
```
jpountz, line 125: Likewise here, I'd rather leave it unimplemented to force implementers to decide if having cost() throw an exception is fine. Presumably, most of the time it's not.

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  /**
+   * Creates a new copy of this {@link KnnVectorValues}. This is helpful when you need to access
+   * different values at once, to avoid overwriting the underlying vector returned.
+   */
+  public abstract KnnVectorValues copy() throws IOException;
+
+  /** Returns the vector byte length, defaults to dimension multiplied by float byte size */
+  public int getVectorByteLength() {
```
jpountz, line 63: should we make it final (or even remove this method since it's a bit trivial?) presumably, there is no other return value that makes sense?

jpountz (commented):
I left a few thoughts/questions. In general, I see how such a random-access API change could help with e.g. your BP reordering work and be valuable in general. I was wondering if this API may be too tailored to HNSW and prevent us from supporting other interesting algorithms, but actually I don't think that this is the case?

jpountz:
Am guessing correctly that you're targeting 10.0 for this change?

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
msokolov: I think it can possibly vary due to quantization --- but you may be right and it refers only to the original size?  Will have to confirm, or maybe @benwtrent can comment

msokolov:
Thanks for the quick review! I will get started on addressing. As for timeline for this change, it would definitely be convenient to get in to 10.0 release. I think you had said 9/22 would be a feature freeze date; it seems we could possibly meet that timeline. I will be traveling starting tomorrow for a week, but I should be able to put in some quality time on the plane LOL

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
msokolov: I'm not sure how to do that since we need to return a DocIndexIterator (that provides `index()` method). I suppose we could move `index()` to `DocIdSetIterator` but that seems like  a lot more disruption could result, or perhaps you meant we would do that and leave it typically unimplemented?

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
msokolov: yes, good

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
msokolov: Let me try - I was also a bit unhappy about this, but at one point along this journey I was relying on being able to recover the shared state - maybe I finally was able to get rid of that and just didn't notice!

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
msokolov: ah, you mean rename this method to `all`? sure, makes sense

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
msokolov: yes, I don't think anything relies on this, makes sense

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
msokolov: hmm I think cost() is rarely used in the vector reader/writers which instead are concerned with KnnVectorValues.size() -- they typically want to know how many vector values there are and to the extent they care about the number of docs it's only when they must iterate through all of them and have no use for an estimate. These iterators aren't really used during searching?

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
msokolov: The way `SortedDocValuesTermsEnum` is, calling its `next` method will overwrite the internal buffer ofd the `SortedDocValues` on which it is built, defeating the  purpose of `copy()` which is to provide two completely independent sources. Another thing we could do is to add `vectorValue(int ord, float[] scratch)` allowing the caller to provide the memory to write to. If we had that, we wouldn't need `copy()`. Maybe we could manage to squeeze that into 10.0 too, but I'd rather do it in a separate PR

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
jpountz: Ah sorry, I had misread DocIdSetIterator when it's a DocIndexIterator!

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
jpountz: But if you call SortedDocValues#termsEnum twice, this would give you two independent sources of terms?

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java
```diff
@@ -361,33 +385,46 @@ private MergedByteVectorValues(List<ByteVectorValuesSub> subs, MergeState mergeS
...
         }
         size = totalSize;
-        docId = -1;
       }
 
       @Override
-      public byte[] vectorValue() throws IOException {
-        return current.values.vectorValue();
+      public byte[] vectorValue(int ord) throws IOException {
+        return current.values.vectorValue(current.values.iterator().index());
```
jpountz, line 392: This part feels a bit hacky, could we instead merge the ord->vector mappings of the various vector values by concatenating them?

jpountz:
>  think you had said 9/22 would be a feature freeze date

I was thinking of doing it next week, but we can backport this PR even though the branch has been cut if it looks ready/safe.

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
jpountz: If we default cost() to returning size(), that would work for me. But I'm not comfortable with having implementations of DocIdSetIterator#cost that may throw, which means e.g. that they cannot be put in a Conjunction DISI(which will want to sort its clauses by cost).

lucene/core/src/java/org/apache/lucene/codecs/lucene95/HasIndexSlice.java
```diff
@@ -14,23 +14,16 @@
...
-
-  ScalarQuantizer getScalarQuantizer();
-
-  float getScoreCorrectionConstant(int vectorOrd) throws IOException;
+public interface HasIndexSlice {
 
-  @Override
-  RandomAccessQuantizedByteVectorValues copy() throws IOException;
+  /** Returns an IndexInput from which to read this instance's values. */
+  IndexInput getSlice();
```
ChrisHegarty, line 28: I very much like this, and had something similar in a past unmarked PR. 👍

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
ChrisHegarty: I always found copy very strange, but I get why it is there. I'd be tempted to leave it as is in this PR, changing the access model and cache of 1 float[] will be a bit tricky.

ChrisHegarty (commented):
I really like this change. I see a lot of refactoring similar to what I half started at one point or the other, but never finished. There are some specific comments to be addressed, but otherwise the approach LGTM.

msokolov:
OK there seem to be some test failures ... I did a complete run, but randomized testing always seems to ferret out something interesting!

msokolov:
Regarding the rename of `fromOrdToDoc` to `all` I think it was not helpful and plan to revert or maybe come up with some other name. The problem is we also have `createDenseIterator` which is also `all`. Essentially we have Sparse and Dense all-iterators. Maybe instead of fromOrdToDoc we can say `createSparseIterator`?

jpountz:
FWIW I started playing with removing copy() by replacing it with a factory method for a dictionary: https://github.com/msokolov/lucene/commit/ae7aca32a690a4b21a3da793258ce17560b551e7. Not sure how far I'll go. :)

msokolov:
I also just started trying to replace `copy()` with the approach of adding `vectorValue(int ord, float[] outValue)` although this does add a copy operation in some cases where previously we would expose internal storage so I'm not sure it's great

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
navneet1v: +1. Even in FloatVecotorValues cost() is returning size() only. https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/FloatVectorValues.java#L46-L48

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java
```diff
@@ -303,29 +314,45 @@ private MergedFloat32VectorValues(List<FloatVectorValuesSub> subs, MergeState me
...
+        };
       }
 
       @Override
-      public int advance(int target) {
-        throw new UnsupportedOperationException();
+      public float[] vectorValue(int ord) throws IOException {
+        // FIXME what can we assert here?
+        // assert ord == iterator.index();
+        return current.values.vectorValue(current.index());
```
benwtrent, line 355: This is the biggest "gotcha" in this whole thing I think. Does `DocIdMerger` allow random access at all?

It seems like `vectorValue(ord)` should also be able to jump between `current` sub iterators and move backwards and forwards. But, I don't think `DocIdMerger` allows backwards movement at all.

benwtrent: I think unless we can fix `DocIdMerger`, we should throw an error here indicating that only forward iteration is allowed.

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
benwtrent: This cannot be final. Its dependent on parameters like `compressed`.

lucene/core/src/java/org/apache/lucene/index/KnnVectorValues.java
benwtrent: I agree here. Either it should default to `size()` via some provided dependency or it shouldn't implement at all and force sub-classes.

lucene/core/src/java/org/apache/lucene/util/quantization/QuantizedByteVectorValues.java
```diff
@@ -17,41 +17,42 @@
...
+  public VectorScorer scorer(float[] query) throws IOException {
+    throw new UnsupportedOperationException();
+  }
 
   @Override
-  public final long cost() {
-    return size();
+  public QuantizedByteVectorValues copy() throws IOException {
+    return this;
   }
```
benwtrent, line 52: defaulting copy to `this` is dangerous. I would recommend against it.

lucene/core/src/java/org/apache/lucene/util/quantization/ScalarQuantizer.java
```diff
@@ -269,11 +270,12 @@ static ScalarQuantizer fromVectors(
     if (totalVectorCount == 0) {
       return new ScalarQuantizer(0f, 0f, bits);
     }
+    KnnVectorValues.DocIndexIterator iterator = floatVectorValues.iterator();
```
benwtrent, line 273: I think this is ok for now, but this quantization code can be made much simpler if indeed we can randomly access even across various merged doc sub iterators.

lucene/core/src/java/org/apache/lucene/index/FloatVectorValues.java
```diff
@@ -27,34 +27,23 @@
...
-   *
-   * @return the number of vectors returned by this iterator
-   */
-  public abstract int size();
-
   @Override
-  public final long cost() {
-    return size();
+  public FloatVectorValues copy() throws IOException {
+    return this;
```
benwtrent, line 37: we shouldn't default to `this` that is dangerous. If an off-heap thing that assumes caching and doesn't allow multi-threaded access overrides but forgets to override, we are in a bad place.

lucene/core/src/java/org/apache/lucene/index/ByteVectorValues.java
```diff
@@ -27,34 +27,23 @@
...
-   *
-   * @return the number of vectors returned by this iterator
-   */
-  public abstract int size();
-
   @Override
-  public final long cost() {
-    return size();
+  public ByteVectorValues copy() throws IOException {
+    return this;
```
benwtrent, line 37: again, this default is dangerous to me.

benwtrent (commented):
I like the simplicity we gain here.

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java
jpountz: I had thought we'd fix it by concatenating the vector dictionaries of each segment. Then based on the requested vector ordinal, you could compute the segment that the vector belongs to via `ReaderUtil#subIndex`.

jpountz:
I iterated a bit on my branch, so that there is no more call site for `FloatVectorValues#copy`: https://github.com/msokolov/lucene/compare/knn-vector-random...jpountz:lucene:knn-vector-random?expand=1. The API looks better to me this way, but I'm keen on getting feedback before pursuing as this is quite a tedious refactoring.

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java
msokolov: There is something similar here in `SlowCompositeCodecReaderWrapper` - binary search across sub-values

msokolov:
FWIW I tried removing `copy()` and using caller-supplied storage in `vectorValue`. In many ways this looks nicer, but it leads to substantial slowdown in indexing/merging because of the additional copy required, so I think it's not really viable. We could provide *both* `vectorValue(int)` and `vectorValue(int, float[])` and avoid `copy()` that way. Maybe that's best

msokolov:
ooh I just saw the Dictionary branch - that looks like a nice approach, I don't think I really understood what you were proposing before. One Q I have is can we remove copy()? Do we need to deprecate it first -- and if so, could we deprecate in 9x branch? Or perhaps it is too late for that

benwtrent:
> One Q I have is can we remove copy()? Do we need to deprecate it first -- and if so, could we deprecate in 9x branch?

You are already removing `RandomAccessVectorValues`, which is the thing that used `copy()`, so, I am not sure what there is to deprecate 😂

msokolov:
I'll post one more iteration here addressing the concerns about dangerous default impls that adds back impls of copy() and cost(). I also added a test-and-throw ensuring that the vectorValues impls that require forward-iteration enforce it. We can fully implement random access later without breaking any APIs. 

I also think we should go ahead with Adrien's Dictionary idea, but do this in two steps because there is a lot going on here already.

benwtrent:
The dictionary idea is OK, but I still don't see how it removes `copy()`. Besides the caching of values, copy gives us multi-threaded safety by copying the underlying index readers. Otherwise we are using the same reader between threads. For concurrent merging of graphs, this is important.

I agree, any further refactoring should be done in another PR.

msokolov:
I think the idea w/Dictionary is that callers, instead of calling `copy().vectorValue(int ord)` would call `dictionary().vectorValue(int ord)`. So then the scratch vector storage (if needed) would be in the Dictionary not in the VectorValues, and thus not shared by multiple users of the same values instance.  In some sense it's not very different, but in the sense that the Dictionary has a much more limited API than the source it came from, it is different.

jpountz:
Exactly. I tried to model it similarly to what doc values do, where `SortedDocValues#termsEnum()` returns a dictionary with a different backing IndexInput clone on every call.

msokolov:
hm interesting there was an EOFException in there - I'll dig

msokolov:
OK, I found an off-by-one error plus a problem with lazy iterator creation that slipped in when we got rid of createIterator(). It makes me a little nervous these didn't show up in earlier testing. I'm now running with `tests.iter=20`

msokolov:
I do not understand what happened here - somehow github is checking out different code than I pushed?? Will try pushing again??

msokolov:
OK, I think this is ready after a few minor issues have been addressed. I opened https://github.com/apache/lucene/issues/13831 to track replacing copy() with dictionary()

javanna:
Should there be a migrate entry added with this change?

msokolov:
> Should there be a migrate entry added with this change?

oh thanks, yes, and a CHANGES entry. I opened https://github.com/apache/lucene/pull/13833 if you want to review

PR #13252, opened 2024-03-30
Use Arrays.compareUnsigned instead of iterating compare.

vsop-479:
Similar to `SegmentTermsEnumFrame`, we can use `Arrays.compareUnsigned` instead of iterating compare, in `SegmentTermsEnum.seekCeil` and `SegmentTermsEnum.seekExact`.

PR #13782, opened 2024-09-13
Use Arrays.compareUnsigned  in IDVersionSegmentTermsEnum and OrdsSegmentTermsEnum.

vsop-479:
### Description

Description
This change similars to https://github.com/apache/lucene/pull/13252.


vsop-479:
@uschindler 

lucene/CHANGES.txt
```diff
@@ -292,6 +292,8 @@ Build
 API Changes
 ---------------------
 
+* GITHUB#13782: Replace handwritten loops compare with Arrays.compareUnsigned in IDVersionSegmentTermsEnum and OrdsSegmentTermsEnum. (zhouhui)
```
jpountz, line 295: We usually put under "API changes" things that require users to change their code upon upgrading, which is not the case here. Can you move it under "Optimizations" instead?

jpountz (approved):

lucene/CHANGES.txt
vsop-479: Thanks @jpountz . That is a mistake, I fixed it.

vsop-479:
@jpountz 
I think this check failure relats to https://github.com/apache/lucene/commit/5045d3c67b18d23c534a32cee1d95827b0b7c482 .

jpountz:
Indeed it likely does, can you merge main back into your branch?

uschindler (approved):
Hi,
We had others like this in former PRs (e.g., https://github.com/apache/lucene/pull/13252). As this old code looks like copy pasted over and over, can we find other occurrences using some regex searches?

jpountz:
I would backport as it looks pretty safe, though I doubt we have anyone using this postings format, so it likely doesn't matter much in practice.

vsop-479:
> can you merge main back into your branch?

Merged.

> can we find other occurrences using some regex searches?

I searched code with `final int targetUptoMid = targetUpto` which these iterating compare start with. 
There  is only two places remains in `org.apache.lucene.backward_codecs.lucene40.blocktree.SegmentTermsEnum`, since this is a bakward class, I am not sure we should modify it.

uschindler:
> I would backport as it looks pretty safe, though I doubt we have anyone using this postings format, so it likely doesn't matter much in practice.

The we should move the changes entry. Do we have a corresponding 9.x section already?

vsop-479:
> The we should move the changes entry. Do we have a corresponding 9.x section already?

We already have a similar change entry for https://github.com/apache/lucene/pull/13252 under 9.11.0, so we should move this change entry?

uschindler:
> > The we should move the changes entry. Do we have a corresponding 9.x section already?
> 
> We already have a similar change entry for #13252 under 9.11.0, so we should move this change entry?

Yes, let's merge them an list both PRs.

vsop-479:
Ok, I will do it.

vsop-479:
> let's merge them an list both PRs.

I merged them into one entry.

vsop-479:
> can we find other occurrences using some regex searches?

Hmm, I also find some suffix's loop comparing in `IDVersionSegmentTermsEnumFrame` and `OrdsSegmentTermsEnumFrame`, similar to `SegmentTermsEnumFrame`, these code maybe could replaced by `Arrays#compareUnsigned` too.

But, firstly I find we attempt to load sub block when scanning a leaf block: https://github.com/apache/lucene/pull/13786.

vsop-479:
> I replaced loop compare suffixe with Arrays#compareUnsigned in IDVersionSegmentTermsEnumFrame and OrdsSegmentTermsEnumFrame.

@jpountz It seems you implemented the same thing in `SegmentTermsEnumFrame`. Please take a look when you get a chance.

vsop-479:
Hello @jpountz , Please take a look when you get a chance.

vsop-479:
> The we should move the changes entry. Do we have a corresponding 9.x section already?

> We already have a similar change entry for https://github.com/apache/lucene/pull/13252 under 9.11.0, so we should move this change entry?

> Yes, let's merge them an list both PRs.

@uschindler Should I split merged change entry, we are on 10.x now.

uschindler (approved):
I think this should be merged. Did not check all details, but looks like a good cleanup.

uschindler:
Yes we should split changes entry

uschindler (approved):

PR #13784, opened 2024-09-13
Change docValuesSkipIndex from a boolean to an enum.

jpountz:
At the moment, our skip indexes record min/max ordinal/value per range of doc IDs. It would be natural to extend it to other pre-aggregated data such as a sum and value count, which facets could take advantage of. This change switches `docValuesSkipIndex` from a boolean to an enum so that we could release such changes in the future in an additive fashion, by adding constants to this enum and new methods to `DocValuesSkipper`.


iverase (approved):
make sense, it will make it more natural to add on it.

PR #13785, opened 2024-09-13
Fix Flaky Test In TestBlockJoinBulkScorer

Mikep86:
Fix the `testSetMinCompetitiveScoreWithScoreModeMax` test, which sometimes failed due to randomizations in how the docs were scored. This fix should also be backported to `branch_9x`.

### DIFF
--- a/lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinBulkScorer.java
+++ b/lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinBulkScorer.java
@@ -27,6 +27,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field;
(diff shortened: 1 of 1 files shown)


lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinBulkScorer.java
```diff
@@ -283,7 +292,33 @@ public void collect(int doc) throws IOException {
...
+        if (actual == null) {
+          foundMatch = true;
+          break;
+        }
+      } else {
+        foundMatch = expected.equals(actual);
+        if (foundMatch) {
+          break;
+        }
+      }
```
jpountz, line 316: nit: maybe merge these two conditions under a single `if (Objects.equals(expected, actual))` which would take care of the `null` case?

lucene/join/src/test/org/apache/lucene/search/join/TestBlockJoinBulkScorer.java
```diff
@@ -283,7 +292,33 @@ public void collect(int doc) throws IOException {
         null,
         0,
         NO_MORE_DOCS);
-    assertEquals(expectedScores, actualScores);
+
+    if (expectedScoresList.size() == 1) {
+      assertEquals(expectedScoresList.getFirst(), actualScores);
+    } else {
+      assertEqualsToOneOf(expectedScoresList, actualScores);
+    }
```
jpountz, line 300: I'm curious why you don't call `assertEqualsToOneOf` all the time, shouldn't it also work in the size==1 case?

Mikep86: It should and I originally did that, but if you use `assertEquals` there are nicer IDE integrations for value comparison on failure. Given that this method is called with a single map of expected scores in nearly all cases, I tried to preserve that functionality. This could help in the case of a failure in `testScoreRandomIndices`, which could have a large random score map.

javanna:
Thanks for fixing this!

PR #13786, opened 2024-09-14
Remove recurse into sub block when scan leaf block in IDVersionSegmentTermsEnumFrame#scanToTermLeaf.

vsop-479:
It is a leaf block.


vsop-479:
@mikemccand 

vsop-479:
Fixed in https://github.com/apache/lucene/pull/13782.

PR #13787, opened 2024-09-14
Fix comment on compare suffix and target in SegmentTermsEnumFrame.

vsop-479:
### Description

Since we do not loop over bytes (hand-written) in the suffix, to compare to the target anymore.


mikemccand (approved):
Thanks @vsop-479

PR #13790, opened 2024-09-14
Remove CollectorManager#forSequentialExecution

javanna:
We have recently (see #13735) introduced this utility method that creates a collector manager which only works when a searcher does not have an executor set, otherwise it throws exception once we attempt to create a new collector for more than one slice.

While we discussed it should be safe to use in some specific scenarios like the monitor module, we should be careful exposing this utility publicly, because while we'd like to ease migration from the search(Query, Collector) method, we may end up making users life even worse, in that it exposes them to failures whenever an executor is set and there is more than one slice created, which is hard to follow and does not provide a good user experience. What is these users do have an executor set to their searcher and think that this method provides a collector manager that is able to turn off concurrency (like search(Query, Collector) does currently), which is not possible?

My proposal is that we use a similar collector manager locally, where safe and required, but we don't expose it to users. In most places, we should rather expose collector managers that do support search concurrency, rather than working around the lack of those. Even for us committers, it has not been entirely clear when this is safe and when it's not, which indicates that this is something that users will likely struggle with.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -311,9 +311,6 @@ API Changes
 * GITHUB#13568, GITHUB#13750: Add DrillSideways#search method that supports any CollectorManagers for drill-sideways dimensions
   or drill-down. (Egor Potemkin)
 
-* GITHUB#13735: Add CollectorManager#forSequentialExecution to make CollectorManager creation more convenient
-  for users of the deprecated IndexSearcher#search(Query, Collector). (Greg Miller)
-
 New Features
 ---------------------
 
--- a/lucene/core/src/java/org/apache/lucene/search/CollectorManager.java
+++ b/lucene/core/src/java/org/apache/lucene/search/CollectorManager.java
@@ -18,8 +18,6 @@ package org.apache.lucene.search;
 
 import java.io.IOException;
 import java.util.Collection;
-import java.util.concurrent.Executor;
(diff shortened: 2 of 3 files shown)


ChrisHegarty (approved):
LGTM

javanna:
@gsmiller how do you feel about this one?

gsmiller:
tl;dr: I agree with removing this. 

I was [initially hesitant](https://github.com/apache/lucene/pull/13735#issuecomment-2338340094) to add this for a lot of the same reasons, but was convinced when realizing we could at least explicitly fail by asserting that more than one collector is never created (so at least we wouldn't be silently creating tear-your-hair-out concurrency bugs). But looking at the bigger picture, the whole point of removing `IndexSearcher#search(Query, Collector)` is to help users avoid traps situations where they think they're setup for concurrent search but not utilizing it. I agree it's a worse trap in a lot of ways to introduce this failure mode. I particularly do not like that this "helper" collector manager can't _actually_ check if it's being used with a concurrency-ready IndexSearcher. So I can imagine a real failure case where IndexSearchers are being created with an Executor but only have a single segment and things appear to work—but then break later (e.g., users that are force-merging their segments, or users that have sparse test coverage that only includes testing over single segments, etc.).

gsmiller (approved):

javanna:
Thanks for the feedback @gsmiller I will go ahead and merge this.

PR #13793, opened 2024-09-16
Make Operations#optional create simpler automata.

jpountz:
In the common case when the input automaton has no transition to state 0, the optional automaton can be created by marking state 0 as accepted.


rmuir (approved):
thanks @jpountz, this is another good practical one, also used by RegExp.

PR #13794, opened 2024-09-16
Deprecate BulkScorer#score(LeafReaderContext, Bits)

javanna:
We have removed BulkScorer#score(LeafReaderContext, Bits) in main in favour of BulkScorer#score(LeafCollector collector, Bits acceptDocs, int min, int max) as part of #13542. This commit deprecates the method in 9.x. Internal usages of it are left untouched as there may be subclasses that override it, which we don't want to break in a minor release.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/BulkScorer.java
+++ b/lucene/core/src/java/org/apache/lucene/search/BulkScorer.java
@@ -34,7 +34,8 @@ public abstract class BulkScorer {
    * @param acceptDocs {@link Bits} that represents the allowed documents to match, or {@code null}
    *     if they are all allowed to match.
    * @deprecated in favour of {@link #score(LeafCollector, Bits, int, int)}. Callers should instead
-   *     call the method variant that takes min and max as arguments. Subclasses that override it
+   *     call the method variant that takes min and max as arguments, providing 0 and {@link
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/search/BulkScorer.java
```diff
@@ -33,7 +33,11 @@ public abstract class BulkScorer {
    * @param collector The collector to which all matching documents are passed.
    * @param acceptDocs {@link Bits} that represents the allowed documents to match, or {@code null}
    *     if they are all allowed to match.
+   * @deprecated in favour of {@link #score(LeafCollector, Bits, int, int)}. Callers should instead
+   *     call the method variant that takes min and max as arguments. Subclasses that override it
```
jpountz, line 37: maybe also suggest passing min=0 and max=DocIdSetIterator.NO_MORE_DOCS here?

jpountz (approved):

lucene/core/src/java/org/apache/lucene/search/BulkScorer.java
javanna: good idea

javanna:
Thanks @jpountz !

PR #13795, opened 2024-09-16
Cleanup redundant allocations and code around Comparator use

original-brownbear:
Noticed some visible allocations in CompetitiveImpactAccumulator during benchmarking and fixed the needless allocation for the comparator in that class as well as a couple other similar spots where needless classes and/or objects could easily be replaced by more lightweight solutions.
Also inlined some single use comparator constants where a class turned into a non-capturing lambda for easier readability.


jpountz (approved):
Regardless of allocation, I like that this makes the code more readable in most cases by moving the comparison logic where it's used.

PR #13803, opened 2024-09-18
Copy stored fields during flush with index sort

dnhatn:
This change avoids copying stored fields individually when flushing with index sorting. This optimization can reduce flushing time, especially in cases with many stored fields.


dnhatn:
@jpountz I tried using merges for this, but it didn't work because merges expect the documents to be already sorted by index sorts.

jpountz:
Oh, interesting. I'm curious if you are able to quantify some speedups, e.g. by modifying luceneutil's StoredFieldsBenchmark (https://github.com/mikemccand/luceneutil/blob/main/src/main/perf/StoredFieldsBenchmark.java) to enable an index sort and measure the speedup?

Do we have good existing tests for this optimization?

dnhatn:
I ran the benchmarks multiple times, but the improvement seems to be more noise than a real gain, so I'm closing this PR.

PR #13816, opened 2024-09-21
Fix contention in StringHelper.randomId

original-brownbear:
Seeing this mutex contended up to O(10ms) in Elasticsearch at times. Moving to CAS and removing the unnecessary alloction of a new instance for the bitwise-and with the mask makes this perform+scale much better.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/StringHelper.java
+++ b/lucene/core/src/java/org/apache/lucene/util/StringHelper.java
@@ -22,6 +22,7 @@ import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.Properties;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * Methods for manipulating strings.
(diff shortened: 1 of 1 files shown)


rmuir:
profilers often lie about this. I think we should keep the synchronization simple. the IDs are important to not duplicate.

original-brownbear:
> profilers often lie about this

True but here I'd say whatever number the profiler outputs is only part of the cost of hard synchronizing on a global like we do here (we also have the indirect cost from cache/memory effects). Plus what the profiler isn't wrong about is the fact that threads actually go into waiting in this spot when benchmarking systems with large core/thread counts.

That said: If I can't convince you of the CAS fix. How about adjusting the math like I did here and moving the `toByteArray` call out of the `synchronized` block? That has pretty much the same effect because the code in the sychronized becomes fast enough to remove the thread-waiting in my testing.

rmuir:
this ID is generated for safety, and we only calculate this **once per segment** when initially flushing it. Doesn't even happen at merge. Really, it should be a few insns under lock, shouldn't need to be fast. That's why there isn't contention, your profiler is lying to you. Sorry, I don't buy 30ms.

And if somehow you are writing THAT many segments concurrently, then you definitely need the safety (and we dont want thread visibility issues where 2 segments end out with the same ID).

If we want to improve it, IMO the best way is to remove custom BigDecimal and xor/shift logic, but instead swap in the OpenJDK implementation. This code was written before OpenJDK had decent PRNGs.

This will clean up the code AND improve the performance.

```java
// initialization
var factory = RandomGeneratorFactory.of("Xoroshiro128PlusPlus");
var generator = factory.create(seed);

// generation
byte[] result = new byte[ID_LENGTH];
synchronized(idLock) {
  random.nextBytes(result);
}
return result;
```

Instead of two BigDecimal operations and an allocation under the lock, it will just do some xors and shifts on two `long` variables behind the scenes. And I think I would trust it more since code is maintained in the JDK.

PR #13819, opened 2024-09-24
Prevent flattening of ordered and unordered interval sources

jimczi:
This commit removes the flattening of ordered and unordered interval sources, as it alters the gap that parent intervals see. For example, ordered("a", ordered("b", "c")) should result in a different gap compared to ordered("a", "b", "c").
The flattening that is currently applied prevent some valid queries to match since the computed gap for ordered and unordered sources are not preserved when combining sources.

Phrase/Block operators will continue to flatten their sub-sources since this does not affect the inner gap (which is always 0 in the case of blocks).


romseygeek (approved):
LGTM

Issue #13821, opened 2024-09-25
Add test backward compatibility indices for 8.11.4


ChrisHegarty:
Nightly smoke tester failed last night:
https://github.com/apache/lucene/actions/runs/11025229258/job/30619746011

Releases that don't seem to be tested:
  8.11.4
Traceback (most recent call last):
...
RuntimeError: some releases are not tested by TestBackwardsCompatibility?

ChrisHegarty:
I tried, but was unsuccessful in building the old _lucene-solr_ repo. :-(

@HoustonPutman  can you please run `dev-tools/scripts/addBackcompatIndexes.py 8.11.4`, and add these backwards compat indices to `main`, `branch_9x`, and `branch_9_12`.

PR #13823, opened 2024-09-25
[9.12] Add support and compat testing for 8.11.4

ChrisHegarty:
This commit adds support and compat testing for 8.11.4, to 9.12.

relates #13821


HoustonPutman:
So I'm looking at the last time I did this for 8.11.3 (https://github.com/apache/lucene/commit/521c26e4fb547bb0aeafdbea838aec038ceb1538), and it looks like there were more things added. Is that not needed anymore? (I know the back compat testing was changing a lot around that time, so I'm not sure what is actually needed now)

ChrisHegarty:
> So I'm looking at the last time I did this for 8.11.3 ([521c26e](https://github.com/apache/lucene/commit/521c26e4fb547bb0aeafdbea838aec038ceb1538)), and it looks like there were more things added. Is that not needed anymore? (I know the back compat testing was changing a lot around that time, so I'm not sure what is actually needed now)

The list of versions has been moved from a string in source to a separate versions file.

HoustonPutman (approved):
Awesome, thanks again!

uschindler (commented):
looks fine.
This also applies to branch_9x, right?

PR #13824, opened 2024-09-25
Add 8.11.4 back compat indices

ChrisHegarty:
This commit add 8.11.4 back compat indices.

relates #13821


uschindler (commented):
Looks fine.

In main, do we not need to add a version constant in the Versions.java file?

PR #12842, opened 2023-11-24
Use group-varint encode the positions

easyice:
Thanks the suggestion from @jpountz , as  discussed in https://github.com/apache/lucene/issues/12826

This PR use group-varint to encode some vint values if `storeOffsets` is true, it's still using class `GroupVIntReader` and `GroupVIntWriter`,  i will update it after https://github.com/apache/lucene/pull/12841 is finished.

Currently i don't use group-vint if `(storeOffsets==false && storePayload==false)`, which means only `token` is stored, because i'm worried that it will use extra memory when bulk decoding. Feel free to correct me.

Then benchmark and file size changes i'll add next week.

PR #12841, opened 2023-11-24
Move group-varint encoding/decoding logic to DataOutput/DataInput

easyice:
From issue: https://github.com/apache/lucene/issues/12826

The JMH benchmark with this PR on my Mac (Intel chip) :

java 17
```
Benchmark                                   (size)   Mode  Cnt  Score   Error   Units
GroupVIntBenchmark.byteArrayReadGroupVInt       64  thrpt    5  5.519 ± 0.270  ops/us
GroupVIntBenchmark.byteArrayReadVInt            64  thrpt    5  4.075 ± 2.868  ops/us
GroupVIntBenchmark.byteBufferReadGroupVInt      64  thrpt    5  7.464 ± 1.618  ops/us
GroupVIntBenchmark.byteBufferReadVInt           64  thrpt    5  5.179 ± 0.470  ops/us
```

java 21
```
Benchmark                                   (size)   Mode  Cnt   Score   Error   Units
GroupVIntBenchmark.byteArrayReadGroupVInt       64  thrpt    5   5.768 ± 0.305  ops/us
GroupVIntBenchmark.byteArrayReadVInt            64  thrpt    5   5.255 ± 0.110  ops/us
GroupVIntBenchmark.byteBufferReadGroupVInt      64  thrpt    5  11.551 ± 0.252  ops/us
GroupVIntBenchmark.byteBufferReadVInt           64  thrpt    5   5.611 ± 0.266  ops/us
```

PR #13825, opened 2024-09-25
Drop final modifier on the public DataInput.readGroupVInts method

### LINKED ISSUE #13820: DataInput class can't be used with delegation pattern
### Description

The change in #12841 brought this pair of methods to DataInput:
```
public final void readGroupVInts(long[] dst, int limit) throws IOException { ... }
protected void readGroupVInt(long[] dst, int offset) throws IOException { ... }
```

If one wishes to write a DataInput delegating to another, existing DataInput, it's no longer possible because the public method is final and the protected method cannot be invoked on the delegate (cross-package access). A corresponding method in DataOutput is not final.

I would remove the final modifier from the API for symmetry. Thoughts?

_No response_

dweiss:
Fixes #13820

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/DataInput.java
+++ b/lucene/core/src/java/org/apache/lucene/store/DataInput.java
@@ -107,7 +107,7 @@ public abstract class DataInput implements Cloneable {
    * @param limit the number of int values to read.
    * @lucene.experimental
    */
-  public final void readGroupVInts(long[] dst, int limit) throws IOException {
+  public void readGroupVInts(long[] dst, int limit) throws IOException {
     int i;
     for (i = 0; i <= limit - 4; i += 4) {
       readGroupVInt(dst, i);
@@ -118,7 +118,7 @@ public abstract class DataInput implements Cloneable {
   }
 
   /**
-   * Override if you have a efficient implementation. In general this is when the input supports
+   * Override if you have an efficient implementation. In general this is when the input supports
    * random access.
    */
   protected void readGroupVInt(long[] dst, int offset) throws IOException {


mikemccand (approved):
Looks good, but I hope this does not somehow mess up hotspot compilation ... let's watch nightly benchy after this merges?

rmuir:
Isn't the root cause of the issue the `protected` method? The `final` method just invokes other methods, so alternatively we could fix the visibility of the protected method to be public like all the others?

dweiss:
> Isn't the root cause of the issue the `protected` method? The `final` method just invokes other methods, so alternatively we could fix the visibility of the protected method to be public like all the others?

Yes, this would also solve the problem. I wonder if it makes sense to expose that protected method though, if it's internal details of the implementation? I'd rather make the primary API method public and let subclasses do whatever they want with it (delegate, reimplement, etc.)

rmuir:
> Yes, this would also solve the problem. I wonder if it makes sense to expose that protected method though, if it's internal details of the implementation? I'd rather make the primary API method public and let subclasses do whatever they want with it (delegate, reimplement, etc.)

I don't understand why there are 2 methods in the first place. There's a little bit of docs on the `final` one, but the `protected` one is basically undocumented. Seems a bit convoluted.

IMO one single, well-documented, public method would be a better fit.

easyice:
Sorry for the confusion. We move group-varint encoding into DataInput to let subclasses make an optimized implementation.

In the first version, the `readGroupVInts` method is `public`. see [delta](https://github.com/apache/lucene/pull/12841/commits/e58154e44a62235450b4f9d837e9307f4f5a8619%0ADataInput.java): 


class DataInput
```java
public void readGroupVInts(long[] docs, int limit) throws IOException {
   fallbackReadGroupVInt(...)
}
```

```java
protected void fallbackReadGroupVInt(long[] docs, int offset) throws IOException {
}
```


subclass MemorySegmentIndexInput
```java
@Override
public void readGroupVInts(long[] docs, int limit) throws IOException {
}
  
private void readGroupVInt(long[] docs, int pos) throws IOException {
    if (...) {
      super.fallbackReadGroupVInt(docs, pos);
      return;
    }
    ....
}
```

The problem with this version is that subclasses have some duplicate code with small changes. then we try to move the duplicate code to a util class in the subclasses, but it has a bit performance regression. see [delta](https://github.com/apache/lucene/pull/12841#issuecomment-1850424855)

class MemorySegmentIndexInput
```java
  @Override
  public void readGroupVInts(long[] dst, int limit) throws IOException {
    try {
      GroupVIntUtil.readGroupVInts(..);
    }
  }

```

Finally, we remove the `readGroupVInts` method from subclasses, and make the base class `DataInput` implement as `final`(because nothing to optimize there anymore), the subclasses just override the `protected readGroupVInt()`.  see [delta](https://github.com/apache/lucene/pull/12841#issuecomment-1851098019)

dweiss:
I understand what the intention is here, @easyice , but this is the problem - you can't override that public method anymore and you can't call the protected method in a delegate class. I sort of agree with Robert that code duplication here is not worth the added API complexity. I suggest we remove the final modifier from that method first, then think of maybe dropping the protected method altogether (although this doesn't bother me as much, since it's protected).

easyice:
> I understand what the intention is here, @easyice , but this is the problem - you can't override that public method anymore and you can't call the protected method in a delegate class. I sort of agree with Robert that code duplication here is not worth the added API complexity. I suggest we remove the final modifier from that method first, then think of maybe dropping the protected method altogether (although this doesn't bother me as much, since it's protected).

+1

rmuir:
the current readGroupVInts is also pretty specific to what the postings are doing. And it just calls readVint() and readGroupVint() behind the scenes. Seems like a good candidate to move to PostingsUtil, leaving just the one readGroupVint() method in DataInput?

easyice:
> the current readGroupVInts is also pretty specific to what the postings are doing. And it just calls readVint() and readGroupVint() behind the scenes. Seems like a good candidate to move to PostingsUtil, leaving just the one readGroupVint() method in DataInput?

Actually, I have tried to use group-varint encoding in other places. just don’t see a significant impact on performance at this point.

* Use group-varint encode the positions  https://github.com/apache/lucene/pull/12842
* Cut over HNSW's neighbor lists to group-varint https://github.com/apache/lucene/issues/12871

rmuir:
i dont think we should add methods to DataInput when we are only "trying to use elsewhere". The method is only used from one place: PostingsUtil, lets move it out

dweiss:
Folks, if you'd like to do anything larger to this, please go ahead. I'm with very limited access to the internet this week and I won't be able to track your comments closely or make any code modifications. I'll return next Monday and then can jump back in.

easyice:
Thank you @rmuir, Maybe we can move `readGroupVInts()` into `GroupVIntUtil` to achieve the same goal? I raised a PR #13830

uschindler:
> Thank you @rmuir, Maybe we can move `readGroupVInts()` into `GroupVIntUtil` to achieve the same goal? I raised a PR #13830

This is much better. Let's keep only the inner loop part in DataInput and replace the protected method by a public method called by the static utility reading the whole think.

Anyways: I was a bit against the whole change from beginning because I was not happy to add new methods to DataInput just for a silly optimization. I am also not happy about the horrible bottom-up code just to get MemorySegment access working.

IMHO, we should move minimum Java version of Lucene main branch (and 10.x) to Java 22 and introduce MemorySegment as a first class citizen and remove all abstractions from DataInput. Just make DataInput read from/to MemorySegment (NIOFSDir) or let it provide "views" (MMap) on the internal buffers. Let's move all our DataInput encoding/decoding logic to work on top of MemorySegment.

That's all I have to say. I don't care about those arguments "wuhaaah, but Java 21 is LTS !!!!1!1111!!!!!!!!!!". Don't care, use Java 22 because this is what we need to be ready for future.

Uwe

dweiss:
I've looked at the merged patch - thank you, it's much nicer than what it was before. I'm closing this issue.

dweiss:
I've backported this to branch_10x and asked the RM to allow this to be cherry picked for the subsequent respin of Lucene 10.0.

javanna:
I get confused that this PR was not merged yet it has a milestone set. Should the milestone be set to the other PR that has been merged instead (#13830) ?

javanna:
I think we are also missing a CHANGES entry , @dweiss could you take care of adding that please?

dweiss:
I've added migration and changes entry and cherry picked on branch_10_0, branch_10x and main.

javanna:
Thanks @dweiss !!!

PR #13826, opened 2024-09-25
Expose more parameters in MultiLeafKnnCollector ctor

### LINKED ISSUE #13699: Allow MultiLeafKnnCollector.greediness to be configurable
### Description

In Lucene 9.10, we introduced a new algorithm for sharing state between segments to speed up the vector search process. This [article](https://www.elastic.co/search-labs/blog/multi-graph-vector-search) explains the features quite well, and it looks like the idea of the greediness parameter is that we can tune it to control the latency-recall tradeoff: low value means more relaxed search (higher recall but higher latency, global queue less important) and high value means more restrictive search (lower latency but lower recall, global queue more important).

However the current code doesn't allow the parameter to be tunable: It will always use the [default value of 0.9f](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/knn/MultiLeafKnnCollector.java#L61C10-L61C20). I think allow it to be configurable would allow each application to use a value suited to its specific use case and document/query patterns.

As per where user can configure this parameters, I'm thinking of 2 places:
- Allow it to be configured in `AbstractKnnVectorQuery`, and propagate the value down to `MultiLeafKnnCollector`
- Allow it to be configured only in `TopKnnCollectorManager`, and users would need to override `AbstractKnnVectorQuery.getKnnCollectorManager()` 

We can make a PR, but would like to hear from the community first.

benwtrent:
There is currently no way to configure two parameters for the multi-leaf collector. For expert extensibility, this commit adds another ctor for advance usage:

closes: https://github.com/apache/lucene/issues/13699


lucene/core/src/java/org/apache/lucene/search/knn/MultiLeafKnnCollector.java
```diff
@@ -58,7 +58,32 @@ public final class MultiLeafKnnCollector implements KnnCollector {
...
+   * @param globalSimilarityQueue the global queue of the highest similarities collected so far
+   * @param subCollector the local collector
+   */
+  public MultiLeafKnnCollector(
+      int k,
+      float greediness,
+      int interval,
+      BlockingFloatHeap globalSimilarityQueue,
+      AbstractKnnCollector subCollector) {
+    if (greediness <= 0 || greediness > 1) {
```
dungba88, line 80: I'm wondering why greediness cannot be 0. If it's 0, then the nonCompetitiveQueue size will have the size of k, which means effectively disable the global queue (but someone could want that for maximum recall?). Even if we disallow 0, someone can still work around by passing a very small positive value (0.0...1), then it will be effectively 0 due to the round?

benwtrent: I am fine with allow `0`.

cpoerschke (approved):

mikemccand (approved):
Thanks @benwtrent!  I like this approach ... complex things (invoking the complex ctor and tweaking greediness/interval) are now possible.

dungba88:
Thank you

dungba88 (approved):

javanna:
Does this deserve a changes entry? :)

benwtrent:
@javanna i guess? I am not sure its really interesting. I can add an entry under a "refactoring" header or something.

PR #13830, opened 2024-09-28
Move DataInput.readGroupVInts into GroupVIntUtil

easyice:
### Description

Discussed in #13825


uschindler (approved):
Much better than before as it gets away with the protected method problem.

rmuir (approved):
thanks!

mikemccand (approved):
Thanks @easyice

PR #13832, opened 2024-09-28
Fix bug 'DefaultPassageFormatter.format' method

Seunghan-Jung:
### Description

In the following PR (https://github.com/apache/lucene/pull/13276), I added an option to allow customizing the sorting of Passages when using the UnifiedHighlighter. As a result, it is no longer guaranteed that the startOffset of the next passage will be greater than that of the previous passage. However, I am now modifying the `DefaultPassageFormatter.format()` method, where a conditional statement was written under the assumption that this was guaranteed. You can refer to the test code for examples and better understanding.


lucene/highlighter/src/java/org/apache/lucene/search/uhighlight/DefaultPassageFormatter.java
```diff
@@ -64,7 +64,7 @@ public String format(Passage[] passages, String content) {
     int pos = 0;
     for (Passage passage : passages) {
       // don't add ellipsis if its the first one, or if its connected.
-      if (passage.getStartOffset() > pos && pos > 0) {
+      if (passage.getStartOffset() != pos && pos > 0) {
```
dsmiley, line 67: I ran this in a debugger to understand.  I think it'd be clearer if the code checks were done in the same order as the comment.  And the "first one" I think would now be clearer if checked via `sb.isEmpty` instead of `pos`, requiring reasoning about what `pos` could be in later iterations with out-of-order passages.

Seunghan-Jung:
```suggestion
      if (!sb.isEmpty() && passage.getStartOffset() != pos) {
```

Is this the expression you intended?

dsmiley: Yes!

Seunghan-Jung: I applied it to commit, [db47d0d](https://github.com/apache/lucene/pull/13832/commits/db47d0dd1ddc17cd5fc74ef93a968dcc1f0b31de)

dsmiley (commented):
Just needs a CHANGES.txt entry.  No more 9.x releases so this will be in 10.

Seunghan-Jung:
@dsmiley
I have added the relevant entry to the CHANGES.txt at commit, [2b2e8a8](https://github.com/apache/lucene/pull/13832/commits/2b2e8a8282489e178b4661d56678172c9f1389bb)

lucene/CHANGES.txt
```diff
@@ -270,6 +270,9 @@ Bug Fixes
 * GITHUB#12878: Fix the declared Exceptions of Expression#evaluate() to match those
   of DoubleValues#doubleValue(). (Uwe Schindler)
 
+* GITHUB#13832: Fixed an issue where the DefaultPassageFormatter.format method did not format passages as intended
+  when they were not sorted by startOffset. (Seunghan Jung)
+
```
javanna, line 275: This is misplaced. 10.0.0 is feature frozen. It should go to either 10.0.1 or 10.1.0

javanna:
@dsmiley 10.0 is feature frozen, a vote is ongoing. this does need to be backported to branch_10x though it seems, and the changes entry moved to the right place.

dsmiley:
I fixed CHANGES.txt and backported to branch_10x now.

javanna:
Thanks @dsmiley !

PR #13833, opened 2024-09-29
CHANGES/MIGRATE entries for KnnVectorValues API change

msokolov:


javanna (approved):
LGTM, should we also cover the new methods added e.g. ordToDoc , copy, getEncoding etc.? Do they actually require migration?

msokolov:
ordToDoc and copy were moved from RandomAccessVectorValues; consumers will mostly just need to replace uses of that class with KnnVectorValues. As for getEncoding, yeah since it's new, there's no migration step as such related to that. I suppose it's conceivable that someone may have implemented their own subclass of Byte or FloatVectorValues and will now need to comply with the new contract, but that seems extremely expert and I don't think we'd need to cover it here?

PR #13834, opened 2024-09-30
Remove synchronization from IndexWriter.isClosed

original-brownbear:
Blocking on this one seems unnecessary as the closed flag is volatile. This tends to block for an extended period in its only production use case `StandardDirectoryWriter.isCurrent()` every now and then which seems unnecessary. Either it's true, in which case it'll stay true and there's no need to wait for anything or it's false in which case the else branch goes into `IndexWriter.nrtIsCurrent` which blocks on the same mutex anyway.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -5820,7 +5820,7 @@ public class IndexWriter
     return isCurrent;
   }
 
-  synchronized boolean isClosed() {
+  boolean isClosed() {
     return closed;
   }
 


javanna:
That makes sense to me but I am not very familiar with this area. I did notice that perhaps, if isClosed needs to be synchronized, isOpen (and ensureOpen perhaps) should follow the same approach but it does not.

original-brownbear:
>  if isClosed needs to be synchronized

I don't see how this could ever be needed in theory. The things does a single read from a volatile field in theory? Even if we race with another thread here that sets that field's value, at worst we'd see the closed state earlier (without waiting for the operation that set closed). Since we never go closed -> open that seems like it couldn't lead to anything bad ?

PR #13835, opened 2024-09-30
Make generated archive files reproducible

breskeby:
This should ensure deterministic archive files and fix issues with changing checksums even though the codebase has not changed.


javanna:
Thanks @breskeby ! This should solve an issue we have observed when building Elasticsearch against a lucene snapshot. We have seen that the snapshot name may stay the same while the verification sha changes, which requires us to wipe the gradle cache or gradle complains about the mismatch.

@dweiss are you the build expect nowadays? Would you like to have a look? :)

rmuir:
Looks awesome, almost too good to be true if it is this simple. IMO: it is a great property for the build to have.

dweiss:
Part of this is actually already in gradle-archives.gradle - I think adding this patch there would be better?

breskeby:
> Part of this is actually already in gradle-archives.gradle - I think adding this patch there would be better?

@dweiss updated this accordingly. boiled down to a one line change now

dweiss (approved):

dweiss:
Thank you. I'll merge this into branch_9x and branch_10x.

Issue #13567, opened 2024-07-12
"gradlew clean check" results in internal gradle error "Unable to make progress running work."


dweiss:
### Description

As reported by Chris, running "gradlew clean check" on main results in this nasty (or its variations):
```
Unable to make progress running work. The following items are queued for execution but none of them can be started:
  - Build ':build-infra':
      - Waiting for nodes:
          - :build-infra:clean (state=SHOULD_RUN, dependencies=COMPLETE_AND_SUCCESSFUL, group=task group 0, dependencies=[Resolve mutations for :build-infra:clean (EXECUTED)], has-failed-dependency=false )
          - :build-infra:forbiddenApis (state=SHOULD_RUN, dependencies=NOT_COMPLETE, group=task group 1, dependencies=[:build-infra:forbiddenApisMain (EXECUTED), :build-infra:forbiddenApisTest (SHOULD_RUN)], waiting-for=[:build-infra:forbiddenApisTest (SHOULD_RUN)], has-failed-dependency=false )
      - Nodes ready to start:
          - :build-infra:clean
          - :build-infra:compileTestJava
          - :build-infra:processTestResources
      - Reachable nodes:
          - :build-infra:compileTestJava (state=SHOULD_RUN, dependencies=COMPLETE_AND_SUCCESSFUL, group=task group 1, dependencies=[Resolve mutations for :build-infra:compileTestJava (EXECUTED), destroyer locations for task group 0 (EXECUTED), :build-infra:classes (EXECUTED), :build-infra:compileJava (EXECUTED)], has-failed-dependency=false )
          - :build-infra:processTestResources (state=SHOULD_RUN, dependencies=COMPLETE_AND_SUCCESSFUL, group=task group 1, dependencies=[Resolve mutations for :build-infra:processTestResources (EXECUTED), destroyer locations for task group 0 (EXECUTED)], has-failed-dependency=false )
          - :build-infra:testClasses (state=SHOULD_RUN, dependencies=NOT_COMPLETE, group=task group 1, dependencies=[:build-infra:compileTestJava (SHOULD_RUN), :build-infra:processTestResources (SHOULD_RUN)], waiting-for=[:build-infra:processTestResources (SHOULD_RUN), :build-infra:compileTestJava (SHOULD_RUN)], has-failed-dependency=false )
[...]

PR #13843, opened 2024-10-01
Release wizard to split clean and check calls to separate calls

javanna:
While preparing Lucene 10 RC1, I had an issue running the release script from branch_10_0. It reproduces on branch_10x as well. The `./gradle clean check` command fails with the following gradle error and some huge tasks dependency output:

```
Unable to make progress running work. There are items queued for execution but none of them can be started
```

@ChrisHegarty confirmed that he could reproduce the same issue on his machine as well. I worked around this by splitting the `clean` and `check` into two separate calls, in which case everything works fine. I [pushed](https://github.com/apache/lucene/commit/f411adfe43163fde25122a88164991369ce3797f) that work-around to `branch_10_0` to unblock the release process, and I am opening this to discuss whether this is a change that we should make to `main` and `branch_10_0`, at least until we have figured out what causes the issue and we have a fix.

### DIFF
--- a/dev-tools/scripts/buildAndPushRelease.py
+++ b/dev-tools/scripts/buildAndPushRelease.py
@@ -112,8 +112,10 @@ def prepare(root, version, pause_before_sign, gpg_key_id, gpg_password, gpg_home
   checkDOAPfiles(version)
 
   if not dev_mode:
-    print('  ./gradlew --stacktrace --no-daemon clean check')
-    run('./gradlew --stacktrace --no-daemon clean check')
+    print('  ./gradlew --stacktrace --no-daemon clean')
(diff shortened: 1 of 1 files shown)


rmuir (approved):
There have been known issues with invoking `clean` at the same time as other targets like this. FYI: I do see the same idiom in `releaseWizard.yaml` as well.

ChrisHegarty (approved):
LGTM

javanna:
Thanks for the feedback, then it sounds like it's a thing that may have happened on branch_9x as well in the past? I have never experienced it and I assumed that it must have been caused by some change we made in main, but that's quite hard to track.

dweiss:
This is the cause, @javanna - https://github.com/apache/lucene/issues/13567

Issue #13844, opened 2024-10-01
Nightly benchmarks hanging during HNSW connectComponents


mikemccand:
### Description

I'm not sure if this is a real bug, or something wrong w/ the nightly benchy env ... but the benchmark is hung right now building the deterministic search index, for ~21 hours not making any progress (it prints number of docs indexed periodically ...).  That indexing is single threaded, and the one thread is doing this:

```
"Index #0" #122 [2684589] prio=5 os_prio=0 cpu=73410815.43ms elapsed=74640.64s tid=0x00007ffff05de700 nid=2684589 runnable  [0x00007ff730187000]
   java.lang.Thread.State: RUNNABLE
        at org.apache.lucene.util.LongHeap.downHeap(LongHeap.java:175)
        at org.apache.lucene.util.LongHeap.pop(LongHeap.java:108)
        at org.apache.lucene.util.hnsw.NeighborQueue.pop(NeighborQueue.java:131)
        at org.apache.lucene.util.hnsw.HnswGraphSearcher.searchLevel(HnswGraphSearcher.java:225)
        at org.apache.lucene.util.hnsw.HnswGraphBuilder.connectComponents(HnswGraphBuilder.java:465)
        at org.apache.lucene.util.hnsw.HnswGraphBuilder.connectComponents(HnswGraphBuilder.java:423)
        at org.apache.lucene.util.hnsw.HnswGraphBuilder.finish(HnswGraphBuilder.java:416)
        at org.apache.lucene.util.hnsw.HnswConcurrentMergeBuilder.finish(HnswConcurrentMergeBuilder.java:123)
        at org.apache.lucene.util.hnsw.HnswConcurrentMergeBuilder.build(HnswConcurrentMergeBuilder.java:94)
        at org.apache.lucene.util.hnsw.IncrementalHnswGraphMerger.merge(IncrementalHnswGraphMerger.java:143)
        at org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsWriter.mergeOneField(Lucene99HnswVectorsWriter.java:373)
        at org.apache.lucene.codecs.perfield.PerFieldKnnVectorsFormat$FieldsWriter.mergeOneField(PerFieldKnnVectorsFormat.java:121)
        at org.apache.lucene.codecs.KnnVectorsWriter.merge(KnnVectorsWriter.java:105)
        at org.apache.lucene.index.SegmentMerger.mergeVectorValues(SegmentMerger.java:271)
[...]

PR #13846, opened 2024-10-01
search on proper level when connecting graph components

msokolov:
spinoff from #13844 but almost certainly not the cause; we're searching for close nodes to connect two graph components and searching level 0 even when connecting higher levels > 0. But the nodes we find should still be on the current level because of the filter we also apply, so this is just an efficiency fix


javanna:
Would it make sense to backport to branch_10x? Could you also set the appropriate milestone? Thanks!

msokolov:
thanks for the reminder - backported to 10x now

PR #13850, opened 2024-10-02
Fix bug where off-heap scorer would kick on even for float vectors

benwtrent:
introduced in the major refactor https://github.com/apache/lucene/pull/13779

Off-heap scoring is only present for `byte[]` vectors, and it isn't enough to verify that the vector provider also satisfies the `HasIndexSlice` interface. The vectors need to be byte vectors otherwise, the slice iterations and scoring are completely nonsensical leading to HNSW graph building to run until the heat-death of the universe.


ChrisHegarty (approved):
Thanks for finding and fixing Ben. LGTM

benwtrent:
I am gonna merge and backport to 10_0_0 & 10x. I am not putting a changes entry as this could be seen as a bugfix for an unreleased feature (the vector API rewrite), if folks disagree with this, I will happily put a bugfix item.

PR #13851, opened 2024-10-02
Add test for float vector values in FlatVectorsScorer impls

ChrisHegarty:
This is a test only change that verifies the behaviour when float vector values are passed to our FlatVectorsScorer implementations. This would have caught the bug causing #13844, subsequently fixed by #13850.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/internal/vectorization/TestVectorScorer.java
+++ b/lucene/core/src/test/org/apache/lucene/internal/vectorization/TestVectorScorer.java
@@ -16,6 +16,7 @@
  */
 package org.apache.lucene.internal.vectorization;
 
+import static java.util.Locale.ROOT;
 import static org.apache.lucene.index.VectorSimilarityFunction.COSINE;
 import static org.apache.lucene.index.VectorSimilarityFunction.DOT_PRODUCT;
 import static org.apache.lucene.index.VectorSimilarityFunction.EUCLIDEAN;
@@ -24,6 +25,8 @@ import static org.apache.lucene.index.VectorSimilarityFunction.MAXIMUM_INNER_PRO
 import com.carrotsearch.randomizedtesting.generators.RandomNumbers;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


benwtrent (approved):
I verified that this test passes with the bug fix but fails without it.

lucene/core/src/test/org/apache/lucene/internal/vectorization/TestVectorScorer.java
```diff
@@ -329,12 +333,63 @@ public void testLarge() throws IOException {
...
+
+      try (IndexInput in = dir.openInput(fileName, IOContext.DEFAULT)) {
+        for (int times = 0; times < TIMES; times++) {
+          for (var sim : List.of(COSINE, EUCLIDEAN, DOT_PRODUCT, MAXIMUM_INNER_PRODUCT)) {
+            var vectorValues = floatVectorValues(1, 1, in, sim);
+            assert vectorValues.getEncoding().byteSize == 4;
+
+            var supplier1 = DEFAULT_SCORER.getRandomVectorScorerSupplier(sim, vectorValues);
+            var supplier2 = MEMSEG_SCORER.getRandomVectorScorerSupplier(sim, vectorValues);
+            // these assertion assumes that the supplier and scorer's toString will have float
```
msokolov, line 353: "assertions assume"?

lucene/core/src/test/org/apache/lucene/internal/vectorization/TestVectorScorer.java
```diff
@@ -329,12 +333,63 @@ public void testLarge() throws IOException {
...
+
+            var scorer1 = DEFAULT_SCORER.getRandomVectorScorer(sim, vectorValues, new float[] {1f});
+            var scorer2 = MEMSEG_SCORER.getRandomVectorScorer(sim, vectorValues, new float[] {1f});
+            assertTrue(scorer1.toString().toLowerCase(ROOT).contains("float"));
+            assertTrue(scorer2.toString().toLowerCase(ROOT).contains("float"));
+            expected = scorer1.score(0);
+            assertEquals(scorer2.score(0), expected, DELTA);
+
+            expectThrows(
+                Throwable.class,
```
msokolov, line 370: Is Throwable the most specific we can assert?

lucene/core/src/test/org/apache/lucene/internal/vectorization/TestVectorScorer.java
msokolov: too bad we can't access the underlying vector values to see what its vectorEncoding is, but this checking toString() seems  sort of OK?

PR #13855, opened 2024-10-04
Replace branch_9x references with branch_10x in workflows

ChrisHegarty:
With the move of the main branch to 11, and the upcoming sunsetting of 9x, we need to replace branch_9x references with branch_10x in workflows.


javanna (approved):
This seems correct to me, thanks!

Issue #13856, opened 2024-10-04
Significant performance regression with search after


ChrisHegarty:
When verifying the Lucene 10.0.0 release candidate builds we observe a significant (10x) performance regression with _search after_, when querying on a mostly (but not all) sorted data. We've narrowed down the change to #13221, and observe very "heavy" behaviour on some datasets.

The luceneutil benchmarks don't show the issue that we see in the Elasticsearch benchmarks, as it seems that there is no coverage for _search after_. Also, the issue we see is very dependent on the actual dataset, and the target competitive range based on the _after_ value. 

For reference, the query is similar to:
```
TopFieldCollectorManager manager =
  new TopFieldCollectorManager(
    new Sort(new SortedNumericSortField("geoid", SortField.Type.LONG)),
    10,
    new FieldDoc(Integer.MAX_VALUE, 0.0f, new Long[] {searchAfter}),
    10);
TopDocs topDocs = searcher.search(new MatchAllDocsQuery(), manager);
```
.. with several million docs, mostly indexed with the `geoid` sorted. This kind of mostly-but-not-all sortedness we see in a number of datasets, e.g. the `timestamp` field in http_logs [1], the `geoid` in Genomes [2]. And not all that uncommon. I'll include some graphs to help visualise the impact:

![Screenshot from 2024-10-04 12-25-54](https://github.com/user-attachments/assets/5bc8fd91-55e1-4f9f-a3b9-586f7a9d9382)
![Screenshot from 2024-10-04 12-26-27](https://github.com/user-attachments/assets/f9d4bcfe-ccdf-4539-a3a9-a726f787e104)
![Screenshot from 2024-10-04 12-27-19](https://github.com/user-attachments/assets/10a460bb-d245-4097-be80-e04faf0c244e)

Ultimately, given where we are in the Lucene 10 release cycle, and the complexity of this area of code, the most responsible approach I can think of is to revert the changes in 13221, allow the Lucene 10 RC process to continue without it. While continuing to work on a fix the underlying issue, with a view to resolving it in 10.x. The changes are implementation only, so can be applied in a minor release.

@original-brownbear @iverase have more details on the specifics and working towards a fix. 

relates #13221

[1] https://elasticsearch-benchmarks.elastic.co/#tracks/http-logs/nightly/default/90d
[2] https://elasticsearch-benchmarks.elastic.co/#tracks/geonames/nightly/default/90d

javanna:
[...]

PR #13857, opened 2024-10-04
Revert "Disjunction as CompetitiveIterator for numeric dynamic pruning (#13221)

javanna:
This reverts commit 1ee4f8a1115d1de623f242014681032d87ed2c1e.

We have observed performance regressions that can be linked to #13221. We will need to revise the logic that such change introduced in main and branch_10x. While we do so, I propose that we bake it out of branch_10_0 and we release Lucene 10 without it.

Relates to #13856


ChrisHegarty (approved):
This looks like an accurate anti-delta of the original commit. LGTM.

I hope that we are revisit this area really soon, with a view to fixing the underlying cause and getting the performance benefits back in an upcoming 10.x release.

PR #13562, opened 2024-07-11
Add IntervalsSource for range and regexp queries

mayya-sharipova:
We already have convenient functions for contructing IntervalsSource for wildcard and fuzzy functions. This adds functions for regexp and range as well.

PR #13859, opened 2024-10-04
Allow open-ended ranges in Intervals range

mayya-sharipova:
Currently IntervalsSource.range function closed intervals. 
This will allow open-ended ranges

Relates to #13562

### DIFF
--- a/lucene/queries/src/java/org/apache/lucene/queries/intervals/Intervals.java
+++ b/lucene/queries/src/java/org/apache/lucene/queries/intervals/Intervals.java
@@ -247,8 +247,10 @@ public final class Intervals {
    * Return an {@link IntervalsSource} over the disjunction of all terms that fall within the given
    * range
    *
-   * @param lowerTerm The term text at the lower end of the range
-   * @param upperTerm The term text at the upper end of the range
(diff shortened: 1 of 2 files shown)


romseygeek (commented):
Can we add a note to the `@param` javadocs on `IntervalsSource#range()` to say that `null` is permitted on `lowerTerm` and `upperTerm` and means an open-ended range at that end?

mayya-sharipova:
@romseygeek Thanks for checking, addressed in 2f2df7951fe416a84fdc7f3b15e5c1283b255ac2

javanna:
Kind reminder to backport if needed and set the appropriate milestone :)

javanna:
Do we also need a changes entry? :)

mayya-sharipova:
@javanna Thanks Luca, can we still backport to 9.12? Are we expecting more releases?

benwtrent:
@mayya-sharipova 10x, there is no longer a 9x.

Lucene main is Lucene 11 now

mayya-sharipova:
@benwtrent Thanks Ben. @javanna Will packport to 10x with a change entry

PR #13863, opened 2024-10-06
Reduce allocations in ByteBuffersDataOutput.writeString

original-brownbear:
Saw this in the nightly benchmark indexing run profiling:

There's no need to allocate a byte array when serializing to heap buffers and the string fits the remaining capacity without further bounds checks. If it doesn't fit we could technically do better than the current `writeLongString` and avoid one round of copying by chunking the string but that might not be worth the complexity.
In either case we can calculate the utf8 length up-front. While this costs extra cycles (in the small case) for iterating the string twice it saves creating an oftentimes 3x oversized byte array, a `BytesRef`, field reads from the `BytesRef`, copying from it to the buffer and the associated GC with cleaning it up. Theory and some quick benchmarking suggests this version is likely faster for any string length than the existing code.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/ByteBuffersDataOutput.java
+++ b/lucene/core/src/java/org/apache/lucene/store/ByteBuffersDataOutput.java
@@ -31,7 +31,6 @@ import java.util.function.Consumer;
 import java.util.function.IntFunction;
 import org.apache.lucene.util.Accountable;
 import org.apache.lucene.util.BitUtil;
-import org.apache.lucene.util.BytesRef;
 import org.apache.lucene.util.RamUsageEstimator;
 import org.apache.lucene.util.UnicodeUtil;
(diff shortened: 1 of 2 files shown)


jpountz (commented):
I see, we do this for every string stored field, hence why it shows up in nightly profiles.

Changes to `ByteBuffersDataOutput` make sense to me. Changes to `DataOutput` are a bit awkward, if this is a better approach, then this is what the `BytesRef()` constructor should do. Maybe revert changes to `DataOutput` here and open a new PR for `BytesRef()` if you think that we should do this all the time?

original-brownbear:
Sounds good @jpountz , reverted the `DataOutput` stuff for now :)

jpountz (approved):

original-brownbear:
Thanks Adrien!

PR #13864, opened 2024-10-06
Make DirectMonotonicReader.Meta more compact

original-brownbear:
Even with the all-zero bit-length optimization recently introduced, these `Meta` instances tend to consume a non-trivial amount of heap still. We can do away with almost 40% of their size by removing the offsets array that can be calculated on the fly when creating the readers which translates into tens of MB of heap or more for some ES use cases with high segment + field count.
Also added a singleton `DirectMonotonicReader` instance for the common all-zero-values that we have a `Meta` singleton for already.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/packed/DirectMonotonicReader.java
+++ b/lucene/core/src/java/org/apache/lucene/util/packed/DirectMonotonicReader.java
@@ -43,7 +43,6 @@ public final class DirectMonotonicReader extends LongValues {
     private final long[] mins;
     private final float[] avgs;
     private final byte[] bpvs;
-    private final long[] offsets;
 
     Meta(long numValues, int blockShift) {
       this.blockShift = blockShift;
@@ -55,9 +54,11 @@ public final class DirectMonotonicReader extends LongValues {
       this.mins = new long[this.numBlocks];
       this.avgs = new float[this.numBlocks];
       this.bpvs = new byte[this.numBlocks];
-      this.offsets = new long[this.numBlocks];
     }
   }
+
+  private static final DirectMonotonicReader SINGLE_ZERO_BLOCK_READER =
+      createInstance(Meta.SINGLE_ZERO_BLOCK, null, false);
 
   /**
    * Load metadata from the given {@link IndexInput}.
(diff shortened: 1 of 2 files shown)


jpountz (commented):
I'm not totally comfortable with this change, I like that the number of written padding bytes doesn't affect backward compatibility today, it makes things simpler.

original-brownbear:
That's alright @jpountz , these things are still eating an absurd amount of heap for nothing. How about leaving the format as is and storing the error vs. my calculation in a `byte[]`, saving 7/8 of what this PR saves is pretty cool too and it doesn't add much complexity? (or ... we could go a tiny bit further and pack the array and it'll be free today because it's all zeros? :) wdyt?)

jpountz:
Looking at places where we used `DirectMonotonicReader`, I suspect that the main offender are doc values terms dictionaries, which maintain two `DirectMonotonicReader`s each. However, they only record every 64th and 1024th value respectively, so we could store them as off-heap long[] instead of `DirectMonotonicReader`, and the storage overhead would likely be negligible without the heap overhead of `DirectMonotonicReader.Meta`?

original-brownbear:
> I suspect that the main offender are doc values terms dictionaries

Jup that's it :)

>  and the storage overhead would likely be negligible without the heap overhead of 

The thing is, we use all the arrays in `Meta` on the hot path when reading from `DirectMonotonic` reader, so just moving this thing off-heap means either taking a (potentially considerable looking at the profiling for this thing) performance hit or a rather complicated solution where we have off-heap meta and lazily set up the concrete arrays?
Or are you suggesting only doing this for offsets?

original-brownbear:
Never mind the above, I think I found a neat solution :)

jpountz:
Sorry, I don't feel good about relying on `paddingBitsNeeded` on the read path. I suggest we close this PR, IMO the better fix would be to change the way we store terms dictionaries to rely less on `DirectMonotonicReader`.

original-brownbear:
yea that's cool sorry forgot about this one, we for starters just store the offsets in a more compact form that'll help already. I'll open a PR once I find a little time :)

PR #13874, opened 2024-10-08
Fix 9.12.0 backcompat break (Lucene 9.12.0 cannot read 9.11.x indices written with quantized HNSW, `Lucene99HnswScalarQuantizedVectorsFormat`)

### LINKED ISSUE #13867: Lucene 9.12 fails reading older versions of Lucene99HnswScalarQuantizedVectorsFormat
### Description

Hello,

I am not sure if I am doing something wrong, but after upgrading to 9.12 all of vector indexes have start failing on init.

In https://github.com/apache/lucene/pull/13767 @mikemccand removed the 8 bit quantization. I believe that the PR had another unwanted side effect that prevents Lucene 9.12 from reading some previously created `Lucene99HnswScalarQuantizedVectorsFormat` files.

The PR also changed the default value of the compress constructor parameter of `Lucene99HnswScalarQuantizedVectorsFormat `from [true to false](https://github.com/apache/lucene/pull/13767/files#diff-2c2e9079edea55e44171f704b7e0ab0058d8251c7c5bd07e855e359eae84d795L70).

In the reader `Lucene99ScalarQuantizedVectorsReader.java` there was a previous check that went from
```
 if (fieldEntry.bits <= 4 && fieldEntry.compress) {
      quantizedVectorBytes = ((dimension + 1) >> 1) + Float.BYTES;
    } else {
      // int8 quantized and calculated stored offset.
      quantizedVectorBytes = dimension + Float.BYTES;
    }
```
to
```
    if (fieldEntry.compress) {
      // two dimensions -> one byte
      quantizedVectorBytes = ((dimension + 1) >> 1) + Float.BYTES;
    } else {
      // one dimension -> one byte
      quantizedVectorBytes = dimension + Float.BYTES;
    }
```

Now, previously created indexes had the default bits set to 7 and the compress set to true. This failed the initial condition in the if. But after the PR, the old ones pass the condition, so the wrong `quantizedVectorBytes` is returned.

When trying to load a previously created index I get the following callstack:
```
Caused by: java.lang.IllegalStateException: Quantized vector data length 1320000 not matching size=10000 * (dim=128 + 4) = 680000
	at org.apache.lucene.codecs.lucene99.Lucene99ScalarQuantizedVectorsReader.validateFieldEntry(Lucene99ScalarQuantizedVectorsReader.java:147)
	at org.apache.lucene.codecs.lucene99.Lucene99ScalarQuantizedVectorsReader.readFields(Lucene99ScalarQuantizedVectorsReader.java:120)
	at org.apache.lucene.codecs.lucene99.Lucene99ScalarQuantizedVectorsReader.<init>(Lucene99ScalarQuantizedVectorsReader.java:89)
	at org.apache.lucene.codecs.lucene99.Lucene99ScalarQuantizedVectorsFormat.fieldsReader(Lucene99ScalarQuantizedVectorsFormat.java:160)
	at org.apache.lucene.codecs.lucene99.Lucene99HnswScalarQuantizedVectorsFormat.fieldsReader(Lucene99HnswScalarQuantizedVectorsFormat.java:154)
	at org.neo4j.kernel.api.impl.schema.vector.codec.LuceneKnnScalarQuantizedVectorFormatV2.fieldsReader(LuceneKnnScalarQuantizedVectorFormatV2.java:55)
	at org.apache.lucene.codecs.perfield.PerFieldKnnVectorsFormat$FieldsReader.<init>(PerFieldKnnVectorsFormat.java:222)
	at org.apache.lucene.codecs.perfield.PerFieldKnnVectorsFormat.fieldsReader(PerFieldKnnVectorsFormat.java:79)
	at org.apache.lucene.index.SegmentCoreReaders.<init>(SegmentCoreReaders.java:158)
	at org.apache.lucene.index.SegmentReader.<init>(SegmentReader.java:96)
	at org.apache.lucene.index.StandardDirectoryReader$1.doBody(StandardDirectoryReader.java:94)
	at org.apache.lucene.index.StandardDirectoryReader$1.doBody(StandardDirectoryReader.java:77)
	at org.apache.lucene.index.SegmentInfos$FindSegmentsFile.run(SegmentInfos.java:821)
	at org.apache.lucene.index.StandardDirectoryReader.open(StandardDirectoryReader.java:109)
	at org.apache.lucene.index.StandardDirectoryReader.open(StandardDirectoryReader.java:67)
	at org.apache.lucene.index.DirectoryReader.open(DirectoryReader.java:60)
	at org.apache.lucene.search.SearcherManager.<init>(SearcherManager.java:113)
....

```

Lucene 9.12, JDK 21, MacOS

mikemccand:
This PR:

  1. Fixes a pre-existing BWC testing bug where our `int8_hnsw*.zip` bwc test files failed to actually use the scalar quantization codec ... they were just ordinary `float32` HNSW since they were using Lucene's default codec to write the bwc indices.  This is why no bwc tests failed with my original PR (#13767), sigh.  I also renamed these files and the test class itself to `int7*` since they are actually all 7 bit not 8 bit quantization.
  1. Regenerates the bwc indices for 9.10.0, 9.11.0, and 9.11.1 (the binary zip files).  This entailed checking out each release tag for these releases, backporting the fix above, regenerating the zip files and copying them into this PR.  It was a bit tricky for 9.10.0 ([see details here](https://github.com/apache/lucene/issues/13867#issuecomment-2400927997)).
  1. Fixes the actual bug from #13767 using the suggested fix from @parnmatt (thank you!).  The bug happened because scalar quantized HNSW indices written with 9.11.0 or 9.11.1 claimed `compress=true` (written into the index) yet vector compression is silently `false` unless `bits` is 7 or 8.  This trap is fixed for newly created indices in 9.12.0, but 9.11.x indices bake in the trap.

I confirmed that after the 2nd step above, the bwc test was indeed failing properly (because of the #13867 bug), and then upon applying the fix in the third step above, bwc tests (and all tests) now pass.

Thank you @neio-ionut for discovering and opening this issue and @parnmatt and @benwtrent for digging and proposing the root cause fix.

Closes #13867


ChrisHegarty:
I think that this is a good change. I'm try to determine why a bwc test fails in the CI.

benwtrent:
I think you have to add it as a backcompat index in another file, but we don't do that until 9.12.1 is cut and released! Since 9.12.1 hasn't gone through the release process, according to backcompat stuff, it doesn't exist. 

So, for 9.12, we should add the other back compat index files, but not 9.12.0 until 9.12.1 has started the release process (then they will be added via that process).

@ChrisHegarty @mikemccand

benwtrent (approved):
Thank you for fixing my bwc mess! It was definitely a learning experience and apparently I didn't learn enough!

mikemccand:
> Thank you for fixing my bwc mess! It was definitely a learning experience and apparently I didn't learn enough!

No worries @benwtrent -- it is crazy complicated and trappy!  I wish we had better automation here (check out release tag, build the bwc index, copy zip files to newer release, etc.).  I will open a spinoff issue...

benwtrent:
@mikemccand I can open PRs to the old versions to add the code. I did the same "copy files and run it" dance that you did because I didn't want to commit code to the old branches. But now I see it would have just been way better to do that :/

mikemccand:
> I'm try to determine why a bwc test fails in the CI.

Thanks @ChrisHegarty.

Can you point me to the CI build that is angry?  I assume you set up a 9.12.x CI build somewhere, and maybe applied this PR?  Why don't I see these failures locally when running `./gradlew check`?  Confused...

> I guess that means we need to generate 9.12.0 indices for the hnsw bwc. You are correct @ChrisHegarty .

Ahhh OK I will generate 9.12.0 bwc zip files and add to this PR!  In fact, I should be able to simply copy down from 10.x, except regenerate the `int7_hnsw.9.12.0.zip` with the bug fix from this PR...

mikemccand:
> I can open the PR -- I still have the dev area where I got it working and generated the bwc `.zip` in this PR.

OK I opened https://github.com/apache/lucene/pull/13879 -- this is just for 9.10.x.

ChrisHegarty:
we need to update `dev-tools/scripts/addBackcompatIndexes.py`. 

```diff
diff --git a/dev-tools/scripts/addBackcompatIndexes.py b/dev-tools/scripts/addBackcompatIndexes.py
index 80272ec0f0c..3056c8268d4 100755
--- a/dev-tools/scripts/addBackcompatIndexes.py
+++ b/dev-tools/scripts/addBackcompatIndexes.py
@@ -40,7 +40,7 @@ def create_and_add_index(source, indextype, index_version, current_version, temp
       'cfs': 'index',
       'nocfs': 'index',
       'sorted': 'sorted',
-      'int8_hnsw': 'int8_hnsw',
+      'int7_hnsw': 'int7_hnsw',
       'moreterms': 'moreterms',
       'dvupdates': 'dvupdates',
       'emptyIndex': 'empty'
@@ -61,7 +61,7 @@ def create_and_add_index(source, indextype, index_version, current_version, temp
     'cfs': 'testCreateCFS',
     'nocfs': 'testCreateNoCFS',
     'sorted': 'testCreateSortedIndex',
-    'int8_hnsw': 'testCreateInt8HNSWIndices',
+    'int7_hnsw': 'testCreateInt7HNSWIndices',
     'moreterms': 'testCreateMoreTermsIndex',
     'dvupdates': 'testCreateIndexWithDocValuesUpdates',
     'emptyIndex': 'testCreateEmptyIndex'
@@ -206,7 +206,7 @@ def main():
   current_version = scriptutil.Version.parse(scriptutil.find_current_version())
   create_and_add_index(source, 'cfs', c.version, current_version, c.temp_dir)
   create_and_add_index(source, 'nocfs', c.version, current_version, c.temp_dir)
-  create_and_add_index(source, 'int8_hnsw', c.version, current_version, c.temp_dir)
+  create_and_add_index(source, 'int7_hnsw', c.version, current_version, c.temp_dir)
   should_make_sorted =     current_version.is_back_compat_with(c.version) \
                        and (c.version.major > 6 or (c.version.major == 6 and c.version.minor >= 2))
```

ChrisHegarty:
if u merge in `branch_9_12`, then `dev-tools/scripts/addBackcompatIndexes.py 9.12.0` should be able to generate the compat indices.

ChrisHegarty:
Argh!! this is very odd.  We cannot just update the addBackcompatIndexes.py script. Since it downloads the 9.12.0 version and tries to run gradle on it!! and of course `testCreateInt7HNSWIndices` does not exist there!! :-(

mikemccand:
> I wonder if the simplest thing here is to just avoid the renaming (from int8 to int7) in 9.x. Otherwise, I don't see how to get `addBackcompatIndexes.py` to work - since 9.12.0 source bundled is released with the int8 names.

Argh!  Write once naming!  A one-way door, sigh.

OK, I will revert that rename.

ChrisHegarty:
> Hmm the problem is, we need to regen the 9.12.0 `int8_hnsw` back compat index using the bug fix from this PR (so that it actually has `int7` quantized vectors, not unquantized `float32`)? So I think I must manually (carefully) do this?

Yes. I did think similar too, then fell into the trap of trying to fix the script, and forgot!!!. You're right. The script is basically borked for this, and unfixable in its current form. In which case, then maybe keep the renames, since it'll avoid the trap I feel into.

ChrisHegarty:
> Ack, I'll re-revert the revert. Thanks @ChrisHegarty.

Thanks. And sorry for the bad suggestion.  

So, I think once the 9.12.0 int7 hnsw bwc file is created, then this PR is ready to merge.   I guess initially to `branch_9_12`, then forward port to `branch_10_0`.   @mikemccand I don't wanna get in ur way, but lemme know if there is anything I can do to help.

mikemccand:
> > Ack, I'll re-revert the revert. Thanks @ChrisHegarty.
> 
> Thanks. And sorry for the bad suggestion.

No worries!

> So, I think once the 9.12.0 int7 hnsw bwc file is created, then this PR is ready to merge. I guess initially to `branch_9_12`, then forward port to `branch_10_0`. @mikemccand I don't wanna get in ur way, but lemme know if there is anything I can do to help.

Yes that is the plan!  I'm running `./gradlew -p lucene/backward-codecs test --tests TestGenerateBwcIndices -Dtests.bwcdir=/l/tmp -Ptests.useSecurityManager=false` to generate the 9.12.0 bwc zip manually, then I'll add/test/push.  I think we are close for 9.12.x.

mikemccand:
Woops, I see my local `branch_9_12` (and where I branched my `fix_back_compat` branch from for this PR) was out of date -- `branch_9_12` had already bumped to 9.12.1 and added bwc indices.  Lemme attempt to merge from its tip and resolve these conflicts ...

mikemccand:
OK phew I think I resolved the conflicts due to stale branch, and tests are passing for me.  I think this is ready?  Let's let GitHub actions chew on it to confirm...

ChrisHegarty:
I was not expecting to see changes in `index.9.12.0-cfs.zip` or `index.9.12.0-no.cfs.zip`. Do you know why these are changed?

mikemccand:
> I was not expecting to see changes in `index.9.12.0-cfs.zip` or `index.9.12.0-no.cfs.zip`. Do you know why these are changed?

Net/net it's because I regen'd all 9.12.0 bwc zip indices ... and each time you gen a Lucene index it'll be different, at least because of the GUID we assign to each segment.

I could revert them -- that would be safer.  Lemme go do that.

ChrisHegarty (approved):
Lgtm

ChrisHegarty:
Oh! Is there an int8_hnsw.9.12.0.zip that can be deleted?

mikemccand:
> Oh! Is there an int8_hnsw.9.12.0.zip that can be deleted?

D'oh!  I'll delete.

ChrisHegarty (approved):

mikemccand:
OK I'll merge shortly once GitHub action is happy!  Thanks everyone, phew!

mikemccand:
> I wish we had better automation here (check out release tag, build the bwc index, copy zip files to newer release, etc.). I will open a spinoff issue...

Actually, this nice `dev-tools/scripts/addBackcompatIndices.py` is exactly the automation I was thinking of!  It just gets tricky in the "who tests the tester" scenario ... when we fix a bug in the bwc index generation and then need to regen from tip of a git branch and not the actual release bits, I guess.

mikemccand:
The `git cherry-pick` of the 9.12.x commit merged cleanly onto 10.0.x branch ... and `gradle check` is happy, so I plan to push that cherry-pick to 10.0.x branch (and 10.x after that) soon.  Any concerns?

mikemccand:
Although, strangely, when I run `./gradlew clean check` I get this very odd (never seen this before) failure:

```
Unable to make progress running work. The following items are queued for execution but none of them can be \
started:
  - Build ':':
      - Waiting for nodes:
          - :lucene:documentation:check (state=SHOULD_RUN, dependencies=NOT_COMPLETE, group=task group 1, d\
ependencies=[:lucene:documentation:checkBrokenLinks (SHOULD_RUN), :lucene:documentation:licenses (EXECUTED)\
, :lucene:documentation:validateSourcePatterns (SHOULD_RUN)], waiting-for=[:lucene:documentation:checkBroke\
nLinks (SHOULD_RUN), :lucene:documentation:validateSourcePatterns (SHOULD_RUN)], has-failed-dependency=fals\
e )
          - :lucene:test-framework:check (state=SHOULD_RUN, dependencies=NOT_COMPLETE, group=task group 1, \
dependencies=[:lucene:test-framework:ecjLint (SHOULD_RUN), :lucene:test-framework:forbiddenApis (SHOULD_RUN\
), :lucene:test-framework:generateEmojiTokenizationTestChecksumCheck (EXECUTED), :lucene:test-framework:gen\
erateJavaUnicodeWordBreakTestChecksumCheck (EXECUTED), :lucene:test-framework:licenses (SHOULD_RUN), :lucen\
e:test-framework:spotlessCheck (SHOULD_RUN), :lucene:test-framework:test (SHOULD_RUN), :lucene:test-framewo\
rk:validateSourcePatterns (SHOULD_RUN)], waiting-for=[:lucene:test-framework:spotlessCheck (SHOULD_RUN), :l\
ucene:test-framework:test (SHOULD_RUN), :lucene:test-framework:ecjLint (SHOULD_RUN), :lucene:test-framework\
:forbiddenApis (SHOULD_RUN), :lucene:test-framework:validateSourcePatterns (SHOULD_RUN), :lucene:test-frame\
work:licenses (SHOULD_RUN)], has-failed-dependency=false )
```

The gradle build then fails having tons of thinsg to do but none of them actionable/schedulable?  Weird.

if I separately run `./gradlew clean` and then `./gradlew check` then it's fine ... weird.  I'm using Java 22.

mikemccand:
OK I pushed 10.0.x fix ... working on 10.x now.  Gradle sees the same weird deadlock when I try to `clean` and `check` in one go.  So weird.

mikemccand:
I cherry-pick'd to 10.x as well.  Let's see if CI builds are happy ... if so, I think this is done!

ChrisHegarty:
> if I separately run `./gradlew clean` and then `./gradlew check` then it's fine ... weird. 

We’ve seen similar recently, reported in #13567. Likely a grade bug, which we’ve worked around as you have, by running clean separately.

javanna:
Heya, I am working on generating the backwards indices after releasing Lucene 10, and here are my observations:

- I think that we need to forward port this change to main as well?
- We need to update the tooling to generate the backwards indices accordingly (python code and java code are no longer in sync)

I am happy to work on this, I think I understand what needs to happen but I wanted to confirm first. Perhaps the forward port is just a cherry-pick to main, hopefully? And once the branches are aligned, I can adjust the tooling and test them direclty as I generate the needed bwc indices for 10.0.0.

javanna:
I did find above a diff that is exactly the same as the PR I opened to update AddBackcompatindices :)  (#13911) . I also opened a PR against main to cherry-pick this change and make the necessary adjustments: #13912

ChrisHegarty:
> * I think that we need to forward port this change to main as well?

Yes, it should be forward ported.

> * We need to update the tooling to generate the backwards indices accordingly (python code and java code are no longer in sync)

Yes, these should be aligned.

Thanks,
-Chris.

PR #13875, opened 2024-10-09
Disable CFS in TestDefaultCodecParallelizesIO.

### LINKED ISSUE #13854: TestDefaultCodecParallelizesIO.testStoredFields fails
### Description

Fails on main & 10x.

```
   >     java.lang.AssertionError: newCount [122] prevCount [38] docs [20]
   >         at __randomizedtesting.SeedInfo.seed([53F3DEA4BDBAD95F:4741334216B03EA]:0)
   >         at org.junit.Assert.fail(Assert.java:89)
   >         at org.junit.Assert.assertTrue(Assert.java:42)
```

I dug a bit into the failure and the count thrashing is being caused by a couple of things.

First, prefetch cache for stored fields only allows 15 blocks. If each doc is in a different block, welp, we are gonna kick something out of cache and obviously this causes more fetches. 

Additionally, I found that the `SerializedIOCountingIndexInput` does its best to keep from incrementing on multiple pre-fetches in a row. Welp, that doesn't work for this test and seed. From what I can tell, `SerializedIOCountingIndexInput` protects incrementing with multiple prefetches here:
```
      if (firstPage >= currentPage && lastPage <= readAheadUpto) {
        // seeking within the current (or next page if ReadAdvice.NORMAL) doesn't increment the
        // counter
      } else if (pendingFetch.get() == false) {
        // If multiple prefetch calls are performed without a readXXX() call in-between, count a
        // single increment as these I/O requests can be performed in parallel.
        counter.increment();
        pendingPages.clear();
        pendingFetch.set(true);
      }
```

However, for every stored fields prefetch, we also call
```
public void prefetch(int docID) throws IOException {
    final long blockID = indexReader.getBlockID(docID);
...
}
```
`indexReader.getBlockID(docID)` might actually read some bytes, thus flipping the `pendingFetch` and causing more increments to occur.

I honestly don't know if this is because we are actually reading bytes that weren't previously prefetched, or if we are reading bytes from another page altogether, etc. etc. etc. 

This testing infrastructure is tricky to understand, and I am not entirely sure its actually testing what it wants to test.

```
./gradlew :lucene:core:test --tests "org.apache.lucene.index.TestDefaultCodecParallelizesIO.testStoredFields" -Ptests.jvms=12 -Ptests.jvmargs= -Ptests.seed=53F3DEA4BDBAD95F -Ptests.gui=true -Ptests.file.encoding=ISO-8859-1 -Ptests.vectorsize=128 -Ptests.forceintegervectors=true
```

jpountz:
`SerialIODirectory` doesn't count reads to files that are open with `ReadAdvice#RANDOM_PRELOAD` as these files are expected to be loaded in memory. Unfortunately, we cannot detect such files on compound segments, so this test now disables compound segments.

Closes #13854

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/index/TestDefaultCodecParallelizesIO.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestDefaultCodecParallelizesIO.java
@@ -40,7 +40,14 @@ public class TestDefaultCodecParallelizesIO extends LuceneTestCase {
     Directory bbDir = new ByteBuffersDirectory();
     try (LineFileDocs docs = new LineFileDocs(random());
         IndexWriter w =
-            new IndexWriter(bbDir, new IndexWriterConfig().setCodec(TestUtil.getDefaultCodec()))) {
+            new IndexWriter(
+                bbDir,
+                new IndexWriterConfig()
+                    // Disable CFS, this test needs to know about files that are open with the
+                    // RANDOM_PRELOAD advice, which CFS doesn't allow us to detect.
+                    .setUseCompoundFile(false)
+                    .setMergePolicy(newLogMergePolicy(false))
+                    .setCodec(TestUtil.getDefaultCodec()))) {
       final int numDocs = atLeast(10_000);
       for (int d = 0; d < numDocs; ++d) {
         Document doc = docs.nextDoc();


javanna (approved):
LGTM

benwtrent (approved):
❤️

PR #13876, opened 2024-10-09
Remove broken .toArray from IntObjectHashMap entirely

### LINKED ISSUE #13761: IntObjectHashMap.values().toArray() method throws ClassCastException
### Description

The ClassCastEx is thrown when the `IntObjectHashMap.values().toArray()` method is called with a type other than `Object`, and of course the same exception occurs with `CharObjectHashMap` and `LongObjectHashMap`. The issue can be reproduced with the following test.
```
@Test
  public void testStringMapValues() {
    IntObjectHashMap<String> map = new IntObjectHashMap<>();
    map.put(1, "a");
    map.put(2, "b");
    map.put(3, "c");
    String[] values = map.values().toArray();
  }
```

There are several options to fix the issue.
1. change the return value to `Object[]`;
2. mimic the `ArrayList.toArray(T[])` method to pass a `T[]` parameter;
3. replace the `toArray` method with a `asList` method, and return a `List<T>`.

IMO, either option 2 or 3 would be acceptable. Any thought?

_No response_

dweiss:
This method is only used in tests. I moved the toList collection utility to the test class to minimize surface API. 
Fixes #13761

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/internal/hppc/TestIntObjectHashMap.java
+++ b/lucene/core/src/test/org/apache/lucene/internal/hppc/TestIntObjectHashMap.java
@@ -18,12 +18,17 @@
 package org.apache.lucene.internal.hppc;
 
 import com.carrotsearch.randomizedtesting.RandomizedTest;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Random;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.lucene.tests.util.LuceneTestCase;
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.Assert;
 import org.junit.Test;
 
 /**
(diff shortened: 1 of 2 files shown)


lucene/core/src/test/org/apache/lucene/internal/hppc/TestIntObjectHashMap.java
```diff
@@ -66,10 +69,8 @@ private static void assertSortedListEquals(int[] array, int... elements) {
   }
 
   /** Check if the array's content is identical to a given sequence of elements. */
-  private static void assertSortedListEquals(Object[] array, Object... elements) {
-    assertEquals(elements.length, array.length);
-    Arrays.sort(array);
-    assertArrayEquals(elements, array);
+  private static void assertSortedListEquals(List<Object> array, Object... elements) {
+    Assert.assertEquals(array.stream().sorted().toList(), Arrays.asList(elements));
```
bruno-roustant, line 73: Should the params order be swapped to have the expected first?

bruno-roustant (approved):

lucene/core/src/test/org/apache/lucene/internal/hppc/TestIntObjectHashMap.java
dweiss: Absolutely - thank you for spotting this.

PR #13877, opened 2024-10-09
Avoid performance regression by constructing lazily the PointTree in NumericComparator (#13498)

iverase:
Forward port of https://github.com/apache/lucene/pull/13498 to address performance issue after reverting https://github.com/apache/lucene/pull/13857.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -218,6 +218,8 @@ Bug Fixes
 * GITHUB#12878: Fix the declared Exceptions of Expression#evaluate() to match those
   of DoubleValues#doubleValue(). (Uwe Schindler)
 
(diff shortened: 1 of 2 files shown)


javanna (approved):
thanks @iverase !

javanna:
Perhaps a changes entry would be good to add?

ChrisHegarty (approved):
LGTM

iverase:
I don't think we need to add an entry in changes as this is part of 9.11.1 and 9.12.0?

iverase:
I add an entry in 77b8098

javanna:
I forward ported this to 10x and main as well, now that #13221 has been fully reverted.

PR #13878, opened 2024-10-09
Reuse Impacts instances across invocations.

jpountz:
Our `ImpactsEnum` currently return a new object on every call to `getImpacts()`, which shows up in nightly profiles.


original-brownbear (approved):
LGTM! :)

small side note: I wonder if we could also make `Collections.singletonList(new Impact(Integer.MAX_VALUE, 1L))` a constant? (not sure how much it saves but reduces method size a bit) The caller is not expected to mutate the list or impact instances ever right?

jpountz:
You are right, I pushed a change.

PR #13879, opened 2024-10-09
Add tooling back on 9.10.x branch to generate int7_hnsw.9.10.zip bwc index

mikemccand:
Spinoff from #13867 to add the tooling back onto `branch_9_10` in case we ever need to regenerate the `int7_hnsw.9.10.0.zip` bwc index again.


lucene/backward-codecs/src/test/org/apache/lucene/backward_index/TestGenerateBwcIndices.java
```diff
@@ -82,6 +82,16 @@ public void testCreateSortedIndex() throws IOException {
     sortedTest.createBWCIndex();
   }
 
+  public void testCreateInt7HNSWIndices() throws IOException {
```
ChrisHegarty, line 85: should we add this to the back compat script? And if so then there could be a collision with future versions.

mikemccand: Hmm I think we can just rename `int8_hnsw` -> `int7_hnsw` in `addBackCompatIndexes.py`?  But I would do that change on our latest branches (9.12.x, 10.x, 10.0.x) as a separate follow-on PR?  This PR is just targetting 9.10.x branch in case we ever need to regen 9.10.x bwc indices again (hope not!).

PR #13884, opened 2024-10-10
Remove broken .toArray from Long/CharObjectHashMap entirely

### LINKED ISSUE #13761: IntObjectHashMap.values().toArray() method throws ClassCastException

bugmakerrrrrr:
### Description

Fixes #13761

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -53,6 +53,7 @@ Bug Fixes
 ---------------------
 * GITHUB#13832: Fixed an issue where the DefaultPassageFormatter.format method did not format passages as intended
   when they were not sorted by startOffset. (Seunghan Jung)
+* GITHUB#13884: Remove broken .toArray from Long/CharObjectHashMap entirely. (Pan Guixin)
 
 Build
 ---------------------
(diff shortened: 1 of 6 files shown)


bugmakerrrrrr:
@dweiss would you mind taking a look at this if you get a chance

dweiss:
Sure, thank you. The linter has failed - seems like an unused method. Please correct and perhaps add a CHANGES.txt entry with an attribution to you?

bugmakerrrrrr:
@dweiss Thanks. I've fixed the lint error, PTAL.

dweiss (approved):

dweiss:
Thank you!

PR #13885, opened 2024-10-10
Lazy initialize ForDeltaUtil and ForUtil in Lucene912PostingsReader

original-brownbear:
Lazy initialize these fields. They consume/cause a lot of memory/GC because they are allocated frequently (~7% of all allocations in luceneutil's wikimedia medium run for me). This does not cause any measurable slowdown as far as runtime is concerned and since these are not even needed for all instances (in fact they are rarely used in the queries the benchmark explores) save quite a few CPU cycles for collecting and allocating them.

@jpountz I know we talked about reusing these instances, but as far as I can tell this deals with the problem much more fundamentally, we simply often do not need these?

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
@@ -632,7 +632,7 @@ public final class Lucene912PostingsReader extends PostingsReaderBase {
   final class EverythingEnum extends PostingsEnum {
 
     private ForDeltaUtil forDeltaUtil;
-    final PForUtil pforUtil = new PForUtil(new ForUtil());
+    private PForUtil pforUtil;
 
     private final long[] docBuffer = new long[BLOCK_SIZE + 1];
     private final long[] freqBuffer = new long[BLOCK_SIZE + 1];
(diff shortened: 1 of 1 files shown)


jpountz:
Even if it doesn't show up in benchmarks it's disappointing to have these conditions in hot code paths. Could we instead initialize these objects in reset() if `docFreq >= BLOCK_SIZE`?

jpountz:
If you're looking at this sort of allocation, you may also want to specialize BlockDocsEnum into one class that decodes only doc IDs and another one that decodes docs and freqs. The former wouldn't need the long[] freq array and the `PForUtil` helper.

original-brownbear:
Hmm I tried and failed miserably to find a solution to initialize in `reset()` that I fully understand :) It's quite hard to reason about all the possible paths here for me, but I see your point.
As far as specialization goes, that one I could find a version that didn't show runtime regressions (because callsites + polymorphism).
Maybe do a little cleanup along the lines of https://github.com/apache/lucene/pull/13892 first then get back to this? :)
It's hard to fix this without complicating the code, this is the simplest version I could find but I see your points. I bet if we clean up all the duplication that we can cleanup without runtime overhead, we'll find an easier fix here :) wdyt?
(this one also isn't so urgent, it's very few cycles runtime, mostly a p100 thing for JVMs under GC pressure probably)

jpountz:
I would expect putting something like `if (forUtil == null && termState.docFreq >= BLOCK_SIZE) { /* initialize forUtil */ }` in reset() to work (untested).

jpountz (approved):
This looks good, you should be able to save the PForUtil instantiation on EverythingEnum by checking totalTermFreq against BLOCK_SIZE.

original-brownbear:
Thanks Adrien!

jpountz:
It's almost certainly this change that sped up TermTitleSort on October 12th, since TermTitleSort is the task that creates the most `PostingsEnum` objects. I pushed an annotation.

PR #13887, opened 2024-10-11
Parse escaped brackets and spaces in range queries

benchaplin:
### Description

[This issue](https://github.com/apache/lucene/issues/13234) raises a question about the QueryParser's ability to handle escaped brackets in a range query's terms.

```
queryParser.parse( "[ a\\[i\\] TO b\\[i\\] ]" );

/* 
org.apache.lucene.queryparser.classic.ParseException: Cannot parse '[a\[i\] TO b\[i\]]': Encountered " "]" "] "" at line 1, column 6.
Was expecting:
    "TO" ...
 */
```

I discovered a workaround highlighted in [a previous PR](https://github.com/apache/lucene/pull/13323). However, I think that escaped special characters (notably: closing brackets and spaces) should be allowed in a range term, so I've updated the JavaCC token definition.

### Backwards Compatibility

There are two JavaCC changes:  
1. **Addition of `| "\\" ~[]`** - by nature of the `|` this should not fail to parse anything that was parsed before. 
2. **Addition of `"\\"` in the negation set: `~[ "\\", " ", "]", "}" ]`** - if the `"\\"` is followed by another character, it will be parsed by change 1. If it's followed by EOF, it would have thrown a `ParseException` for "Term can not end with escape character."

By this logic, nothing that was parsed before the changes will fail to parse after the changes. The one break in backwards compatibility is the message of the `ParseException` in some cases, for example:

`queryParser.parse("[\\ TO abc]");` 
- Before: `org.apache.lucene.queryparser.classic.ParseException: Cannot parse '[\ TO abc]': Term can not end with escape character.`
- After: `org.apache.lucene.queryparser.classic.ParseException: Cannot parse '[\ TO abc]': Encountered " <RANGE_GOOP> "\\] "" at line 1, column 6.
Was expecting:
    "TO" ...`
  - This is to be expected as escaped spaces are now allowed, so the first term parsed is `" TO"`.

---

### Question: other QueryParsers

In previous changes to QueryParser.jj/StandardSyntaxParser.jj, I noticed authors also changed Solr's [QueryParser.jj](https://github.com/apache/solr/blob/main/solr/core/src/java/org/apache/solr/parser/QueryParser.jj). I imagine I'd want to do the same here, but wanted to check with maintainers first.

### DIFF
--- a/lucene/queryparser/src/generated/checksums/javaccParserClassic.json
+++ b/lucene/queryparser/src/generated/checksums/javaccParserClassic.json
@@ -1,9 +1,9 @@
 {
     "lucene/queryparser/src/java/org/apache/lucene/queryparser/classic/ParseException.java": "7a8a8fd5b2ea78f9a17f54cbae8b0e4496e8988e",
     "lucene/queryparser/src/java/org/apache/lucene/queryparser/classic/QueryParser.java": "a2b7d21092d21cbac290cb1ddde5ac161824fb83",
-    "lucene/queryparser/src/java/org/apache/lucene/queryparser/classic/QueryParser.jj": "ed9f248e1a48cadeeab8f0a79e77e986e34ff721",
+    "lucene/queryparser/src/java/org/apache/lucene/queryparser/classic/QueryParser.jj": "64bc53b09c0665afd281f1d85a2db802e5f75266",
     "lucene/queryparser/src/java/org/apache/lucene/queryparser/classic/QueryParserConstants.java": "e59a3fd38b66a3d56779c55955c1e014225a1f50",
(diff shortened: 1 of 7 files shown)


benchaplin:
Sorry for the failed builds, didn't realize I had an incorrect checksum.

benchaplin:
@dweiss you mentioned in my previous PR that I should do some randomized testing. I did, which helped me find the "Addition of "\\" in the negation set" requirement. However I just translated the key test cases to unit tests for this PR.

Did you want me to commit this kind of randomized test? If so, I was thinking I'd have to introduce a snapshotting mechanism to record the output of oldParser(testTerm) as a baseline and compare it against newParser(testTerm) to catch any unintended changes.

PR #13845, opened 2024-10-01
PR 13757 follow-up: add missing with-discountOverlaps Similarity constructor variants, CHANGES.txt entries

cpoerschke:
#13757 follow-up

PR #13891, opened 2024-10-11
PR 13757 follow-up: add missing with-discountOverlaps Similarity constructor variants (#13845)

cpoerschke:
(cherry picked from commit dab731175c86cc25741f37845136d38a69a9d165)

Resolved Conflicts:
	lucene/CHANGES.txt

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -52,9 +52,6 @@ API Changes
   the entire segment should be scored. Subclasses that override the method should instead override its replacement.
   (Luca Cavanna)
 
-* GITHUB#13757: For similarities, provide default computeNorm implementation and remove remaining discountOverlaps setters.
-  (Christine Poerschke, Adrien Grand, Robert Muir)
-
 New Features
 ---------------------
 
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -9,6 +9,11 @@ Bug Fixes
 ---------------------
 (No changes)
 
+API Changes
+---------------------
+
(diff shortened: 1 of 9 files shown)


lucene/CHANGES.txt
```diff
@@ -47,6 +52,9 @@ API Changes
   the entire segment should be scored. Subclasses that override the method should instead override its replacement.
   (Luca Cavanna)
 
+* GITHUB#13757: For similarities, provide default computeNorm implementation and remove remaining discountOverlaps setters.
+  (Christine Poerschke, Adrien Grand, Robert Muir)
```
javanna, line 56: I am afraid is is now too late to modify the changelog for 9.12, as it's already been released?

javanna:
@cpoerschke would you mind getting this in, but backing out the changes to 9.12.0 section of the change log?

lucene/CHANGES.txt
cpoerschke: Yes, https://lucene.apache.org/core/9_12_0/changes/Changes.html cannot be modified but my understanding is that the addition of the missing entry here would be included in https://lucene.apache.org/core/9_12_1/changes/Changes.html in future then, with the 9.12.1 file including 9.12.0 and earlier content also. Whether or not that's worthwhile, hmm, subjective I guess.

lucene/CHANGES.txt
```diff
@@ -47,6 +52,9 @@ API Changes
...
+
```
cpoerschke, line 57:
```suggestion
```

lucene/CHANGES.txt
cpoerschke: https://github.com/apache/lucene/pull/13891/commits/67b95283137fa6169e4898a6cec5655395cf5213 to remove it.

ChrisHegarty (approved):

PR #13892, opened 2024-10-11
Dry up both the two ImpactsEnum Implementation ins Lucene912PostingsReader

original-brownbear:
These two share a lot of code, in particular the impacts implementation is 100% identical. We can save a lot of code and potentially some cycles for method invocations by drying things up. The changes are just mechanical field movements with the following exceptions:

1. One of the two implementations was using a bytes ref builder, one a bytes ref for holding the serialized impacts. The bytesref variant is faster so I used that for both when extracting.
2. Some simple arithmetic simplifications around the levels that should be obvious.
3. Removed the the logic for an index without positions in `BlockImpactsPostingsEnum`, that was dead code, we only set this thing up if there's positions.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
@@ -36,18 +36,9 @@ import org.apache.lucene.codecs.BlockTermState;
 import org.apache.lucene.codecs.CodecUtil;
 import org.apache.lucene.codecs.PostingsReaderBase;
 import org.apache.lucene.codecs.lucene912.Lucene912PostingsFormat.IntBlockTermState;
-import org.apache.lucene.index.FieldInfo;
-import org.apache.lucene.index.Impact;
-import org.apache.lucene.index.Impacts;
-import org.apache.lucene.index.ImpactsEnum;
-import org.apache.lucene.index.IndexFileNames;
-import org.apache.lucene.index.IndexOptions;
-import org.apache.lucene.index.PostingsEnum;
-import org.apache.lucene.index.SegmentReadState;
-import org.apache.lucene.index.SlowImpactsEnum;
+import org.apache.lucene.index.*;
 import org.apache.lucene.internal.vectorization.PostingDecodingUtil;
 import org.apache.lucene.internal.vectorization.VectorizationProvider;
-import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.store.ByteArrayDataInput;
 import org.apache.lucene.store.ChecksumIndexInput;
 import org.apache.lucene.store.DataInput;
@@ -56,7 +47,6 @@ import org.apache.lucene.store.ReadAdvice;
 import org.apache.lucene.util.ArrayUtil;
 import org.apache.lucene.util.BitUtil;
 import org.apache.lucene.util.BytesRef;
-import org.apache.lucene.util.BytesRefBuilder;
 import org.apache.lucene.util.IOUtils;
 
 /**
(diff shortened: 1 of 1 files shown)


lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -1316,8 +1257,88 @@ public BytesRef getPayload() {
...
+          @Override
+          public List<Impact> getImpacts(int level) {
+            if (level == 0 && level0LastDocID != NO_MORE_DOCS) {
+              return readImpacts(level0SerializedImpacts, level0Impacts);
+            }
+            if (level == 1 && level1LastDocID != NO_MORE_DOCS) {
+              return readImpacts(level1SerializedImpacts, level1Impacts);
+            }
+            return DUMMY_IMPACTS;
+          }
```
jpountz, line 1301: I see you refactored this compared with how it's implemented in main. It looks simpler this way, and I believe it behaves the same since `level0LastDocID` being equal to NO_MORE_DOCS should imply `level1LastDocID` being equal to NO_MORE_DOCS too. Have you run luceneutil to check if we indeed return the same impacts, so performance shouldn't be affected?

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -1326,15 +1347,12 @@ private void refillDocs() throws IOException {
 
       if (left >= BLOCK_SIZE) {
         forDeltaUtil.decodeAndPrefixSum(docInUtil, prevDocID, docBuffer);
-
-        if (indexHasFreq) {
-          freqFP = docIn.getFilePointer();
-          PForUtil.skip(docIn);
-        }
```
jpountz, line 1333: Thanks for simplifying.

jpountz (commented):
It's a bit hard to review, since this PR combines code sharing into a common parent class and some unrelated refactorings, but it looks good at first glance.

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
original-brownbear: Jup ran luceneutil yesterday and and this version of things actually is slightly faster even :) (I guess because we have one fewer implementations of the impacts or some other minor effect from reducing codesize.

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -36,18 +36,9 @@
...
-import org.apache.lucene.index.FieldInfo;
-import org.apache.lucene.index.Impact;
-import org.apache.lucene.index.Impacts;
-import org.apache.lucene.index.ImpactsEnum;
-import org.apache.lucene.index.IndexFileNames;
-import org.apache.lucene.index.IndexOptions;
-import org.apache.lucene.index.PostingsEnum;
-import org.apache.lucene.index.SegmentReadState;
-import org.apache.lucene.index.SlowImpactsEnum;
+import org.apache.lucene.index.*;
```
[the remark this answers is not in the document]
original-brownbear: Ah right, I thought we're enforcing that here too but seems not :)

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -1203,70 +1175,53 @@ public long cost() {
...
+
+    private BlockImpactsEnum(IntBlockTermState termState) throws IOException {
+      this.docFreq = termState.docFreq;
       if (docFreq > 1) {
-        docIn = Lucene912PostingsReader.this.docIn.clone();
-        docInUtil = VECTORIZATION_PROVIDER.newPostingDecodingUtil(docIn);
+        this.docIn = Lucene912PostingsReader.this.docIn.clone();
+        this.docInUtil = VECTORIZATION_PROVIDER.newPostingDecodingUtil(docIn);
         prefetchPostings(docIn, termState);
       } else {
```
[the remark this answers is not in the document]
original-brownbear: Right, I guess it became more visible with this change that that's the case :)

original-brownbear:
All addressed Adrien or would you like me to split this up a little? :)

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -1326,8 +1270,88 @@ public BytesRef getPayload() {
...
+            if (level0LastDocID != NO_MORE_DOCS) {
+              numLevels++;
+            }
+            if (level1LastDocID != NO_MORE_DOCS) {
+              numLevels++;
+            }
+            if (numLevels == 0) {
+              numLevels++;
+            }
+            return numLevels;
```
jpountz, line 1294: Nit: this method could be simplified to `return level1LastDocID != NO_MORE_DOCS ? 2 : 1;`

lucene/core/src/java/org/apache/lucene/codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -1326,8 +1270,88 @@ public BytesRef getPayload() {
...
+            }
+            return level == 1 ? level1LastDocID : NO_MORE_DOCS;
+          }
+
+          @Override
+          public List<Impact> getImpacts(int level) {
+            if (level == 0 && level0LastDocID != NO_MORE_DOCS) {
+              return readImpacts(level0SerializedImpacts, level0Impacts);
+            }
+            if (level == 1 && level1LastDocID != NO_MORE_DOCS) {
```
jpountz, line 1310: Nit: The second condition is not necessary I believe (ie. it could be just `if (level == 1)`) as we only ever return 2 levels if `level1LastDocID != NO_MORE_DOCS`.

jpountz (approved):
No, this is fine. LGTM.

original-brownbear:
Thanks Adrien!

PR #13893, opened 2024-10-11
Simplify leaf slice calculation

original-brownbear:
No need to go through the indirection of 2 wrapped functions, just put the logic in plain methods. Also, we can just outright set the field if there's no executor.


javanna (approved):
LGTM

PR #13894, opened 2024-10-11
Use RandomAccessInput instead of seeking in Lucene90DocValuesProducer

original-brownbear:
We use manual seeking and reading instead of random access in this class. That seems just unnecessarily complicated and potentially slower due to the need for updating pointer field in the index input.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
@@ -792,7 +792,7 @@ final class Lucene90DocValuesProducer extends DocValuesProducer {
       return DocValues.emptyBinary();
     }
 
-    final IndexInput bytesSlice = data.slice("fixed-binary", entry.dataOffset, entry.dataLength);
+    final RandomAccessInput bytesSlice = data.randomAccessSlice(entry.dataOffset, entry.dataLength);
     // Prefetch the first page of data. Following pages are expected to get prefetched through
     // read-ahead.
     if (bytesSlice.length() > 0) {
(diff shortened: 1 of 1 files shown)


jpountz (approved):
Good simplification.

For history, `RandomAccessInput` used to not support reading byte[] because `ByteBuffer` didn't support absolute byte[] reads before Java 13 (https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/ByteBuffer.html#get(int,byte%5B%5D,int,int)). We added it when we bumped the minimum required Java version above 13, but there may be other places like this one that still need moving to `RandomAccessInput`.

original-brownbear:
Thanks Adrien!

jpountz:
Looks like there was a slowdown on some taxo facets tasks (which use binary doc values to store the taxonomy). `OrHighMedDayTaxoFacets`, `AndHighHighDayTaxoFacets` and `MedTermDayTaxoFacets` had a slowdown with a low p-value. On the other hand, `BrowseMonthTaxoFacets` went a bit faster (p-value is not so low, but this is the best data point in a long time, which suggests it's not noise). https://benchmarks.mikemccandless.com/2024.10.12.18.20.19.html

We should not revert, but I'll annotate. I'm curious if someone has ideas on the cause. Also cc @stefanvodita @gsmiller since you have interest in facets.

uschindler:
I dont think this is a real slowdown caused by the commit here. It is more caused by the Hotspot optimizer misinterpreting something. We should get the assembly code from the benchmark and compare them to figure out why it gets worse.
Actually the non-random access modifies global state (moves position), so actually optimizer does a harder job, so the slowdown is very strange.

uschindler:
I was a bit afraid that this might have been caused by missing "prefetch()" implementation, but for MemorySegmentIndexInput also the random access slice uses the correct implementation.

My only idea could be that there's a difference between the inlining decisions for *interface* `RandomAccessInput` and the class `IndexInput` (both implemented by MemorySegmentIndexInput, but the first is using `invokeinterface` which the latter is using `invokevirtual` bytecode). This could make a difference here, so we need a dump of which methods were inlined before/after.

original-brownbear:
I think we simply underestimate the variance luceneutil which results in p-values that are too low. See https://github.com/mikemccand/luceneutil/pull/308 for a suggested fix. The QPS have since recovered from the tiny dip seen here .

PR #13895, opened 2024-10-11
Avoid allocating liveDocs when no soft-deletes

dnhatn:
This is a continuation of #13588, where we avoided allocating liveDocs for segments that have the `__soft_deletes` field but no values in it. However, that PR only addressed the reading side. This change fixes the writing scenario with IndexWriter.

Relates #13588

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/PendingSoftDeletes.java
+++ b/lucene/core/src/java/org/apache/lucene/index/PendingSoftDeletes.java
@@ -76,15 +76,14 @@ final class PendingSoftDeletes extends PendingDeletes {
     hardDeletes.onNewReader(reader, info);
     // only re-calculate this if we haven't seen this generation
     if (dvGeneration < info.getDocValuesGen()) {
-      final DocIdSetIterator iterator =
-          FieldExistsQuery.getDocValuesDocIdSetIterator(field, reader);
-      int newDelCount;
-      if (iterator
-          != null) { // nothing is deleted we don't have a soft deletes field in this segment
-        assert info.info.maxDoc() > 0 : "maxDoc is 0";
+      final int newDelCount;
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/index/PendingSoftDeletes.java
```diff
@@ -76,15 +76,14 @@ void onNewReader(CodecReader reader, SegmentCommitInfo info) throws IOException
...
-      final DocIdSetIterator iterator =
-          FieldExistsQuery.getDocValuesDocIdSetIterator(field, reader);
-      int newDelCount;
-      if (iterator
-          != null) { // nothing is deleted we don't have a soft deletes field in this segment
-        assert info.info.maxDoc() > 0 : "maxDoc is 0";
+      final int newDelCount;
+      var iterator = FieldExistsQuery.getDocValuesDocIdSetIterator(field, reader);
+      if (iterator != null && iterator.nextDoc() != DocIdSetIterator.NO_MORE_DOCS) {
+        iterator = FieldExistsQuery.getDocValuesDocIdSetIterator(field, reader);
```
jpountz, line 82: I wonder if we could change `applySoftDeletes` to start with the current doc ID so that we don't need to re-create an iterator here?

jpountz (approved):

lucene/core/src/java/org/apache/lucene/index/PendingSoftDeletes.java
dnhatn: I considered that option as well, but I think creating a new docValues iterator is cheaper than allocating a bitset for liveDocs, so I chose this approach.

dnhatn:
Thanks Adrien!

PR #13899, opened 2024-10-12
Check ahead if we can get the count

### LINKED ISSUE #13890: Check ahead of time if the `count` can be obtained
### Description

Currently, we traverse the `BKD` tree or perform a binary search using `DocValues` first, and then check whether the count can be obtained in the `count()` method of `IndexSortSortedNumericDocValuesRangeQuery`. 

Should we consider providing a mechanism to perform this check beforehand,  avoid unnecessary processing when dealing with a dense range.?   @jpountz

LuXugang:
Check whether the IndexSortSortedNumericDocValuesRangeQuery's count can be obtained before traversing the BKD tree or performing binary search using DocValues.

see https://github.com/apache/lucene/issues/13890

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/IndexSortSortedNumericDocValuesRangeQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/IndexSortSortedNumericDocValuesRangeQuery.java
@@ -186,9 +186,40 @@ public class IndexSortSortedNumericDocValuesRangeQuery extends Query {
       @Override
       public int count(LeafReaderContext context) throws IOException {
         if (context.reader().hasDeletions() == false) {
-          IteratorAndCount itAndCount = getDocIdSetIteratorOrNull(context);
+          IteratorAndCount itAndCount = null;
+          LeafReader reader = context.reader();
+
+          // first use bkd optimization if possible
+          SortedNumericDocValues sortedNumericValues = DocValues.getSortedNumeric(reader, field);
+          NumericDocValues numericValues = DocValues.unwrapSingleton(sortedNumericValues);
+          PointValues pointValues = reader.getPointValues(field);
+          if (pointValues != null && pointValues.getDocCount() == reader.maxDoc()) {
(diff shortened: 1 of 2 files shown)


jpountz (commented):
The logic makes sense to me but it's a bit hard to read, could we avoid touching `getDocIdSetIteratorOrNull` and only have new logic in the `Weight#count` impl?

LuXugang:
> The logic makes sense to me but it's a bit hard to read, could we avoid touching `getDocIdSetIteratorOrNull` and only have new logic in the `Weight#count` impl?

Thank you for your feedback! @jpountz  I really appreciate your suggestion. I’ve made the changes as you recommended

lucene/core/src/java/org/apache/lucene/search/IndexSortSortedNumericDocValuesRangeQuery.java
```diff
@@ -186,10 +186,44 @@ public boolean isCacheable(LeafReaderContext ctx) {
       @Override
       public int count(LeafReaderContext context) throws IOException {
         if (context.reader().hasDeletions() == false) {
-          IteratorAndCount itAndCount = getDocIdSetIteratorOrNull(context);
+          if (lowerValue > upperValue) {
+            return 0;
+          }
```
jpountz, line 191: This could be moved before the check of whether the segment has deletes?

jpountz (approved):
Thanks, the code makes sense and I see that we already have test coverage for counting for our various optimizations.

lucene/core/src/java/org/apache/lucene/search/IndexSortSortedNumericDocValuesRangeQuery.java
LuXugang: if `lowerValue > upperValue` and has deletes, `PointRangeQuery`'s count will reture -1. 
I’ll temporarily maintain the consistency of the two queries for now. Later, we can consider whether to add this optimization to the `PointRangeQuery`. @jpountz

PR #13906, opened 2024-10-14
Avoid slicing memory segments unnecessarily

original-brownbear:
No need to slice when it's a clone that is to be used with random access, we already enforce thread access rules anyway. Also, no point in copying the memory segment instance via noop slicing.

This doesn't really have much of a runtime impact (looking at wikimedium):

```
       BrowseDayOfYearSSDVFacets        6.18      (8.5%)        6.07      (8.7%)   -1.8% ( -17% -   16%) 0.347
                          IntNRQ       21.39      (3.3%)       21.13      (3.8%)   -1.2% (  -8% -    6%) 0.115
            HighIntervalsOrdered        9.76      (4.3%)        9.69      (5.7%)   -0.7% ( -10% -    9%) 0.511
           HighTermDayOfYearSort      156.54      (2.4%)      155.52      (2.9%)   -0.7% (  -5% -    4%) 0.279
                     LowSpanNear       61.87      (2.3%)       61.48      (2.8%)   -0.6% (  -5% -    4%) 0.273
                      TermDTSort       53.99      (1.3%)       53.71      (1.6%)   -0.5% (  -3% -    2%) 0.108
                    HighSpanNear        6.69      (4.4%)        6.66      (4.3%)   -0.4% (  -8% -    8%) 0.697
                       OrHighLow      479.99      (2.1%)      479.02      (2.2%)   -0.2% (  -4% -    4%) 0.670
                          Fuzzy1       62.30      (1.2%)       62.17      (1.3%)   -0.2% (  -2% -    2%) 0.480
               HighTermMonthSort      556.56      (1.3%)      555.47      (1.6%)   -0.2% (  -3% -    2%) 0.546
        AndHighHighDayTaxoFacets        5.60      (4.7%)        5.59      (3.7%)   -0.2% (  -8% -    8%) 0.844
                          Fuzzy2       61.63      (1.3%)       61.54      (1.6%)   -0.1% (  -3% -    2%) 0.651
                         Respell       34.27      (1.5%)       34.22      (1.4%)   -0.1% (  -3% -    2%) 0.702
                 LowSloppyPhrase       19.76      (3.9%)       19.74      (3.5%)   -0.1% (  -7% -    7%) 0.880
             MedIntervalsOrdered       19.12      (2.6%)       19.10      (2.6%)   -0.1% (  -5% -    5%) 0.872
                        Wildcard      115.59      (2.8%)      115.51      (2.8%)   -0.1% (  -5% -    5%) 0.905
                         Prefix3      318.40      (1.4%)      318.82      (1.9%)    0.1% (  -3% -    3%) 0.727
         AndHighMedDayTaxoFacets       52.44      (2.5%)       52.51      (2.3%)    0.1% (  -4% -    5%) 0.794
            BrowseDateSSDVFacets        1.32      (6.3%)        1.33      (4.8%)    0.2% ( -10% -   11%) 0.895
                        PKLookup      157.21      (3.4%)      157.48      (3.8%)    0.2% (  -6% -    7%) 0.829
                      HighPhrase       95.48      (2.3%)       95.69      (2.5%)    0.2% (  -4% -    5%) 0.682
           BrowseMonthTaxoFacets        4.89      (0.4%)        4.90      (0.4%)    0.2% (   0% -    1%) 0.013
                 MedSloppyPhrase       13.62      (2.8%)       13.66      (2.5%)    0.3% (  -4% -    5%) 0.615
                       MedPhrase       23.07      (2.1%)       23.15      (2.4%)    0.3% (  -4% -    4%) 0.495
                HighSloppyPhrase       18.02      (5.3%)       18.09      (4.0%)    0.4% (  -8% -   10%) 0.713
                     MedSpanNear        8.17      (2.9%)        8.20      (2.9%)    0.4% (  -5% -    6%) 0.552
                       LowPhrase      112.30      (2.0%)      112.75      (2.0%)    0.4% (  -3% -    4%) 0.373
                      AndHighLow      607.00      (5.8%)      609.50      (5.4%)    0.4% ( -10% -   12%) 0.740
                         LowTerm      204.43      (2.0%)      205.31      (2.7%)    0.4% (  -4% -    5%) 0.425
               HighTermTitleSort       17.83      (1.4%)       17.91      (1.4%)    0.4% (  -2% -    3%) 0.159
          OrHighMedDayTaxoFacets        3.09      (3.9%)        3.11      (3.8%)    0.5% (  -6% -    8%) 0.599
             LowIntervalsOrdered       20.57      (3.1%)       20.67      (3.9%)    0.5% (  -6% -    7%) 0.550
            HighTermTitleBDVSort        5.50      (3.4%)        5.53      (3.7%)    0.6% (  -6% -    7%) 0.468
                       OrHighMed       89.37      (4.0%)       89.89      (5.4%)    0.6% (  -8% -   10%) 0.583
                      AndHighMed       71.17      (4.3%)       71.61      (5.2%)    0.6% (  -8% -   10%) 0.555
            MedTermDayTaxoFacets       20.66      (4.0%)       20.81      (4.4%)    0.7% (  -7% -    9%) 0.458
                      OrHighHigh       32.44      (6.3%)       32.70      (8.0%)    0.8% ( -12% -   16%) 0.625
           BrowseMonthSSDVFacets        6.29      (7.4%)        6.34     (10.5%)    0.8% ( -15% -   20%) 0.684
                    OrNotHighLow      581.89      (2.8%)      586.83      (2.6%)    0.8% (  -4% -    6%) 0.164
                   OrNotHighHigh      208.03      (5.2%)      209.97      (6.0%)    0.9% (  -9% -   12%) 0.457
                         MedTerm      281.51      (5.0%)      284.37      (6.7%)    1.0% ( -10% -   13%) 0.442
                    OrNotHighMed      231.06      (4.0%)      233.53      (4.1%)    1.1% (  -6% -    9%) 0.240
                    OrHighNotMed      280.55      (4.6%)      283.63      (5.7%)    1.1% (  -8% -   11%) 0.339
                   OrHighNotHigh      141.67      (5.3%)      143.24      (6.5%)    1.1% ( -10% -   13%) 0.403
                     AndHighHigh        8.71      (5.8%)        8.82      (7.4%)    1.2% ( -11% -   15%) 0.419
                    OrHighNotLow      247.33      (5.3%)      250.51      (6.5%)    1.3% (  -9% -   13%) 0.332
                        HighTerm      202.30      (5.5%)      205.03      (7.4%)    1.4% ( -10% -   15%) 0.355
     BrowseRandomLabelSSDVFacets        4.16      (3.6%)        4.27      (7.1%)    2.4% (  -7% -   13%) 0.054
     BrowseRandomLabelTaxoFacets        4.69      (5.5%)        4.97     (22.8%)    5.9% ( -21% -   36%) 0.109
            BrowseDateTaxoFacets        5.44     (12.0%)        5.91     (27.5%)    8.6% ( -27% -   54%) 0.069
       BrowseDayOfYearTaxoFacets        5.54     (12.4%)        6.04     (28.9%)    8.9% ( -28% -   57%) 0.072
```

but it does save about 1% of all allocations overall (effectively `jdk.internal.foreign.MappedMemorySegmentImpl#dup()` disappears from the profile).


lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
```diff
@@ -563,6 +563,13 @@ public final MemorySegmentIndexInput slice(String sliceDescription, long offset,
     return buildSlice(sliceDescription, offset, length);
   }
 
+  public RandomAccessInput randomAccessSlice(long offset, long length) throws IOException {
+    if (offset == 0 && length == this.length) {
+      return this;
```
jpountz, line 568: I'm curious if this code path is properly tested, I would assume most call sites of `randomAccessSlice` to only use a subset of the file/clone.

jpountz (commented):
This looks legal, and looks like it would help terms, postings and vectors (but not norms and doc values, which only use slices, not clones).

One thing that is a bit awkward to me is that it makes clones cheaper than slices, so e.g. refactoring `TermsEnum#postings` to work on a slice that contains just the postings of the current term would be a regression vs. cloning the whole file like today. Not a blocker, it just gets me thinking.

uschindler (changes requested):
Please add a test. The random BaseDirectoryTestCase may not fully check full random slices.

It should also test that the code below works corectly in the single or multi-segment case. The TestMultiMMap testcase is ideal for checking this, as it creates MMapDirs with random chunk sizes.

uschindler:
P.S.: The runtime impact is slow, because most code should not clone IndexInputs all the time, especially not in loops. When it creates a clone before a search starts it should not affect any code.
I am not sure if this really helps much....

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
uschindler: But this code is perfectly fine as random access slices are stateless, so we can simply return the current instance.

original-brownbear:
> One thing that is a bit awkward to me is that it makes clones cheaper than slices, so e.g. refactoring TermsEnum#postings to work on a slice that contains just the postings of the current term would be a regression vs. cloning the whole file like today. Not a blocker, it just gets me thinking.

I had the same thought :D that's what got me here in part. Moving the offset handling into whatever object that uses an `IndexInput` now and replacing that with a stateless shared `RandomAccessInput` should allow for some performance gains here and there and more importantly, should allow for massive heap savings.
It's still quite wild today for ES, I took a heap dump of a http_logs track benchmark node and somehow tens of shards today translate into MBs of `IndexInput` instances for us. I get that there's two sides to this, ES should probably have fewer segments but also still there's no need to redundantly clone (stateful) `IndexInput` when we often also track file pointers in the users of those things anyway in Lucene?

> I am not sure if this really helps much....

See above. But also, even in the Lucene benchmarks it's about 1% of allocations in wikimedium saved for me. Not a massive win, but not entirely irrelevant either :)
Also, this is somewhat helpful as far as making heap-dumps easier to interpret actually. Always easier to find redundancies if object sharing isn't super far down the reference chains :)  

On it, looking looking into a test :)

uschindler:
P.P.S.: Just some background: The code was copied from ByteBufferIndexInput where the clones were necessary. With MemorySegment we no longer create `dup()`s, as MemorySegment is stateless.

So the changes here look fine, but need more tests. Maybe we can refactor the method a bit, so the `clone()` does not go through `buildSlice()` at all.

lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
jpountz: It's correct indeed, but I wonder if we even have a single call site that would benefit from this optimization. If not, then maybe we should simply not optimize this case.

original-brownbear: Right ... sorry this path was only exercised in tests and for some reason my brain decided this was a common prod usage. I reverted this.

original-brownbear:
Thanks @jpountz and @uschindler. I reverted the `randomAccessSlice` thing now, special cased `clone` to not go through slice as asked.
Tests we seem to have for this, doing random slicing and cloning, I take it that request was only related to my previous special casing of the random access slice?

jpountz (approved):
The clone() opto should be well tested. Regarding slicing, I see that `BaseDirectoryTestCase#testSliceOfSlice` creates slices over the whole parent input, for both the outer slice and the inner slice, so I'm good with the existing testing.

uschindler (approved):
Looks good. I agree with @jpountz that testing is fine.

original-brownbear:
Thanks you two!

PR #13909, opened 2024-10-14
Add changelog verifier

### LINKED ISSUE #13898: Create a bot to check if there is a CHANGES entry for new PRs
### Description

We always have to remind ourselves and others to add a CHANGES entry. We can probably write a GitHub action to check for changes to that file on every new PR and leave a comment warning the author if that's missing.

stefanvodita:
### Description

This runs along the checks we already have for PR creation/update and warns us if there is no CHANGES.txt entry.

Closes #13898

### DIFF
--- a/.github/workflows/verify-changelog-and-set-milestone.yml
+++ b/.github/workflows/verify-changelog-and-set-milestone.yml
@@ -23,7 +23,7 @@ jobs:
       - name: ChangeLog Entry Verifier and Milestone Setter
         run: |
           echo "################## STEP 1 ##################"
-          echo "Get all labels from the PR"
+          echo "Get all labels from PR #${{ github.event.number }}"
           mapfile -t labels < <(gh pr view ${{ github.event.number }} --json labels -q '.labels[].name')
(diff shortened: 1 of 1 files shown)


jpountz:
Is it possible to disable this check based on files that are touched (e.g. test-only fixes don't require a changelog entry) and based on the presence of a label (e.g. mechanical refactorings that have no user-facing impact don't need a changelog entry)?

stefanvodita:
Right now this would show up like a separate automated check on PRs, which would fail if the changelog is untouched and we don't have the `Skip-Changelog` label. We can override it in cases where we think a changelog entry is not required. To reduce noise from build failure emails, we can run it only for new PRs, once.

We could also make it smarter about the files that are changed in the PR. Useful example: https://fredrikmeyer.net/2021/05/12/github-actions-changelog.html

My preference would have been to make it a comment instead of a failing check, but I haven't found an action that checks file diffs and comments.

stefanvodita:
This was done better in #14279. Plus, I've accidentally overwritten the branch here. Closing.

PR #13911, opened 2024-10-14
Align TestGenerateBwcIndices.java with AddBackcompatindices.py

javanna:
We updated TestGenerateBwcIndices to create int7 HNSW indices instead of int8 with #13874. The corresponding python code part of the release wizard needs to be updated accordingly.

This should be good for branch_10_0, as well as branch_10x and main. I just used it to generate backwards indices for the 10.0.0 release branch. See https://github.com/apache/lucene/commit/e821a1b4f87b49008368754e2e0d773572e2b1c0 .


jpountz (approved):

ChrisHegarty (approved):

mikemccand:
Thanks @javanna.

PR #13912, opened 2024-10-14
Forward port "Fix 9.12.0 backcompat break" to main

javanna:
This is the forward port of #13874 to main. Additionally, I renamed the int7 backwards indices from 9.x to unsupported, like we had already done with the corresponding int8 indices.


ChrisHegarty (approved):
Looks like an accurate forward port to me. 👍

mikemccand:
Woops, thank you @javanna ... I had the sinking sensation that in all the flurry of cherry-picks to different branches I missed something!

PR #13915, opened 2024-10-15
Early reset scratchBytes in Lucene90BlockTreeTermsWriter.compileIndex.

vsop-479:


mikemccand:
Hello again @vsop-479 -- can you please explain what this change is trying to accomplish?  Thanks.

vsop-479:
Maybe help saving memory because resetting can early happens? I am not sure.

vsop-479:
I will close it, since it is insignificant.

PR #13917, opened 2024-10-15
Skip madvise calls on tiny inner files of compound files.

jpountz:
This commit skips `madvise` calls on inner files of compound files that are smaller than the page size.


mikemccand:
Does our madvise impl otherwise do the right thing for compound files?  I.e. just madvises the right slice of bytes?

mikemccand (approved):
Yeah OK I answered my own question: clearly we do do the right thing (advise on just the slice of bytes).  Thanks @jpountz.

jpountz:
Thanks Mike!

uschindler (commented):
Looks fine.
For small files calling madvise is not really necessary. As it is only one page, the madvise call takes the same amount of time like reading that file in full.
I agree not to take special care if the file starts out of page size.

P.S.: At some point during Lucene 9 I changed the CFS file writer to align all file starts, but we decided to use multiplies of 8 to not force the CPU to read unaliged longs (maximum size of data type). I thought about aligning at page size, but this had several problems:
- Too much wasted space in the CFS file for little bit of improvement
- Possibly incorrect page size if index is used on other platform as it needs to be determined while writing data.

PR #13926, opened 2024-10-17
Include java21 source folders to gradle source sets

javanna:
This makes it so that IDEs automatically add java files under java21 as source files.

### DIFF
--- a/gradle/java/folder-layout.gradle
+++ b/gradle/java/folder-layout.gradle
@@ -19,7 +19,7 @@
 allprojects {
   plugins.withType(JavaPlugin) {
     sourceSets {
-      main.java.srcDirs = ['src/java']
+      main.java.srcDirs = ['src/java', 'src/java21']
       main.resources.srcDirs = ['src/resources']
       test.java.srcDirs = ['src/test']
       test.resources.srcDirs = ['src/test-files']


javanna:
This change does make these file visible as source files in IntelliJ, which is the issue I meant to address. I do need to sort out the build issues that come up though with my change (lack of `--enable-preview` flag)  :)

dweiss (commented):
The main branch should be on Java 21 anyway so we could just move those files to src?

For newer jdk versions, it won't be so easy because the APIs used in those folders won't be supported by the baseline jdk version (and won't compile). I think this should be a conditional setup - if your detected compilation JDK is higher than X then include those folders as source directories for intellij's benefit.  See intellij-idea.gradle for existing hacks that target just the IDE.

javanna:
Thanks @dweiss for chiming in, I clearly don't know what I am doing here :)

> The main branch should be on Java 21 anyway so we could just move those files to src?

Wait, branch_10x as well? We bumped the minimum java version for Lucene 10, hence we can simplify things quite a bit for both main and branch_10x I think, no conditionals needed.

msokolov:
Thank you for addressing this!

javanna:
I will close this then and figure out a different change for branch_10x and main. Thanks for the help!

dweiss:
Thank you, Luka. I think we may want to keep the build infrastructure for mr-jars - just in case it comes in handy in the future.

dweiss:
(but java21 folders can definitely be folded in).

javanna:
> (but java21 folders can definitely be folded in).

I talked with @ChrisHegarty and he kindly explained that that is not the case. it's complicated :) While we are on java 21, these functionalities require the `--enable-preview` flag, and we can't simply copy all under `src/java` just yet.

javanna:
> Thank you for addressing this!

I wish I did! But I know that Chris has been looking at finding some solution around having IDEs import these classes as source sets.

dweiss:
Right... I forgot about that. If so then adding this to intellij will be even more complicated because it'd have to be another module, with a different set of compilation flags. Perhaps you can do it manually - I'm not sure.

javanna:
Yes @dweiss indeed it's complicated. I have tried manually and I did not succeed yet.

dweiss:
It is complicated also because there is some trickery in how Lucene compiles against those preview APIs - we don't use the preview option but instead fool the compiler into thinking these APIs are not in preview-mode (so that they can be used without the preview flag at runtime). It is definitely not something IDEs will easily digest.

I managed to get the compilation working in IntelliJ using IntelliJ compilation mode, followed by manual tweaks. Sorry for lame Windows paths, but maybe it'll be helpful -

1) add the right sources to the main21 module -

image

and redirect its output to a separate folder:

image

2) add custom compiler options to the java21 module:

image

rebuild and the sources in main21 will compile cleanly.

image

Sadly, any time you re-import the project from gradle, this configuration tweak will likely be destroyed. For one-time use it may be an option though.

dweiss:
Also - this basically adds syntax highlighting and suggestions, forget about running tests with these classes - I don't think it'll work from the IDE.

javanna:
Thanks @dweiss ! Those manual steps worked for me as well. Good to know that there is at least a way to get the IDE to do something useful with those files, especially needing to make changes in them.

PR #13927, opened 2024-10-17
Fix StoredFieldsConsumer finish

linfn:
### Description

The `lastDoc` has already invoked `startDocument`, and the next `startDocument` docId should be `lastDoc + 1`.
Additionally, the `lastDoc` will increment automatically within `startDocument`


jpountz:
Ouch, thanks for finding it and fixing it. Can you add a test?

linfn:
> Ouch, thanks for finding it and fixing it. Can you add a test?

Test added. Let me know if anything else is needed.

jpountz (approved):
Thanks, this looks good to me. Can you add an entry to `lucene/CHANGES.txt` under 10.1.0?

linfn:
@jpountz Done. Thanks!

PR #13928, opened 2024-10-18
Change StringHelper from abstract to final.

vsop-479:

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/StringHelper.java
+++ b/lucene/core/src/java/org/apache/lucene/util/StringHelper.java
@@ -28,7 +28,7 @@ import java.util.Properties;
  *
  * @lucene.internal
  */
-public abstract class StringHelper {
+public final class StringHelper {
 
   /**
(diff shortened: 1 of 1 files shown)


jpountz:
I believe that the abstract modifier aimed at preventing instantiation. Can you add a private constructor then?

vsop-479:
> Can you add a private constructor then?

There already is a private constructor in it (line 66).

vsop-479:
I moved the private constructor to the beginning place.

uschindler:
I am not sure if this change is needed at all. There are multiple patterns to make static only classes. The most important one is to add a private ctor. The final or not doesn't really matter.

vsop-479:
Thanks @uschindler — Do you think it is necessary that just remove the abstract modifier?

uschindler:
If you really want to change it, make it final. But it's not necessary.

vsop-479:
> it's not necessary.

I will close it.

PR #13939, opened 2024-10-21
Reduce the compiled size of the collect() method on `TopScoreDocCollector`.

jpountz:
This comes from observations on https://tantivy-search.github.io/bench/ for exhaustive evaluation like `TOP_100_COUNT`. `collect()` is often inlined, but other methods that we'd like to see inlined like `PostingsEnum#nextDoc()` are not always inlined. This PR decreases the compiled size of `collect()` to make more room for other methods to be inlined.

It does so by moving an assertion to `AssertingScorable` and extracting an uncommon code path to a method.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
+++ b/lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java
@@ -70,17 +70,16 @@ public abstract class TopScoreDocCollector extends TopDocsCollector<ScoreDoc> {
         public void collect(int doc) throws IOException {
           float score = scorer.score();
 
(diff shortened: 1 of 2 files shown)


jpountz:
For reference, luceneutil shows no difference.

original-brownbear (approved):
LGTM, I think this makes sense. The benchmarks tend to show `collectCompetitiveHit` as a cold path and in fact on one of my benchmarking machines (faster Threadripper), I get  a slight reproducible win for `TermDTSort` which afaik is one of the tasks that exercise this logic the hardest. No such win on a weaker i7 with less cache, but maybe we can take that as an indication that we can push this further by making other code smaller :)

PR #13942, opened 2024-10-22
Removing the deprecated parameters, -fast, -slow, -crossCheckTermVectors from CheckIndex.

slow-J:
Removing the deprecated parameters, -fast, -slow, -crossCheckTermVectors from CheckIndex.
Their usage is replaced with `-level` with respective values of `1`, `3`, `3`.

Follow-up on the deprecation done in #11023.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -7,7 +7,7 @@ http://s.apache.org/luceneversions
 
 API Changes
 ---------------------
-(No changes)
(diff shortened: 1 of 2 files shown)


stefanvodita (approved):
Thanks for coming back to this! I see that [MIGRATE.md](https://github.com/apache/lucene/blob/main/lucene/MIGRATE.md) doesn't list `crossCheckTermVectors`, but it did print a warning, so I hope it's fine to remove.

jpountz (approved):
Oh, our first 11.0 change. :)

PR #193, opened 2021-06-23
For stability of DisjunctionIntervalsSource.toString(), sort subSources

magibney:
Iterators over subSources of DisjunctionIntervalsSource may return elements in indeterminate order, requiring explicit sorting to make toString() output stable across equivalent instances.

PR #13944, opened 2024-10-22
Ensure stability of clause order for DisjunctionMaxQuery toString

ljak:
Since https://github.com/apache/lucene/pull/110, the disjuncts elements of DisjunctionMaxQueries don't have an order anymore, which is impacting the `toString` method. In isolation, that does not matter. But, in Solr, when the debug component is needed for a distributed query, every shard can return a different toString representation of the same query... and the different toString keys of the debug response will have an array value, containing those different representations (instead of having one value for one same representation).

Example with the `parsedquery_toString` key (of a json response within Solr):
`parsedquery_toString":["((docIdentifiers:\"Okarandeep Osingh\" docIdentifiers:Otest) | (docTitle:\"Okarandeep Osingh\" docTitle:Otest) | (docBody:\"Okarandeep Osingh\" docBody:Otest))","((docBody:\"Okarandeep Osingh\" docBody:Otest) | (docTitle:\"Okarandeep Osingh\" docTitle:Otest) | (docIdentifiers:\"Okarandeep Osingh\" docIdentifiers:Otest))"]`

When PR110 was merged, Solr adapted its unit tests this way: https://github.com/apache/solr/pull/117 but, later on within Lucene, the toString method of DisjuctionIntervalsSource was adapted in prevision of a potential similar future change: https://github.com/apache/lucene/pull/193. 

I adapted the toString method of DisjunctionMaxQueries similarly to this PR.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -70,6 +70,7 @@ Bug Fixes
 * GITHUB#13884: Remove broken .toArray from Long/CharObjectHashMap entirely. (Pan Guixin)
 * GITHUB#12686: Added support for highlighting IndexOrDocValuesQuery. (Prudhvi Godithi)
 * GITHUB#13927: Fix StoredFieldsConsumer finish. (linfn)
+* GITHUB#13944: Ensure deterministic order of clauses for `DisjunctionMaxQuery#toString`. (Laurent Jakubina)
 
 Build
 ---------------------
--- a/lucene/core/src/java/org/apache/lucene/search/DisjunctionMaxQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/DisjunctionMaxQuery.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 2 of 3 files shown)


jpountz (changes requested):
I'm a bit unhappy about making `toString()` run in O(n log(n)) with the number of clauses. Could we instead save a list copy of the queries in the constructor and use the list for `toString()`? `BooleanQuery` does something similar.

ljak:
Thanks for the feedback. Looking at `BooleanQuery`, it "only" has one list `List<BooleanClause> clauses`. So, is the idea to have 2 structures for the `DisjunctionMaxQuery`, the unordered multiset of queries and a sorted list of queries, where the latter is only used for the `toString` method ?

jpountz:
See `BooleanQuery#clauseSets`, which is used for equals()/hashcode() and `BooleanQuery#clauses`, which is used for toString().

ljak:
I needed to sort the `Query`s in some ways, so I compare them according to their toString representation:

`orderedQueries.sort(Comparator.comparing(Query::toString));`

Not sure if it's the right way.

jpountz:
I wouldn't sort them, and just rely on the order that the caller supplied?

jpountz:
Yes, exactly.

jpountz (approved):

jpountz:
Can you add an entry to `lucene/CHANGES.txt` under version 10.1.0? Then I'll merge.

ljak:
Done. Thanks for reviewing!

dsmiley:
Thanks for this!  I had to redo a bunch of tests over this matter at work.

PR #13947, opened 2024-10-23
Fix ord-to-doc mapping when searching Lucene 9.0.0 hnsw indices

benwtrent:
Bug discovered in https://github.com/apache/lucene/pull/13910

This corrects the logic so that users in lucene 9.12.1 will be able to correctly read from Lucene 9.0.0 HNSW indices. Previously, if they created an index in lucene 9.0.0 and not all documents contained a vector, reading in 9.8+ would produce non-sensical scores.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -7,7 +7,8 @@ http://s.apache.org/luceneversions
 
 Bug Fixes
 ---------------------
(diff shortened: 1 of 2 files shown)


jpountz (approved):

msokolov (commented):
thanks!

PR #12460, opened 2023-07-25
Allow reading binary doc values as a DataInput

iverase:
see https://github.com/apache/lucene/issues/12459

Issue #12459, opened 2023-07-25
Allow reading binary doc values as a DataInput


iverase:
### Description

Binary doc values allow to store a variable number of bytes on a doc value. In order to read those bytes, we currently get a BytesRef from the API which contains the bytes on heap. In order to do that, the current implementation preallocates a byte array with the size equals to the biggest doc value. This strategy has two main drawbacks:

1) We are using a byte array as a middle data structure so we are copying each doc value in this byte array. In many cases this is an unnecessary overhead.

2) when one of the doc values is big, in the order of few megabytes, it can cause issues with small heaps (or even big heaps if big enough). This is due to the allocation of a big byte arrays upfront, that can be consider humongous allocations by the G1 garbage collector and it can cause heap issues under high load.

Therefore I would like to propose to add a new API on top of binary doc values that allows reading them using a DataInput. This data input can read directly from the underlaying IndexInput and therefore we don't need to copy data into an intermediate data structure and we don't need to preallocate a byte array.

The new API would be built on top of the existing binary doc values and it would look something like:

```
public abstract class DataInputDocValues extends DocValuesIterator {

  /** Sole constructor. (For invocation by subclass constructors, typically implicit.) */
  protected DataInputDocValues() {}

  /**
   * Returns the binary value wrapped as a {@link DataInput} for the current document ID. It is
   * illegal to call this method after {@link #advanceExact(int)} returned {@code false}.
   *
   * @return the binary value wrapped as a {@link DataInput}
   */
  public abstract DataInput dataInputValue() throws IOException;
}
```

iverase:
I wrote a prototype for this change here: https://github.com/apache/lucene/pull/12460

jpountz:
Another argument in favor of this proposal that I can think of is that many call sites of this API want to decode numbers from the byte[] and end up re-inventing `DataInput#readInt` and similar methods.

iverase:
I expect that the use case of wrapping the byte[] from a binary doc values with a `ByteArrayDataInput` to read the contents to be common, so this approach will give that in a more efficient way.

I think we should provide the possibility to be able to have random access so I would like to refine my original proposal by adding a DataInputDocValue which is just a DataInput which is positional aware. The API would look like:

```
public abstract class DataInputDocValues extends DocValuesIterator {

  /** Sole constructor. (For invocation by subclass constructors, typically implicit.) */
  protected DataInputDocValues() {}

  /**
[...]

PR #13948, opened 2024-10-23
Allow reading binary doc values as a RandomAccessInput

iverase:
Following up this [suggestion](https://github.com/apache/lucene/pull/12460#issuecomment-1715126194) from @jpountz, here I propose to add a new method to the BinaryDocValues API that returns the content of the doc values as a  RandomAccessInput. This will allow to have an off-heap version of the binary doc value whenever possible.

The main two changes here are:

1) Lucene90DocValuesProducer allocates the BytesRef for binary doc values lazily.

2) To make transparent when the binary doc value is already on heap, BytesRef implements RandomAccessInput so we can just return the result from binaryValue() in those cases. If that's not ok, we can create a RandomAccessInput wrapper for BytesRef instead.

relates https://github.com/apache/lucene/issues/12459


lucene/core/src/java/org/apache/lucene/store/RandomAccessInputDataInput.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  public void setPosition(long pos) {
+    this.pos = pos;
+  }
+
+  public void reset(RandomAccessInput input) {
+    this.input = input;
+    pos = 0;
+  }
```
jpountz, line 52: Can you add javadocs to the above methods?

lucene/core/src/java/org/apache/lucene/store/RandomAccessInputDataInput.java
```diff
@@ -0,0 +1,104 @@
...
+  public void skipBytes(long count) {
+    pos += count;
+  }
+
+  @Override
+  public short readShort() throws IOException {
+    try {
+      return input.readShort(pos);
+    } finally {
+      pos += Short.BYTES;
```
jpountz, line 68: This finally block feels wrong: the position should not be updated if the call raises an exception.

jpountz (commented):
Using `RandomAccessInput` addresses problems with multiple callers wanting to consume the value, that helps.

In my opinion, we should not have both a `binaryValue()` and a `randomAccessValue()` getters in the long term. Can we remove the `binaryValue()` getter on the main branch and make it deprecated when we backport to 10.x?

> we can create a RandomAccessInput wrapper for BytesRef instead.

I would like it better this way.

lucene/core/src/java/org/apache/lucene/index/BinaryDocValues.java
```diff
@@ -33,4 +34,15 @@ protected BinaryDocValues() {}
    * @return binary value
    */
   public abstract BytesRef binaryValue() throws IOException;
+
+  /**
+   * Returns the binary value as a {@link RandomAccessInput} for the current document ID. The bytes
+   * start at position 0 up to {@link RandomAccessInput#length()}. It is illegal to call this method
+   * after {@link #advanceExact(int)} returned {@code false}.
```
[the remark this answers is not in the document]
iverase: done

lucene/core/src/java/org/apache/lucene/store/RandomAccessInputDataInput.java
iverase: done

jpountz:
In my experience, binary doc values are more often used to encode structured data, such as maps that help build scoring signals, geo shapes, etc. than actual binary content, so this change makes sense to me. I'm interested in having more opinions though.

Would be nice to extend AssertingBinaryDocValues to make sure that all reads in the input are within bounds.

iverase:
I have a second iteration and I realise that this change can be easily done if we introduce a new abstraction. I have called it `RandomAccessInputRef` and it is  equivalent to BytesRef but backed by a RandomAccessInput  instead of a byte array. I am thinking of it s the off-heap version of BytesRef.

Using this, it makes the changes almost mechanical. The big difference between BytesRef and RandomAccessInputRef is that the latter cannot implement efficient the Compare interface so it cannot be use for comparisons and equality. In that case we need to read bytes on-heap.

With this new abstraction, we just need to add three new methods:

- BytesRefBuilder#copyBytes(RandomAccessInputRef in) throws IOException
- BytesRefArray#append(RandomAccessInputRef bytes) throws IOException 
- ByteBlockPool#append(final RandomAccessInput bytes, long offset, int length) throws IOException: for supporting the method above)
- 
And two helper classes:

- ByteArrayRandomAccessInput: To represent and byte array as a RandomAccessInput
- RandomAccessInputDataInput: to wrap a RandomAccessInput as a DataInput so we can use DataOutput#copyBytes(DataInput input, long numBytes)

There are few places that still needs to be adapted to take full advantage of the new API but  they can be done in a follow up as they require a bit of work :

- MatchingFacetSetsCounts
- BaseShapeDocValuesQuery
- SerializedDVStrategy
- MonitorQuerySerializer

And there are two places that we will need to put the data on-heap as they are using equality:

- TermValComparator
- BinaryDocValueSelector

lucene/codecs/src/java/org/apache/lucene/codecs/simpletext/SimpleTextDocValuesReader.java
```diff
@@ -383,17 +392,31 @@ public long cost() {
...
+          return true;
+        }
+        return false;
       }
 
       @Override
-      public BytesRef binaryValue() throws IOException {
-        return values.apply(docsWithField.docID());
+      public RandomAccessInputRef randomAccessInputValue() {
+        return inputRef;
```
jpountz, line 419: What was the problem with calling `values#apply` in `randomAccessInputValue()` rather than in `nextDoc()`/`advance()`/`advanceExact()`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90DocValuesProducer.java
```diff
@@ -791,25 +792,26 @@ public BinaryDocValues getBinary(FieldInfo field) throws IOException {
       return DocValues.emptyBinary();
     }
 
-    final RandomAccessInput bytesSlice = data.randomAccessSlice(entry.dataOffset, entry.dataLength);
+    final RandomAccessInputRef bytesSlice =
+        new RandomAccessInputRef(data.randomAccessSlice(entry.dataOffset, entry.dataLength));
```
jpountz, line 796: Nit, maybe extract two variables: `bytesSlice` and `byteSliceRef` to make it a bit easier to read?

lucene/core/src/java/org/apache/lucene/search/FieldComparator.java
```diff
@@ -234,7 +235,8 @@ public TermValComparator(int numHits, String field, boolean sortMissingLast) {
 
     private BytesRef getValueForDoc(int doc) throws IOException {
       if (docTerms.advanceExact(doc)) {
-        return docTerms.binaryValue();
+        bytesRefBuilder.copyBytes(docTerms.randomAccessInputValue());
+        return bytesRefBuilder.get();
```
jpountz, line 239: Nit: it's a bit sad to load all bytes on top heap, when in many cases we could identify that the value is not competitive after reading a few bytes only?

lucene/core/src/java/org/apache/lucene/store/RandomAccessInputDataInput.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @lucene.experimental
+ */
+public final class RandomAccessInputDataInput extends DataInput {
+
+  private RandomAccessInput input;
+  private long offset;
+  private long length;
+
+  private long pos;
```
jpountz, line 34: I wonderif `pos` should start at `offset` rather than `0` so that we don't have to sum up `offset` and `pos` whenever we need to know the current position?

lucene/core/src/java/org/apache/lucene/util/UnicodeUtil.java
```diff
@@ -627,35 +629,58 @@ public static String toHexString(String s) {
...
-   * Interprets the given byte array as UTF-8 and converts to UTF-16. It is the responsibility of
-   * the caller to make sure that the destination array is large enough.
+   * Interprets the given {@link RandomAccessInput} slice as UTF-8 and converts to UTF-16. It is the
+   * responsibility of the caller to make sure that the destination array is large enough.
+   *
+   * <p>NOTE: Full characters are read, even if this reads past the length passed (and can result in
+   * an IOException if invalid UTF-8 is passed). Explicit checks for valid UTF-8 are not performed.
+   */
+  // TODO: broken if chars.offset != 0
+  public static int UTF8toUTF16(RandomAccessInput input, long offset, int length, char[] out)
```
jpountz, line 639: Having `UnicodeUtil` need to work on a `RandomAccessInput` doesn't feel right. Do you know if this is performance-sensitive? Otherwise I'd rather make `RandomAccessInputRef` implement `utf8ToString()` by converting to a `BytesRef` first.

jpountz (commented):
The general direction makes sense to me. Using `RandomAccessInput` vs. `DataInput` helps since the former is stateless. I first didn't like that it's a `RandomAccessInputRef` as opposed to a plain `RandomAccessInput` but I have come to like it when looking deeper at the PR.

I ran the sorting tasks on wikibigall and saw no regression to `HighTermTitleBDVSort` (but no speedup either):

```
                      TermDTSort      259.60      (5.7%)      258.98      (6.8%)   -0.2% ( -12% -   13%) 0.904
            HighTermTitleBDVSort       14.52      (3.1%)       14.51      (3.0%)   -0.1% (  -5% -    6%) 0.944
           HighTermDayOfYearSort      476.39      (3.4%)      478.38      (5.7%)    0.4% (  -8% -    9%) 0.780
                        PKLookup      233.90      (3.4%)      235.06      (3.3%)    0.5% (  -5% -    7%) 0.636
               HighTermMonthSort     1661.20      (3.6%)     1677.10      (4.4%)    1.0% (  -6% -    9%) 0.449
```

lucene/codecs/src/java/org/apache/lucene/codecs/simpletext/SimpleTextDocValuesReader.java
iverase: None, I rework it so we call `values#apply` only  in `randomAccessInputValue()`

lucene/core/src/java/org/apache/lucene/store/RandomAccessInputDataInput.java
iverase: Done, I rework it so it looks like `ByteArrayDataInput`

lucene/core/src/java/org/apache/lucene/util/UnicodeUtil.java
iverase: I revert this changes. If we see it is performance sensitive we can re add it.

lucene/core/src/java/org/apache/lucene/search/FieldComparator.java
iverase: agree, it will require a bit of rework here but should be doable. Maybe as a follow up?

lucene/core/src/test/org/apache/lucene/index/Test2BBinaryDocValues.java
```diff
@@ -88,12 +89,13 @@ public void testFixedBinary() throws Exception {
       LeafReader reader = context.reader();
       BinaryDocValues dv = reader.getBinaryDocValues("dv");
       for (int i = 0; i < reader.maxDoc(); i++) {
+        ;
```
jpountz, line 92: remove?

lucene/core/src/test/org/apache/lucene/util/TestBytesRefArray.java
```diff
@@ -43,8 +44,17 @@ public void testAppend() throws IOException {
...
+          BytesRef bytesRef = spare.get();
+          RandomAccessInputRef inputRef =
+              new RandomAccessInputRef(
+                  new ByteArrayRandomAccessInput(bytesRef.bytes), bytesRef.offset, bytesRef.length);
+          assertEquals(i + initSize, list.append(inputRef));
+        } else {
+          assertEquals(i + initSize, list.append(spare.get()));
+        }
         stringList.add(randomRealisticUnicodeString);
+        ;
```
jpountz, line 57: remove?

lucene/monitor/src/java/org/apache/lucene/monitor/MonitorQuerySerializer.java
```diff
@@ -48,38 +47,31 @@ public interface MonitorQuerySerializer {
...
+      public MonitorQuery deserialize(RandomAccessInputRef input) throws IOException {
+        RandomAccessInputDataInput data = new RandomAccessInputDataInput();
+        data.reset(input);
+        String id = data.readString();
+        String query = data.readString();
+        Map<String, String> metadata = new HashMap<>();
+        for (int i = data.readInt(); i > 0; i--) {
+          metadata.put(data.readString(), data.readString());
         }
+        return new MonitorQuery(id, parser.apply(query), query, metadata);
```
jpountz, line 61: Nice that it doesn't load everything it a giant byte[] anymore!

lucene/core/src/java/org/apache/lucene/util/RandomAccessInputRef.java
```diff
@@ -0,0 +1,107 @@
...
+    return new String(ref, 0, len);
+  }
+
+  /**
+   * Creates a new BytesRef that points to a copy of the bytes from <code>input</code> starting at
+   * offset for length.
+   *
+   * <p>The returned BytesRef will have a offset of zero.
+   */
+  public static BytesRef toBytesRef(RandomAccessInputRef input) throws IOException {
```
jpountz, line 67: I'd be keen to make it a method rather than a static function?

jpountz (approved):
This looks good to me!

lucene/core/src/java/org/apache/lucene/util/RandomAccessInputRef.java
iverase: I think I make it static to mimic the  DeepCopy method on BytesRef but it loos nicer as a method.

PR #13949, opened 2024-10-23
Add new Directory implementation for AWS S3

albogdano:
### Description

This PR adds a new module `s3directory` to Lucene, containing a new `Directory` implementation for AWS S3.
The code was adapted from the [lucene-s3-directory](https://github.com/albogdano/lucene-s3directory) project as requested by @jpountz in #13868

Currently, there are a few issues which cause the build to fail:
- AWS SDK module not found or could not be resolved in `module-info.java`
- tests require an AWS account and valid credentials to execute


lucene/s3directory/build.gradle
```diff
@@ -0,0 +1,38 @@
...
+      strictly '1.13'
+    }
+  }
+
+  moduleApi project(':lucene:core')
+
+  moduleTestImplementation project(':lucene:test-framework')
+  moduleTestImplementation project(':lucene:analysis:common')
+  moduleTestImplementation project(':lucene:backward-codecs')
+  moduleTestImplementation project(':lucene:queryparser')
```
jpountz, line 36: We shouldn't need to depend on the `analysis-common`, `backward-codecs` and `queryparser` modules for tests. For analysis, our tests usually use `MockAnalyzer` rather than whitespace or keyword analyzers.

lucene/s3directory/src/test/org/apache/lucene/store/s3/TestS3Directory.java
```diff
@@ -0,0 +1,342 @@
...
+  @Override
+  public void setUp() {
+    s3Directory = new S3Directory(TEST_BUCKET, "", S3LockFactory.INSTANCE);
+    s3Directory.create();
+
+    try {
+      ramDirectory = new MMapDirectory(FileSystems.getDefault().getPath("target/index"));
+      fsDirectory = FSDirectory.open(FileSystems.getDefault().getPath("target/index"));
+    } catch (IOException ex) {
+      logger.log(Level.SEVERE, null, ex);
```
jpountz, line 79: Let's throw this exception instead of swallowing it, we should fail the test suite if this throws an exception?

lucene/s3directory/src/test/org/apache/lucene/store/s3/TestS3Directory.java
```diff
@@ -0,0 +1,342 @@
...
+      iwriter.addDocument(doc);
+      if (iwriter.hasUncommittedChanges()) {
+        iwriter.commit();
+      }
+      if (iwriter.isOpen()) {
+        iwriter.getDirectory().close();
+      }
+      iwriter.forceMerge(1, true);
+    } catch (Exception e) {
+      logger.log(Level.SEVERE, null, e);
```
jpountz, line 156: likewise here, let's not swallow exceptions in general

lucene/s3directory/src/test/org/apache/lucene/store/s3/TestS3Directory.java
```diff
@@ -0,0 +1,342 @@
...
+    }
+    // Now search the index:
+    try (DirectoryReader ireader = DirectoryReader.open(s3Directory)) {
+      final IndexSearcher isearcher = new IndexSearcher(ireader);
+      // Parse a simple query that searches for "text":
+
+      final QueryParser parser = new QueryParser("fieldname", analyzer);
+      final Query query = parser.parse("text");
```
jpountz, line 164: Let's create the query manually instead and remove the dependency on the query parser module? This could be:

```java
final Query query = new TermQuery(new Term("fieldname", "text"));
```

lucene/s3directory/src/test/org/apache/lucene/store/s3/TestS3Directory.java
```diff
@@ -0,0 +1,342 @@
...
+import org.apache.lucene.store.FlushInfo;
+import org.apache.lucene.store.IOContext;
+import org.apache.lucene.store.IndexInput;
+import org.apache.lucene.store.IndexOutput;
+import org.apache.lucene.store.MMapDirectory;
+import org.apache.lucene.tests.util.LuceneTestCase;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TestS3Directory extends LuceneTestCase {
```
jpountz, line 53: Could we make this test extend `BaseDirectoryTestCase`, it has a few tests for various things that directories are supposed to be able to do like cloning and slicing.

lucene/s3directory/src/java/org/apache/lucene/store/s3/S3Directory.java
```diff
@@ -0,0 +1,1281 @@
...
+
+/**
+ * A S3 based implementation of a Lucene <code>Directory</code> allowing the storage of a Lucene
+ * index within S3. The directory works against a single bucket, where the binary data is stored in
+ * <code>objects</code>. Each "object" has an entry in the S3, and different FileEntryHandler can be
+ * defines for different files (or files groups).
+ */
+public class S3Directory extends Directory {
+
+  private static final Logger logger = Logger.getLogger(S3Directory.class.getName());
```
jpountz, line 65: As a library, Lucene never logs, so let's remove this logger and make sure we propagate exceptions when appropriate?

lucene/s3directory/src/java/org/apache/lucene/store/s3/S3Directory.java
```diff
@@ -0,0 +1,1281 @@
...
+   * @param pathName The S3 path (path prefix) within the bucket
+   * @param lockFactory A lock factory implementation
+   */
+  public S3Directory(final String bucketName, final String pathName, LockFactory lockFactory) {
+    initialize(bucketName, pathName, lockFactory);
+  }
+
+  private void initialize(final String bucket, final String path, LockFactory lockFactory) {
+    this.s3 = S3Client.create();
+    this.bucket = bucket.toLowerCase(Locale.getDefault());
```
jpountz, line 105: Nit: let's not rely on the default locale. I would expect our forbidden-apis build task to fail the build because of this.

lucene/s3directory/src/java/org/apache/lucene/store/s3/S3Directory.java
```diff
@@ -0,0 +1,1281 @@
...
+    // and code duplication between this method and S3BufferedIndexInput.
+    // Performance is much better this way!
+    @Override
+    protected void refill() throws IOException {
+      ResponseInputStream<GetObjectResponse> res =
+          s3Directory
+              .getS3()
+              .getObject(b -> b.bucket(s3Directory.getBucket()).key(getPath() + name));
+
+      synchronized (this) {
```
jpountz, line 559: Synchronization shouldn't be needed as `IndexInputs` should be consumed from a single thread anyway?

lucene/s3directory/src/java/org/apache/lucene/store/s3/S3Directory.java
```diff
@@ -0,0 +1,1281 @@
...
+        }
+        doAfterClose();
+      }
+    }
+
+    /**
+     * A simple base class that performs index output memory based buffering. The buffer size if
+     * configurable.
+     */
+    // NEED TO BE MONITORED AGAINST LUCENE
```
jpountz, line 1142: This is Lucene now. :)

jpountz (commented):
Thanks for contributing this!

I'm surprised at the number of dependencies, does the S3 SDK actually have this many dependencies? E.g. https://central.sonatype.com/artifact/software.amazon.awssdk/s3/dependencies suggests that netty-transport is a test-only dependency? (or is it also a transitive dependency which is why you had to add it here?)

Regarding testing, a quick search suggests that there are mocks for S3 that avoid connecting to an actual S3 bucket, would it work? https://stackoverflow.com/questions/6615988/how-to-mock-amazon-s3-in-an-integration-test

I left some other comments from a quick pass over the code, it would be nice to make the code conform to some Lucene expectations, such as no logging.

rmuir:
Have we explored trying a fuse-based mount instead of writing java code?

There seems to be several potentially viable approaches there: https://github.com/s3fs-fuse/s3fs-fuse, but maybe higher-performant https://github.com/kahing/goofys + https://github.com/kahing/catfs

josefschiefer27:
Amazon S3 also offers its own solution for mounting S3 buckets, available at https://github.com/awslabs/mountpoint-s3/.

There are benchmarks at https://github.com/awslabs/mountpoint-s3/blob/main/doc/BENCHMARKING.md

[throughput chart](https://awslabs.github.io/mountpoint-s3/dev/bench/)
    [throughput (with cache) chart](https://awslabs.github.io/mountpoint-s3/dev/cache_bench/)
    [throughput (s3-express) chart](https://awslabs.github.io/mountpoint-s3/dev/s3-express/bench)
    [latency chart](https://awslabs.github.io/mountpoint-s3/dev/latency_bench/)
    [latency (s3-express) chart](https://awslabs.github.io/mountpoint-s3/dev/s3-express/latency_bench/)

albogdano:
@rmuir @josefschiefer27 I haven't yet tried experimenting with S3 fuse mounts but I doubt the performance will be any better than using the S3 API directly. Also, there are some specific API calls around the legal hold locks used in this project as a locking mechanism. I don't know if those are supported by the libraries you mentioned.
Right now I'm focused on minimizing dependencies by integrating an alternative S3 client - this [lightweight AWS client](https://github.com/davidmoten/aws-lightweight-client-java). Ideally IMHO, we should have a lightweight S3 Directory implementation that is suitable for serverless deployments.

reta:
Just to share similar developments, we did S3 filesystem plugin store prototype for OpenSearch based on the same library [1], it works but the latency wise, it is not great (obviously, no caching etc have been done). 

[1] https://github.com/reta/OpenSearch/tree/s3.fs/plugins/store-s3

jpountz:
Since we have a S3 Directory implemented already, let's run a comparison with the fuse mount approach?

albogdano:
I refactored the code and got rid of the AWS SDK in favor of the minimal S3 Client [aws-lightweight-client-java](https://github.com/davidmoten/aws-lightweight-client-java). 
Also, I added an S3 mocking library - [S3Mock](https://github.com/adobe/S3Mock) but it requires additional security permissions and I suppose `tests.policy` has to be modified. 

Otherwise, this `Directory` implementation passes most of the tests in `BaseDirectoryTestCase` except for these four:

- `testPrefetchOnSlice` (:lucene:s3directory)
- `testCopyBytes` (:lucene:s3directory)
- `testPrefetch` (:lucene:s3directory)
- `testCopyBytesWithThreads` (:lucene:s3directory)

I couldn't figure out why that is the case and I would be grateful if anyone could give me some pointers here.

dsmiley:
Couldn't S3 and other file storage be implemented as an NIO FileSystem instead?  AKA JSR-203.  Would the Lucene Directory abstraction level have certain advantages (what)?  Ideally we'd use such abstractions that the Java platform offers so that Lucene isn't tempted to have to maintain custom implementations at its higher layer.

Some google searching seems to show this one as the most/best maintained: https://github.com/carlspring/s3fs-nio

dsmiley:
By the way, the Apache Solr project has an impressive "[BlockCache](https://github.com/apache/solr/blob/07943e87fb762b69a66932f777d56eb14cc72e78/solr/modules/hdfs/src/java/org/apache/solr/hdfs/store/blockcache/BlockCache.java)" Directory wrapper to cache data either on or off heap.  It's about to be deleted merely because of its adjacency with Solr's outgoing HDFS support, but the BlockCache Java package is kind of dependency-free except for Caffeine-Cache.  Technically there are some trivial references to other things that aren't essential.  I think it would make sense for it to live in Lucene for use by people using Directories with remote storage, where the OS-cache isn't in-play.

PR #13036, opened 2024-01-26
Optimize counts on two clause term disjunctions

jfreden:
Calculate `count(clause1 OR clause2)` as `count(clause1) + count(clause2) - count(clause1 AND clause2)`  as discussed in https://github.com/apache/lucene/issues/12644. 

**Todo:** 
- Update nightly benchmarks to either have deletes or add tasks on disjunctions of 3 terms or more to retain coverage on actual disjunction counts

PR #13950, opened 2024-10-23
Make some BooleanQuery methods public and a new `#add(Collection)` method for BQ builder

shubhamvishu:
### Description

Changes in this PR :

1. Makes `getClauses(Occur occur` method public to get the collection of one type of clauses. Without public I had to call `getClauses` and then iterate to store this info whereas `BoolenQuery` is [already storing this info in clauseSets](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java#L111) which seems like a good candidate to be exposed for reuse this method. This would also simplify not having checks like eg: [1](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java#L366-L370), [2](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java#L461-L465), [3](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java#L530-L532), [4](https://github.com/apache/lucene/blob/main/lucene/monitor/src/java/org/apache/lucene/monitor/QueryAnalyzer.java#L86-L87) etc along with the `add(Collection)` method(point 3 below). 
2. Similarly makes `#isPureDisjunction` and `isTwoClausePureDisjunctionWithTerms` public which looks could be something useful and right thing to be exposed by the `BooleanQuery` class.
3. Adds an `add(Collection<BooleanClause> collection)` method to `BooleanQuery.Builder`. While doing some query manipulation I had to insert a set of clause but one by one. I'd be useful if there was a `add(Collection<BooleanClause> collection)` to add the to the maintained list of clauses and simplify it for cases where users are manipulating the BQ. I see some places in code that could be refactored to avoid looping to insert the list of clauses as mentioned in point 1.

I'd love to hear thoughts and suggestions. I don't know if there are any rationale of not having these methods public earlier(or a previous attempt) but would love to know.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
@@ -88,6 +88,28 @@ public class BooleanQuery extends Query implements Iterable<BooleanClause> {
     }
 
     /**
+     * Add a collection of BooleanClause's to this {@link Builder}. Note that the order in which
+     * clauses are added does not have any impact on matching documents or query performance.
+     *
+     * @throws IndexSearcher.TooManyClauses if the new number of clauses exceeds the maximum clause
+     *     number
+     */
+    public Builder add(Collection<BooleanClause> collection) {
+      // We do the final deep check for max clauses count limit during
+      // <code>IndexSearcher.rewrite</code> but do this check to short
+      // circuit in case a single query holds more than numClauses
+      //
+      // NOTE: this is not just an early check for optimization -- it's
+      // neccessary to prevent run-away 'rewriting' of bad queries from
+      // creating BQ objects that might eat up all the Heap.
+      if ((clauses.size() + collection.size()) > IndexSearcher.maxClauseCount) {
+        throw new IndexSearcher.TooManyClauses();
+      }
+      clauses.addAll(collection);
+      return this;
+    }
+
+    /**
      * Add a new clause to this {@link Builder}. Note that the order in which clauses are added does
      * not have any impact on matching documents or query performance.
      *
@@ -136,7 +158,7 @@ public class BooleanQuery extends Query implements Iterable<BooleanClause> {
   }
 
   /** Return the collection of queries for the given {@link Occur}. */
-  Collection<Query> getClauses(Occur occur) {
+  public Collection<Query> getClauses(Occur occur) {
     return clauseSets.get(occur);
   }
 
@@ -144,12 +166,12 @@ public class BooleanQuery extends Query implements Iterable<BooleanClause> {
    * Whether this query is a pure disjunction, ie. it only has SHOULD clauses and it is enough for a
    * single clause to match for this boolean query to match.
    */
-  boolean isPureDisjunction() {
+  public boolean isPureDisjunction() {
     return clauses.size() == getClauses(Occur.SHOULD).size() && minimumNumberShouldMatch <= 1;
   }
 
   /** Whether this query is a two clause disjunction with two term query clauses. */
-  boolean isTwoClausePureDisjunctionWithTerms() {
+  public boolean isTwoClausePureDisjunctionWithTerms() {
     return clauses.size() == 2
         && isPureDisjunction()
         && clauses.get(0).query() instanceof TermQuery
--- a/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
@@ -88,20 +88,14 @@ public class BooleanQuery extends Query implements Iterable<BooleanClause> {
     }
 
     /**
-     * Add a collection of BooleanClause's to this {@link Builder}. Note that the order in which
+     * Add a collection of BooleanClauses to this {@link Builder}. Note that the order in which
      * clauses are added does not have any impact on matching documents or query performance.
      *
      * @throws IndexSearcher.TooManyClauses if the new number of clauses exceeds the maximum clause
      *     number
      */
     public Builder add(Collection<BooleanClause> collection) {
-      // We do the final deep check for max clauses count limit during
-      // <code>IndexSearcher.rewrite</code> but do this check to short
-      // circuit in case a single query holds more than numClauses
-      //
-      // NOTE: this is not just an early check for optimization -- it's
-      // neccessary to prevent run-away 'rewriting' of bad queries from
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
```diff
@@ -87,6 +87,28 @@ public Builder add(BooleanClause clause) {
...
 
+    /**
+     * Add a collection of BooleanClause's to this {@link Builder}. Note that the order in which
+     * clauses are added does not have any impact on matching documents or query performance.
+     *
+     * @throws IndexSearcher.TooManyClauses if the new number of clauses exceeds the maximum clause
+     *     number
+     */
+    public Builder add(Collection<BooleanClause> collection) {
+      // We do the final deep check for max clauses count limit during
```
msokolov, line 98: maybe just say "see addClause()" ?

msokolov (approved):
This LGTM but it has been this way for so long I wonder if someone with more historical knowledge will have a rationale for hiding these things?

lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
```diff
@@ -87,6 +87,28 @@ public Builder add(BooleanClause clause) {
       return this;
     }
 
+    /**
+     * Add a collection of BooleanClause's to this {@link Builder}. Note that the order in which
```
mikemccand, line 91: Remove the `'` -- just `BooleanClauses`.

gsmiller:
> This LGTM but it has been this way for so long I wonder if someone with more historical knowledge will have a rationale for hiding these things?

I took a quick peek at the git history and it looks like:
1. `#getClauses` was added back in LUCENE-6944 as a new method and hasn't been touched since. I would guess it was added with the most limited access visibility required as a default practice?
2.  `#isPureDisjunction` was added in LUCENE-7386 in a pattern that looks similar to above (only added with necessary pkg-internal visibility).
3. `#isTwoClausePureDisjunctionWithTerms` was added recently in #13036, and looks like yet another situation where only the necessary visibility was given for the reason it was being added.

So tl;dr: it looks like access was restricted as a good practice for the use-cases they were being added for.

IMO, I think it's fine to expose `#getClauses(Occur)` since it does provide a better way to access data already being publicly exposed (i.e., makes sense to not require the client to iterate all the clauses when there's a better way to directly do this). I'm a little less convinced on the other two though. Is there a concrete use-case for exposing these? While, on one hand, it's not really leaking any internals to do this, it's also something the calling code can figure out on its own just as easily with what's already provided. While it's nice for calling code to not have to reimplement this logic, it's also nice to keep the API surface tight. Maybe there's a good motivating use-case though (and sorry if I missed this somewhere)?

lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
```diff
@@ -136,20 +158,20 @@ public List<BooleanClause> clauses() {
...
    * single clause to match for this boolean query to match.
    */
-  boolean isPureDisjunction() {
+  public boolean isPureDisjunction() {
     return clauses.size() == getClauses(Occur.SHOULD).size() && minimumNumberShouldMatch <= 1;
   }
 
   /** Whether this query is a two clause disjunction with two term query clauses. */
-  boolean isTwoClausePureDisjunctionWithTerms() {
+  public boolean isTwoClausePureDisjunctionWithTerms() {
```
jpountz, line 174: I can understand why someone would want to make `getClauses` public, but I wouldn't make the two above methods public, these are just implementation details of some rewrite rules?

jpountz: Oh, just saw that @gsmiller wrote this in a much better way before :)

lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
shubhamvishu: Sure

lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
shubhamvishu: Sure

shubhamvishu:
Thanks everyone for taking a look! As per comments the `#isPureDisjunction` and `#isTwoClausePureDisjunctionWithTerms` seems to be the debatable ones :

- About `#isTwoClausePureDisjunctionWithTerms`, I'm myself not super convinced considering its targeting a very specific type of query and I agree it might only be useful in rewrite optimizations. Its good to have these helper functions but since it only has one usage(same for `#isPureDisjunction`) currently prolly keeping it more local to the only usage is better than in highly visible class like `BooleanQuery` (but then it loses the chance of being visible and reused :) ). Anyways I'll remove it. 
- `#isPureDisjunction` does seem and also sound more generic/useful comparatively to above one but I'm not able to think of any specific use case generic enough to convince exposing(though there maybe one for this).

I removed these 2 for now and fixed the comments. Thanks!

jpountz (approved):

uschindler (commented):
Please fix this or revert this before 10.1 release.

lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
```diff
@@ -136,7 +152,7 @@ public List<BooleanClause> clauses() {
   }
 
   /** Return the collection of queries for the given {@link Occur}. */
-  Collection<Query> getClauses(Occur occur) {
+  public Collection<Query> getClauses(Occur occur) {
     return clauseSets.get(occur);
```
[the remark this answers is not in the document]
uschindler: Whenever you add getters for collections, make sure to make the unmodifiable (either before or clone them).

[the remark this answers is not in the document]
uschindler: I am preparing a fix for this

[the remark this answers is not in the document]
uschindler: Fix is here: #14074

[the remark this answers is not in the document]
shubhamvishu: Nice catch! Thanks @uschindler !

PR #13951, opened 2024-10-23
Support multi-tenant RAM buffers for IndexWriter

mdmarshmallow:
### Description

Draft PR to outline my initial approach. I introduced `IndexWriterRamManager` to control writer flushes. I also have a function `IndexWriterRamManager#chooseWriterToFlush` that lets the user choose which writer they specifically want to be flushed, sort of like a simple flush policy. Currently it defaults to just choosing the writer with the most RAM usage. 

One thing I wanted to avoid was starting another thread to just poll the IndexWriter memory usage, so I just added a listener in `IndexWriter#maybeProcessEvents` which is called whenever docs are updated or deleted (`IndexWriterRamManager#flushIfNecessary`). I wanted this method to be called whenever `FlushPolicy#onChange` was called, as I believe they both kinda do the same thing

No unit tests yet, but will add them! I just wanted to have my approach sanity checked for now

### Revision 2
* Failing some unit tests but changed the general approach to address some of the feedback I received
* Still need to write new unit tests to test this code
* Now this will always create a `IndexWriterRAMManager` that is tied to the `IndexWriterConfig`. The user can either pass in their own if they want multiple `IndexWriter` instances to share a single buffer or just let the `IndexWriterRAMManager` be created per `IndexWriter`

### Revision 3
* Fixed unit tests

### Revision 4
* Added unit tests and fixed some bugs

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -941,14 +941,11 @@ public class IndexWriter
    * @param d the index directory. The index is either created or appended according <code>
    *     conf.getOpenMode()</code>.
    * @param conf the configuration settings according to which IndexWriter should be initialized.
-   * @param indexWriterRAMManager The RAM manager used for multi-tenant RAM management
    * @throws IOException if the directory cannot be read/written to, or if it does not exist and
    *     <code>conf.getOpenMode()</code> is <code>OpenMode.APPEND</code> or if there is any other
    *     low-level IO error
    */
-  public IndexWriter(
-      Directory d, IndexWriterConfig conf, IndexWriterRAMManager indexWriterRAMManager)
-      throws IOException {
+  public IndexWriter(Directory d, IndexWriterConfig conf) throws IOException {
     enableTestPoints = isEnableTestPoints();
     conf.setIndexWriter(this); // prevent reuse by other instances
     config = conf;
@@ -1216,27 +1213,9 @@ public class IndexWriter
         writeLock = null;
       }
     }
-
-    if (indexWriterRAMManager != null) {
-      this.indexWriterRAMManager =
-          new IndexWriterRAMManager.PerWriterIndexWriterRAMManager(this, indexWriterRAMManager);
-    } else {
-      this.indexWriterRAMManager = null;
-    }
-  }
-
-  /**
-   * Constructor for IndexWriter's that don't require multi-tenant RAM management
-   *
-   * @param d the index directory. The index is either created or appended according <code>
-   *     conf.getOpenMode()</code>.
-   * @param conf the configuration settings according to which IndexWriter should be initialized.
-   * @throws IOException if the directory cannot be read/written to, or if it does not exist and
-   *     <code>conf.getOpenMode()</code> is <code>OpenMode.APPEND</code> or if there is any other
-   *     low-level IO error
-   */
-  public IndexWriter(Directory d, IndexWriterConfig conf) throws IOException {
-    this(d, conf, null);
+    this.indexWriterRAMManager =
+        new IndexWriterRAMManager.PerWriterIndexWriterRAMManager(
+            this, config.getIndexWriterRAMManager());
   }
 
   /** Confirms that the incoming index sort (if any) matches the existing index sort (if any). */
@@ -1366,10 +1345,6 @@ public class IndexWriter
       }
       rollbackInternal(); // if we got that far lets rollback and close
     }
-
-    if (indexWriterRAMManager != null) {
-      indexWriterRAMManager.removeWriter();
-    }
   }
 
   /**
@@ -1395,6 +1370,7 @@ public class IndexWriter
    */
   @Override
   public void close() throws IOException {
+    indexWriterRAMManager.removeWriter();
     if (config.getCommitOnClose()) {
       shutdown();
     } else {
@@ -6042,9 +6018,7 @@ public class IndexWriter
       seqNo = -seqNo;
       processEvents(true);
     }
-    if (indexWriterRAMManager != null) {
-      indexWriterRAMManager.flushIfNecessary();
-    }
+    indexWriterRAMManager.flushIfNecessary(config.flushPolicy);
     return seqNo;
   }
 
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -462,7 +462,8 @@ public class IndexWriter
         }
       };
 
-  private final IndexWriterRAMManager.PerWriterIndexWriterRAMManager indexWriterRAMManager;
+  /** The id that is associated with this writer for {@link IndexWriterRAMManager} */
+  public final int ramManagerId;
 
   /**
    * Expert: returns a readonly reader, covering all committed as well as un-committed changes to
@@ -1213,9 +1214,7 @@ public class IndexWriter
         writeLock = null;
       }
     }
-    this.indexWriterRAMManager =
-        new IndexWriterRAMManager.PerWriterIndexWriterRAMManager(
-            this, config.getIndexWriterRAMManager());
+    this.ramManagerId = config.indexWriterRAMManager.registerWriter(this);
   }
 
   /** Confirms that the incoming index sort (if any) matches the existing index sort (if any). */
@@ -1370,7 +1369,7 @@ public class IndexWriter
    */
   @Override
   public void close() throws IOException {
-    indexWriterRAMManager.removeWriter();
+    config.indexWriterRAMManager.removeWriter(ramManagerId);
     if (config.getCommitOnClose()) {
       shutdown();
     } else {
@@ -2451,7 +2450,7 @@ public class IndexWriter
     // Ensure that only one thread actually gets to do the
     // closing, and make sure no commit is also in progress:
     if (shouldClose(true)) {
-      indexWriterRAMManager.removeWriter();
+      config.indexWriterRAMManager.removeWriter(ramManagerId);
       rollbackInternal();
     }
   }
@@ -6019,7 +6018,7 @@ public class IndexWriter
       seqNo = -seqNo;
       processEvents(true);
     }
-    indexWriterRAMManager.flushIfNecessary(config.flushPolicy);
+    config.flushPolicy.flushRamManager(this);
     return seqNo;
   }
 
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -2451,6 +2451,7 @@ public class IndexWriter
     // Ensure that only one thread actually gets to do the
     // closing, and make sure no commit is also in progress:
     if (shouldClose(true)) {
+      indexWriterRAMManager.removeWriter();
       rollbackInternal();
     }
   }
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -462,6 +462,8 @@ public class IndexWriter
         }
       };
 
+  private final IndexWriterRAMManager.PerWriterIndexWriterRAMManager indexWriterRAMManager;
+
   /**
    * Expert: returns a readonly reader, covering all committed as well as un-committed changes to
    * the index. This provides "near real-time" searching, in that changes made during an IndexWriter
@@ -939,11 +941,14 @@ public class IndexWriter
    * @param d the index directory. The index is either created or appended according <code>
    *     conf.getOpenMode()</code>.
    * @param conf the configuration settings according to which IndexWriter should be initialized.
+   * @param indexWriterRAMManager The RAM manager used for multi-tenant RAM management
    * @throws IOException if the directory cannot be read/written to, or if it does not exist and
    *     <code>conf.getOpenMode()</code> is <code>OpenMode.APPEND</code> or if there is any other
    *     low-level IO error
    */
-  public IndexWriter(Directory d, IndexWriterConfig conf) throws IOException {
+  public IndexWriter(
+      Directory d, IndexWriterConfig conf, IndexWriterRAMManager indexWriterRAMManager)
+      throws IOException {
     enableTestPoints = isEnableTestPoints();
     conf.setIndexWriter(this); // prevent reuse by other instances
     config = conf;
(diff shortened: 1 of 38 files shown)


lucene/core/src/java/org/apache/lucene/index/IndexWriterRAMManager.java
```diff
@@ -0,0 +1,116 @@
...
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * For managing multiple instances of {@link IndexWriter} sharing the same buffer (configured by
+ * {@link IndexWriterConfig#setRAMBufferSizeMB})
```
jpountz, line 27: It should be the other way around in my opinion, the RAM buffer size should be on `IndexWriterRAMManager`, and setting a `ramBufferSizeMB` on `IndexWriterConfig` would internally create a new `IndexWriterRAMManager` under the hood that is shared with no other `IndexWriter`.

lucene/core/src/java/org/apache/lucene/index/IndexWriterRAMManager.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+    if (totalRam >= config.getRAMBufferSizeMB() * 1024 * 1024) {
+      IndexWriter writerToFlush = chooseWriterToFlush(idToWriter.values(), idToWriter.get(id));
+      writerToFlush.flushNextBuffer();
+    }
+  }
+
+  /**
+   * Chooses which writer should be flushed. Default implementation chooses the writer with most RAM
+   * usage
```
jpountz, line 75: FWIW we ran benchmarks in Elasticsearch, and the approach that worked the best was to flush IndexWriters in a round-robin fashion. This is not intuitive at first sight, but I believe that it works better in practice because it is more likely to flush DWPTs that are little used, and also because otherwise you favor IndexWriters that do little indexing over IndexWriters that do heavy indexing (and are thus more likely to have a large buffer).

jpountz (commented):
Thanks for looking into it. It feels like there should be closer integration between this and the existing `FlushPolicy`/`FlushByRamOrCountsPolicy`?

I think we'll also want to look into the performance overhead of this. It's likely not good to iterate over tens of IndexWriters on every indexed document to check if the total buffer usage is above the limit?

mdmarshmallow:
> Thanks for looking into it. It feels like there should be closer integration between this and the existing `FlushPolicy`/`FlushByRamOrCountsPolicy`?

So I actually had the same thought, but when I took a look at `FlushPolicy`, it seemed very geared towards DW/DWPT, IE:
```
  public abstract void onChange(
      DocumentsWriterFlushControl control, DocumentsWriterPerThread perThread);
```
I couldn't think of a clean way to integrate the two... but I'll give it some more thought

> I think we'll also want to look into the performance overhead of this. It's likely not good to iterate over tens of IndexWriters on every indexed document to check if the total buffer usage is above the limit?

Ah I thought that the `ramBytesUsed()` call was cheap but looks like `AtomicLong#get()` is called under the hood so I'll see if there's a better way to do this. I agree though that performance testing this change makes sense - I haven't quite thought about how to do it though

lucene/core/src/java/org/apache/lucene/index/IndexWriterRAMManager.java
mdmarshmallow: Sorry I'm probably missing something here - so I get what you're saying about having the `IndexWriterRAMManager` in the `IndexWriterConfig`, but what would be the point of creating an `IndexWriterRAMManager` for a single `IndexWriter`? Wouldn't `DocumentWriterFlushControl` be sufficient for this case?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -939,11 +941,14 @@ protected final void ensureOpen() throws AlreadyClosedException {
...
    *     conf.getOpenMode()</code>.
    * @param conf the configuration settings according to which IndexWriter should be initialized.
+   * @param indexWriterRAMManager The RAM manager used for multi-tenant RAM management
    * @throws IOException if the directory cannot be read/written to, or if it does not exist and
    *     <code>conf.getOpenMode()</code> is <code>OpenMode.APPEND</code> or if there is any other
    *     low-level IO error
    */
-  public IndexWriter(Directory d, IndexWriterConfig conf) throws IOException {
+  public IndexWriter(
+      Directory d, IndexWriterConfig conf, IndexWriterRAMManager indexWriterRAMManager)
```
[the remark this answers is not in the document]
mdmarshmallow: Sure that makes sense to me

lucene/core/src/java/org/apache/lucene/index/IndexWriterRAMManager.java
mdmarshmallow: Hmm that's interesting! I can make that the default implementation then

jpountz:
> I couldn't think of a clean way to integrate the two... but I'll give it some more thought

For what it's worth, these classes are package-private, so we can feel free to change their API.

lucene/core/src/java/org/apache/lucene/index/IndexWriterRAMManager.java
```diff
@@ -0,0 +1,116 @@
...
+   *
+   * @param writers list of writers registered with this {@link IndexWriterRAMManager}
+   * @param callingWriter the writer calling {@link IndexWriterRAMManager#flushIfNecessary}
+   * @return the IndexWriter to flush
+   */
+  protected static IndexWriter chooseWriterToFlush(
+      Collection<IndexWriter> writers, IndexWriter callingWriter) {
```
vigyasharma, line 82: Why do we need the calling writer here?

lucene/core/src/java/org/apache/lucene/index/IndexWriterRAMManager.java
```diff
@@ -0,0 +1,116 @@
...
+    idToWriter.remove(id);
+  }
+
+  private void flushIfNecessary(int id) throws IOException {
+    if (idToWriter.containsKey(id) == false) {
+      throw new IllegalArgumentException(
+          "Writer " + id + " has not been registered or has been removed already");
+    }
+    long totalRam = 0L;
+    for (IndexWriter writer : idToWriter.values()) {
```
vigyasharma, line 64: Would it make sense to cache `ramByesUsed()` and only update it for the calling writer?

lucene/core/src/java/org/apache/lucene/index/IndexWriterRAMManager.java
mdmarshmallow: Hmm I took a slightly different approach, will publish a new PR soon. Maybe we can discuss it there but pretty much I just do what @jpountz suggested and move the `ramBufferSizeMB` to be a value held by the `IndexWriterRAMManager`. I think we can then discuss if we should just disable calling writer flushes in the manager if there is only a single writer.

lucene/core/src/java/org/apache/lucene/index/IndexWriterRAMManager.java
mdmarshmallow: I don't use it but I left it there in case someone wants to override the `FlushPolicy` in such a way that they might need the calling writer, IE: just flush the calling writer every time

lucene/core/src/java/org/apache/lucene/index/IndexWriterRAMManager.java
mdmarshmallow: Yeah this makes sense, I didn't realize that `ramBytesUsed()` is not a necessarily cheap operation. Changed it in the next rev

mdmarshmallow:
Thanks for all the comments guys, I've been pretty busy with some life things/work, but hopefully I'll put out another update by tomorrow!

lucene/core/src/java/org/apache/lucene/index/IndexWriterRAMManager.java
```diff
@@ -0,0 +1,217 @@
...
+    void flushIfNecessary(FlushPolicy flushPolicy) throws IOException {
+      manager.flushIfNecessary(flushPolicy, this);
+    }
+
+    long getTotalBufferBytesUsed() {
+      return manager.updateAndGetCurrentBytesUsed(id);
+    }
+  }
+
+  private static class LinkedIdToWriter {
```
vigyasharma, line 116: How about using a java Queue implementation instead of the custom linked-list logic? You could round-robin on elements by removing, processing and add them back to the queue.

I suppose this queue size would be small, so array deque and linked lists are both fine? We can also get some thread safe implementations out of the box.

lucene/core/src/java/org/apache/lucene/index/IndexWriterRAMManager.java
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  private long updateAndGetCurrentBytesUsed(int id) {
+    return idToWriter.getTotalRamTracker(id);
+  }
+
+  /**
+   * For use in {@link IndexWriter}, manages communication with the {@link IndexWriterRAMManager}
+   */
+  public static class PerWriterIndexWriterRAMManager {
```
vigyasharma, line 93: Do we really need this class? `IndexWriter` (and `FlushPolicy` too) should already have a reference to the ram manager via `IndexWriterConfig`.

How about we just store the writer's "ramManagerID" inside `IndexWriter`, and use it to invoke ram manager APIs directly? In fact, can we work with the writer object directly instead of keeping these "id" mappings? Similar to how FlushPolicy accepts the calling DWPT in its APIs...

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
```diff
@@ -142,7 +142,21 @@ public IndexWriterConfig() {
    * problem you should switch to {@link LogByteSizeMergePolicy} or {@link LogDocMergePolicy}.
    */
   public IndexWriterConfig(Analyzer analyzer) {
-    super(analyzer);
+    this(analyzer, new IndexWriterRAMManager(IndexWriterConfig.DEFAULT_RAM_BUFFER_SIZE_MB));
```
vigyasharma, line 145: Any reason for not making this change in the default constructor? We could avoid making changes to all the tests.

vigyasharma (commented):
Sorry it took so long to get to this PR. I have some code refactoring comments that should help simplify the changes.

lucene/core/src/java/org/apache/lucene/index/IndexWriterRAMManager.java
mdmarshmallow: Hmm, so I looked into using both `Queue` and `LinkedHashMap` to do this. The issue I found was that there was still complexity in maintaining the last id that was flushed and the total ram used, so the only function that really became less complex was the `addWriter` function. I think given that this probably won't simplify the overall function too much, I'm inclined the keep the implementation the same way it is right now, but if you disagree feel free to let me know and I can take a second look at it.

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
mdmarshmallow: I think the reason I had to make all the changes in the test was cause I added. this constructor:

`  public IndexWriterConfig(IndexWriterRAMManager indexWriterRAMManager) {`

And then in the tests there were a bunch of `new IndexWriterConfig(null)` calls which became ambiguous. I think that this constructor is potentially useful which is why I took the hit and changed all those tests, but I can remove it to avoid all those test changes?

lucene/core/src/java/org/apache/lucene/index/IndexWriterRAMManager.java
mdmarshmallow: Yeah I think that makes sense, I'll remove it

mdmarshmallow:
Just ran some benchmarks from `luceneutil` and saw a pretty significant slow down in indexing throughput (21 GB/hour -> 16 GB/hour)... trying to figure out why

mikemccand:
> and for some reason the baseline run is much faster

Ugh ... "who goes first" bias?  Maybe the line docs files was cold on your first run?  Though it's surprising that'd make such a big difference overall (21 -> 16) since that usage (reading single file sequentially) is normally well optimized by OS readahead ...

Could you please open a `luceneutil` issue?  Let's get to the bottom of this ... we recently fixed a similar "who goes first" bias (WGF bias?) but for the searcher tasks specifically I think.

mdmarshmallow:
Sure @mikemccand , added an issue [here](https://github.com/mikemccand/luceneutil/issues/320).

I also was able to test my change by just doing two separate runs and manually changing the test to point at whichever version I needed. My test setup involved hacking `Indexer#java` in `luceneutil` a bit to run 2 index writers instead of one. For mainline, I gave both index writers a 100 MB buffer, and for the candidate, I gave the ram manager a 200 MB buffer and passed that into the index writers instead. Both index writers were indexing the exact same dataset (`wikimedium1m`). I found that there was pretty much no change in indexing throughput (20.5 GB/hour for the candidate and 20.1 GB/hour for baseline). I can post more detailed results if anyone is interested.

I realize that this is not a very realistic scenario, so I will try benchmarking with more writers and also skew the indexing rates of each writer to see how the change performs.

mikemccand:
@mdmarshmallow thank you for benchmarking this change!  I know that is tricky and `luceneutil` is not exactly simple to use even when you don't have to make code changes ;)

I think it's fine that we can't measure a performance gain?  Achieving the same performance as N (2 or 5 in your tests) fully separate `IndexWriter`s is great ([Hippocratic Oath: "first do no harm"](https://en.wikipedia.org/wiki/Hippocratic_Oath))!  The goal here is to provide a simple way for users to dynamically share a limited RAM buffer with N `IndexWriter`s...

The only case I'd expect to see a performance win is if the rate of indexing is very unbalanced across the five shards ... e.g. you have five`IndexWriter`s but only one of them is actively indexing at red-line (saturate CPU) throughput.  With your change, that IW should be able to use (nearly) all of the RAM buffer, but prior to your change users had to do something coarse like divide RAM buffer statically up-front by five.

vigyasharma:
> The only case I'd expect to see a performance win is if the rate of indexing is very unbalanced across the five shards

+1, perhaps we could increase the number of index writers and/or the skew across their indexing workload?

I do agree that we should not block this change on an indexing performance win. It's isolated, relatively simple, and provides a meaningful lever to handle skewed indexing workloads without affecting existing setups.

PR #13956, opened 2024-10-24
Add MIGRATE entry about the fact that readVLong() may now read negative values, and up to 10 bytes.

jpountz:
This may be of interest for custom `DataInput`/`IndexInput` implementations that extend `readVLong()`.


ChrisHegarty (approved):
Thanks @jpountz. LGTM

uschindler (approved):

Issue #13805, opened 2024-09-18
TestLucene90DocValuesFormat fails with ArrayIndexOutOfBoundsException


ChrisHegarty:
```
ERROR: The following test(s) have failed:
  - org.apache.lucene.codecs.lucene90.TestLucene90DocValuesFormat.testSparseDocValuesVsStoredFields (:lucene:core)
    Test output: /opt/buildkite-agent/builds/bk-agent-prod-gcp-1726674638633683811/elastic/apache-lucene-nightly/lucene/core/build/test-results/test/outputs/OUTPUT-org.apache.lucene.codecs.lucene90.TestLucene90DocValuesFormat.txt
    Reproduce with: gradlew :lucene:core:test --tests "org.apache.lucene.codecs.lucene90.TestLucene90DocValuesFormat.testSparseDocValuesVsStoredFields" -Ptests.jvms=12 -Ptests.jvmargs= -Ptests.seed=175AB2293A24B66E -Ptests.nightly=true -Ptests.gui=true -Ptests.file.encoding=UTF-8 -Ptests.vectorsize=128 -Ptests.forceintegervectors=true
  - org.apache.lucene.backward_codecs.lucene80.TestBestSpeedLucene80DocValuesFormat.testSparseDocValuesVsStoredFields (:lucene:backward-codecs)
    Test output: /opt/buildkite-agent/builds/bk-agent-prod-gcp-1726674638633683811/elastic/apache-lucene-nightly/lucene/backward-codecs/build/test-results/test/outputs/OUTPUT-org.apache.lucene.backward_codecs.lucene80.TestBestSpeedLucene80DocValuesFormat.txt
    Reproduce with: gradlew :lucene:backward-codecs:test --tests "org.apache.lucene.backward_codecs.lucene80.TestBestSpeedLucene80DocValuesFormat.testSparseDocValuesVsStoredFields" -Ptests.jvms=12 -Ptests.jvmargs= -Ptests.seed=175AB2293A24B66E -Ptests.nightly=true -Ptests.gui=true -Ptests.file.encoding=UTF-8 -Ptests.vectorsize=128 -Ptests.forceintegervectors=true
```

```
 >     java.lang.ArrayIndexOutOfBoundsException: Index 3 out of bounds for length 3
   >         at __randomizedtesting.SeedInfo.seed([175AB2293A24B66E:43816E3AC0A89021]:0)
   >         at org.apache.lucene.util.packed.Packed64.get(Packed64.java:80)
[...]

Issue #13685, opened 2024-08-26
Get field number from the SegmentReadState when get norms


bugmakerrrrrr:
### Description

One of our customers are using Elasticsearch in their own IDC, their cluster has been updated with many versions, including almost all versions from 7.2.0 to 8.9.0, corresponding to Lucene versions 8.0.0 to 9.7.0. Recently, they found that some of the indices would fail to merge. The error stack is as follows:

![image](https://github.com/user-attachments/assets/8881f767-ac7c-4064-ae31-91c471214d8a)

After investigation, it was found that the reason is the inconsistency between the `SegmentReader.fieldInfos` and the `SegmentCoreReader.coreFieldInfos`. The fieldInfo passed to `Lucene80NormsProducer#getNorms` method is from `SegmentReader` and the `Lucene80NormsProducer.norms` is constructed from `SegmentCoreReader.coreFieldInfos`. Currently it is not clear what caused this, but normal querying and reading are not affected. If anyone else in the community has encountered the same problem, welcome to discuss.

Then I imitated the way of getting values from `PointsReader`，that is, getting the field number from the `SegmentReadState`, and found that the merge can be executed successfully.

https://github.com/apache/lucene/blob/4e3945ed548f40f31498dddcdce611c5cbbec245/lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90PointsReader.java#L135-L136

I wonder if there is any reason why we don't get field number from `SegmentReadState` for norms just like points. Can we align the way of obtaining field number to avoid some rare error like this(maybe)?

PR #13961, opened 2024-10-28
replace Map<String,Object> with IntObjectHashMap for DV producer

bugmakerrrrrr:
### Description

Add back #13686, which is reverted previously due to #13805.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/SlowCompositeCodecReaderWrapper.java
+++ b/lucene/core/src/java/org/apache/lucene/index/SlowCompositeCodecReaderWrapper.java
@@ -20,8 +20,10 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import org.apache.lucene.codecs.DocValuesProducer;
 import org.apache.lucene.codecs.FieldsProducer;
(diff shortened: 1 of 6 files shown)


lucene/core/src/java/org/apache/lucene/index/SlowCompositeCodecReaderWrapper.java
```diff
@@ -377,7 +376,7 @@ private static class SlowCompositeDocValuesProducerWrapper extends DocValuesProd
     private final CodecReader[] codecReaders;
     private final DocValuesProducer[] producers;
     private final int[] docStarts;
-    private final Map<String, OrdinalMap> cachedOrdMaps = new HashMap<>();
+    private final IntObjectHashMap<OrdinalMap> cachedOrdMaps = new IntObjectHashMap<>();
```
jpountz, line 379: I would rather keep this one keyed on the String field name since there's some ambiguity about whether it doc values sohuld be retrieved by segment field infos or by merged field infos.

jpountz (changes requested):
I would rather keep `SlowCompositeReaderWrapper` on the field name than field number, otherwise LGTM. Can you add an entry to lucene/CHANGES.txt?

lucene/core/src/java/org/apache/lucene/index/SlowCompositeCodecReaderWrapper.java
bugmakerrrrrr: Reverted. Btw, I wonder if we should change the DocValuesProducer/NormsProducer API to take a String rather than a FieldInfo right now as mentioned in [issue#13805](https://github.com/apache/lucene/issues/13805#issuecomment-2360784108) and [issue#13685](https://github.com/apache/lucene/issues/13685).

jpountz (approved):

Issue #13447, opened 2024-06-03
Explore bypassing HNSW graph building for tiny segments


benwtrent:
### Description

For both quantized, and non-quantized vectors, if there are many flushes with tiny segments, it may not be worth building the HNSW graph.

For example, we already do a brute-force query against the segment if the `k` requested is larger than the number of vectors in the segment. So, having the graph means nothing at query time and is unnecessary work.

In my mind, this would behave as follows (given some empirically determined threshold N):

 - For tiny segments, we would only scalar quantize, or just store in a flat index (if its a non-quantizing format)
 - When merging tiny segments, we determine if the total number of vectors is > N, and if so, we start building the graph
 - During indexing, we don't construct the HNSW graph builder until we have N vectors, then we construct the builder, replay the seen vectors, and handle new docs. 

N needs to be determined empirically, the HNSW codecs already have so many configuration items, it would be very nice to simply pick one where the trade-off makes most sense. I would suspect it is somewhere around `10k`. 

Also, would N take into account if Panama Vector is enabled and if things are quantized or not?

Possibly, we may want to consider the scenario where Panama Vector is disabled, though, I am not sure I would suggest anyone using vector search with it off :/.

jpountz:
> Also, would N take into account if Panama Vector is enabled and if things are quantized or not?

FWIW I'd optimize for simplicity over picking the perfect heuristic. As an example, `IndexOrDocValues` applies an arbitrary 8x penalty to doc values and only uses doc values when it would result in 8x fewer comparisons than running the query against the points index. This 8x number is made up, it only tries to account for the fact that comparisons are a bit cheaper on the points index than on doc values. But I like that it's simple and still makes good decisions more often than wrong decisions.

Maybe we could do something similar here and only build a HNSW graph if doing a top-1000 search would visit less than 1/8th of the documents that have a vector.

jpountz:
> Maybe we could do something similar here and only build a HNSW graph if doing a top-1000 search would visit less than 1/8th of the documents that have a vector.

It looks like the new `expectedVisitedNodes` heuristic that you introduced in #14836 could be used for that purpose?

jpountz:
> I would suspect it is somewhere around 10k.

Interestingly, it looks like your intuition roughly aligns with my suggestion if using topK=100. `expectedVisitedNodes(100, 10_000) = 921 ~= 1250 = 10_000 / 8`

benwtrent:
[...]

PR #13964, opened 2024-10-29
Account for 0 graph size when initializing HNSW graph

mayya-sharipova:
When initializing a joint graph from one of the segments' graphs,
 we always assume that a segment's graph is present. But later we want
 to explore an option where some segments will not have graphs (#13447).

This change allows to account for missing graphs.


john-wagster (approved):
Right off this LGTM

Do we not need any additional tests for when graph size is zero? or is this covered by random tests but didn't surface until now?

mayya-sharipova:
@john-wagster Thanks for the review. I tried to write tests, but it needs a lot of setup and mocks, and I thought it does't worth.

But I plan to write integration kind of test that will cover the changed part as a part of #13447

PR #13958, opened 2024-10-25
Speed up advancing within a block, take 2.

jpountz:
PR #13692 tried to speed up advancing by using branchless binary search, but while this yielded a speedup on my machine, this yielded a slowdown on nightly benchmarks.

This PR tries a different approach using vectorization. Experimentation suggests that it slows down a bit queries when advancing often goes to the very next doc ID, such as term queries and `OrHighNotXXX` tasks. But it speeds up queries that advance to the next few doc IDs, such as `AndHighHigh`. I think that this is a good trade-off since it slows down some plenty fast queries in exchange for a speedup with some more expensive queries.

Here is a `luceneutil` run on `wikibigall` with `-searchConcurrency 0`:

```
                            TaskQPS baseline      StdDevQPS my_modified_version      StdDev                Pct diff p-value
                   OrHighNotHigh      302.78      (2.4%)      283.75      (2.9%)   -6.3% ( -11% -   -1%) 0.000
                    OrHighNotMed      384.69      (3.0%)      363.33      (2.8%)   -5.6% ( -10% -    0%) 0.000
                         MedTerm      564.86      (2.2%)      537.04      (3.5%)   -4.9% ( -10% -    0%) 0.000
                         LowTerm     1014.02      (2.2%)      967.37      (3.6%)   -4.6% ( -10% -    1%) 0.000
                    OrHighNotLow      446.38      (3.4%)      427.10      (3.3%)   -4.3% ( -10% -    2%) 0.000
                        HighTerm      485.41      (1.9%)      464.49      (3.2%)   -4.3% (  -9% -    0%) 0.000
                   OrNotHighHigh      229.78      (2.4%)      221.51      (3.1%)   -3.6% (  -8% -    1%) 0.000
                    OrNotHighMed      396.63      (2.7%)      382.41      (3.1%)   -3.6% (  -9% -    2%) 0.000
                         Prefix3      145.65      (3.6%)      142.39      (3.7%)   -2.2% (  -9% -    5%) 0.051
                          IntNRQ      158.04      (4.7%)      154.77      (5.6%)   -2.1% ( -11% -    8%) 0.205
                       CountTerm     8320.96      (3.2%)     8198.56      (4.7%)   -1.5% (  -9% -    6%) 0.246
                        PKLookup      273.35      (3.6%)      269.71      (5.2%)   -1.3% (  -9% -    7%) 0.345
                        Wildcard       83.30      (3.4%)       82.28      (3.1%)   -1.2% (  -7% -    5%) 0.234
[...]

PR #13969, opened 2024-10-31
minor javadoc correction on VectorUtilSupport#findNextGEQ

gsmiller:


gsmiller:
@jpountz was looking through #13958 retroactively to understand the change and I _think_ I spotted a small javadoc error. Can you take a peek? Even though this is super trivial, I wanted to check with you prior to merging to make sure I'm not missing something. Thanks!

jpountz:
I wrote the javadocs this way on purpose, so that it would still work to create an IntVector/LongVector that starts before `from` and count the number of values that are less than the target. E.g. something like this:

```java
if (length >= LONG_SPECIES.length() && length - from < LONG_SPECIES.length()) {
  // less than LONG_SPECIES.length() doc IDs
  LongVector vector = LongVector.fromArray(LONG_SPECIES, values, length - LONG_SPECIES.length());
  VectorMask<Long> mask = vector.compare(VectorOperators.LT, target);
  return length - LONG_SPECIES.length() + mask.trueCount();
} else {
  // other cases
}
```

The current implementation doesn't take advantage of it, so I don't mind removing it, we could add it back later on if we want to take advantatge of it since it's an internal API.

gsmiller:
@jpountz thanks for the clarification! Let's not make the javadoc change then since it sounds like a reasonable reason to keep the requirement that values are sorted beginning at index `0` and not `from`.  (We could always change it later if it seemed like there was a useful reason to not require values `[0, from]` to be sorted).

gsmiller:
@jpountz after thinking a little more, I wonder if an `assert` would make sense to guard against unsorted data between index `0` and `from`? Probably quite unlikely, but would also be nice if that use-case tripped an assert now instead of silently working and then failing later because it didn't adhere to the contract outlined in the javadoc? We could do something like `assert IntStream.range(0, length - 1).noneMatch(i -> buffer[i] > buffer[i + 1]);`. It's trivial but I'm happy to add this if you think it would be reasonable.

jpountz:
This sounds good to me, maybe extract it to a function to avoid increasing the method size too much? Now that you made me look harder at this code, I'm also considering renaming `length` to `to` since `length` usually is a number of entries after `from` while it's an absolute end offset here.

gsmiller:
> I'm also considering renaming length to to since length usually is a number of entries after from while it's an absolute end offset here

+1. I noticed this as well when writing the assertion.

gsmiller:
Added an assertion and tweaked parameter naming as discussed here over in #13972.

PR #13970, opened 2024-10-31
Simplify codec setup in vector-related tests.

jpountz:
Many of vector-related tests set up a codec manually by extending the current codec. This makes bumping the current codec a bit painful as all these files need to be touched. This commit migrates to `TestUtil#alwaysKnnVectorsFormat`, similarly to what we do for postings and doc values.


benwtrent (approved):

tteofili (approved):
nice :)

PR #13971, opened 2024-11-01
Revert "Disjunction as CompetitiveIterator for numeric dynamic pruning (#13221)"

javanna:
This reverts commit 1ee4f8a1115d1de623f242014681032d87ed2c1e.

We have observed performance regressions that can be linked to #13221. This needs further work before it can be pushed to main and branch_10x

Relates to #13856

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/comparators/NumericComparator.java
+++ b/lucene/core/src/java/org/apache/lucene/search/comparators/NumericComparator.java
@@ -18,9 +18,6 @@
 package org.apache.lucene.search.comparators;
 
 import java.io.IOException;
-import java.util.ArrayDeque;
-import java.util.Deque;
-import java.util.function.Consumer;
 import org.apache.lucene.index.DocValues;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.LeafReaderContext;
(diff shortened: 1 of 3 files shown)


lucene/core/src/java/org/apache/lucene/util/IntArrayDocIdSet.java
```diff
@@ -34,23 +34,15 @@ public final class IntArrayDocIdSet extends DocIdSet {
...
-   * Build an IntArrayDocIdSet by an int array and len.
-   *
-   * @param docs A docs array whose length need to be greater than the param len. It needs to be
-   *     sorted from 0(inclusive) to the len(exclusive), and the len-th doc in docs need to be
-   *     {@link DocIdSetIterator#NO_MORE_DOCS}.
-   * @param len The valid docs length in array.
-   */
-  public IntArrayDocIdSet(int[] docs, int len) {
-    if (docs[len] != DocIdSetIterator.NO_MORE_DOCS) {
+  public IntArrayDocIdSet(int[] docs, int length) {
```
javanna, line 37: @jpountz this class used to be package private. With the revert that this PR applies to main, `TestBooleanOr` would be the only caller of this constructor outside of `o.a.l.util`, requiring it to be public. That looks odd, any chance we could update `TestBooleanOr` to not rely on this class being public?

javanna:
@gf2121 pinging you explicitly because you worked on #13221. Sadly, we observed some severe regressions caused by it, and we were not able to somehow adjust the updated algorithm to avoid those. We already reverted that commit in branch_10_0 for the Lucene 10.0 release,  and I hereby propose to do the same on main and branch_10x until we manage to fill the gap around the search after regressions.

lucene/core/src/java/org/apache/lucene/util/IntArrayDocIdSet.java
jpountz: Could you add a pkg-private copy of it next to `TestBooleanOr` maybe?

javanna: Done.

PR #13972, opened 2024-11-01
A few small tweaks to VectorUtil#findNextGEQ

gsmiller:
1. Rearrange/rename the parameters to be more idiomatic (e.g., follow conventions of Arrays#... methods)
2. Add assert to ensure expected sortedness we may rely on in the future (so we're not trappy)
3. Migrate PostingsReader to call VectorUtil instead of VectorUtilSupport (so it benefits from the common assert)

### DIFF
--- a/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene912/Lucene912PostingsReader.java
+++ b/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene912/Lucene912PostingsReader.java
@@ -593,7 +593,7 @@ public final class Lucene912PostingsReader extends PostingsReaderBase {
         }
       }
 
-      int next = findNextGEQ(docBuffer, docBufferSize, target, docBufferUpto);
+      int next = findNextGEQ(docBuffer, target, docBufferUpto, docBufferSize);
       this.doc = (int) docBuffer[next];
       docBufferUpto = next + 1;
       return doc;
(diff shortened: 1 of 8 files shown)


lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene912/Lucene912PostingsReader.java
```diff
@@ -1755,13 +1755,13 @@ static long readVLong15(DataInput in) throws IOException {
     }
   }
 
-  private static int findNextGEQ(long[] buffer, int length, long target, int from) {
-    for (int i = from; i < length; ++i) {
+  private static int findNextGEQ(long[] buffer, long target, int from, int to) {
```
gsmiller, line 1758: Made this change for consistency but don't have any strong feelings about it. If there's a preference to just leave it be, I'm fine reverting this file.

lucene/core/src/java/org/apache/lucene/codecs/lucene101/Lucene101PostingsReader.java
```diff
@@ -601,7 +599,7 @@ public int advance(int target) throws IOException {
         }
       }
 
-      int next = VECTOR_SUPPORT.findNextGEQ(docBuffer, docBufferSize, target, docBufferUpto);
+      int next = VectorUtil.findNextGEQ(docBuffer, target, docBufferUpto, docBufferSize);
```
gsmiller, line 602: @jpountz I made this change so that everything goes through `VectorUtil` and the `assert`. It also seemed a little more consistent with how we leverage `VectorUtil` vs. `VectorUtilSupport` elsewhere, but I'm not sure if you had a good reason for writing it the way you did. If there's a performance (or other) consideration here, I'd be happy to understand that better.

jpountz: I did it this way after doing something similar with PostingDecodingUtil, which didn't really belong to `VectorUtil`. I agree it's nicer the way you did it, I would not expect performance implications if inlining works properly, nightly benchmarks will tell us.

jpountz (approved):
Thank you!

PR #13974, opened 2024-11-03
SortedSet DV Multi Range query

mkhludnev:
### Description

Here's a sketch of DV analog of MultiRangeQuery. This draft uses bitset for expanding multiple ranges. 
It passes random test, but internals are rough yet.
What do you think about this idea overall? 

@atris, may I request your comments as an author of MultiRangeQuery?


lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
```diff
@@ -0,0 +1,300 @@
...
+            LongBitSet matchingOrdsShifted = null;
+            long minOrd = 0, maxOrd = values.getValueCount() - 1;
+            long matchesAbove =
+                values.getValueCount(); // it's last range goes to maxOrd, by default - no match
+            long maxSeenOrd = values.getValueCount();
+            TermsEnum.SeekStatus seekStatus = TermsEnum.SeekStatus.NOT_FOUND;
+            for (int r = 0; r < rangeClauses.size(); r++) {
+              MultiRangeQuery.RangeClause range = rangeClauses.get(r);
+              long startingOrd;
+              seekStatus = termsEnum.seekCeil(new BytesRef(range.lowerValue));
```
gsmiller, line 164: There's a fair amount of `BytesRef` creation going on here. Should we just store bytes ref instances in the ranges instead of storing byte arrays then converting here?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
```diff
@@ -0,0 +1,300 @@
...
+                new TwoPhaseIterator(values) {
+                  // TODO unwrap singleton?
+                  @Override
+                  public boolean matches() throws IOException {
+                    for (int i = 0; i < values.docValueCount(); i++) {
+                      long ord = values.nextOrd();
+                      if (ord >= finalMinOrd
+                          && ((finalMatchesAbove < values.getValueCount()
+                                  && ord >= finalMatchesAbove)
+                              || finalMatchingOrdsShifted.get(ord - finalMinOrd))) {
```
gsmiller, line 235: This approach of densely memorizing all ordinals that fall within the ranges has me a bit worried. This could potentially be a quite large bitset. I wonder if we could instead just keep track of all the ranges actually represented in the query and then just check against them for each doc ord? Like what we do in `SortedSetDocValuesRangeQuery` but keeping tracking of multiple min/max tuples to check against instead of only one? Is there a reason you didn't go with that approach and feel the dense bitset is necessary here?

gsmiller (commented):
Thanks @mkhludnev ! I haven't fully combed through this change but did an initial pass and wanted to share some thoughts since I meant to look at this a while ago and it slipped through the cracks. I appreciate you taking this on!

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
```diff
@@ -0,0 +1,300 @@
...
+            true,
+            true);
+      }
+      return new SortedSetMultiRangeQuery(name, clauses, this.bytes, comparator);
+    }
+  }
+
+  @Override
+  public Query rewrite(IndexSearcher indexSearcher) throws IOException {
+    ArrayList<MultiRangeQuery.RangeClause> sortedClauses = new ArrayList<>(this.rangeClauses);
```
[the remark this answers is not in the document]
mkhludnev: ook. let's move it to constructor.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
```diff
@@ -0,0 +1,300 @@
...
+import org.apache.lucene.search.Scorer;
+import org.apache.lucene.search.ScorerSupplier;
+import org.apache.lucene.search.TwoPhaseIterator;
+import org.apache.lucene.search.Weight;
+import org.apache.lucene.util.ArrayUtil;
+import org.apache.lucene.util.BytesRef;
+import org.apache.lucene.util.LongBitSet;
+
+/** A union multiple ranges over SortedSetDocValuesField */
+public class SortedSetMultiRangeQuery extends Query {
```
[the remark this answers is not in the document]
mkhludnev: Aghh right. Let me gather my considerations: 
 - do you mean to rename it to `SortedSetDocValuesMultiRangeQuery` or somehow abbreviate it? I'm totally for it. 
 - `SortedSetDocValuesRangeQuery` which is a singular analog of `SortedSetMultiRangeQuery`, has no `bytePerDim` constraint. However, I purpose `SortedSetMultiRangeQuery` for fixed width usages like IP, lately for IP ranges. I'm not sure it's possible at all, but do you think it's worth to drop `bytePerDim` constrain to move it forward?    
 - `sandbox` package is an only option for now, I suppose
 - is it worth to copy `MultiRangeQuery.RangeClause` here? I used it only because it's public. Whether we do it now or later, it doesn't leak from public interface. WDYT?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
```diff
@@ -0,0 +1,300 @@
...
+            for (int r = 0; r < rangeClauses.size(); r++) {
+              MultiRangeQuery.RangeClause range = rangeClauses.get(r);
+              long startingOrd;
+              seekStatus = termsEnum.seekCeil(new BytesRef(range.lowerValue));
+              if (matchingOrdsShifted == null) { // first iter
+                if (seekStatus == TermsEnum.SeekStatus.END) {
+                  return empty(); // no bitset yet, give up
+                }
+                minOrd = termsEnum.ord();
+                if (skipper != null) {
```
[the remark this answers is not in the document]
mkhludnev: Need to think through. Honestly, never see `DocValuesSkipper`, have only vague idea how to test it.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
```diff
@@ -0,0 +1,300 @@
...
+                  minOrd = Math.max(minOrd, skipper.minValue());
+                  maxOrd = Math.min(maxOrd, skipper.maxValue());
+                }
+                if (maxOrd < minOrd) {
```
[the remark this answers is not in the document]
mkhludnev: ok.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
mkhludnev: OK. this would be a reason to copy Range here.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
mkhludnev: This was my first though too. However, then I've found [DocValuesRewriteMethod](https://github.com/apache/lucene/blob/aef16daa76b6efb354c216fbc1415fa63cd4c58c/lucene/core/src/java/org/apache/lucene/search/DocValuesRewriteMethod.java#L176) where there's the set over ordinals. So, if it's feasible for set of point values, why not do the same for set of ranges? 
Perhaps, it needs to set too many 1s for ranges which fully cover all ords? I dont know. 
If we think about Numeric or Binary DVs, there are no ords, and we have to search in the tree of ranges. Also, for me it's purposed for queries with more that 1000 ranges, because smaller sets can be handled via plain BooleanQuery.SHOULD. 
Anyway, I'm really open for suggestions.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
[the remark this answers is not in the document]
mkhludnev: @gsmiller How to test skipper logic via test like https://github.com/apache/lucene/pull/13974/files#diff-5c4a4738d4643b48f34fb00fa3b67098f9bf4f1744e12955372df0f32812670d ?

mkhludnev:
@gsmiller thanks for reviewing. Looking into!

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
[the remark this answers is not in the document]
gsmiller: Cool, thanks! Of course another option would be to do this in the builder if you think that's a better fit. No strong preference on my part.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
gsmiller: Right. As you point out, there are definitely other situations where we use this bitset approach. My only thought here is that we should be able to use less memory by keeping track of the range min/max values and not needing the dense representation (this isn't really as feasible with `DocValuesRewriteMethod` since it's encoding a set of unique terms that aren't particularly likely to be contiguous ranges). If we track ranges, we can take advantage of them being sorted and the doc-level ordinals also being sorted. A binary search approach might be a good fit, and we can keep pushing up the lower bound of the search based on the last range that matched (since doc-level ordinals are sorted). I suspect this would still be less efficient than the dense bitset approach, but I bet it wouldn't be that big of a difference.

Without a way to benchmark and test though, it's just a bit of a guess. I'm OK with keeping it the way you have it for now if that's your preference. We can always benchmark alternatives in the future and evolve this if we want. Should we do that for now?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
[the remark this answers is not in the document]
gsmiller: - I think the name is fine. Including `DocValues` in the naming is a little more consistent with the code base, but it's also pretty verbose (as you point out). I'm fine with either.
- Right. I don't think there is any point to having `bytesPerDim` here is there? That's needed in `MultiRangeQuery` since it needs to unpack multidimensional points, but you're not doing that here. This goes back to my initial misunderstanding of what was being proposed. `MultiRangeQuery` is a multi-range version of `PointRangeQuery`, which are both range queries over n-dimensional points indexed in a points index. To my knowledge, there isn't a "slow" version of these queries that operate over doc values. What I initially thought was being proposed was a doc values version of these queries, but this is something different. As you point out, it's really a multi-range version of `SortedSetDocValuesRangeQuery`. So actually, as I'm writing this, I'm starting to think maybe naming this `SortedSetDocValuesMultiRangeQuery` might be good to make this connection a little more clear?
- Yeah, I think sandbox is still a good fit here. Let's not change that.
- I think it's OK to reuse the existing `RangeClause` for now since it's not part of the public API. As you point out, this would be easy enough to change later if we need to.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
```diff
@@ -0,0 +1,300 @@
...
+  @Override
+  public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost)
+      throws IOException {
+    return new ConstantScoreWeight(this, boost) {
+      @Override
+      public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOException {
+        if (context.reader().getFieldInfos().fieldInfo(field) == null) {
+          return null;
+        }
+        DocValuesSkipper skipper = context.reader().getDocValuesSkipper(field);
```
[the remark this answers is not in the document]
mkhludnev: Turns out, DV access should happen right here since `ScorerSupplier#cost()` needs DV for estimate.

Actually I copied it from 
https://github.com/apache/lucene/blob/0494c824e0ac8049b757582f60d085932a890800/lucene/core/src/java/org/apache/lucene/document/SortedSetDocValuesRangeQuery.java#L122

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
mkhludnev: I mostly agree. Thanks for sharing your considerations. Let me clarify your points:
> isn't really as feasible with DocValuesRewriteMethod  

Why this "set" query can't use a tree of unique values with log(n) access complexity? These queries might be quite close to each other. In the edge case ranges are quite narrow having 1 or 2 ordinals, thus SsDvMultyRange would be the same as TermsInSet. If ranges are wide, it definitely spends more time on setting 1s in the bitset, but maybe it's not lethal due to SIMD intrinsic (who knows) with a gain on search time per doc check. 
I agree it's a field for benchmarking and picking optimizations dynamically (later). Also, something like KD-tree might be applicable here.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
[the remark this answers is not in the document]
mkhludnev: Since I want to limit the scope of this PR to:
- single dimension fields, lately it might be extended to ranges in fields.
- fixed length values.
I'm going to hide constuctor and introduce public factory method.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
mkhludnev: > we can take advantage of them being sorted and the doc-level ordinals also being sorted.

Frankly speaking, I suppose singular values are more common than long sets of values, so, I don't think it get much gain for many users.

mkhludnev: > less memory by keeping track of the range min/max values and not needing the dense representation

I just realized, if we merge overlapping ranges (which turns not so easy, but feasible), for remaining non-overlapping ranges we always have smaller and greater neighbor, that means we can just search ranges by ordinal in `TreeSet`. 
Is it worth to explore instead of dense bit set?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/DocValuesMultiRangeQuery.java
```diff
@@ -0,0 +1,125 @@
...
+
+  /**
+   * Builder for creating a query for matching multi-range field values. Highlights two key points:
+   *
+   * <ul>
+   *   <li>treats multiple or single field value as scalar for range matching
+   *   <li>field values have fixed width
+   * </ul>
+   */
+  public static class SordedSetFieldValueFixedBuilder
```
mkhludnev, line 74: @gsmiller what do you think about such a verbose way of constructing query? Is it worth to piggyback on `java.function` nice interfaces?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
mkhludnev: > My only thought here is that we should be able to use less memory by keeping track of the range min/max values and not needing the dense representation 

Introduced the second implementation treeset instead of bitset. 
https://github.com/apache/lucene/blob/eea9e815c311b006d3297b8d9977fa6cb7c8e3f2/lucene/sandbox/src/java/org/apache/lucene/sandbox/search/DocValuesMultiRangeQuery.java#L198 
It seems doubtful, however if we need to match NumericsDVs this is the only option. 
How to scope this PR?

gsmiller: @mkhludnev apologies for missing this conversation. I was mostly away over the holidays at the end of the year and just catching up more now. I just took a quick look at the latest version and I wonder if we're making this more complex than it needs to be for an initial version? What do you think of going back to the bitset-based version and getting that merged first, then proposing/measuring possible optimizations? WDYT?

mkhludnev: Thanks @gsmiller. Moved range tree to https://github.com/mkhludnev/lucene/tree/dvOnly-multirangeQ-rangetree keep here the bitset only.  
The advantage of bitset impl is that it can handle a set of values and multi-range in the single query that should be common in SIEM (afaik) where we need a set of IPs and masks altogether.

mkhludnev:
open items from my POV:
 - improve testing via copying `TestDocValuesQueries`
 - may DVSkipper limit and reduce the given ranges before matching to limit bitset?
 - if ranges reaches ordinals boundaries: `[0 TO a],[b TO c],...[y TO dv.getValueCount()-1]` we can reduce bitset to `[a+1 TO y-1]`, and check boundary ranges explicitly.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/SortedSetMultiRangeQuery.java
gsmiller: Hey @mkhludnev trying to catch back up on some of the progress here. Looking through `SortedSetDocValuesMultiRangeQuery` in its current implementation, it seems like there are a couple different places you deal with overlapping ranges. There's `resolveOverlaps` at initialization time, then there's also logic when creating the scorer that handles the fact that ranges can overlap (i.e., in `createOrdRanges`). I'm trying to wrap my head around the need for both. Is there a reason we need to handle overlaps during initialization? It seems like we only really care about overlapping ranges so we can avoid re-setting the same bits in the bitset over-and-over right? So this is really an optimization for scorer setup right? In case there was some huge number of ranges provided that all happen to have a bunch of common overlap? Do I have the right understanding here, or am I missing some other important bit of this (trying to make sure I'm not misunderstanding your intent here...). Thanks!

mkhludnev: Thanks for your attention @gsmiller .
You are right.
`resolveOverlaps` literally merges overlapping ranges. Why it's necessary? I think just for some pathological queries. Imagine, many  distinct ranges covered by the huge one, it may drop covered ranges from the future processing whether it be bitset or a tree set approach. I'm open for any stronger arguments. 
Then, [createOrdRanges](https://github.com/apache/lucene/pull/13974/files#diff-7261e59f0cdf2536185e8e55d2fe1be04bbad37473b0eba30b4126326a289551R190) turns bytearrays to ordinals (btw, this step won't be available for NumericDVs and other w/o ords), at here some ranges might become adjacent, not really overlapping, but eligible for merging into bigger ones. I checked the code and comments there, there's no occurrence of `overlap` only `adjacent`.

gsmiller: >  I think just for some pathological queries. Imagine, many distinct ranges covered by the huge one, it may drop covered ranges from the future processing [...]

It makes sense to me that we may want to do some optimizations for pathological cases with many ranges that overlap. I do wonder a bit on the importance of this. My reasoning here is that the cost only hits when creating the bitset, which is an upfront cost and not a per-doc cost. I'm not opposed to pursuing this, but I'm trying to find a way to balance some of the added code complexity. In this case, it looks like you're relying on a couple helper classes (`OrdRange` and `Edge`) along with a reasonably complex algorithm for finding all the overlaps (which to be perfectly honest, I haven't dug into, only glanced at).

> [...] whether it be bitset or a tree set approach

I'm not sure I understand this point completely. I think what you're saying is that you might want to follow up this PR with another approach that uses a tree set approach instead of a bitset for matching (I believe you had an iteration of this PR that introduced the idea). I wonder if a common solution for merging overlapping ranges might become more important later if we really do have multiple implementations? Maybe this is a candidate for a spin-off issue that gets pursued as a later PR? (This goes back to my motivation of trying to reduce complexity of this initial change until it becomes more obvious we need the added complexity).

> [...] this step won't be available for NumericDVs and other w/o ords

Ah yes, that's fair. I think this is a continued discussion of the last point. Maybe we can pursue some merging when we do have multiple implementations that could obviously benefit from something like this? I can an optimization becoming more important in these other situations since it will become a per-doc cost as opposed to an up-front setup cost. Again though, I wonder if this could be tackled as a spin-off issue once we have more implementations?

> createOrdRanges [...]  at here some ranges might become adjacent, not really overlapping, but eligible for merging into bigger ones. I checked the code and comments there, there's no occurrence of overlap only adjacent.

Are you suggesting that `createOrdRanges` as you currently have it implemented would not merge ranges that overlap? It looks to me like it would, but I might be missing something. If the ranges are sorted by their `min`, then I believe this method will merge away all the overlaps (as well as adjacent) wouldn't it? What do you think? Maybe I'm missing some cases or not properly wrapping my head around what you've got implemented here?

Apologies if it feels like I'm pushing back on this too much. I'm not against pursuing these optimizations by any means, I'm just seeing if there's a way to simplify this PR so we can get it merged and follow up with spin-off issues. Please let me know what you think (and thanks for your patience!).

mkhludnev: ok. let's drop `resolveOverlaps`, `createOrdRanges`. Let's just loop ranges and call `seekCeil()` twice per range. Does it make sense to find min and max ords attempting to reduce bitset allocation upfront? 
If it does should it turn byte array ranges into ord ranges and store it for the second pass filling bitset?

gsmiller: Thanks for your thoughts @mkhludnev. I'd suggest one-of-two options. I'd be happy to move forward with either:
1. Keep it as simple as possible and drop all range merging, as you outline above (e.g., drop `resolveOverlaps` and `createOrdRanges`). Then consider adding these ideas back in as spin-off SIMs after the initial implementation is merged.
2. Drop `resolveOverlaps` but keep the logic in `createOrdRanges` as you currently have it. I like how you have this implemented currently. Seems like a pretty clean way of reducing unnecessary calls to `seekCeil` and bitset sets by recognizing adjacent or overlapping ranges without adding much code complexity.

Apologies on my above communication. It may not have been as specific as it could have been. The complexity that I'd like to avoid (at least in the initial version of this) is `resolveOverlaps`. I don't mind `createOrdRanges` since it is immediately applicable in this implementation and isn't particularly complex to reason about. The `resolveOverlaps` method seems more future-looking to other implementations, etc., so the value vs. complexity trade-off isn't as clear to me right now.

What do you think? (I'm happy to go with whatever you think will help get this merged and reduce headaches for you since we're discussed plenty at this point and I don't want to slow you down further )

gsmiller:
@mkhludnev took one more look and I think this is in good shape to merge. Would you mind adding a CHANGES entry?

mkhludnev:
Thanks. I'm happy to hear. Here's what I have to work on:
- @gsmiller what's your feeling about the [proposed API](https://github.com/apache/lucene/blob/c56caeb26a5af4b0afc5f2cb04a4f290ceff6189/lucene/sandbox/src/test/org/apache/lucene/sandbox/search/TestInetAddrSsDvMultiRange.java#L145). For me it seems a little bit cumbersome. I'm thinking about extending it further to other docvalues types and picking range tree algorithm instead of bitset.
- I'll put some more untitests and improve `toString()` somehow.
- I still don't know how to utilize DocValuesSkipper, can you give me a clue how to handle it properly?

gsmiller:
Thanks @mkhludnev . 

1. I'm not as worried about getting the API perfect with this initial commit. One benefit of doing this in `sandbox` is that we can iterate on this in breaking ways without being too concerned. So if you're planning to add more use-cases, I can imagine this evolving and getting iterated on over time. So I wasn't going to comment on the API too much right now since this is in `sandbox`.
2. More testing, and toString is great :)
3. I think what you're doing with skipper already is correct. It's possible you could do more, but I'd save that for a follow up. The skipper gives you the min and max values represented by a range of docs. So what you're essentially doing right now is saying that if a given doc falls into a skip interval that doesn't intersect with the full min-max range that covers all of your multi-ranges, then it cannot match any of your ranges so you can skip it without loading its doc value and checking explicitly. If you're unsure of what you're doing with skipper though, you could also remove it as an optimization for now and add it back in separately (e.g., if you're concerned it might be incorrect).

mkhludnev:
> Would you mind adding a CHANGES entry?

done

gsmiller (approved):
Looks good to me. I'll go ahead and get this merged. (I made one small change to move the changes entry under 10.2 instead of 10.1)

mkhludnev:
It would be interesting to extend it to Numerics DVs and ranges aiming IP/masks

mkhludnev:
one idea: 
 - optimize one seekCeil() call `if r.upper==r.lower || r.equals(r.lower)` TBC

mkhludnev:
>.. ranges aiming IP/masks

Turns out there are many varieties in range fields: single vs multiple values, number of dimensions and range search operations. Let's postpone it till the real life usecase.

PR #13981, opened 2024-11-08
[DRAFT] Change vector input from IndexInput to RandomAccessInput

### LINKED ISSUE #13938: Move vector search from IndexInput to RandomAccessInput
### Description

Vector search currently loads vectors from disk by issuing a `seek()` followed by a `readFloats()`. We should instead:
 - Add an absolute `readFloats()` method to `RandomAccessInput`
 - Refactor the latest vector search file format to use `RandomAccessInput` instead of `IndexInput` to read vectors from disk.

dungba88:
### Description

This PR changes the input of vector reader from IndexInput to RandomAccessInput, as we mostly random-access the vectors with seek() followed by readFloats().

One thing I'm uncertain is the need for cloning RandomAccessInput.

TODO:
- Add tests

fixes #13938

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/RandomAccessInput.java
+++ b/lucene/core/src/java/org/apache/lucene/store/RandomAccessInput.java
@@ -23,7 +23,7 @@ import org.apache.lucene.util.BitUtil; // javadocs
  * Random Access Index API. Unlike {@link IndexInput}, this has no concept of file position, all
  * reads are absolute. However, like IndexInput, it is only intended for use by a single thread.
  */
-public interface RandomAccessInput {
+public interface RandomAccessInput extends Cloneable {
 
   /** The number of bytes in the file. */
   long length();
@@ -46,6 +46,14 @@ public interface RandomAccessInput {
       bytes[offset + i] = readByte(pos + i);
     }
   }
+
+  /**
+   * Reads a specified number of floats starting at a given position into an array at the specified
+   * offset.
+   *
+   * @see DataInput#readFloats
+   */
+  void readFloats(long pos, float[] floats, int offset, int length) throws IOException;
 
   /**
    * Reads a short (LE byte order) at the given position in the file
@@ -77,4 +85,6 @@ public interface RandomAccessInput {
    * @see IndexInput#prefetch
    */
   default void prefetch(long offset, long length) throws IOException {}
+
+  Object clone();
 }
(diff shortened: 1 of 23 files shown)


lucene/core/src/java21/org/apache/lucene/internal/vectorization/Lucene99MemorySegmentByteVectorScorer.java
```diff
@@ -40,8 +41,14 @@ abstract sealed class Lucene99MemorySegmentByteVectorScorer
    * returned.
    */
   public static Optional<Lucene99MemorySegmentByteVectorScorer> create(
-      VectorSimilarityFunction type, IndexInput input, KnnVectorValues values, byte[] queryVector) {
+      VectorSimilarityFunction type,
+      RandomAccessInput slice,
+      KnnVectorValues values,
+      byte[] queryVector) {
     assert values instanceof ByteVectorValues;
+    if (!(slice instanceof IndexInput input)) {
```
shubhamvishu, line 49: Nit : input is not used

lucene/core/src/java/org/apache/lucene/store/RandomAccessInput.java
```diff
@@ -77,4 +85,6 @@ default void readBytes(long pos, byte[] bytes, int offset, int length) throws IO
    * @see IndexInput#prefetch
    */
   default void prefetch(long offset, long length) throws IOException {}
+
+  Object clone();
```
shubhamvishu, line 89: Change this to `RandomAccessInput clone();` so you don't have to cast in all places?

lucene/core/src/java21/org/apache/lucene/internal/vectorization/Lucene99MemorySegmentByteVectorScorer.java
dungba88: It's used in the next block though

```
input = FilterIndexInput.unwrapOnlyTest(input);
```

lucene/core/src/java/org/apache/lucene/store/RandomAccessInput.java
dungba88: This baffles me a lot! I can't let the return type to be `RandomAccessInput` because it will conflict with the `DataInput.clone()` which returns `DataInput`.

Adding the clone method seems a bit weird to me, but I'm not sure if there is a better way to go around it.

benwtrent: I am not sure we should add a `clone()` here to random access input. This is a wide ranging public interface in lucene. Adding something here should be carefully considered.

dungba88: I agree it's not ideal, but clone is needed for `copy()` method. If not, maybe we need to have a CloneableRandomAccessInput?

PR #13982, opened 2024-11-08
Remove duplicate test code

LuXugang:
### Description

The only diff between `doIterate1 ` and `doIterate2 ` is `bb < b.length()` in `doIterate1 `, but this condition seems always true, so should we remove it?


jpountz (approved):
I walked the history of this file to understand the original intention, but it's always been this way. LGTM

Issue #13724, opened 2024-09-05
CheckIndex could assert more things about the HNSW graph?


mikemccand:
### Description

I came across a spooky comment `// we see repeated neighbor nodes?!` in [this `luceneutil` PR](https://github.com/mikemccand/luceneutil/pull/290/files) and discussed with @msokolov and we realized `CheckIndex` tool could assert things about the HNSW graph like "outgoing neighbor nodes are in order" and "there are no duplicate neighbor nodes".

Maybe there are other "HNSW graph integrity" checks that `CheckIndex` could do?

benchaplin:
I've been interested in working on some HNSW stuff so I took a shot at this. I added only the two checks you mentioned to start, but I'm planning to look through some issues and have a think on what other checks could be useful.

PR #13984, opened 2024-11-08
Add some basic HNSW graph checks to CheckIndex

benchaplin:
### Description

This is a first pass on #13724 - I've added the following checks:
- verify neighbor list is in order 
- verify no duplicate neighbors

The code was adapted from [KnnGraphTester](https://github.com/mikemccand/luceneutil/blob/3a6235c2038de6c6d5d0575d8f58cf06c2836793/src/main/knn/KnnGraphTester.java#L521) in luceneutil.

Here's some test results from an index containing the cohere-wikipedia-docs-768d dataset:
```
    test: open reader.........OK [took 0.010 sec]
    test: check integrity.....OK [took 2.207 sec]
    test: check live docs.....OK [took 0.000 sec]
    test: field infos.........OK [2 fields] [took 0.000 sec]
    test: field norms.........OK [0 fields] [took 0.000 sec]
    test: terms, freq, prox...    test: stored fields.......OK [1500000 total field count; avg 1.0 fields per doc] [took 0.398 sec]
    test: term vectors........OK [0 total term vector count; avg 0.0 term/freq vector fields per doc] [took 0.000 sec]
    test: docvalues...........OK [0 docvalues fields; 0 BINARY; 0 NUMERIC; 0 SORTED; 0 SORTED_NUMERIC; 0 SORTED_SET; 0 SKIPPING INDEX] [took 0.000 sec]
    test: points..............OK [0 fields, 0 points] [took 0.000 sec]
    test: vectors.............OK [1 fields, 1500000 vectors] [took 0.497 sec]
    test: hnsw graph..........OK [4 levels, 1547684 nodes (over all levels)] [took 0.485 sec]

No problems were detected with this index.
```

I manually corrupted some neighbors lists in order to check the failure cases:
```
    test: open reader.........OK [took 0.012 sec]
    test: check integrity.....OK [took 3.633 sec]
    test: check live docs.....OK [took 0.000 sec]
    test: field infos.........OK [2 fields] [took 0.000 sec]
    test: field norms.........OK [0 fields] [took 0.000 sec]
    test: terms, freq, prox...    test: stored fields.......OK [1500000 total field count; avg 1.0 fields per doc] [took 0.393 sec]
    test: term vectors........OK [0 total term vector count; avg 0.0 term/freq vector fields per doc] [took 0.000 sec]
    test: docvalues...........OK [0 docvalues fields; 0 BINARY; 0 NUMERIC; 0 SORTED; 0 SORTED_NUMERIC; 0 SORTED_SET; 0 SKIPPING INDEX] [took 0.000 sec]
    test: points..............OK [0 fields, 0 points] [took 0.000 sec]
    test: vectors.............OK [1 fields, 1500000 vectors] [took 0.504 sec]
    test: hnsw graph..........ERROR: org.apache.lucene.index.CheckIndex$CheckIndexException: Neighbors out of order for node 31142: 839445<1326517 1st=33190
org.apache.lucene.index.CheckIndex$CheckIndexException: Neighbors out of order for node 31142: 839445<1326517 1st=33190
	at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CheckIndex.testHnswGraph(CheckIndex.java:2798)
	at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CheckIndex.testSegment(CheckIndex.java:1109)
	at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CheckIndex.checkIndex(CheckIndex.java:806)
	at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CheckIndex.checkIndex(CheckIndex.java:576)
	at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CheckIndex.doCheck(CheckIndex.java:4479)
	at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CheckIndex.doMain(CheckIndex.java:4316)
	at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CheckIndex.main(CheckIndex.java:4248)
```

I intend to dig into the HNSW creation and think about more checks that would be useful here, but I wanted to start small with these neighbor checks only.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
+++ b/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
@@ -52,6 +52,7 @@ import org.apache.lucene.codecs.PostingsFormat;
 import org.apache.lucene.codecs.StoredFieldsReader;
 import org.apache.lucene.codecs.TermVectorsReader;
 import org.apache.lucene.codecs.hnsw.FlatVectorsReader;
+import org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsReader;
 import org.apache.lucene.codecs.perfield.PerFieldKnnVectorsFormat;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.DocumentStoredFieldVisitor;
@@ -91,6 +92,7 @@ import org.apache.lucene.util.automaton.Automaton;
 import org.apache.lucene.util.automaton.ByteRunAutomaton;
 import org.apache.lucene.util.automaton.CompiledAutomaton;
 import org.apache.lucene.util.automaton.Operations;
+import org.apache.lucene.util.hnsw.HnswGraph;
 
 /**
  * Basic tool and API to check the health of an index and write a new segments file that removes
@@ -897,6 +899,74 @@ public final class CheckIndex implements Closeable {
               + lastCommit.counter
               + " is not greater than max segment name "
               + result.maxSegmentName);
+    }
+
+    try (DirectoryReader reader = DirectoryReader.open(dir)) {
+      for (LeafReaderContext context : reader.leaves()) {
+        LeafReader leafReader = context.reader();
+        KnnVectorsReader vectorsReader =
+            ((PerFieldKnnVectorsFormat.FieldsReader) ((CodecReader) leafReader).getVectorReader())
+                .getFieldReader("knn");
+        HnswGraph knnValues = ((Lucene99HnswVectorsReader) vectorsReader).getGraph("knn");
+        msg(
+            infoStream,
+            String.format(
+                Locale.ROOT, "Leaf %d has %d layers\n", context.ord, knnValues.numLevels()));
+        msg(
+            infoStream,
+            String.format(
+                Locale.ROOT, "Leaf %d has %d documents\n", context.ord, leafReader.maxDoc()));
+        final int numLevels = knnValues.numLevels();
+        for (int level = numLevels - 1; level >= 0; level--) {
+          int count = 0;
+          int min = Integer.MAX_VALUE, max = 0, total = 0;
+          long sumDelta = 0;
+          HnswGraph.NodesIterator nodesIterator = knnValues.getNodesOnLevel(level);
+          while (nodesIterator.hasNext()) {
+            int node = nodesIterator.nextInt();
+            int n = 0;
+            knnValues.seek(level, node);
+            int nbr, lastNeighbor = -1, firstNeighbor = -1;
+            while ((nbr = knnValues.nextNeighbor()) != NO_MORE_DOCS) {
+              if (firstNeighbor == -1) {
+                firstNeighbor = nbr;
+              }
+              // we see repeated neighbor nodes?!
+              assert nbr >= lastNeighbor
+                  : "neighbors out of order for node "
+                      + node
+                      + ": "
+                      + nbr
+                      + "<"
+                      + lastNeighbor
+                      + " 1st="
+                      + firstNeighbor;
+              assert nbr != lastNeighbor
+                      : "there are repeated neighbors of node " + node + " with value " + nbr;
+              lastNeighbor = nbr;
+              ++n;
+            }
+            max = Math.max(max, n);
+            min = Math.min(min, n);
+            if (n > 0) {
+              ++count;
+              total += n;
+              sumDelta += lastNeighbor - firstNeighbor;
+            }
+          }
+          msg(
+              infoStream,
+              String.format(
+                  Locale.ROOT,
+                  "Graphh level=%d size=%d, Fanout min=%d, mean=%.2f, max=%d, meandelta=%.2f\n",
+                  level,
+                  count,
+                  min,
+                  total / (float) count,
+                  max,
+                  sumDelta / (float) total));
+        }
+      }
     }
 
     if (result.clean) {
--- a/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
+++ b/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
@@ -2800,14 +2800,10 @@ public final class CheckIndex implements Closeable {
       if (fieldInfos.hasVectorValues()) {
         for (FieldInfo fieldInfo : fieldInfos) {
           if (fieldInfo.hasVectorValues()) {
-            if (vectorsReader instanceof PerFieldKnnVectorsFormat.FieldsReader) {
-              KnnVectorsReader fieldReader =
-                  ((PerFieldKnnVectorsFormat.FieldsReader) vectorsReader)
-                      .getFieldReader(fieldInfo.name);
-              if (fieldReader instanceof HnswGraphProvider) {
-                HnswGraph hnswGraph = ((HnswGraphProvider) fieldReader).getGraph(fieldInfo.name);
-                testHnswGraph(hnswGraph, fieldInfo.name, status);
-              }
+            KnnVectorsReader fieldReader = getFieldReaderForName(vectorsReader, fieldInfo.name);
+            if (fieldReader instanceof HnswGraphProvider graphProvider) {
+              HnswGraph hnswGraph = graphProvider.getGraph(fieldInfo.name);
+              testHnswGraph(hnswGraph, fieldInfo.name, status);
             }
           }
         }
@@ -2820,7 +2816,7 @@ public final class CheckIndex implements Closeable {
               status.hnswGraphsStatusByField.size(),
               nsToSec(System.nanoTime() - startNS)));
       printHnswInfo(infoStream, status.hnswGraphsStatusByField);
-    } catch (Throwable e) {
+    } catch (Exception e) {
       if (failFast) {
         throw IOUtils.rethrowAlways(e);
       }
@@ -2832,6 +2828,15 @@ public final class CheckIndex implements Closeable {
     }
 
     return status;
+  }
+
+  private static KnnVectorsReader getFieldReaderForName(
+      KnnVectorsReader vectorsReader, String fieldName) {
+    if (vectorsReader instanceof PerFieldKnnVectorsFormat.FieldsReader fieldsReader) {
+      return fieldsReader.getFieldReader(fieldName);
+    } else {
+      return vectorsReader;
+    }
   }
 
   private static void printHnswInfo(
--- a/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
+++ b/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
@@ -2819,7 +2819,7 @@ public final class CheckIndex implements Closeable {
               "OK [%d fields] [took %.3f sec]",
               status.hnswGraphsStatusByField.size(),
               nsToSec(System.nanoTime() - startNS)));
-      printHnswInfo(status.hnswGraphsStatusByField);
+      printHnswInfo(infoStream, status.hnswGraphsStatusByField);
     } catch (Throwable e) {
       if (failFast) {
         throw IOUtils.rethrowAlways(e);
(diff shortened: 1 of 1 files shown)


tteofili:
this might also check for graph connectivity, see #12627

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -406,6 +411,21 @@ public static final class VectorValuesStatus {
       public Throwable error;
     }
 
+    /** Status from testing HNSW graph */
+    public static final class HnswGraphStatus {
+
+      HnswGraphStatus() {}
+
+      /** Number of nodes in the graph */
+      public int hnswGraphSize;
```
mikemccand, line 420: Maybe rename to `graphNumNodes` or so?  `size` is a bit ambiguous/tricky, and since `hnsw` is in the class name we can prolly leave it off the member name?

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -406,6 +411,21 @@ public static final class VectorValuesStatus {
...
+
+      /** Number of levels of the graph */
+      public int hsnwGraphNumLevels;
```
mikemccand, line 423: `graphNumLevels`?

Naming is the hardest part!

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -2746,6 +2769,84 @@ public static Status.VectorValuesStatus testVectors(
     return status;
   }
 
+  private static HnswGraph getHnswGraph(CodecReader reader) throws IOException {
+    KnnVectorsReader vectorsReader = reader.getVectorReader();
+    if (vectorsReader instanceof PerFieldKnnVectorsFormat.FieldsReader) {
+      vectorsReader = ((PerFieldKnnVectorsFormat.FieldsReader) vectorsReader).getFieldReader("knn");
```
mikemccand, line 2775: Hmm the field can be anything (not necessarily `knn` like `luceneutil` uses)?  Can we change this so we iterate `FieldInfos` and for any field that has KNN vectors enabled (`FieldInfo.hasVectorValues`, hmm but also has the HNSW graph (since I think a field can be only "flat" vectors?)) we do this check?

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -2746,6 +2769,84 @@ public static Status.VectorValuesStatus testVectors(
...
+                        + firstNeighbor);
+              } else if (nbr == lastNeighbor) {
+                throw new CheckIndexException(
+                    "There are repeated neighbors of node " + node + " with value " + nbr);
+              }
+              lastNeighbor = nbr;
+            }
+            status.hnswGraphSize++;
+          }
+          status.hsnwGraphNumLevels++;
```
mikemccand, line 2825: Instead of incrementing here, you could just set it to `numLevels` from above?

mikemccand (commented):
Thank you for tackling this @benchaplin!  This is a blind spot for `CheckIndex` today, which is scary.  I left some minor comments...

mikemccand:
> this might also check for graph connectivity, see #12627

+1 -- this is a tricky thing about these HNSW graphs (that they are not necessarily born connected but rather must be stitched up somehow afterwards).

But maybe do this separately?  I'd love to just get this initial coverage in first...

mikemccand:
Actually, `CheckIndex` does have some coverage for vectors and KNN graph (it confirms it can enumerate all vectors, and also runs some searches on it if it's not just flat vectors (`checkByte/FloatVectorValues`) , so it's not completely blind :)

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
benchaplin: Good point - renamed it `totalNumNodes` to be extra clear.

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
benchaplin: I could, but wanted to respect the object as a "status" - so if a check fails, `numLevels` will indicate on which level it failed. However we don't print the status if an exception is thrown. What do you think?

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
benchaplin: Thanks, I didn't quite understand fields when I wrote this - I think I get it now. Alright, I've done what you suggested (as is also done in `testVectors`) and iterated over `FieldInfos`, performing the check only when it applies. 

Because we might now parse several HNSW graphs, I've restructured the status object to support per-graph data. Successful output will now look like:

```
    test: open reader.........OK [took 0.010 sec]
    test: check integrity.....OK [took 2.216 sec]
    test: check live docs.....OK [took 0.000 sec]
    test: field infos.........OK [2 fields] [took 0.000 sec]
    test: field norms.........OK [0 fields] [took 0.000 sec]
    test: terms, freq, prox...    test: stored fields.......OK [1500000 total field count; avg 1.0 fields per doc] [took 0.390 sec]
    test: term vectors........OK [0 total term vector count; avg 0.0 term/freq vector fields per doc] [took 0.000 sec]
    test: docvalues...........OK [0 docvalues fields; 0 BINARY; 0 NUMERIC; 0 SORTED; 0 SORTED_NUMERIC; 0 SORTED_SET; 0 SKIPPING INDEX] [took 0.000 sec]
    test: points..............OK [0 fields, 0 points] [took 0.000 sec]
    test: vectors.............OK [1 fields, 1500000 vectors] [took 0.496 sec]
    test: hnsw graphs.........OK [2 fields: (field name: knn1, levels: 4, total nodes: 1547684), (field name: knn2, levels: 4, total nodes: 1547684)] [took 0.979 sec]
```

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -2746,6 +2781,106 @@ public static Status.VectorValuesStatus testVectors(
...
+      final int numLevels = hnswGraph.numLevels();
+      // Perform tests on each level of the HNSW graph
+      for (int level = numLevels - 1; level >= 0; level--) {
+        HnswGraph.NodesIterator nodesIterator = hnswGraph.getNodesOnLevel(level);
+        while (nodesIterator.hasNext()) {
+          int node = nodesIterator.nextInt();
+          hnswGraph.seek(level, node);
+          int nbr, lastNeighbor = -1, firstNeighbor = -1;
+          while ((nbr = hnswGraph.nextNeighbor()) != NO_MORE_DOCS) {
+            if (firstNeighbor == -1) {
```
msokolov, line 2858: perhaps at some point we could also check for uniqueness of the neighbors, and also check that the neighbors are in range [0, numGraphNodes], and finally on levels > 0, we would want to assert that the neighbors are on this level. But this can all be done separately; maybe we could add a comment here?

benchaplin: Sounds good, I'll add a TODO (and add those checks in a followup PR). A couple questions first: 
- Doesn't (neighbors are in order) and (neighbors are not repeated) verify uniqueness?
- Also, why wouldn't we assert neighbors are on this level for level 0?

msokolov:
> Doesn't (neighbors are in order) and (neighbors are not repeated) verify uniqueness?

duh, yes :)
    
    > Also, why wouldn't we assert neighbors are on this level for level 0?

yes, we should, it's just that for level 0 this is the same as node in [0, size) wheras other levels are sparse - they have the same max = size-1, but not every node is included

benchaplin:
I've added @msokolov's node-in-range check + neighbor-on-same-level check, and @tteofili's connectivity check. I chose not to throw an exception for disconnectedness, as I've found there's often a couple disconnected nodes at level 0 (at least in my indices). Maybe we could throw if connectivity is below some threshold? 

So overall, exceptions are thrown if:
- A node is not in the range [0, graphSize - 1]
- A node's neighbor is not on the same level as the node
- A node's neighbors are out of order 
- A node's neighbors contain duplicates

I implemented a more detailed report in the logs: node counts and connectivity are printed for each level of the graph, e.g.
```
test: open reader.........OK [took 0.012 sec]
test: check integrity.....OK [took 0.489 sec]
test: check live docs.....OK [took 0.000 sec]
test: field infos.........OK [2 fields] [took 0.000 sec]
test: field norms.........OK [0 fields] [took 0.000 sec]
test: terms, freq, prox...    test: stored fields.......OK [200000 total field count; avg 1.0 fields per doc] [took 0.055 sec]
test: term vectors........OK [0 total term vector count; avg 0.0 term/freq vector fields per doc] [took 0.000 sec]
test: docvalues...........OK [0 docvalues fields; 0 BINARY; 0 NUMERIC; 0 SORTED; 0 SORTED_NUMERIC; 0 SORTED_SET; 0 SKIPPING INDEX] [took 0.000 sec]
test: points..............OK [0 fields, 0 points] [took 0.000 sec]
test: vectors.............OK [1 fields, 200000 vectors] [took 0.111 sec]
test: hnsw graphs.........OK [1 fields] [took 0.317 sec]
  hnsw field name: knn
    level 3: 11 nodes, 11/11 connected
    level 2: 141 nodes, 141/141 connected
    level 1: 6022 nodes, 6022/6022 connected
    level 0: 200000 nodes, 199999/200000 connected
```

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -2746,6 +2785,191 @@ public static Status.VectorValuesStatus testVectors(
...
+      }
+      msg(
+          infoStream,
+          String.format(
+              Locale.ROOT,
+              "OK [%d fields] [took %.3f sec]",
+              status.hnswGraphsStatusByField.size(),
+              nsToSec(System.nanoTime() - startNS)));
+      printHnswInfo(infoStream, status.hnswGraphsStatusByField);
+    } catch (Throwable e) {
```
msokolov, line 2823: I think it should be enough to catch  Exception here and let an outer frame handle Errors

msokolov (commented):
This looks great - I just have one nit about the exception handling, otherwise I think it's ready to push

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
```diff
@@ -2746,6 +2785,191 @@ public static Status.VectorValuesStatus testVectors(
...
+        for (FieldInfo fieldInfo : fieldInfos) {
+          if (fieldInfo.hasVectorValues()) {
+            if (vectorsReader instanceof PerFieldKnnVectorsFormat.FieldsReader) {
+              KnnVectorsReader fieldReader =
+                  ((PerFieldKnnVectorsFormat.FieldsReader) vectorsReader)
+                      .getFieldReader(fieldInfo.name);
+              if (fieldReader instanceof HnswGraphProvider) {
+                HnswGraph hnswGraph = ((HnswGraphProvider) fieldReader).getGraph(fieldInfo.name);
+                testHnswGraph(hnswGraph, fieldInfo.name, status);
+              }
```
benwtrent, line 2810:
```suggestion
            if (vectorsReader instanceof PerFieldKnnVectorsFormat.FieldsReader fieldsReader) {
              KnnVectorsReader fieldReader = fieldsReader.getFieldReader(fieldInfo.name);
              if (fieldReader instanceof HnswGraphProvider graphProvider) {
                HnswGraph hnswGraph = graphProvider.getGraph(fieldInfo.name);
                testHnswGraph(hnswGraph, fieldInfo.name, status);
              }
```

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
benwtrent: agreed, only do exceptions here.

benwtrent (commented):
One minor modernization comment.

mikemccand:
Thank you for persisting on this important change @benchaplin!

I applied this PR to my local Lucene clone and ran `CheckIndex` on the vector index created by [last night's nightly Lucene benchmarks](https://benchmarks.mikemccandless.com/) index but there was no new output covering the HNSW graph.

I think the reason is because the benchmarks [create an index using `Lucene99HnswVectorsFormat`](https://github.com/mikemccand/luceneutil/blob/main/src/main/perf/Indexer.java#L477-L493) without the `PerField` wrapper.

I think it'd be easy to also handle this case in your change?  In addition to checking if the `PerFieldKnnVectorsFormat` is in use, you could check whether the `reader.getVectorReader() instanceof HnswGraphProvider`?  So both routes (per-field case and not per-field case) could yield an `HnswGraphProvider` which you can then check for integrity ...

It's OK to do this in a follow-on issue too... either way.

msokolov:
Ooh, I was wondering about this `PerFieldKnnVectorsFormat` case - thanks for testing @mikemccand

lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
benchaplin: thanks, much nicer!

mikemccand:
> Let me know if you're able to test this revision!

Yay, I just tested again and I see HNSW output, and the nightly benchy index passed all the new tests here!  Thanks @benchaplin.

```
    test: hnsw graphs.........OK [1 fields] [took 0.804 sec]
      hnsw field name: vector
        level 4: 2 nodes, 2/2 connected
        level 3: 85 nodes, 85/85 connected
        level 2: 1808 nodes, 1808/1808 connected
        level 1: 31568 nodes, 31565/31568 connected
        level 0: 497804 nodes, 497794/497804 connected
```

msokolov:
FYI @benchaplin this is merged now, but then I realized there is no CHANGES entry -- would you like to add one? After that, we can probably also backport to `branch_10x` using cherry-picks

benchaplin:
Thanks @mikemccand and @msokolov! And oh, forgot about the CHANGES - here's an entry (I listed under 10.2.0) https://github.com/apache/lucene/pull/14120.

Issue #13920, opened 2024-10-16
Performance difference between files getting opened with IOContext.RANDOM vs IOContext.READ during merges


navneet1v:
## Description
For past 1 month we have been testing difference in performance for a files getting opened with IOContext.RANDOM vs IOContext.READ specially during merges with Lucene version 9.11.1 and Opensearch version 2.17. We started this deep-dive when we saw an increase in time for our force merges.

### Background
Opensearch k-NN plugin provides the vector search capabilities with Opensearch. The architecture of k-NN plugin is very similar to how Lucene implements the vector search with few small key differences. Before 2.17 version of Opensearch, opensearch was using dvd files to store the raw the vectors and graphs files are stored separately. These graphs are not built using Lucene HNSW but with some external libraries like Faiss.
In 2.17 version of Opensearch, we started using KNNFlatVectorFormat to store and read the raw vectors in place of dvd files. As reading vectors from .vec files as float[] is more efficient than reading byte[] and then converting to float[]. Ref: https://github.com/opensearch-project/k-NN/issues/1853

## Observations
After doing the switch what we observed that our merge time for a 10M 768D dataset is increased by [20%](https://github.com/opensearch-project/k-NN/issues/2134). We did an extensive deep-dive/experiments on the root cause([ref1](https://github.com/opensearch-project/k-NN/issues/2134#issuecomment-2415496614), [ref2](https://github.com/opensearch-project/k-NN/issues/2134#issuecomment-2415542711)) and code difference between dvd files and .vec files format and was able to see that IOContext.RANDOM with .vec files is causing this regression. 

This regression comes because during merges for every Lucene99FlatVectorsReader there are some operations happens like [checkIntegrity](https://github.com/apache/lucene/blob/branch_9_11/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatVectorsReader.java#L227-L229)(which does checksum of whole file), [reading of all vector values to create new segment](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsWriter.java#L242-L266) which are more of sequential reads than random reads. 

I do believe that having a `RANDOM` madvise on `.vec` file is more beneficial for search and graph merges as Lucene uses this file as a way to store raw vectors for HNSW. BTW This PR added the capability: https://github.com/apache/lucene/pull/13267 for Random IOContext to .vec files.

## Solutions Tried:
[...]

PR #13985, opened 2024-11-09
Introduces IndexInput#updateReadAdvice to change the ReadAdvice while merging vectors 

shatejas:
The change is needed to be able to reduce the force merge time. Lucene99FlatVectorsReader is opened with IOContext.RANDOM, this optimizes searches with madvise as RANDOM. For merges we need sequential access and ability to preload pages to be able to shorten the merge time.

The change updates the ReadAdvice.SEQUENTIAL before the merge starts and reverts it to ReadAdvice.RANDOM at the end of the merge for
Lucene99FlatVectorsReader.

### Description

Benchmarking results coming up. Opening as a draft to get initial feedback

### Related issues
https://github.com/apache/lucene/issues/13920


shatejas:
cc: @uschindler @jpountz @navneet1v

lucene/test-framework/src/java/org/apache/lucene/tests/store/BaseDirectoryTestCase.java
```diff
@@ -1554,6 +1555,42 @@ public void testPrefetchOnSlice() throws IOException {
...
+          if (random().nextBoolean()) {
+            orig.updateReadAdvice(randomReadAdvice());
+          }
+        }
+      }
+    }
+  }
+
+  private ReadAdvice randomReadAdvice() {
+    return ReadAdvice.values()[TestUtil.nextInt(random(), 0, ReadAdvice.values().length - 1)];
```
jpountz, line 1591: You could use `RandomPicks#randomFrom`.

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsReader.java
```diff
@@ -123,4 +123,11 @@ public abstract void search(
   public KnnVectorsReader getMergeInstance() {
     return this;
   }
+
+  /**
+   * Optional: reset or close merge resources used in the reader
+   *
+   * <p>The default implementation is empty
+   */
+  public void finishMerge() throws IOException {}
```
jpountz, line 132: I wonder if we should reuse the close() method of merge instances for this, what do you think @uschindler ?

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatVectorsReader.java
```diff
@@ -170,6 +170,15 @@ public void checkIntegrity() throws IOException {
     CodecUtil.checksumEntireFile(vectorData);
   }
 
+  @Override
+  public FlatVectorsReader getMergeInstance() {
+    try {
+      return new MergeLucene99FlatVectorsReader(this);
+    } catch (IOException exception) {
+      throw new RuntimeException(exception);
```
jpountz, line 178:
```suggestion
      throw new UncheckedIOException(exception);
```

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java
```diff
@@ -113,6 +114,25 @@ public Lucene99HnswVectorsReader(SegmentReadState state, FlatVectorsReader flatV
     }
   }
 
+  private Lucene99HnswVectorsReader(
+      Lucene99HnswVectorsReader reader, KnnVectorsReader flatVectorsReader) {
+    assert flatVectorsReader instanceof FlatVectorsReader;
```
jpountz, line 119: I would rather make this ctor take a `FlatVectorsReader` and push the responsibility to callers to make the cast (or maybe we don't even need a cast if we make getMergeInstance() return a `FlatVectorsReader`?)

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsReader.java
jpountz: Separately, it would be nice to improve the asserting framework (`AssertingKnnVectorsReader` in this case) to validate that `finishMerge()` is always called on merge instances.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatVectorsReader.java
```diff
@@ -327,4 +336,61 @@ static FieldEntry create(IndexInput input, FieldInfo info) throws IOException {
           info);
     }
   }
+
+  private static final class MergeLucene99FlatVectorsReader extends FlatVectorsReader {
+
+    private final Lucene99FlatVectorsReader delegate;
+
+    MergeLucene99FlatVectorsReader(final Lucene99FlatVectorsReader flatVectorsReader)
```
jpountz, line 344: Nit: I'm not sure I would create a wrapper, maybe we should just add a copy constructor and have a `boolean merging` instance like we do e.g. on stored fields (see `Lucene90CompressingStoredFieldsReader`).

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsReader.java
shatejas: > I wonder if we should reuse the close() method of merge instances for this, what do you think @uschindler ?

I went with that solution at first, `close` is called way to late, to be able to benefit the search requests going on as per my understanding. I was looking for something to switch back to random access earlier and thats how I landed with this approach. Happy to move it to `close()` if there is a preference

>> Separately, it would be nice to improve the asserting framework (AssertingKnnVectorsReader in this case) to validate that finishMerge() is always called on merge instances.

Wasn't aware of AssertingKnnVectorsReader. I can take a look

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java
shatejas: >> I would rather make this ctor take a FlatVectorsReader and push the responsibility to callers to make the cast

This makes sense to me, I will change it

>> maybe we don't even need a cast if we make getMergeInstance() return a FlatVectorsReader

This seems difficult as of now, [MergeState](https://github.com/shatejas/lucene/blob/12ca4779b962c96367f3e6a8b06523837e5e6434/lucene/core/src/java/org/apache/lucene/index/MergeState.java#L157) expects KNNVectorsReader and FlatVectorsReader is wrapped inside of KNNVectorsReader

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatVectorsReader.java
shatejas: This seemed a bit more clean, Open to the approach if there is a strong preference

lucene/test-framework/src/java/org/apache/lucene/tests/store/BaseDirectoryTestCase.java
shatejas: Will switch

ChrisHegarty:
@shatejas I'm curious how much this actually helps, and I know that you said that benchmark results would be posted.

I do like that we can update the ReadAdvice on an index input 👍  What is unclear is how much the sequential advise here helps over something like a `load` (similar to `MemorySegment::load`), that touches every page.

uschindler:
Hi,
I am currently on travel, so I can't review this. Will look into it posisbly later this week. Greetings from Costa Rica!

shatejas:
> I'm curious how much this actually helps, and I know that you said that benchmark results would be posted.

@ChrisHegarty Preliminary results showed approximately 40mins (~13%) reduction in total force merge time for 10m dataset. You should be able to find details here https://github.com/opensearch-project/k-NN/issues/2134#issuecomment-2420793541. 
Though the setup does not use LuceneHNSW it does use Lucene99FlatVectorsReader.

> What is unclear is how much the sequential advise here helps over something like a load (similar to MemorySegment::load), that touches every page.

I did try preload which does `MemorySegment::load` and there was an improvement in force merge time. The advantage I see with updateAdvice is being able to delay the loading of vectors till needed.

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsReader.java
ChrisHegarty: > I wonder if we should reuse the close() method of merge instances for this, what do you think ... ?

It might be possible to use `close` here, if we override it to restore the previous read advice, and ensure that the merge calls close.  Additionally, either avoid closing the underlying index input or slice the input upon creation of the new instance. (closing the slice will not close the underlying input). 

I see that this is all per-thread, so this may be fine. But just to say it explicitly here, changing the read advice, whether it be on a slice or not, will affect all instances that share the underlying index.  This should be fine - since the reader is just being used for merging at this point.

ChrisHegarty:
Thanks @shatejas 

For clarity, the bottleneck that is being fixed here is with the reading of all vector data from the to-be-merged segments, when copying that data to the new segment - all non-deleted vectors are accessed sequential, just once, as they are copied.  We can then switch back to random access when building the graph.

ChrisHegarty:
Generally, I think that the direction in this PR is good. I wanna help get it moved forward. I'll do some local testing and perf runs to verify the impact. I can also commit some tests and improvements directly to the PR branch. @shatejas ?

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsReader.java
ChrisHegarty: How "sneaky" would it be to just update the read advice and return the same instance. Then restore the random read advice on finishMerge.

shatejas: > How "sneaky" would it be to just update the read advice and return the same instance. Then restore the random read advice in finishMerge. The reason I suggest this is that so far all options and discussions lead to sharing the same underlying memory, so regardless on the actual instance here, then the advice is going to change on the memory.

Makes sense to me, and if we switch to `close` we can switch to the approach mentioned by @jpountz in https://github.com/apache/lucene/pull/13985#discussion_r1835532510

ChrisHegarty:
I added some asserts to the asserting wrappers to ensure that finishMerge is always called. I think that the code is in good shape now. Pending some benchmark runs..

@shatejas no need to force-push. If merged, we collapse the commits into a single one.

shatejas:
### Benchmarks

#### Setup 1 - Opensearch cluster

Ran with [opensearch benchmarks](https://github.com/opensearch-project/opensearch-benchmark)

Total data nodes - 3
Total shards - 6 (2 per node), no replicas
Memory - 128gb
vCPU - 16

Dataset used: cohere-10m

Baseline - OS 2.18 and lucene 9.12
Candidate - OS 2.16 and lucene [9.12 with readAdvice changes](https://github.com/apache/lucene/compare/branch_9_12...shatejas:lucene:branch_9_12)

**Why was this tested with lucene 9.12?**
Opensearch is not using lucene >9.12 for any of its version. Upgrading it to use lucene 10 requires significant changes. For candidate, required commits were cherry-picked


Run 1: sequence of operations: delete-index -> create-index -> add documents -> force-merge -> search

##### Results
|           | Force-merge(ms)   | Force-merge(hrs) | Search p50 | Search p90 | Search p99 |
|-----------|-------------------|------------------|------------|------------|------------|
| Baseline  | 15795889.88313920 | 4hrs 23 mins     | 9.6        | 10.8       | 14.7       |
| Candidate | 15204143.95724240 | 4hrs 13mins      | 10.7       | 12.0       | 15.0       |

Run 2:  Search performed on already indexed data from above run

|           | Search p50 | Search p90 | Search p99 |
|-----------|------------|------------|------------|
| Baseline  | 9.7        | 10.6       | 12.1       |
| Candidate | 10.4       | 11.3       | 12.5       |

#### Setup 2: Used lucene-utils knnPerfTest.py

Baseline - Lucene main
Candidate - Lucene main with current commit

**Baseline**

| recall |  latency (ms)  | nDoc | topK  |  fanout   | maxConn |   beamWidth |   quantized  |  index s   | index docs/s  |  force merge s   | num segments  |  index size (MB) | 
|-----------|------------|------------|------------|-----------|------------|------------|------------|-----------|------------|------------|------------|------------|
| 0.644  |       0.428 | 50000 |   10 |     64 |      64  |      250  |       no |   18.97  |     2635.18  |         1.89   |          1    |        20.62 |

**Candidate**

| recall |  latency (ms)  | nDoc | topK  |  fanout   | maxConn |   beamWidth |   quantized  |  index s   | index docs/s  |  force merge s   | num segments  |  index size (MB) | 
|-----------|------------|------------|------------|-----------|------------|------------|------------|-----------|------------|------------|------------|------------|
|  0.644    |      0.436 |   50000  |   10    |   64   |     64      |   250    |      no  |   20.20    |    2474.76    |        1.77  |            1   |          20.62 | 

There is a small affect on search latencies, its hard to say if its due to the change or just a fluctuation in the runs. I couldn't think of a reason that would of search latencies

@jpountz @ChrisHegarty thoughts?

ChrisHegarty (approved):
Restoring sequential advice is the right thing to do here, and has clear perf benefits. The code LGTM.  lemme know if u need anything further to get this merged. And backported to 10.x

shatejas:
Thanks @ChrisHegarty! 

I do need help with merging this. I have updated `CHANGES.txt`. Do let me know the process of backporting it - if it needs cherry-picking and raising a new PR on 10.x branch, I can do it.

Let me know if I need anything else to merge this

ChrisHegarty:
The `org.apache.lucene.index.TestConcurrentMergeScheduler.testNoWaitClose` test hits a new assert that I added - sorry. I need to look to see if it is a test issue or more of a design issue with finishMerge.

shatejas:
> The `org.apache.lucene.index.TestConcurrentMergeScheduler.testNoWaitClose` test hits a new assert that I added - sorry. I need to look to see if it is a test issue or more of a design issue with finishMerge.
> 
> ```
> ./gradlew test --tests TestConcurrentMergeScheduler.testNoWaitClose -Dtests.file.encoding=UTF-8 -Dtests.iters=10
> ```

I took a look at the failure. Here is what I found, `mergeInstanceCount` does not decrement if the merge is aborted between `getMergeInstance` and  when the merge starts ([ref](https://github.com/apache/lucene/blob/9b5edabf21217699a11ecc8aee7e7d1bf39de4fb/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java#L5258)). In that scenario finishMerge is never called and the readers are closed in the [finally block](https://github.com/apache/lucene/blob/9b5edabf21217699a11ecc8aee7e7d1bf39de4fb/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java#L5474) failing the `assert mergeInstanceCount == 0` in AssertingKnnVectorsReader

ChrisHegarty:
I broke the tests again, by adding more asserts, as I'm a little uncomfortable with the `finishMergeCount <= 0`. Now I broke something else. I'll get to fixing this, or backout the extra asserts. For reference this reproduces:

```
./gradlew :lucene:core:test --tests "org.apache.lucene.codecs.perfield.TestPerFieldKnnVectorsFormat.testByteVectorScorerIteration" -Ptests.jvms=6 "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1" -Ptests.seed=C47EA6F578AC0A04 -Ptests.iters=1 -Ptests.gui=false -Ptests.file.encoding=UTF-8 -Ptests.vectorsize=128 -Ptests.forceintegervectors=true -Dtests.iters=10
```

shatejas:
> I broke the tests again, by adding more asserts, as I'm a little uncomfortable with the finishMergeCount <= 0. Now I broke something else. I'll get to fixing this, or backout the extra asserts. For reference this reproduces:

Took a look, The mismatch between `mergeInstanceCount` and `mergeInstance` is because mergeInstanceCount is being updated in parent and mergeInstance is updated to true during getMergeInstance(). If search is happening during merge there is a good chance that it runs into a scenario where `mergeInstanceCount = 1` but `mergeInstance` is false. 

Considering the test has commit() call followed by search, I think its running into the scenario

ChrisHegarty (approved):

shatejas:
Thanks a lot @ChrisHegarty for adding tight tests and merging this!

uschindler:
Thanks @ChrisHegarty for taking care. If I have any additional comments about API I will open another followup PR.
Still in 🇨🇷, flying back this evening.

msokolov:
The impact on search times is surprising, as you said. Can you clarify one thing a about the benchmark setup: does it perform searches concurrently with indexing (and merging) on the same box, or are they completely separate?

shatejas:
> The impact on search times is surprising, as you said. Can you clarify one thing a about the benchmark setup: does it perform searches concurrently with indexing (and merging) on the same box, or are they completely separate?

searches were after the completion of indexing and force-merge

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatVectorsReader.java
```diff
@@ -250,6 +262,13 @@ public RandomVectorScorer getRandomVectorScorer(String field, byte[] target) thr
         target);
   }
 
+  @Override
+  public void finishMerge() throws IOException {
+    // This makes sure that the access pattern hint is reverted back since HNSW implementation
+    // needs it
+    this.vectorData.updateReadAdvice(ReadAdvice.RANDOM);
```
msokolov, line 269: This is too opinionated. If we want to have this swapping behavior I think we need to find a way to cache the original ReadAdvice and restore it, since we don't know what it was - that decision is made by the Directory and its hints and iocontexts and sysprops and so on. But I don't think we currently have `IOContext.getReadAdvice()`?

navneet1v: I think this make sense. We should flip back to the read advise which was originally present

msokolov:
I'm really not sure about this change any more -- looking back at the (merging) performance improvement, it was not very large, and somewhat offset by the mysterious search slowdowns. At the same time, to get this working we need to add lots of stuff: reference counting to ensure proper closing, access to memory mapping hints. Are we still convinced it is  worth it?

shatejas:
>  it was not very large, and somewhat offset by the mysterious search slowdowns

We can switch to an approach where the madvise is not changed at all. The approach leverages prefetch,  read advise still changes but the reads are not as aggressive as sequential. A prefetch call after each read (inside merge) will make sure that merges don't slow down. At the same time this minimizes the search perf impact considering prefetch is not aggressive. 

https://github.com/apache/lucene/pull/14076#issuecomment-2563503703

msokolov:
Perhaps if we have a case where there is no random access (from Lucene) at all and we are only using Lucene to store the vector data - any searh indexing is being done by a native plugin (I think this is what you are targeting?) then we don't really want to be switching back and forth between access modes, but rather we'd want to set to SEQUENTIAL and leave it that way, ideally?  In that case, what if the reader could recognize that it has no associated HNSW graph (somehow)  and provide some kind of hint to the directory that it could then use to decide to use sequential?  Also, wondering if SEQUENTIAL is really that much better than NORMAL? If not, then we could simply revert this given that we are retruning to NORMAL as the default.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatVectorsReader.java
shatejas: > Directory and its hints and iocontexts and sysprops and so on. But I don't think we currently have IOContext.getReadAdvice()?

At the time of the change the initial advice was random so this is reverting back to random. But I agree, we can simply cache it and revert it from the cached value

msokolov:
I'll open an issue for backporting https://github.com/apache/lucene/pull/14702 to 10x to fix this

shatejas:
> Perhaps if we have a case where there is no random access (from Lucene) at all and we are only using Lucene to store the vector data - any searh indexing is being done by a native plugin (I think this is what you are targeting?) then we don't really want to be switching back and forth between access modes

In certain scenarios, random access is critical for optimal performance. Our initial hypothesis, which we are currently validating through benchmarks, is that random access lookups are significantly more efficient than sequential lookup even for flat vector data under high memory pressure, especially when doing a exact search on filtered documents. This is because random access avoids data prefetching, thereby reducing memory swapping.

To provide users with greater control, its best to allow them to configure initial IOContext value based on their specific workloads, rather than keeping it a fixed constant. This approach will offer flexibility while maintaining sensible defaults.

Furthermore, we can mitigate the impact on search performance during merges by implementing a dedicated prefetch functionality for vector merges. This eliminates the need to switch between access methods, ensuring a minimal impact on ongoing searches.

PR #13986, opened 2024-11-11
Allow easier verification of the Panama Vectorization provider with newer Java versions

ChrisHegarty:
This commit allows easier verification of the Panama Vectorization provider with newer Java versions.

The upper bound Java version of the Vectorization provider is hardcoded to the version that has been tested and is known to work. This is a bit inflexible when experimenting with and verifying newer JDK versions. This change proposes to add a new system property that allows to set the upper bound of the range of Java versions supported.

With this change, and the accompanying small gradle change, then one can verify newer JDKs as follows:

```
CI=true; RUNTIME_JAVA_HOME=/Users/chegar/binaries/jdk-24.jdk-ea-b23/Contents/Home
./gradlew :lucene:core:test -Dorg.apache.lucene.vectorization.upperJavaFeatureVersion=24
```

This change helps both testing and verifying with Early Access JDK builds, as well as allowing to override the upper bound when the JDK is known to work fine.

### DIFF
--- a/gradle/testing/defaults-tests.gradle
+++ b/gradle/testing/defaults-tests.gradle
@@ -132,8 +132,10 @@ allprojects {
       def v = JavaVersion.toVersion(Integer.parseInt(prop)).majorVersion
       if (rootProject.vectorIncubatorJavaVersions.contains(rootProject.runtimeJavaVersion) ||
           rootProject.runtimeJavaVersion.majorVersion <= v) {
-        systemProperty 'org.apache.lucene.vectorization.upperJavaFeatureVersion', v
         jvmArgs '--add-modules', 'jdk.incubator.vector'
+        if (rootProject.runtimeJavaVersion.majorVersion <= v) {
+          systemProperty 'org.apache.lucene.vectorization.upperJavaFeatureVersion', v
+        }
       }
       
       jvmArgs '--enable-native-access=' + (project.path in [
--- a/gradle/testing/defaults-tests.gradle
+++ b/gradle/testing/defaults-tests.gradle
@@ -128,12 +128,10 @@ allprojects {
       jvmArgs '--add-modules', 'jdk.management'
 
       // Enable the vector incubator module on supported Java versions:
-      def v = JavaVersion.VERSION_1_1
-      def prop = providers.systemProperty("org.apache.lucene.vectorization.upperJavaFeatureVersion")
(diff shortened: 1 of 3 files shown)


gradle/testing/defaults-tests.gradle
```diff
@@ -128,7 +128,13 @@ allprojects {
       jvmArgs '--add-modules', 'jdk.management'
 
       // Enable the vector incubator module on supported Java versions:
-      if (rootProject.vectorIncubatorJavaVersions.contains(rootProject.runtimeJavaVersion)) {
+      def v = JavaVersion.VERSION_1_1
+      def prop = providers.systemProperty("org.apache.lucene.vectorization.upperJavaFeatureVersion")
```
dweiss, line 132: It'd be probably more consistent to use the propertyOrDefault "function" that we defined globally to allow passing such properties via -P (gradle's project properties) or -D (system properties). You can provide the default as the second argument - look at any existing call of that function.

ChrisHegarty: thanks @dweiss, that's a bit cleaner. Done.

dweiss (commented):
lgtm.

uschindler (approved):
To me looks fine. I was expecting some fancy reflection checks if required methods are available, but adding a system property to override the latest supported java version looks like a good compromize.

I have no time to closer test the whole setup, but security handling looks fine.

Personally I would prefer a less if/else/default handling using `Optional` like done in the previous sysprops.

Gradle code looks fine.

Greetings from Costa Rica!

ChrisHegarty:
> > Personally I would prefer a less if/else/default handling using Optional like done in the previous sysprops.
> 
> I'll make that change before merging.

Well... I tried, and it was awful! The reason being the need to parseInt the value and catch the exception, which, while possible, just lead to more convoluted code. Easy to refactor later, if at all.

uschindler:
> > > Personally I would prefer a less if/else/default handling using Optional like done in the previous sysprops.
> > 
> > 
> > I'll make that change before merging.
> 
> Well... I tried, and it was awful! The reason being the need to parseInt the value and catch the exception, which, while possible, just lead to more convoluted code. Easy to refactor later, if at all.

I would have left the code to bail out if the Sysprop is empty or no valid integer, so not catch exception at all.

Anyways, it's good as is.

PR #13987, opened 2024-11-12
Update lastDoc in ScoreCachingWrappingScorer

msfroh:
### Description

I noticed that ScoreCachingWrappingScorer never updates lastDoc, so it's always -1. Technically, it's probably fine, since it still ends up returning the same score for multiple score() calls between collect calls, but I think this change better reflects the intended logic. (In particular, if the same doc was somehow collected multiple times, then the score would get recalculated.)

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/TestPositiveScoresOnlyCollector.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestPositiveScoresOnlyCollector.java
@@ -114,8 +114,9 @@ public class TestPositiveScoresOnlyCollector extends LuceneTestCase {
     Collector c = new PositiveScoresOnlyCollector(tdc);
     LeafCollector ac = c.getLeafCollector(ir.leaves().get(0));
     ac.setScorer(s);
-    while (s.iterator().nextDoc() != DocIdSetIterator.NO_MORE_DOCS) {
-      ac.collect(0);
+    int docId;
+    while ((docId = s.iterator().nextDoc()) != DocIdSetIterator.NO_MORE_DOCS) {
+      ac.collect(docId);
     }
     TopDocs td = tdc.topDocs();
     ScoreDoc[] sd = td.scoreDocs;
--- a/lucene/core/src/test/org/apache/lucene/search/TestScoreCachingWrappingScorer.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestScoreCachingWrappingScorer.java
@@ -157,4 +157,40 @@ public class TestScoreCachingWrappingScorer extends LuceneTestCase {
     ir.close();
     directory.close();
   }
+
+  private static class CountingScorable extends FilterScorable {
+    int count = 0;
+
+    public CountingScorable(Scorable in) {
+      super(in);
+    }
+
+    @Override
+    public float score() throws IOException {
+      count++;
+      return in.score();
+    }
+  }
+
+  public void testRepeatedCollectReusesScore() throws Exception {
+    Scorer s = new SimpleScorer();
+    CountingScorable countingScorable = new CountingScorable(s);
+    ScoreCachingCollector scc = new ScoreCachingCollector(scores.length * 2);
+    LeafCollector lc = scc.getLeafCollector(null);
+    lc.setScorer(countingScorable);
+
+    // We need to iterate on the scorer so that its doc() advances.
+    int doc;
+    while ((doc = s.iterator().nextDoc()) != DocIdSetIterator.NO_MORE_DOCS) {
+      lc.collect(doc);
+      lc.collect(doc);
+    }
+
+    for (int i = 0; i < scores.length; i++) {
+      assertEquals(scores[i], scc.mscores[i * 2], 0f);
+      assertEquals(scores[i], scc.mscores[i * 2 + 1], 0f);
+    }
+
+    assertEquals(scores.length, countingScorable.count);
+  }
 }
--- a/lucene/core/src/test/org/apache/lucene/search/TestScoreCachingWrappingScorer.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestScoreCachingWrappingScorer.java
@@ -159,7 +159,7 @@ public class TestScoreCachingWrappingScorer extends LuceneTestCase {
   }
 
   private static class CountingScorable extends FilterScorable {
-    int count = 0;
+    int count;
 
     public CountingScorable(Scorable in) {
       super(in);
@@ -175,6 +175,7 @@ public class TestScoreCachingWrappingScorer extends LuceneTestCase {
   public void testRepeatedCollectReusesScore() throws Exception {
     Scorer s = new SimpleScorer();
     CountingScorable countingScorable = new CountingScorable(s);
+    // We're going to collect every document twice, so we need to allocate a 2x larger array.
     ScoreCachingCollector scc = new ScoreCachingCollector(scores.length * 2);
     LeafCollector lc = scc.getLeafCollector(null);
     lc.setScorer(countingScorable);
@@ -186,11 +187,14 @@ public class TestScoreCachingWrappingScorer extends LuceneTestCase {
       lc.collect(doc);
     }
 
+    // Verify that each score was collected twice.
     for (int i = 0; i < scores.length; i++) {
       assertEquals(scores[i], scc.mscores[i * 2], 0f);
       assertEquals(scores[i], scc.mscores[i * 2 + 1], 0f);
     }
 
+    // But we only call score() on the underlying scorer once per document, because scores are
+    // cached.
     assertEquals(scores.length, countingScorable.count);
   }
 }
--- a/lucene/core/src/test/org/apache/lucene/search/TestTopFieldCollector.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestTopFieldCollector.java
@@ -359,7 +359,7 @@ public class TestTopFieldCollector extends LuceneTestCase {
       leafCollector.collect(1);
 
       scorer.score = 4;
-      leafCollector.collect(1);
+      leafCollector.collect(2);
 
       TopDocs topDocs = collector.topDocs();
       assertEquals(totalHitsThreshold < 4, scorer.minCompetitiveScore != null);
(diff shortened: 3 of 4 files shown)


lucene/core/src/test/org/apache/lucene/search/TestPositiveScoresOnlyCollector.java
```diff
@@ -114,8 +114,9 @@ public void testNegativeScores() throws Exception {
     Collector c = new PositiveScoresOnlyCollector(tdc);
     LeafCollector ac = c.getLeafCollector(ir.leaves().get(0));
     ac.setScorer(s);
-    while (s.iterator().nextDoc() != DocIdSetIterator.NO_MORE_DOCS) {
-      ac.collect(0);
+    int docId;
+    while ((docId = s.iterator().nextDoc()) != DocIdSetIterator.NO_MORE_DOCS) {
+      ac.collect(docId);
```
mikemccand, line 119: Wow, good catch!  So this was a latent pre-existing test bug, and with your above bug fix this test is now (correctly!) failing, and with your fix to this separate test bug, the test now passes?

lucene/core/src/test/org/apache/lucene/search/TestTopFieldCollector.java
```diff
@@ -359,7 +359,7 @@ public void testTotalHitsWithScore() throws Exception {
       leafCollector.collect(1);
 
       scorer.score = 4;
-      leafCollector.collect(1);
+      leafCollector.collect(2);
```
mikemccand, line 362: Another pre-existing test bug fixed?

mikemccand (approved):
Phew, this was a great catch of a long standing bug ... I wonder how much net performance was lost across all Lucene applications that were using this class thinking it actually saved CPU by caching the score like it claims it does.  Perhaps this class used to work and then a refactoring (maybe the addition of `Scorable`?) accidentally introduced the bug ...

The change looks great to me, and it's sweet you also then found two separate pre-existing test bugs.  I'm hoping we can also add a dedicated test case confirming this class is actually doing its one job.

jpountz (commented):
Thanks for catching this! This suggests that the caching of scores did not actually work, can you add a test that fails when a score is computed twice for the same doc even though `ScoreCachingWrapperScorer` is used?

msfroh:
Luckily, this is a Lucene 10-only bug (from when `docId()` was removed from `Scorable`). 

I came across it when updating OpenSearch to support Lucene 10 and needed to refactor some code that used `ScoreCachingWrapperScorer`.

msfroh:
I think the caching worked fine, albeit in a funny way. 

When you call `collect()` with any valid doc ID, it invalidates the cache, causing scores to be computed. After the score was computed, `currDoc` was set to `-1`.  Repeated calls to `score()` (until the next `collect()`) will reuse the cached value because `-1 == -1`. 

As long as you don't call `collect()` on the same document multiple times (which you probably shouldn't be doing anyway -- those existing unit tests were a little weird), it should be okay.

lucene/core/src/test/org/apache/lucene/search/TestScoreCachingWrappingScorer.java
```diff
@@ -157,4 +157,40 @@ public void testGetScores() throws Exception {
...
+    CountingScorable countingScorable = new CountingScorable(s);
+    ScoreCachingCollector scc = new ScoreCachingCollector(scores.length * 2);
+    LeafCollector lc = scc.getLeafCollector(null);
+    lc.setScorer(countingScorable);
+
+    // We need to iterate on the scorer so that its doc() advances.
+    int doc;
+    while ((doc = s.iterator().nextDoc()) != DocIdSetIterator.NO_MORE_DOCS) {
+      lc.collect(doc);
+      lc.collect(doc);
```
jpountz, line 186: It's illegal to call collect() multiple times on the same doc, this shouldn't be necessary as `collect()` already calls `Scorable#score()` multiple times?

jpountz (approved):
I'm curious about the questions that @mikemccand raised, otherwise the change looks good to me.

lucene/core/src/test/org/apache/lucene/search/TestScoreCachingWrappingScorer.java
```diff
@@ -157,4 +157,40 @@ public void testGetScores() throws Exception {
     ir.close();
     directory.close();
   }
+
+  private static class CountingScorable extends FilterScorable {
+    int count = 0;
```
mikemccand, line 162: We don't need the `= 0` since it's java's default already?

lucene/core/src/test/org/apache/lucene/search/TestScoreCachingWrappingScorer.java
mikemccand, line 186: Hmm why are we collecting the same `doc` twice?  (I thought this is smelly -- we fixed the other two tests to stop doing that?).

Oh I see, is so `score()` happens more than once while on the same doc?  Actually, since `ScoreCachingCollector`'s `LeafCollector` is calling `.score()` three times for each `.collect()`, you should be able to do only one `lc.collect(doc)` here?

Does this new test case fail if you revert the bug fix?

lucene/core/src/test/org/apache/lucene/search/TestScoreCachingWrappingScorer.java
```diff
@@ -157,4 +157,40 @@ public void testGetScores() throws Exception {
...
+    public float score() throws IOException {
+      count++;
+      return in.score();
+    }
+  }
+
+  public void testRepeatedCollectReusesScore() throws Exception {
+    Scorer s = new SimpleScorer();
+    CountingScorable countingScorable = new CountingScorable(s);
+    ScoreCachingCollector scc = new ScoreCachingCollector(scores.length * 2);
```
mikemccand, line 178: Maybe add a comment about why we need `* 2` here?

lucene/core/src/test/org/apache/lucene/search/TestScoreCachingWrappingScorer.java
```diff
@@ -157,4 +157,40 @@ public void testGetScores() throws Exception {
...
+      lc.collect(doc);
+      lc.collect(doc);
+    }
+
+    for (int i = 0; i < scores.length; i++) {
+      assertEquals(scores[i], scc.mscores[i * 2], 0f);
+      assertEquals(scores[i], scc.mscores[i * 2 + 1], 0f);
+    }
+
+    assertEquals(scores.length, countingScorable.count);
```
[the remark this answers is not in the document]
mikemccand: Ahh I see -- you want to assert that indeed `score()` was called (up on top) multiple times per hit, but then down below in the "true" scorer, it was dedup'd to just once per hit.  So that's why you `.collect(doc)` twice, OK ...

mikemccand (commented):
Thanks @msfroh -- does the new test fail when you revert the bug fix?

lucene/core/src/test/org/apache/lucene/search/TestScoreCachingWrappingScorer.java
msfroh: The new test does fail if I revert the fix. 

As I mentioned in https://github.com/apache/lucene/pull/13987#issuecomment-2472044035, the condition for this being an actual bug is probably not really valid (i.e. you need to call `collect()` on the `ScoreCachingWrappingLeafCollector` multiple times for the same document).

The caching *does* work if the wrapped `LeafCollector` calls `score()` multiple times, which can happen in `MultiCollector` (for example). The layering is roughly:

```
ScoreCachingWrappingLeafCollector wraps
  Some other LeafCollector whose score is
    ScoreCachingWrappingScorer, which wraps
      The real Scorable
```

The multiple calls to `score()` from `Some other LeafCollector` work just fine. Again, it's the call to `collect()` on `ScoreCachingWrappingLeafCollector` that was invalidating the cache.

Those other two tests broke because my "fix" causes the caching to work (but they didn't expect it).

msfroh: Yeah -- that's why I'd hesitate to really call this a "bug fix". 

I'm pretty sure you meant to write `lastDoc = currDoc` in https://github.com/apache/lucene/pull/12407, but functionally `currDoc = lastDoc` works fine as long as you don't call `collect()` multiple times on the same doc.

I only noticed this because I was [copy/pasting the logic](https://github.com/opensearch-project/OpenSearch/pull/16366/files#diff-45fac6161bed1775163b916fba9ea296fbe9a8adf931a48593672c5c558a114b) to adapt it for `LeafBucketCollector` in OpenSearch and IntelliJ warned me that `lastDoc` could be `final`.

lucene/core/src/test/org/apache/lucene/search/TestTopFieldCollector.java
msfroh: With the change, the call to `leafCollector.collect(1)` would get a score of 3 (since it's cached for document 1), instead of the 4 that the test expected.

lucene/core/src/test/org/apache/lucene/search/TestScoreCachingWrappingScorer.java
jpountz: Thanks for explaining, then my suggestion would be to keep the test simple and not call collect() multiple times on the same doc. Your change looks good to me otherwise, so I can merge after that.

jpountz:
@msfroh FWIW I'm happy to merge this PR when we remove the double call to LeafCollector#collect on the same doc ID in tests.

msfroh:
> @msfroh FWIW I'm happy to merge this PR when we remove the double call to LeafCollector#collect on the same doc ID in tests.

In that case, the unit test that I added can be removed (since the double-call to LeafCollector#collect was the whole point).

The old code had the correct behavior if we don't call collect twice for the same doc. We can simplify/clarify it like https://github.com/apache/lucene/pull/14012 (which is functionally equivalent to what was there, but maybe a little clearer).

jpountz:
> In that case, the unit test that I added can be removed

This works for me. Sorry for putting you on the wrong track by suggesting that a test is added, it took me a while to understand how the existing logic was accidentally working.

msfroh:
Closing in favor of https://github.com/apache/lucene/pull/14012

Issue #13983, opened 2024-11-08
KnnFloatVectorQuery#toString should show the filter


jpountz:
I was looking at the `toString()` of a `KnnFloatVectorQuery` and got surprised that it was not pre-filtered, when I later discovered that it actually was, `toString()` was just omitting to display the pre-filter.

viswanathk:
Seems like a good first issue -  I can contribute this @jpountz.

benwtrent:
This is now fixed: https://github.com/apache/lucene/pull/13990

PR #13990, opened 2024-11-12
Adding filter to the toString() method of KnnFloatVectorQuery

viswanathk:
Addresses the issue #13983

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/KnnFloatVectorQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/KnnFloatVectorQuery.java
@@ -112,7 +112,14 @@ public class KnnFloatVectorQuery extends AbstractKnnVectorQuery {
 
   @Override
   public String toString(String field) {
-    return getClass().getSimpleName() + ":" + this.field + "[" + target[0] + ",...][" + k + "]";
+    StringBuilder buffer = new StringBuilder();
+    buffer.append(getClass().getSimpleName() + ":");
+    buffer.append(this.field + "[" + target[0] + ",...]");
+    buffer.append("[" + k + "]");
+    if (this.filter != null) {
+      buffer.append(",filter:[" + this.filter + "]");
+    }
+    return buffer.toString();
   }
 
   @Override
--- a/lucene/core/src/java/org/apache/lucene/search/KnnFloatVectorQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/KnnFloatVectorQuery.java
@@ -117,7 +117,7 @@ public class KnnFloatVectorQuery extends AbstractKnnVectorQuery {
     buffer.append(this.field + "[" + target[0] + ",...]");
     buffer.append("[" + k + "]");
     if (this.filter != null) {
-      buffer.append(",filter:[" + this.filter + "]");
+      buffer.append("[" + this.filter + "]");
     }
     return buffer.toString();
   }
--- a/lucene/core/src/test/org/apache/lucene/search/TestKnnFloatVectorQuery.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestKnnFloatVectorQuery.java
@@ -29,13 +29,7 @@ import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field;
 import org.apache.lucene.document.KnnFloatVectorField;
 import org.apache.lucene.document.StringField;
-import org.apache.lucene.index.DirectoryReader;
-import org.apache.lucene.index.IndexReader;
-import org.apache.lucene.index.IndexWriter;
-import org.apache.lucene.index.IndexWriterConfig;
-import org.apache.lucene.index.LeafReaderContext;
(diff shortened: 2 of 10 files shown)


lucene/core/src/test/org/apache/lucene/search/TestKnnFloatVectorQuery.java
```diff
@@ -29,13 +29,7 @@
...
 import org.apache.lucene.document.KnnFloatVectorField;
 import org.apache.lucene.document.StringField;
-import org.apache.lucene.index.DirectoryReader;
-import org.apache.lucene.index.IndexReader;
-import org.apache.lucene.index.IndexWriter;
-import org.apache.lucene.index.IndexWriterConfig;
-import org.apache.lucene.index.LeafReaderContext;
-import org.apache.lucene.index.QueryTimeout;
-import org.apache.lucene.index.VectorSimilarityFunction;
+import org.apache.lucene.index.*;
```
benwtrent, line 32: Let's not collapse to `*`

lucene/core/src/java/org/apache/lucene/search/KnnFloatVectorQuery.java
```diff
@@ -112,7 +112,14 @@ VectorScorer createVectorScorer(LeafReaderContext context, FieldInfo fi) throws
...
   @Override
   public String toString(String field) {
-    return getClass().getSimpleName() + ":" + this.field + "[" + target[0] + ",...][" + k + "]";
+    StringBuilder buffer = new StringBuilder();
+    buffer.append(getClass().getSimpleName() + ":");
+    buffer.append(this.field + "[" + target[0] + ",...]");
+    buffer.append("[" + k + "]");
+    if (this.filter != null) {
+      buffer.append(",filter:[" + this.filter + "]");
+    }
```
benwtrent, line 121:
```suggestion
    if (this.filter != null) {
      buffer.append("[" + this.filter + "]");
    }
```

I think that would be better.

benwtrent (changes requested):
Could you update the byte knn query & DiversifyingChildern* knn queries as well?

viswanathk:
> Could you update the byte knn query & DiversifyingChildern* knn queries as well?

I made the changes fo DiversifyingChildren*, but by byteknn do you mean `ByteVectorSimilarityQuery`?

jpountz:
I believe that @benwtrent meant `KnnByteVectorQuery`.

viswanathk:
I hope I got all of them now.

lucene/join/src/java/org/apache/lucene/search/join/DiversifyingChildrenByteKnnVectorQuery.java
```diff
@@ -154,7 +154,14 @@ protected TopDocs approximateSearch(
 
   @Override
   public String toString(String field) {
-    return getClass().getSimpleName() + ":" + this.field + "[" + query[0] + ",...][" + k + "]";
```
benwtrent, line 157: test updates for this and the float version? The tests are in: TestParentBlockJoinByteKnnVectorQuery and of course the float version as well.

viswanathk: yeah, i did check them. they didn't have existing tests that cover `toString` so skipped them. will add that coverage though.

viswanathk: @benwtrent for the Diversifying* does it make sense to capture the `childFilter` and `parentsFilter` in toString() as well?

jpountz: It's not obvious to me, `parentsFilter` is more a property of the index than of the query. You could have multiple levels of parents, which would make this property relevant to print, but at the same time I don't think they have proper toString() implementations. Let's leave it out for now.

jpountz (approved):

jpountz:
> Could you update the byte knn query & DiversifyingChildern* knn queries as well?

Unrelated: I noticed that the `DiversifyingChildren*` queries don't use the filter to evaluate the query, is this a bug?

benwtrent:
> Unrelated: I noticed that the DiversifyingChildren* queries don't use the filter to evaluate the query, is this a bug?

I am unsure what you mean by this? The filter is utilized when building the accepted docs set in the super() classes.

vigyasharma (approved):

benwtrent (approved):
I think a `CHANGES` entry is in order. This seems like a nice little bug fix to aid folks in debugging issues.

@viswanathk once you add the changes entry, I can merge and backport to 10.x

viswanathk:
> I think a `CHANGES` entry is in order. This seems like a nice little bug fix to aid folks in debugging issues.
> 
> @viswanathk once you add the changes entry, I can merge and backport to 10.x

@benwtrent Added in `CHANGES` for 10.1. Also corrected some git history mess up.

PR #13991, opened 2024-11-13
Add a Multi-Vector Similarity Function

vigyasharma:
This is a small first change towards adding support for multi-vectors. We start with adding a `MultiVectorSimilarityFunction` that can handle (late) interaction across multiple vector values.

This is the first of a series of splits for the larger prototype PR #13525


vigyasharma:
I am thinking we can leverage the `NONE` aggregation (in #13525) for non-ColBERT passage vector use-cases, by making each graph node correspond to a single value in the multi-vector i.e. index time aggregation becomes "none". The resultant graph would be similar to what we construct with parent-child docs today, while flat storage with multi-vectors could allow for aggregated similarity checks at *query* time. This could help with recall while making mutli-vector usage easier to use (no overquery or index time joins).

It'll need some work: a mechanism to address each vector value directly, and corresponding changes in VectorValues. I'm thinking maybe an "ordinal" for the multi-vector, and a "sub-ordinal" for values within the multi-vector. Both ints can be packed into a long for node value?

Since I haven't chalked out all the details yet, I decided to **remove** the `NONE` aggregation for now, and focus first on the ColBERT use case. Will go with the `isMultiVector` flag in FieldInfos to identify multi-vector storage requirements.

cc: @cpoerschke , @benwtrent

lucene/core/src/java/org/apache/lucene/index/MultiVectorSimilarityFunction.java
```diff
@@ -0,0 +1,202 @@
...
+        }
+        for (int i = 0; i < inner.length; i += dimension) {
+          innerList.add(ArrayUtil.copyOfSubArray(inner, i, i + dimension));
+        }
+
+        float result = 0f;
+        for (float[] o : outerList) {
+          float maxSim = Float.MIN_VALUE;
+          for (float[] i : innerList) {
+            maxSim = Float.max(maxSim, vectorSimilarityFunction.compare(o, i));
```
dungba88, line 64: If we add another `compare` method with start and end indexes for both inner and outer, I guess we won't need to copy the array?

lucene/core/src/java/org/apache/lucene/index/MultiVectorSimilarityFunction.java
```diff
@@ -0,0 +1,202 @@
...
+          }
+          result += maxSim;
+        }
+        return result;
+      }
+
+      @Override
+      public float aggregate(
```
dungba88, line 72: It's so unfortunate that Java doesn't support generic primitive array and has to have duplicate code.

lucene/core/src/java/org/apache/lucene/index/MultiVectorSimilarityFunction.java
vigyasharma: Right, but it needs to go all the way down to `VectorUtilSupport`, which I think should be a PR of its own.

PR #13994, opened 2024-11-14
Speed up top-k retrieval on filtered conjunctions.

jpountz:
A while back we added an optimized bulk scorer that implements block-max AND, this yielded a good speedup on nightly benchmarks, see annotation `FP` at https://benchmarks.mikemccandless.com/AndHighHigh.html. With this PR, filtered conjunctions now also run through this optimized bulk scorer by doing two things:
 - It flattens inner conjunctions. This makes queries initially written as something like `+(+term1 +term2) #filter` rewritten to `+term1 +term2 #filter`.
 - It evaluates queries that have a mix of MUST and FILTER clauses evaluated through `BlockMaxConjunctionBulkScorer` by treating FILTER clauses as scoring clauses that produce a score of 0.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
@@ -268,6 +268,11 @@ public class BooleanQuery extends Query implements Iterable<BooleanClause> {
       return new MatchNoDocsQuery("empty BooleanQuery");
     }
 
+    // Queries with no positive clauses have no matches
+    if (clauses.size() == clauseSets.get(Occur.MUST_NOT).size()) {
+      return new MatchNoDocsQuery("pure negative BooleanQuery");
+    }
+
     // optimize 1-clause queries
     if (clauses.size() == 1) {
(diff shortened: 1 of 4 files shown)


lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
```diff
@@ -558,6 +557,41 @@ public Query rewrite(IndexSearcher indexSearcher) throws IOException {
...
+                builder.add(innerClause.query(), Occur.FILTER);
+              }
+            }
+          } else {
+            builder.add(clause);
+          }
+        } else {
+          builder.add(clause);
+        }
+      }
```
benwtrent, line 589: Holy smokes I had to read this like 5 times to fully grok what it is doing. 

Could you name `occur` to `innerClauseOccur` and change `clause` to `outerClause`?

benwtrent (approved):
The inlining and optimization makes sense to me. Minor thoughts around naming things differently so that the flattening logic is easier to follow.

jpountz:
Thanks @benwtrent !

PR #13999, opened 2024-11-15
Make CombinedFieldQuery eligible for WAND/MAXSCORE.

jpountz:
`CombinedFieldQuery` currently returns an infinite maximum score. We can do better by returning the maximum score that the sim scorer can return, which in the case of BM25 is bounded by the IDF. This makes CombinedFieldQuery eligible for WAND/MAXSCORE (not their block-max variants though, since we return the same score upper bound for the whole index).

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -99,6 +99,9 @@ Optimizations
 * GITHUB#13985: Introduces IndexInput#updateReadAdvice to change the ReadAdvice
   while merging vectors. (Tejas Shah)
 
+* GITHUB#13999: CombinedFieldQuery now returns non-infinite maximum scores,
+  making it eligible to dynamic pruning. (Adrien Grand)
+
 Bug Fixes
 ---------------------
 * GITHUB#13832: Fixed an issue where the DefaultPassageFormatter.format method did not format passages as intended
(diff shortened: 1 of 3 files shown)


ChrisHegarty (approved):
LGTM

jpountz:
Interestingly, it looks like this change caused a significant regression to top-level disjunctions, see e.g. https://benchmarks.mikemccandless.com/OrHighHigh.html.

It seems to me that this is due to the fact that `MaxScoreBulkScorer#scoreInnerWindowAsConjunction` now sometimes runs with iterators implemented as `DisjunctionDISIApproximation` while it only ran with iterators implemented as `BlockImpactsDocsEnum` before this change. So `nextDoc()` / `advance()` become bimorphic (or maybe even polymorthic I'm not 100% sure if we went from 1 to 2 or from 2 to 3).

I won't revert since it's a pre-existing issue, we would have seen the same earlier if we ran more disjunctions of <something else than term queries> in nightly benchmarks. It would be interesting to look into mitigating this sort of slowdown though.

jpountz:
So it went from bimorphic to polymorphic indeed, we had 2 iterator impls: `Lucene101PostingsReader$BlockImpactsDocsEnum` and `SlowImpactsEnum` and we added `DisjunctionDISIApproximation`. Removing `SlowImpactsEnum` should be easy, I'll look into it.

jpountz:
I opened #14017.

PR #14000, opened 2024-11-18
Only consider clauses whose cost is less than the lead cost to compute block boundaries in WANDScorer.

jpountz:
WANDScorer implements block-max WAND and needs to recompute score upper bounds whenever it moves to a different block. Thus it's important for these blocks to be large enough to avoid re-computing score upper bounds over and over again. With this commit, WANDScorer no longer uses clauses whose cost is higher than the cost of the filter to compute block boundaries. This effectively makes blocks larger when the filter is more selective.


benwtrent (approved):
Such a simple change, but nice results!

Could you add a CHANGES?

jpountz:
Will do, thanks @benwtrent!

PR #14009, opened 2024-11-22
Add Query for reranking KnnFloatVectorQuery with full-precision vectors

### LINKED ISSUE #13564: Add refinement of quantized vector scores with fp distance calculations
### Description

With quantization techniques that are compressing vectors in memory further and further, because of how much information is lost, recall is going to drop. However, with the current quantization support, we already store the full precision float vector values. With this, we could create a two phased search process that oversamples vectors from the quantized index (i.e. r\*k results), and then lazily loads vectors from disk for the r\*k results, re-scoring the vectors. This has been discussed directly or indirectly in a couple different places, but I figured itd make sense to create a separate issue for it:
* #13251 
* #13468
* #12615

It has been shown to work with binary compression techniques and some PQ in a few different places (https://medium.com/qdrant/hyperfast-vector-search-with-binary-quantization-865052c5c259, https://huggingface.co/blog/embedding-quantization#binary-rescoring). We also have done some experiments in OpenSearch with IVFPQ and re-scoring and its shown pretty strong promise (assuming that the disk IOPS/throughput are strong enough - see https://github.com/opensearch-project/k-NN/issues/1779#user-content-appendix-b-baseline-rescore-experiments).

That being said, it can provide similar benefits to the DiskANN approach.

One challenge I foresee is the approach requires the quantized ANN index to be resident in memory. However, I am unsure if loading the full precision vectors from disk will cause the page cache to evict the quantized ANN index or if the page cache will naturally adapt to the access pattern. If it is the former, we would probably need some way to pin the quantized ANN index and its vectors in memory.

In the future, it could probably be fine tuned quite a bit with access pattern optimizations and/or some kind of hierarchical quantization and re-scoring.

dungba88:
### Description

fixes #13564 

Added a new Query which wraps around KnnFloatVectorQuery and does re-ranking for quantized index using full precision vectors. The idea is to first run KnnFloatVectorQuery with over-sampled k (x1.5, x2, x5, etc) and then re-rank the docs using full-precision (original, non-quantized) vector, and finally take top-k.

Questions:
- Should we expose the `target` inside `KnnFloatVectorQuery` so that users don't need to pass the target twice? Currently it only exposes the `getTargetCopy()` which requires array copy so it's inefficient, but I assume the intention is to encapsulate the array so that it won't be modified from outside?
- Maybe out of scope for this PR, but I'm curious how people think about using `mlock` for preventing the quantized vectors from being swapped out, as loading fp vectors (although only a small set per query) means there will more pressure on RAM.

Usage:
```
KnnFloatVectorQuery knnQuery = ...; // create the KnnFloatVectorQuery with some over-sampled k
RerankKnnFloatVectorQuery query = new RerankKnnFloatVectorQuery(knnQuery, targetVector, k);
TopDocs topDocs = searcher.search(query, k);
```


dungba88:
The build fails with `The import org.apache.lucene.codecs.lucene100 cannot be resolved`, I thought this is already in mainline. Will check.

dungba88:
I have a preliminary benchmark here (top-k=100, fanout=0) using Cohere 768 dataset.

image

Anyhow I can see these 2 things that should be addressed:
- If we access the full-sized vectors, it will swap the memory that is allocated (either through preloading, or through mmap) for quantized vectors (main search phase) if there's not enough memory. Eventually, some % part of the quantized index will be swapped out which will slower the search. If we have to load all full-precision vectors to memory, then that kinda defeats the purpose of quantization. I'm wondering if there could be a way we can access full-precision vectors without interfering with the space of quantized vectors.
- The latency could be better. With oversample=1.5 (second dot) for 4_bit, we have around the same latency and recall as baseline. Although one can argue that we can save memory compared to baseline, with new access pattern of two-phase search that saving might be diminished. Otherwise it seems to have little benefit over just using plain HNSW.

lucene/core/src/java/org/apache/lucene/search/RerankKnnFloatVectorQuery.java
```diff
@@ -0,0 +1,117 @@
...
+  public Query rewrite(IndexSearcher indexSearcher) throws IOException {
+    IndexReader reader = indexSearcher.getIndexReader();
+    Query rewritten = indexSearcher.rewrite(query);
+    // short-circuit: don't re-rank if we already got all possible results
+    if (query.getK() <= k) {
+      return rewritten;
+    }
+    Weight weight = indexSearcher.createWeight(rewritten, ScoreMode.COMPLETE_NO_SCORES, 1.0f);
+    HitQueue queue = new HitQueue(k, false);
+    for (var leaf : reader.leaves()) {
```
shatejas, line 64: Should this be switched to parallel execution similar to [AbstractKnnVectorQuery](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L93)?

dungba88:
I reran the benchmark, this time comparing the two-phase search against no-reranking as well. It seems the re-ranking phase only adds a trivial amount of latency so that's a good thing.

image

Benchmark parameters:
- topk = 100
- fanout = 0
- Datasets: Cohere 768

lucene/core/src/java/org/apache/lucene/search/RerankKnnFloatVectorQuery.java
dungba88: Good question, I was using single-thread as a simple version and try to benchmark the latency first, since multi-thread could add some overhead as well. This class only does vector loading and similarity computation for a small set of vectors (k * oversample * num_segments) so it's not as CPU-intensive as the [AbstractKnnVectorQuery](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L93)

I'll change this to multi-thread and run the benchmark again. From the below benchmark, the re-ranking phase only adds a trivial amount of latency it might not help much. Also the benchmark code seems to force merge so there's only a single partition, we need to change so that there are multiple partitions.

dungba88:
Also this is the luceneutil branch I used for benchmarking: https://github.com/dungba88/luceneutil/tree/dungba88/two-phase-search, which incorporates the test for BQ implementation by @benwtrent and the two-phase search.

lucene/core/src/test/org/apache/lucene/search/TestRerankKnnFloatVectorQuery.java
```diff
@@ -0,0 +1,124 @@
...
+        int id = retrievedDoc.getField("id").numericValue().intValue();
+        float[] docVector = vectors.get(id);
+        assert docVector != null : "Vector for id " + id + " not found";
+        float expectedScore = VECTOR_SIMILARITY_FUNCTION.compare(targetVector, docVector);
+        Assert.assertEquals(
+            "Score does not match expected similarity for doc ord: " + scoreDoc.doc + ", id: " + id,
+            expectedScore,
+            scoreDoc.score,
+            1e-5);
+      }
```
huynmg, line 113: We can test that the results are sorted by exact distance.

Maybe we can also test that the result of the same query with oversample will be "at lease the same or better" than without oversample ? By "better" I mean we should have higher recall. But I'm not sure if it's deterministic

lucene/core/src/java/org/apache/lucene/search/RerankKnnFloatVectorQuery.java
shubhamvishu, line 64: Here we have the access to [IndexSearcher#getTaskExecutor](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java#L1150-L1156) and could use it to parallelize the work across segments(like we did earlier with some other query rewrites). But the `HitQueue` here isn't thread-safe. I don't know if using concurrency after making `insertWithOverflow` thread-safe would be really helpful? Might be?

dungba88: That's right. In order to apply parallelism we need to use a per-segment queue, then merge it like in [AbstractKnnVectorQuery.mergeLeafResults](https://code.amazon.com/packages/Lucene/blobs/c059819a59a99873081833c0f957452ae4be70f0/--/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L231). I think the added latency is already low, but still want to try if it helps.

mikemccand:
I think this is a nice overall approach, adding a new `RerankKnnFloatVectorQuery` that wraps a KNN query that used quantization to get the initial results.

It's reminiscent of Lucene's existing `QueryRescorer`, to implement multi-phased ranking, except that class doesn't wrap another `Query`... maybe it should (separately)!

I'm curious about [your results here](https://github.com/apache/lucene/pull/14009#issuecomment-2502665806) -- why is recall better for 1bit and 4bit than 7bit, when reranking?

dungba88:
> I'm curious about https://github.com/apache/lucene/pull/14009#issuecomment-2502665806 -- why is recall better for 1bit and 4bit than 7bit, when reranking?

If we compare the recall with the same oversample, then 7-bit is always better or same. The difference becomes smaller at higher oversample. E.g, at oversample=1 , 7 bit has 20% higher recall than 1-bit but at oversample=1 they are mostly the same.

vigyasharma:
> The graph is a bit confusing, but the dots are the oversample (from 1 to 5).

This is the piece I was missing! Can we edit the [benchmark comment](https://github.com/apache/lucene/pull/14009#issuecomment-2502665806) and add a line saying each dot represent oversample factor (from 1 to 6, i count 6 dots).

So 4x oversample with 1_bit outperforms 2x oversample with 7_bit, in both latency and recall. And 3x oversample with 4_bit has better recall and latency than 7_bit with no oversample.

Sorry I'm late to the party. Will look at the PR and benchmark more closely and try to think of latency laggers.

vigyasharma (commented):
This is excellent functionality! The ability to **rerank results from a faster, lower-precision initial query with a higher-quality second-pass scoring** is a big enabler.

I wonder if we should generalize this concept more broadly, decoupling it from `KnnFloatVectorQuery.` How about introducing a new query type, perhaps `RerankQuery` or `LateInteractionQuery?` This query would accept `query,` `field,` `target,` and `k` as inputs. It would then score hits from the inner query using floating-point vectors from the specified field and target, returning the top-k results. This new query would manage its own weight and scorer, utilizing the scorer from the inner query. Importantly, the inner `query` wouldn't even need to be a vector query, although it likely often would be.

This generalized approach could also pave the way for future extensions, such as late interaction models where the inner query uses single-value vectors, while the outer query leverages a multi-vector field and a multi-vector target representation of the query. 

Even in its basic form, this generalized query would achieve what this PR currently implements.

dungba88:
Thanks @vigyasharma for the comment! I updated the comment to make it less confusing.

I'll think about generalization, but the idea is that as long as the field can expose the float vectors through `getFloatVectorValues(fieldInfo)` then it can be used with this Query? Or do you think we need to abstract/generalize this part as well?

> utilizing the scorer from the inner query

I can't use the inner scorer here because for quantized codec it will be the quantized vector scorer and only use the quantized vectors.

vigyasharma:
@dungba88 : I shared an [alternate implementation](https://github.com/apache/lucene/pull/14698) for RerankVectorQuery that I feel can generalize more broadly. 

**Main Differences:**
1. While the primary use case would be to rerank results from a quantized `KnnFloatVectorQuery` against full precision vectors, this approach has no tight coupling. Users could provide any inner query and rerank against vectors stored in any field.
2. This can, in future, be extended to support reranking using Late Interaction models stored in a different field. We just need to override add some checks to ensure multi-vectors in provided `field` are compatible with provided `target`. This idea has been floated [here](https://github.com/apache/lucene/pull/13525#issuecomment-2445295372) earlier
3. Maintains a regular query flow of keeping heavier computation in Weight and Scorer, instead of heavy matching and scoring logic in rewrite.
4. This should implicitly use any slicing or per leaf parallel execution provided at searchers. We don't need to rewrite logic for per leaf threads.

Please know that I don't mean to override your work, just needed to write it down to clarify my ideas.  I'm happy to collaborate on whatever direction makes most sense.

dungba88:
No worry, I think generalization makes sense to me too. And I like the idea of moving this forward.

The use of `KnnFloatVectorQuery` in this implementation only adds a small benefit that we can short-circuit when the results are less than k, which is marginal to be honest. Totally agree that we can make it a generic Query.

I'm not sure about the weight part though. Asking just for my understanding of Lucene, I thought Lucene typically rewrites the Query first, and then createWeight on the simplest form. Here we are doing the reverse (createWeight, and inside that call rewrite). Can this break any assumption (e.g caching) in Lucene?

dungba88:
I changed the query to be generic but keep other the same. (Didn't have JDK24 yet so build would fail.)

msokolov:
I haven't dug in deep yet, but along the lines of further generality, would it make sense to accept a DoubleValuesSource (that can compute dot products against a vector field)  rather than limiting this to vector dot products only?

vigyasharma:
> Lucene typically rewrites the Query first, and then createWeight on the simplest form.

That's a good callout. I'll override the `rewrite` method to rewrite the inner query. That will keep the query cacheable. If inner query rewrite is only called in `createWeight` then each weight can get a different set of knn hits which goes against caching.

> we no longer limit the final results by K, so if the inner queries is doing over-sampling, the rerank queries will return the full over-sampled results as well.

This gets handled by the `k` you pass to searcher.search(), and the top docs collector you use. It's same as any other query.

> would it make sense to accept a DoubleValuesSource (that can compute dot products against a vector field) rather than limiting this to vector dot products only

interesting idea, let me see what that would look like.

msokolov:
Is it this? https://github.com/apache/lucene/blob/main/lucene/queries/src/java/org/apache/lucene/queries/function/FunctionScoreQuery.java

dungba88:
> This gets handled by the k you pass to searcher.search(), and the top docs collector you use. It's same as any other query.

I guess the main difference is when the results got trimmed down. For the alternative implementation we leave that up to the final hit collection to return the top-k based on a ranking mechanism (as this Query is usually just part of the final Query we send to IndexSearcher) while this implementation of the Query would output top-k by its own. If the final score is determined independently from the score returned by this Query, then the re-ranking we did would be a waste (and we can just over-sample and let the final scorer to do the work)?

With that being said, I have no strong opinion. Wondering what would make sense for the actual use case that it will use (quantization and late-interaction model). We can also use special value (like negative number) to tell us not to trim down the results if we do need both way.

lucene/core/src/java/org/apache/lucene/search/RerankFloatVectorQuery.java
```diff
@@ -0,0 +1,116 @@
...
+      }
+      FieldInfo fi = leaf.reader().getFieldInfos().fieldInfo(field);
+      if (fi == null) {
+        continue;
+      }
+      VectorSimilarityFunction comparer = fi.getVectorSimilarityFunction();
+      DocIdSetIterator iterator = scorer.iterator();
+      while (iterator.nextDoc() != DocIdSetIterator.NO_MORE_DOCS) {
+        int docId = iterator.docID();
+        float[] vectorValue = floatVectorValues.vectorValue(docId);
```
vigyasharma, line 78: This will only work for dense vectors where `docId == ordinal` I think. For sparse vectors, we need to pass the ordinal for this docId. You can use a DocIndexIterator, see [PR diff](https://github.com/apache/lucene/pull/14698/files#diff-6af08c57495bc31ea199e84da23b8804569f025f843169379aecdf16b5a11ba5R86-R117) for reference

vigyasharma:
> For the alternative implementation we leave that up to the final hit collection to return the top-k based on a ranking mechanism

Those final TopDocs collectors will use the score generated by the Rerank query, which is the rerank mechanism we are implementing.

I'm not really sure I follow the concern here. In both cases, you need to score all the docs to find top-K. In this one, you do it during rewrite and maintain your own HitQueue. In the other, you do it in Scorer and leverage TopDocs collector or whatever collector is needed. The alternate approach is more flexible, you can vary how many hits you want to work with (upto oversample limit). But computationally, it's not more expensive.

lucene/core/src/java/org/apache/lucene/search/RerankFloatVectorQuery.java
dungba88: Thanks for the pointer.

dungba88:
Let me rephrase it better. I'm not worrying about the computation cost, as you mentioned they are the same.

In Amazon (where both Vigya and I are working at), we usually run KnnFloatVectorQuery along with keyword matching (such as TermQuery) disjunctively amongst with other constraint filtering. We don't use the scoring in the KnnFloatVectorQuery as well as TermQuery (as we will wrap them with ConstantScoreQuery), and after we get all the hits from both KNN and keyword matching we will use our own scorer to compute the final top-k score. One of the thing I'm hoping to do here is for KnnFloatVectorQuery with low-bit quantization, we can get a better recall with over-sampling and re-ranking with the full-precision vector and cut-off the hits to the original top-k. The end result is that the KnnFloatVectorQuery will still output top-k results but with better recall due to the oversample and reranking. Moreover we don't have to rely on the final scorer to remove/downrank the defects which maybe introduced with the oversampling. If we keep the full over-sample during the Query match phase and only get the final top-k when collecting hits at the end as the alternative approach, then:
- Wouldn't the `Scorer.score()` be skipped/not called at all and the RerankQuery will essentially be no-op? Asking just for my education.
- It would not possible (due to the how Weight/Scorer are used) to limit the hits to a fixed K for the RerankQuery alone, before applying it conjunctively/disjunctively with other Query (similar to how AbstractKnnVectorQuery works), wouldn't it? Then it might not be possible to apply for this use case.

If we allow a way to flexibly choose either to trim or not the results from RerankQuery, do you think that would be better and serve both use cases we have here?

vigyasharma:
> Is it this? https://github.com/apache/lucene/blob/main/lucene/queries/src/java/org/apache/lucene/queries/function/FunctionScoreQuery.java

@msokolov : Gosh, you're right! This is pretty much what we need. Just pass it a `DoubleValuesSource` for full precision vector scoring, along with the query you want to wrap.

I noticed we already had `DoubleValuesSource`s for vector similarity, but they didn't compute full precision for quantized vectors. Have added that support here – #14708 

Perhaps, we could also add some syntactic sugar on the FunctionScoreQuery for this specific use case, maybe a static getter that uses these double values sources to return a query that reranks by full precision vector similarity.

dungba88:
I'm thinking that instead of using `vectorValues(...)` we can probably use DoubleValueSource as per @msokolov suggestion for a generic rescorer. Then we can use scorer from another (vector) field for re-ranking. This could help with the use case to e.g use 1-bit quantization for matching and 7-bit quantization for re-ranking by indexing an additional fields.

Either way we would not be able to reuse the `FunctionScoreQuery` due to the fact that we are using `ConstantScoreQuery` and we need to trim the returned matches from KNN alone to just K.

dungba88:
I published another revision to support a custom DoubleValueSource for re-ranking, and we can reuse the `FloatVectorSimilarityValuesSource` in case we want to rescore with other field, or the `FullPrecisionFloatVectorSimilarityValuesSource` in https://github.com/apache/lucene/pull/14708.

I also checked the `Rescorer` in https://github.com/apache/lucene/pull/14776, but it seems that will only affect the final results, and may not work when there are hybrid of keyword and semantic (KNN) matching and that we only need to re-score the KNN part.

lucene/CHANGES.txt
```diff
@@ -857,7 +858,7 @@ Improvements
 
 * GITHUB#13285: Early terminate graph searches of AbstractVectorSimilarityQuery to follow timeout set from
   IndexSearcher#setTimeout(QueryTimeout). (Kaival Parikh)
-
```
dungba88: Seems like my IDE automatically removes extra spaces. If there is objection I'll revert that in next rev, along with other feedbacks.

lucene/CHANGES.txt
```diff
@@ -2453,7 +2454,7 @@ New Features
 * LUCENE-10385: Implement Weight#count on IndexSortSortedNumericDocValuesRangeQuery
   to speed up computing the number of hits when possible. (Lu Xugang, Luca Cavanna, Adrien Grand)
 
-* LUCENE-10422: Monitor Improvements: `Monitor` can use a custom `Directory` 
+* LUCENE-10422: Monitor Improvements: `Monitor` can use a custom `Directory`
```
vigyasharma: Looks like a lot of unrelated changes, probably needs a merge from main?

lucene/core/src/java/org/apache/lucene/search/RescoreTopNQuery.java
```diff
@@ -0,0 +1,128 @@
...
+ * limitations under the License.
+ */
+package org.apache.lucene.search;
+
+import java.io.IOException;
+import java.util.Objects;
+import org.apache.lucene.index.IndexReader;
+
+/**
+ * A Query that re-scores another Query with a DoubleValueSource function and cut-off the results at
```
vigyasharma: This sounds generic to any query, but rewrite always returns a `KnnFloatVectorQuery` ?

lucene/core/src/java/org/apache/lucene/search/RescoreTopNQuery.java
```diff
@@ -0,0 +1,128 @@
...
+ * @lucene.experimental
+ */
+public class RescoreTopNQuery extends Query {
+
+  private final int n;
+  private final Query query;
+  private final DoubleValuesSource valuesSource;
+
+  /**
+   * Execute the inner Query, re-score using a customizable DoubleValueSource and trim down the
```
vigyasharma, line 40: What capabilities does this query add over `FunctionScoreQuery`? It seems to do the exact same thing, only it doesn't rescore in the rewrite phase.

lucene/CHANGES.txt
dungba88: It's due to my IDE automatically remove extra white spaces, will revert in next rev.

lucene/core/src/java/org/apache/lucene/search/RescoreTopNQuery.java
dungba88: The `createRewrittenQuery` method would return a `DocAndScoreQuery` which is currently internal to KNN, but from the client/API point of view it's the same as any other Query. Moreover it only contains the docId and respective scores.

We can extract `createRewrittenQuery` into separate class for more reusability if needed. I can put a prerequisite PR if that makes sense.

lucene/core/src/java/org/apache/lucene/search/RescoreTopNQuery.java
dungba88: That's one of them, it does not depend on the score mode (`FunctionScoreQuery` will be a no-op if `needsScores = false`) and it also trim down the results to N. 

So it's more similar to `Rescorer` class, except that it can rescore a sub-Query instead of top-level Query in the final `IndexSearcher.search()`.

The nuance is important in cases where we don't need the overall score (using ConstantScoreQuery) but still want to rescore and trim down results in some sub-Query (like KnnFloatVectorQuery).

It's heavily influenced by KnnFloatVectorQuery but API-wise there would be no such limitation.

vigyasharma:
Thanks for the explanations, @dungba88. I suppose the scenario you're trying to solve for, is when users want to change the matchset of a KnnVectorQuery using full-precision or other reranking. I'm open to this change if it's a valid requirement. 

My perception is that it should be solvable by plumbing vector query results into a rescorer, then combining top-K hits with other (lexical) hits. For e.g. is this also a problem for hybrid search in OpenSearch/Elasticsearch ? I suspect they might have independent queries for both with some way to combine results.

dungba88:
> is when users want to change the matchset of a KnnVectorQuery using full-precision or other reranking

Yes that's correct, @vigyasharma. We are using a hybrid search where KnnFloatVectorQuery and TermQuery (amongst others) are combined into a single BooleanQuery. Thus it is important to change the matchset of KnnFloatVectorQuery individually.

For hybrid search in OpenSearch/Elastic Search, I'm wondering if @jmazanec15 and @benwtrent have any input. I'm having a feeling that it's quite common to combine lexical + KNN matching into a single BooleanQuery.

benwtrent:
kNN queries are completed in the rewrite phase, if any rescoring needs to be done, it should be done during that full phase. 

I would expect the experience to be:

`RescoreQueryWithVectorQueryThingy(KnnQuery)` and the rescore will occur during rewrite (or atleast provide a scorer that iterates the kNN query results calculating the higher fidelity scores).

kNN should be "just another query" and should be combinable with any other query. I realize this is a bit tricky as kNN is unique in that it effectively "collects" its results up-front.

lucene/core/src/java/org/apache/lucene/search/RescoreTopNQuery.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+      }
+    }
+    int i = 0;
+    ScoreDoc[] scoreDocs = new ScoreDoc[queue.size()];
+    for (ScoreDoc topDoc : queue) {
+      scoreDocs[i++] = topDoc;
+    }
+    TopDocs topDocs =
+        new TopDocs(new TotalHits(queue.size(), TotalHits.Relation.EQUAL_TO), scoreDocs);
```
vigyasharma: Instead of setting this to configured `n`, should we retain the total no. of hits and relation from original query?

lucene/core/src/java/org/apache/lucene/search/RescoreTopNQuery.java
```diff
@@ -0,0 +1,128 @@
...
+    return KnnFloatVectorQuery.createRewrittenQuery(reader, topDocs, 0);
```
vigyasharma: Hmm, so while this is called `KnnFloatVectorQuery.createRewrittenQuery`, it isn't specific to vector queries, right? It just creates a new `DocAndScoreQuery`. Should we rename the method (or create a different one) to fit better with this generic query?

lucene/core/src/java/org/apache/lucene/search/RescoreTopNQuery.java
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * @lucene.experimental
+ */
+public class RescoreTopNQuery extends Query {
+
+  private final int n;
+  private final Query query;
+  private final DoubleValuesSource valuesSource;
+
+  /**
```
vigyasharma, line 39: This query will override the `rewrite` phase of any provided query to gather all hits, find top N based on provided DoubleValuesSource, and return a DocAndScoreQuery. This only applies to vector search queries right now, but in spirit is generic to all approx search queries (not sure what other candidates might use it)?

I think we should call this out in the query doc string, to disambiguate what this query does v/s other rescoring methods like `Rescorer`s and `FunctionScoreQuery`.

lucene/core/src/test/org/apache/lucene/search/TestRescoreTopNQuery.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    // Step 2: Run TwoPhaseKnnVectorQuery with a random target vector
+    try (IndexReader reader = DirectoryReader.open(directory)) {
+      IndexSearcher searcher = new IndexSearcher(reader);
+      float[] targetVector = randomFloatVector(VECTOR_DIMENSION, random);
+      int k = 10;
+      double oversample = random.nextFloat(1.5f, 3.0f);
+
+      FloatVectorSimilarityValuesSource valueSource =
```
vigyasharma: Can we test it against the [FullPrecisionValuesSource](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/FullPrecisionFloatVectorSimilarityValuesSource.java) instead? I understand this predates us merging the FP values source but let's update it now? Otherwise this is essentially a no-op, the KnnFloatVectorQuery was already scored by quantized scores, and we used the same scores again.

lucene/core/src/test/org/apache/lucene/search/TestRescoreTopNQuery.java
```diff
@@ -0,0 +1,140 @@
...
+  public void testInvalidN() {
+    expectThrows(
+        IllegalArgumentException.class,
+        () ->
+            new RescoreTopNQuery(
+                new TermQuery(new Term("test")), DoubleValuesSource.constant(0), 0));
+  }
+
+  @Test
+  public void testRescoreField() throws Exception {
```
vigyasharma, line 75: Can we also add a test for when the document is missing from double values source? I think we should get a `0f` score based on your query?

lucene/core/src/java/org/apache/lucene/search/RescoreTopNQuery.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.Objects;
+import org.apache.lucene.index.IndexReader;
+
+/**
+ * A Query that re-scores another Query with a DoubleValueSource function and cut-off the results at
+ * top N.
+ *
+ * @lucene.experimental
+ */
+public class RescoreTopNQuery extends Query {
```
vigyasharma, line 33: How about we add a static function to create a RescoreTopNQuery for full precision vector rescoring? It would accept the relevant args, internally create a [FullPrecisionFloatVectorSimilarityValuesSource](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/FullPrecisionFloatVectorSimilarityValuesSource.java), and use it to create the `RescoreTopNQuery`.

vigyasharma (commented):
Thanks for persisting on this @dungba88 , changes look good. I have a few suggestions but this looks almost ready!

lucene/CHANGES.txt
rmuir: this problem is fixed for all files in `main` now, and any new trailing whitespaces will fail the CI build.

lucene/core/src/java/org/apache/lucene/search/RescoreTopNQuery.java
dungba88: I'll rename it to `createDocAndScoreQuery`. We can move it out of `AbstractKnnVectorQuery` too.

lucene/core/src/test/org/apache/lucene/search/TestRescoreTopNQuery.java
dungba88: Good idea, I'll use that to test. The current test demonstrates the idea of using a different field for rescoring.

lucene/core/src/java/org/apache/lucene/search/RescoreTopNQuery.java
dungba88: I have no preference, but I can't find where Scorer or Weight would expose the relation of the original query.

dungba88:
Sorry for spamming the replies! I should have gone to the Files changed tab, which allow sending all replies in the same message.

jmazanec15:
> For hybrid search in OpenSearch/Elastic Search, I'm wondering if @jmazanec15 and @benwtrent have any input. I'm having a feeling that it's quite common to combine lexical + KNN matching into a single BooleanQuery.

Not sure Im following the discussion completely. It is common to combine lexical and k-NN in a boolean query, but I think there is a lot of variety in what/how users are implementing hybrid search, so flexibility is great to have. Like @benwtrent mentioned, result computation for k-NN is done upfront, but queries can also be used to re-score after the initial phase via the [QueryRescorer](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/QueryRescorer.java) (as @mikemccand mentioned awhile ago), so I dont think a separate rescorer is necessary. 

I also like the point around lazy iteration: "or atleast provide a scorer that iterates the kNN query results calculating the higher fidelity scores". For expensive re-scoring (I think multi-vector will be), this might be nice to have too for hybrid/boolean queries - I think this approach is take in [FloatVectorSimilarityQuery](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/FloatVectorSimilarityQuery.java#L107). But, this can probably be taken for future consideration.

lucene/core/src/java/org/apache/lucene/search/RescoreTopNQuery.java
dungba88: I used the original result count as totalHits, but keep the relation as EQUAL_TO, as it doesn't seem it was exposed anywhere. Usually this would only be visible on `search(query, n)` with TopDocsCollector.

I also realized that this `totalHitsRelation` value would not be used anywhere so it would not matter.

vigyasharma (approved):

dungba88:
> result computation for k-NN is done upfront, but queries can also be used to re-score after the initial phase via the [QueryRescorer](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/QueryRescorer.java) (as @mikemccand mentioned awhile ago), so I dont think a separate rescorer is necessary.

Rescorer can be used, but IIUC Rescorer works only in the collection phase. After the first pass collection we will rescore the final results. This would not work if we combine semantic and lexical matching into a single Query, in that case we can only rescore the combined matches. Like @benwtrent mentioned, rescoring should be done in the rewrite phase. This will work for both the cases where semantic and lexical are combined or where semantic matching alone.

> "or atleast provide a scorer that iterates the kNN query results calculating the higher fidelity scores"

This is also handled by this PR, technically not a Scorer but a DoubleValueSources. User can use it to either use full-precision vectors, or even use another field for rescoring (amongst other use cases, as DoubleValueSources is extensible). A potential idea is to have 1-bit vector field for matching and another 4-bit or 7-bit vector field for rescoring. However the quantization cost of query vector even for scalar 7-bit is a bit high. We will tackle it as future optimization to this new Query.

Also thanks @vigyasharma for approving the PR! Can you help to merge it if there is no objection?

vigyasharma:
> Also thanks @vigyasharma for approving the PR! Can you help to merge it if there is no objection?

Yes, I'll merge in these changes tonight. Was waiting a day to allow people to give feedback on the latest revision if they want to.

vigyasharma:
@dungba88 – Can this change go in 10.3 instead of waiting for 11.0? I didn't see anything blocking so I updated the changes entry. But I'm running into some merge issues while backporting, likely because of the `DocAndScoreQuery` refactor? 

If you would like to raise a separate PR for 10.3 backport (against branch_10x), and I can help review it. Or if this can only go in 11.0, then I'll update back the changes entry. Let me know.

dungba88:
It should be possible to backport to 10.3. I'll raise a PR. Thanks @vigyasharma for merging!

dungba88:
I put a backport PR to 10.3 here: https://github.com/apache/lucene/pull/14860

PR #14010, opened 2024-11-22
Taxonomy counts are incorrect due to ordinal sorting (#14008)

### LINKED ISSUE #14008: Taxonomy counts are incorrect due to ordinal sorting
### Description

The following line implicitly sorts in `ordinal` array (for efficiency reasons I presume):
https://github.com/apache/lucene/blob/main/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacets.java#L326

But the following line is needing the original ordinal values:
https://github.com/apache/lucene/blob/main/lucene/facet/src/java/org/apache/lucene/facet/taxonomy/TaxonomyFacets.java#L333

This is giving me incorrect count values in the `labelValues` returned when I call `TaxonomyFacetIntAssociations#getTopChildren`

I am using JDK21, Lucene 10.0.0.

paulk-asert:
I just moved collecting the counts to before the sort occurs. Added a test which currently fails but passes after the change.

Closes #14008

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -123,6 +123,8 @@ Other
 Bug Fixes
 ---------------------
 
+* GITHUB#14008: Counts provided by taxonomy facets in addition to another aggregation are now returned together with
+  their corresponding ordinals. (Paul King)
 
 ======================= Lucene 10.0.0 =======================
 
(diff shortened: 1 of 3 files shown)


stefanvodita (approved):
Thank you for the fix and the test @paulk-asert!
I think we should push this to main, 10x, and 9x. I'll leave it up for a little while and if no one objects I'll add a CHANGES section for 9.12.1 and mention this bug fix here and in 10.0.1.

stefanvodita:
At the same time, I don't think we need to rush bug fix releases, since this functionality was broken from the time it was released.

PR #14011, opened 2024-11-22
Reduce allocation rate in HNSW concurrent merge

viliam-durina:
The PR removes the allocation of a new `LockedRow` for each locking operation in `HnswLock`. Even though the object was very quickly released, and JIT supports on-stack allocation, it didn't happen in my experiments on OpenJDK 21 - it's easy to avoid the allocation, rather than rely on the JIT.

Comparison of the change processing the same workload. The effect on overall execution time was negligible.

Before:
image

After:
image


viliam-durina:
@benwtrent https://github.com/apache/lucene/pull/14018

PR #14012, opened 2024-11-22
Simplify logic in ScoreCachingWrappingScorer

msfroh:
### Description

This is functionally equivalent to the logic that was present, but makes the behavior clearer.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/ScoreCachingWrappingScorer.java
+++ b/lucene/core/src/java/org/apache/lucene/search/ScoreCachingWrappingScorer.java
@@ -31,8 +31,7 @@ import java.util.Collections;
  */
 public final class ScoreCachingWrappingScorer extends Scorable {
 
-  private int lastDoc = -1;
-  private int curDoc = -1;
+  private boolean computeScore;
   private float curScore;
   private final Scorable in;
 
@@ -64,7 +63,7 @@ public final class ScoreCachingWrappingScorer extends Scorable {
     @Override
     public void collect(int doc) throws IOException {
       if (scorer != null) {
-        scorer.curDoc = doc;
+        scorer.computeScore = true;
       }
       super.collect(doc);
     }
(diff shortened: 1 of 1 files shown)


lucene/core/src/java/org/apache/lucene/search/ScoreCachingWrappingScorer.java
```diff
@@ -31,8 +31,7 @@
  */
 public final class ScoreCachingWrappingScorer extends Scorable {
 
-  private int lastDoc = -1;
-  private int curDoc = -1;
+  private boolean computeScore;
```
vigyasharma, line 34: Minor: how about renaming this to something that tells whether the score is cached already e.g. `scoreCached` ?And the flag gets unset whenever we collect a new doc (we would reverse the boolean meaning with this name).

vigyasharma (approved):
I like this change. It's the first time I read this class, and this PR did make it easier for me. Any reason this is still in draft?

lucene/core/src/java/org/apache/lucene/search/ScoreCachingWrappingScorer.java
jpountz: +1 I had a similar reaction when reading this PR. It's easier for me to reason about when the cached score should be invalidated than when the score should be computed.

jpountz (approved):
Nice simplification.

msfroh:
Thanks folks! I made the suggested change and took this PR out of draft. 

I think this one makes more sense than my previous one over at https://github.com/apache/lucene/pull/13987

vigyasharma (approved):
Looks good, @msfroh.. Can you also add a Changes entry?

jpountz (approved):

jpountz:
@msfroh Do you think you'll have time to add a CHANGES entry? Otherwise I'll do it for you and merge.

msfroh:
> Looks good, @msfroh.. Can you also add a Changes entry?

I wasn't planning to add a CHANGES entry. Nothing functionally changes, there's no optimization. It's just an extremely minor code cleanup.

jpountz:
Fair.

PR #14015, opened 2024-11-22
fix JavaDoc: use TopDocs instead of Hits

saschaszott:
### Description

Class `Hits` was removed from the Lucene API with Lucene version 3.0. Use class `TopDocs` instead.


vigyasharma (approved):
Good catch! Thanks for fixing.

PR #14018, opened 2024-11-25
Fix changelog for GITHUB#14011

viliam-durina:


msokolov:
Thanks for that, I will move to ``Optimizations` section and merge

viliam-durina:
> @viliam-durina would you also like to attempt the backport of this and the related commit to 10.1? Generally this is a matter of cherry-picking the change on to the branch, in this case it would be `branch_10_1`

The `branch_10_1` doesn't exist yet, not sure what to do.

msokolov:
uh, got it. Sorry, I was confused, I think we can backport to `branch_10x` and that will get copied over to 10.1 when that release happens.

viliam-durina:
Backport ready in #14037

PR #14019, opened 2024-11-25
Grammar and typo fixes

viliam-durina:
A random collection while reading through the code.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThreadPool.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThreadPool.java
@@ -30,13 +30,13 @@ import org.apache.lucene.util.ThreadInterruptedException;
 
 /**
  * {@link DocumentsWriterPerThreadPool} controls {@link DocumentsWriterPerThread} instances and
- * their thread assignments during indexing. Each {@link DocumentsWriterPerThread} is once a
- * obtained from the pool exclusively used for indexing a single document or list of documents by
- * the obtaining thread. Each indexing thread must obtain such a {@link DocumentsWriterPerThread} to
+ * their thread assignments during indexing. Each {@link DocumentsWriterPerThread} is, once obtained
+ * from the pool, exclusively used for indexing a single document or list of documents by the
+ * obtaining thread. Each indexing thread must obtain such a {@link DocumentsWriterPerThread} to
  * make progress. Depending on the {@link DocumentsWriterPerThreadPool} implementation {@link
  * DocumentsWriterPerThread} assignments might differ from document to document.
  *
- * <p>Once a {@link DocumentsWriterPerThread} is selected for flush the {@link
+ * <p>Once a {@link DocumentsWriterPerThread} is selected for flush, the {@link
  * DocumentsWriterPerThread} will be checked out of the thread pool and won't be reused for
  * indexing. See {@link #checkout(DocumentsWriterPerThread)}.
  */
@@ -184,7 +184,7 @@ final class DocumentsWriterPerThreadPool implements Iterable<DocumentsWriterPerT
   /**
    * Removes the given DWPT from the pool unless it's already been removed before.
    *
-   * @return <code>true</code> iff the given DWPT has been removed. Otherwise <code>false</code>
+   * @return <code>true</code>, iff the given DWPT has been removed. Otherwise <code>false</code>
    */
   synchronized boolean checkout(DocumentsWriterPerThread perThread) {
     // The DWPT must be held by the current thread. This guarantees that concurrent calls to
--- a/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThreadPool.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThreadPool.java
@@ -184,7 +184,7 @@ final class DocumentsWriterPerThreadPool implements Iterable<DocumentsWriterPerT
   /**
    * Removes the given DWPT from the pool unless it's already been removed before.
    *
-   * @return <code>true</code>, iff the given DWPT has been removed. Otherwise <code>false</code>
+   * @return <code>true</code> iff the given DWPT has been removed. Otherwise <code>false</code>
    */
   synchronized boolean checkout(DocumentsWriterPerThread perThread) {
     // The DWPT must be held by the current thread. This guarantees that concurrent calls to
--- a/lucene/core/src/java/org/apache/lucene/index/IndexCommit.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexCommit.java
@@ -25,9 +25,9 @@ import org.apache.lucene.store.Directory;
  * Expert: represents a single commit into an index as seen by the {@link IndexDeletionPolicy} or
  * {@link IndexReader}.
  *
- * <p>Changes to the content of an index are made visible only after the writer which made that
- * change commits by writing a new segments file (<code>segments_N</code>). This point in time, when
- * the action of writing of a new segments file to the directory is completed, is an index commit.
+ * <p>Changes to the content of an index are made visible only after the writer who made that change
+ * commits by writing a new segments file (<code>segments_N</code>). This point in time, when the
+ * action of writing of a new segments file to the directory is completed, is an index commit.
  *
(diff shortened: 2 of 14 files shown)


lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThreadPool.java
```diff
@@ -184,7 +184,7 @@ List<DocumentsWriterPerThread> filterAndLock(Predicate<DocumentsWriterPerThread>
   /**
    * Removes the given DWPT from the pool unless it's already been removed before.
    *
-   * @return <code>true</code> iff the given DWPT has been removed. Otherwise <code>false</code>
+   * @return <code>true</code>, iff the given DWPT has been removed. Otherwise <code>false</code>
```
shubhamvishu, line 187: Maybe fix /iff/if/ since you are at it?

viliam-durina: "Iff" is correct, it means "if and only if", see https://www.merriam-webster.com/dictionary/iff

msokolov: Since this is all about syntax, grammar, etc, we might as well italicize _iff_ since that is how it is generally done, as in the example you cited.

msokolov: also, the comma you inserted is not correct (try replacing iff with "if and only if" - we wouldn't use a comma before "if").

lucene/core/src/java/org/apache/lucene/index/IndexCommit.java
```diff
@@ -25,9 +25,9 @@
  * Expert: represents a single commit into an index as seen by the {@link IndexDeletionPolicy} or
  * {@link IndexReader}.
  *
- * <p>Changes to the content of an index are made visible only after the writer who made that change
- * commits by writing a new segments file (<code>segments_N</code>). This point in time, when the
- * action of writing of a new segments file to the directory is completed, is an index commit.
+ * <p>Changes to the content of an index are made visible only after the writer which made that
```
msokolov, line 28: sorry this is wrong. We could use `that`, but `who` is also correct - no need to "fix" it,

lucene/core/src/java/org/apache/lucene/index/IndexCommit.java
```diff
@@ -25,9 +25,9 @@
...
+ * change commits by writing a new segments file (<code>segments_N</code>). This point in time, when
```
msokolov, line 29: OK this sentence is awkward, but what you've done here doesn't really address the core problem. I think it really would be most correct as:  "_The_ point in time (no comma) when ... is completed (no comma) is an index commit".

lucene/core/src/java/org/apache/lucene/index/IndexOptions.java
```diff
@@ -33,7 +33,7 @@ public enum IndexOptions {
   DOCS,
   /**
    * Only documents and term frequencies are indexed: positions are omitted. This enables normal
-   * scoring, except Phrase and other positional queries will throw an exception.
+   * scoring, except phrase and other positional queries will throw an exception.
```
msokolov, line 36: I think I would prefer to avoid this ambiiguity by using  `PhraseQuery` explicitly

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -101,8 +101,8 @@
  * whether a new index is created, or whether an existing index is opened. Note that you can open an
  * index with {@link OpenMode#CREATE} even while readers are using the index. The old readers will
  * continue to search the "point in time" snapshot they had opened, and won't see the newly created
- * index until they re-open. If {@link OpenMode#CREATE_OR_APPEND} is used IndexWriter will create a
- * new index if there is not already an index at the provided path and otherwise open the existing
+ * index until they re-open. If {@link OpenMode#CREATE_OR_APPEND} is used, IndexWriter will create a
```
msokolov, line 104: Again, not a great sentence, but fixing it requires a bigger rewrite; I don't think that this change is better (the "will" has no subject).

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -123,13 +123,13 @@
  * IndexWriterConfig#setRAMBufferSizeMB}) or the number of added documents (see {@link
  * IndexWriterConfig#setMaxBufferedDocs(int)}). The default is to flush when RAM usage hits {@link
  * IndexWriterConfig#DEFAULT_RAM_BUFFER_SIZE_MB} MB. For best indexing speed you should flush by RAM
- * usage with a large RAM buffer. In contrast to the other flush options {@link
- * IndexWriterConfig#setRAMBufferSizeMB} and {@link IndexWriterConfig#setMaxBufferedDocs(int)},
+ * usage with a large RAM buffer. In contrast to the other flush options, in case of {@link
```
msokolov, line 126: let's leave this as it was. It was OK.

msokolov (commented):
I started reviewing these one by one, but as I go I find that many of these changes insert unneeded commas and other agrammatical constructions that are sometimes worse than the original (which admittedly could stand to be improved). If you really want to work on this, I am happy to give out more grammar pedantry (it tickles some kind of weird obsession I have), but maybe it is not worth our time?

viliam-durina:
Thanks for the review :+1: 

I reverted all the disputed changes, agree we should not go into unproductive discussions. I added a few new ones, they are in a separate commit to simplify the review.

I'll only have note to the `writer ... who` (https://github.com/apache/lucene/pull/14019#discussion_r1864958353) - "who" should be used only for persons, this writer is not a person. But I reverted it anyway.

msokolov (approved):

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -141,7 +141,7 @@
  * soon as a new commit is done. Creating your own policy can allow you to explicitly keep previous
  * "point in time" commits alive in the index for some time, either because this is useful for your
  * application, or to give readers enough time to refresh to the new commit without having the old
- * commit deleted out from under them. The latter is necessary when multiple computers take turns
+ * commit deleted under their hands. The latter is necessary when multiple computers take turns
```
msokolov, line 144: this sounds like an idiom of sort, but I've never heard it -- I think the existing phrasing is more idiomatic of the English I know anyway.

lucene/core/src/java/org/apache/lucene/index/IndexWriterConfig.java
```diff
@@ -509,28 +509,28 @@ public IndexWriterConfig setCheckPendingFlushUpdate(boolean checkPendingFlushOnU
...
    * <p>Soft deletes allow to retain documents across merges if the merge policy modifies the live
    * docs of a merge reader. {@link SoftDeletesRetentionMergePolicy} for instance allows to specify
-   * an arbitrary query to mark all documents that should survive the merge. This can be used to for
-   * example keep all document modifications for a certain time interval or the last N operations if
-   * some kind of sequence ID is available in the index.
+   * an arbitrary query to mark all documents that should survive the merge. This can be used, for
+   * example, to keep all document modifications for a certain time interval or the last N
+   * operations if some kind of sequence ID is available in the index.
    *
    * <p>Currently there is no API support to un-delete a soft-deleted document. In oder to un-delete
```
msokolov, line 527: Can we fix "In oder" to be "In order"?

msokolov (changes requested):
Thank you for all the fixes! I reviewed them all this time, and I think there are just a couple of changes I'd like to see before merging.

msokolov (approved):
Looks good, thank you!

msokolov:
BTW I am fascinated by your github avatar pic! I feel like I have seen it before but not sure where? I tried image search but it does not find it. Is it something well known?

viliam-durina:
> BTW I am fascinated by your github avatar pic! I feel like I have seen it before but not sure where? I tried image search but it does not find it. Is it something well known?

It's an illustration from a children book by Albín Brunovský :smile:

msokolov:
wow those books would scare any child that I know!

PR #14027, opened 2024-11-29
Make SegmentInfos#readCommit(Directory, String, int) public

javanna:
The corresponding readLatestCommit method is public and can be used to read segment infos from indices that are older than N - 1. The same should be possible for readCommit, but that requires the method that takes the minimum supported version as an argument to be public.


jpountz (approved):
Thanks for catching this, it indeed makes little sense to have it public on readLatestCommit and not on readCommit.

PR #14038, opened 2024-12-04
Reduce specialization in TopScoreDocCollector.

jpountz:
The specialization of `SimpleCollector` vs. `PagingCollector` only helps save a null check, so it's probably not worth the complexity. Benchmarks cannot see a difference with this change.


rmuir (approved):

javanna:
++ thanks for simplifying this!

PR #14031, opened 2024-12-02
Ensure Panama float vector distance impls inlinable 

ChrisHegarty:
This commit reduces the Panama vector distance float implementations to less than the maximum bytecode size of a hot method to be inlined (325).

E.g. Previously: ` org.apache.lucene.internal.vectorization.PanamaVectorUtilSupport::dotProductBody (355 bytes)   failed to inline: callee is too large`.

After: `org.apache.lucene.internal.vectorization.PanamaVectorUtilSupport::dotProductBody (311 bytes)   inline (hot)`

This helps things a little.

PR #14041, opened 2024-12-04
Revert "Ensure Panama float vector distance impls inlinable "

rmuir:
Reverts apache/lucene#14031

Reverting because of query results in nightly benchmark. Let's figure out separately what is happening. We may need to test on AMD cpu (i have some, and so does AWS).

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -130,9 +130,6 @@ Optimizations
 * GITHUB#14032: Speed up PostingsEnum when positions are requested.
   (Adrien Grand)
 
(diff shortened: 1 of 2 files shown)


ChrisHegarty:
Thanks @rmuir

ChrisHegarty (approved):

rmuir:
I will backport from 10.x too, and open followup issue to figure it out. 

Would definitely love for us to consider taking a step back and making nightly easier to debug here, maybe we can safely add `-XX:+PrintCompilation` or `-XX:+LogCompilation` when running it, to be able to see better what is happening? I don't know if it will disrupt performance.

and separately, test  changes on AMD.

PR #14046, opened 2024-12-06
Refactor dummy scorables.

jpountz:
This makes our scorers that produce dummy scorables share the same implementation.


benwtrent (approved):
Love seeing deleted code 😍

jpountz:
There was a regression on nightly benchmarks last night (https://benchmarks.mikemccandless.com/2024.12.09.18.05.31.html) that is either due to this PR or to newly added tasks (`CountOrMany`, `CountFilteredOrMany`). I'm reverting this PR to figure out which one is the culplit. I'll add it back if performance doesn't improve.

PR #14048, opened 2024-12-06
Reduce specialization in `ForUtil` and `ForDeltaUtil`.

jpountz:
These classes specialize all bits per value up to 24. But performance of high numbers of bits per value is not very important, because they are used by short postings lists, which are fast to iterate anyway. So this PR only specializes up to 16 bits per value.

For instance, if a postings list uses blocks of 17 bits per value, it means that one can find gaps of 65,536 consecutive doc IDs that do not contain the term. Such rare terms do not drive query performance.


jpountz:
luceneutil doesn't notice any difference.

rmuir (approved):

Issue #14047, opened 2024-12-06
Randomize KNN vector format parameters in tests


msokolov:
### Description

I discovered while doing some sweeping refactoring changes that we lack coverage of many vector field features, like scalar compression and concurrent merging. We even lack randomization for basic settings like graph connectivity (we *always* test with the default = 32).

The idea here is to add randomization to `RandomCodec` similar to what we do for other field types. I made an initial, but many tests do make assumptions that were broken.  I think initially we will need a way to "opt out" of such randomization. Is there a standard way to do that (for a specific format, or for the whole codec)?

msokolov:
also pushed 4f8035c01393735e347f42c0bdc5e7b932db7447 to address a build failure by disabling randomization in  more tests

PR #14049, opened 2024-12-06
Randomize KnnVector codec params in RandomCodec; addresses gh-14047

msokolov:
This adds the randomization to RandomCodec and addresses some test issues.

* In a few places I had to disable the randomization; basically wherever we are carefully testing scores. In these cases we won't have the same testing for quantized scores, but I think those would require more targeted testing since it's unclear what kinds of guarantees we can make about score values under quantization.
* In a few other palces I added some slop to the score comparisons; basically if we had a score comparison in a test that was primarily focused on other things (like testing Explain)
* I fixed a divide-by-zero edge case in the ScalarQuantizer (when there is only a single vector in a segment)
* I standardized bounds checking so we have consistent exception messages for dimension mismatches

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/quantization/ScalarQuantizer.java
+++ b/lucene/core/src/java/org/apache/lucene/util/quantization/ScalarQuantizer.java
@@ -97,12 +97,18 @@ public class ScalarQuantizer {
     }
     assert maxQuantile >= minQuantile;
     assert bits > 0 && bits <= 8;
-    this.minQuantile = minQuantile;
-    this.maxQuantile = maxQuantile;
     this.bits = bits;
     final float divisor = (float) ((1 << bits) - 1);
-    this.scale = divisor / (maxQuantile - minQuantile);
-    this.alpha = (maxQuantile - minQuantile) / divisor;
+    if (minQuantile == maxQuantile) {
+      // avoid divide-by-zero with an arbitrary but plausible choice (leads to alpha = scale = 1)
+      this.minQuantile = minQuantile - divisor;
+      this.maxQuantile = maxQuantile + divisor;
+    } else {
+      this.minQuantile = minQuantile;
+      this.maxQuantile = maxQuantile;
+    }
+    this.scale = divisor / (this.maxQuantile - this.minQuantile);
+    this.alpha = (this.maxQuantile - this.minQuantile) / divisor;
   }
 
   /**
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -24,7 +24,8 @@ Optimizations
 
 Bug Fixes
 ---------------------
-(No changes)
+* GITHUB#14049: Randomize KNN codec params in RandomCodec. Fixes scalar quantization div-by-zero
+  when all values are identical. (Mike Sokolov)
 
 Other
 ---------------------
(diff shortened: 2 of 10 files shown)


msokolov:
oops a test failed; my bad for only beasting the ones in core ... I'll look at this one:

    gradlew :lucene:join:test --tests "org.apache.lucene.search.join.TestBlockJoin.testSimpleKnn" -Ptests.jvms=2 -Ptests.jvmargs= -Ptests.seed=E773842CB2A422CA

msokolov:
Another test fails now (this is going to be fun, it's hard to ferret all these things out proactively):

```
TestBasicBackwardsCompatibility > testIndexOldIndex {Lucene-Version:10.0.0; Pattern: index.%1$s-cfs.zip} FAILED
    org.junit.ComparisonFailure: wrong first document expected:<4[4]> but was:<4[1]>
        at __randomizedtesting.SeedInfo.seed([22D75A8559728A8D:93B6C125DE4C6E18]:0)
```

Looking at this test, I can't tell where these assertions came from: I think they might be too ambitious? @s1monw did you invent this?  The entire class was created out of whole cloth, but I assume at least some of these tests were pre-existing?

s1monw:
@msokolov I didn't invent this test. Yet, we never tested the current version obeying with it's own BWC expecations. Please make sure nothing randomizes anything in this test. BWC tests are not made for this. you need to supress anything that potentially randomizes stuff here. 
> Another test fails now (this is going to be fun, it's hard to ferret all these things out proactively):

maybe revert then? That's what it takes I guess?

msokolov:
Thank Simon, that's good feedback about avoiding randomization for these tests. There's nothing to revert - this hasn't been committed, it's just github running tests that is failing. I do wonder though if we run the risk of a user with a customized Codec (just applying settings like graph bushiness and enabling quantization requires a custom Codec, so it would be typical) failing to be able to load an older index version for some reason, and we'd have no advance warning from the back-compat tests.  Maybe that shouldn't be a concern, I'm not sure. In any case I will disable the randomization for these tests.

benwtrent (approved):
I do think this should have a "CHANGES" entry for the quantization bug fix. That bug will happen with uniform vector values which is common in testing infrastructure.

msokolov:
makes sense, I'll add a CHANGES entry and commit

lucene/core/src/java/org/apache/lucene/util/quantization/ScalarQuantizer.java
```diff
@@ -97,12 +97,18 @@ public ScalarQuantizer(float minQuantile, float maxQuantile, byte bits) {
     }
     assert maxQuantile >= minQuantile;
     assert bits > 0 && bits <= 8;
-    this.minQuantile = minQuantile;
-    this.maxQuantile = maxQuantile;
     this.bits = bits;
     final float divisor = (float) ((1 << bits) - 1);
-    this.scale = divisor / (maxQuantile - minQuantile);
-    this.alpha = (maxQuantile - minQuantile) / divisor;
+    if (minQuantile == maxQuantile) {
```
mccullocht, line 102: @msokolov mind if I backport just this changes in this file into branch_10x tonight?

msokolov: No, although it might be too late to do it on the "tonight" of which you were writing!

PR #14053, opened 2024-12-10
IndexInput.isLoaded seems to return false for mmap index inputs on Windows

### LINKED ISSUE #14050: IndexInput.isLoaded seems to return false for mmap index inputs on Windows
### Description

There is a check added to BaseDirectoryTest case in #13998 which has been failing since that PR was merged (nearly all Windows builds failing on jenkins):
```
        var loaded = in.isLoaded();
        if (FilterDirectory.unwrap(dir) instanceof MMapDirectory
            // direct IO wraps MMap but does not support isLoaded
            && !(dir.getClass().getName().contains("DirectIO"))) {
          assertTrue(loaded.isPresent());
          assertTrue(loaded.get());
        } else {
          assertFalse(loaded.isPresent());
        }
```

The isLoaded optional has a value of false, even though it is a mmap directory. Seems like isLoaded in MemorySegment always returns false on Windows?

_No response_

dweiss:
Fixes #14050

### DIFF
--- a/lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
+++ b/lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
@@ -27,6 +27,7 @@ import java.util.Objects;
 import java.util.Optional;
 import org.apache.lucene.util.ArrayUtil;
 import org.apache.lucene.util.BitUtil;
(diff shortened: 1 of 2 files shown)


ChrisHegarty (approved):
LGTM

rmuir (approved):

PR #14054, opened 2024-12-10
Add github on-commit tests on MacOS and Windows

dweiss:
There are some low-level APIs used now that may be surprising and behave differently on different platforms. I suggest we enable MacOS and Windows builds to have a broader coverage and earlier feedback. 

Windows builds are significantly slower compared to unix-ish platforms. I don't think there is a way around it (unless somebody works at github and could arrange a higher-cpu runner ;). But since those builds are shifted to the background and don't interfere with normal dev workflow, I don't think it's such a big issue.

This said, it'd be nice to have the tests run great again...

### DIFF
--- a/.github/workflows/run-checks-all.yml
+++ b/.github/workflows/run-checks-all.yml
@@ -38,7 +38,7 @@ jobs:
       - uses: ./.github/actions/prepare-for-build
 
       - name: Run gradle check (without tests)
-        run: ./gradlew check -x test -Ptask.times=true --max-workers 2
+        run: ./gradlew check -x test "-Ptask.times=true" --max-workers 2
 
 
   # This runs all tests without any other validation checks.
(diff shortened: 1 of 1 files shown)


dweiss:
The failing check on Windows is due to #14053 - shows it's working. ;)

rmuir (approved):
we need this, looks good to me, after isLoaded fixed is merged.

msokolov:
I thought we did have macos tests already? Maybe that is not tests, but some other task?

dweiss:
I think distribution tests (run-checks-mod-distribution.tests.yml) run on MacOSX. Also, Uwe has his own jenkins box, which runs Windows and Mac tests (on a virtual machine). These github workers are running on real hardware though and I think it's worth keeping them. Even if they are somehow "virtualized" or sandboxed, it'll be an earlier warning sign for pull requests that something is wrong.

PR #14058, opened 2024-12-11
Adjust test for new random quantization formats

### LINKED ISSUE #14051: TestKnnFloatVectorQuery.testScoreNegativeDotProduct failing seed
### Description

FAILED:  org.apache.lucene.search.TestKnnFloatVectorQuery.testScoreNegativeDotProduct

Error Message:
java.lang.AssertionError: expected:<0.0> but was:<0.5>

Stack Trace:
java.lang.AssertionError: expected:<0.0> but was:<0.5>
        at __randomizedtesting.SeedInfo.seed([20B8DBD2266326B8:1CAAD9AAEB2AB64F]:0)
        at org.junit.Assert.assertEquals(Assert.java:577)
        at org.junit.Assert.assertEquals(Assert.java:701)
        at org.apache.lucene.search.TestKnnFloatVectorQuery.testScoreNegativeDotProduct(TestKnnFloatVectorQuery.java:134)

_No response_

benwtrent:
lucene now randomly adjusts quantization params, etc. and consequently this test sometimes doesn't return exactly a score of `0`. Adjusting to ensure we are never negative.

Related: https://github.com/apache/lucene/pull/14049
closes: https://github.com/apache/lucene/issues/14051

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/TestKnnFloatVectorQuery.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestKnnFloatVectorQuery.java
@@ -131,7 +131,7 @@ public class TestKnnFloatVectorQuery extends BaseKnnVectorQueryTestCase {
         DocIdSetIterator it = scorer.iterator();
         assertEquals(2, it.cost());
         assertEquals(0, it.nextDoc());
-        assertEquals(0, scorer.score(), EPSILON);
+        assertTrue(0 <= scorer.score());
         assertEquals(1, it.advance(1));
         assertEquals(1, scorer.score(), EPSILON);
       }


msokolov (commented):
do we still also want to assert that it is < EPSILON?

msokolov:
I meant previously the test asserted (implicitly) that the value was in [-EPSILON, EPSILON]. You changed it to assert that the value is > 0. I was proposing a more restrictive criterion: that it is in [0, EPSILON].

benwtrent:
@msokolov FYI the test failed because the score was `0.5`

msokolov:
oh I see; I think in other cases like this, I disabled the randomness, but easing the constraint seems OK too.

msokolov (approved):

PR #14062, opened 2024-12-13
Check uptTo before score in ReqExclBulkScorer#score.

vsop-479:

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/ReqExclBulkScorer.java
+++ b/lucene/core/src/java/org/apache/lucene/search/ReqExclBulkScorer.java
@@ -67,7 +67,7 @@ final class ReqExclBulkScorer extends BulkScorer {
       }
     }
 
-    if (upTo == max) {
+    if (upTo != DocIdSetIterator.NO_MORE_DOCS && upTo == max) {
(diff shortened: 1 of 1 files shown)


jpountz:
It's not clear to me what problem this is fixing as it is legal to call `bulkScorer.score(collector, acceptDocs, NO_MORE_DOCS, NO_MORE_DOCS)`.

vsop-479:
> it is legal to call bulkScorer.score(collector, acceptDocs, NO_MORE_DOCS, NO_MORE_DOCS).

Thanks for reviewing @jpountz . Do you think we can return `NO_MORE_DOCS` immediately when `min` is NO_MORE_DOCS in `BulkScorer#score`?

jpountz:
This would be correct, but I don't think we should do it. If there is a `DocIdSetIterator` that is slow at `advance(NO_MORE_DOCS)`, this is where we should add a special check.

vsop-479:
I added this check in `ReqExclBulkScorer#score` because i found `upTo` would be `NO_MORE_DOCS` after the loop in some case, e.g: `max` is `NO_MORE_DOCS`, and result in call `req.score(collector, acceptDocs, NO_MORE_DOCS, NO_MORE_DOCS)`.

Maybe it is useless, I will close it.

PR #14064, opened 2024-12-13
Allow doc ID set iterators and scorers to optimize bulk iteration / scoring.

jpountz:
While trying to understand while Tantivy is still much faster than Lucene at exhaustive evaluation of disjunctions, I noticed that there is significant potential for vectorization that Lucene doesn't take advantage of.

This PR keeps the same overall logic but refactors the code in a way that the compiler can more easily notice when operations are candidates for auto-vectorization (confirmed by the produced assembly).


jpountz:
I made this PR a draft while I'm still thinking of whether there is a better way of getting the same benefits.

PR #14067, opened 2024-12-15
BlockJoinBulkScorer could check for parent deletions (not children)

jimczi:
This change ensures that BlockJoinBulkScorer verifies deletions at the parent level.

### DIFF
--- a/lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
+++ b/lucene/join/src/java/org/apache/lucene/search/join/ToParentBlockJoinQuery.java
@@ -533,7 +533,8 @@ public class ToParentBlockJoinQuery extends Query {
       }
 
       BatchAwareLeafCollector wrappedCollector = wrapCollector(collector, acceptDocs);
-      // We don't propagate the acceptDocs since only parents are checked for deletion in the wrapped collector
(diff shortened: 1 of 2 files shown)


jpountz:
Block-joins require deletes to be consistent across parent and child documents. So I wouldn't call it a bug, the implementation should be free to check deletes on child documents or parent documents depending on whichever is more convenient/efficient.

ChrisHegarty:
> Block-joins require deletes to be consistent across parent and child documents.

This seems a little fragile, and place where bugs could hide. I wonder if we should assert this constraint, which would at least catch some of these potential issue usage issues, while not hurt run time performance (too much).

jpountz:
Indeed. We have a few such assertions already, e.g. to check that docs returned by the child query don't match the parent bitset. Adding a few more would be good if it can be done in a reasonable fashion (e.g. not forcing nextDoc() / advance() to be called just for the sake of validation).

The block-join module also has `CheckJoinIndex`, which is expected to be used in a more offline fashion and will catch such bugs. https://github.com/apache/lucene/blob/a8d8d6b3d9efd483d29cb1aa8ff85672ad2be049/lucene/join/src/java/org/apache/lucene/search/join/CheckJoinIndex.java#L63-L84

PR #14072, opened 2024-12-17
aws jmh benchmark cleanups

rmuir:
This script lets you benchmark a patch across 10 different cpu types in parallel, just copy the branch name from PR and pass to `make`, along with any jmh args if you dont want to run all benchmarks, for example:
 
```
make PATCH_BRANCH=ChrisHegarty:unroll_by_2 JMH_ARGS='floatDotProductVector -p size=1024'
```
It creates a combined `build/report.txt` by CPU, suitable for copy/pasting into a PR.

Apply some cleanups:
* Remove hardcoded lucene 10 jar, which was breaking the benchie: use a glob pattern instead.
* Add graviton4 instance type (c8g) to the benchmark.

### DIFF
--- a/dev-tools/aws-jmh/README.md
+++ b/dev-tools/aws-jmh/README.md
@@ -2,15 +2,19 @@ Runs lucene microbenchmarks across a variety of CPUs in EC2.
 
 Example:
 
+```console
 export AWS_ACCESS_KEY_ID=xxxxx
 export AWS_SECRET_ACCESS_KEY=yyyy
 make PATCH_BRANCH=rmuir:some-speedup
+```
 
 Results file will be in build/report.txt
 
 You can also pass additional JMH args if you want:
 
+```console
 make PATCH_BRANCH=rmuir:some-speedup JMH_ARGS='float -p size=756'
+```
 
 Prerequisites:
 
--- a/dev-tools/aws-jmh/README.md
+++ b/dev-tools/aws-jmh/README.md
@@ -1,3 +1,20 @@
+<!--
+    Licensed to the Apache Software Foundation (ASF) under one or more
+    contributor license agreements.  See the NOTICE file distributed with
+    this work for additional information regarding copyright ownership.
+    The ASF licenses this file to You under the Apache License, Version 2.0
+    the "License"); you may not use this file except in compliance with
+    the License.  You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+ -->
+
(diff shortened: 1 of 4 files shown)


ChrisHegarty (approved):
This is incredibly useful. I didn't realise that we had such a thing!

Trivially, I had issues with ssh setup, and just changed the URLs to, https://github.com/...

rmuir:
OK, we can change that to https. I'm pretty sure it doesn't work for you because no agent was forwarded. i have in my ssh config:
```
AddKeysToAgent yes
ForwardAgent yes
```

This is just old habits because https clones are unreliable.

rmuir:
Yeah, i commented out the forwarding and I think i reproduced your issue:

```
fatal: [graviton4]: FAILED! => 
    changed: false
    cmd: /usr/bin/git ls-remote git@github.com:apache/lucene.git -h refs/heads/main
    msg: |-
        Warning:********@github.com: Permission denied (publickey).
        fatal: Could not read from remote repository.

        Please make sure you have the correct access rights
        and the repository exists.
    rc: 128
    stderr: |-
        Warning: Permanently added 'github.com' (ED25519) to the list of known hosts.
        git@github.com: Permission denied (publickey).
        fatal: Could not read from remote repository.

        Please make sure you have the correct access rights
        and the repository exists.
    stderr_lines: <omitted>
    stdout: ''
    stdout_lines: <omitted>
```

Will push a commit with change to https after i test that it works.

rmuir:
@ChrisHegarty take another look?

ChrisHegarty:
@rmuir this LGTM.  I clearly have some misconfiguration and/or quirkiness in my ssh setup, so no need to do too much to appease me.  I'm also not proficient with ansible!

I'm silly and it took me too long to determine that to debug a failure I need to use the `lucene` user to log into the VM, e.g. `ssh lucene@..` (rather than ec2-user)

I also needed to set `ServerAliveInterval` to say `30` in my case, otherwise the script just appeared to "hang" in `Assemble Sources`.  ( I just created an  `ansible.cfg` with `[ssh_connection]  ssh_args = -o ServerAliveInterval=30`). Again, a small gotcha  on my side. But worth a note about ssh config?

ChrisHegarty (approved):

rmuir:
@ChrisHegarty I want to make it so anybody can use it :)

I understand the SSH reliability issue as I have experienced it before. The issue is that i have sub-10ms ping time to us-east-1 region, so I don't see your problem. I will fix the ssh defaults here, I'll do it as a followup PR.

Typically I set them to something like this, to not overwrite the defaults of reusing-connections and compressing, which are immensely helpful (see "default" in https://docs.ansible.com/ansible/2.7/reference_appendices/config.html#ansible-ssh-args), but just add the keepalives:

```
-C -o ControlMater=auto -o ControlPersist=60s -o ServerAliveInterval=30 -o ServerAliveCountMax=10
```

In conjunction with this, I also set `retries = 3` so that when network fails it will try to reconnect instead of just dying.

I will make followup issue, and just try to set these as env variables in the Makefile so that nobody needs to fight it for good network reliability.

rmuir:
merging this one for now: will report back with a PR to make the SSH more robust.

PR #14073, opened 2024-12-17
Add comment to the SIMD intersection logic.

jpountz:
I did not know it when I checked in the code, but this is almost exactly the v1 intersection algorithm from the "SIMD compression and the intersection of sorted integers" paper.


ChrisHegarty (approved):
Very cool!! Lgtm

PR #14074, opened 2024-12-17
This fixes immutability of clauseSets (broken by #13950)

uschindler:
This fixes immutability of `BooleanQuery` caused by PR #13950.

See comment: https://github.com/apache/lucene/pull/13950#pullrequestreview-2508662206

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
@@ -124,7 +124,8 @@ public class BooleanQuery extends Query implements Iterable<BooleanClause> {
 
   private final int minimumNumberShouldMatch;
   private final List<BooleanClause> clauses; // used for toString() and getClauses()
-  private final Map<Occur, Collection<Query>> clauseSets; // used for equals/hashcode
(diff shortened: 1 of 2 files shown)


jpountz (approved):
Thanks @uschindler !

uschindler:
P.S.: I added the `Collections.immutableCollection` only in the getter, because making `EnumMap clauseSets` does not work:
- The inner values of the map are custom classes. As we have no unmodifiable Multiset, we cant wrap it without breaking equals (If we map it as unmodifiableCollection, the resulting equals/hashCode inspects all elements and respects order.
- we can make the outer map immutable, but as this one is not exposed, I don't care. I just added another warning.

For the future: If people submit PRs about making private members which are collections public, always check if the immutability could be violated. This is a major problem in Java where collections are by default mutable!

mikemccand:
> For the future: If people submit PRs about making private members which are collections public, always check if the immutability could be violated. This is a major problem in Java where collections are by default mutable!

+1

Is this something a static checker could catch?

gsmiller:
Ah great catch. Thanks @uschindler!

shubhamvishu:
Thanks for fixing this @uschindler !

PR #14079, opened 2024-12-18
hunspell: tolerate REP rule count mismatches

rmuir:
Similar to support for tolerating PFX/SFX count mismatches, add the ability to tolerate REP count mismatches.

The issue arises in recent updates to LibreOffice mongolian dictionary and is currently failing all PRs that change the analyzers:

https://bugs.documentfoundation.org/show_bug.cgi?id=164366

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -44,7 +44,9 @@ New Features
 
 Improvements
 ---------------------
-(No changes)
+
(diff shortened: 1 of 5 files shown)


donnerpeter (approved):
LGTM overall, thank you! Update CHANGES.txt?

rmuir:
I added a CHANGES. Also the underlying issue is now fixed in the LibreOffice. I will try to chase it down in the bataak repo after brushing up on some mongolian :)

PR #14081, opened 2024-12-19
Fix urls describing why NIOFS is not recommended for Windows

YeonghyeonKO:
### Description
- Lucene opens segment files whether JRE_IS_64BIT is true or false as below:
```java
  public static FSDirectory open(Path path, LockFactory lockFactory) throws IOException {
    if (Constants.JRE_IS_64BIT) {
      return new MMapDirectory(path, lockFactory);
    } else {
      return new NIOFSDirectory(path, lockFactory);
    }
  }
```
- In FSDirectory.java and NIOFSDirectory.java, there are [url](https://bugs.java.com/bugdatabase/view_bug?bug_id=6265734) links describing why NIOFSDirectory is not appropriate way to open file only for Windows. Unfortunately these have no longer been valid url, so I fixed them in this PR.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -31,7 +31,7 @@ Bug Fixes
 
 Other
 ---------------------
-(No changes)
+* GITHUB#14081: Fix urls describing why NIOFS is not recommended for Windows (Marcel Yeonghyeon Ko)
 
 ======================= Lucene 10.2.0 =======================
 
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -31,7 +31,7 @@ Bug Fixes
 
 Other
 ---------------------
-* GITHUB#14081: Fix urls describing why NIOFS is not recommended for Windows (Marcel Yeonghyeon Ko)
+(No changes)
 
 ======================= Lucene 10.2.0 =======================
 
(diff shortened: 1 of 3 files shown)


lucene/CHANGES.txt
```diff
@@ -31,7 +31,7 @@ Bug Fixes
 
 Other
 ---------------------
-(No changes)
+* GITHUB#14081: Fix urls describing why NIOFS is not recommended for Windows (Marcel Yeonghyeon Ko)
```
danmuzi, line 34: Let's include this patch in Lucene 10.2.0 :)
(Merge into the main -> cherry-pick to the branch_10x)
Lucene 11 is a long way from being released.

danmuzi (changes requested):
Thanks for your patch! 👍

lucene/CHANGES.txt
YeonghyeonKO: @danmuzi Thanks for reviewing my PR. I fixed CHANGES.txt to include in Lucene 10.2

danmuzi (approved):
LGTM

YeonghyeonKO:
@danmuzi I appreciate your review. Let me know any remaining steps I need to take before merging this PR.

PR #14083, opened 2024-12-20
Optimize BitSetIterator#intoBitSet.

jpountz:
Calling `DocIdSetIterator#intoBitSet` on a `BitSetIterator` copies set bits one by one from one `BitSet` to another. We can do better.


jpountz:
This gave the expected speedup on nightly benchmarks: https://benchmarks.mikemccandless.com/CountFilteredIntNRQ.html

PR #14091, opened 2024-12-31
Cover all DataType

LuXugang:
A change made ten years ago—did you forget to change `DataType.PACKED` back to `DataType.values()`? 😊  @jpountz


jpountz (approved):
Wow, you're probably right! Thanks for fixing.

LuXugang:
Sorry for misssing backreport this issue, I would fix it soon

PR #14096, opened 2025-01-02
Improve use of generics in DisjunctionMaxQuery construction

ChrisHegarty:
This is a small usability improvement which allows construction of a DisjunctionMaxQuery to play nicer with type inference. Follows PECS.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/TestDisjunctionMaxQuery.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestDisjunctionMaxQuery.java
@@ -638,6 +638,7 @@ public class TestDisjunctionMaxQuery extends LuceneTestCase {
   }
 
   // Non-functional. Compile only - to ensure generics and type inference play nicely together
+  @SuppressWarnings("unused")
   public void testGenerics() {
     var dmq1 =
         new DisjunctionMaxQuery(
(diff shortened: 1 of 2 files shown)


lucene/core/src/test/org/apache/lucene/search/TestDisjunctionMaxQuery.java
```diff
@@ -637,13 +637,27 @@ private void doTestRandomTopDocs(int numFields, double... freqs) throws IOExcept
     dir.close();
   }
 
+  // Non-functional. Compile only - to ensure generics and type inference play nicely together
```
dweiss, line 640: I'd just use ```@Ignore("Compile only - to ensure generics and type inference play nicely together") and then perhaps add some sinkhole for the variables so that suppress warnings isn't needed; something like:

```
assertTrue(Arrays.asList(dmq1, disjuncts, dmq2).size() > 0)
```

Or whatever so that the variables are consumed somehow?

dweiss (approved):

lucene/core/src/test/org/apache/lucene/search/TestDisjunctionMaxQuery.java
ChrisHegarty: I was simpler to just make the test run, as normal, with a trivial asset. It's sufficient to ensure that things work as expected.

dweiss (approved):

uschindler:
Very nice!

Issue #13565, opened 2024-07-11
Try applying bipartite graph reordering to KNN graph node ids


msokolov:
### Description

We recently added BPIndexReorderer to enable sorting indexed docids using the BP approach to cluster documents with similar postings in some chosen set of fields. The same approach can also be used to sort node ids in a graph, eg see https://dl.acm.org/doi/10.1145/2939672.2939862 and https://arxiv.org/pdf/2403.01797 applies similar ideas to shards. I think we could apply the algorithm we have already implemented to HNSW node ids using document/document dot products as the distant metric that drives the partitioning and we might get some compression benefits since nearby nodes will tend to be linked together and this should reduce the overall node id deltas. We might also get memory locality benefits; possibly the memory access patterns could get a bit less random.

jpountz:
> We might also get memory locality benefits

This is what got me to thinking of BP for HNSW search: intuitively, it could help a lot when the dataset size exceeds the size of the page cache?

msokolov:
> This is what got me to thinking of BP for HNSW search: intuitively, it could help a lot when the dataset size exceeds the size of the page cache?

Yes, and as we keep compressing the vectors themselves, the graph takes up more and more space, relatively speaking

tteofili:
this [paper](https://dl.acm.org/doi/abs/10.1145/3626772.3657906) from SIGIR'24 seems to do exactly this as a first step in their Block Max Pruning technique.

mikemccand:
> This is what got me to thinking of BP for HNSW search: intuitively, it could help a lot when the dataset size exceeds the size of the page cache?

I think that gains might be astounding?

Similar vectors would be stored near each other in the `.vec` / `.veq` file, so paging in larger blocks / OS readahead could be very effective (though we may have to turn off `MADV_RANDOM` and see if it helps).  It should also mean less broad exploration of the graph: once you find your "neighborhood" of similar-ish vectors you spend some effort there and more quickly get to the top K.

msokolov:
[...]

PR #14097, opened 2025-01-02
HNSW BP reordering 

msokolov:
### Description

This is similar to the previous PR (#13683) on the same issue, but the difference here is:

1. rebased on main (there was a 10.0 major release in between)
2. refactored to make `BpVectorReorderer` compatible with `BPIndexReorderer` so that either can plug into `BPReorderingMergePolicy`

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/hnsw/IncrementalHnswGraphMerger.java
+++ b/lucene/core/src/java/org/apache/lucene/util/hnsw/IncrementalHnswGraphMerger.java
@@ -68,7 +68,7 @@ public class IncrementalHnswGraphMerger implements HnswGraphMerger {
   @Override
   public IncrementalHnswGraphMerger addReader(
       KnnVectorsReader reader, MergeState.DocMap docMap, Bits liveDocs) throws IOException {
-    if (!noDeletes(liveDocs) || !(reader instanceof HnswGraphProvider)) {
+    if (hasDeletes(liveDocs) || !(reader instanceof HnswGraphProvider)) {
       return this;
     }
     HnswGraph graph = ((HnswGraphProvider) reader).getGraph(fieldInfo.name);
@@ -189,16 +189,16 @@ public class IncrementalHnswGraphMerger implements HnswGraphMerger {
     return oldToNewOrdinalMap;
   }
 
-  private static boolean noDeletes(Bits liveDocs) {
+  private static boolean hasDeletes(Bits liveDocs) {
     if (liveDocs == null) {
-      return true;
+      return false;
     }
 
     for (int i = 0; i < liveDocs.length(); i++) {
       if (!liveDocs.get(i)) {
-        return false;
+        return true;
       }
     }
-    return true;
+    return false;
   }
 }
--- a/lucene/core/src/java/org/apache/lucene/util/hnsw/IncrementalHnswGraphMerger.java
+++ b/lucene/core/src/java/org/apache/lucene/util/hnsw/IncrementalHnswGraphMerger.java
@@ -21,7 +21,6 @@ import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
 import java.io.IOException;
 import org.apache.lucene.codecs.KnnVectorsReader;
 import org.apache.lucene.codecs.hnsw.HnswGraphProvider;
-import org.apache.lucene.codecs.perfield.PerFieldKnnVectorsFormat;
 import org.apache.lucene.index.ByteVectorValues;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.FloatVectorValues;
@@ -63,33 +62,30 @@ public class IncrementalHnswGraphMerger implements HnswGraphMerger {
 
   /**
    * Adds a reader to the graph merger if it meets the following criteria: 1. Does not contain any
-   * deleted docs 2. Is a HnswGraphProvider/PerFieldKnnVectorReader 3. Has the most docs of any
-   * previous reader that met the above criteria
+   * deleted docs 2. Is a HnswGraphProvider 3. Has the most docs of any previous reader that met the
+   * above criteria
    */
   @Override
   public IncrementalHnswGraphMerger addReader(
       KnnVectorsReader reader, MergeState.DocMap docMap, Bits liveDocs) throws IOException {
-    KnnVectorsReader currKnnVectorsReader = reader;
-    if (reader instanceof PerFieldKnnVectorsFormat.FieldsReader candidateReader) {
-      currKnnVectorsReader = candidateReader.getFieldReader(fieldInfo.name);
-    }
-
-    if (!(currKnnVectorsReader instanceof HnswGraphProvider) || !noDeletes(liveDocs)) {
+    if (!noDeletes(liveDocs) || !(reader instanceof HnswGraphProvider)) {
       return this;
     }
-
+    HnswGraph graph = ((HnswGraphProvider) reader).getGraph(fieldInfo.name);
+    if (graph == null || graph.size() == 0) {
+      return this;
+    }
     int candidateVectorCount = 0;
     switch (fieldInfo.getVectorEncoding()) {
       case BYTE -> {
-        ByteVectorValues byteVectorValues =
-            currKnnVectorsReader.getByteVectorValues(fieldInfo.name);
+        ByteVectorValues byteVectorValues = reader.getByteVectorValues(fieldInfo.name);
         if (byteVectorValues == null) {
           return this;
         }
         candidateVectorCount = byteVectorValues.size();
       }
       case FLOAT32 -> {
-        FloatVectorValues vectorValues = currKnnVectorsReader.getFloatVectorValues(fieldInfo.name);
+        FloatVectorValues vectorValues = reader.getFloatVectorValues(fieldInfo.name);
         if (vectorValues == null) {
           return this;
         }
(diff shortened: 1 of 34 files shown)


msokolov:
note: addresses issue #13565

msokolov:
One interesting detail that came out while testing with the merge policy is that indexing actually becomes *faster* with this change since the additional work we do in the BP reordering is outweighed by the savings we have performing HNSW searches during merging. Indeed the savings can be dramatic:

|condition| recall|  latency (ms)|     nDoc|  topK|  fanout|  maxConn|  beamWidth|  quantized|  index s|  index docs/s|  force merge s|  num segments|  index size (MB)|  vec disk (MB)|  vec RAM (MB)|
|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
|no BP| 0.794        | 0.996 | 2000000   | 10     | 64      | 64       | 250    | 7 bits | 1073.04      | 1863.87        | 271.95            | 1         | 2607.89      | 2441.406      | 488.281
|with BP|0.797        | 0.887 | 2000000   | 10     | 64      | 64       | 250    | 7 bits  | 987.76      | 2024.78        | 143.48            | 1         | 2558.39      | 2441.406      | 488.281

mikemccand:
These results look amazing @msokolov!

mikemccand:
@msokolov do you have changes to luceneutil's `knnPerfTest.py` to enable this?  Let's merge those upstream (to luceneutil) too ... I'm working on getting nightly benchy to run `knnPerfTest` so maybe we can cutover to this new merge policy there!

The nightly benchy already indexes KNN vectors (with and without quantization), so that's another place where we could turn on this new feature... I'll open a luceneutil issue to iterate!  Exciting...

benwtrent:
These are exciting numbers! Its interesting how improved search latency is dropping the index build time.

Do we know why the searching times are so much better? Is it simply because during merge, the merger sees the docs in a specific order that aids better graph building?

I am a bit confused by how this all works.

msokolov:
I am not sure, but surmising that search performance is improved because of some combination of (1) graph ordinal decoding being faster (since we encode using VInts and these are now smaller), and (2) improved memory locality something something - I mean we are in theory comparing vectors that are nearer to each other in virtual memory space - not sure how that could/would affect anything though. I admit this is pretty hand-wavy, but it's the best I've got -- the only change here is that more similar vectors are now nearer to each other in docid (and thus vector ordinal) space, and a follow-on from that is that delta encoding in the HNSW graph is more compact/efficient.

msokolov:
Regarding merging luceneutil tooling - I will open a PR, but suggest we hold off merging until this change hits Lucene

msokolov:
I will note that tests with smaller indexes don't show such dramatic improvements - more support for the theory that graph decoding is what is helped, because there are no real compression savings on small graphs where all the deltas are 2-bytes already. In the previous round of this Adrien had suggested looking into different encodings for the graph node ids, as we have for postings - I suspect there is room for improvement there

mikemccand:
I like the more efficient delta encoding theory.

Decoding `vInt` is a hotspot for HNSW graph traversal ... so if we can use 2 bytes instead of 3, or 1 byte instead of 2, thanks to the improved locality, that might be needle moving!

benwtrent:
If we really think `vint` is the cause, I wonder if we should switch encoding to the `readGroupVInts` stuff? https://github.com/apache/lucene/issues/12871


My thought around the bp reordering helping was dependent on the order of the vectors read is actually helping the vector indexing because clusters of vectors are indexed together. Honestly, bootstrapping HNSW via clusters is a really nice idea (and could help solve the "merge cost problem" by reusing clusters during merge...).

The numbers here are really nice. I just want to understand why they were better, especially as recall changes, which seems to indicate that the graph building itself is being changed.

msokolov:
> The numbers here are really nice. I just want to understand why they were better, especially as recall changes, which seems to indicate that the graph building itself is being changed.

benwtrent:
> I didnt' see a big impact on recall beyond what is typical from noise -- even with the same graph settings we see variance in recall due to randomness in the graph creation when there are multiple threads. I had 'numMergeWorker': (12,),
'numMergeThread': (4,) so that could be the explanation for the variance?

Yeah, that could explain variance. For repeatable tests, I generally only merge with one thread, but this is because I am usually doing things that effect recall.

But that force merge time difference is 🤯 `143.48` vs `271.95` is HUGE. It seems strange that we would get an almost 2x improvement just from faster vint reading?

lucene/misc/src/java/org/apache/lucene/misc/index/BpVectorReorderer.java
```diff
@@ -0,0 +1,788 @@
...
+      for (int i = ids.offset; i < ids.offset + ids.length; i++) {
+        VectorUtil.add(centroid, vectors.vectorValue(ids.ints[i]));
+      }
+      switch (vectorSimilarity) {
+        case EUCLIDEAN, MAXIMUM_INNER_PRODUCT -> vectorScalarMul(1 / (float) ids.length, centroid);
+        case DOT_PRODUCT, COSINE ->
+            vectorScalarMul(
+                1 / (float) Math.sqrt(VectorUtil.dotProduct(centroid, centroid)), centroid);
+      }
+      ;
```
jpountz, line 284: looks like it could be removed?

lucene/misc/src/java/org/apache/lucene/misc/index/BpVectorReorderer.java
```diff
@@ -0,0 +1,788 @@
...
+        int midPoint,
+        float[] leftCentroid,
+        float[] rightCentroid,
+        float[] scratch,
+        float[] biases)
+        throws IOException {
+
+      // Computing biases is typically a bottleneck, because each iteration needs to iterate over
+      // all postings to recompute biases, and the total number of postings is usually one order of
+      // magnitude or more than the number of docs. So we try to parallelize it.
```
jpountz, line 300: This comment is probably not true for vectors.

lucene/misc/src/java/org/apache/lucene/misc/index/BpVectorReorderer.java
```diff
@@ -0,0 +1,788 @@
...
+        maxLeftBias = Math.max(maxLeftBias, biases[i]);
+      }
+      float minRightBias = Float.POSITIVE_INFINITY;
+      for (int i = midPoint, end = ids.offset + ids.length; i < end; ++i) {
+        minRightBias = Math.min(minRightBias, biases[i]);
+      }
+      float gain = maxLeftBias - minRightBias;
+      // This compares the gain of swapping the doc from the left side that is most attracted to the
+      // right and the doc from the right side that is most attracted to the left against the
+      // average vector length (/500) rather than zero.
```
jpountz, line 328: Where does this 500 come from, is it a heuristic?

jpountz (approved):
This looks pretty cool! I left a few questions, nothing major.

msokolov:
Thanks for the reviews. I agree the measurements are not well explained. I have other runs that show no or less impact on search times, unchanged index sizes, and sometimes negative impact (on indexing times) -- I believe we expect indexing times to increase due to the additional work done when merging. But I don't think any of that should block merging this since it doesn't impact any existing use case, and testing to get to the bottom of what's going on is laborious since it requires building large indexes, so it could be done over time

lucene/core/src/java/org/apache/lucene/util/hnsw/IncrementalHnswGraphMerger.java
```diff
@@ -63,41 +62,38 @@ public IncrementalHnswGraphMerger(
...
+   * deleted docs 2. Is a HnswGraphProvider 3. Has the most docs of any previous reader that met the
+   * above criteria
    */
   @Override
   public IncrementalHnswGraphMerger addReader(
       KnnVectorsReader reader, MergeState.DocMap docMap, Bits liveDocs) throws IOException {
-    KnnVectorsReader currKnnVectorsReader = reader;
-    if (reader instanceof PerFieldKnnVectorsFormat.FieldsReader candidateReader) {
-      currKnnVectorsReader = candidateReader.getFieldReader(fieldInfo.name);
+    if (!noDeletes(liveDocs) || !(reader instanceof HnswGraphProvider)) {
```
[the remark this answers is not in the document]
msokolov: hah, true. Not sure where this came from - will change to `hasDeletes`. Q: is it truly necessary to check every bit of liveDocs? If we have a non-null liveDocs doesn't that imply one of the bits will be zero?

lucene/misc/src/java/org/apache/lucene/misc/index/BpVectorReorderer.java
msokolov: yes, but something analogous is true (cost is like N(docs) * D(dimension)  -- I'll update the comment

lucene/misc/src/java/org/apache/lucene/misc/index/BpVectorReorderer.java
msokolov: yes, I'll update the comment

lucene/core/src/java/org/apache/lucene/util/hnsw/IncrementalHnswGraphMerger.java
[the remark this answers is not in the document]
jpountz: > Q: is it truly necessary to check every bit of liveDocs? If we have a non-null liveDocs doesn't that imply one of the bits will be zero?

I'd say yes. IMO `null` is only an optimization for the case when all docs are live.

lucene/misc/src/java/org/apache/lucene/misc/index/BpVectorReorderer.java
jpountz: I had overlooked the number of dimensions, good point.

jpountz:
> I don't think any of that should block merging

Agreed, especially for the misc module.

benwtrent:
@msokolov I agree, we shouldn't block merging as it doesn't adjust default behavior. 

I was just surprised at the numbers and curious to the cause.

Issue #13896, opened 2024-10-12
Create a community metrics dashboard


stefanvodita:
### Description

Beam has a rich community metrics [dashboard](http://35.193.202.176/d/code_velocity/code-velocity?orgId=1) ([discussion](https://lists.apache.org/thread/0mqo7jfz0ooc1bzvggvw6h0gdlmbd1kl)). Would it help to have something similar?

prudhvigodithi:
OpenSearch Project has one https://github.com/opensearch-project/opensearch-metrics and I'm one of the maintainer for this metrics project. Happy to drive this here and be part of this discussion :).

- [Project Insights Dashboard](https://metrics.opensearch.org/_dashboards/app/dashboards#/view/a987a4b0-d801-11ee-8a84-e3710560950c)

- [Project Ops Metrics Dashboard](https://metrics.opensearch.org/_dashboards/app/dashboards#/view/f1ad21c0-e323-11ee-9a74-07cd3b4ff414)

Thank you.

stefanvodita:
@prudhvigodithi - that's great! Your help is very welcome, I wouldn't really know where to start if I was doing this myself.

prudhvigodithi:
To begin with we need:

-  A server that can run and host the metrics portal (the frontend).
- The application that will daily use the Github API to scrape the issues, pulls, contributions etc and save it to a database or any distributed service.
      - Since Github API's are limited, the best possible way is to collect daily incremental data (to begin fetching the data from last successful collection).
      - If required to use any external datasource, should be leveraging its own API's to fetch the data.
      - An existing Github bot can be used as well, either to use the bot token or push on an Github event.
- A side car (or a process in the same server) that will use the raw saved data from database or any distributed service and create the desired metrics that will be displayed by the frontend.

Thank you

stefanvodita:
That sounds like a plan! I'd be thrilled if you were driving this @prudhvigodithi - thank you for offering your help.
It would also be great to get feedback from others on which metrics they'd find most useful.

prudhvigodithi:
Thanks @stefanvodita, I would like to have the maintainers of the repo to take the initial stab at this metrics dashboard (come up the server, app, choosing the database and deploying the resources), and later as mentioned I will be happy to contribute to the dashboard design and features.

PR #14101, opened 2025-01-04
Use github wf to add module labels for PR based on file changes

### LINKED ISSUE #13897: Create bot to label PRs based on the files they touch
### Description

It would be useful to filter by "module:facet", "module:spatial", etc. These labels already exist, but I haven't seen them used. We could probably write a bot to add them to all new PRs.

pseudo-nymous:
### Description
Adds a new bot to automate labelling of pull requests with module tags based on file changes.
This bot uses `.github/labeler.yml` file to maintain mapping for module labels to corresponding regex paths to track. 

This workflow is triggered using `pull_request_target` event which provides necessary write permission for adding labels to PRs. More details around this event: [StackOverflow](https://stackoverflow.com/questions/74957218/what-is-the-difference-between-pull-request-and-pull-request-target-event-in-git)

### Testing
Tested changes by creating pr on my own fork. https://github.com/pseudo-nymous/lucene/pull/2

Further testing with sync flag: https://github.com/pseudo-nymous/lucene/pull/3

Closes #13897

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,133 @@
+# This file defines module label mappings for the Lucene project.
+# Each module is associated with a set of file globs that, when matched,
+# will trigger the corresponding label to be applied to pull requests.
+#
+# This configuration is used by the workflow defined in .github/workflows/label-pull-request.yml.
+# If we are adding new labels or refactoring modules, we will need to modify this file globs here to ensure that the correct labels are applied.
+
+# For more information on how to define globs, visit: https://github.com/actions/labeler
+
+module:analysis:
+  - changed-files:
+      - any-glob-to-any-file: 'lucene/analysis/**'
+
+module:benchmark:
+  - changed-files:
+      - any-glob-to-any-file: 'lucene/benchmark/**'
+
+module:classification:
+  - changed-files:
+      - any-glob-to-any-file: 'lucene/classification/**'
+
+module:core/codecs:
+  - changed-files:
+      - any-glob-to-any-file: ['lucene/core/src/java/org/apache/lucene/codecs/**', 'lucene/core/src/test/org/apache/lucene/codecs/**']
+
+module:core/FSTs:
+  - changed-files:
+      - any-glob-to-any-file: ['lucene/core/src/java/org/apache/lucene/util/fst/**', 'lucene/core/src/test/org/apache/lucene/util/fst/**']
+
+module:core/hnsw:
+  - changed-files:
+      - any-glob-to-any-file: ['lucene/core/src/java/org/apache/lucene/util/hnsw/**', 'lucene/core/src/test/org/apache/lucene/util/hnsw/**']
+
+module:core/index:
+  - changed-files:
+      - any-glob-to-any-file: ['lucene/core/src/java/org/apache/lucene/index/**', 'lucene/core/src/test/org/apache/lucene/index/**']
+
+module:core/search:
+  - changed-files:
+      - any-glob-to-any-file: ['lucene/core/src/java/org/apache/lucene/search/**', 'lucene/core/src/test/org/apache/lucene/search/**']
+
+module:core/store:
+  - changed-files:
+      - any-glob-to-any-file: ['lucene/core/src/java/org/apache/lucene/store/**', 'lucene/core/src/test/org/apache/lucene/store/**']
+
+module:core/other:
+  - changed-files:
+      - any-glob-to-any-file: ['lucene/core/**']
+      - all-globs-to-all-files:
+            - '!lucene/core/src/java/org/apache/lucene/codecs/**'
+            - '!lucene/core/src/test/org/apache/lucene/codecs/**'
+            - '!lucene/core/src/java/org/apache/lucene/util/fst/**'
+            - '!lucene/core/src/test/org/apache/lucene/util/fst/**'
+            - '!lucene/core/src/java/org/apache/lucene/util/hnsw/**'
+            - '!lucene/core/src/test/org/apache/lucene/util/hnsw/**'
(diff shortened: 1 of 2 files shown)


stefanvodita:
@pseudo-nymous, I'm only seeing this now, sorry! At first glance, it matches what I had in mind - thank you for addressing that issue! I'll do an in-depth review soon, but I'd appreciate others' thoughts as well. One potential concern is our GitHub action budget - do you have a sense of the cost of this workflow?

.github/labeler.yml
```diff
@@ -0,0 +1,134 @@
...
+    - changed-files:
+        - any-glob-to-any-file: ['lucene/spatial-extras/**', 'lucene/spatial-test-fixtures/**']
+
+module:spatial3d:
+  - changed-files:
+      - any-glob-to-any-file: 'lucene/spatial3d/**'
+
+module:spatial-extras:
+  - changed-files:
+      - any-glob-to-any-file: 'lucene/spatial-extras/**'
```
stefanvodita, line 126: Curious - why give this its own label and the `module:spatial` label as well?

.github/labeler.yml
```diff
@@ -0,0 +1,134 @@
+# This file defines module label mappings for the Lucene project.
+# Each module is associated with a set of file globs that, when matched,
+# will trigger the corresponding label to be applied to pull requests.
+#
+# This configuration is used by the workflow defined in .github/workflows/label-pull-request.yml.
+# If we are adding new labels or refactoring modules, we will need to modify this file globs here to ensure that the correct labels are applied.
+
+# For more information on how to use the `actions/checkout` action, visit:
```
stefanvodita, line 8: Is this relevant? I don't see us using `checkout` here.

.github/workflows/label-pull-request.yml
```diff
@@ -0,0 +1,21 @@
...
+#
+# For more information on the `actions/labeler` GitHub Action, refer to https://github.com/actions/labeler
+
+name: "Pull Request Labeler"
+run-name: Labelling pull request with module tags based on changed files in the PR
+on:
+  - pull_request_target
+
+jobs:
+  labeler:
```
stefanvodita, line 15: Can we add `sync-labels` to make it so labels get removed when they are no longer valid after an update to the PR?

.github/labeler.yml
```diff
@@ -0,0 +1,134 @@
...
+      - any-glob-to-any-file: ['lucene/core/src/java/org/apache/lucene/search/**', 'lucene/core/src/test/org/apache/lucene/search/**']
+
+module:core/store:
+  - changed-files:
+      - any-glob-to-any-file: ['lucene/core/src/java/org/apache/lucene/store/**', 'lucene/core/src/test/org/apache/lucene/store/**']
+
+module:core/other:
+  - changed-files:
+      - any-glob-to-any-file: ['lucene/core/**']
+      - all-globs-to-all-files:
```
stefanvodita, line 50: Cool!

stefanvodita (approved):
This is great @pseudo-nymous! I left minor comments that I would like to see addressed, but this is very good already. I retract my concern about the action budget. This is running when a PR is created or updated, which is not that often.

.github/labeler.yml
pseudo-nymous: It was a miss. I'll remove `lucene/spatial-extras/**` from `module:spatial` label.

.github/labeler.yml
pseudo-nymous: It's not relevant anymore. I'll remove it.

.github/workflows/label-pull-request.yml
pseudo-nymous: Great suggestion! I'll update.

pseudo-nymous:
Did further testing with new sync flag, change is working as expected.
Ref: https://github.com/pseudo-nymous/lucene/pull/3

pseudo-nymous:
> This is great @pseudo-nymous! I left minor comments that I would like to see addressed, but this is very good already. I retract my concern about the action budget. This is running when a PR is created or updated, which is not that often.

Thanks. 

Yes, this bot is only triggered when when pr is created or updated.

I have addressed all the comments. Please take a look.

.github/labeler.yml
stefanvodita: Sorry, I should have made a recommendation. I think it would be better to list both paths under `module:spatial` and not have `module:spatial-extras` at all. `spatial-extras` is really the core spatial module ([overview](https://github.com/apache/lucene/blob/main/lucene/spatial-extras/src/java/overview.html)).

stefanvodita (approved):
Thank you for responding to the comments so quickly! I've one more small thing to add and if that's sorted I'll leave it a few days to see if others have feedback, then merge.

.github/labeler.yml
pseudo-nymous: I see. It makes sense to have it in `module:spatial` label itself.

pseudo-nymous: Updated.

mikemccand:
Wow, this looks really awesome!  I wish we could retroactively apply it to PRs missing their module labels ... but let's start with this change (so new PRs are labeled).

stefanvodita:
Thank you for this contribution @pseudo-nymous, I merged it just now and I'm really looking forward to see the first PR using it!

stefanvodita:
Brilliant! Let us know if you find anything. If you're interesting in this area of GitHub workflows / community tools, there are a few others that would be great to do:
1. #13898
2. #14190
3. #13896

pseudo-nymous:
I'll be happy to work on these. Thanks!

stefanvodita:
Linking the PR related to the question above: #14721.

PR #14105, opened 2025-01-06
Optimize DFS while marking connected components (#14022)

viswanathk:
Backporting changes for #14022 into branch_10x.


viswanathk:
My bad. Made the changes.

PR #14109, opened 2025-01-07
prefetch may select the wrong memory segment for multi-segment slices

ChrisHegarty:
This commit fixes a bug where by prefetch may select the wrong memory segment for multi-segment slices.

The issue was discovered when debugging a large test scenario, where the index input was backed by several memory segments. When sliced, a multi-segment index input uses an offset into the initial memory segment. This offset should be added to the prefetch offset to determine the absolute offset. The unfortunate part is that the bounds checks need to either be moved to the subclasses, or somehow coerced to look at segment byteSizes rather than length. I found the former to be more straightforward.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -60,7 +60,9 @@ Optimizations
 
 Bug Fixes
 ---------------------
-(No changes)
+
+* GITHUB#14109: prefetch may select the wrong memory segment for
+  multi-segment slices. (Chris Hegarty)
 
 Other
 ---------------------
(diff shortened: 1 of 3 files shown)


uschindler (approved):
Looks fine. Thus is consistent to all other methods in subclasses.
Sorry for this code complexity, it's way too easy to add new methods but forget to look at code of methods. You're not the first one that broke this. But with normal read methods it's causing test failures, but prefetch just works.

Code is consistent so keep it like it is. The trap is there but when dealing with the IO layer you've to be careful.

jpountz (approved):

PR #14117, opened 2025-01-08
fix for gh#14110: stablize BpVectorReordered heuristic

msokolov:
It turned out that the initial shuffled order resulted in a left / right split with centroids that were very close together, the heuristic would terminate without doing any swaps, causing the test to fail. The test assumes that at least one swap will happen in the first iteration level. I changed the heuristic to be based on the size of the difference between the two centroids rather than the sum of their sizes.


msokolov:
BTW this also fixed failures like this one:

```
FAILED:  org.apache.lucene.misc.index.TestBpVectorReorderer.testQuantizedIndex

Error Message:
java.lang.AssertionError: out of order at 18

Stack Trace:
java.lang.AssertionError: out of order at 18
        at __randomizedtesting.SeedInfo.seed([CC2D27D6F33B8476:7613F311A04F402E]:0)
```

Issue #14118, opened 2025-01-08
backwards compat tests fail with "too many neighbors: 11"


msokolov:
### Description

repros for me with

```
gradlew :lucene:backward-codecs:test --tests "org.apache.lucene.backward_index.TestIndexUpgradeBackwardsCompatibility.testUpgradeOldIndex {Lucene-Version:10.1.0; Pattern: index.%1$s-cfs.zip}" -Ptests.jvms=4 -Ptests.jvmargs= -Ptests.seed=9C147645265876CD -Ptests.multiplier=2 -Ptests.directory=MMapDirectory -Ptests.gui=true -Ptests.file.encoding=UTF-8 -Ptests.vectorsize=128 -Ptests.forceintegervectors=true
```

msokolov:
It seems this was a bug introduced with https://github.com/apache/lucene/pull/14097 where we made a subtle change to preserve HNSW graph max-connections parameter from existing graphs when reordering rather than using the codec's configured value. This also applies to merging when we merge on to an existing graph as we attempt to do to save work. Ordinarily these values will always be the same, but when a graph that was written with a different Codec is read, the values can vary.

PR #14121, opened 2025-01-08
Preserve max-conn when merging onto existing graph Fixes gh#14118

msokolov:
preserve the HNSW graph max-conn when writing


msokolov:
fixes #14118

Issue #14106, opened 2025-01-07
Optimize DirectIOIndexInput


ChrisHegarty:
This issue has been filed to track various low-level optimisations to the DirectIOIndexInput.

A few specific points:
- [x] add overloads for primitive access, `getShort`, `getInt`, etc. #14107
- [x] add overloads for bulk retrieval, e.g. `readInts(int[] dst, int offset, int len)` #14124
- [x] avoid double buffering with `BufferedIndexInput` #14103

PR #14124, opened 2025-01-09
DirectIOIndexInput - add overloads for bulk retrieval

ChrisHegarty:
This commit adds overloads for bulk retrieval to DirectIOIndexInput. The implementation of these methods is identical to that of BufferedIndexInput, and it already covered by existing tests.

relates: #14106


original-brownbear (approved):
LGTM test failure looks unrelated :)

PR #14129, opened 2025-01-09
Upgrade commons-codec from 1.13.0 to 1.17.2

msfroh:
### Description

Similar to https://github.com/apache/lucene/pull/13269, we should use a less ancient version of commons-codec. 

Whereas commons-codec was broken with JPMS for versions 1.16.1, 1.17.0, and 1.17.1, it's now been fixed in 1.17.2 (see https://issues.apache.org/jira/projects/CODEC/issues/CODEC-324).


dweiss (approved):

msokolov (approved):

msokolov:
Thanks, @msfroh

PR #14131, opened 2025-01-10
Integrating GPU based Vector Search using cuVS

chatman:
Implementation of support for vector search on GPU with cuVS described in https://github.com/apache/lucene/issues/14243

This is an in-progress PR at the moment. Here is a way to test it out:
* Clone the cuvs repository from the PR branch.
* ./build.sh libcuvs && ./build.sh java
* (The above will install the cuvs-java artifacts in local Maven repository)
* Compile and use this branch in an IDE.

Contributors: @narangvivek10, @punAhuja, @chatman, @ChrisHegarty, along with help from @cjnolet.


build-tools/build-infra/build.gradle
```diff
@@ -22,6 +22,7 @@ plugins {
 }
 
 repositories {
+  mavenLocal()
```
dweiss, line 25: Remove mavenLocal before merging, if it happens. There will be issues with it - some are very cryptic and hard to diagnose (like different artifact hashes). It's going to be a major headache if it's left in.

navneet1v:
@chatman thanks for creating the PR. This looks very interesting. is the idea here is the Lucene library will on a GPU machine and running the CUVS.

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/CuVSVectorsWriter.java
```diff
@@ -0,0 +1,402 @@
...
+  }
+
+  @SuppressForbidden(reason = "A temporary java.util.File is needed for Cagra's serialization")
+  private byte[] createCagraIndex(float[][] vectors, List<Integer> mapping) throws Throwable {
+    CagraIndexParams indexParams =
+        new CagraIndexParams.Builder(resources)
+            .withNumWriterThreads(cuvsWriterThreads)
+            .withIntermediateGraphDegree(intGraphDegree)
+            .withGraphDegree(graphDegree)
+            .withCagraGraphBuildAlgo(CagraGraphBuildAlgo.NN_DESCENT)
```
navneet1v, line 118: I have some experience with building the Cagra index, and I think NN_DESCENT is faster in cagra index creation but it has a high GPU memory footprint. Should we use IVF_PQ here? Or can we have a hybrid approach where if doc count is < a specific number then we use NN_DESCENT else IVF_PQ.

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/CuVSVectorsWriter.java
```diff
@@ -0,0 +1,402 @@
...
+            .build();
+
+    // log.info("Indexing started: " + System.currentTimeMillis());
+    cagraIndexForHnsw =
+        new CagraIndex.Builder(resources).withDataset(vectors).withIndexParams(indexParams).build();
```
benwtrent, line 167: The cagra index builder should instead accept a file handle or something. Having to feed it on-heap vectors is not good.

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/CuVSVectorsReader.java
```diff
@@ -0,0 +1,340 @@
...
+    // log.info("initialFilePointerLoc: " + initialFilePointerLoc);
+
+    List<StackFrame> stackTrace = StackWalker.getInstance().walk(this::getStackTrace);
+
+    boolean isMergeCase = false;
+    for (StackFrame s : stackTrace) {
+      if (s.toString().startsWith("org.apache.lucene.index.IndexWriter.merge")) {
+        isMergeCase = true;
+        // log.info("Reader opening on merge call");
+        break;
```
benwtrent, line 101: You should instead use `getMergeInstance` which allows you do set merge settings, or whatever for a given reader.

benwtrent (changes requested):
This seems pretty far from ready yet. I left some comments on some glaring issues. However, there are other things like:

 - tests for queries
 - tests for the format
 - preventing bad behavior (e.g. using `byte[]`)

I haven't touched on the validity of having an NVIDIA only GPU backed index in Lucene sandbox directly. The new dependencies are huge. IDK if whomever downloads and builds lucene then now have to download and build with these? I am unsure how the sandbox stuff works.

chatman:
> This seems pretty far from ready yet. I left some comments on some glaring issues. However, there are other things like:
> 
>     * tests for queries
> 
>     * tests for the format
> 
>     * preventing bad behavior (e.g. using `byte[]`)
> 
> 
> I haven't touched on the validity of having an NVIDIA only GPU backed index in Lucene sandbox directly. The new dependencies are huge. IDK if whomever downloads and builds lucene then now have to download and build with these? I am unsure how the sandbox stuff works.

Indeed, Ben. This is WIP at the moment. More tests are WIP. As for loading the entire index in byte[], this is something that we're working with the NVIDIA/cuVS team to see if streaming can be supported (right now it is not).

build-tools/build-infra/build.gradle
chatman: Sure, Dawid, will add a nocommit comment there. I had added nocommits earlier, but had to remove them to make "./gradlew check" work. Is there a way to have it check everything other than the nocommits?

chatman:
> @chatman thanks for creating the PR. This looks very interesting. is the idea here is the Lucene library will on a GPU machine and running the CUVS.

Yes, exactly.

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/CuVSVectorsWriter.java
chatman: Unfortunately, we encountered some crashes while trying IVF_PQ, hence using NN_DESCENT as the default. Maybe we can make it configurable once those crashes can be thoroughly verified.

chatman:
> I haven't touched on the validity of having an NVIDIA only GPU backed index in Lucene sandbox directly. The new dependencies are huge. IDK if whomever downloads and builds lucene then now have to download and build with these? I am unsure how the sandbox stuff works.

This is something we need to work out as we want this to be. Here are my thoughts at the moment, and things we need consensus on:
* Right now, the `cuvs-java` dependency comes from local Maven and it should come from Maven central once the artifacts are available. 
* TODO: If a system doesn't have cuda or GPUs, these codepaths should gracefully fail and indicate that support not available.
* Continuous testing can be enabled on GPU enabled Jenkins instances (we can have a discussion around that).
* To validate the integration at API level, some mock tests (that simulate the same functionality using the CPU) can be added to the cuvs-java API.
* If we can choose, we can ignore shipping this by default.

build-tools/build-infra/build.gradle
dweiss: I think you can skip the entire task that looks for nocommits (and other things) by running ./gradlew check -x validateSourcePatterns

lucene/sandbox/build.gradle
```diff
@@ -19,9 +19,16 @@ apply plugin: 'java-library'
...
+  mavenLocal()
+}
+
+
 dependencies {
   moduleApi project(':lucene:core')
   moduleApi project(':lucene:queries')
   moduleApi project(':lucene:facet')
   moduleTestImplementation project(':lucene:test-framework')
+  moduleImplementation deps.commons.lang3
```
uschindler, line 32: It would really be great to remove the commons-lang3 (bullshit) library. Sorry for this, but this library is mostly useless. Sometimes some of its rarely used funtions can be replicated (if they are of good use for other parts). But in general the broken null handling in this library (it ignores nulls) and lots of legacy code that can be much easier written with more modern Java 21 APIs.

My suggestion: Actually the code here only uses `SerializationUtils.deserialize()` and `SerializationUtils.serialize()`. Maybe just copy those to the Utils class of the package as it seems to be of not much use elsewhere in Lucene - and remove this dependency.

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/CagraFieldVectorsWriter.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * CuVS based fields writer
+ */
+public class CagraFieldVectorsWriter extends KnnFieldVectorsWriter<float[]> {
+
+  public final String fieldName;
+  public final ConcurrentHashMap<Integer, float[]> vectors =
+      new ConcurrentHashMap<Integer, float[]>();
+  public int fieldVectorDimension = -1;
```
uschindler, line 33: this should be final and only initialized by ctor.

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/CuVSCodec.java
```diff
@@ -0,0 +1,58 @@
...
+    this("CuVSCodec", new Lucene101Codec());
+  }
+
+  public CuVSCodec(String name, Codec delegate) {
+    super(name, delegate);
+    KnnVectorsFormat format;
+    try {
+      format = new CuVSVectorsFormat(1, 128, 64, MergeStrategy.NON_TRIVIAL_MERGE);
+      setKnnFormat(format);
+    } catch (LibraryNotFoundException ex) {
```
uschindler, line 42: This should fail hard. If cuvs is not available it should throw exception and not just log a severe error. Instead this throws NPE later when the knn format is retrieved.

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/CuVSCodec.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  KnnVectorsFormat knnFormat = null;
+
+  @Override
+  public KnnVectorsFormat knnVectorsFormat() {
+    return knnFormat;
+  }
+
+  public void setKnnFormat(KnnVectorsFormat format) {
```
uschindler, line 55: No setters please, codecs should be unmodifiable! It should initialize on ctor or fail if library cannot be loaded.

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/CuVSSegmentFile.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+/**
+ * Methods to deal with a CuVS composite file inside a segment
+ */
+public class CuVSSegmentFile implements AutoCloseable {
+  private final ZipOutputStream zos;
+
+  private Set<String> filesAdded = new HashSet<String>();
```
uschindler, line 35: this must be final.

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/CuVSVectorsFormat.java
```diff
@@ -0,0 +1,91 @@
...
+  public static final String VECTOR_DATA_CODEC_NAME = "Lucene99CagraVectorsFormatData";
+  public static final String VECTOR_DATA_EXTENSION = "cag";
+  public static final String META_EXTENSION = "cagmf";
+  public static final int VERSION_CURRENT = 0;
+  public final int maxDimensions = 4096;
+  public final int cuvsWriterThreads;
+  public final int intGraphDegree;
+  public final int graphDegree;
+  public MergeStrategy mergeStrategy;
+  public static CuVSResources resources;
```
uschindler, line 41: There's something wrong: this is a static field, but it is initialized in ctor!? This looks wrong,

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/CuVSVectorsFormat.java
```diff
@@ -0,0 +1,91 @@
...
+
+  public CuVSVectorsFormat() {
+    super("CuVSVectorsFormat");
+    this.cuvsWriterThreads = 1;
+    this.intGraphDegree = 128;
+    this.graphDegree = 64;
+    try {
+      resources = new CuVSResources();
+    } catch (Throwable e) {
```
uschindler, line 50: don't catch Throwable, change to more exact types and rethrow them using our Lucene Mechanisms (keep RuntimeException and Error unmodified) and only rethrow Exceptions wrapped as something else (preferable IOException or similar).

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/CuVSVectorsFormat.java
```diff
@@ -0,0 +1,91 @@
...
+  public CuVSVectorsWriter fieldsWriter(SegmentWriteState state) throws IOException {
+    return new CuVSVectorsWriter(
+        state, cuvsWriterThreads, intGraphDegree, graphDegree, mergeStrategy, resources);
+  }
+
+  @Override
+  public CuVSVectorsReader fieldsReader(SegmentReadState state) throws IOException {
+    try {
+      return new CuVSVectorsReader(state, resources);
+    } catch (Throwable e) {
```
uschindler, line 82: same here.

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/CuVSVectorsFormat.java
```diff
@@ -0,0 +1,91 @@
...
+
+  public static final String VECTOR_DATA_CODEC_NAME = "Lucene99CagraVectorsFormatData";
+  public static final String VECTOR_DATA_EXTENSION = "cag";
+  public static final String META_EXTENSION = "cagmf";
+  public static final int VERSION_CURRENT = 0;
+  public final int maxDimensions = 4096;
+  public final int cuvsWriterThreads;
+  public final int intGraphDegree;
+  public final int graphDegree;
+  public MergeStrategy mergeStrategy;
```
uschindler, line 40: should also be final and why is this one and the previous ones public?

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/CuVSVectorsReader.java
```diff
@@ -0,0 +1,343 @@
...
+
+/**
+ * KnnVectorsReader instance associated with CuVS format
+ */
+public class CuVSVectorsReader extends KnnVectorsReader {
+
+  // protected Logger log = Logger.getLogger(getClass().getName());
+
+  IndexInput vectorDataReader = null;
+  public String fileName = null;
```
uschindler, line 62: all fields should not be public, and if they need to be public they should be final at least. I assume they can be package private or private.

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/CuVSVectorsWriter.java
```diff
@@ -0,0 +1,411 @@
...
+import org.apache.lucene.util.SuppressForbidden;
+
+/**
+ * KnnVectorsWriter for CuVS, responsible for merge and flush of vectors into GPU
+ */
+public class CuVSVectorsWriter extends KnnVectorsWriter {
+
+  // protected Logger log = Logger.getLogger(getClass().getName());
+
+  private List<CagraFieldVectorsWriter> fieldVectorWriters = new ArrayList<>();
```
uschindler, line 52: check which fields can be final.

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/PerLeafCuVSKnnCollector.java
```diff
@@ -0,0 +1,92 @@
...
+import org.apache.lucene.search.ScoreDoc;
+import org.apache.lucene.search.TopDocs;
+import org.apache.lucene.search.TotalHits;
+
+/**
+ * KnnCollector for CuVS
+ */
+public class PerLeafCuVSKnnCollector implements KnnCollector {
+
+  public List<ScoreDoc> scoreDocs;
```
uschindler, line 31: again why is all this public?

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/Util.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+import org.apache.commons.lang3.SerializationUtils;
+
+/**
+ * Some Utils used in CuVS integration
+ */
+public class Util {
```
uschindler, line 31: this class is internal, make it pkg private

uschindler (changes requested):
Next to Dawid's comments about the dependencies and the code, I have some major problems with:

- cleanup the API and do not make everything public. Only the Codec and format class needs to be public, all other pkg-private.
- remove public fields that are modifiable.
- make all fields final, if possible.
- there is an issue with a static field which gets initialized in the ctor. This looks wrong!
- don't swallow exceptions and log errors. If you have no CUVS, fail hard. It makes no sense to proceed, because without the native library and a graphics adapter the whole codec won't work.

Anyways, I am happy to see this and that the underlying library was moved to Panama. So it was really a pleasure to talk to your colleagues last summer at berlinbuzzwords!

lucene/sandbox/src/java/module-info.java
```diff
@@ -20,6 +20,9 @@
   requires org.apache.lucene.core;
   requires org.apache.lucene.queries;
   requires org.apache.lucene.facet;
+  requires java.logging;
+  requires com.nvidia.cuvs;
```
ChrisHegarty, line 24: This makes cuVS a non-optional dependency. I would have expected to see this feature as optional, i.e. if not present you get a nice error message or something. I guess it kinda depends on how the cuVS and native implementation are tied together? That is, can one use or even instantiate types from the cuVS Java API without the native library being present?

uschindler: I don't see this in the snapshot module a problem. The code itsself should be written in a way that the codec can only be used, if the native library is there. Currently the code is a bit broken as it does not fail hard. I'd like to change this (see my review),
But if the dependency is on Maven central, it should work.
Of course we could make it also optional in module system, but then you would get CNFE when loading codecs or related classes.

ChrisHegarty: @uschindler Agree - this feature should be optional.  If the Java API is present and loadable without the native library - has at least one callable method, which we can call and catch, then that is fine. I'll try it out.

uschindler: The current code catches this, but just logs a warning and then fails later (due to NPE). I argued that the Codec should fail hard or at least fails when it is to be used.

This must be refactored before release:
- Allow to load the Codec, but delay initialization of native library until the codec is actually used to create components.
- Fail hard when an index with the custom codec is loaded and theres no native access.

uschindler: When looking at the code I think best wozuld be the following: Create a small static "holder" class (inside the Codec impl) which has a static initializer loading and setting up the CUVS code. This should fail hard with some LinkageError if library is not there. The holder class should have getters for some CUVS entry points.

All getter methods to create formats and readers in the Codec delegate to the getters in the holder and pass through any exceptions. In addition, all CUVS classes should be package private (I mentioned this already) and only let the codec and a essential config classes public.

uschindler: The same for the postings formats and other public components loaded via SPI.

ChrisHegarty:
Hi,  `cuvs-java-25.02` is currently compiled with JDK 22, so has a minimum class file version of 66. Lucene compiles with a minimum of JDK 21, class file version 65.  The reason why cuvs-java has a minimum of JDK 22 is that is uses Panama/foreign APIs directly, which is fine.  

We have two options here: 
1. restructure cuvs-java so that it compiles to a minimum JDK 21, with an mr jar/version specific loading. Maybe it can even strip use class file version 65 and strip the preview bit, similar to what Lucene does. Depends on the nature of what the code is doing - which I've yet to fully look into. All said, this is quite involved, but may make that cuvs-java api more generally useful.   ( one can at least develop against it more easily, if not run tests, etc )

2. Restructure the code in Lucene so that it uses a toolchain and javac from JDK 22 and a `java22` sourceset. Also make it's loading conditional on the JDK version and/or use an mrjar.

dweiss:
The third option is to bump the minimum Java requirement to Java 22 on main? I know it's an interim release but maybe we should just do it, anticipating the next major lts (due to be released in September)?

uschindler:
> > restructure cuvs-java so that it compiles to a minimum JDK 21, with an mr jar/version specific loading. Maybe it can even strip use class file version 65 and strip the preview bit, similar to what Lucene does. Depends on the nature of what the code is doing - which I've yet to fully look into. All said, this is quite involved, but may make that cuvs-java api more generally useful. ( one can at least develop against it more easily, if not run tests, etc )
> 
> If this is possible, it would be very nice! I think @narangvivek10 has more details, but as I understand, the FFM functionality is being used in Project Panama, but those APIs are available only since JDK22.

Tha APIs are available in Java 21, too (with minimal changes regarding some specific parts like string handling). If you omit those, you can compile against java 21 and later strip the preview bit from all classes (Lucebe did this a while ago). Nowadays Lucene extracted a stub-jar with all public class signatures (without code) from java 21 and compiles against it with some compiler tricks.

I don't know if the APIs which are different are used by cuvs, so I can't give a recommendation.

If cuvs is only available with Java 22 due to incompatibility of APIs, we need to either upgrade minimum version of Lucene or we need toolkit magic to only compile this for Java 22 (which I'd like to avoid).

P.S. I'd like to bite into the apple and make Java 22 minimum requirement. Then we can use Memory segments in our public API and clean up a lot.

benwtrent:
@dweiss While I think that might work for unused/snapshot Lucene releases, I think @chatman et. al. is aiming for usage in the current Lucene Main so that Lucene focused search engines (e.g. OpenSearch, Elasticsearch, etc.) can take advantage of the cuvs format if they want.

@chatman the only way to get this to work in current major releases (e.g. this year) for Apache Lucene engines is to handle this Java version code and use the MR-Jar module stuff. Meaning, it cannot have a hard compilation dependency on java 22.

> I'd like to bite into the apple and make Java 22 minimum requirement. Then we can use Memory segments in our public API and clean up a lot.

You mean bump the minimum Java for every Java release until the next LTS, assuming that Lucene 11 will require Java 25? I would be for this myself :). Releasing Lucene 11 with Java 25 will have many nice benefits.

But, I think there is a strong desire for @chatman and company to get this into production without waiting another year+.

uschindler:
P.P.S. Elasticsearch has a Gradle plugin to strip preview flags.

ChrisHegarty:
> Tha APIs are available in Java 21, too (with minimal changes regarding some specific parts like string handling). If you omit those, you can compile against java 21 and later strip the preview bit from all classes (Lucebe did this a while ago). Nowadays Lucene extracted a stub-jar with all public class signatures (without code) from java 21 and compiles against it with some compiler tricks.
> 
> I don't know if the APIs which are different are used by cuvs, so I can't give a recommendation.

From a quick skim, then I think we can manage to make this work. For the small shimmer in the foreign API, we can provide static version specific stubs. We do similar in Elasticsearch.

> If cuvs is only available with Java 22 due to incompatibility of APIs, we need to either upgrade minimum version of Lucene or we need toolkit magic to only compile this for Java 22 (which I'd like to avoid).

At a minimum, the cuvs-java API exposes MemorySegment, so all consumers - Lucene in this case - would have to deal with the preview-ness nonsense!. But again, this may be doable with a little fiddling in gradle and the class file.

> P.S. I'd like to bite into the apple and make Java 22 minimum requirement. Then we can use Memory segments in our public API and clean up a lot.

As you know, I'm strongly in favour of moving to newer java versions. However, this is quite a move. I will result in every project consuming Lucene to keep upgrading constantly until the next Java LTS.  Which I'm personally ok with, but I'm not sure anyone has really done this before.

ChrisHegarty:
> P.P.S. Elasticsearch has a Gradle plugin to strip preview flags. Basically it patches one byte in all class files that are created by Javac.

Yes, we can do this. Along with an MRJAR plugin. In ES we use the toolchain to resolve newer JDKs, e.g. 22, etc

uschindler:
One additional important thing: no public API in this new Lucene module/codec must export any preview API, so it must all be private/pkg-private.

ChrisHegarty:
I've made the cuvs-java api Java 21 friendly, with an spi and a java-22 specific impl in the versioned section of an mrjar - MemorySegment and Arena have been removed from the api, so only appear in the 22 impl code.   https://github.com/rapidsai/cuvs/pull/628

The pattern is that the api entry points will throw UOE on platforms where there is no support, e.g. < JDK 22, Mac, Windows, or if the native library is not installed.

lucene/sandbox/src/java/org/apache/lucene/sandbox/vectorsearch/CuVSKnnFloatVectorQuery.java
```diff
@@ -0,0 +1,53 @@
...
+import java.io.IOException;
+import org.apache.lucene.index.LeafReader;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.search.KnnFloatVectorQuery;
+import org.apache.lucene.search.TopDocs;
+import org.apache.lucene.search.knn.KnnCollectorManager;
+import org.apache.lucene.util.Bits;
+
+/** Query for CuVS */
+/*package-private*/ class CuVSKnnFloatVectorQuery extends KnnFloatVectorQuery {
```
chatman, line 28: @ChrisHegarty for a user / downstream project of Lucene, how should he/she call this query? Should they use a KnnFloatVectorQuery, which internally uses this?

benwtrent: This is a POC, for sure, a package private query won't work. 

Additionally, I am not sure why it would need a custom collector or custom collection manager at all. I think the approximate search interface as it stands is perfectly fine. 

If the format requires weird collection manipulation, that smells like a design bug to me.

rmuir:
> P.S. I'd like to bite into the apple and make Java 22 minimum requirement.

At least for main branch what is the harm? could we do 23 or 24? I'd really like https://openjdk.org/jeps/467 !

benwtrent:
I think bumping main only for each non LTS release would be cool. Then we keep it at the next LTS (Java 25)?

Or, if its just as long from Lucene 10 -> 11, more likely the next LTS for Lucene 11 would be Java 30 🙈 . Maybe project Valhalla and Panama will be done by then.

My only concerns are:

 -  be developer complications (requiring devs to bump java versions all the time, which maybe they should be anyways...) 
 - backport complications
 - Friendliness to 3rd party CI stuff (folks who perf test and test against main for some reason, maybe they shouldn't and just stick to 10x).

ChrisHegarty:
> I think bumping main only for each non LTS release would be cool. Then we keep it at the next LTS (Java 25)?

I filed the following issue to help facilitate the discussion related to bumping the minimum compile version ( since it's no longer relevant in this PR - cvus-java is now at Java 21 ) - https://github.com/apache/lucene/issues/14229

ChrisHegarty:
I just committed a rewrite for the cuVS format implementation.

After the rewrite all the BaseKnnVectorsFormatTestCase tests pass. There are still some lurking intermittent failures, but the tests pass successfully the majority of the time.

Summary of the most significant changes:

1. Use the flat vectors reader/writer to support the raw float32 vectors and ordinal to docId mapping. This is similar to how HNSW is supported in Lucene. And keeps the code aligned with how other formats are layered atop each other.
2. The cuVS indices (Cagra, brute force, and HNSW) are stored directly in the format, so can be mmap'ed directly.
3. Merges are physical, all raw vectors are retrieved and used to create new cuVS indices.
4. A standard KnnCollector is used, no need for a special one for cuVS, unless one wants to customise some very specific parameters.

A number of workarounds have been put in place, which will eventually be lifted.

1. pre-filter and deleted docs over sample the topK, since the cuvs-java do not yet support a pre-filter.
2. Ignore Cagra failures indexing with small numbers of docs, fail over to just brute force.
3. We need to move to the cuvs-java merge api, to avoid bringing the vectors on-heap during merge.

zacksiri:
How is this coming along I'm looking forward to this shipping.

chatman:
FYI, there were some important building blocks missing in cuVS that were needed for a good first class integration. As a result, this PR represented several inefficiencies and workarounds.

At the moment, we are working on the cuVS side of things that are needed to make this production ready. Once they are done and released, we can open a new PR. For now, this PR represents a reasonable proof of concept.

Important pieces that were not available for use during this PR:
* Prefiltering in CAGRA (done)
* CAGRA Merge API (done)
* Efficient dataset loading (avoiding float[][]) while creating the CAGRA index (done, also some improvements WIP)

In future, in cuVS, we plan to work on:
* Quantization support
* Multi-GPU support
* HNSW index build support using GPU

I'm closing this PR here, and shall open a new PR soon, containing some of these changes incorporated.

FYI @ChrisHegarty @benwtrent.

PR #14132, opened 2025-01-10
Add nullability annotations to IndexSearcher APIs

Edarke:
Annotate public APIs with an internal definition of `@Nullable` where appropriate.

This proposal is a follow up from the discussion [here](https://lists.apache.org/thread/oh09k7nlvntzf49gx7d399d1chdk1z0n). This change adds an internal copy of JSpecify's `@Nullable` to lucene-util and uses Intellij's "infer nullability" feature to automatically apply the annotation to `IndexSearcher` and its dependencies. If this looks good, there will be a follow up PR to automatically annotate other APIs.

The motivation for this change is to increase the null safety of downstream projects using Lucene by allowing users to configure a javac plugin to lint their build. This change was tested with NullAway, which recognizes any annotation with the simple class name `@Nullable`

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
+++ b/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
@@ -581,7 +581,8 @@ public class IndexSearcher {
    * @throws TooManyClauses If a query would exceed {@link IndexSearcher#getMaxClauseCount()}
    *     clauses.
    */
-  public TopDocs searchAfter(@Nullable ScoreDoc after, Query query, int numHits) throws IOException {
+  public TopDocs searchAfter(@Nullable ScoreDoc after, Query query, int numHits)
+      throws IOException {
     final int limit = Math.max(1, reader.maxDoc());
     if (after != null && after.doc >= limit) {
       throw new IllegalArgumentException(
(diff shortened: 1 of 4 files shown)


msokolov:
This seems reasonable to me, but it implies a future promise to maintain it, and I don't know how we would ever know if we added some new usage that isn't properly annotated

rmuir:
For the java language:

Specifying that something is nullable doesn't provide any value: all types are nullable by default already.

Specifying that something is NOT-nullable would allow for additional smarts: it is the only possible way to do "gradual nullness typing".

I agree too, that we would need linter to verify this stuff in builds. Maybe the errorprone could do it.

rmuir:
One option we could do for correctness is to turn on ecj's null analysis. ecj is pretty fast and runs as part of gradle checks already, and it is the compiler often using this feature in the IDEs anyway.

https://github.com/apache/lucene/blob/main/gradle/validation/ecj-lint/ecj.javadocs.prefs#L2-L10

rmuir:
I will look in on the ECJ side, last time I used it, their null analysis had issues on lucene's enormous codebase. It was many years ago though, maybe it has solidified.

error-prone seems to only support Nullable-style vs NotNullable-style annotations, but maybe just reflects google's choices. It does have some good checks though, I esp like the ones where Nullable is interpreted differently by different tools, e.g.
https://errorprone.info/bugpattern/NullableWildcard
https://errorprone.info/bugpattern/NullableTypeParameter

PR #14138, opened 2025-01-14
Implement IntersectVisitor#visit(IntsRef) whenever it makes sense

iverase:
In https://github.com/apache/lucene/pull/13149, the method Implement IntersectVisitor#visit(IntsRef) was introduced which is a nice. trick to avoid multiple virtual calls to IntersectVisitor#visit(int).

This was only implemented for PointRangeQuery but it is useful for all implementation that expects hitting that code path. Therefore in this PR we proposed to implement this method wherever it make sense.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -61,6 +61,10 @@ Improvements
 
 * GITHUB#14113: Remove unnecessary ByteArrayDataInput allocations from `Lucene90DocValuesProducer$TermsDict.decompressBlock`. (Ankit Jain)
 
+* GITHUB#14138: Implement IntersectVisitor#visit(IntsRef) in  many of the current implementations and add
+  BulkAdder#add(IntRef) method. They should provide better perfromance due to less virtual method calls and
+  more efficient bulk processing. (Ignacio Vera)
+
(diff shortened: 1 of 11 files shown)


jpountz:
This makes sense to me, if the `visit()` call is inlined, then feeding the IntsRef into either an int[] (`BufferAdder`) or a `FixedBitSet` (`BitSetAdder`) can be auto-vectorized. Otherwise it can't.

This makes me wonder if we should instead introduce `BulkAdder#add(IntsRef)` and implement these `visit(IntsRef)` methods by calling `adder.add(intsRef)` directly, this would make auto-vectorizing less likely to get disabled by an apparently innocent change.

iverase:
Unrelated test error: ```

iverase:
I made BulkAdder sealed so we are sure they will always just two implementations.

iverase:
Actually we can make BulkAdder a sealed interface and the implementations are Java records.

jpountz (approved):

jpountz:
> Unrelated test error

@jimczi This test is failing because we go through a slightly different code path on the first doc of a segment, which generates a slightly different exception message it seems.

PR #14141, opened 2025-01-15
Publish build scans to develocity.apache.org

clayburn:
### Description

This PR migrates the Lucene project to publish Build Scans to the the new Develocity instance at develocity.apache.org.

Additionally, this PR migrates from the legacy Gradle Enterprise plugin to the renamed Develocity plugin and sets a projectId for use by Develocity.

@dweiss - I tested [here](https://github.com/clayburn/lucene/actions/runs/12789297319) on my fork (albeit with a different Develocity instance I have access to)

### DIFF
--- a/.github/workflows/run-checks-all.yml
+++ b/.github/workflows/run-checks-all.yml
@@ -13,7 +13,7 @@ on:
       - 'branch_10x'
 
 env:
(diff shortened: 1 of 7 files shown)


dweiss:
Thank you. One of the jobs is failing due to #14144 - I'll fix that and return.

dweiss:
Applied to 10x and main - thank you. Seems to work too.

PR #14147, opened 2025-01-16
move MultiLeafKnnCollector to decorator and remove unnecessary code

benwtrent:
MultiLeafKnnCollector was missed during the switch to the KnnCollector.Decorator.

Fixes this and removes duplicate.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/knn/MultiLeafKnnCollector.java
+++ b/lucene/core/src/java/org/apache/lucene/search/knn/MultiLeafKnnCollector.java
@@ -19,7 +19,6 @@ package org.apache.lucene.search.knn;
 
 import org.apache.lucene.search.AbstractKnnCollector;
 import org.apache.lucene.search.KnnCollector;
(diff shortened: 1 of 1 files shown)


lucene/core/src/java/org/apache/lucene/search/knn/MultiLeafKnnCollector.java
```diff
@@ -77,6 +76,7 @@ public MultiLeafKnnCollector(
       int interval,
       BlockingFloatHeap globalSimilarityQueue,
       AbstractKnnCollector subCollector) {
+    super(subCollector);
```
cpoerschke, line 79: Perhaps `collector` in the decorator could be given protected visibility and then the `subCollector` member in this class would not be needed.

benwtrent: For this particular use, it needs to be a `AbstractKnnCollector` which has different methods than the regular `KnnCollector`. So, changing the visibility doesn't really simplify things much.

PR #14154, opened 2025-01-20
Add UnwrappingReuseStrategy for AnalyzerWrapper

mayya-sharipova:
Add a new reuse strategy WrappingReuseStrategy that consults 
he wrapped analyzer's strategy to decide if components can be 
reused or need to be updated.

Add an optional constructor to CompletionAnalayzer that uses this strategy.


john-wagster (approved):
LGTM

lucene/suggest/src/java/org/apache/lucene/search/suggest/document/CompletionAnalyzer.java
```diff
@@ -112,6 +116,25 @@ public CompletionAnalyzer(
...
+      boolean preservePositionIncrements,
+      ReuseStrategy fallbackStrategy) {
+    super(new WrappingReuseStrategy(fallbackStrategy));
+    // häckidy-hick-hack, because we cannot call super() with a reference to "this":
+    ((WrappingReuseStrategy) getReuseStrategy()).setUp(this, analyzer, wrappedComponents);
+    this.analyzer = analyzer;
+    this.preserveSep = preserveSep;
+    this.preservePositionIncrements = preservePositionIncrements;
+    this.maxGraphExpansions = ConcatenateGraphFilter.DEFAULT_MAX_GRAPH_EXPANSIONS;
+  }
```
benwtrent, line 136:
```
   /**
   * JAVA DOCS about when you should use this and what it does
   */
   public static CompletionAnalyzer withWrappingReuseStrategy(
      Analyzer analyzer,
      boolean preserveSep,
      boolean preservePositionIncrements,
      ReuseStrategy fallbackStrategy) {
    CompletionAnalyzer completionAnalyzer = new CompletionAnalyzer(analyzer, preserveSep, preservePositionIncrements, fallbackStrategy);
    ((WrappingReuseStrategy)completionAnalyzer.getReuseStrategy()).setUp(completionAnalyzer, analyzer, completionAnalyzer.wrappedComponents);
    return completionAnalyzer;
  }
```

What if we made the ctor private, removed the setup hack and did it instead within a public static method?

lucene/core/src/java/org/apache/lucene/analysis/AnalyzerWrapper.java
```diff
@@ -151,4 +152,57 @@ protected final Reader initReaderForNormalization(String fieldName, Reader reade
...
+    }
+
+    public void setUp(
+        AnalyzerWrapper wrapper,
+        Analyzer wrappedAnalyzer,
+        CloseableThreadLocal<TokenStreamComponents> wrappedComponents) {
+      this.wrapper = wrapper;
+      this.wrappedAnalyzer = wrappedAnalyzer;
+      this.wrappedComponents = wrappedComponents;
+    }
```
benwtrent, line 179: This bugs me. We allow a reuse strategy to be mutated after construction. It seems ready to cause a bug.

lucene/core/src/java/org/apache/lucene/analysis/AnalyzerWrapper.java
```diff
@@ -151,4 +152,57 @@ protected final Reader initReaderForNormalization(String fieldName, Reader reade
...
+
+    @Override
+    public TokenStreamComponents getReusableComponents(Analyzer analyzer, String fieldName) {
+      if (analyzer == wrapper) {
```
benwtrent, line 183: Do we really need to know the instance, or can we do `Analyzer analyzer instanceof WrapperAnalyzer wrapperAnalyzer` then get the strategy from `wrapperAnalyzer.getWrappedAnalyzer(fieldName).getReuseStategy().getReusableComponents...`?

lucene/core/src/java/org/apache/lucene/analysis/AnalyzerWrapper.java
```diff
@@ -151,4 +152,57 @@ protected final Reader initReaderForNormalization(String fieldName, Reader reade
...
+    public void setReusableComponents(
+        Analyzer analyzer, String fieldName, TokenStreamComponents components) {
+      if (analyzer == wrapper) {
+        setStoredValue(analyzer, components);
+        wrappedAnalyzer
+            .getReuseStrategy()
+            .setReusableComponents(wrappedAnalyzer, fieldName, wrappedComponents.get());
+      } else {
+        fallbackStrategy.setReusableComponents(analyzer, fieldName, components);
+      }
```
benwtrent, line 205: Do we really need to know the instance, or can we do `Analyzer analyzer instanceof WrapperAnalyzer wrapperAnalyzer` then get the wrapped analyzer via `getWrappedAnalyzer(fieldName)`?

lucene/core/src/java/org/apache/lucene/analysis/AnalyzerWrapper.java
```diff
@@ -151,4 +152,57 @@ protected final Reader initReaderForNormalization(String fieldName, Reader reade
...
     return getWrappedAnalyzer(fieldName).attributeFactory(fieldName);
   }
+
+  /**
+   * A {@link org.apache.lucene.analysis.Analyzer.ReuseStrategy} that checks the wrapped analyzer's
+   * strategy for reusability. If the wrapped analyzer's strategy returns null, components need to
+   * be re-created. During components creation, this analyzer must store the wrapped analyzer's
+   * components in {@code wrappedComponents} local thread variable.
+   */
+  public static final class WrappingReuseStrategy extends ReuseStrategy {
```
benwtrent, line 162: It seems to me that the better name is `UnWrapping` as if we find that the analyzer is a wrapped analyzer, we unwrap it and use the underlying analyzer?

lucene/core/src/java/org/apache/lucene/analysis/AnalyzerWrapper.java
benwtrent, line 179: Ah, I see that the "wrappedComponents" is key here. its tricky, as the analyzer provides access directly to `storedValue` and there is no way to override :(. 

That seems like a bad API already, but we don't want to change all these long lived interfaces unnecessarily.

benwtrent (commented):
The API is a little concerning. I do not immediately know a better way to handle both wrappedComponents and also the inner storedValue :(

jpountz:
I don't like that `CompletionAnalyzer` needs to track a thread-local, the point of reuse strategy is to avoid this kind of thing. Also I'm not sure I understand why `CompletionAnalyzer` is different from other analyzer wrappers?

mayya-sharipova:
@benwtrent Thanks for the review, I am not happy with the design either, will see how I can incorporate your feedback. 

> I don't like that CompletionAnalyzer needs to track a thread-local, the point of reuse strategy is to avoid this kind of thing. Also I'm not sure I understand why CompletionAnalyzer is different from other analyzer wrappers?

@jpountz  Thanks for looking.   The idea of having an extra thread-local is for AnalyzerWrapper to keep track of `wrappedComponents`, but may be it doesn't need to be thread-local and can be just a local variable.  
You are right, `CompletionAnalyzer` is not different from other analyzer wrappers, and I can extend this policy to all AnalyzerWrapper. `CompletionAnalyzer` is where we experienced a problem in Elasticsearch, where it did not respect a reuse policy of the wrapped analyzer.

mayya-sharipova:
@jpountz @benwtrent I've addressed your comments in the last commit, please continue to review

lucene/core/src/java/org/apache/lucene/analysis/AnalyzerWrapper.java
```diff
@@ -151,4 +160,46 @@ protected final Reader initReaderForNormalization(String fieldName, Reader reade
...
+        Analyzer analyzer, String fieldName, TokenStreamComponents components) {
+      reuseStrategy.setReusableComponents(analyzer, fieldName, components);
+      if (analyzer instanceof AnalyzerWrapper wrapper) {
+        final Analyzer wrappedAnalyzer = wrapper.getWrappedAnalyzer(fieldName);
+        wrappedAnalyzer
+            .getReuseStrategy()
+            .setReusableComponents(
+                wrappedAnalyzer, fieldName, wrapper.getWrappedComponents(fieldName));
+      }
+    }
```
benwtrent, line 203: I am not gonna lie, I have a very hard time understanding all this reusable stuff. It seems like we are storing things twice?

The analyzer that is wrapped has a `storedValue` thread local field. That `storedValue` is conditionally globally or to a hashmap that is per field. The stored value can be anything really.

So, here we store the components of the analyzer. And, if the analyzer is a wrapper, we will un-wrap it, look at a hashmap that isn't thread safe to get the wrapped components. 

But, aren't the components added to `setReusableComponents` already the wrapped ones?

Are we trying to store for reuse both the wrapped components and the unwrapped ones?

This is very confusing to me.

benwtrent (commented):
I am sorry @mayya-sharipova I am having a difficult time understanding the change and what its supposed to be doing.

Is the idea that we don't reuse wrapped components if the unwrapped component themselves cannot be reused?

lucene/core/src/java/org/apache/lucene/analysis/AnalyzerWrapper.java
mayya-sharipova: Addressed in the last commit.

mayya-sharipova:
I syncrhonized with @benwtrent offline, and  we decided to use to extend TokenStreamComponents, so that it has access to wrapped components.

@benwtrent please review the last changes. Thank you.

benwtrent:
I will have to review later, but first glance is way cleaner than before!

Thank you

benwtrent (commented):
I have a couple of comments, but its looking good.

lucene/core/src/java/org/apache/lucene/analysis/AnalyzerWrapper.java
```diff
@@ -151,4 +156,62 @@ protected final Reader initReaderForNormalization(String fieldName, Reader reade
...
+      return reuseStrategy.getReusableComponents(analyzer, fieldName);
+    }
+
+    @Override
+    public void setReusableComponents(
+        Analyzer analyzer, String fieldName, TokenStreamComponents components) {
+      reuseStrategy.setReusableComponents(analyzer, fieldName, components);
+
+      if (analyzer instanceof AnalyzerWrapper wrapper) {
+        final TokenStreamComponentsWrapper wrapperComponents =
```
benwtrent, line 190:
```suggestion
        assert components instanceof TokenStreamComponentsWrapper;
        final TokenStreamComponentsWrapper wrapperComponents =
```

Lets force tests to fail on an assertion instead of a casting error.

benwtrent (approved):
Looks so much better! Thank you for iterating on the API. I like it.

john-wagster (approved):

jpountz:
I don't feel good about this change (at least not yet). It looks like there is an analyzer that changes its components over time somewhere, and this change aims at making sure that an analyzer that wrap this analyzer would also update its components. I worry that this is just one place where the assumption that analysis components are immutable is violated, and that there are other ones that haven't been found yet. I'm curious if other alternatives were considered, e.g. atomically swapping the analyzer or updating the behavior of analysis components without needing to replace them.

For reference, a side-effect of this change is that analysis components are now cached both on the wrapper and on the wrapped analyzer, while they were only cached on the wrapper before.

mayya-sharipova:
@jpountz Thank you for your feedback. I am happy to discuss alternatives. 

Isn't the whole idea of ReuseStrategy to decide if an analyzer's components can be reused or should be recreated/swapped?  

> I worry that this is just one place where the assumption that analysis components are immutable is violated, and that there are other ones that haven't been found yet. I'm curious if other alternatives were considered, e.g. atomically swapping the analyzer or updating the behavior of analysis components without needing to replace them.

Yes, indeed in this PR analysis components are immutable, they are recreated/ swapped in an analyzer

----

For the context: this PR came from a need to have an updateable synonyms analyzer when it is wrapped  in another analyzer, for example CompletionAnalyzer. In Elasticsearch an analyzer with updateable synonyms has a custom ReuseStrategy that creates a new components when synonyms are updated. We now want CompletionAnalyzer also update its components when its wrapped synonyms analyzer updated synonyms.

jpountz:
I suspect that `ReuseStrategy` hasn't been built with the idea that it could be use to invalidate components, but only as a way to describe an efficient caching strategy for analysis components, typically whether analysis components can be cached globally (`GLOBAL_REUSE_STRATEGY`) or need a different cache entry on each field (`PER_FIELD_REUSE_STRATEGY`).

@uschindler I believe that you are the one who designed this logic, I'd be interested in your thoughts.

Issue #14127, opened 2025-01-09
TestBPReorderingMergePolicy fails CheckIndex.testHnswGraph


msokolov:
### Description

```
TestBPReorderingMergePolicy > testReorderOnAddIndexes FAILED
    org.apache.lucene.index.CheckIndex$CheckIndexException: Field "vector" has repeated neighbors of node 6978 with value 7004
        at __randomizedtesting.SeedInfo.seed([B1FA21A9AAC314F2:A0D1EE47A3A0C942]:0)
        at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CheckIndex.testHnswGraph(CheckIndex.java:2931)
        at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CheckIndex.testHnswGraphs(CheckIndex.java:2806)
        at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CheckIndex.testSegment(CheckIndex.java:1125)
        at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CheckIndex.checkIndex(CheckIndex.java:822)
        at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CheckIndex.checkIndex(CheckIndex.java:592)
        at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.util.TestUtil.checkIndex(TestUtil.java:338)
        at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.store.MockDirectoryWrapper.close(MockDirectoryWrapper.java:921)
        at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.util.IOUtils.close(IOUtils.java:85)
        at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.util.IOUtils.close(IOUtils.java:72)
        at app//org.apache.lucene.misc.index.TestBPReorderingMergePolicy.testReorderOnAddIndexes(TestBPReorderingMergePolicy.java:241)
```

B1FA21A9AAC314F2

### Gradle command to reproduce

./gradlew :lucene:misc:test --tests "org.apache.lucene.misc.index.TestBPReorderingMergePolicy" -Dtests.seed=B1FA21A9AAC314F2

[...]

PR #14157, opened 2025-01-21
gh-14127: remove duplicate neighbors when writing HNSW graphs

msokolov:
should fix #gh-14127 test failures.

I believe there are no back-compat concerns here since this would be a no-op for graphs with no duplicates (I think that is what we were always producing before), and even if we did produce dups before, this will produce a functionally-equivalent graph, just a bit more compact.

I also don't think we need to take any special action w.r.t. to the CheckIndex check that was added since if we never produced dups without reordering then it won't randomly start failing on indexes that checked out OK before.  Although there is some chance I'm wrong - it's difficult to *prove* we could not have produced duplicates before, I think we would have seen problems by now with the CheckIndex check having been out in the wild on main and 10x branch?

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
@@ -408,6 +408,7 @@ public final class Lucene99HnswVectorsWriter extends KnnVectorsWriter {
     // write vectors' neighbours on each level into the vectorIndex file
     int countOnLevel0 = graph.size();
     int[][] offsets = new int[graph.numLevels()][];
+    int[] scratch = new int[graph.maxConn() * 2];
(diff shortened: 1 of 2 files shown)


benwtrent (approved):

iverase:
I was thinking in a solution more like this: https://github.com/apache/lucene/pull/14159. Just open it for discussion, I am ok if the preferred solution is done at this level.

The only merit of the other solution is that we will still create new connections, not sure if there is a benefit on doing that.

msokolov:
@iverase I see what you did there ... that would also solve this problem, but I think it is less desirable since it (1) requires extending the HNSW search API in a way I think we wouldn't use elsewhere, and (2) overall seems to do a more work. I don't think we need the extra links, at least not to satisfy the intended use case here which is to guarantee some notion of connectivity (all nodes reachable from an entry point).

iverase:
Sounds good to me @msokolov, I didn't like to add  yet a new  parameter in the search api. Thanks for taking the time to review it.

PR #230, opened 2021-07-29
LUCENE-7020: Remove TieredMergePolicy#setMaxMergeAtOnceExplicit.

jpountz:
TieredMergePolicy no longer bounds the number of segments that can be merged via
a forced merge.

PR #14165, opened 2025-01-23
Remove `maxMergeAtOnce` option from `TieredMergePolicy`.

jpountz:
`maxMergeAtOnce` increases merge amplification by running multiple merges when it could run a single merge, without giving significant benefits in exchange. We removed this parameter for forced merges in #230. Let's row remove it for regular merges as well?

In practice, merges above the floor segment size will be bounded by the number of segments per tier, and merges below the floor segment size will be unbounded.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -9,6 +9,8 @@ API Changes
 ---------------------
 * GITHUB#11023: Removing deprecated parameters from CheckIndex. (Jakub Slowinski)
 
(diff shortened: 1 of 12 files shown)


mikemccand (approved):
It is indeed weird to have a separate control (`maxMergeAtOnce`) from `segsPerTier`.   I think the past motivation was because merging was somewhat RAM heavy, maybe?  It allocates stuff for the docid map ... but I think we've improved its ram efficiency, and anyway if that really is somehow a problem, users can lower `segsPerTier`.  Thanks @jpountz.

PR #14166, opened 2025-01-23
Allow `LogMergePolicy` to merge more than `mergeFactor` segments together when the merge is below the min merge size.

jpountz:
This is essentially porting #266 to `LogMergePolicy`. By allowing more than `mergeFactor` segments to be merged together for small merges, the merge policy gets a lower write amplification and indexes have fewer small segments.


jpountz:
@original-brownbear I remember you looked at indexes that had more segments (and more small segments in particular) than you would have liked, so you may be interested in this.

original-brownbear (approved):
I don't have massive experience with the performance impact of merging more segments but seeing how all of this involves tiny segments only anyway + code looks good => LGTM, thanks! Looking at the problematic segments I've observed here and there, this seems like it should prevent a good chunk of those :)

PR #12446, opened 2023-07-17
Enable rank-unsafe optimizations for MAXSCORE/WAND.

jpountz:
Both MAXSCORE and WAND can easily be tuned to perform rank-unsafe optimizations, by skipping doc IDs that are unlikely to make it to the top-k. The main challenge is how to expose this kind of optimization. One approach could consist of artificially increasing the minimum competitive score as suggested in the original WAND paper. The approach I'm considering here is to configure a target evaluation cost, giving the scorer a budget of documents that it can visit and asking it to compute the best hits it can identify with this budget.

This draft PR tries to give an idea of how it could look like. It's currently only implemented for our MAXSCORE implementation but could easily be ported to our WAND scorer too.

An interesting follow-up could be to integrate this into the timout mechanism, so that `IndexSearcher` would progressively reduce the target cost as the amount of remaining time reduces.

I'm interested in gathering feedback on this approach.

PR #14167, opened 2025-01-23
Add knn result consistency test

benwtrent:
Inspired by some weird behavior I have seen, adding a consistency test. 

I found that indeed, this fails over some seeds.

Frustratingly, the seeded failures do not seem to be repeatable. But, running 

```
./gradlew :lucene:core:test --tests "org.apache.lucene.search.TestSeededKnnFloatVectorQuery.testRandomConsistency" -Dtests.iters=1000
```

Results in failures, though, not consistently. This seems to indicate some funky race condition.

Obviously, this shouldn't be merged until we figure out the consistency issue.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
+++ b/lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
@@ -489,6 +489,7 @@ abstract class BaseKnnVectorQueryTestCase extends LuceneTestCase {
     assertRandomConsistency(false);
   }
 
+  @AwaitsFix(bugUrl = "https://github.com/apache/lucene/issues/14180")
   public void testRandomConsistencyMultiThreaded() throws IOException {
     assertRandomConsistency(true);
   }
--- a/lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
+++ b/lucene/core/src/test/org/apache/lucene/search/BaseKnnVectorQueryTestCase.java
@@ -23,6 +23,7 @@ import static org.apache.lucene.search.DocIdSetIterator.NO_MORE_DOCS;
 
 import java.io.IOException;
 import java.util.HashSet;
+import java.util.Random;
 import java.util.Set;
 import org.apache.lucene.codecs.KnnVectorsFormat;
 import org.apache.lucene.document.Document;
@@ -40,7 +41,9 @@ import org.apache.lucene.index.IndexWriterConfig;
 import org.apache.lucene.index.KnnVectorValues;
 import org.apache.lucene.index.LeafReader;
 import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.index.NoMergePolicy;
 import org.apache.lucene.index.QueryTimeout;
+import org.apache.lucene.index.SerialMergeScheduler;
 import org.apache.lucene.index.StoredFields;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.index.VectorEncoding;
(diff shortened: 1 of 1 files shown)


mikemccand:
> Frustratingly, the seeded failures do not seem to be repeatable. 

Hmm that is bad ... it means there is a test bug or test infra bug (separate from the scary bug this test is chasing!)?

Oh, maybe force `SerialMergeScheduler` to your `RandomIndexWriter`?  Since `CMS` (Lucene's default and RIW will sometimes pick that) launches threads and we don't know how to determinize JVM's/OS's thread scheduling, that might explain the non-reproducibility?  E.g.:

```
    IndexWriterConfig iwc = LuceneTestCase.newIndexWriterConfig(r, new MockAnalyzer(r)), true, r.nextBoolean();
    iwc.setMergeScheduler(new SerialMergeScheduler());
    RandomIndexWriter riw = new RandomIndexWriter(random(), dir, iwc);
```

or so?

msokolov:
As for the reproducibility problem, that may be caused by concurrent HNSW merging, which is nondeterministic.

msokolov:
I think our comments relate to the observation that the test does not reproducibly fail with the same seed

benwtrent:
> I think our comments relate to the observation that the test does not reproducibly fail with the same seed

🤦 for sure. Let me see if I can shore it up.

benwtrent:
OK, I cleaned it all up, and have two separate tests, one for multi-threaded one for single threaded.

The multi-threaded one is the only one that fails periodically, which explains the difficulty in replicating. Threads might be racing to explore their segments first and thus stop exploring other graphs sooner than other runs. 

As for the single-threaded, I haven't had it fail in 10s of thousands of runs. Which doesn't 100% mean there isn't an issue there as well. I just haven't had a failure yet.

benwtrent:
OK, if I change to never use `MultiLeafKnnCollector`, the multi-threaded consistency test passes. But with using that collector, it will fail a couple times over 10k+ repeats.

benwtrent:
@mayya-sharipova maybe a search time flag is possible, but it would stink to have a "inconsistent but fast" flag that users then have to worry about.

I don't know of another query where multiple passes over a static dataset can return different docs.

It seems that the default behavior should be consistency.

I think we need to do one of the following:

 - fix multi-threaded consistency with information sharing (my first choice if I had a magic wand)
 - turn off direct multi-threaded queries in kNN (my second choice)
 - turn off information sharing between segments (the last resort)

I would rather keep doing less work with information sharing and use less threads than to do more work while also using more threads. However, if I can magically have both, I prefer it.

I am curious to the opinions of others here: @jpountz @msokolov @mikemccand 

Another consideration is if this is enough for a bugfix in Lucene 9.12.x.

jpountz:
> I don't know of another query where multiple passes over a static dataset can return different docs.

Currently, this does not happen because Lucene only enables so-called "rank-safe" optimizations to top-k query processing for lexical search. So regardless of how search threads race with one another, `Top(ScoreDoc|Field)CollectorManager` are guaranteed to always return the same (correct) hits. However, would we enable "rank-unsafe" optimizations (e.g. https://github.com/apache/lucene/pull/12446), we would be observing the same issue that you are seeing here.

I suspect that users may indeed struggle with this behavior, e.g. if running the same query multiple times on an e-commerce website doesn't return the same hits every time. It probably makes it hard to write integration tests as well. I believe that the Anserini IR toolkit wouldn't be happy either given how much it cares about reproducibility. The direction that you are suggesting makes sense to me, I have no idea how hard it is.

jpountz:
Somewhat related, thinking out loud: I have been wondering about what is the best way to parallelize top-k query processing. Lexical search has a similar issue as knn search in that it is not very CPU-efficient to let search threads independently make similar decisions about what it means for a hit to be competitive. This made me wonder if it would be a better trade-off to let just one slice run on its own first, and then let all other N-1 slices run in parallel with one another, taking advantage of what we "learned" from processing the first slice. If these N-1 slices would only look at what we learned from this first slice and ignored everything about any other slice, I believe that there wouldn't be any consistency due to races while query processing would still be mostly parallel and likely more CPU-efficient (as in total CPU time per query).

benwtrent:
> This made me wonder if it would be a better trade-off to let just one slice run on its own first, and then let all other N-1 slices run in parallel with one another,

I really like this idea. For kNN search, it seems best to take the largest tiers, gather information from them, and then run the smaller tiers in parallel. 

The major downside of kNN is that there is no slicing at all. Every segment is just its own worker, which is sort of crazy. We should at a minimum combine all the tiny segments together into a single thread.

What do you think @mayya-sharipova ? Slicing the segments and then picking the "largest" slice and search that in current thread. Then using that information to help the future parallel threads?

msokolov:
I was thinking of another approach based on pro-rating. On its own this is deterministic and close to optimally efficient, but risks missing the best results when the index is skewed. If me that if the HNSW search could be made re-entrant, by preserving the state in the HnswSearcher (visited list, priority queues) then we could examine all the per-segment results after completing a pass through the graphs, and then revisit some segments more deeply if the results appear skewed. Basically the information-sharing would be done in a sequential, periodic fashion

benwtrent:
To aid in the conversation, I opened an issue: https://github.com/apache/lucene/issues/14180

I plan on merging this new test, but with the multi-threaded case muted until we can fix: https://github.com/apache/lucene/issues/14180

PR #14169, opened 2025-01-24
Optimize ContextQuery with big number of contexts

mayya-sharipova:
When there are big number of contexts, ContextQuery may take a 
lot of time because of how context automata are constructed. 
Instead of the current approach of repeatedly concatenating and unioning 
context automata, this PR first constructs all individual context automata
and then does one single union at the end.

Thus for the added test with 1000 contexts, the performance improved from 4000 ms to 18 ms.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -76,6 +76,8 @@ Optimizations
 * GITHUB#14133: Dense blocks of postings are now encoded as bit sets.
   (Adrien Grand)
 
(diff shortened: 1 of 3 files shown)


benwtrent (approved):
Please add CHANGES.txt under optimizations.

It makes sense to me that:

`Union(Union(Union(Union(Union(..., ctxN-3), ctxN-2), ctxN-1), ctxN)`

Is the same as `Union(ctx0, ..., ctxN)`

This is indeed tricky. Looks like we already have a fair number of multi-context tests as well to cover to make sure this didn't break any weird automaton logic.

PR #14170, opened 2025-01-24
Adjusts Seeded knn searches to clean up user and internal interfaces

benwtrent:
This is a bugfix and refactor for seeded knn searches. 

First, Since we are using collectors, we don't actually need unique queries for every input type. Consequently, I have collapsed the two individual seeded queries into a single query that delegates to a provided kNN query. Then the collector manager is simply wrapped, so that the entry points can be provided.

Second, the interactions in the hnsw graph were not clear. Consequently, I did a minor refactor of HNSW searcher to have a "SeededSearcher", where instead of searching the graph for the entry points, it provides them directly. 

Third, instead of continually overloading collectors, I opted to add a new "searchstrategy" value to KnnCollector. This way various strategies can be executed with different options. I think `Seeded` could eventually be replaced with something.


tteofili (approved):

john-wagster (approved):
LGTM

lucene/core/src/java/org/apache/lucene/search/SeededKnnVectorQuery.java
```diff
@@ -0,0 +1,321 @@
...
+   */
+  public static SeededKnnVectorQuery fromByteQuery(KnnByteVectorQuery knnQuery, Query seed) {
+    return new SeededKnnVectorQuery(knnQuery, seed, null);
+  }
+
+  SeededKnnVectorQuery(AbstractKnnVectorQuery knnQuery, Query seed, Weight seedWeight) {
+    super(knnQuery.field, knnQuery.k, knnQuery.filter);
+    this.seed = Objects.requireNonNull(seed);
+    this.delegate = knnQuery;
+    this.seedWeight = seedWeight;
```
cpoerschke, line 76: minor/subjective: init order to match member declaration order
```suggestion
    this.seedWeight = seedWeight;
    this.delegate = knnQuery;
```

lucene/core/src/java/org/apache/lucene/search/knn/KnnSearchStrategy.java
```diff
@@ -0,0 +1,63 @@
...
+      if (numberOfEntryPoints > 0 && entryPoints == null) {
+        throw new IllegalArgumentException("entryPoints must not be null");
+      }
+      this.entryPoints = entryPoints == null ? DocIdSetIterator.empty() : entryPoints;
+    }
+
+    /**
+     * Iterator of valid entry points for the kNN search
+     *
+     * @return DocIdSetIterator of entry points, default is empty iterator
```
cpoerschke, line 48:
```suggestion
     * @return DocIdSetIterator of entry points
```

lucene/core/src/java/org/apache/lucene/search/knn/KnnSearchStrategy.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public DocIdSetIterator entryPoints() {
+      return entryPoints;
+    }
+
+    /**
+     * Number of valid entry points for the kNN search
+     *
+     * @return number of entry points, default is 0
```
cpoerschke, line 57:
```suggestion
     * @return number of entry points
```

lucene/core/src/java/org/apache/lucene/util/hnsw/SeededHnswGraphSearcher.java
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @lucene.experimental
+ */
+final class SeededHnswGraphSearcher extends AbstractHnswGraphSearcher {
+
+  private final AbstractHnswGraphSearcher delegate;
+  private final int[] seedOrds;
+
+  static SeededHnswGraphSearcher fromEntryPoints(
+      AbstractHnswGraphSearcher delegate, int numEps, DocIdSetIterator eps, HnswGraph graph)
```
cpoerschke, line 38: maybe
```suggestion
      AbstractHnswGraphSearcher delegate, int numEps, DocIdSetIterator eps, int graphSize)
```

lucene/join/src/java/org/apache/lucene/search/join/DiversifyingNearestChildrenKnnCollector.java
```diff
@@ -42,7 +43,20 @@ class DiversifyingNearestChildrenKnnCollector extends AbstractKnnCollector {
...
+    this(k, visitLimit, null, parentBitSet);
+  }
+
+  /**
+   * Create a new object for joining nearest child kNN documents with a parent bitset
+   *
+   * @param k The number of joined parent documents to collect
+   * @param visitLimit how many child vectors can be visited
+   * @param parentBitSet The leaf parent bitset
+   * @param searchStrategy The search strategy to use
```
cpoerschke, line 55:
```suggestion
   * @param searchStrategy The search strategy to use
   * @param parentBitSet The leaf parent bitset
```

tteofili (approved):

PR #14172, opened 2025-01-27
Adjust knn merge stability testing

### LINKED ISSUE #13640: testMergeStability failing for Knn formats

benwtrent:
With the new connected components work, merging is unstable. It generally seems that between indexing, there might be new connections as `vex` index size increases.

I am not 100% sure the cause here. Opening this PR as one potential solution. Its interesting indeed that it occurs at all.

closes: https://github.com/apache/lucene/issues/13640


lucene/test-framework/src/java/org/apache/lucene/tests/index/RandomCodec.java
```diff
@@ -191,6 +193,9 @@ public PostingsFormat getPostingsFormatForField(String name) {
       // Safety:
       assert previousMappings.size() < 10000 : "test went insane";
     }
+    if (VERBOSE) {
+      System.out.println("RandomPostingsFormat: " + name + " -> " + codec);
```
msokolov, line 197: Isn't `System.out` forbidden? Or is it OK in tests?

msokolov (changes requested):
seems OK to mark this as unstable, but something is failing in the checks, maybe related to the print statements

lucene/test-framework/src/java/org/apache/lucene/tests/index/RandomCodec.java
benwtrent: Yeah, it's likely forbidden. I found this verbose output helpful in debugging. But I think it can be removed in preference to setting the info stream

I will fix up the PR and merge.

PR #14179, opened 2025-01-29
Do not enable security manager on JDK 24+

ChrisHegarty:
This commit avoids setting the security manager on JDK 24+ - since it is not longer possible to enable it in JDK 24+

This is the minimum required to start testing with JDK 24 EA.


dweiss (approved):

cwperks (approved):

uschindler:
Works well with RC of Java 24: 
https://jenkins.thetaphi.de/job/Lucene-main-Linux/53317/consoleFull

Also Java 25 EA is fine:
https://jenkins.thetaphi.de/job/Lucene-main-Linux/53323/consoleFull

uschindler:
In case we may need another Lucene 9 release and want to enable Jenkins we may need to backport.

PR #14181, opened 2025-01-29
Add updateable random scorer interface for vector index building

benwtrent:
As stated by @ChrisHegarty and @msokolov the amount of garbage we create during vector index creation is pretty astounding. 

This adjusts the interface to allow an "Updateable" random vector interface (@msokolov 's idea if I remember correctly) and refactors the usage to keep it threadsafe. 

I still need to do some larger scale tests, I imagine the actual indexing times are not effected much.

I think I caught all the weird edge cases.

### DIFF
--- a/lucene/codecs/src/java/org/apache/lucene/codecs/bitvectors/FlatBitVectorsScorer.java
+++ b/lucene/codecs/src/java/org/apache/lucene/codecs/bitvectors/FlatBitVectorsScorer.java
@@ -63,8 +63,7 @@ public class FlatBitVectorsScorer implements FlatVectorsScorer {
     private final byte[] query;
 
     BitRandomVectorScorer(ByteVectorValues vectorValues, byte[] query) {
-      this.query = new byte[query.length];
-      System.arraycopy(query, 0, this.query, 0, query.length);
+      this.query = query;
       this.bitDimensions = vectorValues.dimension() * Byte.SIZE;
       this.vectorValues = vectorValues;
     }
@@ -98,17 +97,21 @@ public class FlatBitVectorsScorer implements FlatVectorsScorer {
 
   static class BitRandomVectorScorerSupplier implements RandomVectorScorerSupplier {
     protected final ByteVectorValues vectorValues;
-    protected final ByteVectorValues vectorValues1;
+    protected final ByteVectorValues targetVectors;
 
     public BitRandomVectorScorerSupplier(ByteVectorValues vectorValues) throws IOException {
       this.vectorValues = vectorValues;
-      this.vectorValues1 = vectorValues.copy();
+      this.targetVectors = vectorValues.copy();
     }
 
     @Override
-    public UpdateableRandomVectorScorer scorer(int ord) throws IOException {
-      byte[] query = vectorValues1.vectorValue(ord);
-      return new BitRandomVectorScorer(vectorValues1, query);
+    public UpdateableRandomVectorScorer scorer(Integer ord) throws IOException {
+      byte[] query = new byte[vectorValues.dimension()];
+      if (ord == null) {
+        return new BitRandomVectorScorer(vectorValues, query);
+      }
+      System.arraycopy(targetVectors.vectorValue(ord), 0, query, 0, query.length);
+      return new BitRandomVectorScorer(targetVectors, query);
     }
 
     @Override
--- a/lucene/codecs/src/java/org/apache/lucene/codecs/bitvectors/FlatBitVectorsScorer.java
+++ b/lucene/codecs/src/java/org/apache/lucene/codecs/bitvectors/FlatBitVectorsScorer.java
@@ -105,13 +105,9 @@ public class FlatBitVectorsScorer implements FlatVectorsScorer {
     }
 
     @Override
-    public UpdateableRandomVectorScorer scorer(Integer ord) throws IOException {
+    public UpdateableRandomVectorScorer scorer() throws IOException {
       byte[] query = new byte[vectorValues.dimension()];
-      if (ord == null) {
-        return new BitRandomVectorScorer(vectorValues, query);
-      }
-      System.arraycopy(targetVectors.vectorValue(ord), 0, query, 0, query.length);
-      return new BitRandomVectorScorer(targetVectors, query);
+      return new BitRandomVectorScorer(vectorValues, query);
     }
 
     @Override
--- a/lucene/codecs/src/java/org/apache/lucene/codecs/bitvectors/FlatBitVectorsScorer.java
+++ b/lucene/codecs/src/java/org/apache/lucene/codecs/bitvectors/FlatBitVectorsScorer.java
@@ -26,6 +26,7 @@ import org.apache.lucene.util.Bits;
 import org.apache.lucene.util.VectorUtil;
 import org.apache.lucene.util.hnsw.RandomVectorScorer;
 import org.apache.lucene.util.hnsw.RandomVectorScorerSupplier;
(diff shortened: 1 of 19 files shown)


lucene/codecs/src/java/org/apache/lucene/codecs/bitvectors/FlatBitVectorsScorer.java
```diff
@@ -58,7 +59,7 @@ public RandomVectorScorer getRandomVectorScorer(
     throw new IllegalArgumentException("vectorValues must be an instance of ByteVectorValues");
   }
 
-  static class BitRandomVectorScorer implements RandomVectorScorer {
+  static class BitRandomVectorScorer implements UpdateableRandomVectorScorer {
     private final ByteVectorValues vectorValues;
     private final int bitDimensions;
     private final byte[] query;
```
benwtrent, line 65: agh, `query` isn't a copy here. need to fix that...

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswConcurrentMergeBuilder.java
```diff
@@ -190,12 +191,25 @@ private int getStartPos(int maxOrd) {
...
+    }
+
     @Override
     public void addGraphNode(int node) throws IOException {
       if (initializedNodes != null && initializedNodes.get(node)) {
         return;
       }
-      super.addGraphNode(node);
+      if (scorer == null) {
+        scorer = scorerSupplier.scorer(node);
```
msokolov, line 208: would it make sense to add this (maybe with node==0?) when creating the builder so it's never null

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
```diff
@@ -191,8 +191,14 @@ protected void addVectors(int minOrd, int maxOrd) throws IOException {
     if (infoStream.isEnabled(HNSW_COMPONENT)) {
       infoStream.message(HNSW_COMPONENT, "addVectors [" + minOrd + " " + maxOrd + ")");
     }
+    UpdateableRandomVectorScorer scorer = null;
     for (int node = minOrd; node < maxOrd; node++) {
-      addGraphNode(node);
+      if (scorer == null) {
+        scorer = scorerSupplier.scorer(node);
```
msokolov, line 197: ditto here - these lazy inits are okay, but proactively initializing is better

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
```diff
@@ -296,6 +282,30 @@ to the newly introduced levels (repeating step 2,3 for new levels) and again try
...
+       we always add all the outgoing links first before adding incoming link such that
+       when a search visits this node, it can always find a way out
+    4. If the node has level that is less or equal to graph level, then we're done here.
+       If the node has level larger than graph level, then we need to promote the node
+       as the entry node. If, while we add the node to the graph, the entry node has changed
+       (which means the graph level has changed as well), we need to reinsert the node
+       to the newly introduced levels (repeating step 2,3 for new levels) and again try to
+       promote the node to entry node.
+    */
+    UpdateableRandomVectorScorer scorer = scorerSupplier.scorer(node);
```
msokolov, line 305: how did we get away with not doing this before? Is it that we pushed the supplier.scorer() call deeper in the call graph?

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
```diff
@@ -463,7 +478,11 @@ private boolean connectComponents(int level) throws IOException {
 
           beam.clear();
           eps[0] = c0.start();
-          RandomVectorScorer scorer = scorerSupplier.scorer(c.start());
+          if (scorer == null) {
```
msokolov, line 481: oh yes, here it is -- again could we initialize outside the loop, maybe even initialize to all zeros? Can ScorerSupplier accept null??

msokolov (approved):
Thanks for tackling this! I have a few questions, possible suggestions for further improvement, but they shouldn't block anything.

lucene/core/src/java/org/apache/lucene/codecs/hnsw/DefaultFlatVectorScorer.java
```diff
@@ -90,23 +91,29 @@ public String toString() {
   private static final class ByteScoringSupplier implements RandomVectorScorerSupplier {
     private final ByteVectorValues vectors;
     private final ByteVectorValues vectors1;
-    private final ByteVectorValues vectors2;
     private final VectorSimilarityFunction similarityFunction;
 
     private ByteScoringSupplier(
         ByteVectorValues vectors, VectorSimilarityFunction similarityFunction) throws IOException {
       this.vectors = vectors;
       vectors1 = vectors.copy();
```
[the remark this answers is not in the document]
benwtrent: I can happily rename everywhere. We use the same pattern many times.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswConcurrentMergeBuilder.java
```diff
@@ -190,12 +191,25 @@ private int getStartPos(int maxOrd) {
...
+    @Override
+    public void addGraphNode(int node, UpdateableRandomVectorScorer scorer) throws IOException {
+      if (initializedNodes != null && initializedNodes.get(node)) {
+        return;
+      }
+      super.addGraphNode(node, scorer);
+    }
+
     @Override
     public void addGraphNode(int node) throws IOException {
```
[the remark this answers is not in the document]
benwtrent: I was unsure about adjusting the full interface :/ There are things that inherit from HNSW graph, and I wanted it to "just work" without any fuss.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswConcurrentMergeBuilder.java
benwtrent: Let me see if I can adjust the interface a bit to ensure that we allow a "null" scorer supplier that simply initializes the buffer.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
benwtrent: yeah, supplier got pushed all the way down everywhere. Creating new scorers during diversity checking, etc. So, a single node add would likely create many many short lived objects for no useful purpose.

benwtrent:
Ugh, I just tested recall over a larger data set with multiple indexing and merging threads. Recall is negatively effected by this change, indicating I have a silly bug somewhere.

msokolov:
sorry, one more idea; please feel to go ahead with what you have, but I wonder, now that the Scorer is updateable, do we even need to initialize it with an ord at all? Won't we generally speaking be updating it? If that were the case then we could change to `scorer()` and wouldn't need the null checks (it would always be the null case).

benwtrent:
OK, benchmarks show pretty much no difference when I was testing on my machine. But, recall numbers all checkout and the API is nicer and makes more sense for vector merging.

So, I am gonna merge.

ChrisHegarty:
Thank you @benwtrent  - let's wait for the next lucene nightly. If no perf improvement, that's ok. There should be a lot less garbage created, and CPU devoted to cleaning young heap regions.

PR #14187, opened 2025-01-30
Disable the query cache by default.

jpountz:
The query cache trades heap for faster queries. Given all the progress that has been made on making uncached queries faster (`IndexOrDocValuesQuery`, bitset encoding of blocks of postings, etc.), it's not obviously a good trade-off anymore. So I suggest that we make it an opt-in in Lucene 11.


msokolov:
I agree with the judgment, but maybe this just indicates we need to improve the cache! Still, until we can figure out how to do so, +1 to disable by default

lucene/CHANGES.txt
```diff
@@ -30,6 +30,10 @@ Bug Fixes
 
 * GITHUB#14075: Remove duplicate and add missing entry on brazilian portuguese stopwords list. (Arthur Caccavo)
 
+Changes in Runtime Behavior
+---------------------
+* GITHUB#14187: The query cache is now disabled by default. (Adrien Grand)
```
mikemccand, line 35: Maybe add a sentence about how to opt-in?  Something like "call `IndexSearcher.setQueryCache(new LRUQueryCache(1000, maxRamBytesUsed)) to get back to 10.x behavior" or so?  Or maybe add these details to `MIGRATE.md` instead?

mikemccand (approved):
My my how times have changed :)  RAM is precious, and it's now better spent on letting the OS keep the hot index pages hot, than remembering past query results!

jpountz:
For the record, another downside of the query cache: the fact that it caches per segment doesn't play nicely with intra segment concurrency. E.g. if you have a single large segment that is searched with N threads, putting clauses of the query into the cache would still run single-threaded.

sgup432:
>I agree with the judgment, but maybe this just indicates we need to improve the cache!

Agree here that we need to improve the cache itself. I had opened an issue [here](https://github.com/apache/lucene/issues/14222) with some suggestions. 

We can further consider using off-heap or even disk as a cache where we store precomputed values. A hybrid tiered cache model approach where we use a combination of on-heap and disk tier might also work pretty well. I personally favor the latter.

lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -77,7 +77,8 @@
 public class IndexSearcher {
 
   static int maxClauseCount = 1024;
-  private static QueryCache DEFAULT_QUERY_CACHE;
+  // Caching is disabled by default.
+  private static QueryCache DEFAULT_QUERY_CACHE = null;
   private static QueryCachingPolicy DEFAULT_CACHING_POLICY = new UsageTrackingQueryCachingPolicy();
```
msfroh, line 82: If the query cache is disabled by default, what do folks think about removing `DEFAULT_QUERY_CACHE`  altogether?

I feel like anyone who would benefit from caching probably has some kind of SearcherManager singleton that can hand the cache instance to IndexSearchers as they're created. I don't think we need a mutable global variable for it.

jpountz: Good question, I wondered about this too, and for `DEFAULT_CACHING_POLICY` as well. I agree that setting them in a centralized place where searchers get instantiated is a better practice. These defaults are still quite convenient for tests though, which don't have such a central place where searchers get created (not all tests rely on `LuceneTestCase#newSearcher`).

In general, I'm in favor of this, would looking into it in a follow-up work for you?

msfroh: > In general, I'm in favor of this, would looking into it in a follow-up work for you?

Sounds good to me, thanks!

dsmiley (commented):
I am glad to see this!  Libraries should be opt-in to caching IMO.

PR #14189, opened 2025-02-01
Bump floor segment size to 16MB.

jpountz:
This bumps the floor segment size from 2MB (`TieredMergePolicy`) / 1.6MB (`LogByteSizeMergePolicy`) to 16MB.

My motivation is that such small segment sizes don't make index structures actually helpful vs. linear scans, so we should avoid them. Furthermore, there has been progress on merging rules for segments below the floor size, in particular merge policies no longer perform quadratic merging (#900) so this change will not make indexing/merging absurdly slow if an application flushes tiny segments.


mikemccand (approved):
Maybe we could make this change in 10.x too?  It is not really a bwc break, improving merge selection...

jpountz:
Thanks for the feedback, I was hesitating. Let's pull this in 10.2 then.

jpountz:
None other than me being a bit anxious about side-effects, e.g. this floor segment size also affects the behavior of merge-on-full-flush. But it shouldn't be a big deal. I'll merge shortly!

jainankitk (approved):

PR #14193, opened 2025-02-04
Add Automata.makeCharSet/makeCharClass to optimize regexp

rmuir:
Previously caseless matching was implemented via code such as this:

```java
  Operations.union(Automata.makeChar('x'), Automata.makeChar('X'))
```

Proposed unicode caseless matching (#14192) implements it with repeated unions:

```java
  a1 = Operations.union(Automata.makeChar('x'), Automata.makeChar('X'))
  a2 = Operations.union(a1, Automata.makeChar('y'))
  a3 = Operations.union(a2, Automata.makeChar('Y'))
```
The union operation doesn't return a minimal automaton: improving union would always be nice, but this change offers a simple api for the task that returns half the number of states.

Before: caseless match of "a":
a_before

After:
a_after

Before: caseless match of "lucene":
lucene_before

After:
lucene_after

Just like the `union`, the `concatenate` adds some useless states, but they are less of a problem than the ones from before.

I didn't try anything more such as repeated union or kleene star, to see if I could make a really bad case, I felt like this was good enough, to get it to a better place. We can look at optimizing union/concatenate separately still, but that's always more dangerous and tricky.


rmuir:
also `makeCharUnion()` comes to mind as a compelling alternative name, since there is already a `makeStringUnion()`. Naming is hard. just want to get the idea out there, since caseless regex is a common usecase.

jpountz (approved):
This looks good to me. Worst-case scenario, we make this new API delegate to union() when we improve it. I'm curious if you considered passing the code points as var-args for convenience?

rmuir:
I considered it but then didn't use any varargs after Dawid's email about compiler performance problems coming from them.

rmuir:
I generalized this to `makeCharClass(int[],int[])`, added a "character class" node to use it instead of unioning many nodes, replaced the pre-built class functionality with it too.

The build failure is bogus: tests pass. It seems to be bug in compiler:
```
An exception has occurred in the compiler (21.0.5). Please file a bug against the Java compiler via the Java bug reporting page (https://bugreport.java.com/) after checking the Bug Database (https://bugs.java.com/) for duplicates. Include your program, the following diagnostic, and the parameters passed to the Java compiler in your report. Thank you.
```

rmuir:
That's error-prone that's broke trying to do some null analysis :)

rmuir:
anyway, I think this is the right path, rather than fight with union(), let's just get it out of our way. with this change union() is only used for union operator (`|`) and not internally.

rmuir:
My example for this one, if you have something like `[^a-gklM-O\s]`, with the case-insensitive flag maybe, it just calls the new `makeCharClass(int[],int[])` method and you get minimal automaton with the ranges you expect from state1->state2. No more walking up and down the parse tree and collecting sub-automata into lists and unioning them, no more sub-regexp creation and unioning for prebuilt classes.

rmuir:
I will figure out what angers the error-prone tomorrow. I am rusty with java, so this PR needs assistance LOL. but all the tests pass.

rmuir:
a few more notes:
* maybe we should deprecate `union(Automaton, Automaton)` and only leave `union(List<Automaton>)`. I see the former approach has proven trappy, let's guide developers to do it the faster way? The deprecation warning itself can be useful here to explain why. Nothing outside of tests uses this signature anymore.
* by having a way to create a character class node, if we wanted, we could give it additional boolean `@param complement True if the character class should be negated`. It would not help performance but would allow us to cleanup on the API side: the problematic complement() is needed and awesome for negated character classes, but the general algorithm is not sustainable and requires determinization. It is a crazy trap: complement() on one of these simple character classes is the perfect algorithm: totalize() to one dead state, then nuke the previous accept state. I definitely wrote the code and deleted it several times to just have this flag, to remove the general-purpose complement() trap elsewhere, which is not efficient at all for all situations. 
* the PR needs more tests but I fought the existing toString-based ones to exhaustion already. At least tests work and confirm it does what we expect
* warning: there may be dead code, that we could remove that I missed. We should look and try to fix RegExp class to not do everything package-private, because it makes it impossible for our tooling to detect dead code. If it used private correctly, tools would tell us if a function is unused by anything.

dweiss (approved):
This looks great. The only thing that caught my attention would be parallel arguments (from/to) - we could replace it with the more high-level and perhaps more obvious List<record CodepointRange>. But it's such a low-level class that perhaps we shouldn't try to be that fancy and just keep it as is... It's clear what the code is doing (to me at least).

rmuir:
I tried to update the error prone to fix its bugs, it is angry about the way we do gradle. I will YOLO my way thru this stuff. 

> The default --should-stop=ifError policy (INIT) is not supported by Error Prone, pass --should-stop=ifError=FLOW instead

rmuir:
Like i literally have no idea what this tool is trying to tell me there. But I think errorprone is broken, it depends on too many internals of the java compiler apis.

rmuir:
I would propose replacing this checker with ast-grep rules for whatever we need: it is not a good one. the use of internal java APIs is too crazy.

rmuir:
I opened https://github.com/apache/lucene/issues/14200 for the error-prone situation

lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
```diff
@@ -648,13 +645,16 @@ private Automaton toAutomaton(
...
         if (check(ASCII_CASE_INSENSITIVE)) {
-          a = toCaseInsensitiveChar(c);
+          a = Automata.makeCharSet(toCaseInsensitiveChar(c));
         } else {
           a = Automata.makeChar(c);
         }
         break;
       case REGEXP_CHAR_RANGE:
-        a = Automata.makeCharRange(from, to);
+        a = Automata.makeCharRange(from[0], to[0]);
```
mikemccand, line 654: Do we check that `from.length == 1` and `to.length == 1` somewhere?

lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
```diff
@@ -140,6 +141,32 @@ public static Automaton makeCharRange(int min, int max) {
     return a;
   }
 
+  /** Returns a new minimal automaton that accepts any of the provided codepoints */
+  public static Automaton makeCharSet(int[] codepoints) {
```
mikemccand, line 145: I guess it's harmless if you have the same codepoint more than once in the incoming array?  Automaton will dedup those dups in `a.finishState()`?  Maybe we should add a test case making sure that's OK ...

lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
```diff
@@ -828,13 +838,10 @@ void toStringBuilder(StringBuilder b) {
...
         b.append(s2).append(">");
         break;
-      case REGEXP_PRE_CLASS:
-        b.append("\\").appendCodePoint(from);
-        break;
     }
   }
 
-  /** Like to string, but more verbose (shows the higherchy more clearly). */
+  /** Like to string, but more verbose (shows the hierarchy more clearly). */
```
mikemccand, line 844: Oooh nice typo fix!

lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
```diff
@@ -1050,14 +1059,25 @@ static RegExp makeDeprecatedComplement(int flags, RegExp exp) {
...
+    return newLeafNode(flags, Kind.REGEXP_CHAR, null, c, 0, 0, 0, null, null);
   }
 
   static RegExp makeCharRange(int flags, int from, int to) {
     if (from > to)
       throw new IllegalArgumentException(
           "invalid range: from (" + from + ") cannot be > to (" + to + ")");
-    return newLeafNode(flags, Kind.REGEXP_CHAR_RANGE, null, 0, 0, 0, 0, from, to);
+    return newLeafNode(
+        flags, Kind.REGEXP_CHAR_RANGE, null, 0, 0, 0, 0, new int[] {from}, new int[] {to});
```
mikemccand, line 1070: Oh if this is the only place where we create `REXEGP_CHAR_RANGE` (the API is not public) then we don't need to otherwise check `from.length == 1` and `to.length == 1`.

lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
```diff
@@ -1195,60 +1215,132 @@ final RegExp parseCharClassExp() throws IllegalArgumentException {
...
+    } else if (peek("W")) {
+      // non-word: [^0-9A-Z_a-z]
+      starts.add(Character.MIN_CODE_POINT);
+      ends.add('0' - 1);
+      starts.add('9' + 1);
+      ends.add('A' - 1);
+      starts.add('Z' + 1);
+      ends.add('_' - 1);
+      starts.add('_' + 1);
+      ends.add('a' - 1);
```
mikemccand, line 1321: Heh this range is just one character: ` (back quote).

lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
```diff
@@ -1195,60 +1215,132 @@ final RegExp parseCharClassExp() throws IllegalArgumentException {
...
+      // whitespace: [\t-\n\r ]
+      starts.add((int) '\t');
+      ends.add((int) '\n');
+      starts.add((int) '\r');
+      ends.add((int) '\r');
+      starts.add((int) ' ');
+      ends.add((int) ' ');
+      next();
+    } else if (peek("S")) {
+      // non-whitespace: [^ \t\n\r]
```
mikemccand, line 1291: Hmm maybe fix this comment to put the space char at the end for consistency (the chars are sorted by their ascii/codepoint, ascending)?

lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
```diff
@@ -1195,60 +1215,132 @@ final RegExp parseCharClassExp() throws IllegalArgumentException {
...
+      starts.add(Character.MIN_CODE_POINT);
+      ends.add('\t' - 1);
+      starts.add('\n' + 1);
```
mikemccand, line 1294: Maybe add a comment that you do not do the `\t-\n` case because they are adjacent codepoints (`0x9` and `0xa`) so there are no characters in the negated range in between them?

mikemccand (approved):
This looks great!  I left minor comments... thanks @rmuir.

Crazy how union makes dead states like that, but leave that battle for another day...

lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
rmuir: I have a test for dups :) I don't have many tests but I tested that one.

lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
rmuir: yeah it is. The only reason I didn't nuke these RANGE and CHAR nodes is ... toString()!

rmuir:
OK I tried out a List-based API  as alternative to array-based API. It isn't fully correct, which is part of my issue, but see it here: https://github.com/apache/lucene/commit/8b535a1c2fb4008dbe4e0b007ac95155dfe640ba

I was excited that it might cleanup the code, but it presented some problems. I think most problems are because we are talking about `int` arrays (codepoint ranges), but common use cases will of course use java `char`. 

array-based API:
```java
Automata.makeCharSet(new int[] {'a', 'A'});
```

List-based API:
```java
Automata.makeCharSet(List.of((int) 'a', (int) 'A'));
```

It requires user to make casts because the boxed types in java suck.

Also I'm not happy that with the List-based API, error-prone keeps catching me doing this:
```
error: [BoxedPrimitiveEquality] Comparison using reference equality instead of value equality. Reference equality of boxed primitive types is usually not useful, as they are value objects, and it is bug-prone, as instances are cached for some values but not others.
          if (from.get(i) == to.get(i)) {
                          ^
    (see https://errorprone.info/bugpattern/BoxedPrimitiveEquality)
  Did you mean 'if (Objects.equals(from.get(i), to.get(i))) {' or 'if (from.get(i).equals(to.get(i))) {'?
```

It happens because I keep forgetting to call `.intValue()` in special places.

So, I'd like to stick with array API, net-net I think it is better. I will look at messy List code that I have in this parser and see if I can improve that to reduce chances of problems when parsing lists of ranges. I also have other changes in that commit that are unrelated good ones and will poach them into here.

lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
rmuir: I looked. toString is a problem. so I want to leave this one be for now. To me it also doesn't make sense to keep adding more "variables" to the RegExp class , it has too many already for these nodes.

rmuir:
Finally! The concatenate() issue was an easy fix, it neglected to clean up its dead states. All of its partners in crime do this, but the fact we neglect it for concatenate messes up too many As.

Screen_Shot_2025-02-05_at_23 55 55

rmuir:
I'm feeling good about this one now, with the change, a lot of regexps now come out minimal from the start, which is a good thing. 

We also eliminate overhead of tons of nodes, which is important if we ever want to support caseless range matches (e.g. `[a-z]` with case-insensitive flag matching 'A'). To implement it generally (not just for ascii), we need to iterate the range and add tons of alternatives, but its an `int[]` and will all be folded into a single state: basically as good as we can get.

lucene/core/src/test/org/apache/lucene/util/automaton/TestAutomaton.java
```diff
@@ -667,11 +667,14 @@ public void testConcatenatePreservesDet() throws Exception {
...
+    Automaton a = new Automaton();
+    int s1 = a.createState();
+    a.createState(); // create dead state
+    a.setAccept(s1, true);
+    a.finishState();
+    assertEquals(2, a.getNumStates());
     a = Operations.removeDeadStates(a);
-    assertEquals(3, a.getNumStates());
+    assertEquals(1, a.getNumStates());
   }
```
rmuir, line 678: cc @jpountz this is how i fixed this test.

jpountz: Much better, thanks.

dweiss:
> Finally! The concatenate() issue was an easy fix, it neglected to clean up its dead states. All of its partners in crime do this, but the fact we neglect it for concatenate messes up too many As.
> 
> Screen_Shot_2025-02-05_at_23 55 55

God this is so much nicer compared to that first/ original automaton you included!

Issue #13002, opened 2024-01-09
migrate OpenNLP 'ant train-test-models' to Gradle


cpoerschke:
### Description

ref: https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.11.2/lucene/analysis/opennlp/build.xml#L52-L84

PR #14194, opened 2025-02-04
Add posTagFormat parameter for OpenNLPPOSFilter

### LINKED ISSUE #14188: Benefit from OpenNLP's new UD models
### Description

We recently upgraded Lucene's dependency on OpenNLP to 2.5. This upgrade offers a new part-of-speech tagging model that works across more languages. The update maintained backward compatibility with the old Penn model by hardcoding it in the token filter.

We should expose the UD model as an option.

I'd like to work on this, so please assign it to me.

@mawiesne, you're the OpenNLP expert, please let me know about potential pitfalls. I don't know OpenNLP, but this seems fun.

msfroh:
### Description

This allows users to use either a Penn or UD part-of-speech tagging model, but output tags in the other format. This allows users to combine a Penn POS tagging model with a lemmatizer model trained on UD tags, for example.

For a quick reference on the two:
* Penn: https://cs.nyu.edu/~grishman/jet/guide/PennPOS.html
* UD: https://universaldependencies.org/u/pos/

The conversion rules are also defined in https://github.com/apache/opennlp/blob/6daacd319b95c5937abca5ef99e24566825fe89f/opennlp-tools/src/main/java/opennlp/tools/postag/POSTagFormatMapper.java#L40

This commit also changes the default POSTagFormat to `CUSTOM` (whereas I previously set it to `PENN`), which just passes through the tag format from the POSTaggerModel. I believe this is a reasonable default, since new users are likely to use just the new UD models published at https://opennlp.apache.org/models.html, whereas existing users likely have Penn models

Users only need to specify a POSTagFormat if they have a combination of models and need to convert between UD and Penn tag formats (to convert from a POSTaggerModel in one format to a lemmatizer or chunker model in the other format).

Currently, the models used for the unit tests all use the Penn tag format. Retraining the models using the UD format can be addressed as part of https://github.com/apache/lucene/issues/13002 (which I may work on next). To verify the downstream consumption of UD tags by another filter, I manually updated the lemmatizer dictionary (a non-binary model) to add UD tags.

Resolves https://github.com/apache/lucene/issues/14188

### DIFF
--- a/lucene/analysis/opennlp/src/java/org/apache/lucene/analysis/opennlp/OpenNLPPOSFilterFactory.java
+++ b/lucene/analysis/opennlp/src/java/org/apache/lucene/analysis/opennlp/OpenNLPPOSFilterFactory.java
@@ -18,7 +18,9 @@
 package org.apache.lucene.analysis.opennlp;
 
 import java.io.IOException;
+import java.util.Locale;
 import java.util.Map;
+import opennlp.tools.postag.POSTagFormat;
 import org.apache.lucene.analysis.TokenFilterFactory;
 import org.apache.lucene.analysis.TokenStream;
 import org.apache.lucene.analysis.opennlp.tools.OpenNLPOpsFactory;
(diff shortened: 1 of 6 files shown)


msfroh:
Test failure:

```
Reproduce with: gradlew :lucene:core:test --tests "org.apache.lucene.search.TestSeededKnnByteVectorQuery.testSeedWithTimeout" -Ptests.jvms=1 -Ptests.jvmargs= -Ptests.seed=533751C89ADE5499 -Ptests.useSecurityManager=true -Ptests.gui=true -Ptests.file.encoding=UTF-8 -Ptests.vectorsize=256 -Ptests.forceintegervectors=true
```

The test failure reliably reproduces with that seed on my Mac laptop.

msfroh:
> Test failure:
> 
> ```
> Reproduce with: gradlew :lucene:core:test --tests "org.apache.lucene.search.TestSeededKnnByteVectorQuery.testSeedWithTimeout" -Ptests.jvms=1 -Ptests.jvmargs= -Ptests.seed=533751C89ADE5499 -Ptests.useSecurityManager=true -Ptests.gui=true -Ptests.file.encoding=UTF-8 -Ptests.vectorsize=256 -Ptests.forceintegervectors=true
> ```
> 
> The test failure reliably reproduces with that seed on my Mac laptop.

The test was fixed. Everything is good after rebasing to latest 👍

epugh (commented):
The changes make sense, however some more documentation would be great on the UD and PENN formats...

epugh:
@cpoerschke (and anyone else) I haven't done a commit on Lucene in a long time so I want to get another set of eyes on this..    And I need to remember what all is in the workflow as well ;-)

PR #14198, opened 2025-02-05
Migrate OpenNLP 'ant train-test-models' to Gradle

### LINKED ISSUE #13002: migrate OpenNLP 'ant train-test-models' to Gradle

msfroh:
### Description

This resurrects the OpenNLP model training task from Ant (https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.11.2/lucene/analysis/opennlp/build.xml#L52-L84) to Gradle.

I ran the new `regenerate` task to regenerate the models with the new OpenNLP 2.5.3, which broke some of the chunking tests. I don't really understand the chunking output, so I just updated the test to reflect the new output. (The differences are pretty small.)

Edit: I checked the OpenNLP chunker docs and I think I understand the difference now. I think the older model wasn't properly the identifying verb group. So whereas the first sentence was previously a single long noun group:

```
Sentence number 1 has 6 words.
```

Now it's three groups (because the verb phrase gets its own group, which is consistent with the example in the OpenNLP manual):

```
Sentence number 1
has
6 words.
```

Resolves https://github.com/apache/lucene/issues/13002


lucene/analysis/opennlp/build.gradle
```diff
@@ -26,3 +26,33 @@ dependencies {
...
+def trainModel(String command, String lang, String data, String model, List extraArgs = []) {
+  javaexec {
+    classpath = sourceSets.main.compileClasspath
+    mainClass = 'opennlp.tools.cmdline.CLI'
+    workingDir = testModelDataDir
+    args = [command, '-lang', lang, '-data', data, '-model', "${testModelDir}/${model}"] + extraArgs
+  }
+}
+
+tasks.register('regenerate') { dependsOn 'trainTestModels' }
```
dweiss, line 58: Those regenerate tasks are often more complex than just dependencies - they have inputs and outputs to detect when the generation task can be skipped. Is the output from this opennlp always the same (deterministic)? If so, we could plug it into the regenerate pipeline - I can help, probably. But if it's not deterministic then I'd remove this regenerate line entirely and leave the task to be manually invoked when needed.

msfroh: Ahh... the LLM that helped me to migrate the Ant build to Gradle (after I spent an hour trying to do it by hand) copied the regenerate task from this line: https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.11.2/lucene/analysis/opennlp/build.xml#L117

I *think* the output is deterministic, but will change when a) OpenNLP is upgraded again or b) someone updates the training files.

Assuming neither of those happen very often (as has been the case recently), I think leaving it manual makes sense.

msfroh: Nope -- I checked by running the `trainTestModels` task again and the model files all had *something* change. So, I'm going to remove this line and we can stick to manual training as needed. 👍

dweiss: > the LLM that helped me to migrate the Ant build to Gradle (after I spent an hour trying to do it by hand)

It did a nice job. :) I'll take another look in the evening and we can merge, I think. Thank you.

dweiss:
So these generated model files are zip files and inside is a property file, which contains a timestamp... that's why they're different (this, plus zip compression may not be predictably the same, depending on the implementation)... 

So in theory, this could be rewritten so that there are no changes upon regeneration but for now I think it's already an improvement. 

Some things will not work with newer gradle but let's worry about it later.

PR #14201, opened 2025-02-05
Error prone back from the dead

### LINKED ISSUE #14200: improve the error prone situation
### Description

Currently this tool struggles to work with JDK21. maybe we can find a version that works or some workaround?
It is also slow, so slow that we only run it in the CI, which isn't great.

I spent a lot of time previously trying to limit the number of rules so that it isn't so slow, but when it barfs completely on NoSuchMethodError or whatever, there is no choice but to disable it.

Another idea that has worked, revisit what checks are available in the ecj (which is fast). Maybe it has some new ones that we can turn on that overlap with the functionality. Maybe we can use another fast syntax-based tool such as ast-grep to implement some of these checks.

Otherwise we have to fix it to play well with the JDK. This isn't a problem with JDK24 or anything, issues are being hit with JDK21.

rmuir:
Error prone is failing with NoSuchMethodError on JDK21 at least, due to their use of internal compiler apis.

Upgrading to 2.33.0 fixes the issue. Upgrading to 2.34+ is not possible as you will then encounter a different incompatibility with the JDK compiler.

But this gets error-prone working again.

Closes #14200


risdenk:
This looks fine. 

I do think there are new rules that could be disabled like [DefaultLocale](https://errorprone.info/bugpattern/DefaultLocale)

We were keeping a list

https://github.com/apache/lucene/blob/main/gradle/validation/error-prone.gradle#L73

I understand what a PITA it is to keep that up to date and disable the unneeded rules though.

I can take a stab at it tomorrow most likely if it would help.

risdenk (approved):

rmuir:
@risdenk yes that would absolutely be awesome and help us keep it all sorted

rmuir:
@risdenk thanks for looking. Will get this thing turned on again with hopefully less trouble than the previous version.

risdenk:
Took a little longer than expected but the followup is here - https://github.com/apache/lucene/pull/14216

PR #14176, opened 2025-01-28
Reduce virtual calls when visiting bpv24-encoded doc ids in BKD leaves

gf2121:
**UPDATE:**

This PR changed to only reduce virtual call of bpv24 encoded doc ids. Vectorized decoding optimization will be in follow-up PRs.

**Background**
* https://github.com/apache/lucene/pull/541 tried to introduce the vectorized decoding for BKD leaf blocks
* https://github.com/apache/lucene/pull/706 reverted the PR since we see a regression in nightly benchmark.
* https://github.com/apache/lucene/pull/797 introduced the bpv16 and bpv32 bulk decoding again. We find the regression mainly comes from bpv24 decoding because one pass loop is faster than simd decoding + loop (more detail can be found in [LUCENE-10315](https://issues.apache.org/jira/browse/LUCENE-10315))

**Proposal**
This PR tries to introduce the bpv24 vectorized decoding again and use the new bulk visit method to reduce virtual call, in favor of https://github.com/apache/lucene/pull/13149 and https://github.com/apache/lucene/pull/14138.

Luceneutil now can load 3 implementors of IntersectVisitor: RangeQuery Visitor, RangeQuery InverseVisitor and DynamicPruning Visitor. Here is the result on  `wikimediumall` and `taskCountPerCat=5`

```
                            TaskQPS baseline      StdDevQPS my_modified_version      StdDev                Pct diff p-value
               TermDayOfYearSort      259.87      (3.9%)      269.26      (4.2%)    3.6% (  -4% -   12%) 0.005
             CountFilteredIntNRQ       61.70      (7.1%)       85.00      (2.0%)   37.8% (  26% -   50%) 0.000
                      TermDTSort      149.65      (6.2%)      232.85      (9.6%)   55.6% (  37% -   76%) 0.000
                  FilteredIntNRQ       82.76     (10.0%)      135.48      (3.7%)   63.7% (  45% -   85%) 0.000
                          IntNRQ       84.62     (10.5%)      139.05      (2.6%)   64.3% (  46% -   86%) 0.000
```

Tasks
```
TermDayOfYearSort: dayofyeardvsort//0 # freq=708472
TermDayOfYearSort: dayofyeardvsort//names # freq=402762
TermDayOfYearSort: dayofyeardvsort//nbsp # freq=492778
TermDayOfYearSort: dayofyeardvsort//part # freq=588644
TermDayOfYearSort: dayofyeardvsort//st # freq=306811

[...]

PR #14203, opened 2025-02-06
Decode doc ids in BKD leaves with auto-vectorized loops

gf2121:
Context: https://github.com/apache/lucene/pull/14176

I find that when running with constant block size (512), JIT can auto-vectorize the decoding loop. But it does not work when block size become variable, which can be true in real BKD leaves. This PR proposes to use vector API to decode DocIds in BKD.

**MAC M2**
```
Benchmark                        (bpv)  (countVariable)   Mode  Cnt    Score   Error   Units
BKDCodecBenchmark.current           16             true  thrpt    5   85.316 ± 2.181  ops/ms
BKDCodecBenchmark.current           16            false  thrpt    5  208.971 ± 2.734  ops/ms
BKDCodecBenchmark.current           24             true  thrpt    5   85.752 ± 2.129  ops/ms
BKDCodecBenchmark.current           24            false  thrpt    5  147.652 ± 1.786  ops/ms
BKDCodecBenchmark.currentVector     16             true  thrpt    5  186.534 ± 2.376  ops/ms
BKDCodecBenchmark.currentVector     16            false  thrpt    5  213.891 ± 4.671  ops/ms
BKDCodecBenchmark.currentVector     24             true  thrpt    5  140.298 ± 2.189  ops/ms
BKDCodecBenchmark.currentVector     24            false  thrpt    5  134.398 ± 1.640  ops/ms
BKDCodecBenchmark.legacy            16             true  thrpt    5   87.278 ± 1.432  ops/ms
BKDCodecBenchmark.legacy            16            false  thrpt    5  201.612 ± 3.277  ops/ms
BKDCodecBenchmark.legacy            24             true  thrpt    5   87.148 ± 1.704  ops/ms
BKDCodecBenchmark.legacy            24            false  thrpt    5   84.830 ± 8.852  ops/ms
```

**Linux X86 (AVX512 supported)**
```
Benchmark                        (bpv)  (countVariable)   Mode  Cnt    Score    Error   Units
BKDCodecBenchmark.current           16             true  thrpt    5   27.711 ?  2.777  ops/ms
BKDCodecBenchmark.current           16            false  thrpt    5  132.859 ? 16.914  ops/ms
BKDCodecBenchmark.current           24             true  thrpt    5   34.672 ?  5.730  ops/ms
BKDCodecBenchmark.current           24            false  thrpt    5   33.017 ?  5.080  ops/ms
BKDCodecBenchmark.currentVector     16             true  thrpt    5   99.538 ? 11.813  ops/ms
BKDCodecBenchmark.currentVector     16            false  thrpt    5  107.525 ? 11.693  ops/ms
BKDCodecBenchmark.currentVector     24             true  thrpt    5   69.268 ? 10.351  ops/ms
BKDCodecBenchmark.currentVector     24            false  thrpt    5   64.134 ?  7.790  ops/ms
BKDCodecBenchmark.legacy            16             true  thrpt    5   27.531 ?  3.810  ops/ms
BKDCodecBenchmark.legacy            16            false  thrpt    5  125.707 ?  9.652  ops/ms
BKDCodecBenchmark.legacy            24             true  thrpt    5   22.528 ?  4.724  ops/ms
BKDCodecBenchmark.legacy            24            false  thrpt    5   23.903 ?  3.505  ops/ms
```


gf2121:
E2E result is disappointing:
```
                          IntSet      839.86      (3.1%)      835.78      (3.7%)   -0.5% (  -7% -    6%) 0.677
                      TermDTSort      203.80      (5.1%)      206.38      (5.0%)    1.3% (  -8% -   11%) 0.464
             CountFilteredIntNRQ       91.06      (2.3%)       92.48      (3.5%)    1.6% (  -4% -    7%) 0.123
               TermDayOfYearSort      203.77      (3.7%)      208.42      (3.7%)    2.3% (  -4% -   10%) 0.075
                          IntNRQ      148.29      (1.9%)      151.84      (3.2%)    2.4% (  -2% -    7%) 0.008
                  FilteredIntNRQ      146.00      (2.6%)      149.63      (3.2%)    2.5% (  -3% -    8%) 0.013
```

Profile suggests bottleneck is `FixedBitset#set` rather than decoding
image

gf2121:
On a AVX-512 Linux X86 machine:
```
                          IntSet      247.99      (4.1%)      244.08      (2.5%)   -1.6% (  -7% -    5%) 0.143
                      TermDTSort       82.84      (6.5%)       83.46      (8.8%)    0.8% ( -13% -   17%) 0.759
               TermDayOfYearSort       83.58      (4.6%)       85.12      (6.3%)    1.8% (  -8% -   13%) 0.290
             CountFilteredIntNRQ       38.61      (2.9%)       42.32      (2.6%)    9.6% (   4% -   15%) 0.000
                  FilteredIntNRQ       64.02      (3.6%)       75.48      (3.7%)   17.9% (  10% -   26%) 0.000
                          IntNRQ       66.28      (4.5%)       79.10      (2.9%)   19.3% (  11% -   28%) 0.000
```

jpountz:
Thanks for looking into it. Were you able to confirm that the difference with the variable count is indeed that auto-vectorization not getting enabled as opposed to something else such as different loop unrolling? I'm curious if you can compare the produced assembly and/or trick the JVM into generating more efficient code by writing the loop a bit differently, e.g. by having a fixed-size inner loop?

gf2121:
Thanks for feedback! I implement the fixed-size inner loop and print out assembly for all. [perf_asm.log](https://github.com/user-attachments/files/18752147/perf_asm.log)

* When profiling enabled, `#current countVariable=true` and `#current countVariable=false` has same (slow) speed. It seems like profiling prevented some optimization.

* According to the assembly, `#current bpv=16` does not get auto-vectorized. `#current bpv=24` gets vectorized on the shift loop, but not for the remainder loop.

* According to the assembly, the innerloop get auto-vectorized, but slower than vector API.

**MAC M2**
```
Benchmark                        (bpv)  (countVariable)   Mode  Cnt    Score    Error   Units
BKDCodecBenchmark.current           16             true  thrpt    5  103.490 ±  6.785  ops/ms
BKDCodecBenchmark.current           16            false  thrpt    5  212.488 ±  5.383  ops/ms
BKDCodecBenchmark.current           24             true  thrpt    5   91.203 ±  1.023  ops/ms
BKDCodecBenchmark.current           24            false  thrpt    5  149.742 ±  1.953  ops/ms
BKDCodecBenchmark.currentVector     16             true  thrpt    5  213.162 ±  1.598  ops/ms
BKDCodecBenchmark.currentVector     16            false  thrpt    5  216.529 ±  2.518  ops/ms
BKDCodecBenchmark.currentVector     24             true  thrpt    5  153.970 ±  1.101  ops/ms
BKDCodecBenchmark.currentVector     24            false  thrpt    5  140.103 ±  3.001  ops/ms
BKDCodecBenchmark.innerLoop         16             true  thrpt    5  129.281 ±  0.471  ops/ms
BKDCodecBenchmark.innerLoop         16            false  thrpt    5  131.083 ±  8.775  ops/ms
BKDCodecBenchmark.innerLoop         24             true  thrpt    5   99.597 ±  2.850  ops/ms
BKDCodecBenchmark.innerLoop         24            false  thrpt    5   96.235 ± 14.875  ops/ms
BKDCodecBenchmark.legacy            16             true  thrpt    5  104.314 ±  0.557  ops/ms
BKDCodecBenchmark.legacy            16            false  thrpt    5  202.175 ± 10.863  ops/ms
BKDCodecBenchmark.legacy            24             true  thrpt    5   86.016 ±  1.315  ops/ms
BKDCodecBenchmark.legacy            24            false  thrpt    5   85.609 ±  5.733  ops/ms
```

**Linux X86 AVX512 profiling disabled**
```
Benchmark                        (bpv)  (countVariable)   Mode  Cnt    Score    Error   Units
BKDCodecBenchmark.current           16             true  thrpt    5   41.138 ±  1.770  ops/ms
BKDCodecBenchmark.current           16            false  thrpt    5  142.277 ±  0.943  ops/ms
BKDCodecBenchmark.current           24             true  thrpt    5   43.104 ±  0.066  ops/ms
BKDCodecBenchmark.current           24            false  thrpt    5   42.760 ±  0.496  ops/ms
BKDCodecBenchmark.currentVector     16             true  thrpt    5   86.565 ±  0.904  ops/ms
BKDCodecBenchmark.currentVector     16            false  thrpt    5   86.624 ±  0.395  ops/ms
BKDCodecBenchmark.currentVector     24             true  thrpt    5   80.064 ±  2.604  ops/ms
BKDCodecBenchmark.currentVector     24            false  thrpt    5   76.638 ± 18.692  ops/ms
BKDCodecBenchmark.innerLoop         16             true  thrpt    5   43.810 ±  1.096  ops/ms
BKDCodecBenchmark.innerLoop         16            false  thrpt    5   42.485 ±  0.073  ops/ms
BKDCodecBenchmark.innerLoop         24             true  thrpt    5   37.255 ±  0.994  ops/ms
BKDCodecBenchmark.innerLoop         24            false  thrpt    5   37.243 ±  0.593  ops/ms
BKDCodecBenchmark.legacy            16             true  thrpt    5   41.415 ±  0.079  ops/ms
BKDCodecBenchmark.legacy            16            false  thrpt    5  145.713 ±  0.381  ops/ms
BKDCodecBenchmark.legacy            24             true  thrpt    5   27.758 ±  4.210  ops/ms
BKDCodecBenchmark.legacy            24            false  thrpt    5   28.519 ±  1.839  ops/ms
```

**Linux X86 AVX512 profiling enabled**
```
Benchmark                            (bpv)  (countVariable)   Mode  Cnt   Score   Error   Units
BKDCodecBenchmark.current               16             true  thrpt    5  29.878 ± 0.130  ops/ms
BKDCodecBenchmark.current:asm           16             true  thrpt          NaN             ---
BKDCodecBenchmark.current               16            false  thrpt    5  29.314 ± 0.229  ops/ms
BKDCodecBenchmark.current:asm           16            false  thrpt          NaN             ---
BKDCodecBenchmark.current               24             true  thrpt    5  34.874 ± 0.320  ops/ms
BKDCodecBenchmark.current:asm           24             true  thrpt          NaN             ---
BKDCodecBenchmark.current               24            false  thrpt    5  33.987 ± 0.055  ops/ms
BKDCodecBenchmark.current:asm           24            false  thrpt          NaN             ---
BKDCodecBenchmark.currentVector         16             true  thrpt    5  79.717 ± 5.983  ops/ms
BKDCodecBenchmark.currentVector:asm     16             true  thrpt          NaN             ---
BKDCodecBenchmark.currentVector         16            false  thrpt    5  81.924 ± 3.799  ops/ms
BKDCodecBenchmark.currentVector:asm     16            false  thrpt          NaN             ---
BKDCodecBenchmark.currentVector         24             true  thrpt    5  65.615 ± 8.901  ops/ms
BKDCodecBenchmark.currentVector:asm     24             true  thrpt          NaN             ---
BKDCodecBenchmark.currentVector         24            false  thrpt    5  74.759 ± 2.173  ops/ms
BKDCodecBenchmark.currentVector:asm     24            false  thrpt          NaN             ---
BKDCodecBenchmark.innerLoop             16             true  thrpt    5  40.869 ± 3.407  ops/ms
BKDCodecBenchmark.innerLoop:asm         16             true  thrpt          NaN             ---
BKDCodecBenchmark.innerLoop             16            false  thrpt    5  41.825 ± 1.644  ops/ms
BKDCodecBenchmark.innerLoop:asm         16            false  thrpt          NaN             ---
BKDCodecBenchmark.innerLoop             24             true  thrpt    5  37.251 ± 3.447  ops/ms
BKDCodecBenchmark.innerLoop:asm         24             true  thrpt          NaN             ---
BKDCodecBenchmark.innerLoop             24            false  thrpt    5  37.419 ± 1.238  ops/ms
BKDCodecBenchmark.innerLoop:asm         24            false  thrpt          NaN             ---
BKDCodecBenchmark.legacy                16             true  thrpt    5  28.477 ± 3.747  ops/ms
BKDCodecBenchmark.legacy:asm            16             true  thrpt          NaN             ---
BKDCodecBenchmark.legacy                16            false  thrpt    5  29.838 ± 0.163  ops/ms
BKDCodecBenchmark.legacy:asm            16            false  thrpt          NaN             ---
BKDCodecBenchmark.legacy                24             true  thrpt    5  28.295 ± 1.224  ops/ms
BKDCodecBenchmark.legacy:asm            24             true  thrpt          NaN             ---
BKDCodecBenchmark.legacy                24            false  thrpt    5  27.915 ± 0.911  ops/ms
BKDCodecBenchmark.legacy:asm            24            false  thrpt          NaN             ---
```

jpountz:
> #current bpv=24 gets vectorized on the shift loop, but not for the remainder loop.

This is an interesting observation. I wonder if a small refactoring could help it get auto-vectorized? E.g. what if we applied the `0xFF` mask to `scratch` in the shift loop rather than the remainder loop? Or if we split the remainder loop into 3 loops, one for each 8 bits that get contributed to the value?

Sorry for pushing, but if we could get auto-vectorization to do the right thing, then this would automatically benefit all users, not only those who enable the vector module.

gf2121:
> applied the 0xFF mask to scratch in the shift loop

This helps generate `vpand` in assembly, but not help performance too much.

> Sorry for pushing

Not at all, it's interesting & exciting to play with how loops get auto-vectorized!

>  if we could get auto-vectorization to do the right thing, then this would automatically benefit all users, not only those who enable the vector module.

+1

I try to minimize the code to see when loop will be vectorized. I write a simple `ShiftMaskBenchmark`(I add it to this patch temporarily to show code). It tells me that JIT can not vectorize loop when the offset is not predictable. This might explain why remainder decode / bpv16 decoding can not be vectorized.

asm: 
[perf_asm.log](https://github.com/user-attachments/files/18763358/perf_asm.log)

```
java -version

openjdk version "23.0.2" 2025-01-21
OpenJDK Runtime Environment (build 23.0.2+7-58)
OpenJDK 64-Bit Server VM (build 23.0.2+7-58, mixed mode, sharing)

Mac M2

Benchmark                      Mode  Cnt     Score     Error   Units
ShiftMaskBenchmark.fixOffset  thrpt    5  9131.716 ±  49.914  ops/ms
ShiftMaskBenchmark.varOffset  thrpt    5  3019.323 ± 120.261  ops/ms

Linux AVX512

Benchmark                          Mode  Cnt     Score    Error   Units
ShiftMaskBenchmark.fixOffset      thrpt    5  5021.907 ? 13.400  ops/ms
ShiftMaskBenchmark.fixOffset:asm  thrpt            NaN              ---
ShiftMaskBenchmark.varOffset      thrpt    5  1115.164 ?  0.958  ops/ms
ShiftMaskBenchmark.varOffset:asm  thrpt            NaN              ---
```

InnerLoop seems like a good option if the conclusion is true. I'd try to see if i can improve it a bit.

gf2121:
I refactor code to inner-loop. Result on wikimediumall AVX512

```
                      TermDTSort       84.41      (4.4%)       82.84      (7.4%)   -1.9% ( -13% -   10%) 0.337
               TermDayOfYearSort       83.32      (4.6%)       84.33      (5.8%)    1.2% (  -8% -   12%) 0.463
                          IntSet      246.30      (4.1%)      254.93      (4.0%)    3.5% (  -4% -   12%) 0.006
             CountFilteredIntNRQ       38.18      (3.5%)       42.08      (3.4%)   10.2% (   3% -   17%) 0.000
                          IntNRQ       66.06      (4.9%)       81.60      (3.6%)   23.5% (  14% -   33%) 0.000
                  FilteredIntNRQ       63.17      (4.4%)       79.11      (3.4%)   25.2% (  16% -   34%) 0.000
```

jpountz:
These results look even better than the results that you had previously reported for the vector API, is my understanding correct that it performs even better?

gf2121:
>  is my understanding correct that it performs even better?

Yeah!

gf2121:
Comparison of VectorAPI(Baseline) and InnerLoop(Candidate)

```
                      TermDTSort       86.75      (4.5%)       85.67      (4.7%)   -1.2% ( -10% -    8%) 0.398
               TermDayOfYearSort       84.66      (4.8%)       84.50      (7.1%)   -0.2% ( -11% -   12%) 0.924
                          IntSet      249.81      (3.7%)      253.82      (2.3%)    1.6% (  -4% -    7%) 0.100
             CountFilteredIntNRQ       40.98      (1.9%)       42.26      (3.0%)    3.1% (  -1% -    8%) 0.000
                          IntNRQ       77.45      (3.0%)       81.24      (3.3%)    4.9% (  -1% -   11%) 0.000
                  FilteredIntNRQ       73.63      (2.5%)       77.89      (3.1%)    5.8% (   0% -   11%) 0.000
```

jpountz:
Thanks for iterating and running benchmarks. I played with the micro-benchmark and I get almost the same result if I use a single batch size of 16 of 32 (AMD Ryzen with AVX2 but no AVX-512), which I like a bit better since it's simpler (single batch size) and should better work on lower numbers of entries, e.g. if we decide to decrease the BKD leaf size in the future. Do you get similar results?

gf2121:
Thanks for feedback! 

And sorry for my poor english.. Do you mean something like this by `single batch size of 16 of 32` ?
```
private static void readDelta16(IndexInput in, int count, int[] docIds) throws IOException {
  final int min = in.readVInt();
  int k = 0;
  for (int bound = count - 31; k < bound; k += 32) {
    in.readInts(docIds, k, 16);
    // Can be inlined to make offsets consistent so that loop get auto-vectorized.
    inner16(k, docIds, 16, min);
  }
  for (; k < count; k++) {
    docIds[k] = Short.toUnsignedInt(in.readShort()) + min;
  }
}

private void readInts24(IndexInput in, int count, int[] docIDs) throws IOException {
  int k = 0;
  for (int bound = count - 31; k < bound; k += 32) {
    in.readInts(scratch, k, 24);
    shift(k, docIDs, scratch, 24);
    // Can be inlined to make offsets consistent so that loop get auto-vectorized.
    remainder24(k, docIDs, scratch, 8, 16, 24);
  }
  readScalarInts24(in, count - k, docIDs, k);
}
```

jpountz:
Yes, exactly.

gf2121:
Results on my machines are a bit disappointing
```
java -version
openjdk version "23.0.2" 2025-01-21
OpenJDK Runtime Environment (build 23.0.2+7-58)
OpenJDK 64-Bit Server VM (build 23.0.2+7-58, mixed mode, sharing)

Mac M2

Benchmark                 (bpv)   Mode  Cnt   Score    Error   Units
BKDCodecBenchmark.scalar     16  thrpt    5  97.169 ±  0.454  ops/ms
BKDCodecBenchmark.scalar     24  thrpt    5  86.957 ±  0.929  ops/ms
BKDCodecBenchmark.vector     16  thrpt    5  79.996 ± 17.609  ops/ms
BKDCodecBenchmark.vector     24  thrpt    5  35.612 ±  0.346  ops/ms

Linux AVX512

Benchmark                 (bpv)   Mode  Cnt   Score   Error   Units
BKDCodecBenchmark.scalar     16  thrpt    5  32.064 ? 2.751  ops/ms
BKDCodecBenchmark.scalar     24  thrpt    5  27.612 ? 0.931  ops/ms
BKDCodecBenchmark.vector     16  thrpt    5  17.617 ? 1.296  ops/ms
BKDCodecBenchmark.vector     24  thrpt    5  10.724 ? 0.176  ops/ms
```

jpountz:
Thanks for running benchmarks. I'm confused as to why running inner loops of size 512 would be to much better than inner loops of size 32. This doesn't feel right? Does luceneutil also report worse performance with the inner loop of size 32?

gf2121:
Confused +1 ... but the comparison of step512(baseline) and step32(candidate): 

```
                          IntNRQ       77.37      (4.1%)       54.55      (3.4%)  -29.5% ( -35% -  -22%) 0.000
                  FilteredIntNRQ       73.44      (4.0%)       52.70      (2.2%)  -28.2% ( -33% -  -22%) 0.000
             CountFilteredIntNRQ       40.78      (3.4%)       33.68      (2.4%)  -17.4% ( -22% -  -12%) 0.000
                          IntSet      242.01      (4.1%)      224.59      (3.9%)   -7.2% ( -14% -    0%) 0.000
```

gf2121:
> There must be something that happens with this 512 step that doesn't happen otherwise such as using different instructions, loop unrolling, better CPU pipelining or something else.

Thanks for pointing out this. I studied the asm profile again and i can see at least loop unrolling differs there. According to the asm printed by jmh, i can see for bpv24 decoding:

* VectorAPI unrolled shift loop x8 (add 0x40 once) and remainder loop x4 (add 0x20 once) 
* InnerLoop 512 step unrolled shift loop x4 (add 0x20 once) and remainder loop x2 (add 0x10 once) 
* InnerLoop 128 step does not get loop unrolling for either shift loop (add 0x8 once) or remainder loop (add 0x8 once).

This is corresponding to the result of jmh: vector API > InnerLoop step-512 > InnerLoop step-128. 

Things might change in luceneutil because we find InnerLoop step-512 faster than Vector API there. I confirmed the result of  luceneutil of step-512(baseline) vs step-128(candidate):

```
                  FilteredIntNRQ       80.02      (4.0%)       71.31      (3.0%)  -10.9% ( -17% -   -4%) 0.000
                          IntNRQ       80.94      (2.5%)       72.60      (3.6%)  -10.3% ( -16% -   -4%) 0.000
             CountFilteredIntNRQ       42.93      (2.9%)       40.22      (2.3%)   -6.3% ( -11% -   -1%) 0.001
                          IntSet       93.36      (2.1%)       93.85      (0.7%)    0.5% (  -2% -    3%) 0.633
```

jpountz:
Thanks for running benchmarks. So it looks like the JVM doesn't think these shorter loops (with step 128) are worth unrolling? This makes me wonder how something like that  performs on your AVX-512 CPU. I think you had something similar in one of your previous iterations. On my machine it's on par with the current version.

```java
  private void readInts24(IndexInput in, int count, int[] docIDs) throws IOException {
    if (count == BKDConfig.DEFAULT_MAX_POINTS_IN_LEAF_NODE) {
      // Same format, but enabling the JVM to specialize the decoding logic for the default number
      // of points per node proved to help on benchmarks
      doReadInts24(in, 512, docIDs);
    } else {
      doReadInts24(in, count, docIDs);
    }
  }

  private void doReadInts24(IndexInput in, int count, int[] docIDs) throws IOException {
    // Read the first (count - count % 4) values
    int quarter = count >> 2;
    int numBytes = quarter * 3;
    in.readInts(scratch, 0, numBytes);
    for (int i = 0; i < numBytes; ++i) {
      docIDs[i] = scratch[i] >>> 8;
      scratch[i] &= 0xFF;
    }
    for (int i = 0; i < quarter; ++i) {
      docIDs[numBytes + i] = scratch[i]
         | (scratch[quarter + i] << 8)
         | (scratch[2 * quarter + i] << 16);
    }
    // Now read the remaining 0, 1, 2 or 3 values
    for (int i = quarter << 2; i < count; ++i) {
      docIDs[i] = (in.readShort() & 0xFFFF) | (in.readByte() & 0xFF) << 16;
    }
  }
```

gf2121:
On the AVX-512 machine:

* Specialized read does not vectorize the remainder loop, it seems the complier failed to inline it.
* Specialized decode vectorizes the remainder loop.
* Push masks to remainder loop seems to get a better performance.

I pushed the benchmark code to the patch, here is result on my machine:

```
Benchmark                                                         Mode  Cnt   Score   Error   Units
InnerLoopDecodingBenchmark.hybridInnerLoop                       thrpt    5  76.311 ? 0.177  ops/ms
InnerLoopDecodingBenchmark.hybridInnerLoop:asm                   thrpt          NaN             ---
InnerLoopDecodingBenchmark.specializedDecode                     thrpt    5  73.600 ? 0.123  ops/ms
InnerLoopDecodingBenchmark.specializedDecode:asm                 thrpt          NaN             ---
InnerLoopDecodingBenchmark.specializedDecodeMaskInRemainder      thrpt    5  80.902 ? 0.046  ops/ms
InnerLoopDecodingBenchmark.specializedDecodeMaskInRemainder:asm  thrpt          NaN             ---
InnerLoopDecodingBenchmark.specializedRead                       thrpt    5  37.195 ? 0.099  ops/ms
InnerLoopDecodingBenchmark.specializedRead:asm                   thrpt          NaN             ---
```

**LuceneUtil**

hybridInnerLoop (baseline) vs specializedRead (candidate)
```
                          IntNRQ       83.50      (3.7%)       78.07      (4.1%)   -6.5% ( -13% -    1%) 0.000
                  FilteredIntNRQ       81.66      (3.6%)       76.50      (3.8%)   -6.3% ( -13% -    1%) 0.000
             CountFilteredIntNRQ       44.30      (1.9%)       43.07      (2.8%)   -2.8% (  -7% -    1%) 0.000
                          IntSet       94.41      (1.7%)       94.38      (1.0%)   -0.0% (  -2% -    2%) 0.940
```

hybridInnerLoop (baseline) vs specializedDecodeMaskInRemainder (candidate)
```
                          IntSet       96.10      (1.1%)       94.85      (1.5%)   -1.3% (  -3% -    1%) 0.039
             CountFilteredIntNRQ       43.67      (2.3%)       44.78      (2.8%)    2.6% (  -2% -    7%) 0.036
                          IntNRQ       81.12      (2.9%)       84.52      (4.1%)    4.2% (  -2% -   11%) 0.013
                  FilteredIntNRQ       78.94      (3.0%)       83.07      (4.1%)    5.2% (  -1% -   12%) 0.002
```

jpountz:
Again, thanks a lot for running benchmarks.

> I can refactor the code to the specialized decoding if it makes sense to you

That would be great, thank you. Sorry for making it hard for you to move this PR forward, I was a bit annoyed that we needed something complicated to speed things up, I like the simplicity of specializedDecodeMaskInRemainder.

> BTW, should we disable the flexibility of changing maxPointsInLeafNode if we make optimization only work on default config?

In my opinion, it's good enough the way things are today as the default codec doesn't allow configuring the number of points per leaf node.

gf2121:
> Sorry for making it hard for you to move this PR forward, I was a bit annoyed that we needed something complicated to speed things up, I like the simplicity of specializedDecodeMaskInRemainder.

No apologies needed, it was exciting to play with the asm code, vectorize loops and make the code simpler! I learned so much through these iterations :)

> In my opinion, it's good enough the way things are today as the default codec doesn't allow configuring the number of points per leaf node.

LGTM. IMO we should bind OpenStreet benchmark to DEFAULT_MAX_POINTS_IN_LEAF_NODE after this get merged.
https://github.com/mikemccand/luceneutil/blob/9724b69c9ef53715dae27baa3493a0ee7949748e/src/main/perf/IndexAndSearchOpenStreetMaps.java#L542

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90PointsWriter.java
```diff
@@ -105,15 +107,22 @@ public Lucene90PointsWriter(
     }
   }
 
+  public Lucene90PointsWriter(
+      SegmentWriteState writeState, int maxPointsInLeafNode, double maxMBSortInHeap)
+      throws IOException {
+    this(writeState, maxPointsInLeafNode, maxMBSortInHeap, Lucene90PointsFormat.VERSION_CURRENT);
+  }
```
jpountz, line 114: let's make all constructors that take a version pkg-private?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
```diff
@@ -139,6 +141,28 @@ public BKDWriter(
...
+        maxDoc,
+        tempDir,
+        tempFileNamePrefix,
+        config,
+        maxMBSortInHeap,
+        totalPointCount,
+        BKDWriter.VERSION_CURRENT);
+  }
+
+  public BKDWriter(
```
jpountz, line 154: Can you add javadocs that this ctor should be only used for testing with older versions?

lucene/core/src/java/org/apache/lucene/util/bkd/DocIdsWriter.java
```diff
@@ -114,31 +117,28 @@ void writeDocIds(int[] docIds, int start, int count, DataOutput out) throws IOEx
...
+          final int quarter = count >> 2;
+          final int numInts = quarter * 3;
+          for (int i = 0; i < numInts; i++) {
+            scratch[i] = docIds[i + start] << 8;
+          }
+          for (int i = 0; i < quarter; i++) {
+            final int longIdx = i + numInts + start;
+            scratch[i] |= docIds[longIdx] >>> 16;
+            scratch[i + quarter] |= (docIds[longIdx] >>> 8) & 0xFF;
+            scratch[i + quarter * 2] |= docIds[longIdx] & 0xFF;
```
jpountz, line 133: nit: maybe write bytes in little-endian order for consistency, so `scratch[i] |= docIds[longIdx] & 0xFF; scratch[i + quarter] |= (docIds[longIdx] >>> 8) & 0xFF; scratch[i + quarter * 2] |= docIds[longIdx] >>> 16;`, etc. ?

lucene/core/src/java/org/apache/lucene/util/bkd/DocIdsWriter.java
```diff
@@ -248,21 +281,68 @@ private void readBitSet(IndexInput in, int count, int[] docIDs) throws IOExcepti
...
+          BKDConfig.DEFAULT_MAX_POINTS_IN_LEAF_NODE / 4,
+          BKDConfig.DEFAULT_MAX_POINTS_IN_LEAF_NODE / 4 * 3);
+      assert quarter * 4 == BKDConfig.DEFAULT_MAX_POINTS_IN_LEAF_NODE
+          : " we are assuming DEFAULT_MAX_POINTS_IN_LEAF_NODE is a multiple of 4 here.";
+    } else {
+      decode24(docIDs, scratch, quarter, numInts);
+      // Now read the remaining 0, 1, 2 or 3 values
+      for (int i = quarter << 2; i < count; ++i) {
+        docIDs[i] = (in.readShort() & 0xFFFF) | (in.readByte() & 0xFF) << 16;
+      }
```
jpountz, line 329: out of curiosity, does it hurt performance if we add this as part of decode24? This would help save the above assertion?

jpountz (commented):
It looks good in general, just left minor comments. Thank you!

lucene/core/src/java/org/apache/lucene/util/bkd/DocIdsWriter.java
gf2121: I want to keep decode24 small so i put it under the `if else` block to save the assertion. luceneutil and jmh proved it has similar performance.

jpountz (approved):
Looks great!

gf2121:
Nightly benchmark confirmed the speed up https://benchmarks.mikemccandless.com/2025.03.16.18.04.58.html.

Thanks again for profile guide and helping figure out simpler and faster codes!

gf2121:
I raised an PR for annotation. https://github.com/mikemccand/luceneutil/pull/354.

jpountz:
Fantastic speedup. Nice to see tasks like `TermDayOfYearSort` also take advantage from this change.

PR #14205, opened 2025-02-06
Automatically remove dead states from concatenated automata.

jpountz:
Concatenating automata frequently creates dead states. This PR suggests that `Operations#concatenate` automatically removes these dead states. This is not unseen: `Operations#repeat`, `Operations#union` and other automaton manipulation utilities also automatically remove dead states.

Here is an example of what concatenating automata that accept "f" and "g" returns, without this patch:
automaton

With this patch, state 3 gets removed.

Relates #14193.


rmuir:
It's in my PR over there too. I think we should avoid addEpsilon in the test. We don't even need any transitions.

jpountz:
Superseded by #14193

PR #14207, opened 2025-02-06
Make Operations#union merge accept states that have no outgoing transition.

jpountz:
This helps generate simpler automata, especially when these automata are later combined through other operations such as `Operations#concat`.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/Operations.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/Operations.java
@@ -508,7 +508,7 @@ public final class Operations {
 
     result.finishState();
 
-    return removeDeadStates(result);
+    return mergeAcceptStatesWithNoTransition(removeDeadStates(result));
   }
 
   // Simple custom ArrayList<Transition>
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/util/automaton/Operations.java
```diff
@@ -1052,6 +1052,77 @@ public static Automaton removeDeadStates(Automaton a) {
     return result;
   }
 
+  /** Merge all accept states that don't have outgoing transitions to a single shared state. */
```
rmuir, line 1055: I like the idea, couldn't we potentially use this for other operations such as reverse() which might cause the same inefficiency as union() today?

Can we hit it with a simple random test for good feelings about it?
```java
for (int i = 0; i < N; i++) {
  Automaton expected = AutomatonTestUtil.randomAutomata(random);
  Automaton actual = mergeAcceptStates(expected);
  assertSameLanguage(expected, actual);
}
```

rmuir (approved):
Thanks for tackling this!

PR #14209, opened 2025-02-06
Deprecate Operations.concat(a1, a2) and Operations.union(a1, a2)

### LINKED ISSUE #14202: Deprecate Operations.concatenate(a1, a2) and Operations.union(a1, a2)
### Description

These automata operations have two forms today, using concatenate() as an example:

```java
  /**
   * Returns an automaton that accepts the concatenation of the languages of the given automata.
   *
   * <p>Complexity: linear in total number of states.
   */
  public static Automaton concatenate(Automaton a1, Automaton a2) {
    return concatenate(Arrays.asList(a1, a2));
  }

  /**
   * Returns an automaton that accepts the concatenation of the languages of the given automata.
   *
   * <p>Complexity: linear in total number of states.
   */
  public static Automaton concatenate(List<Automaton> l) {
    ...
```

I think these helpers are trappy, in that it makes it too easy to use Automaton like a StringBuilder:
```java
while (someloop) {
  automaton = Operations.concatenate(automaton, other);
}
```
But if you do it this way, it does not run in linear time.

So I'd propose to just deprecate the `concatenate(a1, a2)` in favor of `concatenate(List.of(a1, a2))` and the same thing with the `union()` if you really just want to concatenate/union two things, it isn't causing you too much additional pain. If you want to concatenate/union 3 or more things, it encourages that you do it in the performant way.

rmuir:
These algorithms run in linear time: it is trappy to offer two-arg options: they encourage users to use them in a loop and create quadratic time.

Send a strong signal to the user's editor/IDE to avoid trouble in the first place:
* Deprecate two-arg methods in favor of the List-based versions
* Rename parameter 'l' to 'list'
* Add javadoc `@param` to parameter 'list'

Closes #14202

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -46,6 +46,9 @@ API Changes
 * GITHUB#14134: Added Bits#applyMask API to help apply live docs as a mask on a
   bit set of matches. (Adrien Grand)
 
+* GITHUB#14209: Deprecate Operations.union(Automaton,Automaton) and 
+  concatenate(Automaton,Automaton) in favor of the methods taking List.  (Robert Muir)
+
 New Features
 ---------------------
 
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/Operations.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/Operations.java
@@ -81,6 +81,7 @@ public final class Operations {
    * Returns an automaton that accepts the concatenation of the languages of the given automata.
    *
    * <p>Complexity: linear in total number of states.
+   *
    * @param list List of automata to be joined
    */
   public static Automaton concatenate(List<Automaton> list) {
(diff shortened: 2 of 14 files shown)


mikemccand (approved):
Yay.  Thanks @rmuir.

rmuir:
I will look into the test fail, I did run them many times, but I was also pretty aggressive about trying to cleanup tests, so that we can just remove the deprecations in a followup commit, and improve efficiency in some places too.

Code in the tests and suggester were impacted in this way.

rmuir:
The problem is not related to this PR from what I can tell, the issue is that `concatenate()` turns an NFA into a DFA and this fails the test.

I can reproduce it in main: I will deal with it separately.

rmuir:
https://github.com/apache/lucene/issues/14210

jpountz (approved):
This looks good to me, though I wonder if we really need the deprecation dance as there are internal APIs (expert at best), so we could proceed with the removal directly?

rmuir:
To me the deprecation is easy enough, developer is usually responsive to such things. 

It comes across different than just a hard break in a few ways, usually you question why the previous thing was deprecated and it causes you to take a second look. Versus just trying to make the code compile again. I'd rather us give the helpful nudge to make the code more efficient?

PR #12581, opened 2023-09-21
Allow reading / writing binary stored fields as DataInput

iverase:
Currently, the only way to handle binary data on stored fields is via byte arrays (wrapped as BytesRef). THis means we are allocating a new byte array everytime we read the value which is wasteful and it can be problematic as those arrays can be humongous and add pressure to the GC. 

In this PR we proposed to add the possibility to read / write binary stored values using a DataInput and the number of bytes. By default the implementations will allocate those bytes in a newly created byte array and call the already existing method. 

This should speed up the merging of stored fields as we are not using an intermediate data structure any longer and allow implementoirs to read the binary fields without having to allocate a byte array.

**EDIT**: merges may be faster if merge strategy is `MergeStrategy.VISITOR`

closes https://github.com/apache/lucene/issues/12556

PR #14213, opened 2025-02-06
Support DataInput as source for StoredField

Tim-Brooks:
Allowing indexing stored-only StoredField directly from DataInput.
Introduces StoredFieldDataInput record to associate a length with
DataInput.


Tim-Brooks:
I am opening this proposed change to support writing a stored field from a byte source which does not require a contiguous array allocation. The reason I am proposing this is because there are times when we would like to store large stored fields and the requirement to provide a fully contiguous byte array can cause issues on smaller heaps. Particularly when the original data is already on heap in a non-contiguous source.

I took a stab at this using a `DataInput` as a source for the indexing. I went with this initial approach as it aligns with the fact that `StoredFieldsWriter` already supports `DataInput` (seemingly for merges).

I wrapped the `DataInput` in a record style class `StoredFieldDataInput` to associated `length` with it.

If this approach has support I will continue to refine the PR. In particular, I was uncertain whether Lucene would want `DataInput` to be fully support in `Field` similar to `stringValue`, `readerValue`, `doubleValue`, etc, etc (with getters and setters). Or stick with what I did where it is only really supported in `StoredField` as a `storedValue`. Also I would be interested in what additional test classes I should modify with this type of ensure coverage.

Finally, would we want to modify `StoredFieldsWriter#writeField(FieldInfo info, DataInput value, int length)` to use this `StoredFieldDataInput` abstraction instead of also having the length int (if others support the introduction of this abstraction)?

DataInput is only one potential approach. I took it because there was already some work around `DataInput` with stored fields.

A `ByteRef[]` or `ByteBuffer[]` would also work for our use. `DataInput` has the downside of requiring a local intermediate buffer in `ByteBuffersDataOutput` to copy into direct bytes. `BytesRef[]` would work but then not allow direct memory as a source (doesn't matter to our use case but worth noting). `ByteBuffer[]` supports everything (direct, no intermediate buffer) but is theoretically a bit less flexible than `DataInput` which is a very flexible abstraction.

Any of these approaches are fine for my use case and I would be happy to work on whichever has the most support and consensus.

jpountz:
> I went with this initial approach as it aligns with the fact that StoredFieldsWriter already supports DataInput (seemingly for merges).

Indeed, the change that introduced this capability had a similar motivation to yours: #12581.

> I wrapped the DataInput in a record style class StoredFieldDataInput to associated length with it.

I'm curious if we should make it an actual record?

> I was uncertain whether Lucene would want DataInput to be fully supported in Field similar to stringValue, readerValue, doubleValue, etc,

I like what you did better. I don't want to have to deal with the case when a value is provided through a `DataInput` and needs to be analyzed / indexed.

> Finally, would we want to modify StoredFieldsWriter#writeField(FieldInfo info, DataInput value, int length) to use this StoredFieldDataInput abstraction

It's slightly awkward to have an argument called "value" that doesn't fully encapsulate all the information that is necessary to know what the actual value is, so I like your proposal to modify it.

> DataInput is only one potential approach. I took it because there was already some work around DataInput with stored fields.

This looks like the best approach to me, I like the symmetry with the read API in `StoredFieldVisitor`.

In general the change makes sense to me, I'm curious to get @iverase 's opinion since he's worked on the same change on the read side (`StoredFieldsVisitor`). It is also consistent with the fact that analyzed fields support passing a `java.io.Reader` instead of a fully materialized `String`.

Tim-Brooks:
> I'm curious if we should make it an actual record?

Haha probably. I actually did not check Lucene's language level when producing the PR.

I'll continue to refine this a bit more based on the premise that `DataInput` is the preferred abstraction while we wait for more feedback.

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene50/compressing/Lucene50CompressingStoredFieldsReader.java
```diff
@@ -25,12 +25,7 @@
 import org.apache.lucene.codecs.CodecUtil;
 import org.apache.lucene.codecs.StoredFieldsReader;
 import org.apache.lucene.document.Document;
-import org.apache.lucene.index.CorruptIndexException;
-import org.apache.lucene.index.FieldInfo;
-import org.apache.lucene.index.FieldInfos;
-import org.apache.lucene.index.IndexFileNames;
-import org.apache.lucene.index.SegmentInfo;
-import org.apache.lucene.index.StoredFieldVisitor;
+import org.apache.lucene.index.*;
```
iverase, line 28: your IDE is misbehaving :)

iverase:
+1 I like the idea of encapsulating the DataInput and length inside `StoredFieldDataInput`, and yes, it gives symmetry between the read and the write side.

lucene/core/src/java/org/apache/lucene/index/StoredFieldVisitor.java
```diff
@@ -41,15 +40,17 @@ public abstract class StoredFieldVisitor {
   protected StoredFieldVisitor() {}
 
   /**
-   * Expert: Process a binary field directly from the {@link DataInput}. Implementors of this method
-   * must read {@code length} bytes from the given {@link DataInput}. The default implementation
-   * reads all byes in a newly created byte array and calls {@link #binaryField(FieldInfo, byte[])}.
+   * Expert: Process a binary field directly from the {@link StoredFieldDataInput}. Implementors of
+   * this method must read {@code StoredFieldDataInput#length} bytes from the given {@link
+   * StoredFieldDataInput}. The default implementation reads all byes in a newly created byte array
```
iverase, line 45: it is an existing typo.

```suggestion
   * StoredFieldDataInput}. The default implementation reads all bytes in a newly created byte array
```

iverase (approved):
LGTM

iverase:
@Tim-Brooks Could you add an entry in CHANGES.txt? It should be under the 10.2 version, thanks!

Tim-Brooks:
> @Tim-Brooks Could you add an entry in CHANGES.txt? It should be under the 10.2 version, thanks!

I made this change and added some more tests. Let me know if any additional tests make sense.

iverase:
Thanks @Tim-Brooks, tests look good.

I will be merging this soon.

iverase:
Thank you @Tim-Brooks !

Issue #14214, opened 2025-02-07
HNSW connect components can take an inordinate amount of time


benwtrent:
### Description

Connect components on Flush or merge, while good for graphs that are "almost OK" but need to be better connected, can just destroy performance if the vector distribution is poor. 

I don't readily have test data, but if you have tightly clustered, or many duplicate vectors, it can take until the "heat death of the universe" to complete. 

It seems to me that since connect Components is a "best effort fix up" of the graph, we should add a "cap" on the amount of work this does.

Marking as a bug as I have seen this for real users and it effectively takes a CPU hostage for hours (maybe days).

tteofili:
this can be reproduced with either of the following tests

```java
public void testSameVectorIndexedMultipleTimes() throws IOException {
    try (Directory d = newDirectory()) {
      try (IndexWriter w = new IndexWriter(d, new IndexWriterConfig())) {
        for (int j = 1; j <= 1_000_000; j++) {
          Document doc = new Document();
          doc.add(
                  getKnnVectorField(
                          "field", randomVector(16), DOT_PRODUCT));
          w.addDocument(doc);
          if (j % 100 == 0) {
            w.flush();
          }
        }
        w.commit();
      }

    }
  }

  public void testFewDistinctVectors() throws IOException {
    try (Directory d = newDirectory()) {
      try (IndexWriter w = new IndexWriter(d, new IndexWriterConfig())) {
        float[][] f = new float[16][];
        for (int i = 0; i < 16; i++) {
          f[i] = new float[16];
          Arrays.fill(f[i], (float) i);
        }
        for (int i =0; i < 1_000_000; i++) {
          Document doc = new Document();
          doc.add(getKnnVectorField("field", f[random().nextInt(16)], DOT_PRODUCT));
          w.addDocument(doc);
        }
      }
    }
  }
```

they take tens of minutes and thread dump shows thread stuck at `hnsw.HnswGraphBuilder.connectComponents`

whereas a similar test even with random vectors takes 10x less time to finish.

Vikasht34:
```
private void connectComponents() {
    BitSet visited = new BitSet(graph.size());
    List<Integer> entryPoints = new ArrayList<>();
    
    for (int node = 0; node < graph.size(); node++) {
        if (!visited.get(node)) {
            List<Integer> component = new ArrayList<>();
            exploreComponent(node, visited, component);

            if (!entryPoints.isEmpty()) {
                int bestCandidate = findBestCandidate(entryPoints, component);
[...]

PR #12770, opened 2023-11-06
Stop exploring HNSW graph if scores are not getting better.

benwtrent:
I noticed while testing lower dimensionality and quantization, we would explore the HNSW graph way too much. I was stuck figuring out why until I noticed the searcher checks for distance equality (not just if the distance is better) when exploring neighbors-of-neighbors. This seems like a bad heurstic, but to double check I looked at what nmslib does. This pointed me back to this commit: https://github.com/nmslib/nmslib/pull/106

Seems like this performance hitch was discovered awhile ago :). 

This commit adjusts HNSW to only explore the graph layer if the distance is actually better.

PR #14215, opened 2025-02-07
Bugfix/fix hnsw search termination check

benwtrent:
previously related PR: https://github.com/apache/lucene/pull/12770

While my original change to help move us towards a saner HNSW search behavior, it is will still actually explore a candidate if its score is `==` min accepted. This will devolve in the degenerate case where all vectors are the same.

Here are some test runs. One test indexes the same vector many times. The other indexes the same 16 vectors many times. 

There isn't much difference with the "few unique vectors" case from what I can tell. However, the super degenerate case where all scores are exactly the same, this is magnitudes faster.

Logically, it makes sense to make the condition to skip a candidate the exact same for adding a candidate.

Also note that this degenerate case with uniform vector scores got WAY worse with the connected components change.

[Archive.zip](https://github.com/user-attachments/files/18710710/Archive.zip)

related to (but doesn't fully solve): https://github.com/apache/lucene/issues/14214

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java
+++ b/lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java
@@ -52,7 +52,7 @@ public class TopKnnCollector extends AbstractKnnCollector {
 
   @Override
   public float minCompetitiveSimilarity() {
-    return queue.size() >= k() ? Math.nextUp(queue.topScore()) : Float.NEGATIVE_INFINITY;
+    return queue.size() >= k() ? queue.topScore() : Float.NEGATIVE_INFINITY;
   }
 
   @Override
--- a/lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java
+++ b/lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java
@@ -52,7 +52,7 @@ public class TopKnnCollector extends AbstractKnnCollector {
 
   @Override
   public float minCompetitiveSimilarity() {
-    return queue.size() >= k() ? queue.topScore() : Float.NEGATIVE_INFINITY;
+    return queue.size() >= k() ? Math.nextUp(queue.topScore()) : Float.NEGATIVE_INFINITY;
   }
 
   @Override
--- a/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java
+++ b/lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java
@@ -212,7 +212,7 @@ public class HnswGraphSearcher extends AbstractHnswGraphSearcher {
 
     // A bound that holds the minimum similarity to the query vector that a candidate vector must
     // have to be considered.
-    float minAcceptedSimilarity = results.minCompetitiveSimilarity();
+    float minAcceptedSimilarity = Math.nextUp(results.minCompetitiveSimilarity());
     while (candidates.size() > 0 && results.earlyTerminated() == false) {
       // get the best candidate (closest or best scoring)
       float topCandidateSimilarity = candidates.topScore();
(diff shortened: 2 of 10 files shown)


jimczi (approved):
That's a nasty performance bug, great catch @benwtrent

benwtrent:
@iverase is the one who found it :D

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java
```diff
@@ -216,7 +216,7 @@ void searchLevel(
     while (candidates.size() > 0 && results.earlyTerminated() == false) {
       // get the best candidate (closest or best scoring)
       float topCandidateSimilarity = candidates.topScore();
-      if (topCandidateSimilarity < minAcceptedSimilarity) {
+      if (minAcceptedSimilarity > topCandidateSimilarity) {
```
jpountz, line 219: I am confused, aren't these conditions exactly the same?

shubhamvishu: ++ I think this was supposed to be `>=`

benwtrent: They are not. Previously, the candidate would be searched if was equal to the min score (e.g. break if the score is less than, but continue if equal to or greater than)

Now it must be strictly better.

jimczi: Right, that should be the inverse of https://github.com/apache/lucene/pull/14215/files#diff-ba3b6b03fc9c4dde56640064013726c8d5334076a6f8622c5f9eaee048b4ad47R237

To make it less ambiguous I wonder if `KnnCollector#minCompetitiveSimilarity` should be `Math.nextUp(minimumHeapScore)` like we do for top hits collectors?

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphSearcher.java
```diff
@@ -216,7 +216,7 @@ void searchLevel(
     while (candidates.size() > 0 && results.earlyTerminated() == false) {
       // get the best candidate (closest or best scoring)
       float topCandidateSimilarity = candidates.topScore();
-      if (minAcceptedSimilarity > topCandidateSimilarity) {
+      if (topCandidateSimilarity < minAcceptedSimilarity) {
```
iverase, line 219: I guess we need to change the consition `friendSimilarity > minAcceptedSimilarity` to `friendSimilarity >= minAcceptedSimilarity`?

lucene/core/src/java/org/apache/lucene/search/TopKnnCollector.java
```diff
@@ -52,7 +52,7 @@ public boolean collect(int docId, float similarity) {
 
   @Override
   public float minCompetitiveSimilarity() {
-    return queue.size() >= k() ? queue.topScore() : Float.NEGATIVE_INFINITY;
+    return queue.size() >= k() ? Math.nextUp(queue.topScore()) : Float.NEGATIVE_INFINITY;
```
mayya-sharipova, line 55: May be instead of adding `Math.nextUp` in different collectors, some of which we may miss, it is better to update local `minAcceptedSimilarity` in `HnswGraphSearcher:searchLevel` ?

```java
minAcceptedSimilarity =Math.nextUp( results.minCompetitiveSimilarity());
```

benwtrent: I am fine with that as well. I suppose the only thing that actually uses this is the graph searcher :/

jimczi: It was more about the high-level contract of minCompetitiveSimilarity. It's not a big deal, but it does place the responsibility on the various implementers.

tteofili: > I am fine with that as well. I suppose the only thing that actually uses this is the graph searcher :/

we now possibly have multiple class for graph searchers (extensions of `AbstractHnswGraphSearcher`), but much better than going over all of the `KnnCollectors`

tteofili (approved):

tteofili:
only perhaps it'd be nice if we could add an @Monster / nightly test to make sure we don't run in this again in the future.

PR #14218, opened 2025-02-11
Clean up more dead states creation by Automata/Operations

rmuir:
Add helper methods to AutomatonTestUtil:
* `assertCleanNFA()`: checks for no dead states and tells you what type of "dead" they are when it fails
* `assertCleanDFA()`: above, plus checks that it is deterministic
* `assertMinimalDFA()`: above, plus checks that it is minimal

Improve the simple non-random tests in `TestAutomaton` and `TestOperations`:
* remove unnecessary calls to `determinize()`
* remove unnecessary calls to `removeDeadStates()`
* insert checks of `assertMinimalDFA()`/`assertCleanDFA()`/`assertCleanNFA()`
* split some large tests into smaller tests for easier management

Fix dead-state issues found:
* `Operations.concatenate()` [corner-case missed by previous PR!]
* `Operations.repeat()`
* `Automata.makeDecimalInterval()`

The goal of this is not to solve any hard computer science problems, but just to make sure practical cases behave as we expect, and have a way to debug if they do not. Currently the tests just determinize everything which is too lenient and won't find issues or prevent future ones.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
@@ -491,6 +491,7 @@ public final class Automata {
    */
   public static Automaton makeDecimalInterval(int min, int max, int digits)
       throws IllegalArgumentException {
+    // TODO: can this be improved to always return a DFA?
     String x = Integer.toString(min);
     String y = Integer.toString(max);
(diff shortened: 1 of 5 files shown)


jpountz (approved):
Wonderful

dweiss (approved):
LGTM!

Issue #14152, opened 2025-01-18
Design: Are BlendedInfixSuggester's LINEAR_COEF and DEFAULT_NUM_FACTOR fields intended to be mutable?


paulirwin:
(Found during reviewing code while porting to the [Lucene.NET](https://github.com/apache/lucenenet/) project.)

BlendedInfixSuggester has two static, mutable fields that seem like they might have been intended to be constants instead: [`LINEAR_COEF and DEFAULT_NUM_FACTOR`](https://github.com/apache/lucene/blob/dce24c8621748d51954c4845eab7626dae166db9/lucene/suggest/src/java/org/apache/lucene/search/suggest/analyzing/BlendedInfixSuggester.java#L59-L64). We wanted to know whether it was intentional that these were mutable, or an oversight that should be corrected and both marked `final`. 

`DEFAULT_NUM_FACTOR` does not seem to have much utility if it is modified, and could result in bugs if it is modified during execution. Perhaps there is good reasoning for modifying it, though, that could be documented on the field.

`LINEAR_COEF` seems like a possible use case for modifying it would be to change the algorithm's math, but if so it is unexpected then that it is protected instead of public.

Affects Lucene 4.7.0 through current (10.1.0).

Issue #14151, opened 2025-01-18
Design: TaskSequence.REPEAT_EXHAUST is mutable


paulirwin:
### Description

(Found during reviewing code while porting to the [Lucene.NET](https://github.com/apache/lucenenet/) project.)

The static [`TaskSequence.REPEAT_EXHAUST`](https://github.com/apache/lucene/blob/dce24c8621748d51954c4845eab7626dae166db9/lucene/benchmark/src/java/org/apache/lucene/benchmark/byTask/tasks/TaskSequence.java#L33) field has been perhaps accidentally left as a mutable static field. It appears that this field is only ever used a sentinel value that is not intended to be mutated, and should be marked `final`. As far as I can tell, there is no justification to have this field be mutable. Although this is just benchmark code, and thus I'm sure very low priority, it could potentially result in bugs if it is mutated during execution. 

### Version and environment details

Affects Lucene 4.0.0 through current (10.1.0)

PR #14216, opened 2025-02-08
Upgrade errorprone to 2.36.0

risdenk:
### Description

Builds on top of https://github.com/apache/lucene/pull/14201 which started with 2.18.0 and helps with the error described here https://github.com/apache/lucene/issues/14200. This adds all the rules to `error-prone.gradle` and upgrades to the latest error prone plugin version (https://github.com/tbroyer/gradle-errorprone-plugin/releases) and error prone version.

This was checked with

```
./gradlew check -x test -Pvalidation.errorprone=true -Pjavac.failOnWarnings=false -Pvalidation.git.failOnModified=false
```

These are ones that potentially have problems but are commented out since the fixes were not trivial
* `ArrayRecordComponent`
* `ClassInitializationDeadlock`
* `LockOnNonEnclosingClassLiteral`

All the rules from the different versions are linked below:

https://github.com/google/error-prone/releases/tag/v2.19.0
- [NotJavadoc](https://errorprone.info/bugpattern/NotJavadoc)
- [StringCaseLocaleUsage](https://errorprone.info/bugpattern/StringCaseLocaleUsage)
- [UnnecessaryTestMethodPrefix](https://errorprone.info/bugpattern/UnnecessaryTestMethodPrefix)

https://github.com/google/error-prone/releases/tag/v2.20.0
- [InlineTrivialConstant](https://errorprone.info/bugpattern/InlineTrivialConstant)
- [UnnecessaryStringBuilder](https://errorprone.info/bugpattern/UnnecessaryStringBuilder)
- [BanClassLoader](https://errorprone.info/bugpattern/BanClassLoader)
- [DereferenceWithNullBranch](https://errorprone.info/bugpattern/DereferenceWithNullBranch)
- [DoNotUseRuleChain](https://errorprone.info/bugpattern/DoNotUseRuleChain)
- [LockOnNonEnclosingClassLiteral](https://errorprone.info/bugpattern/LockOnNonEnclosingClassLiteral)
- [MissingRefasterAnnotation](https://errorprone.info/bugpattern/MissingRefasterAnnotation)
- [NamedLikeContextualKeyword](https://errorprone.info/bugpattern/NamedLikeContextualKeyword)
- [NonApiType](https://errorprone.info/bugpattern/NonApiType)

https://github.com/google/error-prone/releases/tag/v2.21.0
[...]

PR #14221, opened 2025-02-11
Clean up public non-final statics

msfroh:
### Description

Following up on https://github.com/apache/lucene/issues/14151 and https://github.com/apache/lucene/issues/14152, I decided to grep for other `public static` non-`final` variables in the Lucene codebase.

This change makes `static`s `final` where it is both possible and seems appropriate. (That is, the variables really do seem to be constants.) I made a couple `private` instead, since they are modified, but privately.

I also updated some old-school code where IntelliJ flagged opportunities to use newer Java idioms.

### DIFF
--- a/lucene/analysis/smartcn/src/java/org/apache/lucene/analysis/cn/smart/AnalyzerProfile.java
+++ b/lucene/analysis/smartcn/src/java/org/apache/lucene/analysis/cn/smart/AnalyzerProfile.java
@@ -34,13 +34,9 @@ import java.util.Properties;
 public class AnalyzerProfile {
 
   /** Global indicating the configured analysis data directory */
-  public static final String ANALYSIS_DATA_DIR;
+  public static final String ANALYSIS_DATA_DIR = resolveDataDir();
 
-  static {
-    ANALYSIS_DATA_DIR = init();
-  }
-
-  private static String init() {
+  private static String resolveDataDir() {
     String dirName = "analysis-data";
     String propName = "analysis.properties";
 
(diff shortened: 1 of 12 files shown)


dweiss (approved):
Thanks! I think this can be merged as is, but I'll wait a bit if anybody else wants to chime in.

msokolov:
LGTM, thanks!

lucene/analysis/smartcn/src/java/org/apache/lucene/analysis/cn/smart/AnalyzerProfile.java
```diff
@@ -34,45 +34,44 @@
 public class AnalyzerProfile {
 
   /** Global indicating the configured analysis data directory */
-  public static String ANALYSIS_DATA_DIR = "";
+  public static final String ANALYSIS_DATA_DIR;
 
   static {
-    init();
+    ANALYSIS_DATA_DIR = init();
```
risdenk, line 40: Does this need to be in the static block? cant the line above just be

`public static final String ANALYSIS_DATA_DIR = init();`

risdenk:
https://github.com/apache/lucene/pull/14216 may help with this specifically `NonFinalStaticField` which was off since it was super noisy. https://errorprone.info/bugpattern/NonFinalStaticField

lucene/analysis/smartcn/src/java/org/apache/lucene/analysis/cn/smart/AnalyzerProfile.java
msfroh: You are absolutely right! Thanks!

risdenk (approved):

msfroh:
> #14216 may help with this specifically `NonFinalStaticField` which was off since it was super noisy. https://errorprone.info/bugpattern/NonFinalStaticField

Once #14216 and this PR are merged, I'm happy to make the required changes to enable the `NonFinalStaticField` check. It should be a lot less noisy.

dweiss:
Done. Thank you!

Issue #14180, opened 2025-01-29
Multi-threaded vector search over multiple segments can lead to inconsistent results


benwtrent:
### Description

Related to: https://github.com/apache/lucene/pull/14167

But multi-threaded search over multiple segments in addition to multi-leaf collection (e.g. information sharing) can lead to consistent results at low `k` values. 

It is possible to get more consistent results, and likely remove most inconsistencies by simply gathering more neighbors (`k` in the query, `fanout` in lucene util, or `efSearch` if you will...). However, while HNSW search is approximate, we should strive for consistency.

Zona-hu:
> ### 描述
> 相关：[#14167](https://github.com/apache/lucene/pull/14167)
> 
> 但是，除了多叶收集（例如信息共享）之外，对多个段进行多线程搜索也可以在低值下获得一致的结果`k`。
> 
> 有可能获得更一致的结果，并且可能通过简单地收集更多邻居（`k`在查询中、`fanout`在 lucene 工具中，或者`efSearch`如果你愿意的话......）来消除大多数不一致性。然而，虽然 HNSW 搜索是近似的，但我们应该努力保持一致性。
> 
> ### 版本和环境详细信息
> _没有回应_

Hello, I filed this issue:
https://github.com/apache/lucene/issues/14180
Could you please let me know which future version of Elasticsearch will resolve the vector search consistency problem?

tteofili:
> Could you please let me know which future version of Elasticsearch will resolve the vector search consistency problem?

we are investigating on a proper solution to this issue that addresses the determinism while retaining good efficiency (e.g., see some early experiments in this [draft PR](https://github.com/apache/lucene/pull/14191))

tteofili:
this should be now fixed via https://github.com/apache/lucene/pull/14226

Zona-hu:
> this should be now fixed via [#14226](https://github.com/apache/lucene/pull/14226)

Has the problem been resolved? Will the next version of Elasticsearch fix this issue?

benwtrent:
> Has the problem been resolved? Will the next version of Elasticsearch fix this issue?

It has been resolved in Lucene. There is no set fix version for anything that depends on Lucene, like Elasticsearch.

PR #14160, opened 2025-01-22
Add new Acorn-esque filtered HNSW search heuristic

benwtrent:
This is a continuation and completion of the work started by @benchaplin in https://github.com/apache/lucene/pull/14085

The algorithm is fairly simple:

 - Only score and then explore vectors that actually match the filtering criteria
 - Since this will make the graph even sparser, the search spread is increased to also include the candidate's neighbor neighbors (e.g. generally maxConn * maxConn exploration)
 - Additionally, even more scored candidates for a given NSW are considered to combat the increased sparsity

Some of the changes to the baseline Acorn algorithm are:

 - There is some general threshold of filtering that bypasses this algorithm altogether. Default suggestion is 60%.
 - The number of additional neighbors explored is predicated on the percentage of the immediate neighborhood that is filtered out
 - More extended neighborhoods will be explored if the filter is exceptionally restrictive. This attempts to find valid vectors to score and explore for every candidate.
 - Only look at the extended neighbors if less than 90% of the current neighborhood matches the filter.

Here are some numbers for 1M vectors, float32 and then int4 quantized. 

https://docs.google.com/spreadsheets/d/1GqD7Jw42IIqimr2nB78fzEfOohrcBlJzOlpt0NuUVDQ

Here is the "nightly" dataset (but I merged to a single segment)

https://docs.google.com/spreadsheets/d/1gk1uybtqleVtDUfhWXActyhW8q_lgG1mlMrOohnJRJA

Since this changes the behavior significantly, and there are still some weird edge cases, I am exposing as a parameter within a new idea called `KnnSearchStrategy` that the collectors can provide. This strategy object can be provided to the queries.

closes: https://github.com/apache/lucene/issues/13940

PR #14226, opened 2025-02-12
OptimisticKnnVectorQuery

msokolov:
### Description

This is a WIP patch to work out an idea for Knn hit collection that is deterministic and efficient in the sense that the number of hits collected per leaf scales with the size of the leaf. The idea is:

1. Run pro-rated search assuming uniform random topK-document distribution among leaves (optimistically) by scaling k to kLeaf according to the proportion of documents in the leaf.
3. Examine the list of per-leaf results. Any leaves whose minimum score `among` its kLeaf results is >= global minimum score (in top K across all leaves, merged) is submitted for further exploration using seeded search starting with the previous best results.
4. repeat 2, 3 until all leaves min scores are worse than the global min score or other limiting conditions are reached.

When the presumption of uniform distribution is valid, we would be able to skip steps 3 and 4 so we should get similar performance as we do with simple pro-rated algorithm. Otherwise we pay the cost of re-entering search, but we are guaranteed to get global (approximate) top K in a deterministic way (without cross-thread communication during search), and hopefully the cost is minimized by seeding with the best results so far.

The key point is that we make an optimisitic assumption initially, but we check and if it proved to be incorrect, we make up for it with some more work. In many circumstances the original assumption will hold, and in any case the cost of further searching *shouldn't* be too much because of the seeding.  I plan to do some perf tests; no results yet beyond unit testing. I did run TestOptimisticKnnVectorQuery with 1000 repeats and it didn't fail, including the new consistency test. in #14180.

This is a WIP only because the class design isn't really great.  It re-uses a bunch of stuff from SeededKnnVectorQuery. It's for floats only as it is. My thinking is a better way might be to make this be the default and merge it directly into AbstractKnnVectorQuery?  Or it could possilby be triggered by a KnnSearchStrategy?


benwtrent:
This is indeed interesting. I eagerly wait some of the perf numbers ;).

But, this re-entering the graph makes me think that collectors will need to track visitation. This way the same vector path isn't visited multiple times. Its possible that once you enter the lower level, that you accidentally "back track" to previously hit vectors.


> My thinking is a better way might be to make this be the default and merge it directly into AbstractKnnVectorQuery? Or it could possilby be triggered by a KnnSearchStrategy?

I do think that if this proves fruitful, it should go straight into the default behavior. As for a strategy or not, it depends on how extreme the performance & recall differences are :/. 



The reason I am thinking a search strategy for the filter stuff I have been doing here: https://github.com/apache/lucene/pull/14160 is that the recall behavior is significantly different for certain kinds of filters. Users might want to turn it off (if they have particular kinds of filters they run that behave wonky).

msokolov:
Hmm github's test run failed with:

```
gradlew :lucene:core:test --tests "org.apache.lucene.search.TestByteVectorSimilarityQuery.testFallbackToExact" -Ptests.jvms=1 -Ptests.jvmargs= -Ptests.seed=9F91AA4B44864767 -Ptests.useSecurityManager=true -Ptests.gui=true -Ptests.file.encoding=ISO-8859-1 -Ptests.vectorsize=default -Ptests.forceintegervectors=false
```

however this did not repro locally. I'm going to kick off another run

msokolov:
> But, this re-entering the graph makes me think that collectors will need to track visitation. This way the same vector path isn't visited multiple times. Its possible that once you enter the lower level, that you accidentally "back track" to previously hit vectors.

msokolov:
Well I ran some tests, and surprisingly, I saw a significant different in both recall and latency (decreases in both). This surprised me: I expected to see more-or-less similar results, so I want todig a little further.

benwtrent:
Ah, on other thought is that we are definitely scoring every seeded entry point twice. Once when they are gathered during initial query phase, then later through the seeded provisioning. We could adjust the internal API to allow cached scores to be provided (if we find that we can get the recall up).

msokolov:
You can tune it by changing the magic number 3 to a bigger number. I found that with 15 I get slight better recall and slightly lower latencies than the baseline for my test case

msokolov:
also - we are not really tracking "visited" properly I think

msokolov:
I think maybe what happens here is that the K controls not only how many hits are returned from each segment, but also what the "beam width" is during search, so we could have gotten better hits even from the segments that appeared to have been exhausted (because their worst in the top K weas below the min competitive score)

benwtrent:
I am not 100% sure whats up with the behavior. However, I switched to `16` (also happens to be the graph conn) instead of `3`.

Its interesting how visited is lower, but recall is on par with the current non-consistent baseline. I don't like the magic "16" here, without knowing exactly why its working.

I also fixed visited in Lucene util (accounting for the multiple topdoc merges).

Here is NOT reusing the entry point scores
```
recall  latency(ms)     nDoc  topK  fanout  visited  num segments  selectivity
 0.942       10.400  8000000   100     100    68605           128        1.000
```

Here IS reusing the entry point scores (have a local refactor). So, visited drops just a little.
```
recall  latency(ms)     nDoc  topK  fanout  visited  num segments  selectivity
 0.942        7.400  8000000   100     100    68477           128        1.000
```

here is reusing entry point, single threaded:
```
recall  latency(ms)     nDoc  topK  fanout  visited  num segments  selectivity
 0.942       49.300  8000000   100     100    68477           128        1.000
```


Note, the visited for a SINGLE graph is around 3000.

Also, as a further baseline, here is no information sharing (e.g. what it was before our in-consistent multi-threaded results). e.g. just fully exploring every graph:

```
recall  latency(ms)     nDoc  topK  fanout  visited  num segments  selectivity
 1.000      213.400  8000000   100     100   412280           128        1.000
```

msokolov:
The latency improvement from re-using scores is surprisingly large. I would have expected costs to be dominated by newly-explored nodes, but this is cool.

msokolov:
I don't believe 16 is "special" except in the sense that it happens to be a sweet spot is this context.  We expect that as we increase that per-segment factor we will get increased recall because it is the equivalent of the old "fanout" parameter we used to have - it expands the beam of the search while keeping the number of results returned fixed.  This is different from "fanout" because that was a global setting, and this one scales with the relative size of the segment.

What I'd like to understand is whether there is a fixed value for this sweet spot, or whether it changes with {data, graph construction parameters, something else}. It seems clear that my idea that it is a probabilistic thing is incorrect since 16 standard deviations is just crazy.  But we can at least look at it empirically.

msokolov:
@benwtrent I'm curious how you managed to re-use the scores. I poked around a little and I guess it requires some new plumbing / class casts since the API isn't really designed for it? Do you have a patch you wouldn't mind posting?

benwtrent:
@msokolov I can provide a quick patch tomorrow against main. As I said, it would work for stuff as it is now.

lucene/core/src/java/org/apache/lucene/search/OptimisticKnnVectorQuery.java
```diff
@@ -0,0 +1,205 @@
...
+          new TimeLimitingKnnCollectorManager(
+              new ReentrantKnnCollectorManager(
+                  getKnnCollectorManager(kInLoop, indexSearcher), perLeafResults),
+              indexSearcher.getTimeout());
+      // System.out.println("k=" + k + " kloop=" + kInLoop);
+      Iterator<LeafReaderContext> ctxIter = leafReaderContexts.iterator();
+      while (ctxIter.hasNext()) {
+        LeafReaderContext ctx = ctxIter.next();
+        TopDocs perLeaf = perLeafResults.get(ctx.ord);
+        if (perLeaf.scoreDocs.length > 0
```
dungba88, line 108: Can we publish the metrics for each case here to see how often/how many passes we usually need to take?

msokolov: yes that would be critical to understanding what's going on here! Maybe we can extend TopK to include some metrics? Not sure what the best API is here

lucene/core/src/java/org/apache/lucene/search/OptimisticKnnVectorQuery.java
```diff
@@ -0,0 +1,205 @@
...
+    KnnCollectorManager manager =
+        (visitedLimit, strategy, context) -> {
+          @SuppressWarnings("resource")
+          float leafProportion =
+              context.reader().maxDoc() / (float) context.parent.reader().maxDoc();
+          int perLeafTopK = perLeafTopKCalculation(k, leafProportion);
+          assert perLeafTopK
+              > 0; // if we divided by zero above, leafProportion can be NaN and then this would be
+          // 0
+          return new TopKnnCollector(perLeafTopK, visitedLimit);
```
dungba88, line 146: Would it make sense to cap `perLeafTopK` by the original k? I think k is double on every iteration, and `perLeafTopK` can theoretically go over the original k, which is excessive.

lucene/core/src/java/org/apache/lucene/search/OptimisticKnnVectorQuery.java
msokolov: Yes, OK, pass in some kind of abstract Counter?  But maybe a simple `int[]` is enough? The consumer can always convert to their metric system.

lucene/core/src/java/org/apache/lucene/search/OptimisticKnnVectorQuery.java
msokolov: yes, we can do it in perLeafTopkCalculation

dungba88:
I ran some benchmark with Cohere 768 dataset for 3 different algorithms: (1) the baseline "greedy", (2) this PR "optimistic", and (3) with only "pro-rata". (2) and (3) will converge with fanout >= 20, it kinda makes sense because as we increase the `ef` parameters, it would already look deeper into the graph. (2) increase ~6% recall compared to (3) at fanout=0, and the effect got diminished the higher fanout, such as ~3% at fanout=5 and 1% at fanout=10.

(2) still missed some matches (about 8%) when compared to (1), I'm wondering why is that, as we tried to exhaustively search in segments.

Other params: ndoc=500K, topK=100

```
recall  latency_(ms)  fanout  quantized  greediness  index_s  index_docs/s  force_merge_s  num_segments  index_size_(MB)  vec_disk_(MB)  vec_RAM_(MB)   algorithm 
 0.936        13.285       0         no        0.00   117.24       4264.94           0.00             7          1490.05       1464.844      1464.844  greediness 
 0.933        12.766       0         no        0.10   116.45       4293.84           0.00             7          1490.14       1464.844      1464.844  greediness 
 0.927        12.045       0         no        0.30   116.69       4285.04           0.00             7          1490.12       1464.844      1464.844  greediness 
 0.923        11.709       0         no        0.50   117.74       4246.79           0.00             7          1490.06       1464.844      1464.844  greediness 
 0.917        10.950       0         no        0.70   117.68       4248.67           0.00             7          1490.06       1464.844      1464.844  greediness 
 0.907        10.996       0         no        0.90   117.40       4258.80           0.00             7          1489.98       1464.844      1464.844  greediness 
 0.836        10.349       0         no        1.00   117.07       4271.13           0.00             7          1490.03       1464.844      1464.844  greediness 
 0.846         6.682       0         no       -1.00   116.50       4291.77           0.00             7          1490.09       1464.844      1464.844  optimistic
 0.862         6.699       5         no       -1.00   118.33       4225.44           0.00             7          1489.97       1464.844      1464.844  optimistic
 0.871         7.172      10         no       -1.00   116.10       4306.63           0.00             7          1490.05       1464.844      1464.844  optimistic
 0.897         7.962      20         no       -1.00   116.11       4306.15           0.00             7          1490.06       1464.844      1464.844  optimistic
 0.925        10.975      50         no       -1.00   116.71       4284.23           0.00             7          1490.10       1464.844      1464.844  optimistic
 0.941        15.635     100         no       -1.00   116.58       4289.08           0.00             7          1490.13       1464.844      1464.844  optimistic
 0.786         5.727       0         no       -1.00   117.86       4242.25           0.00             7          1490.10       1464.844      1464.844    prorata
 0.831         6.176       5         no       -1.00   117.27       4263.70           0.00             7          1490.07       1464.844      1464.844    prorata
 0.857         6.746      10         no       -1.00   116.79       4281.04           0.00             7          1490.08       1464.844      1464.844    prorata
 0.891         7.763      20         no       -1.00   116.21       4302.59           0.00             7          1490.09       1464.844      1464.844    prorata
 0.921        11.083      50         no       -1.00   118.02       4236.46           0.00             7          1490.06       1464.844      1464.844    prorata
 0.940        15.694     100         no       -1.00   116.63       4286.91           0.00             7          1490.00       1464.844      1464.844    prorata
```

Screenshot 2025-02-22 at 21 17 40

dungba88:
Here is the graph with various values of stdev (3, 9, 16, 25). Again both (2) and (3) converge with a stdev value of 9, and all 3 converge at a stdev value of 16 and greediness of 0.7. With this dataset and settings, using 9 seems to be the most efficient, but then there is virtually no difference between (2) and (3).

Screenshot 2025-02-23 at 21 31 42

msokolov:
I pushed a version that re-uses scores *and* limits per-leaf topK to global topK.  The former didn't make very much difference, but the latter change did improve things quite a bit.  Here are some numbers from cohere/768d:

### mainline
recall  latency (ms)    nDoc  topK  fanout  maxConn  beamWidth  quantized  index s  index docs/s  num segments  index size (MB)  vec disk (MB)  vec RAM (MB)
0.954        12.919  500000    50       0       64        250         no    13786     0.00      Infinity             8          1501.70       1464.844      1464.844
0.981        18.488  500000    50      50       64        250         no    20371     0.00      Infinity             8          1501.70       1464.844      1464.844
0.989        22.948  500000    50     100       64        250         no    24963     0.00      Infinity             8          1501.70       1464.844      1464.844

### wih reused scores *and* limiting perLeafK <= K
Results:
recall  latency (ms)    nDoc  topK  fanout  maxConn  beamWidth  quantized  visited  index s  index docs/s  num segments  index size (MB)  vec disk (MB)  vec RAM (MB)
 0.959        11.375  500000    50       0       64        250         no    12086   308.23       1622.15             8          1501.70       1464.844      1464.844
 0.979        14.926  500000    50      50       64        250         no    16724     0.00      Infinity             8          1501.70       1464.844      1464.844
 0.987        17.858  500000    50     100       64        250         no    20277     0.00      Infinity             8          1501.70       1464.844      1464.844

it would be awesome if you could produce similar comparisons for this version, @dungba88 !

benwtrent:
> so I rebased and removed the reuse-scores part of this since it was conflicting with other changes and doesn't seem worth preserving

@msokolov thanks for confirming and digging into that :). It was worth a shot! But its not super surprising as the dominate cost will be the further exploration.

dungba88:
@msokolov I didn't see the change to cap `perLeafTopK` in the [latest commit](https://github.com/apache/lucene/pull/14226/commits/5b06e168c683e3edf36987379091357c298f0f28#diff-6bf79d1f0edcb0d2cf0c56c71540794f9e6edfd35973cfe47f30b6cac7046396), maybe it has not been published yet?

msokolov:
Also, I forgot about this comment:

> Would it make sense to cap perLeafTopK by the original k? I think k is double on every iteration, and perLeafTopK can theoretically go over the original k, which is excessive.

I added an additional cap on this, but then realized we are already implicitly imposing such a limit here:

```
if (perLeaf.scoreDocs.length > 0  
    && perLeaf.scoreDocs[perLeaf.scoreDocs.length - 1].score >= minTopKScore
    && perLeafTopKCalculation(kInLoop / 2, ctx.reader().maxDoc() / (float) reader.maxDoc())
           <= k + 1) { 
```

by the way another thing we might want to try is relaxing this reentry check a bit by looking at the second- or third-worst per-leaf score, because in theory `lambda` created a buffer that should cause leaves to collect deeper than the best top K.  This could enable this per-leaf strategy to outperform the global fanout? Anyway it's easy to try

dungba88:
> I added an additional cap on this, but then realized we are already implicitly imposing such a limit here:

@msokolov that checks if the *previous* iteration (kInLoop / 2) has exceeded topk, not *next* iteration. E.g if global k is 100, and the current loop has kInLoop = 60, then the next one would be 120 and can still exceed the global k. I think we should still cap the effective per-leaf k.

Unrelatedly, another observation I have is that, at some iterations when there is only 1 or few segments left, we are still using per-leaf pro-rata as if we are still running with every segments. For example, in some extreme cases where one segments should have all the best matches, but that segment is so small so the per-leaf top k can only increase slowly after each pass (2 -> 4 -> 8 -> 16). That makes me think:
- Could we readjust the pro-rata rate, not based on the whole index, but based on the effective segments?
- What if we just set the per-leaf k to the same as global k in the second pass, and stop at second pass? I'm curious about the overhead of re-entry vs the benefit

dungba88:
Sorry I'm back today and will try this idea first

> - What if we just set the per-leaf k to the same as global k in the second pass, and stop at second pass? I'm curious about the overhead of re-entry vs the benefit

msokolov:
Another wrinkle here is that KnngraphTester currently does not ever run its queries multithreaded; it does not pass an Executor when it creates IndexSearcher. Because of that we are not truly testing the behavior of the existing AbstractKnnVectorQuery, that can potentially behave differently when run multithreaded, possibly consuming more total CPU and even yielding different results, because in the single-threaded case where each segment is searched in succession, the threshold derived from prior segments is used when searching the successor segments, but when searched concurrently, information propagates *while* searching, so some prospective unnecessary searching may be happening, and for all we know, may be yielding some additional matches too.  Net/net I think we ought to be adding some multithreaded test capability to KnnGraphTester.

benwtrent:
> Net/net I think we ought to be adding some multithreaded test capability to KnnGraphTester.

Agreed, I think a "num_search_threads" parameter would be beneficial. Then the searcher that is created can have a fixed executor `ExecutorService executorService = num_search_threads > 1 ? Executors.newFixedThreadPool(num_search_threads) : null;` passed to it in `testSearch`. 

```
          IndexSearcher searcher = new IndexSearcher(reader, executorService);
```

msokolov:
> Agreed, I think a "num_search_threads" parameter would be beneficial.

dungba88:
I tried the idea of stopping at second pass with the original k, but the benchmark results look weird for all algorithms, as if it doesn't matter at all. This is much different from the last runs.

```
recall  latency_(ms)  fanout  quantized  greediness  index_s  index_docs/s  force_merge_s  num_segments  index_size_(MB)  vec_disk_(MB)  vec_RAM_(MB)   algorithm  stdev
 0.959       150.588       0         no       -1.00   117.12       4269.31           0.00             7          1490.07       1464.844      1464.844     prorata      3
 0.959       152.532       0         no       -1.00   116.56       4289.64           0.00             7          1489.97       1464.844      1464.844     prorata      6
 0.956       160.835       0         no       -1.00   117.71       4247.84           0.00             7          1489.95       1464.844      1464.844     prorata      9
 0.960       157.832       0         no       -1.00   116.22       4302.11           0.00             7          1490.00       1464.844      1464.844     prorata     16
 0.960       155.564       0         no       -1.00   117.92       4240.05           0.00             7          1490.03       1464.844      1464.844  optimistic      3
 0.961       152.263       0         no       -1.00   118.04       4235.96           0.00             7          1490.08       1464.844      1464.844  optimistic      6
 0.962       158.384       0         no       -1.00   118.79       4209.25           0.00             7          1490.06       1464.844      1464.844  optimistic      9
 0.959       162.569       0         no       -1.00   117.15       4268.14           0.00             7          1489.99       1464.844      1464.844  optimistic     16
 0.956       152.338       0         no       -1.00   119.60       4180.57           0.00             7          1489.98       1464.844      1464.844  simplified      3
 0.960       161.406       0         no       -1.00   117.71       4247.69           0.00             7          1490.12       1464.844      1464.844  simplified      6
 0.960       151.540       0         no       -1.00   117.11       4269.42           0.00             7          1490.02       1464.844      1464.844  simplified      9
 0.959       159.400       0         no       -1.00   116.79       4281.34           0.00             7          1490.10       1464.844      1464.844  simplified     16
```

Will try to look deeper.

dungba88:
Ran again with the second idea that the pro-rata rate will be based on the active segments in the current pass instead of the whole index (called "adaptive" in the graph). However I didn't see much improvement and it only increases the latency, doesn't seem to worth pursuing.

The "simplified" version now only has a slight more latency compared to "optimistic" with lambda=2. I think it's comparable to the "optimistic", subjected to some noise. Maybe we can "opt" for this simplified 2-pass collection instead multi-pass?

Screenshot 2025-03-10 at 18 31 22

msokolov:
I think the simplified version makes the most sense -- Just confirming: this is where we first search pro-rated across segments, and then if any segment has its output queue full with competitive hits, we revisit it using the prior hits as entry points and a queue size of K (the full original K).

My only question is whether lambda=2 is correct - the graphs show lambda 6+ looking more reasonable, and indeed they don't seem to show a data point for 2 at all?

dungba88:
> this is where we first search pro-rated across segments, and then if any segment has its output queue full with competitive hits, we revisit it using the prior hits as entry points and a queue size of K (the full original K).

Yes that's correct

> My only question is whether lambda=2 is correct - the graphs show lambda 6+ looking more reasonable, and indeed they don't seem to show a data point for 2 at all?

I've benchmarked with the minimum of 3 so there is no data points for 2. For lambda=3, it's interesting that the simplified version is slower than the optimistic version. This is (somewhat) expected as with the simplified version on the second pass we immediately use the queue size of k (k=100 here) while the optimistic version will gradually increase the queue size, so that it may end up with a size much less than k. When we increase lambda, the difference becomes diminished as there is less re-entry needed and the queue size difference would also be smaller.

dungba88:
Ah right, sorry it was a typo :)

mikemccand:
> > Net/net I think we ought to be adding some multithreaded test capability to KnnGraphTester.
> 
> Agreed, I think a "num_search_threads" parameter would be beneficial. Then the searcher that is created can have a fixed executor `ExecutorService executorService = num_search_threads > 1 ? Executors.newFixedThreadPool(num_search_threads) : null;` passed to it in `testSearch`.
> 
> ```
>           IndexSearcher searcher = new IndexSearcher(reader, executorService);
> ```

I opened https://github.com/mikemccand/luceneutil/issues/350 to enable intra-query concurrency for KNN benchy, but +1 to making sure we report total CPU ticks (aggregated across all threads that participated for each query) in addition to "wall clock latency".  Both metrics are useful.

dungba88:
I published the simplified version here for reference: https://github.com/dungba88/lucene/commit/278d7c919bc6ca6e1618868a892bcf3d4970cea5

benwtrent:
@dungba88 @msokolov 

Do we know where we stand with this? I wonder if we should simply use this to replace the current buggy behavior as fanning out to segments multiple times with sane sync points should fix our consistency issues and if the performance of this is on-par with the current information sharing logic, seems like a net-win.

Anything I can do to help here?

dungba88:
I'm also in favor of making this the default behavior. Maybe we can let the lambda parameter to be tuned with a reasonable default value (similar to greediness in the current collector) so that expert can tune if they need. From the benchmark we also saw that the tuning range and impact of this optimistic algorithm is larger than the global sharing algorithm, so it may satisfies a wider audiences.

dungba88:
@msokolov I can publish the PR for the simplified version with directly changing the default behavior if there is no objection (if you are not available for it).

msokolov:
> I'm also in favor of making this the default behavior. Maybe we can let the lambda parameter to be tuned with a reasonable default value (similar to greediness in the current collector) so that expert can tune if they need. From the benchmark we also saw that the tuning range and impact of this optimistic algorithm is larger than the global sharing algorithm, so it may satisfies a wider audiences.

Regarding lambda setting, I'm in favor of fixing it as a  magic number that replicates something close to the current behavior (or better recall if we can and retain the same latency) and letting users tune further using fanout. I think these are roughly equivalent and I don't think we should be exposing a lot of knobs. Experts who want to can write their own Query.

benwtrent:
> , I'm in favor of fixing it as a magic number that replicates something close to the current behavior (or better recall if we can and retain the same latency as we found with lambda=16) and letting users tune further using fanout. I think these are roughly equivalent and I don't think we should be exposing a lot of knobs.

I am for this as well. Exposing more and more knobs makes things way too complicated.

msokolov:
As far as making this the default, that sounds OK to me, but let's not backport until we've had a chance to verify no harm for a while in some pre-production environments

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -93,22 +100,62 @@ public Query rewrite(IndexSearcher indexSearcher) throws IOException {
...
-    TopDocs topK = mergeLeafResults(perLeafResults);
+    Map<Integer, TopDocs> perLeafResults = new HashMap<>();
+    TopDocs topK = runSearchTasks(tasks, taskExecutor, perLeafResults, leafReaderContexts);
+    int reentryCount = 0;
+    if (topK.scoreDocs.length > 0 && perLeafResults.size() > 1) {
+      float minTopKScore = topK.scoreDocs[topK.scoreDocs.length - 1].score;
+      TimeLimitingKnnCollectorManager knnCollectorManagerInner =
+          new TimeLimitingKnnCollectorManager(
+              new ReentrantKnnCollectorManager(
+                  getKnnCollectorManager(k, indexSearcher), perLeafResults),
```
dungba88, line 116: I think this is the second pass right? If so, we need to get the TopKnnCollector with full k instead of per leaf?

dungba88 (commented):
Oh I see the latest commit already incorporates the simplified version and make it the default behavior. That's nice! Will just comment here instead.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
msokolov: right! duh - I just collapsed all the code, haven't had a chance to retest w/luceneutil. Thanks for checking!  This is hard to unit test, grr

msokolov:
I did another round of testing with luceneutil over a couple different vector datasets and it is consistently outperforming the baseline. Given that and @benwtrent's recent results, I think we can go ahead.

PR #14228, opened 2025-02-13
Enable error-prone checks for NonFinalStaticField

msfroh:
### Description

On https://github.com/apache/lucene/pull/14221, I foolishly asserted that the error-prone `NonFinalStaticField` checks would be "a lot less noisy". It seems that PR and its mere 12 files had only scratched the surface.

A couple of hours, 153 Java source files (plus one Perl script that generates a Java file) later, we can enable the `NonFinalStaticField` check.

The good (?) news is that I "only" needed to resort to suppressing the warnings in 52 files, and was able to fix the other 101. 🎉

### DIFF
--- a/gradle/validation/error-prone.gradle
+++ b/gradle/validation/error-prone.gradle
@@ -486,7 +486,7 @@ allprojects { prj ->
             // '-Xep:NonApiType:OFF', // noisy
             // '-Xep:NonAtomicVolatileUpdate:OFF', // TODO: there are problems
             // '-Xep:NonCanonicalType:OFF', // noisy
-            // '-Xep:NonFinalStaticField:OFF', // noisy
+            '-Xep:NonFinalStaticField:WARN',
             '-Xep:NonOverridingEquals:WARN',
             '-Xep:NotJavadoc:WARN',
             '-Xep:NullOptional:WARN',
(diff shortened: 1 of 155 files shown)


rmuir:
The hunspell test failure is likely an upstream issue with libreoffice dictionaries. It happened recently to me with mongolian dictionaries.

Problems:
* the test doesn't execute enough: then if your PR touches hunspell, you have to figure it out.
* the test does not use pinned copy of LO dictionaries, so it may suddenly start failing for unrelated reasons.

I'm not sure what to suggest to make it better. I can try to look into this failure tonight.

dweiss:
I'll create a separate issue for hunspell tests and take care of that tomorrow, no worries.

rmuir:
@msfroh if you get a chance, I think you just need to merge in `main` and it will take care of that CI issue.

dweiss (approved):

dweiss:
I've merged this. There are conflicts when backporting to branch_10x and they sort of depend on previous fixes to spotless, so I leave this patch on main only. I think a backport to 10x would be nice but perhaps it's worth another issue since it's not trivial.

msfroh:
>  I think a backport to 10x would be nice but perhaps it's worth another issue since it's not trivial.

I can take care of that. 

After going down the rabbit hole once with this PR, I think I'll use it as reference to split out the different solutions into their own smaller PRs.

dweiss:
Thanks, @msfroh ! I think it should be possible to cherry pick a series of commits from main - perhaps with minor adjustments. When you try to cherry pick this commit from main, you'll see a conflict in gradle's error-prone configuration - you'd need to follow back on the history of commits to that file and perhaps apply/consolidate these changes first, then this commit would apply cleanly. I hope. :)

PR #14232, opened 2025-02-13
Fix TestForTooMuchCloning flakiness

### LINKED ISSUE #14220: TestForTooMuchCloning.test fails
### Description

I have seen this fail in periodic CI builds and on some PR builds.

```
TestForTooMuchCloning > test FAILED
--
  | java.lang.AssertionError: too many calls to IndexInput.clone during merging: 519
  | at __randomizedtesting.SeedInfo.seed([274680B90E738EFA:AF12BF63A08FE302]:0)
  | at org.junit.Assert.fail(Assert.java:89)
  | at org.junit.Assert.assertTrue(Assert.java:42)
  | at org.apache.lucene.index.TestForTooMuchCloning.test(TestForTooMuchCloning.java:63)
  | at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
  | at java.base/java.lang.reflect.Method.invoke(Method.java:580)

```

Or:

```
TestForTooMuchCloning > test FAILED
    java.lang.AssertionError: too many calls to IndexInput.clone during merging: 503
        at __randomizedtesting.SeedInfo.seed([FB5D79942E80A9F0:7309464E807CC408]:0)
        at org.apache.lucene.index.TestForTooMuchCloning.test(TestForTooMuchCloning.java:63)
```

```
./gradlew test --tests TestForTooMuchCloning.test -Dtests.seed=FB5D79942E80A9F0 -Dtests.locale=sl-Latn-SI -Dtests.timezone=America/Mexico_City -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```

benwtrent:
Ever since: https://github.com/apache/lucene/pull/14165

This test has been flaky. It fails as the number of clone calls during indexing exceeds 500. 

I tried only updating the merge scheduler to serial, but i would still get clone calls > 500. 

I tried only upping the limit to 600, but with concurrent merge scheduler, I would still see clone calls > 600.

So, switching to serial merge scheduler and 600 as the new limit. I do not know if this is expected.

closes: https://github.com/apache/lucene/issues/14220

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/index/TestForTooMuchCloning.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestForTooMuchCloning.java
@@ -42,6 +42,8 @@ public class TestForTooMuchCloning extends LuceneTestCase {
             random(),
             dir,
             newIndexWriterConfig(new MockAnalyzer(random()))
+                // to reduce flakiness on merge clone count
+                .setMergeScheduler(new SerialMergeScheduler())
                 .setMaxBufferedDocs(2)
                 // use a FilterMP otherwise RIW will randomly reconfigure
                 // the MP while the test runs
@@ -62,7 +64,7 @@ public class TestForTooMuchCloning extends LuceneTestCase {
     // System.out.println("merge clone count=" + cloneCount);
     assertTrue(
         "too many calls to IndexInput.clone during merging: " + dir.getInputCloneCount(),
-        dir.getInputCloneCount() < 500);
+        dir.getInputCloneCount() < 600);
 
     final IndexSearcher s = newSearcher(r);
     // important: set this after newSearcher, it might have run checkindex


jpountz:
Thanks for looking into this and sorry for missing the build failures. The fact that this test has failures makes sense to me since merging is a bit more aggressive now, though I don't expect it to be a problem in practice. Another way of making the number of clones more predictable would be to avoid merges below the floor size, ie.

```patch
diff --git a/lucene/core/src/test/org/apache/lucene/index/TestForTooMuchCloning.java b/lucene/core/src/test/org/apache/lucene/index/TestForTooMuchCloning.java
index 92470083540..4834ad9219f 100644
--- a/lucene/core/src/test/org/apache/lucene/index/TestForTooMuchCloning.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestForTooMuchCloning.java
@@ -37,6 +37,7 @@ public class TestForTooMuchCloning extends LuceneTestCase {
     final MockDirectoryWrapper dir = newMockDirectory();
     final TieredMergePolicy tmp = new TieredMergePolicy();
     tmp.setSegmentsPerTier(2);
+    // Disable merges below the floor size, which are otherwise allowed to merge more than segmentsPerTier segments together
+    tmp.setFloorSegmentMB(Double.MIN_VALUE);
     final RandomIndexWriter w =
         new RandomIndexWriter(
             random(),

```

benwtrent:
@jpountz I tried that while reverting my other changes and it was still over 500 over many runs. I will just merge what I got. Thanks!

jpountz:
Thanks @benwtrent

PR #14236, opened 2025-02-13
Move `CombinedFieldQuery` to the core module.

jpountz:
`CombinedFieldQuery` is Lucene's most robust way of scoring across multiple fields, let's move it to core and recommend using it to query multiple fields.

While moving the class, I modified the API so that it supports configuring a single term.


jpountz:
Thanks for correcting the title @gsmiller.

This doesn't look controversial to me and there haven't been objections, so I'm feeling free to merge.

gsmiller:
> Thanks for correcting the title @gsmiller.

Ha, sorry couldn't help fixing the typo. I had this open to have a look then got busy with other stuff and never came back to it (wanted to dig in a bit more on `CombinedFieldQuery` since I'm unfamiliar with it, but agreed that moving it out of sandbox doesn't appear to be a concern)

PR #14239, opened 2025-02-14
hunspell: improve tolerateAffixRuleCountMismatches() for common problems

rmuir:
When the user opts-in to tolerateAffixRuleCountMismatches(), it tolerates some, but not all affix mismatches. Support two cases:

"EOF case": count is wrong and then we encounter end of file. previously the exception looked like this:

    java.text.ParseException: Premature end of rules for SFX B Y 2

"stampede case": in this case count is wrong and another instruction follows (such as another affix's rule).
Previously, the exception looked like this:

    java.text.ParseException: Affix rule mismatch. Header: SFX A Y 2; rule: SFX B Y 2

This PR should fix the issue in CI with the spanish dictionary (I think, we'll see). I will followup with the upstream.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -82,6 +82,9 @@ Improvements
   the wrapped analyzer's strategy to decide if components can be reused or need
   to be updated. (Mayya Sharipova)
 
+* GITHUB#14239: Hunspell's option to tolerate affix rule count mismatches was
+  improved to tolerate more instances of this problem.  (Robert Muir)
+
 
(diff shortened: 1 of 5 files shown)


rmuir:
using the `mark()/reset()` like this can be invitation for trouble, but the situation is contained: the parser will always make forward progress so it can't go crazy or infinite. Also, the user opted into the lenient parse.

donnerpeter (approved):
Great, thank you!

dweiss (approved):

PR #14242, opened 2025-02-14
Simplify operations by avoiding calculation of extra bitwise operations

### LINKED ISSUE #14240: Simplify bitwise operations in DefaultVectorUtilSupport
### Description

Currently, `DefaultVectorUtilSupport` contains methods that use unnecessary bitwise operations. These can be replaced with direct values to eliminate extra computations. For example, instead of `a.length & ~(4 - 1)`, simply use `a.length & -4`, and so on. A minor change would be necessary for this

giorgigagnidze16:
### Description
A minor change to eliminate unnecessary bitwise calculations, replaced with actual expected value

Example: 
`int upperBound = a.length & ~(4 - 1); ` -> `int upperBound = a.length & -4;`

Resolves #14240

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/internal/vectorization/DefaultVectorUtilSupport.java
+++ b/lucene/core/src/java/org/apache/lucene/internal/vectorization/DefaultVectorUtilSupport.java
@@ -45,7 +45,7 @@ final class DefaultVectorUtilSupport implements VectorUtilSupport {
       float acc2 = 0;
       float acc3 = 0;
       float acc4 = 0;
-      int upperBound = a.length & ~(4 - 1);
+      int upperBound = a.length & -4;
       for (; i < upperBound; i += 4) {
         acc1 = fma(a[i], b[i], acc1);
         acc2 = fma(a[i + 1], b[i + 1], acc2);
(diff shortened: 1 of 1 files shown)


jpountz:
Thank you. There shouldn't be any performance difference since I would expect the compiler to pre-compute the result of `~(4-1)` since these are constants, so this change is only about readability. The use of negative numbers works for me as a way to convey the number of trailing zeros that are clear, it looks slightly simpler.

Double checking that it works for you too @rmuir?

rmuir:
This was written this way (with a bit of redundancy) on purpose to make it 100% clear it is consistent with `VectorSpecies.loopBound`: 

> As long as VLENGTH is a power of two, then the result is also equal to length & ~(VLENGTH - 1).

https://docs.oracle.com/en/java/javase/23/docs/api/jdk.incubator.vector/jdk/incubator/vector/VectorSpecies.html#loopBound(int)

rmuir:
> Double checking that it works for you too @rmuir?

As long as we fix javadocs of VectorSpecies.loopBound to match :) 

Vector operations loops are not the place to optimize or try to reshuffle code. I can't tell you how many times I have "slightly rearranged loop" and then it is not longer vectorized by the compiler. It is stupidly annoying when trying to use the vector API. So that's why the code was written "by the book" in this way.

PR #14244, opened 2025-02-14
Relation check within 1D BKD Leaves

gf2121:
This patch tries to introduce additional relation check to shortcut visiting within BKD leaves. This helps queries that hit few docs like PointInSetQuery or narrow RangeQueries on high-cardinality fields, without too much overhead to many-hits queries since their bottleneck is not visiting edge leaves.

The goal that visits less points can be achieved by reducing `maxPointsInLeafNode` as well. This patch mainly aims to optimize default config and help cases that need many-hits queries and few-hits queries on the same format.

wikimediumall:

```
                          IntNRQ       66.70      (3.8%)       65.73      (3.5%)   -1.5% (  -8% -    6%) 0.215
               TermDayOfYearSort       83.39      (6.3%)       82.40      (7.6%)   -1.2% ( -14% -   13%) 0.591
                  FilteredIntNRQ       64.46      (3.7%)       63.90      (4.9%)   -0.9% (  -9% -    8%) 0.522
             CountFilteredIntNRQ       38.03      (2.7%)       37.94      (3.9%)   -0.2% (  -6% -    6%) 0.821
                      TermDTSort       84.62      (3.4%)       85.39      (3.8%)    0.9% (  -6% -    8%) 0.431
                          IntSet      248.07      (3.9%)      277.50      (4.9%)   11.9% (   2% -   21%) 0.000
```

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
@@ -443,12 +443,7 @@ public abstract class PointRangeQuery extends Query {
             new IntersectVisitor() {
               @Override
               public void visit(int docID) {
-                // this branch should be unreachable
-                throw new UnsupportedOperationException(
-                    "This IntersectVisitor does not perform any actions on a "
-                        + "docID="
-                        + docID
-                        + " node being visited");
+                matchingNodeCount[0]++;
               }
 
               @Override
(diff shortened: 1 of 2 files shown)


jpountz:
I understand the idea, but this looks a bit like a hack to me. I'm also not too fond of optimizing point-in-set queries, as terms would be a better way to index the data than points if the goal is to search for exact values?

PR #14246, opened 2025-02-16
Repeatable hunspell tests

### LINKED ISSUE #14235: Figure out why hunspell tests occasionally fail and make them more consistent
### Description

A spinoff from #14228 

_No response_

dweiss:
Fixes #14235 

Sorry this took surprisingly long. I've tried to clean up a few things along the way:
* the repositories are now pulled at constant refs embedded inside the gradle file, unless an explicit reference is passed using gradle properties for each repository. So `-Ptests.hunspell.libreoffice.ref=master` will test against the master branch (or any other sha provided there). For woorm, it's `-Ptests.hunspell.woorm.ref=main`. 
* I've added a scheduled workflow run every Monday against latest dev branches of both projects.

I also cleaned up gradle scripts a bit.
* removed the separate test task and replaced it with a property that controls whether we pull/ test against remote repositories: `-Ptests.hunspell.regressions=true`. The separate task wasn't configured properly (it wasn't running in module mode) and in general I think it's better to just keep the 'test' task for running tests.
* gh workflows run just the `TestAllDictionaries` class, not everything.

I also had to open up some core packages to the test framework - this is a side effect of TestAllDictionaries measuring memory of Dictionary objects it creates. I think it'd be nice to make Dictionary implement Accountable and remove opens statements but it's probably a follow up issue.

### DIFF
--- a/.github/workflows/run-scheduled-hunspell.yml
+++ b/.github/workflows/run-scheduled-hunspell.yml
@@ -4,7 +4,8 @@ on:
   workflow_dispatch:
 
   schedule:
-    - cron: '13 4 * * 1' // 4:13 on Mondays
+    # 4:13 on Mondays
+    - cron: '13 4 * * 1'
 
 env:
   DEVELOCITY_ACCESS_KEY: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
--- a/.github/workflows/run-scheduled-hunspell.yml
+++ b/.github/workflows/run-scheduled-hunspell.yml
@@ -1,0 +1,31 @@
+name: "Run scheduled checks: Hunspell tests against latest dictionaries"
+
+on:
+  workflow_dispatch:
+
+  schedule:
+    - cron: '13 4 * * 1' // 4:13 on Mondays
+
+env:
+  DEVELOCITY_ACCESS_KEY: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
+
+jobs:
+  test:
+    name: Hunspell regression tests against latest dictionaries
+    timeout-minutes: 15
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/prepare-for-build
+
+      - name: Run Hunspell regression tests against latest commits in dictionary repositories
+        run: >
+          ./gradlew -p lucene/analysis/common 
+          -Ptests.hunspell.regressions=true
+          -Ptests.verbose=true
+          -Ptests.hunspell.libreoffice.ref=master
+          -Ptests.hunspell.woorm.ref=main
+          test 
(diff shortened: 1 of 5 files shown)


.github/workflows/run-scheduled-hunspell.yml
```diff
@@ -0,0 +1,32 @@
...
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/prepare-for-build
+
+      - name: Run Hunspell regression tests against latest commits in dictionary repositories
+        run: >
+          ./gradlew -p lucene/analysis/common 
+          -Ptests.hunspell.regressions=true
+          -Ptests.verbose=true
+          -Ptests.hunspell.libreoffice.ref=master
+          -Ptests.hunspell.woorm.ref=main
```
dweiss, line 30: Here we pick the 'latest' commit from both branches for the scheduled workflow.

lucene/analysis/common/build.gradle
```diff
@@ -24,51 +26,104 @@ dependencies {
...
+        "name": "libreoffice",
+        "url": "https://github.com/LibreOffice/dictionaries",
+        "ref": "762abe74008b94b2ff06db6f4024b59a8254c467" // head: master
+    ],
+    [
+        "name": "woorm",
+        "url": "https://github.com/wooorm/dictionaries",
+        "ref": "8cfea406b505e4d7df52d5a19bce525df98c54ab" // head: main
+    ]
+]
```
dweiss, line 53: This will need to be updated by hand, occasionally. The frequency of changes in those repositories isn't that great so no problem there, I think?

lucene/core/src/java/module-info.java
```diff
@@ -63,6 +63,14 @@
...
   opens org.apache.lucene.document to
       org.apache.lucene.test_framework;
+  opens org.apache.lucene.util.fst to
+      org.apache.lucene.test_framework;
+  opens org.apache.lucene.store to
+      org.apache.lucene.test_framework;
+  opens org.apache.lucene.util.automaton to
+      org.apache.lucene.test_framework;
+  opens org.apache.lucene.util to
+      org.apache.lucene.test_framework;
```
dweiss, line 73: Had to add these to make ram testing happy now that the tests run in module mode.

lucene/analysis/common/build.gradle
rmuir: yes, honestly any revision will work for the intended purpose of preventing regressions. If we want to discover new and interesting ways that dictionaries can be malformed in the wild that we should try to detect, that is different :)

rmuir (approved):
This is great, thank you so much. I'll keep watch on the new weekly builds and meanwhile we'll prevent regressions in PRs without noise!

dweiss:
Thanks. I'll backport to 10x as well.

PR #14251, opened 2025-02-18
Improve user-facing docs.

jpountz:
This improves user-facing docs in Lucene's package javadocs:
 - Make some docs up-to-date, e.g. some of them were still referring to oal.index.Fields.
 - More emphasis of structured search and knn search.
 - More guidance on how to index data in `oal.document`.
 - Add API docs for impacts.
 - Discourage less strongly to change the similarity at search time.
 - Recommend using Query factory methods on Field classes instead of creating queries manually.


rmuir (approved):
wonderful improvement, thank you for modernizing this!

PR #14256, opened 2025-02-18
Reuse entry point scores and provide mechanisms to provide scores for directly entry points

benwtrent:
Spinning out of: https://github.com/apache/lucene/pull/14226

That particular evolution of kNN querying is attempting to re-entry individual segment graphs with new exit and search criteria. To prevent having to rescore the new entry points, this PR provides the ability to (optionally) keep track of the scores for entry points. 

Additionally, this will take advantage of entry point score retention during graph building and searching. The performance improvements are marginal.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/knn/MappedDISI.java
+++ b/lucene/core/src/java/org/apache/lucene/search/knn/MappedDISI.java
@@ -1,0 +1,102 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.search.knn;
+
+import java.io.IOException;
+import org.apache.lucene.index.KnnVectorValues;
+import org.apache.lucene.search.DocIdSetIterator;
+
+/**
+ * A {@link DocIdSetIterator} that maps the doc ids from a {@link TopDocsDISI} to the corresponding
+ * vector ordinal
+ *
+ * @lucene.internal
+ */
+public sealed class MappedDISI extends DocIdSetIterator {
+  KnnVectorValues.DocIndexIterator indexedDISI;
+  DocIdSetIterator sourceDISI;
+
+  /**
+   * Create a new instance of a {@link MappedDISI}
+   *
+   * @param indexedDISI the mapping from doc to vector ordinal iterator
+   * @param sourceDISI the source iterator
+   * @return a new {@link MappedDISI}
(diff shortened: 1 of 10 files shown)


msokolov:
Hi, @benwtrent , this bakes support for this supplying scores feature pretty deeply. I was thinking if we were to use this only for the SeededKnnVectorQuery, it might suffice to create a wrapping `RandomVectorScorer ` that would supply the cached scores, while delegating the others to the underlying scorer?

lucene/core/src/java/org/apache/lucene/search/knn/MappedDISI.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.lucene.index.KnnVectorValues;
+import org.apache.lucene.search.DocIdSetIterator;
+
+/**
+ * A {@link DocIdSetIterator} that maps the doc ids from a {@link TopDocsDISI} to the corresponding
+ * vector ordinal
+ *
+ * @lucene.internal
+ */
+public sealed class MappedDISI extends DocIdSetIterator {
```
msokolov, line 29: whoa I had to look up what a sealed class is - I guess it is like final but you can allow some classes to inherit?

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
```diff
@@ -229,25 +230,32 @@ public void addGraphNode(int node, UpdateableRandomVectorScorer scorer) throws I
...
         int level = i + lowestUnsetLevel;
         candidates.clear();
-        graphSearcher.searchLevel(candidates, scorer, level, eps, hnsw, null);
-        eps = candidates.popUntilNearestKNodes();
+        graphSearcher.searchLevel(candidates, scorer, level, eps, epsScores, epCount, hnsw, null);
+        candidates.popUntilNearestKNodes();
+        eps = ArrayUtil.growExact(eps, candidates.size());
+        epsScores = ArrayUtil.growExact(epsScores, candidates.size());
+        epCount = candidates.size();
+        candidates.collectNodesAndScores(eps, epsScores);
```
msokolov, line 258: you are re-using the scores from upper levels when scoring the lower levels?

msokolov:
I also tried something based on the simpler approach I mentioned and also saw very minor gains in the seeded search with reentry when reusing scores.

lucene/core/src/java/org/apache/lucene/search/knn/MappedDISI.java
benwtrent: correct

benwtrent:
> I also tried something based on the simpler approach I mentioned and also saw very minor gains in the seeded search with reentry when reusing scores.

Yeah, I don't expect this to be a huge performance gain. I would expect the actual bottlenecks addressed in your changes (thrashing, over exploration, etc.) would dominate any savings.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
benwtrent: Yes, we were recalculating entry point scores every time we add a node to a layer.

benwtrent:
>  it might suffice to create a wrapping RandomVectorScorer  that would supply the cached scores, while delegating the others to the underlying scorer?

🤔 Maybe we could wrap the scorer at the vector format level. But an iterator/score provider will need to still be given via the collector so that the format even knows that these things are providing valid scores.

This PR is certainly making a ton of changes. I can split out some of the work, focusing at query time, instead of fully extending the HNSW builder to take advantage of the changes.

benwtrent:
This proved not particularly useful. Maybe it can be a future optimization, but for now, it seems the added complexity isn't worth its cost.

PR #14258, opened 2025-02-19
[Unit] Increase Dynamic Range Faceting coverage and address edge cases

houserjohn:
### Summary

This is a continuation of [GH#14238](https://github.com/apache/lucene/pull/14238) which was reverted due to a bug discovered during a random test with seed 50D292E371306B1. The bug was that the randomized test had randomly chosen `topN=0`. This caused all the checks to fail because no ranges were returned. This has been addressed in this PR by returning early if `topN=0`.

This PR has been created to address that bug and others. This PR contains:

- A new test that previously failed [GH#13914](https://github.com/apache/lucene/pull/13914):`SameWeightsOutOfOrder`.
- A new test for `topN=0`: `ZeroTopN`
- A new check for `topN=0` for `assertDynamicNumericRangeValidProperties`

This is in addition to the original PR which:

- Adds tests for varying TopN values
- Adds test for inputs with the same weights
- Adds tests that randomly shuffle the inputs
- Adds test for a single misplaced value
- Adds randomized test that is checked based on the properties of the ranges

### Testing

```
./gradlew clean
./gradlew tidy
./gradlew build
```

Additionally, I confirmed that this PR fixed the bug with seed 50D292E371306B1:
```
./gradlew test -Dtests.seed=50D292E371306B1 -Dtests.locale=fr-MG -Dtests.timezone=Asia/Almaty -Dtests.asserts=true -Dtests.file.encoding=UTF-8 
```

Also, to ensure that it is unlikely we will have to revert this PR, I have run the randomized tests contained in this PR over 100 times and confirmed that none of them fail.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -88,6 +88,7 @@ Improvements
 * GITHUB#14239: Hunspell's option to tolerate affix rule count mismatches was
   improved to tolerate more instances of this problem.  (Robert Muir)
 
+* GITHUB#14238: Improve test coverage of Dynamic Range Faceting. (John Houser)
 
 Optimizations
 ---------------------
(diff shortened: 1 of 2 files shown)


stefanvodita (approved):
Thanks for the quick fix!

gsmiller (approved):
Thanks for the quick fix (randomized testing is tricky)! This looks good to me.

PR #14262, opened 2025-02-19
Add support for querying multiple fields to QueryBuilder.

jpountz:
`QueryBuilder#createBooleanQuery` now supports querying multiple fields. Scoring-wise, it supports two options: either BM25F, the best choice but requires fields to be indexed/searched with the same analyzer, index options and similarity, or taking the max score across fields, a poor option but that doesn't make any assumption about how fields are indexed/searched.

To make it work properly, I added support for multiple terms back to `CombinedFieldQuery` (like in sandbox) to support synonyms.


jpountz:
After discussing with @rmuir on options when fields don't have the same analyzer, index options, etc. I'm leaning towards only supporting BM25F for now. There is simply no good way of scoring on multiple fields when they are indexed differently.

PR #14265, opened 2025-02-20
Reduce knn recall test flakiness

benwtrent:
I have noticed some additional flakiness for knn recall. 

For example:

```
./gradlew test --tests TestPerFieldKnnVectorsFormat.testRecall -Dtests.seed=FAEFE5196FDED25B -Dtests.locale=twq-Latn-NE -Dtests.timezone=America/New_York -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```

So, I am making two minor adjustments to the test:

 - Increasing hnsw efsearch to 25 (still only gathering 10 nearest docs in the end)
 - Better filling vectors for tiny lines

I also added some nicer `toString` methods and output for the test, so if we notice it failing again, we can quickly see for what codec and options.


msokolov (approved):
thanks, @benwtrent !

PR #13599, opened 2024-07-22
Delegating the matches in PointRangeQuery weight to relate method

jainankitk:
### Description
Delegating the matches in PointRangeQuery weight to relate method

### Issue
Resolves #13598

PR #14267, opened 2025-02-21
Reduce the number of comparisons when lowerPoint is equal to upperPoint

hanbj:
### Description
When lowerPoint is equal to upperPoint. In fact, there is no need to compare lowerPoint and upperPoint at the same time. The number of comparisons can be reduced by half when collecting document ids to construct bitsets and deeply traversing the bkd tree.
During my reading of Elasticsearch related code, I found that when executing term or terms queries on the date field, the query is rewritten, and a single term is rewritten as a rang query (lowerTerm is the same as upperTerm). The terms query uses BooleanQuery to encapsulate multiple range queries. Therefore, it is more suitable for this scenario, reducing the number of comparisons and improving performance when collecting a large number of documents.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
@@ -132,416 +132,447 @@ public abstract class PointRangeQuery extends Query {
   public final Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost)
       throws IOException {
 
+    if (this.equalValues) { // lowerPoint==upperPoint
+      return new SinglePointConstantScoreWeight(this, scoreMode, boost);
+    }
     // We don't use RandomAccessWeight here: it's no good to approximate with "match all docs".
     // This is an inverted structure and should be used in the first pass:
+    return new MultiPointsConstantScoreWeight(this, scoreMode, boost);
+  }
 
-    return new ConstantScoreWeight(this, boost) {
+  /**
+   * Essentially, it is to reduce the number of comparisons. This is an optimization, used for the
+   * case of lowerPoint==upperPoint.
+   */
+  protected class SinglePointConstantScoreWeight extends MultiPointsConstantScoreWeight {
 
-      private final ByteArrayComparator comparator = ArrayUtil.getUnsignedComparator(bytesPerDim);
+    public SinglePointConstantScoreWeight(Query query, ScoreMode scoreMode, float boost) {
+      super(query, scoreMode, boost);
+    }
 
-      private boolean matches(byte[] packedValue) {
-        int offset = 0;
-
-        if (equalValues) {
-          for (int dim = 0; dim < numDims; dim++, offset += bytesPerDim) {
-            if (comparator.compare(packedValue, offset, lowerPoint, offset) != 0) {
-              return false;
-            }
-          }
-          return true;
-        }
-
-        for (int dim = 0; dim < numDims; dim++, offset += bytesPerDim) {
-          if (comparator.compare(packedValue, offset, lowerPoint, offset) < 0) {
-            // Doc's value is too low, in this dimension
-            return false;
-          }
-          if (comparator.compare(packedValue, offset, upperPoint, offset) > 0) {
-            // Doc's value is too high, in this dimension
-            return false;
-          }
-        }
-        return true;
-      }
-
-      private Relation relate(byte[] minPackedValue, byte[] maxPackedValue) {
-
-        boolean crosses = false;
-        int offset = 0;
-
-        if (equalValues) {
-          for (int dim = 0; dim < numDims; dim++, offset += bytesPerDim) {
-
-            int cmpMin = comparator.compare(minPackedValue, offset, lowerPoint, offset);
-            if (cmpMin > 0) {
-              return Relation.CELL_OUTSIDE_QUERY;
-            }
-
-            int cmpMax = comparator.compare(maxPackedValue, offset, lowerPoint, offset);
-            if (cmpMax < 0) {
-              return Relation.CELL_OUTSIDE_QUERY;
-            }
-
-            if (cmpMin != 0 || cmpMax != 0) {
-              crosses = true;
-            }
-          }
-
-          if (crosses) {
-            return Relation.CELL_CROSSES_QUERY;
-          } else {
-            return Relation.CELL_INSIDE_QUERY;
-          }
-        }
-
-        for (int dim = 0; dim < numDims; dim++, offset += bytesPerDim) {
-
-          if (comparator.compare(minPackedValue, offset, upperPoint, offset) > 0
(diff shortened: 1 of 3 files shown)


stefanvodita (commented):
Sorry it's taken a while to review. Have you checked whether the [failing test case](https://github.com/apache/lucene/actions/runs/13450612987/job/37584306320?pr=14267) is related? Do we have any performance tests?

Edit: Retry succeeded, but it's still worth checking that previous run. I haven't had a look at it yet.

hanbj:
The previous failed test case was org.apache.Lucene.index TestKnnGraph.testMultiThreadedSearch.
I have confirmed the testMultiThreadedSearch method, which uses KnnFloatVectorQuery for search and does not use PointRangeQuery, so it is not relevant to this range query.

jainankitk (commented):
Thanks @hanbj for creating this PR. Can you make suggested refactoring to avoid any potential regression for the `lowerPoint != upperPoint` code path, which is also more common? Can you also add unit test for this new code path?

lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
```diff
@@ -129,6 +141,16 @@ public final Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, fl
...
         int offset = 0;
+
+        if (equalValues) {
+          for (int dim = 0; dim < numDims; dim++, offset += bytesPerDim) {
+            if (comparator.compare(packedValue, offset, lowerPoint, offset) != 0) {
+              return false;
+            }
+          }
+          return true;
+        }
```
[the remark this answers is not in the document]
hanbj: This implementation will result in a lot of code duplication. I will implement it first. Thank you for your hard work in the review.

lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
```diff
@@ -120,381 +132,447 @@ public void visit(QueryVisitor visitor) {
...
     // This is an inverted structure and should be used in the first pass:
+    return new MultiPointsConstantScoreWeight(this, scoreMode, boost);
+  }
 
-    return new ConstantScoreWeight(this, boost) {
+  /**
+   * Essentially, it is to reduce the number of comparisons. This is an optimization, used for the
+   * case of lowerPoint==upperPoint.
+   */
+  protected class SinglePointConstantScoreWeight extends MultiPointsConstantScoreWeight {
```
jainankitk, line 147: I am assuming we are reusing some of the methods from `MultiPointsConstantScoreWeight`. That's why we are extending from that class. May, I suggest creating class say `PointRangeQueryWeight` that extends from `ConstantScoreWeight`? And, both `SinglePointRangeQueryWeight` and `MultiPointRangeQueryWeight` extend from `PointRangeQueryWeight`?

jainankitk (commented):
Thanks @hanbj for refactoring the code and adding the unit test. Looks much better now. Just a minor suggestion to improve the readability further.

lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
hanbj: This suggestion is great, SinglePointRangeQueryWeight and MultiplaPointRangeQueryWeight only need to implement their own point value matching logic and relationship judgment.

jainankitk (approved):

jainankitk:
@hanbj - Thanks for patiently addressing the review comments. While I don't see any performance regression risk myself, I am wondering if we can do one quick performance benchmark run, just to ensure we are not missing anything obvious?

lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
```diff
@@ -517,6 +623,11 @@ public byte[] getUpperPoint() {
     return upperPoint.clone();
   }
 
+  // for test
+  public boolean isEqualValues() {
```
gsmiller, line 627: I'd really prefer we do not increase our public API surface area only for testing on a class like this. Can we find another way to test without exposing this please?

gsmiller (commented):
Thanks for proposing this! I didn't review in deep detail but commented on a couple API/visibility related concerns that jumped out to me.

Also, +1 to running some benchmarks on this before merging to ensure we're not regressing current behavior. Adding this polymorphic indirection may actually hurt performance in interesting, non-obvious ways and we should verify this is actually beneficial.

As an alternative direction, I'd also be curious how `PointInSetQuery` with a single point performs. A really simple thing to try would be to change the query factory methods (e.g., `LongPoint#newExactQuery`) to build a set query instead of range if it's better. Or another option could be to create a specialized query that does an exact point comparison. Benchmarks would be a great help in figuring out what of these approaches is best.

Thanks again!

lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
jainankitk: Good catch @gsmiller! Can we make this package-private?

lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
```diff
@@ -120,381 +132,475 @@ public void visit(QueryVisitor visitor) {
...
 
-    return new ConstantScoreWeight(this, boost) {
+  /**
+   * Single-point range query weight implementation class, used to handle the special case where the
+   * lower and upper bounds are equal (i.e. single-point query).
+   *
+   * <p>Optimize query performance by reducing the number of comparisons between dimensions. This
+   * implementation is used when the upper and lower bounds of all dimensions are exactly the same.
+   */
+  protected class SinglePointRangeQueryWeight extends PointRangeQueryWeight {
```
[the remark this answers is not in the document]
jainankitk: Yeah, they can be `private` in my opinion as well.

hanbj:
@gsmiller @jainankitk I haven't carefully studied the implementation of benchmark sin Lucene too, which may take sometime.

jainankitk:
> @gsmiller @jainankitk I haven't carefully studied the implementation of benchmark sin Lucene too, which may take sometime.

@hanbj - Please take your time. You can follow the instructions here - https://github.com/mikemccand/luceneutil/blob/main/README.md. It is fairly straightforward, did it recently for [another PR](https://github.com/apache/lucene/pull/13599), coincidentally in `PointRangeQuery` only. Please let me know if you get stuck anywhere

jainankitk:
Closing in favor of linked PR #14625 that addresses review comments with performance benchmark results

lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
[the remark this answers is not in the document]
jainankitk: While `MultiPointRangeQueryWeight` and `SinglePointRangeQueryWeight` are used for asserting the weight instance type, `PointRangeQueryWeight` can be made private

PR #14268, opened 2025-02-21
PointInSetQuery early exit on non-matching segments

hanbj:
### Description
When creating a PointInSetQuery object, the data in the packedPoints parameter is returned in order, so the maximum and minimum values ​​can be determined when iterating over packedPoints.

With the maximum and minimum values, the query can be returned early in special cases during the serial search of the segment.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -67,6 +67,8 @@ Optimizations
 ---------------------
 * GITHUB#14418: Quick exit on filter query matching no docs when rewriting knn query. (Pan Guixin)
 
+* GITHUB#14268: PointInSetQuery early exit on non-matching segments. (hanbj)
+
 Bug Fixes
 ---------------------
 (No changes)
--- a/lucene/core/src/java/org/apache/lucene/search/PointInSetQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/PointInSetQuery.java
@@ -111,7 +111,8 @@ public abstract class PointInSetQuery extends Query implements Accountable {
       if (previous == null) {
         previous = new BytesRefBuilder();
         lowerPoint = new byte[bytesPerDim * numDims];
-        System.arraycopy(current.bytes, current.offset, lowerPoint, 0, lowerPoint.length);
+        assert lowerPoint.length == current.length;
+        System.arraycopy(current.bytes, current.offset, lowerPoint, 0, current.length);
       } else {
         int cmp = previous.get().compareTo(current);
         if (cmp == 0) {
@@ -129,7 +130,8 @@ public abstract class PointInSetQuery extends Query implements Accountable {
     if (previous != null) {
       BytesRef max = previous.get();
       upperPoint = new byte[bytesPerDim * numDims];
-      System.arraycopy(max.bytes, max.offset, upperPoint, 0, upperPoint.length);
+      assert upperPoint.length == max.length;
+      System.arraycopy(max.bytes, max.offset, upperPoint, 0, max.length);
     }
     ramBytesUsed =
         BASE_RAM_BYTES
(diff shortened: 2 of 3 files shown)


stefanvodita (commented):
Thank you for your contribution @hanbj! Would it make sense to add unit tests that exercise these new code paths and serve as examples for the type of situation you want to capture? Also, if I understand correctly, this is meant as an optimisation, so performance tests that show an improvement would be great!

hanbj:
@stefanvodita Thank you for the review. Unit testing has been added

lucene/core/src/java/org/apache/lucene/search/PointInSetQuery.java
```diff
@@ -108,6 +110,8 @@ protected PointInSetQuery(String field, int numDims, int bytesPerDim, Stream pac
       }
       if (previous == null) {
         previous = new BytesRefBuilder();
+        lowerPoint = new byte[bytesPerDim * numDims];
+        System.arraycopy(current.bytes, current.offset, lowerPoint, 0, lowerPoint.length);
```
gsmiller, line 114: minor: I think it'd be slightly more readable if you used `current.length` here instead of ` lowerPoint.length` (and I might also throw an `assert lowerPoint.length == current.length` immediately before this line to make it clear they should be equal).

lucene/core/src/java/org/apache/lucene/search/PointInSetQuery.java
```diff
@@ -122,6 +126,11 @@ protected PointInSetQuery(String field, int numDims, int bytesPerDim, Stream pac
     }
     sortedPackedPoints = builder.finish();
     sortedPackedPointsHashCode = sortedPackedPoints.hashCode();
+    if (previous != null) {
+      BytesRef max = previous.get();
+      upperPoint = new byte[bytesPerDim * numDims];
+      System.arraycopy(max.bytes, max.offset, upperPoint, 0, upperPoint.length);
```
gsmiller, line 132: minor: same comment here about using `previous.length` in place of `upperPoint.length`. I don't feel strongly about this so please feel free to disagree if you have a different perspective.

lucene/core/src/java/org/apache/lucene/search/PointInSetQuery.java
```diff
@@ -153,6 +162,21 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
           return null;
         }
 
+        if (values.getDocCount() == 0) {
+          return null;
+        } else if (lowerPoint != null && upperPoint != null) {
```
gsmiller, line 167: Should it be true that either 1) both of these are null, or 2) both are non-null? I think so right? If that's right, I would check `lowerPoint != null` then put an `assert upperPoint != null` in the condition branch.

lucene/core/src/java/org/apache/lucene/search/PointInSetQuery.java
```diff
@@ -153,6 +162,21 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
           return null;
         }
 
+        if (values.getDocCount() == 0) {
```
gsmiller, line 165: I'm not 100% sure of this but I don't think it's possible to get back a non-null instance from `reader#getPointValues` that has a zero doc count. I believe you'll always get back a null instance if the points field has no docs in a segment. Can you confirm with a test and/or some debugging?

gsmiller (commented):
This optimization makes sense to me. I left you a few small comments. Thanks for suggesting this change!

lucene/core/src/java/org/apache/lucene/search/PointInSetQuery.java
hanbj: You're right, usually the length of the copied array is used.

lucene/core/src/java/org/apache/lucene/search/PointInSetQuery.java
hanbj: You're right, usually the length of the copied array is used. I used the length of max here.

lucene/core/src/java/org/apache/lucene/search/PointInSetQuery.java
hanbj: This has been modified.

lucene/core/src/java/org/apache/lucene/search/PointInSetQuery.java
hanbj: I am referring to the implementation in PointRangeQuery here.

gsmiller (commented):
Thanks for the iteration!

lucene/CHANGES.txt
```diff
@@ -186,6 +186,8 @@ Optimizations
 
 * GITHUB#14272: Use DocIdSetIterator#range for continuous-id BKD leaves. (Guo Feng)
 
+* GITHUB#14268: PointInSetQuery clips segments by lower and upper (hanbj)
```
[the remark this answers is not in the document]
hanbj: This has been modified.

gsmiller:
This looks great! Taking care of the merge now. Thank you @hanbj !

PR #14270, opened 2025-02-21
Address completion fields testing gap and truly allow loading FST off heap

javanna:
We currently have a gap in testing the completion postings format. While it allows to load FSTs off heap, and the format has a constructor that takes the fst load mode, at read time SPI goes through the default constructor that has fst load mode hardcoded to on heap at all times. This means that despite we randomize fst load mode in `TestSuggestField`, we never effectively load the completion FST off heap, we only set the load mode to something different than on heap when writing, where the load mode takes no effect.

This commit introduces a static method that allows to switch the loading of the FST for all the completion postings format.

This unveiled a bug when loading FST off heap and the index input is a `MemorySegmentIndexInput` which is fixed by #14271

Besides the testing concerns, an important question is: what are users supposed to do to enable off heap loading of completion fields? Is the intention that they need to register custom postings format, although it's only the loading part that needs overriding? Or do we want to make that easier somehow reusing the existing postings format? I think it is rather odd to require users to customize the postings format, and have to support their own format for the time being, if the write side is identical to the standard one.

### DIFF
--- a/lucene/suggest/src/java/org/apache/lucene/search/suggest/document/CompletionPostingsFormat.java
+++ b/lucene/suggest/src/java/org/apache/lucene/search/suggest/document/CompletionPostingsFormat.java
@@ -104,13 +104,12 @@ public abstract class CompletionPostingsFormat extends PostingsFormat {
   static final String INDEX_EXTENSION = "cmp";
   static final String DICT_EXTENSION = "lkp";
 
-  /**
-   * Default way of loading FSTs when reading a completion field
-   */
+  /** Default way of loading FSTs when reading a completion field */
   public static final FSTLoadMode DEFAULT_FST_LOAD_MODE = FSTLoadMode.ON_HEAP;
+
   /**
-   * The {@link FSTLoadMode} that drives how {@link CompletionPostingsFormat} subclasses load their FSTs,
-   * used whenever a completion field is read.
+   * The {@link FSTLoadMode} that drives how {@link CompletionPostingsFormat} subclasses load their
+   * FSTs, used whenever a completion field is read.
    */
   public static FSTLoadMode FST_LOAD_MODE = DEFAULT_FST_LOAD_MODE;
 
--- a/lucene/suggest/src/java/org/apache/lucene/search/suggest/document/CompletionPostingsFormat.java
+++ b/lucene/suggest/src/java/org/apache/lucene/search/suggest/document/CompletionPostingsFormat.java
@@ -104,6 +104,9 @@ public abstract class CompletionPostingsFormat extends PostingsFormat {
   static final String INDEX_EXTENSION = "cmp";
   static final String DICT_EXTENSION = "lkp";
 
+  public static final FSTLoadMode DEFAULT_FST_LOAD_MODE = FSTLoadMode.ON_HEAP;
+  public static FSTLoadMode FST_LOAD_MODE = DEFAULT_FST_LOAD_MODE;
+
   /** An enum that allows to control if suggester FSTs are loaded into memory or read off-heap */
   public enum FSTLoadMode {
     /**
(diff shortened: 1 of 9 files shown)


lucene/suggest/src/java/org/apache/lucene/search/suggest/document/CompletionPostingsFormat.java
```diff
@@ -104,6 +104,9 @@ public abstract class CompletionPostingsFormat extends PostingsFormat {
   static final String INDEX_EXTENSION = "cmp";
   static final String DICT_EXTENSION = "lkp";
 
+  public static final FSTLoadMode DEFAULT_FST_LOAD_MODE = FSTLoadMode.ON_HEAP;
+  public static FSTLoadMode FST_LOAD_MODE = DEFAULT_FST_LOAD_MODE;
```
javanna, line 108: A mutable static field is certainly not elegant. I am out of ideas of how we can allow to swap reading between on_heap and off_heap without requiring a specific postings format for each load mode, which would be odd given loading is merely a read aspect and does not affect the writing side in any way. Having to rely on the default constructor is quite limiting. On the other hand this is an improvement compared to what we had until now.

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestSuggestField.java
```diff
@@ -949,10 +952,7 @@ static IndexWriterConfig iwcWithSuggestField(Analyzer analyzer, final Set<String
     iwc.setMergePolicy(newLogMergePolicy());
     Codec filterCodec =
         new FilterCodec(TestUtil.getDefaultCodec().getName(), TestUtil.getDefaultCodec()) {
-          final CompletionPostingsFormat.FSTLoadMode fstLoadMode =
-              RandomPicks.randomFrom(random(), CompletionPostingsFormat.FSTLoadMode.values());
-          final PostingsFormat postingsFormat = new Completion101PostingsFormat(fstLoadMode);
```
javanna, line 954: While this looked like randomizing the load mode, it did not as SPI would not go through this constructor. Setting fst load mode this way would take no effect.

jpountz:
Do we need to let users configure whether to load their completion fields on-heap or off-heap? Our inverted indexes, doc values, etc. don't give the option - if users want to load data on-heap, they need to create their own codec. Could it be the same with `CompletionPostingsFormat`? Plus who wants to load GBs of data on heap?

javanna:
I think that we do we need to have a way to truly make the switch at least for testing purposes, otherwise we only test one approach which is error prone. That's the original issue that this PR addresses

From a users perspective, I don't have data at hand to show that off-heap is as performant as on heap. Given how difficult it's been to use off heap completion fields, I'd assume that everyone always relied on on-heap, as that has been the default since day one years ago. I think exposing a way to make the switch would be handy.

I personally find it not so user-friendly to require users to define their own codec only for loading purposes, when there is no effect on how stuff is written. Is that a common thing users may do in other scenarios like inverted indices, doc values etc. ?

If you are suggesting switching to off-heap by default, I am +1 for that. 

FYI, another problem we have around providing custom completion postings formats is going to be addressed by #14275, so that existing SPI services can be predictably replaced. Once that is fixed, either changing the default or requiring users to provide their own postings format to override the default behaviour is probably good enough. In that case I would leave the static handling I am adding in this PR only for testing purposes.

jpountz:
> I personally find it not so user-friendly to require users to define their own codec only for loading purposes, when there is no effect on how stuff is written. Is that a common thing users may do in other scenarios like inverted indices, doc values etc. ?

The way I'm seeing it: for other data structures like inverted indexes and doc values, Lucene takes full ownership of what should live on-heap (terms index, doc values metadata, etc.) and what should live off-heap (terms, postings, etc.). Users don't have a way to configure it. I would like `completion` to be the same: Lucene makes a decision of whether this FST should be on-heap or off-heap once for all, and removes the ability to tune whether to load it on-heap or off-heap.

javanna:
That's good with me. Shall we clearly document this then as a follow-up, and shall I make the fst load mode static field package private perhaps, so that we at least test both modes?

javanna:
I opened #14364 to still address the testing gap, but also change the default load mode to off heap.

PR #14272, opened 2025-02-21
Use DocIdSetIterator#range for continuous-id BKD leaves

gf2121:
Use`DocIdSetIterator#range` instead of `DocBaseBitsetIterator` as `DocIdSetIterator#range` implemented `#intoBitset`.


jpountz (approved):
That makes sense to me.

PR #12415, opened 2023-07-05
Optimize disjunction counts.

jpountz:
This introduces `LeafCollector#collect(DocIdStream)` to enable collectors to collect batches of doc IDs at once. `BooleanScorer` takes advantage of this by creating a `DocIdStream` whose `count()` method counts the number of bits that are set in the bit set of matches in the current window, instead of naively iterating over all matches.

On wikimedium10m, this yields a ~20% speedup when counting hits for the `title OR 12` query (2.9M hits).

Relates #12358

PR #14273, opened 2025-02-21
Speed up histogram collection in a similar way as disjunction counts.

jpountz:
This generalizes the `IndexSearcher#count` optimization from PR #12415 to histogram facets by introducing specialization for counting the number of matching docs in a range of doc IDs.

Currently, disjunctions and dense conjunctions both internally collect `DocIdStream`s backed by a bitset. In the future, we could make more queries collect whole `DocIdStream`s at once to speed up collection, e.g. `MatchAllDocsQuery` or doc-value range queries that take advantage of a sparse index.


jpountz:
@epotyom You may be interested in taking a look.

epotyom (commented):
I like the idea! Looks like we can do similar trick for range facets and long values facets?

gsmiller:
+1 to this optimization. Love the idea!

jpountz:
Quick update: we now have more queries that collect hits using `collect(DocIdStream)`, which makes this optimization more appealing.

lucene/core/src/java/org/apache/lucene/search/DocIdStream.java
```diff
@@ -34,12 +33,35 @@ protected DocIdStream() {}
...
     int[] count = new int[1];
     forEach(doc -> count[0]++);
     return count[0];
   }
+
+  /**
+   * Count the number of doc IDs in this stream that are below the given {@code upTo}. These doc IDs
+   * may not be consumed again later.
+   */
+  public int count(int upTo) throws IOException {
```
gsmiller, line 59: This only becomes an optimization if we specialize this method right? The specializations I'm aware of rely on `FixedBitSet#cardinality`. Are you thinking of peeking into these bit sets to provide cardinality up to the specific doc? (Or maybe I'm missing something?)

jpountz:
> If we have a skipper, I think we ought to also be able to use competitive iterators to jump over blocks of docs we know we won't collect based on their values?

This is correct. I plan on doing something similar when sorting: it is safe to skip blocks whose values all compare worse than the current k-th value. It's similar to what block-max WAND/MAXSCORE do: when a block's best possible score is less than the k-th best score so far, it can safely be skipped.

jpountz:
It should be ready for review now. Now that `DocIdStream` has become more sophisticated, I extracted impls to proper classes that could be better tested. This causes some diffs in our boolean scorers, hence the high number of lines changed.

jpountz:
I'll try to run some simple benchmarks next.

jpountz:
I played with the geonames dataset, by filtering out docs that don't have a value for the `elevation` field (2.3M docs left), enabling index sorting on the `elevation` field and computing histograms on the `elevation` field with a bucket width of 100.

| Query |  | Latency on main (ms) | Latency on branch (ms) |
| - | - | - | - |
| `MatchAllDocsQuery` | Uses `RangeDocIdStream` under the hood | 6.9 | 4.3 |
| `featureClass:(S P)` | Matches spots or cities, 1.2 matching docs, uses `BitSetDocIdStream` under the hood | 4.8 | 2.4 |

I also checked wikibigall, no slowdowns were detected.

lucene/core/src/java/org/apache/lucene/search/BooleanScorer.java
```diff
@@ -207,8 +164,32 @@ private void scoreWindowIntoBitSetAndReplay(
       acceptDocs.applyMask(matching, base);
     }
 
-    docIdStreamView.base = base;
-    collector.collect(docIdStreamView);
+    if (buckets == null) {
+      if (acceptDocs != null) {
+        // In this case, live docs have not been applied yet.
+        acceptDocs.applyMask(matching, base);
+      }
```
gsmiller, line 171: If I'm looking at this diff properly, I don't think you should need this block of code at all since you're still applying `acceptDocs` immediately prior?

lucene/core/src/java/org/apache/lucene/search/BitSetDocIdStream.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @Override
+  public boolean mayHaveRemaining() {
+    return upTo < max;
+  }
+
+  @Override
+  public void forEach(int upTo, CheckedIntConsumer<IOException> consumer) throws IOException {
+    if (upTo >= this.upTo) {
```
gsmiller, line 42: I think you want `>` here instead? (`>=` still functionally works since `forEach` is tolerant of the `==` case, but I think you want to short circuit if `upTo == this.upTo`?)

(Also possible I'm getting this wrong... but I did make the changes locally and all the testing still seems to pass)

lucene/core/src/java/org/apache/lucene/search/BitSetDocIdStream.java
```diff
@@ -0,0 +1,60 @@
...
+      upTo = Math.min(upTo, max);
+      bitSet.forEach(this.upTo - offset, upTo - offset, offset, consumer);
+      this.upTo = upTo;
+    }
+  }
+
+  @Override
+  public int count(int upTo) throws IOException {
+    if (upTo >= this.upTo) {
```
gsmiller, line 51: Another place where I think you want `>`

lucene/core/src/java/org/apache/lucene/search/RangeDocIdStream.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  @Override
+  public boolean mayHaveRemaining() {
+    return upTo < max;
+  }
+
+  @Override
+  public void forEach(int upTo, CheckedIntConsumer<IOException> consumer) throws IOException {
+    if (upTo >= this.upTo) {
```
gsmiller, line 41: Another place where I think you want `>`

lucene/core/src/java/org/apache/lucene/search/RangeDocIdStream.java
```diff
@@ -0,0 +1,61 @@
...
+      for (int doc = this.upTo; doc < upTo; ++doc) {
+        consumer.accept(doc);
+      }
+      this.upTo = upTo;
+    }
+  }
+
+  @Override
+  public int count(int upTo) throws IOException {
+    if (upTo >= this.upTo) {
```
gsmiller, line 52: And one more spot where I think you want `>`?

gsmiller (commented):
Left some minor feedback. Looks good though!

lucene/core/src/java/org/apache/lucene/search/DISIDocIdStream.java
```diff
@@ -0,0 +1,68 @@
...
+import java.io.IOException;
+import org.apache.lucene.util.FixedBitSet;
+
+final class DISIDocIdStream extends DocIdStream {
+
+  private final DocIdSetIterator iterator;
+  private final int to;
+  private final FixedBitSet spare;
+
+  DISIDocIdStream(DocIdSetIterator iterator, int to, FixedBitSet spare) {
```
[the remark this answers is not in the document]
jpountz: I'm leaning towards waiting until there's a need for it before adding it.

lucene/core/src/java/org/apache/lucene/search/BitSetDocIdStream.java
jpountz: You're right, I'll change this.

lucene/core/src/java/org/apache/lucene/search/BooleanScorer.java
jpountz: Thanks for catching, this bit of code was mistakenly duplicated.

lucene/core/src/java/org/apache/lucene/search/BitSetDocIdStream.java
jpountz: (For reference, it's not as much for short-circuiting as for the fact that logic under the if block would fail if trying to move backwards. But since we're doing a check anyway, I agree it's also worth excluding the trivial case when the range of docs to collect is empty.)

lucene/core/src/java/org/apache/lucene/util/FixedBitSet.java
```diff
@@ -204,6 +205,40 @@ public int cardinality() {
...
+      assert (from & 0x3F) == 0;
+    }
+
+    for (int i = from >> 6, end = to >> 6; i < end; ++i) {
+      cardinality += Long.bitCount(bits[i]);
+    }
+
+    // Now handle bits between the last complete word and to
+    if ((to & 0x3F) != 0) {
+      long bits = this.bits[to >> 6] << -to;
```
gsmiller, line 235: minor: one other small comment. I noticed in the new `forEach` method you added you use `long bits = this.bits[to >> 6] & ((1L << to) - 1);`. Is the rationale here that you need to persist the correct number of trailing zeros in the `forEach` implementation but not in this case since you're doing a bit count? Is this approach (shifting by `-to`) more performant (I ask since you could use the same approach as `forEach` here too for consistency, so I assume you had a reason ;))

lucene/core/src/java/org/apache/lucene/search/DISIDocIdStream.java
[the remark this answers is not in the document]
gsmiller: I think that's fair. Thanks!

lucene/core/src/java/org/apache/lucene/util/FixedBitSet.java
gsmiller: Thanks for the detailed explanation. This makes sense to me, it just left me scratching my head for a little bit initially to figure out the intention behind the different approaches :)

gsmiller (approved):
Looks good. Thanks for incorporating the minor feedback!

PR #14277, opened 2025-02-22
Introduce DocIdStream#intoBitset to speed up cache

gf2121:
I was looking for a way to use `DocIdSetIterator#intoBitset` to speed up caching docs into `FixedBitset`. But it seems challenging to take advantage of both specialized BulkScorers and `Scorer#iterator#intoBitset` at the same time by current API. 

This PR tries to introduce `DocIdStream#intoBitset` to do so. It can help dense conjunctions and disjunctions for now, and hopefully point queries or other bitset-result queries in the future.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/DocsWithFieldSet.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocsWithFieldSet.java
@@ -72,6 +72,7 @@ public final class DocsWithFieldSet extends DocIdSet {
   }
 
   /** Return the number of documents of this set. */
+  @Override
   public int cardinality() {
     return cardinality;
   }
--- a/lucene/core/src/java/org/apache/lucene/search/BooleanScorer.java
+++ b/lucene/core/src/java/org/apache/lucene/search/BooleanScorer.java
@@ -114,6 +114,18 @@ final class BooleanScorer extends BulkScorer {
     }
 
     @Override
+    public int intoBitset(FixedBitSet bitSet) throws IOException {
+      if (buckets != null) {
+        return super.intoBitset(bitSet);
+      }
+      int length = Math.min(matching.length(), bitSet.length() - base);
+      if (length > 0) {
+        FixedBitSet.orRange(matching, 0, bitSet, base, length);
+      }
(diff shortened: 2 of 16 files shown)


jpountz:
I believe that this could help `FacetCollector` as well.

Intuitively, I had thought of `BulkScorer` as a better place for this API that would more easily help the queries that you mentioned. Can you describe the challenges that you identified?

gf2121:
Thanks for feedback!

>  I had thought of BulkScorer as a better place for this API that would more easily help the queries that you mentioned. Can you describe the challenges that you identified?

I meant it would be a challenge if no new API was added :) Adding the API to either `DocIdStream` or `BulkScorer` should work. I agree that adding it to `BulkScorer` would make things much simpler, I like it!

> I believe that this could help FacetCollector as well.

+1. I notice that `FacetCollector` use a `DocIdSetBuilder` instead of `FixedBitSet`. `LRUQueryCache` also tries to reduce memory usage by `RoaringDocIdSet`. I'm thinking about an API like `DocIdSet scoreAsDocIdSet(Bits acceptDocs, int min, int max)`, WDYT?

gf2121:
> an API like DocIdSet scoreAsDocIdSet(Bits acceptDocs)

I roughly implemented the idea and use it the range query. I'll add tests if it makes sense to you.

jpountz:
I had thought of something in-between your previous PR and this one. E.g. adding `BulkScorer#intoBitSet` with a similar signature and contract as `DocIdSetIterator#intoBitSet`.

jpountz:
To be honest, I'm a bit on the fence about introducing specialization for caching. This is why I was wondering if faceting could benefit from it too, though I'm not a fan of the fact that the facet module loads all matches into a bit set either. So I'm now wondering whether we should actually introduce specialization. (Longer-term, I'm thinking of removing the query cache #14187)

gf2121:
> (Longer-term, I'm thinking of removing the query cache https://github.com/apache/lucene/pull/14187)

Thanks for explanation, I understand the motivation that we are focusing on skipping docs instead of loading everything into RAM, that makes sense to me.

PR #14278, opened 2025-02-22
Enhance DictionaryCompoundWordTokenFilter

renatoh:
Adding option to consume characters if a matching word is found, and not used for further potential matches anymore. E.g. if the word "schwein" is extracted, the sub-word "wein" is  not extracted anymore.

### DIFF
--- a/lucene/analysis/common/src/test/org/apache/lucene/analysis/compound/TestCompoundWordTokenFilter.java
+++ b/lucene/analysis/common/src/test/org/apache/lucene/analysis/compound/TestCompoundWordTokenFilter.java
@@ -251,7 +251,8 @@ public class TestCompoundWordTokenFilter extends BaseTokenStreamTestCase {
             CompoundWordTokenFilterBase.DEFAULT_MIN_WORD_SIZE,
             CompoundWordTokenFilterBase.DEFAULT_MIN_SUBWORD_SIZE,
             CompoundWordTokenFilterBase.DEFAULT_MAX_SUBWORD_SIZE,
-            true);
+            true,
+            false);
 
     assertTokenStreamContents(
         tf,
@@ -275,6 +276,7 @@ public class TestCompoundWordTokenFilter extends BaseTokenStreamTestCase {
             CompoundWordTokenFilterBase.DEFAULT_MIN_WORD_SIZE,
             CompoundWordTokenFilterBase.DEFAULT_MIN_SUBWORD_SIZE,
             CompoundWordTokenFilterBase.DEFAULT_MAX_SUBWORD_SIZE,
+            false,
             false);
 
     assertTokenStreamContents(
@@ -297,6 +299,7 @@ public class TestCompoundWordTokenFilter extends BaseTokenStreamTestCase {
             CompoundWordTokenFilterBase.DEFAULT_MIN_WORD_SIZE,
             CompoundWordTokenFilterBase.DEFAULT_MIN_SUBWORD_SIZE,
             CompoundWordTokenFilterBase.DEFAULT_MAX_SUBWORD_SIZE,
+            false,
             false);
 
     // since "d" is shorter than the minimum subword size, it should not be added to the token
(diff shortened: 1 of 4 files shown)


rmuir:
looks good to me. I wonder about the name of the parameter, maybe "greedy" would be more intuitive as a way to describe what it is doing?

lucene/analysis/common/src/test/org/apache/lucene/analysis/compound/TestCompoundWordTokenFilter.java
```diff
@@ -682,4 +687,41 @@ protected TokenStreamComponents createComponents(String fieldName) {
...
+      Tokenizer tokenizer, String searchTerm, CharArraySet dict) {
+    tokenizer.setReader(new StringReader(searchTerm));
+    return new DictionaryCompoundWordTokenFilter(
+        tokenizer,
+        dict,
+        CompoundWordTokenFilterBase.DEFAULT_MIN_WORD_SIZE,
+        CompoundWordTokenFilterBase.DEFAULT_MIN_SUBWORD_SIZE,
+        CompoundWordTokenFilterBase.DEFAULT_MAX_SUBWORD_SIZE,
+        true,
+        true);
```
rmuir, line 725: could we add a case with:
```java
longestMatch = false;
consumeChars = true;
```

If the combination doesn't make sense, lets just throw an `IllegalArgumentException` in the constructor and have the test `expectThrows()` that?

renatoh:
> looks good to me. I wonder about the name of the parameter, maybe "greedy" would be more intuitive as a way to describe what it is doing?

not saying "consumeChars" is a good name, happy to change it. but "greedy", as we know it from Regex, is something else, we are not capturing as much as possible, we are skipping forward on a match
what do you think of "reuseChars" with a default of true?

rmuir:
for changing defaults, my goto would be, if we could do that as a followup PR, for a major release. 

We can expose this parameter in a minor release without hurting anyone, but if we change the default it could cause some reindexing for users.

renatoh:
I would argue, at least in German, nothing but longestMatch=true and skipping forward does make any sense. Without skipping forward the filter extracts a lot of nonsense and in my opinion is unusable, at least in German. I've seen this Filter being dropped from projects because of that unexpected behavior and at first I thought this is a bug.
But with an additional parameter "reuseChars" with a default of false we would not change the default behavior.

rmuir:
Yes, I'm just suggesting to split it. We can add this new parameter here, backport to minor release 10.2.0, no breaking changes. Separately we can default it to `true` for 11.0?

renatoh:
> Yes, I'm just suggesting to split it. We can add this new parameter here, backport to minor release 10.2.0, no breaking changes. Separately we can default it to `true` for 11.0?

Sure, do you think reuseChars ( with default true to keep current behavior) is a better name?

rmuir:
Thanks @renatoh !

renatoh:
> Thanks @renatoh !

thanks for your inputs and review it!

rmuir:
@renatoh Feel free to open another PR, if you have time, to try to improve defaults around this for the next version of lucene. If i ask for "longest match" I don't expect to have additional "shorter" subwords coming out of the analyzer, so it seems like a good improvement.

renatoh:
@rmuir could we reduce it to only two 'valid' behavior: onlyLongestMatch=true  with reuseChars=false and onlyLongestMatch=false with reuseChars=true. if we think only these two cases make sense, we could actually reduce it to one flag and try to come up with a different name for that only flag. or should onlyLongestMatch=true  with reuseChars=true, the default behavior of today, also be an option?

rmuir:
@renatoh the way i'd go about that, would be to replace `reuseChars` with `newAndAwesomeLongestMatch` (obviously needs a better name). And deprecate `onlyLongestMatch` in favor of it in 10.2.0. Remove `onlyLongestMatch` completely from 11.0. If enough people complain that they want "old onlyLongestMatch" then we can add it back before releasing 11.0

Issue #14190, opened 2025-02-02
Create a bot to add milestones to new PRs


stefanvodita:
### Description

We risk forgetting to add a milestone to our PRs for the next release, which makes the release manager's job harder. Is there some GitHub workflow that could do that for us?

mikemccand:
It might in theory check the `lucene/CHANGES.txt` to look for an entry (with the PR/issue number) summarizing the PR?  Then it could see which Lucene release the issue is under. 
 This would also serve to validate that `CHANGES.txt` was updated (we also often forget this).  But ... many PRs are trivial and don't require a `CHANGES.txt` entry, so I guess we'd need some escape hatch to tell the bot it's OK.

stefanvodita:
I think #14672 is the first PR to be added to a milestone by the bot. Calling it a success and resolving this issue!

stefanvodita:
#14697 is a nice example of the bot modifying the milestone after we moved the CHANGES entry to a different section!

PR #14279, opened 2025-02-23
Adds github action to verify changelog entry and set milestone to PRs

### LINKED ISSUE #13898: Create a bot to check if there is a CHANGES entry for new PRs

pseudo-nymous:
### Description
Adds github action workflow to verify change log entry and set milestone based on the location of change log entry.

This workflow does the following step by step:
1. Check if the PR has `skip changelog` label to skip the further steps. `Committers should add this label if they want skip adding change log entry`. This is helpful when change log entry is not required.
2. Verifies if there is an entry in change log file (lucene/CHANGES.txt). 
3. Extracts the milestone version based on the location of the change log entry.
4. Checks if the milestone exist in github then applies (adds/updates) the milestone version to the PR.
5. Flags if milestone version do not exist. Currently, we are only logging it. We can either add a comment to create the milestone or create one automatically if a new banner is created (=== Lucene XXX ===) under change log file.

Currently, I have only added logs in the action workflow testing all the flow. It's not blocking the merging process. 
We can make it blocking based on the discussion in this PR.

Tested changes here: https://github.com/pseudo-nymous/lucene/pull/13

Closes: https://github.com/apache/lucene/issues/13898 https://github.com/apache/lucene/issues/14190

### DIFF
--- a/.github/workflows/verify-changelog-and-set-milestone.yml
+++ b/.github/workflows/verify-changelog-and-set-milestone.yml
@@ -6,7 +6,7 @@ on:
 env:
   GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   ISSUE: ${{ github.event.issue.html_url }}
-  ESCAPE_HATCH: ${{ '__NA__' }}
+  SKIP_CHANGELOG_LABEL: ${{ '"skip changelog"' }}
   CHANGE_LOG_FILE: ${{ 'lucene/CHANGES.txt' }}
 
 jobs:
@@ -23,15 +23,16 @@ jobs:
       - name: ChangeLog Entry Verifier and Milestone Setter
         run: |
           echo "################## STEP 1 ##################"
-          echo "Checking for escape hatch in the top comment"
-          pr_comment=$(gh pr view ${{ github.event.number }} --json body -q '.body')
-          if echo "$pr_comment" | grep -q ${{ env.ESCAPE_HATCH }}; then
-            echo "Skipping checks as escape hatch: ${ESCAPE_HATCH} is found in the top comment."
-            exit 0
-          else
-            echo "No escape hatch found in the top comment. Proceeding with next steps."
-          fi
-          
+          echo "Get all labels from the PR"
+          mapfile -t labels < <(gh pr view ${{ github.event.number }} --json labels -q '.labels[].name')
+          IFS=','; echo "${labels[*]}"
+          for label in "${labels[@]}"; do
+            if [[ $label == ${{ env.SKIP_CHANGELOG_LABEL }} ]]; then
+              echo "Skipping github action workflow as label: ${SKIP_CHANGELOG_LABEL} is found in the PR."
+              exit 0
+            fi
+          done
+          echo "${{ env.SKIP_CHANGELOG_LABEL }} not found in the PR. Proceeding with next steps."
           
           echo -e "\n"
           echo "################## STEP 2 ##################"
@@ -97,4 +98,6 @@ jobs:
             echo "Milestone:${lucene_version} does not exist. Please create the milestone and run the workflow again."
             exit 0
           fi
-          gh pr edit ${{ github.event.number }} --milestone "${lucene_version}"
+          # Will uncomment this line after we ensure it runs perfectly in production.
+          # gh pr edit ${{ github.event.number }} --milestone "${lucene_version}"
+          echo "Adding/Updating milestone for the PR to:${lucene_version}"
--- a/.github/workflows/verify-changelog-and-set-milestone.yml
+++ b/.github/workflows/verify-changelog-and-set-milestone.yml
@@ -6,7 +6,7 @@ on:
 env:
   GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   ISSUE: ${{ github.event.issue.html_url }}
-  SKIP_CHANGELOG_LABEL: ${{ '"skip changelog"' }}
+  SKIP_CHANGELOG_CHECK_LABEL: ${{ '"skip-changelog-check"' }}
   CHANGE_LOG_FILE: ${{ 'lucene/CHANGES.txt' }}
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/verify-changelog-and-set-milestone.yml
```diff
@@ -0,0 +1,100 @@
...
+            if [[ $milestone == "$lucene_version" ]]; then
+              milestone_exists=true
+              break
+            fi
+          done
+          if [[ $milestone_exists == false ]]; then
+            echo "Milestone:${lucene_version} does not exist. Please create the milestone and run the workflow again."
+            exit 0
+          fi
+          gh pr edit ${{ github.event.number }} --milestone "${lucene_version}"
```
stefanvodita, line 100: If I understand correctly, this is the only place where we're changing the PR and not just writing logs. I think I'd like to merge with this line commented out first, see how it runs, and then come back here and uncomment the line. Does that make sense?

.github/workflows/verify-changelog-and-set-milestone.yml
```diff
@@ -0,0 +1,100 @@
...
+          fi
+          
+          
+          echo -e "\n"
+          echo "################## STEP 2 ##################"
+          echo "Checking for change log entry in ${{ env.CHANGE_LOG_FILE }}"
+          git log  --pretty=oneline | tail -n 2 | cat
+          echo "merge base sha: ${{ github.event.pull_request.base.sha }}, merge head sha: ${{ github.event.pull_request.head.sha }}"
+          if ! git diff ${{ github.event.pull_request.base.sha }} --name-only | grep -q "${{ env.CHANGE_LOG_FILE }}"; then
+            echo "Change log file:${{ env.CHANGE_LOG_FILE }} does not contains an entry corresponding to changes introduced in PR. Please add a changelog entry."
```
stefanvodita, line 42: It's for this to be in the log now, but could we make it a PR comment eventually?

.github/workflows/verify-changelog-and-set-milestone.yml
```diff
@@ -0,0 +1,100 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2
+
+      - name: ChangeLog Entry Verifier and Milestone Setter
+        run: |
+          echo "################## STEP 1 ##################"
+          echo "Checking for escape hatch in the top comment"
```
stefanvodita, line 26: This runs when a PR is updated. Would it make more sense to check the description? Better yet, have a tag we add to skip this workflow?

stefanvodita (commented):
This is great work and really helpful @pseudo-nymous! I left some comments, but I'm looking forward to seeing this in action.

.github/workflows/verify-changelog-and-set-milestone.yml
pseudo-nymous: To clarify, here top comment refers to the description itself. 

But I also like the idea of adding a `skip changelog` label to pr to skip the workflow. This way it's more intuitive rather than adding a specific skip hatch string in description.

I'll update with this change.

.github/workflows/verify-changelog-and-set-milestone.yml
pseudo-nymous: Yes, I was also thinking the same. We can add the related comment change in the next PR when we uncomment the milestone setter line.

To clarify requirements:
1. Do we keep adding a comment after each PR update if user has not added changelog entry yet?
2. Do we just remind the user one time to add a changelog entry?

What do you suggest?

.github/workflows/verify-changelog-and-set-milestone.yml
pseudo-nymous: Yes, this line access the PR. 

It makes sense to just add a log for now and uncomment to pr after we observe how it runs.
I'll update.

pseudo-nymous: Can we create a new label for this purpose? I suggest adding `skip changelog` label.

dweiss:
Please do not force-push your changes. Keep prior commits so that the history is preserved and comments are attached where they should be. Thanks!

pseudo-nymous:
There were small fixes which I missed in previous commit. This didn't had any history attached to it and I kept the commit which had comments. 

Also, I'll keep this in check in next time. Thanks

.github/workflows/verify-changelog-and-set-milestone.yml
stefanvodita: Thanks for the clarification!

.github/workflows/verify-changelog-and-set-milestone.yml
stefanvodita: I created a [label](https://github.com/apache/lucene/labels/skip-changelog-check) and added it to this PR.

stefanvodita:
Thank you for making the changes @pseudo-nymous! I think once we make the action use the new label I created we're ready to test this out.

pseudo-nymous:
I have updated the script to use new label. We are ready to test the changes.

.github/workflows/verify-changelog-and-set-milestone.yml
pseudo-nymous: Thanks for clarification.

stefanvodita (approved):
This looks good to me! If no one objects, I'll merge in the next few days and we can see how it runs. If we like what we see in the logs, we can come back and enable it to set milestones and comment on the PR.

javanna:
Hey @stefanvodita the changelog entry for this was filed under 10.2, but I don't believe the change itself was backported. Can you double check and either backport or move the changelog entry? Thanks!

stefanvodita:
Thanks for pointing that out @javanna! Funny how that happened on a PR that's specifically about the changelog. We should only push this to main. I'll actually delete the entry for now since we're still iterating on this workflow to make it work. See #13898.

PR #14280, opened 2025-02-24
ExceptionInInitializerError in ScorerUtil

imario42:
lazy initialize the ScorerUtil DEFAULT_IMPACTS_ENUM_CLASS to prevent initialization issues with this class if the thread gets interrupted.

### Description

The ScorerUtils fails to initialize if a thread gets interrupted right when it executes its static initialization routine.
One has to restart the JVM then to make it work again.

`Caused by: java.lang.ExceptionInInitializerError: Exception org.apache.lucene.util.ThreadInterruptedException: java.lang.InterruptedException [in thread "/ng_api/infoHub/suggestions"]
	at org.apache.lucene.index.IndexWriter$EventQueue.close(IndexWriter.java:354)
	at org.apache.lucene.index.IndexWriter.rollbackInternalNoCommit(IndexWriter.java:2506)
	at org.apache.lucene.index.IndexWriter.rollbackInternal(IndexWriter.java:2456)
	at org.apache.lucene.index.IndexWriter.shutdown(IndexWriter.java:1341)
	at org.apache.lucene.index.IndexWriter.close(IndexWriter.java:1369)
	at org.apache.lucene.search.ScorerUtil.<clinit>(ScorerUtil.java:56)`

I made the patch so that it will ignore double-initialization of that constant in case of multithreading. I thought it is not worth to tackle with that in that scenario.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/ScorerUtil.java
+++ b/lucene/core/src/java/org/apache/lucene/search/ScorerUtil.java
@@ -38,9 +38,13 @@ import org.apache.lucene.util.PriorityQueue;
 /** Util class for Scorer related methods */
 class ScorerUtil {
 
-  private static final Class<?> DEFAULT_IMPACTS_ENUM_CLASS;
+  private static Class<?> DEFAULT_IMPACTS_ENUM_CLASS = null;
 
-  static {
+  private static Class<?> getDefaultImpactsEnumClass() {
+    if (DEFAULT_IMPACTS_ENUM_CLASS != null) {
+      return DEFAULT_IMPACTS_ENUM_CLASS;
+    }
+
     try (Directory dir = new ByteBuffersDirectory();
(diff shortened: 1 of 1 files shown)


jpountz:
One concern I have with this change is that this code now runs as part of evaluating a query, when users care about query latency.

@uschindler You usually have informed opinions on this sort of things, I wonder if you have thoughts.

jpountz:
It looks like another problem with the current code is that codec randomization may randomly run before this block of code, making it unreliable when running tests. I opened #14306 as an alternative.

uschindler:
This is better. But best would be to add a getter for the class in our internal package (like we do for other stuff). The usual SharedSecrets approach.

uschindler:
But this is fine. We currently only have shared secrets for tests.
Making that class public does not hurt.

jpountz:
This sounded like a good idea so I applied it.

jpountz:
I merged my other PR, which should supersede this one. Closing.

PR #14282, opened 2025-02-24
Remove field argument from DocIdSetBuilder constructor

javanna:
The argument is unused, its callers can stop providing it.


jpountz (approved):
LGTM. For reference, this was initially required because all methods on `PointValues` used to take a `String field`. This became no longer needed with #8545 when all fields got their own `PointValues` instance and the `String field` arguments were removed.

javanna:
thanks for the reference, I found that in the history but I did not have all the context, it's good to know. I was also wondering if this could be perceived as a breaking change and I think that should not be the case, hence I am planning to backport this to 10.x if there are no objections.

PR #14284, opened 2025-02-24
Remove IOException from DocIdSet#iterator signature

javanna:
There is no implementation of DocIdSet#iterator that throws IOException. This commit proposes removing throwing IOException from its signature.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -14,6 +14,8 @@ API Changes
 * GITHUB#14160: Adjusts the default HNSW search behavior so that filter optimized search is enabled
   when 60% or less of vectors pass the pre-filter. (Ben Chaplin, Ben Trent)
 
+* GITHUB#14284: Remove IOException from DocIdSet#iterator signature (Luca Cavanna)
+
 New Features
 ---------------------
 * GITHUB#14097: Binary partitioning merge policy over float-valued vector field. (Mike Sokolov)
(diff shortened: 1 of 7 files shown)


jpountz (commented):
`DocIdSet` used to be an important API for filtering in the past, but since the merge of Query and Filter, it has become mostly an implementation-detail of the query cache. I don't mind removing the exception: since it's used for in-memory caching, pulling an iterator shouldn't throw an I/O exception (as opposed to a `DocIdSetIterator` that would read from `Directory` for instance)?

Should this be a 11.0 change rather than 10.2? The fact that `DocIdSet` was an important part of the API in the past worries me that there may be lots of usage and implementations of it in the wild.

FWIW it looks like there are other cleanups to be done in this class, e.g. removing `DocIdSet#all(int)` and `DocIdSet#bits()` (another remainder from the query/filter merge).

javanna:
> pulling an iterator shouldn't throw an I/O exception

agreed, that was also my thinking when I made the change.

> Should this be a 11.0 change rather than 10.2? 

I don't have a strong opinion, but would we then document it as a breaking change? Isn't this technically an internal API? Also, a bit of a recurring question, but do we need to change anything around how we mark things in the javadocs?

javanna:
> FWIW it looks like there are other cleanups to be done in this class, e.g. removing DocIdSet#all(int) and DocIdSet#bits() (another remainder from the query/filter merge).

Thanks for raising that. I opened #14288 as a starter, will look further into ` bits` as well.

PR #14290, opened 2025-02-25
Remove bits method from DocIdSet

javanna:
The bits method is now redundant in the DocIdSet base class. This commit removes it, while leaving a single leftover usage in BitDocIdSet as a new method that is specific to that implementation.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -67,6 +67,8 @@ API Changes
 * GITHUB#14289: The unused public static DocIdSet#all method has been deprecated.
   (Luca Cavanna)
 
+* GITHUB#14292: The redundant DocIdSet#bits method has been deprecated. (Luca Cavanna)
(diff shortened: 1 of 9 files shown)


jpountz:
> leaving a single leftover usage in BitDocIdSet

Is it because of `QueryBitSetProducer`? If so it looks like this class could be refactored to not use `BitDocIdSet`. All it needs is caching bit sets, and I suspect that it only uses `DocIdSet` as a way to have an `EMPTY` sentinel value. But it could do the same with a simple record that wraps a `BitSet`?

javanna:
Yes, the only usage is in QueryBitSetProducer, that was already previously casting anyways. I will try to apply your suggesting as a follow-up once I merged this change. Also, that one may probably be backported while this one can't.

jpountz (approved):
That makes sense.

PR #14294, opened 2025-02-25
Fix optimization to help inline calls to live docs.

jpountz:
While doing benchmarks on indexes with deletions, I found a bug in `ScorerUtil`, which optimizes live docs for the wrong class: `FixedBitSet` instead of `FixedBit`. Another performance bug is that `FixedBits` did not override `Bits#applyMask` with a more efficient implementation.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/ScorerUtil.java
+++ b/lucene/core/src/java/org/apache/lucene/search/ScorerUtil.java
@@ -39,6 +39,7 @@ import org.apache.lucene.util.PriorityQueue;
 class ScorerUtil {
 
   private static final Class<?> DEFAULT_IMPACTS_ENUM_CLASS;
+  private static final Class<?> DEFAULT_ACCEPT_DOCS_CLASS;
 
   static {
     try (Directory dir = new ByteBuffersDirectory();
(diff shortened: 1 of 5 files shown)


lucene/core/src/java/org/apache/lucene/search/ScorerUtil.java
```diff
@@ -113,22 +116,21 @@ static Scorable likelyTermScorer(Scorable scorable) {
...
-    } else if (acceptDocs != null) {
-      return new FilterBits(acceptDocs);
     } else {
-      return null;
+      return new FilterBits(acceptDocs);
     }
   }
 
-  private static class FilterBits implements Bits {
+  static class FilterBits implements Bits {
```
gf2121, line 133: Why we need to change this?

gf2121 (approved):
LGTM

lucene/core/src/java/org/apache/lucene/search/ScorerUtil.java
jpountz: This is because I needed it in the test, which I had failed to check in. I have now checked it in, but your comment made me want to avoid making this class pkg-private, so I refactored the test. This class is now private again.

gf2121 (approved):

PR #14296, opened 2025-02-26
Fix TestSysoutLimits occasionally failing.

dweiss:
VectorizationProvider may print warnings, which clashes with the test's assumptions. I simply forced any static initializations to occur before the test starts.


dweiss:
Failing repro line, before the fix:
```
./gradlew test --tests TestSysoutsLimits -Dtests.seed=945AAB8255E7F992
```

uschindler (approved):
looks ok.

uschindler:
could the same problem not also appear with MMapDirectory when it prints warnings on startup? For MMapDirectory, loading the class would be enough.

Maybe it is not an issue here, but we should have that in mind.

dweiss:
Thanks, Uwe. I don't know, to be honest. This is internal infrastructure test... maybe it won't trigger or launch MMapDirectory. I'll just merge this one and monitor builds?

PR #14297, opened 2025-02-26
Remove BitDocIdSet#bits method

javanna:
We removed DocIdSet#bits in #14290. This commit removes the last usage of the former bits method from QueryBitSetProducer. A BitSet can be used directly instead of a DocIdSet, introducing a sentinel that's specific to QueryBitSetProducer.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -22,6 +22,8 @@ API Changes
 
 * GITHUB#14291: Remove IOException from ScorerSupplier#setTopLevelScoringClause signature (Luca Cavanna)
 
+* GITHUB#14297: Replace last usage of BitDocIdSet#bits and remove the method. (Luca Cavanna)
+
 New Features
 ---------------------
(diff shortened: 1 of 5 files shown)


lucene/CHANGES.txt
```diff
@@ -22,6 +22,8 @@ API Changes
 
 * GITHUB#14291: Remove IOException from ScorerSupplier#setTopLevelScoringClause signature (Luca Cavanna)
 
+* GITHUB#14297: Replace last usage of BitDocIdSet#bits and remove the method. (Luca Cavanna)
```
jpountz, line 25: I'm unsure if this would make sense to end users, especially as this will land in the same release as #14290. Maybe just add this GH PR number to the CHANGES entry of #14290 that says that `#bits()` goes away?

jpountz (approved):
LGTM

lucene/CHANGES.txt
javanna: yea that's a good point I was also not sure. Probably one change in the log is enough.

PR #14301, opened 2025-02-27
Avoid unnecessary evaluations and skipping documents

hanbj:
### Description

When reading the code related to sorting optimization in NumericComparator, replacing the fixed value 0x1f with 'currentSkipInterval-1' can improve the performance of merging posting lists. After reading and confirming that LongDistanceEigenQuery also uses this method, testing found that the overall query performance improved by about 20%~30%, so it was migrated over.


jpountz (approved):
The change looks good, can you add a CHANGES entry under version 10.2, section "Optimizations"?

jpountz (approved):

PR #14308, opened 2025-02-27
fix lambdas for java 23

rmuir:
After the upgrade to java 23, my editor is flooded with warnings of unused variables from lambdas. Fix them.

I also downloaded eclipse, installed it, checked all possible compiler options, compared the resulting file and ensured out eclipse compiler file is synced up, so that everything is explicit.


jpountz (approved):
I skimmed through the change, it looks good to me.

rmuir:
Thanks, the change is just a bit annoying  from noise perspective. I will merge up main first, to make sure there aren't any new lambdas in recent commits that anger the check.

PR #14310, opened 2025-02-27
Recommend multi-stage retrieval pipelines in oal.search javadocs.

jpountz:
`oal.search` currently only describes how to tweak ranking in a candidate retrieval stage. This enhances docs to recommend multiple stages, using reranking and/or RRF.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/package-info.java
+++ b/lucene/core/src/java/org/apache/lucene/search/package-info.java
@@ -382,7 +382,8 @@
  * produce scores that not only have different ranges, but also different distributions within this
  * range. A robust way of combining multiple retrieval pipelines consists of combining the top hits
  * that they produce through their ranks rather than through their scores using reciprocal rank
- * fusion. This is exposed via {@link org.apache.lucene.search.TopDocs#rrf(int topN, int k, TopDocs[] hits)}.
+ * fusion. This is exposed via {@link org.apache.lucene.search.TopDocs#rrf(int topN, int k,
+ * TopDocs[] hits)}.
  *
  * <h2>Custom Queries &mdash; Expert Level</h2>
  *
--- a/lucene/core/src/java/org/apache/lucene/search/package-info.java
+++ b/lucene/core/src/java/org/apache/lucene/search/package-info.java
@@ -382,7 +382,7 @@
  * produce scores that not only have different ranges, but also different distributions within this
  * range. A robust way of combining multiple retrieval pipelines consists of combining the top hits
  * that they produce through their ranks rather than through their scores using reciprocal rank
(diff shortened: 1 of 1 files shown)


lucene/core/src/java/org/apache/lucene/search/package-info.java
```diff
@@ -350,6 +350,40 @@
...
+ *
+ * <h3>Top hits fusion</h3>
+ *
+ * <p>Sometimes, multiple retrieval pipelines may make sense, having their own pros and cons. A
+ * typical example would be a lexical retrieval pipeline, matching exactly what the user requested,
+ * and a semantic retrieval pipeline, matching documents that are closest to the user's query from a
+ * semantic perspective. Combining scores is hazardous as different retrieval pipelines often
+ * produce scores that not only have different ranges, but also different distributions within this
+ * range. A robust way of combining multiple retrieval pipelines consists of combining the top hits
+ * that they produce through their ranks rather thank through their scores using reciprocal rank
```
rmuir, line 384:
```suggestion
 * that they produce through their ranks rather than through their scores using reciprocal rank
```

jpountz: Woops, thanks for catching.

lucene/core/src/java/org/apache/lucene/search/package-info.java
```diff
@@ -350,6 +350,40 @@
...
+ * <h3>Top hits fusion</h3>
+ *
+ * <p>Sometimes, multiple retrieval pipelines may make sense, having their own pros and cons. A
+ * typical example would be a lexical retrieval pipeline, matching exactly what the user requested,
+ * and a semantic retrieval pipeline, matching documents that are closest to the user's query from a
+ * semantic perspective. Combining scores is hazardous as different retrieval pipelines often
+ * produce scores that not only have different ranges, but also different distributions within this
+ * range. A robust way of combining multiple retrieval pipelines consists of combining the top hits
+ * that they produce through their ranks rather than through their scores using reciprocal rank
+ * fusion. This is exposed via {@link org.apache.lucene.search.TopDocs#rrf(int, int, TopDocs[])}.
```
rmuir, line 385:
```suggestion
 * fusion. This is exposed via {@link org.apache.lucene.search.TopDocs#rrf(int topN, int k, TopDocs[] hits)}.
```
it is at least legal to do this, and can significantly include readability, since types aren't always enough to understand the parameters when reading. 

As far as the auto-generated label that `javadoc` tool makes from it, you'd have to test it out. of course that can always be specified, but maybe this is an easier approach.

jpountz: Ah, thanks, I didn't even know that one could do this. I just checked the generated javadocs, they look good with these parameter names.

rmuir (approved):

PR #14311, opened 2025-02-27
introduce new parameter onlyLongestMatchNoSubwords replacing onlyLongestMatch

renatoh:


renatoh:
@rmuir onlyLongestMatchNoSubwords is basically what was onlyLongestMatch=true + reuseChars=false

renatoh:
@rmuir  any thoughts on my changes?

rmuir:
@renatoh oh, sorry for the slow feedback, did not realize you had deconflicted it.

changes look good to me. I wish there was a way to really use a `@deprecated/@Deprecated` here, but I haven't dug in, it may be challenging due to telescoping ctor. If you have ideas on how to solve it, let me know, otherwise I think we should just merge the change.

renatoh:
@rmuir 
the field is on the super class, hence, it hence we cannot deprecated it. we could deprecated the current constructor and introduce another constructor without the onlyLongestMatch and just call the super-constructor with onlyLongestMatch=false. what do you think?

renatoh:
@rmuir Sorry for rushing, but have you seen my suggestion regarding deprecating the constructor?

rmuir:
@renatoh that seems fine, If you have a way to do it so it works. Because they are just booleans I wasn't sure?

I also can't remember if there is a way that you can signal a deprecation from parameter specified in the TokenFilterFactory (even if its just a hook method for e.g. solr, elasticsearch, etc to use). Maybe @uschindler knows.

renatoh:
> @renatoh that seems fine, If you have a way to do it so it works. Because they are just booleans I wasn't sure?
> 
> I also can't remember if there is a way that you can signal a deprecation from parameter specified in the TokenFilterFactory (even if its just a hook method for e.g. solr, elasticsearch, etc to use). Maybe @uschindler knows.
the constructor with both,onlyLongestMatch and onlyLongestMatchIgnoreSubwords is now deprecated. please have a look if this makes sense to you

rmuir:
@renatoh it is good, I just wanted to check. thank you @uschindler. I will get this merged soon.

rmuir:
I'm just doing final tests. Thanks again @renatoh. I will backport it to 10.2. We can followup to remove the deprecated "sorta-kinda-longest-match" from lucene's `main` branch, and see if we can tidy things up? And only revert it if someone really complains, but this one behaves really more like a bug to me (just my opinion)

renatoh:
> I'm just doing final tests. Thanks again @renatoh. I will backport it to 10.2. We can followup to remove the deprecated "sorta-kinda-longest-match" from lucene's `main` branch, and see if we can tidy things up? And only revert it if someone really complains, but this one behaves really more like a bug to me (just my opinion)

Thanks for merging and your support. Happy to clean that up once we are ready for it, just let me know.

rmuir:
@renatoh we can clean up `main` at any time as it is marked deprecated for 10.2 now

PR #12194, opened 2023-03-08
[GITHUB-11915] Make Lucene smarter about long runs of matches via new API on DISI

zacharymorn:
This PR adds a new API to DISI to find / estimate long running matches, which can then be leveraged by higher level code to skip over a range of matches and speed up certain query tasks.

PR #14312, opened 2025-02-27
Make Lucene better at skipping long runs of matches.

### LINKED ISSUE #11915: Make Lucene smarter about long runs of matches
### Description

Lucene's abstractions are good at dealing with long runs of documents that do not match a query, but much less at dealing with long runs of documents that match a query. In such cases, Lucene still needs to linearly scan these matches and do some amount of work for each of the matches.

Is it actually common to have long runs of matches? For full-text indexes, maybe not so much, only stop words may have runs of adjacent matches. For string fields, this may happen if the field has a default value that is the value of most documents in the collection. Also it's possible for users to use index sorting in order to cluster similar documents together, which increases the likelihood to have long runs of adjacent matches.

One idea would be to augment the `DocIdSetIterator` API to add a new `peekNextNonMatchingDocID` method, which would return the next doc ID that may not be a match. The default implementation could return `docID() + 1`. We'd need to implement this API in postings, doc-value iterators and a few other important `DocIdSetIterators` like `BitSetIterator` and `DocIdSetIterator#all`. And propagate this information in disjunctions and conjunctions. Then we could leverage this API in a number of places:
 - `ReqExclScorer` could ask the prohibited clause for the next doc ID that is worth evaluating on the required clause.
 - Conjunctions could ignore non-scoring clauses over ranges of doc IDs that all match.
 - `FixedBitSet#or` could set ranges of doc IDs at once, which would in-turn speed up `DocIdSetBuilder`.

jpountz:
This is an attempt to resurrect #12194 in a (hopefully) better way. Now that many queries run with `DenseConjunctionBulkScorer`, which scores windows of doc IDs at a time, it becomes natural to skip clauses that have long runs of matches by checking if they match the whole window.

This introduces the same `DocIdSetIterator#peekNextNonMatchingDocID()` API that PR #12194 suggested, implements it in `DocIdSetIterator#all`, and uses it in `DenseConjunctionBulkScorer` to skip clauses that match the whole window.

For better test coverage, `DenseConjunctionBulkScorer` was refactored to require at least one iterator, which can be a `DocIdSetIterator#all` instance if all docs match.

In follow-ups, we should look into supporting other queries that are likely to have long runs of matches, in particular doc-value range queries on fields that are part of the index sort and take advantage of a doc-value skipper.

Closes #11915


lucene/core/src/java/org/apache/lucene/search/DenseConjunctionBulkScorer.java
```diff
@@ -128,6 +128,16 @@ private void scoreWindowUsingBitSet(
...
 
+    windowIterators.clear();
+    for (DocIdSetIterator iterator : iterators) {
+      // Skip iterators that fully match the window
+      if (iterator.docID() > windowBase || iterator.peekNextNonMatchingDocID() < windowMax) {
+        windowIterators.add(iterator);
+      }
+    }
+    // Note: iterators may be empty!
+    iterators = windowIterators;
```
gf2121, line 139: lead clause might get changed in this window, `iterators.get(0)` need to advance `windowBase` again. I believe that is why CI tests get failed.

lucene/core/src/java/org/apache/lucene/search/DenseConjunctionBulkScorer.java
```diff
@@ -128,6 +128,16 @@ private void scoreWindowUsingBitSet(
     assert windowMatches.scanIsEmpty();
     assert clauseWindowMatches.scanIsEmpty();
 
+    windowIterators.clear();
+    for (DocIdSetIterator iterator : iterators) {
+      // Skip iterators that fully match the window
+      if (iterator.docID() > windowBase || iterator.peekNextNonMatchingDocID() < windowMax) {
```
gf2121, line 134: Would it be worth passing the comparison value in like `iterator.peekNextNonMatchingDocID(windowMax)`  so that Implementations may reduce the number the blocks it need to check according to the threshold?

gf2121 (commented):
Thanks @jpountz , we have `nextDoc` for sparse docs, `intoBitset` for dense docs, and now we are getting this new `peekNextNonMatchingDocID` for sequential docs. It is exciting to see `DocIdSetIterator` getting smarter on various doc distributions!

lucene/core/src/java/org/apache/lucene/search/DenseConjunctionBulkScorer.java
jpountz: I'm a bit on the fence about it because it makes testing harder (I like it being declarative, like we do for e.g. impacts or positions), and I don't expect this `peekNextNonMatchingDocID` call to ever be a bottleneck?

jpountz:
I've been thinking a bit more about naming since I don't like peekNextNonMatchingDocID much, I'm thinking of renaming to `docIDRunEnd` (using "run" as in "run-length encoding"). I like it better because it just says that there is a run of adjacent doc IDs without implying that the next doc ID doesn't match. It's also shorter.

lucene/core/src/java/org/apache/lucene/search/DenseConjunctionBulkScorer.java
gf2121: >  I don't expect this peekNextNonMatchingDocID call to ever be a bottleneck

That makes sense, thanks for explanation!

gf2121 (approved):
Nice work!

lucene/core/src/java/org/apache/lucene/search/DenseConjunctionBulkScorer.java
```diff
@@ -117,6 +107,65 @@ private static int advance(FixedBitSet set, int i) {
...
+      for (DocIdSetIterator iterator : iterators) {
+        if (iterator.docID() > min) {
+          minDocIDRunEnd = min;
+          break;
+        } else {
+          minDocIDRunEnd = Math.min(minDocIDRunEnd, iterator.docIDRunEnd());
+        }
+      }
+
+      if (minDocIDRunEnd >= bitsetWindowMax) {
```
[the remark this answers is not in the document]
jpountz: I would like to avoid collecting tiny windows of doc IDs at once, so that collectors can feel free to apply logic that has some overhead in `LeafCollector#collect(DocIdStream)` (e.g. https://github.com/apache/lucene/pull/14273/files#diff-05525bb5769d4251279bcd9c76d259f8eb451a16075af357e69bd98890c3db5bR257). But I applied your suggestion of relaxing the window size a bit in the case when everything matches.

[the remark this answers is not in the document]
gf2121: > I would like to avoid collecting tiny windows of doc IDs at once, so that collectors can feel free to apply logic that has some overhead in LeafCollector#collect(DocIdStream) (e.g. https://github.com/apache/lucene/pull/14273/files#diff-05525bb5769d4251279bcd9c76d259f8eb451a16075af357e69bd98890c3db5bR257).

Good point, thanks!

gf2121 (approved):

PR #14318, opened 2025-03-01
lint python scripts in dev-tools

rmuir:
These scripts are important, but have zero static analysis, type-checking, or formatting.

Add simple make-based build with CI hook for this.

Format check is not yet enabled, we should "make reformat" and commit the changes separately.

Any failing violations are disabled, we can iteratively enable them.

This is similar to the hook added to luceneutil to start cleaning up the python there.

### DIFF
--- a/.github/actions/prepare-for-build/action.yml
+++ b/.github/actions/prepare-for-build/action.yml
@@ -2,11 +2,12 @@
 # for java, gradle, caches, etc.
 
 name: Prepare Lucene build
+description: Creates a shared setup for other workflows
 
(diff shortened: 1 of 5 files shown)


rmuir:
If we are ok with it, I would want to backport these changes and cleanups to 10.x branch, too, so that things start to become easier.

dweiss (approved):

rmuir:
I will followup with changes to iterate. mainly want to detect common mistakes as a start, then it won't be so terrifying to change or review these files

PR #14319, opened 2025-03-01
python scripts: fix enough so that undefined variable analysis works

rmuir:
Fix basic errors from linter and get undefined variable analysis working through the type-checker.

This will detect common problems such as typos, instead of at runtime. The changes are straightforward: cleaning up imports, basic bad things.


dweiss (approved):
Nice.

Issue #14315, opened 2025-02-28
TestDirectIODirectory is slow on Ubuntu/Linux systems


dweiss:
### Description

Something is odd here.  This, on main:

```
./gradlew :lucene:misc:test --tests "org.apache.lucene.misc.store.TestDirectIODirectory" -Ptests.jvms=4 -Ptests.jvmargs= -Ptests.seed=DEE483D0402DBE52 -Ptests.multiplier=2 -Ptests.useSecurityManager=true -Ptests.nightly=true -Ptests.vectorsize=512 -Ptests.forceintegervectors=true
```

Takes a veeery long time. Specifically, testRandomByte from the parent class (BaseDirectoryTestCase) takes ages, with this loop:
```
      // subslices
      for (int offset = 1; offset < bytes.length; offset++) {
        RandomAccessInput subslice = input.randomAccessSlice(offset, input.length() - offset);
        assertEquals(input.length() - offset, subslice.length());
        assertBytes(subslice, bytes, offset);
      }
```
taking the most of the time. The slices involved are not large. But here's a sample flame graph from the profiler on Linux -

![Image](https://github.com/user-attachments/assets/1aed6636-5894-43a0-a0cb-1cb318d97f26)

Seems like those native pread0 reads are where it's spending most of the time... I didn't have time to compare to a different system yet (or measure it more accurately deeper into the native stack).

ChrisHegarty:
Any particular JDK, or observable in many ?

dweiss:
I don't know if it's the JDK. Seems like it's direct i/o all the way to the kernel -

![Image](https://github.com/user-attachments/assets/4e087313-c029-450e-a3c2-151625ebdc93)

This is with
```
openjdk version "23.0.2" 2025-01-21
OpenJDK Runtime Environment Temurin-23.0.2+7 (build 23.0.2+7)
OpenJDK 64-Bit Server VM Temurin-23.0.2+7 (build 23.0.2+7, mixed mode, sharing)
```
on  Linux 6.11.0-17-generic #17~24.04.2-Ubuntu SMP PREEMPT_DYNAMIC Mon Jan 20 22:48:29 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

dweiss:
For the seed above, it takes 4m29s to execute this test.

ChrisHegarty:
I can only think that this is a bug that I introduced with recent refactoring. Looking ...

dweiss:
[...]

PR #14320, opened 2025-03-01
Fix DirectIOIndexInput seek to not read when position is within buffer

ChrisHegarty:
This commit changes DirectIOIndexInput::seek so that it repositions the indexInput's position without a read, when the new position is within the bounds of the current buffer.

Prior to this fix (on Linux):
```
times out after more than 10 mins
```

With this fix:
```
The slowest tests (exceeding 500 ms) during this run:
   3.48s TestDirectIODirectory.testRandomByte (:lucene:misc)
```

relates #14315

### DIFF
--- a/lucene/misc/src/java/org/apache/lucene/misc/store/DirectIODirectory.java
+++ b/lucene/misc/src/java/org/apache/lucene/misc/store/DirectIODirectory.java
@@ -373,7 +373,13 @@ public class DirectIODirectory extends FilterDirectory {
     @Override
     public void seek(long pos) throws IOException {
       if (pos != getFilePointer()) {
-        seekInternal(pos);
+        final long absolutePos = pos + offset;
+        if (absolutePos >= filePos && absolutePos <= filePos + buffer.limit()) {
+          // the new position is within the existing buffer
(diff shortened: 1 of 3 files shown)


dweiss (approved):

dweiss:
Good find!

ChrisHegarty:
I’m going to see if I can write a few small unit tests for this particular scenario.

lucene/misc/src/java/org/apache/lucene/misc/store/DirectIODirectory.java
```diff
@@ -373,7 +373,13 @@ public long getFilePointer() {
     @Override
     public void seek(long pos) throws IOException {
       if (pos != getFilePointer()) {
-        seekInternal(pos);
+        final long absolutePos = pos + offset;
+        if (absolutePos >= filePos && absolutePos <= filePos + buffer.limit()) {
+          // the new position is within the existing buffer
+          buffer.position(Math.toIntExact(absolutePos - filePos));
```
rmuir, line 379:
```suggestion
          buffer.position(Math.subtractExact(absolutePos, filePos));
```

rmuir (approved):
Looks good to me.

PR #14321, opened 2025-03-02
python scripts: enable standard type checking

rmuir:
Fix all the type issues, enable checking so that no more creep in.

A couple third-party libraries have some issues (such as totally wrong typing), I just disabled in that case.

Fixes are mostly basic stuff such as adding missing null checks, occasional cast or type-hint.


rmuir:
After this PR, i'd like to run `make reformat` and create PR with the result. The mix of 2-space and 4-space indentation is killing me :)

dweiss (approved):

PR #14322, opened 2025-03-02
reformat the python code with 'make reformat' and enable format in CI check

rmuir:
currently the python code has a mix of indentation, styles, imports ordering, etc. for example, it is very difficult to work with mixed indentation levels: the language is sensitive to indentation.

reformat all the code with 'make reformat' and enable format checks when linting. It works like spotless, just don't think about it.

### DIFF
--- a/dev-tools/scripts/Makefile
+++ b/dev-tools/scripts/Makefile
@@ -23,9 +23,7 @@ VENV=${PWD}/.venv
 SOURCES=$(wildcard *.py)
 
 # check formatting, linting, and types
-# TODO: enable this when we're ready to reformat all code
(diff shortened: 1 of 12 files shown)


stefanvodita (approved):
I was looking at the luceneutil PRs, but missed the fact that you were also doing this for Lucene. Thank you @rmuir!

rmuir:
thank you @stefanvodita for reviewing (in both places). I am sure it won't go 100% smooth and there will be short-term problems, but I think in the long term it will make things easier.

rmuir:
it is merged, the squash-based merging is not useful for situations like this.

PR #14326, opened 2025-03-03
python: enable all linting checks and type-hint the code

rmuir:
Review https://github.com/apache/lucene/pull/14322 separate as it is included here to prevent merge conflicts.

So look at the 2nd commit if you want to see what has changed:
* enable all type checking
* type-hint the code to get checker flowing
* disable a few remaining checks with a TODO list
* enable all linter checks
* autofix all safe fixes (e.g. api migrations)
* disable just the remaining checks with a massive TODO list

### DIFF
--- a/dev-tools/scripts/create_line_file_docs.py
+++ b/dev-tools/scripts/create_line_file_docs.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-# -*- coding: utf-8 -*-
 # Licensed to the Apache Software Foundation (ASF) under one or more
 # contributor license agreements.  See the NOTICE file distributed with
 # this work for additional information regarding copyright ownership.
@@ -24,16 +23,17 @@ import subprocess
 import tempfile
 import time
 import urllib.request
+from io import TextIOWrapper
 
 DEBUG = False
 
 TARGET_DOC_CHARS = 1024
 
 
-def compress_with_seek_points(file_name_in, file_name_out, num_seek_points):
+def compress_with_seek_points(file_name_in: str, file_name_out: str, num_seek_points: int):
   bytes_per_chunk = os.path.getsize(file_name_in) / num_seek_points
 
-  seek_points = []
+  seek_points: list[int] = []
 
   if os.path.exists(file_name_out):
     os.remove(file_name_out)
@@ -60,7 +60,7 @@ def compress_with_seek_points(file_name_in, file_name_out, num_seek_points):
         break
 
       bytes_in_chunk += len(line)
-      f_out.write(line)  # false positive in python's crazy typing # pyright: ignore[reportArgumentType]
+      f_out.write(line)
 
       if bytes_in_chunk > bytes_per_chunk and chunk_count < num_seek_points:
         f_out.close()
@@ -77,16 +77,13 @@ re_newlines = re.compile(r"\n+")
 re_space = re.compile(r"\s")
 
 # used to find word break, for splitting docs into ~1 KB sized smaller docs:
-re_next_non_word_character = re.compile(r"\W", re.U)
+re_next_non_word_character = re.compile(r"\W", re.UNICODE)
 
 EUROPARL_V7_URL = "https://www.statmt.org/europarl/v7/europarl.tgz"
 
 
-def split_docs(all_out, title_string, date_string, body_string):
-  """
-  Splits docs into smallish (~1 KB) sized docs, repeating same title and date
-  """
-
+def split_docs(all_out: TextIOWrapper, title_string: str, date_string: str, body_string: str):
+  """Splits docs into smallish (~1 KB) sized docs, repeating same title and date"""
   doc_count = 0
   while len(body_string) > 0:
     char_count = int(random.gauss(TARGET_DOC_CHARS, TARGET_DOC_CHARS / 4))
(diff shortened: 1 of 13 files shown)


dev-tools/scripts/create_line_file_docs.py
```diff
@@ -60,7 +60,7 @@ def compress_with_seek_points(file_name_in, file_name_out, num_seek_points):
         break
 
       bytes_in_chunk += len(line)
-      f_out.write(line)  # false positive in python's crazy typing # pyright: ignore[reportArgumentType]
+      f_out.write(line)
```
stefanvodita, line 63: Curious - what changed that that is no longer necessary?

dev-tools/scripts/pyproject.toml
```diff
@@ -1,21 +1,163 @@
...
 indent-width = 2
 
 [tool.ruff.lint]
-# TODO: improve!
-# select = ["ALL"]
+select = ["ALL"]
 
 # disabling/enabling of rules
 ignore = [
+  # This is a big TODO list of current linter problems.
```
stefanvodita, line 38: Should we open an issue to make it more likely someone finds this and decides to carry on the work?

dev-tools/scripts/releaseWizard.py
```diff
@@ -887,7 +873,7 @@ def get_release_version():
     version = Version.parse(v)
   except Exception:
     print("Not a valid version %s" % v)
-    return get_release_version()
+    raise
```
stefanvodita, line 876: This change doesn't look like the others.

stefanvodita (commented):
Reviewed the 2nd commit. Only small comments.

dev-tools/scripts/create_line_file_docs.py
rmuir: IIRC, it seems like a corner-case bug in the typechecker. typechecker has a grid of options: https://docs.basedpyright.com/latest/configuration/config-files/#diagnostic-settings-defaults

And in the process of getting started here, lots of booleans were being flipped on and off. I think it led to a false positive given the exact circumstances of booleans (ignore this problem, but fail on that problem, etc): basically we are asking it to validate within a broken/incoherent type system. 

Once enough stuff got fixed, then strictness was enabled which allows things to work smoothly. This is just my read on what happened.

dev-tools/scripts/pyproject.toml
rmuir: Yes I was planning to make a "good first issue" in both repositories for this? I thought first, we might want to have a look at the list and see if there is anything we are against, and simply move those further down in the file, away from the TODO list, instead under "list of rules we keep disabled". 

I am sure there are probably things here that at least aren't priorities: policing print() statements, TODOs, and docstrings are examples. Maybe we don't even want to adopt python naming conventions. These are just examples.

So we could remove any bikesheds from  the list first: it would be great to get some of these issues fixed such as variable shadowing, exception handling, etc

dev-tools/scripts/create_line_file_docs.py
rmuir: In this example I'm pretty sure that because `file_name_out` parameter was untyped (Unknown), it made `gzip.open()` call more ambiguous, causing `f_out` to have type of `GzipFile | TextIOWrapper[_WrappedBuffer]`

Once `file_name_out` got a type of `str`, it reduces ambiguity and narrows result of `gzip.open()` and hence `f_out` to just be `GzipFile`.

You can see this yourself if you go in editor and change `file_name_out` to be Any

dev-tools/scripts/releaseWizard.py
rmuir: I think some of the change was automatic. What else should we do if supplied with an invalid version? try to proceed anyway? :)

rmuir: I know when battling this one, it was difficult to fix the issues because of situations like this. It causes helper function `get_release_version()` to have unnecessarily wide type (e.g. allowing None or maybe even other types in some places). And callsites are not checking for those exceptional cases, which makes for tons of chaos.

Better for exceptional cases to be exceptions?

rmuir:
By the way, thank you for reviewing these changes! I know it isnt fun, and I know i'll cause something to break: but these scripts are important (releases, backwards-compatibility, community stuff) and I'll help fix any short-term fallout.

The changes make the scripts easier to work with too besides just linting. You get things like autocomplete and hover documentation in your editor when the types work, its 2025 :)

dev-tools/scripts/pyproject.toml
stefanvodita: That's a good idea!

dev-tools/scripts/create_line_file_docs.py
stefanvodita: Got it, thank you for the explanation!

dev-tools/scripts/releaseWizard.py
stefanvodita: It's hard for me to judge this one without knowing how it's used, but I can imagine it's frustrating to stop a long workflow with this exception because of a typo instead of prompting the user a second time. Now, from looking at the code, it seems that is not the case, so it should be fine to proceed.

stefanvodita (approved):
Happy I could help review and fingers crossed it doesn't break anything too badly!

PR #14328, opened 2025-03-03
Avoid using time zones that emit warnings (jdk25+)

dweiss:
This causes tests that expect exact outputs (like TestReproduceMessage) to occasionally fail under JDK25+. I added some filtering to randomTimeZone so that those warning-emitting time zone codes are never used (on JDKs that don't support them).


ChrisHegarty (approved):
Thanks for tracking this down and fixing the test-infra! LGTM

rmuir (approved):

uschindler:
Looks fine!

dweiss:
I've backported this to branch_10x.

PR #12722, opened 2023-10-25
Disable suffix sharing for block tree index

gf2121:
closes https://github.com/apache/lucene/issues/12702

Issue #12659, opened 2023-10-12
Nightly benchmark regression for term dict queries


gf2121:
### Description

I'm seeing regressions of term-dict queries on nightly benchmark:
https://home.apache.org/~mikemccand/lucenebench/2023.10.10.18.03.55.html

Diff commits: https://github.com/apache/lucene/compare/33a3af426003aea1dbcb60f2eeed35d47d65b9d7...05d26ac44dde056636fb8518b520a9b83fc7668f

IMO the commit that most probably cause these regressions is: https://github.com/apache/lucene/pull/12631. But Mike helped run luceneutil before merge and [result](https://github.com/apache/lucene/pull/12631#issuecomment-1752064474) looks good ...

gf2121:
I write a JMH benchmark to compare reading `vint` vs `msbvint`.
```
Benchmark                         (size)  (valueBit)   Mode  Cnt   Score   Error   Units
ReadVLongBenchmark.readVLong         128           7  thrpt    5  29.406 ± 0.148  ops/us
ReadVLongBenchmark.readVLong         128          14  thrpt    5   3.644 ± 0.013  ops/us
ReadVLongBenchmark.readVLong         128          21  thrpt    5   2.416 ± 0.004  ops/us
ReadVLongBenchmark.readVLong         128          28  thrpt    5   2.074 ± 0.005  ops/us
ReadVLongBenchmark.readVLong         128          35  thrpt    5   1.838 ± 0.001  ops/us
ReadVLongBenchmark.readVLong         128          42  thrpt    5   1.958 ± 0.004  ops/us
ReadVLongBenchmark.readVLong         128          49  thrpt    5   1.511 ± 0.012  ops/us
ReadVLongBenchmark.readVLong         128          56  thrpt    5   1.463 ± 0.005  ops/us
ReadVLongBenchmark.readVLong         128          63  thrpt    5   1.299 ± 0.092  ops/us
ReadVLongBenchmark.writeVLongMSB     128           7  thrpt    5  28.622 ± 0.865  ops/us
ReadVLongBenchmark.writeVLongMSB     128          14  thrpt    5   4.479 ± 0.013  ops/us
ReadVLongBenchmark.writeVLongMSB     128          21  thrpt    5   3.647 ± 0.007  ops/us
ReadVLongBenchmark.writeVLongMSB     128          28  thrpt    5   2.434 ± 0.372  ops/us
[...]

PR #12631, opened 2023-10-07
Write MSB VLong for better outputs sharing in block tree index

gf2121:
closes https://github.com/apache/lucene/issues/12620

PR #14333, opened 2025-03-06
A specialized Trie for Block Tree Index

gf2121:
**Context**

* #12631 introduced a MSBVLong format to encode the first fp of FST output. It is the first time we benefit from the output sharing in blocktree. The change reduces ~13% tip size, in turn caused a performance regression when accumulating output bytes #12659. Then https://github.com/apache/lucene/pull/12722 introduce a complex and tricky OutputAccumulator to get the performance back a bit, while still slower than no output prefix sharing.

* https://github.com/apache/lucene/pull/12722 we disabled suffix sharing as we find that very few suffix get shared in block tree.

**Proposal**

Before the PRs mentioned above, the fst in block tree is almost like a trie - no output prefix sharing and few suffix sharing.  When output prefix get shared, we saved space, but also lost performance. IMO for `tip`, performance is more important than storage size, which is usually a very small part of the whole index, and loaded off-heap.

This PR tries to introduce a simple trie that specialized for blocktree index: 

* It can have assumptions that keys are bytes so numbers are better encoded.
* The output bytes array can be read from `IndexInput` without copy. 
* It also improves performance by replacing vint / vlong with a over-read and mask.
... 

PR is still a draft, but the number looks promising.

**Storage**

<meta http-equiv="Content-Type" content="text/html; charset=utf-8"></meta><byte-sheet-html-origin data-id="1741344596281" data-version="4" data-is-embed="false" data-grid-line-hidden="false" data-importRangeRawData-spreadSource="https://bytedance.larkoffice.com/sheets/HPOosqSs7hkOMOtURy0c5mV9nlf" data-importRangeRawData-range="&#39;Sheet1&#39;!A1:D17">
  | Baseline | Candidate | diff
-- | -- | -- | --
_32_Lucene101_0.tip | 4425601 | 3827402 | -13.52%
_65_Lucene101_0.tip | 4458107 | 3884702 | -12.86%
_98_Lucene101_0.tip | 4791217 | 4185525 | -12.64%
_cb_Lucene101_0.tip | 4832497 | 4183043 | -13.44%
_fe_Lucene101_0.tip | 4807799 | 4173039 | -13.20%
_fp_Lucene101_0.tip | 720343 | 587765 | -18.40%
_g0_Lucene101_0.tip | 721438 | 585146 | -18.89%
_gb_Lucene101_0.tip | 694205 | 566073 | -18.46%
_gm_Lucene101_0.tip | 688145 | 561999 | -18.33%
_gx_Lucene101_0.tip | 819804 | 657046 | -19.85%
_gy_Lucene101_0.tip | 142276 | 100526 | -29.34%
_gz_Lucene101_0.tip | 125578 | 87568 | -30.27%
_h0_Lucene101_0.tip | 109982 | 77425 | -29.60%
_h1_Lucene101_0.tip | 113266 | 80308 | -29.10%
_h2_Lucene101_0.tip | 104672 | 72846 | -30.41%
sum | 27554930 | 23630413 | -14.24%

</byte-sheet-html-origin>

**Search**

```
                          IntSet      595.28      (3.3%)      584.37      (3.9%)   -1.8% (  -8% -    5%) 0.110
                      OrHighRare      108.03      (8.7%)      106.53      (9.2%)   -1.4% ( -17% -   18%) 0.624
                  FilteredPhrase       16.86      (2.9%)       16.69      (3.0%)   -1.0% (  -6% -    5%) 0.288
                IntervalsOrdered        7.89      (3.1%)        7.81      (4.9%)   -1.0% (  -8% -    7%) 0.448
                        SpanNear       12.64      (2.1%)       12.56      (2.2%)   -0.6% (  -4% -    3%) 0.359
                    FilteredTerm      128.98      (3.2%)      128.26      (2.9%)   -0.6% (  -6% -    5%) 0.565
                 CountAndHighMed       84.00      (2.3%)       83.55      (1.6%)   -0.5% (  -4% -    3%) 0.389
                AndMedOrHighHigh       54.23      (4.6%)       53.99      (4.2%)   -0.5% (  -8% -    8%) 0.745
             CountFilteredPhrase       12.76      (2.3%)       12.70      (2.4%)   -0.4% (  -4% -    4%) 0.556
                  CountOrHighMed      172.95      (3.8%)      172.23      (3.6%)   -0.4% (  -7% -    7%) 0.724
                CountAndHighHigh       84.36      (2.0%)       84.03      (1.6%)   -0.4% (  -3% -    3%) 0.496
                     TermGroup1M       23.31      (1.9%)       23.24      (2.1%)   -0.3% (  -4% -    3%) 0.635
                          Phrase       50.78      (3.1%)       50.65      (2.9%)   -0.3% (  -6% -    5%) 0.781
                     CountPhrase       18.16      (1.8%)       18.12      (2.0%)   -0.2% (  -4% -    3%) 0.698
                    TermGroup10K       17.28      (1.4%)       17.24      (1.9%)   -0.2% (  -3% -    3%) 0.665
                  FilteredOrMany        7.97      (1.9%)        7.95      (1.8%)   -0.2% (  -3% -    3%) 0.730
                   TermTitleSort      138.40      (3.0%)      138.24      (3.6%)   -0.1% (  -6% -    6%) 0.910
     FilteredAnd2Terms2StopWords      436.82      (3.3%)      436.31      (3.1%)   -0.1% (  -6% -    6%) 0.906
             CountFilteredOrMany       13.23      (2.1%)       13.22      (2.2%)   -0.1% (  -4% -    4%) 0.889
             CombinedAndHighHigh        9.59      (2.7%)        9.59      (2.2%)   -0.1% (  -4% -    5%) 0.930
             FilteredOrStopWords       25.57      (3.9%)       25.56      (3.3%)   -0.0% (  -6% -    7%) 0.966
                 AndHighOrMedMed       15.54      (1.4%)       15.53      (1.4%)   -0.0% (  -2% -    2%) 0.939
         CountFilteredOrHighHigh       37.22      (1.2%)       37.21      (0.9%)   -0.0% (  -2% -    2%) 0.923
                    TermGroup100       16.94      (2.4%)       16.95      (2.3%)    0.0% (  -4% -    4%) 0.989
                     CountOrMany       15.03      (3.4%)       15.03      (3.4%)    0.0% (  -6% -    7%) 0.988
                 CountOrHighHigh       89.26      (1.9%)       89.28      (1.7%)    0.0% (  -3% -    3%) 0.960
             CountFilteredIntNRQ       23.19      (1.6%)       23.20      (1.7%)    0.0% (  -3% -    3%) 0.944
                  FilteredIntNRQ       35.57      (1.8%)       35.60      (2.3%)    0.1% (  -3% -    4%) 0.887
                          IntNRQ       35.72      (1.9%)       35.76      (2.1%)    0.1% (  -3% -    4%) 0.860
                    TermBGroup1M       23.72      (2.4%)       23.75      (2.2%)    0.1% (  -4% -    4%) 0.848
              CombinedAndHighMed       78.87      (3.2%)       79.00      (3.3%)    0.2% (  -6% -    6%) 0.876
               FilteredOrHighMed      147.44      (4.0%)      147.69      (2.8%)    0.2% (  -6% -    7%) 0.876
          CountFilteredOrHighMed       46.61      (1.1%)       46.69      (1.0%)    0.2% (  -1% -    2%) 0.609
              FilteredOrHighHigh       30.50      (3.9%)       30.58      (3.1%)    0.3% (  -6% -    7%) 0.816
                    AndStopWords        6.02      (6.8%)        6.04      (4.4%)    0.3% ( -10% -   12%) 0.881
                       OrHighMed       55.30      (4.4%)       55.45      (5.8%)    0.3% (  -9% -   10%) 0.868
                          OrMany        5.73      (6.5%)        5.75      (7.0%)    0.3% ( -12% -   14%) 0.898
                      AndHighMed      116.16      (3.6%)      116.50      (3.4%)    0.3% (  -6% -    7%) 0.789
                FilteredOr3Terms      106.20      (3.8%)      106.59      (2.4%)    0.4% (  -5% -    6%) 0.713
                    SloppyPhrase       22.88      (3.1%)       22.98      (3.1%)    0.4% (  -5% -    6%) 0.666
                  TermBGroup1M1P       13.89      (2.4%)       13.96      (1.9%)    0.4% (  -3% -    4%) 0.518
            FilteredAndStopWords       26.26      (4.3%)       26.38      (4.0%)    0.5% (  -7% -    9%) 0.719
               CombinedOrHighMed       15.73      (3.4%)       15.81      (1.5%)    0.5% (  -4% -    5%) 0.564
                 DismaxOrHighMed      100.85      (3.4%)      101.37      (3.6%)    0.5% (  -6% -    7%) 0.642
              CombinedOrHighHigh       12.56      (3.3%)       12.64      (1.0%)    0.6% (  -3% -    5%) 0.403
      FilteredOr2Terms2StopWords      192.94      (2.7%)      194.24      (1.9%)    0.7% (  -3% -    5%) 0.362
                      DismaxTerm      719.66      (5.8%)      724.60      (6.9%)    0.7% ( -11% -   14%) 0.733
                       And3Terms      493.05      (5.2%)      496.54      (5.4%)    0.7% (  -9% -   12%) 0.676
              Or2Terms2StopWords      102.67      (4.8%)      103.43      (5.4%)    0.7% (  -9% -   11%) 0.643
                    CombinedTerm       23.30      (3.5%)       23.48      (1.7%)    0.8% (  -4% -    6%) 0.365
             FilteredAndHighHigh       25.26      (4.6%)       25.46      (4.7%)    0.8% (  -8% -   10%) 0.594
                     OrStopWords       37.40      (4.8%)       37.73      (6.2%)    0.9% (  -9% -   12%) 0.614
                DismaxOrHighHigh       21.65      (3.7%)       21.86      (5.5%)    1.0% (  -7% -   10%) 0.521
               FilteredAnd3Terms       99.12      (5.5%)      100.08      (5.3%)    1.0% (  -9% -   12%) 0.573
                      OrHighHigh       24.82      (4.6%)       25.07      (5.8%)    1.0% (  -9% -   12%) 0.546
                     AndHighHigh       83.00      (4.4%)       83.91      (3.1%)    1.1% (  -6% -    9%) 0.369
                            Term      706.24      (4.0%)      713.95      (4.9%)    1.1% (  -7% -   10%) 0.440
                        Or3Terms      253.05      (5.5%)      256.03      (5.2%)    1.2% (  -9% -   12%) 0.487
                      TermDTSort      243.15      (3.3%)      246.59      (5.2%)    1.4% (  -6% -   10%) 0.305
                   TermMonthSort     2863.45      (7.1%)     2908.45      (7.4%)    1.6% ( -12% -   17%) 0.493
                 FilteredPrefix3      144.48      (2.9%)      146.86      (2.3%)    1.6% (  -3% -    7%) 0.046
             And2Terms2StopWords      247.08      (3.1%)      251.49      (3.1%)    1.8% (  -4% -    8%) 0.066
              FilteredAndHighMed      146.72      (7.4%)      149.37      (7.0%)    1.8% ( -11% -   17%) 0.426
                         Prefix3       68.50      (3.3%)       69.91      (1.6%)    2.1% (  -2% -    7%) 0.012
               TermDayOfYearSort      443.75      (5.2%)      453.71      (4.2%)    2.2% (  -6% -   12%) 0.134
                        Wildcard      100.44      (4.4%)      104.04      (3.3%)    3.6% (  -3% -   11%) 0.004
                          Fuzzy2       47.50      (3.4%)       49.31      (3.6%)    3.8% (  -3% -   11%) 0.001
                         Respell       64.88      (2.9%)       68.90      (3.7%)    6.2% (   0% -   13%) 0.000
                          Fuzzy1      121.48      (2.1%)      129.95      (3.1%)    7.0% (   1% -   12%) 0.000
                       CountTerm     7936.16     (10.6%)     8914.66      (8.8%)   12.3% (  -6% -   35%) 0.000
                        PKLookup      269.87      (3.6%)      361.18      (4.9%)   33.8% (  24% -   43%) 0.000
```


jpountz:
The speedup on PKLookup is exciting!

> IMO for tip, performance is more important than storage size, which is usually a very small part of the whole index, and loaded off-heap.

Since you wrote this, I expected tip files to become bigger, but your data suggests the opposite, tip files are getting smaller? Am I reading it correctly?

gf2121:
Thanks for feedback!

> Since you wrote this, I expected tip files to become bigger, but your data suggests the opposite, tip files are getting smaller? Am I reading it correctly?

Sorry I didn't make that clear. This was my motivation for implementing a specialized trie instead of a specialized FST. From a data structure perspective, an FST should be smaller than a trie because it can share output prefixes (as evidenced by https://github.com/apache/lucene/pull/12631). But in practice, this trie is specialized for block tree index, so the tip is actually smaller because this trie does more bitwise encoding than vint/vlong and it doesn't need to store the length of the floor data since it's already self-contained to know the length during reading.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnum.java
```diff
@@ -857,123 +768,126 @@ public SeekStatus seekCeil(BytesRef target) throws IOException {
...
-          break;
-        }
-        if (f.prefixLength == validIndexPrefix) {
-          isSeekFrame = false;
-        }
-        ord++;
-      }
-    }
-  }
+  /*
```
jpountz, line 771: Should we remove this block or comment it out?

jpountz (commented):
I like it. It looks simpler to me. And you're saying it's also faster and consumes less off-heap memory. So this looks like a net win.

It'd be good for @mikemccand to take a look before moving forward. Maybe keep it this way for now, I suspect that the review will become harder when you properly fork the codec.

mikemccand:
This is an exciting change @gf2121!  Smaller, simpler, and faster!?  I'll try to review soon.

mikemccand (commented):
Phew, I like this idea!  I left tons of small comments (some of which were me just being confused as I slowly caught on...) and had to stop because GitHub seems to have some sort of O(N^3) slowdown as you add more comments...

Trie is indeed a better fit for a terms index since sharing of the suffixes should be relatively uncommon.

This is an exciting change -- thanks @gf2121.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
```diff
@@ -0,0 +1,486 @@
...
+        int header =
+            SIGN_NO_CHILDREN
+                | ((outputFpBytes - 1) << 2)
+                | ((output.hasTerms ? 1 : 0) << 5)
+                | ((output.floorData != null ? 1 : 0) << 6);
+        index.writeByte(((byte) header));
+        writeLongNBytes(output.fp, outputFpBytes, index);
+        if (output.floorData != null) {
+          index.writeBytes(
+              output.floorData.bytes, output.floorData.offset, output.floorData.length);
```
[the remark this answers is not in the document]
gf2121: Floor data is self-contained: it writes the size N of blocks, and write N leading byte and fps. We can easily know the end when reading, like:

```
int size = in.readVInt();
for (int i = 0; i < size; i++) {
    byte b = in.readByte();
    long fp = in.readVLong();
}
```

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
```diff
@@ -0,0 +1,486 @@
...
+      throw new IllegalStateException("tries should be unsaved");
+    }
+    trie.status = Status.DESTROYED;
+    putAll(this.root, trie.root);
+  }
+
+  private static void putAll(Node n, Node add) {
+    assert n.label == add.label;
+    if (add.output != null) {
+      n.output = add.output;
```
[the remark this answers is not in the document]
gf2121: Actually I intended to make it work like `java.util.Map#putAll`, simply overwrite the origin output if any. This makes tests easier.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
```diff
@@ -0,0 +1,486 @@
...
+    }
+    ListIterator<Node> iter = n.children.listIterator();
+    // TODO we can do more efficient if there is no intersection, block tree always do that
```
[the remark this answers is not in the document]
gf2121: I believe it is a O(M + N) as each element only visited once.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
```diff
@@ -0,0 +1,486 @@
...
+      this.label = label;
+      this.output = output;
+      this.children = children;
+    }
+  }
+
+  private Status status = Status.UNSAVED;
+  final Node root = new Node(0, null, new LinkedList<>());
+
+  Trie(BytesRef k, Output v) {
```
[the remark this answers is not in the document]
gf2121: I like it!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
```diff
@@ -0,0 +1,486 @@
...
+      intersect(node.children, key, consumer);
+      key.setLength(key.length() - 1);
+    }
+  }
+
+  void save(DataOutput meta, IndexOutput index) throws IOException {
+    if (status != Status.UNSAVED) {
+      throw new IllegalStateException("only unsaved trie can be saved");
+    }
+    status = Status.SAVED;
```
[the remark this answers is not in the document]
gf2121: > is the act of saving (this method) destructive?

The trie can be saved again if we reset all nodes' fp to -1. We do not actually need that now so i kept it simple.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
```diff
@@ -0,0 +1,486 @@
...
+      n.output = add.output;
+    }
+    ListIterator<Node> iter = n.children.listIterator();
+    // TODO we can do more efficient if there is no intersection, block tree always do that
+    outer:
+    for (Node addChild : add.children) {
+      while (iter.hasNext()) {
+        Node nChild = iter.next();
+        if (nChild.label == addChild.label) {
+          putAll(nChild, addChild);
```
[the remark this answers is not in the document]
gf2121: > but it does mean the length of your terms index prefixes must be under the stack depth recursion limit.

+1, we should avoid this recursive implementation. I think we only have limit of 65535 in BytesRefHash, i do not want to add more strict limitation here.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
```diff
@@ -0,0 +1,486 @@
...
+        index.writeByte(((byte) header));
+        writeLongNBytes(output.fp, outputFpBytes, index);
+        if (output.floorData != null) {
+          index.writeBytes(
+              output.floorData.bytes, output.floorData.offset, output.floorData.length);
+        }
+        continue;
+      }
+
+      Node unCompiled = null;
```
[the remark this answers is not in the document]
gf2121: > shoshin (初心)

Wow!!! sweet translation :)

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
```diff
@@ -0,0 +1,486 @@
...
+      for (Node child : node.children) {
```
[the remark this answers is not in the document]
gf2121: Nice catch, i added an iterator in `Node` to avoid the O(N^2) iteration.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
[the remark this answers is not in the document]
gf2121: I just realized this is a Japanese word. There is an exactly same word 初心 in Chinese, hah :)

jpountz:
I started looking at the code but you would know better: does this new encoding make it easier to know the length of leaf blocks while traversing the terms index so that we could prefetch the right byte range when doing terms dictionary lookups? https://github.com/apache/lucene/blob/661dcae3c25cc548a6df251b79b7bfac81c2dba8/lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/SegmentTermsEnumFrame.java#L147-L148

gf2121:
Thanks for looking :)

> I started looking at the code but you would know better: does this new encoding make it easier to know the length of leaf blocks while traversing the terms index so that we could prefetch the right byte range when doing terms dictionary lookups?

We have the exactly same output as the origin FST index so i do not think it can be easier.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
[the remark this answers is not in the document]
gf2121: I rethink about this and i agree we should have this check, added :)

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+   */
+  record Output(long fp, boolean hasTerms, BytesRef floorData) {}
+
+  private enum Status {
+    BUILDING,
+    SAVED,
+    DESTROYED
+  }
+
+  private static class Node {
```
mikemccand, line 63: At read-time we do not allocate many of these `Node`s right?  Just the "stack" / recursion array down to the iterator (`Intersect` or `TermsEnum`)'s current position in term space?

Even while building I think we don't have too many `Node` instances since we build each child sub-tree of the trie and then absorb those into parent?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+  private TrieBuilder(BytesRef k, Output v) {
+    if (k.length == 0) {
+      root.output = v;
+      return;
+    }
+    Node parent = root;
+    for (int i = 0; i < k.length; i++) {
+      int b = k.bytes[i + k.offset] & 0xFF;
+      Output output = i == k.length - 1 ? v : null;
+      Node node = new Node(b, output, new LinkedList<>());
```
mikemccand, line 102: Do we always pass as new empty `LinkedList` as children?  If so, maybe remove it from ctor arguments and init to empty `LinkedList` in ctor?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+   */
+  void absorb(TrieBuilder trieBuilder) {
+    if (status != Status.BUILDING || trieBuilder.status != Status.BUILDING) {
+      throw new IllegalStateException("tries should be unsaved");
+    }
+    // Use a simple stack to avoid recursion.
+    Deque<Runnable> stack = new ArrayDeque<>();
+    stack.add(() -> absorb(this.root, trieBuilder.root, stack));
+    while (!stack.isEmpty()) {
+      stack.pop().run();
```
mikemccand, line 123: Hmm isn't this still recursive (will use a java/C stack frame per utf8 digit in the prefix)?  It's just that the recursion is now wrapped in a lambda?  Oh, I see, it is actually iterative!  You are making the tail-recursion explicit by forcing a return of the parent's `absorb` call before then handling the children.  Cool!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+    assert n.label == add.label;
+    if (add.output != null) {
+      if (n.output != null) {
+        this.status = Status.DESTROYED;
+        throw new IllegalArgumentException("Duplicated key found in the given trie.");
+      }
+      n.output = add.output;
+    }
+    ListIterator<Node> iter = n.children.listIterator();
+    // TODO we can do more efficient if there is no intersection, block tree always do that
```
mikemccand, line 138: And also, I think for our usage here, there is never intersection?  We should be absorbing completely orthogonal children each time?  We could maybe `assert` that and make a more efficient impl that simply adds the child sub-tree instead of fully recursing each time?

Hmm in fact I think with this impl we have O(N^2) (where N = length of prefix) cost?  Let's make sure we add a stress test case that adds many long-prefix'd terms?  E.g. indexing URLs as whole-terms (admittedly a weird thing to do!) would tickle this in practice.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+    meta.writeVLong(index.getFilePointer());
+    status = Status.SAVED;
+  }
+
+  void saveNodes(IndexOutput index) throws IOException {
+    final long startFP = index.getFilePointer();
+    Deque<Node> stack = new ArrayDeque<>();
+    stack.push(root);
+
+    // Visit and save nodes of this trie in a post-order traversal.
```
mikemccand, line 198: Maybe add `post-order depth-first traversal` to make it explicit?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+
+    while (stack.isEmpty() == false) {
+      Node node = stack.peek();
+      assert node.fp == -1;
+      final int childrenNum = node.children.size();
+
+      if (childrenNum == 0) {
+        assert node.output != null;
```
mikemccand, line 206: Maybe add `// leaf node` comment?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+  private static int bytesRequiredVLong(long v) {
+    return Long.BYTES - (Long.numberOfLeadingZeros(v | 1) >>> 3);
+  }
+
+  private static void writeLongNBytes(long v, int n, DataOutput out) throws IOException {
+    for (int i = 0; i < n; i++) {
+      out.writeByte((byte) v);
+      v >>= 8;
+    }
+  }
```
mikemccand, line 365: Hmm `assert v == 0`?  We are trusting caller to pass the correct `int n`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+
+  private static int bytesRequiredVLong(long v) {
+    return Long.BYTES - (Long.numberOfLeadingZeros(v | 1) >>> 3);
+  }
+
+  private static void writeLongNBytes(long v, int n, DataOutput out) throws IOException {
+    for (int i = 0; i < n; i++) {
+      out.writeByte((byte) v);
+      v >>= 8;
```
mikemccand, line 363: And maybe `assert v > 0` after this line, also confirming caller gave us the right `int n`.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+              output.floorData.bytes, output.floorData.offset, output.floorData.length);
+        }
+        continue;
+      }
+
+      if (node.childrenIterator == null) {
+        node.childrenIterator = node.children.iterator();
+      }
+      if (node.childrenIterator.hasNext()) {
+        stack.push(node.childrenIterator.next());
```
mikemccand, line 235: Hmm do we assert anywhere while saving that the `label` of each child is in strictly monotonically increasing order?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+
+        // [n bytes] floor data
+        // [n bytes] children fps | [n bytes] position data
+        // [1 byte] children count (if floor data) | [n bytes] encoded output fp | [1 byte] label
+        // [5bit] position bytes | 2bit children strategy | [3bit] encoded output fp bytes
+        // [1bit] has output | [3bit] children fp bytes | [2bit] sign
+
+        final int minLabel = node.children.getFirst().label;
+        final int maxLabel = node.children.getLast().label;
+        PositionStrategy positionStrategy = null;
```
mikemccand, line 285: `assert maxLabel > minLabel`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+        int positionBytes = Integer.MAX_VALUE;
+        for (PositionStrategy strategy : PositionStrategy.values()) {
+          int strategyCost = strategy.positionBytes(minLabel, maxLabel, childrenNum);
+          if (strategyCost < positionBytes) {
+            positionStrategy = strategy;
+            positionBytes = strategyCost;
+          } else if (positionStrategy != null
+              && strategyCost == positionBytes
+              && strategy.priority > positionStrategy.priority) {
```
mikemccand, line 294: We could just statically sort `PositionStrategy` by highest to lowest priority (or, reverse of that) and set the tie-breaking accordingly?  We'd have to also add a unit test asserting that the order is correct.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+        final int minLabel = node.children.getFirst().label;
+        final int maxLabel = node.children.getLast().label;
+        PositionStrategy positionStrategy = null;
+        int positionBytes = Integer.MAX_VALUE;
+        for (PositionStrategy strategy : PositionStrategy.values()) {
+          int strategyCost = strategy.positionBytes(minLabel, maxLabel, childrenNum);
+          if (strategyCost < positionBytes) {
+            positionStrategy = strategy;
+            positionBytes = strategyCost;
+          } else if (positionStrategy != null
```
mikemccand, line 292: Hmm will `positionStrategy` ever be `null`?  Oh I see, if `strategyCost` is ever returned as `Integer.MAX_VALUE`, ok.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+              && strategyCost == positionBytes
+              && strategy.priority > positionStrategy.priority) {
+            positionStrategy = strategy;
+          }
+        }
+
+        assert positionStrategy != null;
+        assert positionBytes > 0 && positionBytes <= 32;
+
+        long maxChildDeltaFp = node.fp - node.children.getFirst().fp;
```
mikemccand, line 302: Maybe add comment explaining `// this is max because first child was written first, and we (parent) at written after all children are written` or so?

This is like index-time joins in Lucene!  Parent comes last in the block of documents ... though there have been raging debates over the years about why they didn't come first :)  Reminds me of [The Butter Battle Book](https://en.wikipedia.org/wiki/The_Butter_Battle_Book)!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+                | (minLabel << 16);
+
+        writeLongNBytes(header, 3, index);
+
+        if (node.output != null) {
+          Output output = node.output;
+          long encodedFp = encodeFP(output);
+          writeLongNBytes(encodedFp, encodedOutputFpBytes, index);
+          if (output.floorData != null) {
+            index.writeByte((byte) (childrenNum - 1));
```
mikemccand, line 323: This is confusing :)  What does one (`floorData`) have to do with the other (`childrenNum`)?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+        positionStrategy.save(node.children, childrenNum, positionBytes, index);
+        assert index.getFilePointer() == positionStartFp + positionBytes
+            : positionStrategy.name()
+                + " position bytes compute error, computed: "
+                + positionBytes
+                + " actual: "
+                + (index.getFilePointer() - positionStartFp);
+
+        for (Node child : node.children) {
+          assert node.fp > child.fp;
```
mikemccand, line 337: Add comment `// parent always written after all children`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+          writeLongNBytes(node.fp - child.fp, childrenFpBytes, index);
```
mikemccand, line 338: Oh, hmmmmm I see, the `numBytes` might actually be bigger than is technically required for a `vLong`... so then one of the asserts I suggested (inside the loop) can't be done.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+  enum PositionStrategy {
+
+    /**
+     * Store children labels in a bitset, this is likely the most efficient storage as we can
+     * compute position with bitCount instruction, so we give it the highest priority.
+     */
+    BITS(2) {
+      @Override
+      int positionBytes(int minLabel, int maxLabel, int labelCnt) {
+        int byteDistance = maxLabel - minLabel + 1;
```
mikemccand, line 376: Let's also add a test case that exercises the full binary space of terms (outside of valid UTF8)?  Which would make this `byteDistance` `256`.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+  }
+
+  private static void writeLongNBytes(long v, int n, DataOutput out) throws IOException {
+    for (int i = 0; i < n; i++) {
+      out.writeByte((byte) v);
+      v >>= 8;
+    }
+  }
+
+  enum PositionStrategy {
```
mikemccand, line 367: Maybe rename to `ChildSaveStrategy` or so?  `Positions` is overloaded term in Luceneland!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+          }
+        }
+        return targetLabel - minLabel - low;
+      }
+    };
+
+    private static final PositionStrategy[] STRATEGIES = new PositionStrategy[3];
+
+    static {
+      for (PositionStrategy strategy : PositionStrategy.values()) {
```
mikemccand, line 528: Oh I see, we are statically building this already in priority order (so my comment above about tie-break should work).

Can we add `assert STRATEGIES[strategy.priority] == null` before assigning?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,552 @@
...
+    abstract int positionBytes(int minLabel, int maxLabel, int labelCnt);
+
+    abstract void save(List<Node> children, int labelCnt, int positionBytes, IndexOutput output)
+        throws IOException;
+
+    abstract int lookup(
+        int targetLabel, RandomAccessInput in, long offset, int positionBytes, int minLabel)
+        throws IOException;
+
+    static PositionStrategy byCode(int code) {
```
mikemccand, line 548: Hmm so this `code`/`priority` is written into the index and must not change its meaning across releases?  E.g. we cannot suddenly insert a higher priority strategy on a dot-release or major release since that would renumber the enum values here and break bwc?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieReader.java
```diff
@@ -0,0 +1,228 @@
...
+        0xFFFFL,
+        0xFFFFFFL,
+        0xFFFFFFFFL,
+        0xFFFFFFFFFFL,
+        0xFFFFFFFFFFFFL,
+        0xFFFFFFFFFFFFFFL,
+        0xFFFFFFFFFFFFFFFFL
+      };
+
+  static class Node {
```
mikemccand, line 39: Oh!  Two different `Node` classes, OK.

mikemccand (commented):
This is looking nice!  Thank you @gf2121, what a tricky and vital part of Lucene's Codec.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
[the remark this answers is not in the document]
mikemccand: Ahhh, OK I see, it is like a merge sort ("zipper" two iterators together).  I agree it's O(M+N).

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
```diff
@@ -0,0 +1,486 @@
...
+    // TODO we can do more efficient if there is no intersection, block tree always do that
+    outer:
+    for (Node addChild : add.children) {
+      while (iter.hasNext()) {
+        Node nChild = iter.next();
+        if (nChild.label == addChild.label) {
+          putAll(nChild, addChild);
+          continue outer;
+        }
+        if (nChild.label > addChild.label) {
```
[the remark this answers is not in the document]
mikemccand: Maybe document and `assert` this somewhere?  I think it's non-obvious ... it's the merge sorting in `absorb` that makes it so I think?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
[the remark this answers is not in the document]
mikemccand: +1 to design for today!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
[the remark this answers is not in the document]
mikemccand: Let's explicitly test massive terms sent to `TrieBuilder`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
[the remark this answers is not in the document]
mikemccand: Yay :)  This is my favorite term in open source community.  It's so important yet so hard for old-timers ("API blindness") to achieve ... we forget what it is like to be a new person learning or developing Lucene, and yet those new experiences are so crucial in guiding our development / filing off the warts.  It is a "sharp tool that quickly dulls".

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
```diff
@@ -0,0 +1,486 @@
...
+          unCompiled = child;
+          break;
+        }
+      }
+      if (unCompiled != null) {
+        stack.push(unCompiled);
+        continue;
+      }
+
+      node.fp = index.getFilePointer() - startFP;
```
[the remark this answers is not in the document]
mikemccand: Hmm I see the thumbs up but I don't see this comment in the latest rev ...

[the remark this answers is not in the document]
gf2121: It's here:

https://github.com/apache/lucene/blob/13d5f4bf8b9a52b046b1c39f3a21f9c1443760fc/lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java#L239-L242

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/Trie.java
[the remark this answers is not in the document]
gf2121: +1, so important!!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
gf2121: Double yes on read / write side!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
gf2121: As talked another place, this is a O(M+N), so i'd leave a todo to follow up PR.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
gf2121: As you pointed in another comment, this assertion can not be added :)

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
gf2121: Good point. 
I'm considering make `priority` only a `code`, declare strategy enums in priority order, and have a test for this.
In this way, code is just to identify which strategy is used, and we can easily define a new strategy in any priority.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieReader.java
gf2121: Yeah, `TrieBuilder.Node` and `TrieReader.Node`. I think the class prefix has made it clear :)

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
gf2121: Tests on the trie are all generating random bytes (no binding to UTF8)

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
gf2121: We need this `childrenNum` to compute where the floor data start, i add a comment.

https://github.com/apache/lucene/blob/13d5f4bf8b9a52b046b1c39f3a21f9c1443760fc/lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieReader.java#L171-L175

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/IntersectTermsEnumFrame.java
```diff
@@ -89,8 +89,6 @@ final class IntersectTermsEnumFrame {
 
   final ByteArrayDataInput bytesReader = new ByteArrayDataInput();
 
-  int outputNum;
```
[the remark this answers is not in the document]
gf2121: This was previously used, in IntersectTermsEnum.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
gf2121: Oh sorry i miss read your comment! i'll try to get out the more efficient impl.
One thing makes me a bit unhappy is `java.util.LinkedList` can not easily append one to another in one single operation, so we may need something new here.

gf2121: Done :)

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
gf2121: As talked in another comment, i plan to introduce an array like 
`STRATEGIES_IN_PRIORITY_ORDER = new ChildSaveStrategy[] {BITS, ARRAY, REVERSE_ARRAY}` to define priority, and the code is just a code to identify which strategy used, serialized to disk and used on reading side.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
mikemccand: If we haven't already (I will try to review latest PR soon!), can we add a comment making this "we sometimes must write 0 bytes" explicit?  I.e. explain why these are not "true"/normal `vLong` we are writing.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,632 @@
...
+import org.apache.lucene.store.DataOutput;
+import org.apache.lucene.store.IndexOutput;
+import org.apache.lucene.store.RandomAccessInput;
+import org.apache.lucene.util.BytesRef;
+import org.apache.lucene.util.BytesRefBuilder;
+
+/**
+ * A builder to build prefix tree (trie) as the index of block tree, and can be saved to disk.
+ *
+ * <p>TODO make it a more memory efficient structure
```
mikemccand, line 33: Maybe tweak comment to make it clear this `TODO` is only about the builder, which only runs during indexing?  The search-time data structure is already efficient -- smaller than FSTs!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,632 @@
...
+    }
+  }
+
+  /**
+   * Absorb all (K, V) pairs from the given trie into this one. The given trie builder need to
+   * ensure its keys greater or equals than max key of this one.
+   *
+   * <p>Note: the given trie will be destroyed after absorbing.
+   */
+  void absorb(TrieBuilder trieBuilder) {
```
mikemccand, line 121: Maybe rename to `append`?  The two tries are strictly orthogonal, and, the incoming trie is > `this` one?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,632 @@
...
+    this.maxKey = trieBuilder.maxKey;
+    trieBuilder.status = Status.DESTROYED;
+  }
+
+  Output getEmptyOutput() {
+    return root.output;
+  }
+
+  /**
+   * Used for tests only. The recursive impl need to be avoided if someone plans to use for
```
mikemccand, line 176: ++ thank you for the comment!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,632 @@
...
+        index.writeByte(((byte) header));
+        writeLongNBytes(output.fp, outputFpBytes, index);
+        if (output.floorData != null) {
+          index.writeBytes(
+              output.floorData.bytes, output.floorData.offset, output.floorData.length);
+        }
+        continue;
+      }
+
+      // If there are any children have not been saved, push it into stack and continue.
```
mikemccand, line 250: `push it` -> `push them`

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,632 @@
...
+        | (output.fp << 2);
+  }
+
+  private static int bytesRequiredVLong(long v) {
+    return Long.BYTES - (Long.numberOfLeadingZeros(v | 1) >>> 3);
+  }
+
+  private static void writeLongNBytes(long v, int n, DataOutput out) throws IOException {
+    for (int i = 0; i < n; i++) {
+      out.writeByte((byte) v);
```
mikemccand, line 381: Let's add the comment here?  Something like:
```
// note that we sometimes write trailing 0 bytes here, when the incoming int n is bigger than
// would be required for a "normal" vLong
```

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,632 @@
...
+    return (output.floorData != null ? NON_LEAF_NODE_HAS_FLOOR : 0)
+        | (output.hasTerms ? NON_LEAF_NODE_HAS_TERMS : 0)
+        | (output.fp << 2);
+  }
+
+  private static int bytesRequiredVLong(long v) {
+    return Long.BYTES - (Long.numberOfLeadingZeros(v | 1) >>> 3);
+  }
+
+  private static void writeLongNBytes(long v, int n, DataOutput out) throws IOException {
```
mikemccand, line 379: And maybe also add comment here like:
```
/** This differs from writeVLong because it can write more bytes than would be
    needed for vLong when the incoming int n is larger. */
```

lucene/core/src/java/org/apache/lucene/codecs/lucene90/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,632 @@
...
+      assert node.lastChild == null;
+    } else if (node.childrenNum == 1) {
+      assert node.firstChild == node.lastChild;
+      assert node.firstChild.next == null;
+    } else if (node.childrenNum > 1) {
+      int n = 0;
+      for (Node child = node.firstChild; child != null; child = child.next) {
+        n++;
+        assert child.next == null || child.label < child.next.label
+            : " the label of children nodes should always be in order.";
```
mikemccand, line 399: Maybe `in strictly increasing order`?

mikemccand (commented):
This is looking close -- I left some minor wording / comments this time.

Are there any major items / blockers?  Do all Lucene tests pass if you force tests to use this in their Codec?  Maybe beast things a bit (run nightly, increase `TEST_ITERS`, etc.) :)

Once we think this is ready, we should prolly merge at first as the non-default Codec, and let CI chew on things / let it bake for a while in `main`?  If all goes well, then maybe at some point we backport and make it default for a 10.x?

jpountz:
> Once we think this is ready, we should prolly merge at first as the non-default Codec

We should add this format to `RandomCodec` then, so that it gets included as part of codec randomization. Otherwise it will only be exercised in `TestXXXPostingsFormat`.

We just cut branch_10_2, so this should give us some time to let it bake while still having it released in 10.3.

gf2121:
Thank you very much for all these careful, warm and helpful comments!

> Are there any major items / blockers?

I think I've addressed all of them (hopefully didn't miss any).

> Do all Lucene tests pass if you force tests to use this in their Codec?

The `Lucene90BlockTreeTermsWriter` is now used by default codec so tests should catch it easily. All lucene core tests passed.

> Maybe beast things a bit (run nightly, increase TEST_ITERS, etc.) :)

I've finished a round of `./gradlew :lucene:core:test -Ptests.nightly=true -Ptests.monster=true` and all tests passed. I'll run more.

> Once we think this is ready, we should prolly merge at first as the non-default Codec, and let CI chew on things / let it bake for a while in main? If all goes well, then maybe at some point we backport and make it default for a 10.x?

I understand the idea that merge it into main but not backport immediately. But i do not know how to make CI catch this change and test it heavily without making it a default codec in main? Or you are meaning not backport to 10.x, but make it a default codec in main? (Sorry for my pool English!)

gf2121:
> We should add this format to RandomCodec then, so that it gets included as part of codec randomization.

OK, did not see this. I know how to do it then. Thanks Adrien :)

gf2121:
I roughly implemented the idea. This is my first time forking a new codec, hopefully have not made too many mistakes :)

A few thoughts during my refactoring:

* I thought i only need to fork a `Lucene103BlockTreeTerms` to intersect with `Lucene101Postings`. But it seems challenging based on current API design. I have to fork the new `Lucene103Postings` as well. Maybe this is a point can be improved?

* I'm not sure if it matters to make it a default codec or not in main, as main will not get released anyway? Default to main without backporting sounds good enough to me. If we stick not to make it default, maybe this codec should be moved into test / sandbox / codec module? It would be weird to see multiple codecs in a core module for contributors who don't have the context of this PR.

jpountz:
> I thought i only need to fork a Lucene103BlockTreeTerms to intersect with Lucene101Postings. But it seems challenging based on current API design. I have to fork the new Lucene103Postings as well. Maybe this is a point can be improved?

Right, this is because a Lucene `PostingsFormat` encapsulates both the format of the terms dictionary and of postings lists. I guess that we could make terms dictionaries and postings formats pluggable, but that could make it harder to do interesting things such as inlining data about postings into the terms dictionary.

> f we stick not to make it default, maybe this codec should be moved into test / sandbox / codec module?

I don't have a strong opinion about making it the default right now in main or not, but @mikemccand seemed to care about not making it the default yet. In this case, I would suggest having it in lucene/core, hopefully in one or two weeks it becomes part of the default codec.

jpountz (approved):
Changes look good to me.

jpountz (approved):

mikemccand:
> > f we stick not to make it default, maybe this codec should be moved into test / sandbox / codec module?
> 
> I don't have a strong opinion about making it the default right now in main or not, but @mikemccand seemed to care about not making it the default yet. In this case, I would suggest having it in lucene/core, hopefully in one or two weeks it becomes part of the default codec.

+1 -- let's leave it in core.  We should plan to make it the default in a couple weeks ... I'd just like to bake it a bit on `main` and see if Jenkins uncovers any fun surprises?  With the gains it offers we should absolutely plan to use this in the default codec!

lucene/core/src/java/org/apache/lucene/codecs/lucene103/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,640 @@
...
+
+  /**
+   * Append all (K, V) pairs from the given trie into this one. The given trie builder need to
+   * ensure its keys greater or equals than max key of this one.
+   *
+   * <p>Note: the given trie will be destroyed after appending.
+   */
+  void append(TrieBuilder trieBuilder) {
+    if (status != Status.BUILDING || trieBuilder.status != Status.BUILDING) {
+      throw new IllegalStateException("tries have wrong status.");
```
mikemccand, line 123: Maybe include the wrong status in the exception message?

lucene/core/src/java/org/apache/lucene/codecs/lucene103/blocktree/TrieBuilder.java
```diff
@@ -0,0 +1,640 @@
...
+    for (int i = 0; i < mismatch; i++) {
+      final Node aLast = a.lastChild;
+      final Node bFirst = b.firstChild;
+      assert aLast.label == bFirst.label;
+
+      if (b.childrenNum > 1) {
+        aLast.next = bFirst.next;
+        a.childrenNum += b.childrenNum - 1;
+        a.lastChild = b.lastChild;
+        assertChildrenLabelInOrder(a);
```
mikemccand, line 147: Should this be `assert assertChildrenLabelInOrder(a);`?  So that we don't even call an empty method when assertions are disabled?

mikemccand (approved):
This looks great!  I left some tiny polish comments.

Let's also make this the default codec for `main`?

lucene/core/src/java/org/apache/lucene/codecs/lucene103/blocktree/TrieBuilder.java
gf2121: Oops, yes!

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene101/PostingDecodingUtil.java
```diff
@@ -0,0 +1,57 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.backward_codecs.lucene101;
+
+import java.io.IOException;
+import org.apache.lucene.store.IndexInput;
+
+/** Utility class to decode postings. */
+class PostingDecodingUtil {
```
gf2121, line 23: I do not think we should expose `org.apache.lucene.internal.vectorization` to help bwc. So i just forked the scalar version out. Correct me if i played it wrong :)

gf2121:
I am seeing all tests in `TestInt7HnswBackwardsCompatibility` get failed:

```
org.apache.lucene.index.CorruptIndexException: codec mismatch: actual codec=Lucene101PostingsWriterMeta vs expected codec=Lucene103PostingsWriterMeta (resource=BufferedChecksumIndexInput(MockIndexInputWrapper(NIOFSIndexInput(path="/Users/runner/work/lucene/lucene/lucene/backward-codecs/build/tmp/tests-tmp/lucene.backward_index.TestInt7HnswBackwardsCompatibility_89534029F176782-001/tempDir-002/_0.cfs") [slice=_0_Asserting_0.psm])))
   >         at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.codecs.CodecUtil.checkHeaderNoMagic(CodecUtil.java:205)
   >         at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.codecs.CodecUtil.checkHeader(CodecUtil.java:194)
   >         at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.codecs.CodecUtil.checkIndexHeader(CodecUtil.java:254)
   >         at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.codecs.lucene103.Lucene103PostingsReader.<init>(Lucene103PostingsReader.java:97)
   >         at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.codecs.lucene103.Lucene103PostingsFormat.fieldsProducer(Lucene103PostingsFormat.java:428)
   >         at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.codecs.asserting.AssertingPostingsFormat.fieldsProducer(AssertingPostingsFormat.java:54)
   >         at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.codecs.perfield.PerFieldPostingsFormat$FieldsReader.<init>(PerFieldPostingsFormat.java:319)
   >         at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.codecs.perfield.PerFieldPostingsFormat.fieldsProducer(PerFieldPostingsFormat.java:381)
   >         at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.SegmentCoreReaders.<init>(SegmentCoreReaders.java:98)
   >         at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.SegmentReader.<init>(SegmentReader.java:95)
   >         at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CheckIndex.testSegment(CheckIndex.java:1037)
   >         at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CheckIndex.checkIndex(CheckIndex.java:822)
   >         at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CheckIndex.checkIndex(CheckIndex.java:592)
   >         at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.util.TestUtil.checkIndex(TestUtil.java:338)
   >         at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.store.MockDirectoryWrapper.close(MockDirectoryWrapper.java:921)
   >         at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.util.IOUtils.close(IOUtils.java:85)
   >         at app/org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.util.IOUtils.close(IOUtils.java:72)
   >         at app//org.apache.lucene.backward_index.BackwardsCompatibilityTestBase.tearDown(BackwardsCompatibilityTestBase.java:156)
   >         at java.base@23.0.2/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
   >         at java.base@23.0.2/java.lang.reflect.Method.invoke(Method.java:580)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.RandomizedRunner.invoke(RandomizedRunner.java:1758)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.RandomizedRunner$10.evaluate(RandomizedRunner.java:1004)
   >         at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.util.TestRuleSetupTeardownChained$1.evaluate(TestRuleSetupTeardownChained.java:48)
   >         at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.util.AbstractBeforeAfterRule$1.evaluate(AbstractBeforeAfterRule.java:43)
   >         at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.util.TestRuleThreadAndTestName$1.evaluate(TestRuleThreadAndTestName.java:45)
   >         at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.util.TestRuleIgnoreAfterMaxFailures$1.evaluate(TestRuleIgnoreAfterMaxFailures.java:60)
   >         at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.util.TestRuleMarkFailure$1.evaluate(TestRuleMarkFailure.java:44)
   >         at app/junit@4.13.1/org.junit.rules.RunRules.evaluate(RunRules.java:20)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.ThreadLeakControl$StatementRunner.run(ThreadLeakControl.java:390)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.ThreadLeakControl.forkTimeoutingTask(ThreadLeakControl.java:843)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.ThreadLeakControl$3.evaluate(ThreadLeakControl.java:490)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.RandomizedRunner.runSingleTest(RandomizedRunner.java:955)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.RandomizedRunner$5.evaluate(RandomizedRunner.java:840)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.RandomizedRunner$6.evaluate(RandomizedRunner.java:891)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.RandomizedRunner$7.evaluate(RandomizedRunner.java:902)
   >         at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.util.AbstractBeforeAfterRule$1.evaluate(AbstractBeforeAfterRule.java:43)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
   >         at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.util.TestRuleStoreClassName$1.evaluate(TestRuleStoreClassName.java:38)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.rules.NoShadowingOrOverridesOnMethodsRule$1.evaluate(NoShadowingOrOverridesOnMethodsRule.java:40)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.rules.NoShadowingOrOverridesOnMethodsRule$1.evaluate(NoShadowingOrOverridesOnMethodsRule.java:40)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
   >         at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.util.TestRuleAssertionsRequired$1.evaluate(TestRuleAssertionsRequired.java:53)
   >         at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.util.AbstractBeforeAfterRule$1.evaluate(AbstractBeforeAfterRule.java:43)
   >         at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.util.TestRuleMarkFailure$1.evaluate(TestRuleMarkFailure.java:44)
   >         at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.util.TestRuleIgnoreAfterMaxFailures$1.evaluate(TestRuleIgnoreAfterMaxFailures.java:60)
   >         at app/org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.util.TestRuleIgnoreTestSuites$1.evaluate(TestRuleIgnoreTestSuites.java:47)
   >         at app/junit@4.13.1/org.junit.rules.RunRules.evaluate(RunRules.java:20)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.ThreadLeakControl$StatementRunner.run(ThreadLeakControl.java:390)
   >         at app/randomizedtesting.runner@2.8.1/com.carrotsearch.randomizedtesting.ThreadLeakControl.lambda$forkTimeoutingTask$0(ThreadLeakControl.java:850)
   >         at java.base@23.0.2/java.lang.Thread.run(Thread.java:1575)
  2> NOTE: reproduce with: gradlew test --tests TestInt7HnswBackwardsCompatibility.testReadOldIndices -Dtests.seed=89534029F176782 -Dtests.locale=ff-Adlm-SL -Dtests.timezone=Africa/Blantyre -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```

lucene/core/src/java/org/apache/lucene/codecs/Codec.java
```diff
@@ -56,7 +56,7 @@ static NamedSPILoader<Codec> getLoader() {
     }
 
     @SuppressWarnings("NonFinalStaticField")
-    static Codec defaultCodec = LOADER.lookup("Lucene101");
+    static Codec defaultCodec = LOADER.lookup("Lucene103");
```
mikemccand, line 59: Yay!  This means default Codec is now using trie!

mikemccand (approved):
This looks great!  Thank you for persisting @gf2121!

lucene/core/src/java/org/apache/lucene/codecs/Codec.java
dweiss: I think this is causing the OOMs. See #14487

gf2121:
Nightly confirmed the speedup.

https://benchmarks.mikemccandless.com/2025.04.14.18.05.20.html
https://benchmarks.mikemccandless.com/PKLookup.html

Annotation PR: https://github.com/mikemccand/luceneutil/pull/370

mikemccand:
> Nightly confirmed the speedup.

That's ~1.2 M primary key lookups per second, yay!

PK lookups are used all the time, not just the "key/value store" use case... e.g. updating or deleting a document by its id.

msokolov:
I saw a message raising concern about some OOM test failures possibly related to this? Am I confused? Is that a potential problem?

stefanvodita:
If we think the issue last week was with the tests, should we go ahead and back-port this change?

gf2121:
Hi @stefanvodita 

We plan to allow CI chew on this change for a couple of weeks and backport it if everything goes well. See https://github.com/apache/lucene/pull/14333#issuecomment-2789601238.

When backporting, i'll try to catch all changes made to `Lucene103Codec`, including https://github.com/apache/lucene/pull/14447. So for now, feel free to just merge it into main :)

stefanvodita:
You anticipated my concern with #14447 😄 
Thank you for handling this!

Coqueue:
Thanks for the fantastic change!

Want to share that we adopted and backported this codec to Lucene 912, and ran it against an Amazon Search internal benchmark, from which we observed an increase of 2.7% in Searcher throughput : D

Looking forward to seeing the actual improvements this brings in production!

mikemccand:
Thank you @Coqueue.

> ran it against an Amazon Search internal benchmark, from which we observed an increase of 2.7% in Searcher throughput : D

Small correction: Amazon Product Search internal benchmarks.

Also, we did this backport just to evaluate performance impact to our service from this change since we so heavily use Lucene's terms dictionary, but for actually running this in production we will upgrade to 10.x.

Hmm this change doesn't have the milestone set -- I'll set it.

jpountz:
This is great info, thanks for sharing!

PR #14335, opened 2025-03-07
Optimize ConcurrentMergeScheduler for Multi-Tenant Indexing

### LINKED ISSUE #13883: A multi-tenant ConcurrentMergeScheduler
### Description

`ConcurrentMergeScheduler` computes max thread counts assuming a single `IndexWriter` in the JVM. But it's common with Solr or Elasticsearch to have tens of active `IndexWriter`s running in the same JVM. Could we make `ConcurrentMergeScheduler` more multi-tenant, or introduce a new multi-tenant merge scheduler that accounts for the fact that the JVM is shared across multiple indexes?

DivyanshIITB:
This PR enhances the ConcurrentMergeScheduler by introducing dynamic resource allocation for multi-tenant indexing scenarios.

Key Improvements:
1) Global Counter for Active IndexWriters: Introduced an AtomicInteger to track active IndexWriter instances dynamically.
2) Dynamic Thread Allocation: Adjusted maxThreadCount based on the number of active writers to optimize merge performance.
3) Adaptive Merge Throttling: Distributed merge speed limits among active writers to prevent resource contention and ensure fair indexing performance.

These changes improve efficiency in multi-writer environments by dynamically adapting merge scheduling to system load.

Fixes #13883


jpountz:
This looks too naive to me, we don't want each index writer to have 1/N of the resources, which would prevent one writer from maxing out resources (e.g. if one index has a heavy write load while other indexes are only rarely updated).

DivyanshIITB:
Thank you for the review, @jpountz!

I see your concern regarding equal resource distribution across IndexWriter instances potentially leading to inefficiencies when some writers have a significantly heavier workload than others. My initial goal was to ensure fairness in multi-tenant environments, but I agree that a static 1/N allocation might not be ideal.

To address this, I’m considering an adaptive resource allocation approach, where:

- Recent Merge Activity Tracking: Each IndexWriter's recent merge activity is monitored.
- Dynamic Thread Allocation: Instead of an equal split, threads are assigned based on recent workload.
- Merge Throttling Balance: Prevents a single heavy writer from starving others while ensuring underutilized writers don’t hold 
   excessive resources.

Would this approach align better with Lucene’s design goals?

jpountz:
Merge throttling is now disabled by default, IMO it's fine to ignore merge throttling for now. Regarding thread creation, I'm thinking of a shared fixed (e.g. number of processors / 2) thread pool for merge scheduling.

DivyanshIITB:
Just a gentle reminder

jpountz:
Apologies I had missed your reply.

> should this be a shared global pool across all IndexWriters, or should each writer have its own pool?

It should be shared, we don't want the total number of threads to scale with the number of index writers. The reasoning for the numProcessors/2 number is that merging generally should not be more expensive than indexing, so by reserving only half the CPU capacity for merging, it should still be possible to max out hardware while indexing, while also having a peak number of threads running merges under numProcessors/2.

> If it's a shared pool, how should we handle cases where a few writers are highly active while others are idle? Should we allow active writers to take more resources dynamically, or keep a strict fixed allocation?

Idle writers would naturally submit fewer tasks than highly active writers. IMO the fixed allocation is key here.

DivyanshIITB:
Thanks for the detailed clarification, @jpountz!

I have made the necessary changes to the implementation:

- Implemented a shared global thread pool with a fixed size of numProcessors / 2.
- Ensured that each IndexWriter submits merge tasks to the shared pool without dynamic reallocation, as idle writers naturally 
   submit fewer tasks.
- Updated the logic in MergeScheduler.java to align with these changes.

@jpountz, I’ve made the requested changes. Please review when you get a chance.

jpountz:
I think it'll be simpler to create a new merge scheduler rather than modify ConcurrentMergeScheduler. Also we'll need tests.

DivyanshIITB:
Thanks for your feedback, @jpountz!

I have created a new MultiTenantMergeScheduler as suggested, instead of modifying ConcurrentMergeScheduler.
I have restored ConcurrentMergeScheduler to its original state.
Added a test case (TestMultiTenantMergeScheduler) to validate the new scheduler's behavior.
Please review and let me know if any further changes are needed!

lucene/core/src/java/org/apache/lucene/index/MultiTenantMergeScheduler.java
```diff
@@ -0,0 +1,44 @@
...
+            try {
+                future.get(); // Ensure the task completes
+            } catch (Exception e) {
+                throw new IOException("Merge operation failed", e);
+            }
+        }
+    }
+
+    @Override
+    public void close() {
```
vigyasharma, line 41: This shouldn't be a no-op. We should wait for all running merges of the calling IndexWriter to complete, similar to `sync()` in ConcurrentMergeScheduler. This could be done by maintaining a per writer list of futures for running merges. We could run a clean up loop in `merge()` that removes completed futures.

Also, I get that this thread pool is global and shared, but I don't like the idea of having a threadpool that can never be closed. How about we expose some hooks to close this scheduler if all consuming writers have closed?

lucene/core/src/test/org/apache/lucene/index/TestMultiTenantMergeScheduler.java
```diff
@@ -0,0 +1,30 @@
...
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.RAMDirectory;
+import org.apache.lucene.analysis.standard.StandardAnalyzer;
+import org.apache.lucene.document.Document;
+import org.apache.lucene.document.TextField;
+import org.apache.lucene.util.LuceneTestCase;
+
+public class TestMultiTenantMergeScheduler extends LuceneTestCase {
+
+    public void testMultiTenantMergeScheduler() throws Exception {
```
vigyasharma, line 12: This test is not sufficient. We need to test (and also implement) a lot more scenarios, like multiple writers consuming this scheduler, scheduling different merges and starting/closing at different times. You can refer to the tests in `TestConcurrentMergeScheduler` as a starting point.

vigyasharma (changes requested):
@DivyanshIITB : I appreciate your interest in this issue. A multi-tenant CMS however, has quite a few nuances that this implementation will need to cover. I'd encourage you to look at the different ways in which IndexWriter interacts with merging. Some examples of functionality I find missing include support for forceMerging, aborting a merge, and waiting till all merges complete.

I can help review future iterations that include these changes.

lucene/core/src/java/org/apache/lucene/index/MultiTenantMergeScheduler.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.concurrent.Future;
+import java.io.IOException;
+
+/**
+ * A multi-tenant merge scheduler that shares a global thread pool across multiple IndexWriters.
+ */
+public class MultiTenantMergeScheduler extends MergeScheduler {
+
+    // Shared global thread pool
+    private static final ExecutorService MERGE_THREAD_POOL = Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors() / 2);
```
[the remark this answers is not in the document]
DivyanshIITB: Thanks for the suggestion! I've updated the implementation to use lazy static initialization by wrapping the thread pool in a nested static class. This ensures the thread pool is only created when first accessed, avoiding unnecessary resource allocation. Let me know if you have any further feedback! Now, instead of directly initializing MERGE_THREAD_POOL, I'm using a LazyHolder class with a getMergeThreadPool() method to ensure thread-safe lazy initialization. This should prevent the thread pool from being created when the policy is unused."

lucene/core/src/java/org/apache/lucene/index/MultiTenantMergeScheduler.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Override
+    public void merge(MergeScheduler.MergeSource mergeSource, MergeTrigger trigger) throws IOException {
+        while (true) {
+            MergePolicy.OneMerge merge = mergeSource.getNextMerge();
+            if (merge == null) {
```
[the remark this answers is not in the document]
DivyanshIITB: Thanks for pointing this out! I’ve updated the loop condition to use hasPendingMerges() instead of relying on null. This ensures better adherence to the documented API behavior. Let me know if this aligns with what you were suggesting!

lucene/core/src/java/org/apache/lucene/index/MultiTenantMergeScheduler.java
```diff
@@ -0,0 +1,44 @@
...
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
+            });
+
+            try {
+                future.get(); // Ensure the task completes
+            } catch (Exception e) {
+                throw new IOException("Merge operation failed", e);
+            }
```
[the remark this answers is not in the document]
DivyanshIITB: Thanks for the feedback! You're absolutely right—blocking on future.get(); was making the merges sequential. I've removed it so that merges can now proceed asynchronously in the background. Let me know if this aligns with the intended behavior!

lucene/core/src/java/org/apache/lucene/index/MultiTenantMergeScheduler.java
DivyanshIITB: Thanks for pointing this out! I’ve updated close() to properly wait for all running merges, similar to sync() in ConcurrentMergeScheduler. Now, merge() tracks active tasks, and I've added cleanup for completed futures. Also, I introduced shutdownThreadPool() to allow safe global shutdown. Let me know if this works!

lucene/core/src/test/org/apache/lucene/index/TestMultiTenantMergeScheduler.java
DivyanshIITB: Thanks for the feedback! I’ve expanded the test suite to include multiple writers consuming the scheduler, different merge scheduling times, and staggered start/close behavior. Also, I’ve referred to TestConcurrentMergeScheduler for structuring test cases to verify concurrency. Let me know if you have additional suggestions!

DivyanshIITB:
I have a request to you. Kindly ignore the following two deleted files in the "Files Changed" section :
"KeepOnlyLastCommitDeletionPolicy.java"
"ConcurrentMergeScheduler.java"

lucene/core/src/java/org/apache/lucene/index/MultiTenantMergeScheduler.java
```diff
@@ -0,0 +1,72 @@
...
+                }
+            });
+
+            // Cleanup completed merges
+            activeMerges.removeIf(Future::isDone);
+        }
+    }
+
+    @Override
+    public void close() throws IOException {
```
mccullocht, line 48: IIRC `IndexWriter.close()` will call `MergeScheduler.close()` which means that you would wait for all merges across all indexes using this scheduler to complete. The `IndexWriter` that is currently calling `close()` will stop scheduling new merges, but the other writers may not, which means this may not terminate. To fix this you may need to map active merges back to the `IndexWriter` or `Directory` they belong to.

lucene/core/src/java/org/apache/lucene/index/MultiTenantMergeScheduler.java
```diff
@@ -0,0 +1,72 @@
...
+    private static class LazyHolder {
+        static final ExecutorService MERGE_THREAD_POOL = 
+            Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors() / 2);
+    }
+
+    private static ExecutorService getMergeThreadPool() {
+        return LazyHolder.MERGE_THREAD_POOL;
+    }
+
+    // Use getMergeThreadPool() instead of direct access
```
vigyasharma, line 23: We can skip the `getMergeThreadPool()` function since we only need to access the threadpool internally. Something like:
```java
  private static class MergeThreadPool {
      private static final ExecutorService INSTANCE = 
          Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors() / 2);
  }
  // and then just use MergeThreadPool.INSTANCE inside merge() and other functions.
  // ...
}
```

lucene/core/src/java/org/apache/lucene/index/MultiTenantMergeScheduler.java
```diff
@@ -0,0 +1,72 @@
...
+            MERGE_THREAD_POOL.submit(() -> {
+                try {
+                    mergeSource.merge(merge);
+                } catch (IOException e) {
+                    throw new RuntimeException("Merge operation failed", e);
+                }
+            });
+
+            // Cleanup completed merges
+            activeMerges.removeIf(Future::isDone);
```
vigyasharma, line 43: Where is this declared and initialized? When are merges added to `activeMerges` ?

lucene/core/src/java/org/apache/lucene/index/MultiTenantMergeScheduler.java
vigyasharma: +1; my suggestion in the previous review was similar. We need a mapping of writer to merges and only block on merges specific to the calling writer.

lucene/core/src/test/org/apache/lucene/index/TestMultiTenantMergeScheduler.java
```diff
@@ -0,0 +1,73 @@
...
+        long startTime = System.currentTimeMillis();
+        writer.forceMerge(1);
+        long endTime = System.currentTimeMillis();
+
+        writer.close();
+        scheduler.close();
+        MultiTenantMergeScheduler.shutdownThreadPool();
+
+        // Check if merging took less time than sequential execution would
+        assertTrue("Merges did not happen concurrently!", (endTime - startTime) < 5000);
```
vigyasharma, line 69: It's not reliable to assert or depend on elapsed time. Lucene is run of many different machines and this timing will differ.

lucene/core/src/java/org/apache/lucene/index/MultiTenantMergeScheduler.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    private final ConcurrentHashMap<IndexWriter, List<Merge>> activeMerges = new ConcurrentHashMap<>();
+
+    @Override
+    public void close() throws IOException {
+        IndexWriter currentWriter = getCurrentIndexWriter();  // Method to get the calling writer
```
vigyasharma, line 47: This is not defined anywhere!

lucene/core/src/java/org/apache/lucene/index/MultiTenantMergeScheduler.java
```diff
@@ -0,0 +1,70 @@
...
+        List<Merge> merges = activeMerges.getOrDefault(currentWriter, Collections.emptyList());
+    
+        for (Merge merge : merges) {
+            merge.waitForCompletion(); // Only wait for merges related to this writer
+        }
+    
+        activeMerges.remove(currentWriter); // Cleanup after closing
```
vigyasharma, line 54: This doesn't line up, `activeMerges` is a List, you're calling Map APIs on it. Did this compile?

vigyasharma (changes requested):
Hi @DivyanshIITB, 

Thanks for your work here, a concurrent multi-tenant merge scheduler is an ambitious feature. This area of the codebase, especially dealing with how `IndexWriter` interacts with merging, can be quite complex. 

I'd suggest the following steps to help things along:
1. First, please ensure your code compiles and all existing tests pass. You can run `./gradlew check` to verify this.
2. A helpful approach would be to try swapping out the existing `ConcurrentMergeScheduler` (CMS) with your `MultiTenantMergeScheduler` (MTMS) in your local branch and running the test suite. This can reveal some subtle edge cases that your scheduler will need to handle.
3. You can use the existing tests for CMS as a guide for writing comprehensive tests for your MTMS.

This is a challenging task, and building familiarity with the codebase through smaller contributions can be very beneficial. Would you be interested in exploring some smaller, more focused issues to start? 

Appreciate your effort and here to help.

DivyanshIITB:
Thankyou for your help @vigyasharma !
I would love to explore small and more focused issues to start with !

PR #14337, opened 2025-03-10
Make single value BKDReader instances lighter

original-brownbear:
Found this case in Elasticsearch heap dump's and it'as the same as #13618 but for BKDReader instances. Admittedly, as for terms this is a corner case but observed use cases where the distribution of these values is heavily weighted towards a single value would save up to a O(5M) in heap and some cache pressure from this change.


iverase (approved):
LGTM

PR #14338, opened 2025-03-10
deduplicate standard BKDConfig records

iverase:
In the case you have many BKD readers, it feels wasteful to have individual instances of BKDConfig records as most of the time those instances correspond to standard lucene fields. Therefore lets deduplcate them at creation time and save some heap.


original-brownbear (approved):
LGTM :)

lucene/core/src/java/org/apache/lucene/util/bkd/BKDConfig.java
```diff
@@ -38,6 +39,19 @@ public record BKDConfig(int numDims, int numIndexDims, int bytesPerDim, int maxP
...
+          // config for int / float
+          new BKDConfig(1, 1, 4, DEFAULT_MAX_POINTS_IN_LEAF_NODE),
+          // config for long / double
+          new BKDConfig(1, 1, 8, DEFAULT_MAX_POINTS_IN_LEAF_NODE),
+          // config for inetAddress
+          new BKDConfig(1, 1, 16, DEFAULT_MAX_POINTS_IN_LEAF_NODE),
+          // config for 2D points
+          new BKDConfig(2, 2, 4, DEFAULT_MAX_POINTS_IN_LEAF_NODE),
+          // config for shapes
+          new BKDConfig(7, 4, 4, DEFAULT_MAX_POINTS_IN_LEAF_NODE));
```
jpountz, line 53: It feels a bit arbitrary to not include HaltFloatPoint and LongRange/DoubleRange. I wonder if we should do it in a way that it a bit more agnostic to what specific implementations exist and add an entry for numbers of bytes per value 1, 2, 4, 8 and for dimensions 1 and 2 (plus the config for shapes) instead.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDConfig.java
```diff
@@ -38,6 +39,19 @@ public record BKDConfig(int numDims, int numIndexDims, int bytesPerDim, int maxP
...
+
   public BKDConfig {
```
jpountz, line 55: should we aim at making this private?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDConfig.java
```diff
@@ -68,6 +82,16 @@ public record BKDConfig(int numDims, int numIndexDims, int bytesPerDim, int maxP
     }
   }
 
+  static BKDConfig create(int numDims, int numIndexDims, int bytesPerDim, int maxPointsInLeafNode) {
```
jpountz, line 85: nit: `create` suggests that this returns a new instance, what about `of(...)` instead?

jpountz (approved):
I'm not too comfortable with the heap saving argument, as these objects only consume a couple tens of bytes, and it should be ok for codecs to allocate a couple tens of bytes per field. That said, the change looks simple enough.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDConfig.java
iverase: feels better as you suggested.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDConfig.java
iverase: Changed the test so it can be private now.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDConfig.java
```diff
@@ -38,6 +39,18 @@ public record BKDConfig(int numDims, int numIndexDims, int bytesPerDim, int maxP
...
+      List.of(
+          new BKDConfig(1, 1, 2, DEFAULT_MAX_POINTS_IN_LEAF_NODE),
+          new BKDConfig(1, 1, 4, DEFAULT_MAX_POINTS_IN_LEAF_NODE),
+          new BKDConfig(1, 1, 8, DEFAULT_MAX_POINTS_IN_LEAF_NODE),
+          new BKDConfig(1, 1, 16, DEFAULT_MAX_POINTS_IN_LEAF_NODE),
+          new BKDConfig(2, 2, 2, DEFAULT_MAX_POINTS_IN_LEAF_NODE),
+          new BKDConfig(2, 2, 4, DEFAULT_MAX_POINTS_IN_LEAF_NODE),
+          new BKDConfig(2, 2, 8, DEFAULT_MAX_POINTS_IN_LEAF_NODE),
+          new BKDConfig(2, 2, 16, DEFAULT_MAX_POINTS_IN_LEAF_NODE),
+          new BKDConfig(7, 4, 4, DEFAULT_MAX_POINTS_IN_LEAF_NODE));
```
jpountz, line 52: nit: add a comment that says that this is the config of shapes?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDConfig.java
```diff
@@ -38,6 +39,18 @@ public record BKDConfig(int numDims, int numIndexDims, int bytesPerDim, int maxP
   /** Maximum number of index dimensions */
   public static final int MAX_INDEX_DIMS = 8;
 
+  private static final List<BKDConfig> DEFAULT_CONFIGS =
+      List.of(
+          new BKDConfig(1, 1, 2, DEFAULT_MAX_POINTS_IN_LEAF_NODE),
```
jpountz, line 44: nit: add a comment about the reasoning wrt how this list has been compiled?

jpountz (approved):

jpountz (approved):

PR #14343, opened 2025-03-11
Speedup slice calculation in IndexSearcher

original-brownbear:
It's in the title, some obvious speedups. This is fairly expensive logic for Elasticsearch when run over a larger number of shards. No need for streams, creating comparator instances and so much copying here.


msfroh (approved):
Nice cleanup! Cutting down on the number of unnecessary allocations is nice.

I wonder if it makes sense to require that the input to the `LeafSlice` constructor is already sorted? In general, the way that they get created (even with segment partitions) could be sorted externally. That said, for the common existing case where each slice has one partition, the sort will fast-return anyway.

original-brownbear:
Thanks Michael!

PR #14349, opened 2025-03-12
Case-insensitive TermInSetQuery Implementation (Proof of Concept)

willdickerson:
## Overview
  This PR introduces a proof of concept for a case-insensitive variant of TermInSetQuery. The implementation provides an efficient way to search for terms regardless of case without needing to generate all case variations.

  ## Implementation Details
  - Extends `MultiTermQuery` to leverage existing infrastructure
  - Uses regular expressions with case-insensitive flag for pattern matching
  - Implements filtering through a `ByteRunAutomaton` for efficient matching
  - Ensures proper handling of deterministic automata
  - Follows Lucene's visitor pattern for queries

  ## Limitations
  - Uses Java's standard case folding which may not handle all Unicode special cases
  - For full locale-aware case folding, an analyzer should be used during indexing

  ## Testing
  The implementation includes comprehensive test cases covering:
  - Basic case-insensitive matching
  - Multiple term behavior
  - Unicode character testing with notes on special cases
  - Comparison with standard TermInSetQuery
  - Visitor pattern
  - Contract tests (equals/hashCode)
  - Randomized testing with Unicode characters

  This POC is intended to gather feedback from the community

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/CaseInsensitiveTermInSetQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/CaseInsensitiveTermInSetQuery.java
@@ -19,12 +19,9 @@ package org.apache.lucene.search;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.Objects;
 import java.util.Set;
-import org.apache.lucene.index.FilteredTermsEnum;
 import org.apache.lucene.index.Term;
-import org.apache.lucene.index.Terms;
-import org.apache.lucene.index.TermsEnum;
-import org.apache.lucene.util.AttributeSource;
 import org.apache.lucene.util.BytesRef;
 import org.apache.lucene.util.automaton.Automaton;
 import org.apache.lucene.util.automaton.ByteRunAutomaton;
@@ -32,13 +29,12 @@ import org.apache.lucene.util.automaton.Operations;
 import org.apache.lucene.util.automaton.RegExp;
 
 /**
- * A {@link TermInSetQuery} that matches terms in a case-insensitive manner.
+ * A query that matches terms in a case-insensitive manner using regular expressions.
  *
- * <p>This query performs case-insensitive term matching by using a regexp-based approach to match
- * all case variations of the given terms. For multi-term queries with many terms, this can be more
- * efficient than creating a large boolean query with many disjunctions.
+ * <p>This query creates a {@link RegexpQuery} for each term with case-insensitive matching enabled.
+ * For multiple terms, it builds a boolean query combining the individual regex queries.
  */
-public class CaseInsensitiveTermInSetQuery extends MultiTermQuery {
+public class CaseInsensitiveTermInSetQuery extends Query {
 
   private final String field;
   private final Set<BytesRef> terms;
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/search/CaseInsensitiveTermInSetQuery.java
```diff
@@ -0,0 +1,185 @@
...
+      sb.append(c);
+    }
+    return sb.toString();
+  }
+
+  /**
+   * A specialized TermsEnum implementation that performs case-insensitive matching against the
+   * terms dictionary using a ByteRunAutomaton for proper Unicode case handling.
+   */
+  private class CaseInsensitiveSetEnum extends FilteredTermsEnum {
```
rmuir, line 153: Don't do this, you lose all skipping and it will evaluate every single term in the terms dictionary. extremely slow. just extend Query and rewrite() to RegexpQuery since that's what you are doing.

lucene/core/src/java/org/apache/lucene/search/CaseInsensitiveTermInSetQuery.java
```diff
@@ -0,0 +1,185 @@
...
+    
+    // Create a regular expression with case-insensitive flag
+    RegExp regexp = new RegExp(regexBuilder.toString(), RegExp.ALL, 
+                              RegExp.CASE_INSENSITIVE);
+    Automaton automaton = regexp.toAutomaton();
+    
+    // Ensure the automaton is deterministic
+    if (!automaton.isDeterministic()) {
+      automaton = Operations.determinize(automaton, Operations.DEFAULT_DETERMINIZE_WORK_LIMIT);
+    }
```
rmuir, line 131: The automaton here is probably on the order of excessive if you have many terms. try it with a few thousand

lucene/core/src/test/org/apache/lucene/search/TestCaseInsensitiveTermInSetQuery.java
```diff
@@ -0,0 +1,377 @@
...
+    }
+    
+    IndexReader reader = iw.getReader();
+    IndexSearcher searcher = newSearcher(reader);
+    iw.close();
+    
+    // Test with a random subset of terms
+    List<String> queryBaseTerms = new ArrayList<>(baseTerms);
+    queryBaseTerms = queryBaseTerms.subList(0, Math.min(5, queryBaseTerms.size()));
+
```
msfroh, line 349: I think you'd need to test this with thousands of terms -- 5 isn't going to be enough to detect issues. (It's small enough that you could just do a disjunction of case-insensitive term queries.)

I like @rmuir's suggestion on the mailing list that a case-insensitive version of `Automata.makeStringUnion` *might* be reasonable.

willdickerson: I've added a test case with 5,000 terms in `testLargeNumberOfTerms()`. Instead of using `Automata.makeStringUnion`, I went with a HashSet-based approach that avoids the automaton determinization costs. I can explore the `Automata.makeStringUnion` approach if you prefer that direction.

lucene/core/src/java/org/apache/lucene/search/CaseInsensitiveTermInSetQuery.java
```diff
@@ -81,58 +89,95 @@ public void visit(QueryVisitor visitor) {
...
+
+    // Override the run method with our HashSet-based implementation
+    return new ByteRunAutomaton(emptyAutomaton) {
+      @Override
+      public boolean run(byte[] s, int offset, int length) {
+        // Convert input bytes to string and lowercase for comparison
+        BytesRef slice = new BytesRef(s, offset, length);
+        String termString = slice.utf8ToString().toLowerCase();
+        return lowerCaseTerms.contains(termString);
+      }
```
msfroh: The `ByteRunAutomaton` is only used for highlighting (see `MultiTermHighlighting`). The actual query will use the `Automaton` passed to `AutomatonQuery`.

lucene/core/src/java/org/apache/lucene/search/CaseInsensitiveTermInSetQuery.java
```diff
@@ -81,58 +89,95 @@ public void visit(QueryVisitor visitor) {
...
+   * complex automata with many terms.
+   */
+  private Automaton createAutomaton() {
+    // Create a minimal automaton that accepts any string
+    // The actual matching is done via the ByteRunAutomaton override
+    Automaton automaton = new Automaton();
+    int state = automaton.createState();
+    automaton.setAccept(state, true);
+    return automaton;
   }
```
msfroh: This `Automaton` will only match the empty string. (It's equivalent to `Automata.makeEmptyString()`.)

I believe the query will be very fast, but won't match any terms (except the empty string). In the end, I believe we'd need an `Automaton` that matches all the strings in a case-insensitive fashion.

I think it would be possible to do that with some changes to the `StringsToAutomaton` class to support case-insensitive transitions. Essentially, for each codepoint, we would need to add a transition for the lower- and upper-case versions (if applicable) in the `convert` method.

msfroh:
@willdickerson -- I took a stab at modifying `StringsToAutomaton`, to support case-insensitive matching: https://github.com/apache/lucene/pull/14350

rmuir:
In my head, that's what we need. There is a crazy difference in construction and execution time between a "native" union and using the efficient linear-time algorithm, as opposed to going worst-case exponential. With simple case-folding it may still be possible to do this though. Try using the new test functions available such as `AutomatonTestUtil.assertMinimal` and company to assist with testing. I will dig into the linked PR!

jpountz:
Implementing it as a query that rewrites to the proper query based on the list of terms makes sense to me. You may want to move this query to the lucene-queries module since it doesn't really qualify as a "core" query.

PR #14350, opened 2025-03-13
[DRAFT] Case-insensitive matching over union of strings

msfroh:
### Description

This is a rough attempt to make `StringsToAutomaton` support case-insensitive strings.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
@@ -608,7 +608,7 @@ public final class Automata {
     if (utf8Strings.iterator().hasNext() == false) {
       return makeEmpty();
     } else {
-      return StringsToAutomaton.build(utf8Strings, false, false);
+      return StringsToAutomaton.build(utf8Strings, false, false, false);
     }
   }
 
@@ -620,11 +620,12 @@ public final class Automata {
    * @return An {@link Automaton} accepting all input strings. The resulting automaton is codepoint
    *     based (full unicode codepoints on transitions).
    */
-  public static Automaton makeCaseInsensitiveStringUnion(Iterable<BytesRef> utf8Strings) {
+  public static Automaton makeCaseInsensitiveStringUnion(
+      Iterable<BytesRef> utf8Strings, boolean turkic) {
     if (utf8Strings.iterator().hasNext() == false) {
       return makeEmpty();
     } else {
-      return StringsToAutomaton.build(utf8Strings, false, true);
+      return StringsToAutomaton.build(utf8Strings, false, true, turkic);
     }
   }
 
@@ -641,7 +642,7 @@ public final class Automata {
     if (utf8Strings.iterator().hasNext() == false) {
       return makeEmpty();
     } else {
-      return StringsToAutomaton.build(utf8Strings, true, false);
+      return StringsToAutomaton.build(utf8Strings, true, false, false);
     }
   }
 
@@ -654,7 +655,7 @@ public final class Automata {
    *     based (full unicode codepoints on transitions).
    */
   public static Automaton makeStringUnion(BytesRefIterator utf8Strings) throws IOException {
-    return StringsToAutomaton.build(utf8Strings, false, false);
+    return StringsToAutomaton.build(utf8Strings, false, false, false);
   }
 
   /**
@@ -667,6 +668,6 @@ public final class Automata {
    *     based (UTF-8 encoded byte transition labels).
    */
   public static Automaton makeBinaryStringUnion(BytesRefIterator utf8Strings) throws IOException {
-    return StringsToAutomaton.build(utf8Strings, true, false);
+    return StringsToAutomaton.build(utf8Strings, true, false, false);
   }
 }
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
@@ -608,7 +608,23 @@ public final class Automata {
     if (utf8Strings.iterator().hasNext() == false) {
       return makeEmpty();
     } else {
-      return StringsToAutomaton.build(utf8Strings, false);
+      return StringsToAutomaton.build(utf8Strings, false, false);
+    }
+  }
+
+  /**
+   * Returns a new (deterministic and minimal) automaton that accepts the union of the given
+   * collection of {@link BytesRef}s representing UTF-8 encoded strings.
+   *
+   * @param utf8Strings The input strings, UTF-8 encoded. The collection must be in sorted order.
+   * @return An {@link Automaton} accepting all input strings. The resulting automaton is codepoint
+   *     based (full unicode codepoints on transitions).
+   */
+  public static Automaton makeCaseInsensitiveStringUnion(Iterable<BytesRef> utf8Strings) {
+    if (utf8Strings.iterator().hasNext() == false) {
+      return makeEmpty();
+    } else {
+      return StringsToAutomaton.build(utf8Strings, false, true);
     }
   }
 
@@ -625,7 +641,7 @@ public final class Automata {
     if (utf8Strings.iterator().hasNext() == false) {
       return makeEmpty();
     } else {
-      return StringsToAutomaton.build(utf8Strings, true);
+      return StringsToAutomaton.build(utf8Strings, true, false);
     }
   }
 
@@ -638,7 +654,7 @@ public final class Automata {
    *     based (full unicode codepoints on transitions).
    */
   public static Automaton makeStringUnion(BytesRefIterator utf8Strings) throws IOException {
-    return StringsToAutomaton.build(utf8Strings, false);
+    return StringsToAutomaton.build(utf8Strings, false, false);
   }
 
   /**
@@ -651,6 +667,6 @@ public final class Automata {
    *     based (UTF-8 encoded byte transition labels).
    */
   public static Automaton makeBinaryStringUnion(BytesRefIterator utf8Strings) throws IOException {
-    return StringsToAutomaton.build(utf8Strings, true);
+    return StringsToAutomaton.build(utf8Strings, true, false);
   }
 }
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/CaseFolding.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/CaseFolding.java
@@ -743,4 +743,42 @@ class CaseFolding {
 
     return alts;
   }
+
+  /**
+   * Folds the case of the given character according to {@link Character#toLowerCase(int)}, but with
+   * exceptions if the turkic flag is set.
+   *
+   * @param codepoint to code point for the character to fold
+   * @param turkic if true, then apply tr/az folding rules
+   * @return the folded character
+   */
+  static int foldCase(int codepoint, boolean turkic) {
+    if (turkic) {
+      if (codepoint == 0x00130) { // İ [LATIN CAPITAL LETTER I WITH DOT ABOVE]
+        return 0x00069; // i [LATIN SMALL LETTER I]
+      } else if (codepoint == 0x000049) { //  I [LATIN CAPITAL LETTER I]
+        return 0x00131; // ı [LATIN SMALL LETTER DOTLESS I]
+      }
+    }
+    return Character.toLowerCase(codepoint);
+  }
+
+  /**
+   * Attempts to convert the given character to upper case, acccording to {@link
+   * Character#toUpperCase(int)}, but with exceptions if the turkic flag is set.
+   *
+   * @param codepoint to code point for the character to convert to upper case
+   * @param turkic if true, then apply tr/az folding rules
+   * @return the upper case character
+   */
+  static int upperCase(int codepoint, boolean turkic) {
+    if (turkic) {
+      if (codepoint == 0x00069) { // i [LATIN SMALL LETTER I]
+        return 0x00130; // İ [LATIN CAPITAL LETTER I WITH DOT ABOVE]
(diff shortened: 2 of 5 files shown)


rmuir:
@dweiss understands this one the best, he implemented it.

lucene/core/src/java/org/apache/lucene/util/automaton/StringsToAutomaton.java
```diff
@@ -209,7 +209,25 @@ private static int convert(
...
+          for (int alt : alternatives) {
+            a.addTransition(converted, dest, alt);
+          }
+        } else {
+          int altCase =
+                  Character.isLowerCase(label)
+                          ? Character.toUpperCase(label)
+                          : Character.toLowerCase(label);
+          if (altCase != label) {
+            a.addTransition(converted, dest, altCase);
```
rmuir, line 227: maybe this works because the finishState() will sort the transitions always, and still give us the same efficient build?

rmuir:
To me it seems potentially safe and practical addition. The idea would be that, we can add transition "alternatives" (e.g. `A` vs `a`) and it doesn't break the high-level algorithm, due to internal representation and how we sort transitions when doing e.g. `finishState()`?

I don't really have confidence either way, but it is a good one to explore. Maybe with testing we can be confident it still does the right thing?

rmuir:
OG paper: https://aclanthology.org/J00-1002.pdf

lucene/core/src/java/org/apache/lucene/util/automaton/StringsToAutomaton.java
msfroh: Essentially, I tried to copy what you did for the case-insensitive regex matching to add extra transition arcs for the other letter-cases.

I think the `finish` call is handled at the end.

Note that this implementation will be way more efficient if all of the input strings are the same case. Otherwise, it might miss common (case-insensitive) prefixes. I'm imagining that a query would lowercase everything first.

dweiss:
> @dweiss understands this one the best, he implemented it.

... 15 years ago in LUCENE-3832. Thanks for putting so much trust in my memory. I'll take a look.

dweiss:
I think this will work just fine in most cases and is a rather inexpensive way to implement this case-insensitive matching, but this comes at the cost of the output automaton that may not be minimal. Consider this example:
```
    List<BytesRef> terms = new ArrayList<>(List.of(
            newBytesRef("abc"),
            newBytesRef("aBC")));
    Collections.sort(terms);
    Automaton a = build(terms, false, false);
```
which produces:
image

However, when you naively expand just the transitions for each letter variant, you get this:
image
which clearly isn't minimal (and doesn't pass checkMinimized).

I think the absolutely worst case is for the automaton to double the number of transitions - the number of states remains the same. So it's not like it's going to expand uncontrollably... But it's no longer minimal. Perhaps this is acceptable, given the constrained worst case?

rmuir:
Bigger downside: that example isn't deterministic either.

dweiss:
Crap, you're right. Didn't think of it.

dweiss:
I also don't think you can make it deterministic in any trivial way.

msfroh:
My thinking is that a query that uses this should lowercase, dedupe, and sort the input before feeding it into `StringsToAutomaton`. That would handle @dweiss's example (i.e. that input is "invalid", or at least exact duplicates wouldn't add any new states, I think, since the full string exists as a prior prefix). 

Would a check with `Character.isLowerCase()` on each input codepoint for the case-insensitive case be sufficient to reject that kind of input across all valid Unicode strings?

rmuir:
> Would a check with `Character.isLowerCase()` on each input codepoint for the case-insensitive case be sufficient to reject that kind of input across all valid Unicode strings?

I dont think so for greek. I would step back from that and try to get matching working with simple Character.toLowerCase/toUpperCase first? If the user provides data with a certain order/casing as you suggest, will they always get a DFA?

I'm less concerned about it being minimal, let's start with deterministic. I don't think we should do this if it will explode (e.g. NFA). And union of strings really wants to do that, if handled the naive way, that's why there is a special algorithm for it.

msfroh:
To the best of my understanding from reading the through the code while sketching this PR, I believe it would produce a minimal DFA if every character in a set of alternatives in the input strings have the same canonical representation. (The existing implementation already throws if input is not sorted BytesRefs.)

That is, if you input `cap, cat, cats, cob`, it will generate the minimal DFA. If you input `CAP, CAT, CATS, COB`, you'll end up with the same minimal DFA (albeit with the transitions added in the opposite order, which I think is fine). But if you input `CAP, CATS, cat, cob`, you'll end up with a NFA.

dweiss:
I don't know Unicode as well as Rob so I can't say what these alternate case folding
equivalence classes are... but they definitely don't have a "canonical" representation
with regard to Character.toLowercase. Consider the killer Turkish dotless i, for example:

```
    public void testCornerCase() throws Exception {
        List<BytesRef> terms = Stream.of(
                        "aIb", "aıc")
                .map(s -> {
                    int[] lowercased = s.codePoints().map(Character::toLowerCase).toArray();
                    return new String(lowercased, 0, lowercased.length);
                })
                .map(LuceneTestCase::newBytesRef)
                .sorted()
                .collect(Collectors.toCollection(ArrayList::new));
        Automaton a = build(terms, false, true);
        System.out.println(a.toDot());
        assertTrue(a.isDeterministic());
    }
```
which yields:
image

It would take some kind of character normalization filter on both the index and automaton building/expansion side for this to work.

dweiss:
Or we can just embrace the fact that it can be a non-minimal NFA and justlet it run like that (with NFARunAutomaton).

dweiss:
Ok, fair enough.

msfroh:
This is kind of what I had in mind:

```java
  private static int canonicalize(int codePoint) {
    int[] alternatives = CaseFolding.lookupAlternates(codePoint);
    if (alternatives != null) {
      for (int cp : alternatives) {
        codePoint = Math.min(codePoint, cp);
      }
    } else {
      int altCase = Character.isLowerCase(codePoint) ? Character.toUpperCase(codePoint) : Character.toLowerCase(codePoint);
      codePoint = Math.min(codePoint, altCase);
    }
    return codePoint;
  }

  public void testCornerCase() throws Exception {
    List<BytesRef> terms = Stream.of(
                    "aIb", "aıc")
            .map(s -> {
              int[] lowercased = s.codePoints().map(TestStringsToAutomaton::canonicalize).toArray();
              return new String(lowercased, 0, lowercased.length);
            })
            .map(LuceneTestCase::newBytesRef)
            .sorted()
            .collect(Collectors.toCollection(ArrayList::new));
    Automaton a = build(terms, false, true);
    System.out.println(a.toDot());
    assertTrue(a.isDeterministic());
  }
```

That produces this automaton, which is minimal and deterministic:
automaton

I don't know if that `canonicalize` method is a good idea, though.

rmuir:
It isn't a good idea. If the user wants to "erase case differences" then they should apply `foldcase(ch)`. That's what case-folding means. That CaseFolding class does everything, except, that. Again its why i recommend not messing with it for now and starting simpler.

msfroh:
Hmm... I'm thinking of just requiring that input is lowercase (per `Character.lowerCase(c)`), then check for collisions on uppercase versions when adding transitions, and throw an exception (since it won't be a DFA).

Unfortunately, that would mess with Turkish, if someone tries searching for sınıf (class) and sinirli (nervous). Without locale info, we'd get two transitions from s to I.

rmuir:
+1 to start simple with Character.toLowerCase, thats the best you can get in java.

The problem is java not having a Character.foldCase. A proper function would look like ICU's `UCharacter.foldCase(int, boolean)`: https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/lang/UCharacter.html#foldCase-int-boolean-

The regexp folding code doesn't handle turkish correctly either. dotless and dotted I are DIFFERENT, but it mixes all these characters up and conflates them. So I'd like for us not to perpetuate this further, somehow creating "nonstandard case folding" disagrees with the unicode standard.

rmuir:
Separately, it would be nice to add boolean flag (for turkish/azeri) to that CaseFolding class, and fix it to do the right thing, so it doesn't match unrelated characters in turkish. ultimately if we want to add a function to that class to "fold" (e.g. for purposes like here) it should expose that boolean and match unicode casefolding algorithm.

rmuir:
I think my ask is misunderstood, it is just to follow the Unicode standard. There are two mappings for simple case folding:
* Default
* Alternate (Turkish/azeri)

rmuir:
If you want to do fancy romanian accent removal, use an analyzer and normalize your data. That's what a search engine is all about. 

But if we want to provide some limited runtime expansion (which I'm still very unsure about), let's just stick with the standards and not invent something else. No need for interfaces, abstractions or lambdas. `boolean` is the correct solution.

msfroh:
Okay, got it! That's the piece that I was misunderstanding. I didn't realize that Turkish/Azeri is the **only** other valid folding. I kept thinking of it as just an example where the naïve folding wouldn't work.

I'll go ahead and update this PR with that in mind.

rmuir:
it is confusing. because unicode case folding algorithm is supposed to work for everyone. But here's the problem:

for most of the world:
* lowercase i has a dot, uppercase I has no dot.

for turkish/azeri world:
* lowercase i corresponds to İ with a dot and lowercase ı without dot corresponds to I

that's why they need their own separate mappings.

lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java
```diff
@@ -608,7 +608,24 @@ public static Automaton makeStringUnion(Iterable<BytesRef> utf8Strings) {
...
+
+  /**
+   * Returns a new (deterministic and minimal) automaton that accepts the union of the given
+   * collection of {@link BytesRef}s representing UTF-8 encoded strings.
+   *
+   * @param utf8Strings The input strings, UTF-8 encoded. The collection must be in sorted order.
+   * @return An {@link Automaton} accepting all input strings. The resulting automaton is codepoint
+   *     based (full unicode codepoints on transitions).
+   */
+  public static Automaton makeCaseInsensitiveStringUnion(
```
dweiss, line 623: turkic parameter is missing from the javadoc.

dweiss (commented):
Those turkic flags... they will proliferate throughout the factory methods. 

I don't know. Seems like we're trying to compensate for something that could indeed be done during indexing (analyzer pipeline).

lucene/core/src/java/org/apache/lucene/util/automaton/CaseFolding.java
```diff
@@ -743,4 +743,42 @@ static int[] lookupAlternates(int codepoint) {
...
+   */
+  static int foldCase(int codepoint, boolean turkic) {
+    if (turkic) {
+      if (codepoint == 0x00130) { // İ [LATIN CAPITAL LETTER I WITH DOT ABOVE]
+        return 0x00069; // i [LATIN SMALL LETTER I]
+      } else if (codepoint == 0x000049) { //  I [LATIN CAPITAL LETTER I]
+        return 0x00131; // ı [LATIN SMALL LETTER DOTLESS I]
+      }
+    }
+    return Character.toLowerCase(codepoint);
```
rmuir, line 763: For real case folding we have to do more than this. it is a simple 1-1 mapping but e.g. `Σ`, `σ`, and `ς`, will all fold to σ. Whereas toLowerCase(ς) = ς. Because it is already in lower-case, just in final-form. This is just an example. To see more, compare your function against ICU UCharacter.foldCase(int, bool) across all of unicode.

msfroh: Got it. 

Checking https://www.unicode.org/Public/16.0.0/ucd/CaseFolding.txt, indeed I can see those entries:

```
03A3; C; 03C3; # GREEK CAPITAL LETTER SIGMA
...
03C2; C; 03C3; # GREEK SMALL LETTER FINAL SIGMA
```

Ideally, I'd love to just use those folding rules.

I could get them from `UCharacter.foldCase(int, bool)`, but that involves pulling in icu4j as a dependency, which is an extra 12MB jar.

Would it be worthwhile to write a generator that pulls https://www.unicode.org/Public/16.0.0/ucd/CaseFolding.txt (updated to whatever the current Unicode spec is) and generates a `foldCase` method that's functionally equivalent to `UCharacter.foldcase(int, bool)`?

rmuir: Maybe, depending what we are going to do with it? if done correctly we could replace `LowerCaseFilter`, `GreekLowerCaseFilter`, etc in analysis chain. Of course "correctly" there is a difficult bar, as it would impact 100% of users in a very visible way and could easily bottleneck indexing / waste resources if not done correctly. For example large-arrays-of-objects or even primitives is a big no here. See https://www.strchr.com/multi-stage_tables and look at what JDK and ICU do already.

But for the purpose of this PR, we may want to start simpler (this is the same approach I mentioned on regex caseless PR). We should avoid huge arrays and large data files in lucene-core, just for adding more inefficient user regular expressions that isn't really related to searching. On the other hand, if we are going to get serious benefit everywhere (e.g. improve all analyzers), then maybe the tradeoff makes sense.

And I don't understand why we'd parse text files versus just write any generator itself to use ICU... especially since we already use such an approach in the build already: https://github.com/apache/lucene/blob/main/gradle/generation/icu/GenerateUnicodeProps.groovy

Still I wouldn't immediately jump to generation as a start, it is a lot of work, and we should iterate. First i'd compare `Character.toLowerCase(Character.toUpperCase(x))` to `UCharacter.foldCase(int, false)` to see what the delta really needs to be as far as data. I'd expect this to be very small. You can start prototyping with that instead of investing a ton of up-front time.

rmuir:
Maybe this one helps the issue: https://github.com/apache/lucene/pull/14389

PR #14352, opened 2025-03-13
PointInSetQuery use reverse collection to improve performance

hanbj:
### Description

Performance issues with terms number field encountered in production environments: high query time and very high CPU usage.

Through analysis and localization, it was found that the main time consumption is in the build_scorer stage. The main task of build_scorer is to collect document IDs and fill bitset in memory. When there are a large number of document IDs, it takes up a lot of CPU.

After reading the code and optimizing it, the performance of terms number field is improved by 5-10 times in low cardinality case.

1. values.getDocCount() == reader.maxDoc() : ensures that each document contains this field.
2. values.getDocCount() == values.size() : ensures that each document has only one value for this field.
3. cost() > reader.maxDoc() / 2 : ensure that more than half of the documents matched by all the points to be queried.

Because when each document has only one value for this field, the document id in the bkd tree will not be duplicated. There will be no collection of duplicate document IDs. Therefore, first determine the number of documents that match all the points to be queried. If it exceeds half, use reverse collection to reduce the number of collected document ids.

Another new thought is that I believe SinglePointVisitor is unnecessary, as processing each point separately will traverse the bkd tree multiple times. MergePointVisitor will only traverse the bkd tree once. Therefore, in most cases, SinglePointVisitor will result in slower performance. As long as MergePointVisitor supports both single and multiple dimensions, reverse collection optimization can be used.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/PointInSetQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/PointInSetQuery.java
@@ -34,10 +34,13 @@ import org.apache.lucene.index.PrefixCodedTerms.TermIterator;
 import org.apache.lucene.util.Accountable;
 import org.apache.lucene.util.ArrayUtil;
 import org.apache.lucene.util.ArrayUtil.ByteArrayComparator;
+import org.apache.lucene.util.BitSetIterator;
 import org.apache.lucene.util.BytesRef;
 import org.apache.lucene.util.BytesRefBuilder;
 import org.apache.lucene.util.BytesRefIterator;
 import org.apache.lucene.util.DocIdSetBuilder;
+import org.apache.lucene.util.FixedBitSet;
+import org.apache.lucene.util.IntsRef;
 import org.apache.lucene.util.RamUsageEstimator;
 
 /**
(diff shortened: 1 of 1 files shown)


msfroh:
> Another new thought is that I believe SinglePointVisitor is unnecessary, as processing each point separately will traverse the bkd tree multiple times. MergePointVisitor will only traverse the bkd tree once. Therefore, in most cases, SinglePointVisitor will result in slower performance. As long as MergePointVisitor supports both single and multiple dimensions, reverse collection optimization can be used.

Right now, `MergePointVisitor` only supports 1-dimension, since it relies on the fact that it can traverse the tree from left to right at the same time as it's traversing the points from left to right (like a merge sort).

I think it might be possible to do an N-dimensional approach like a quick sort, though.

Suppose you have a list of N-dimensional points. At the root of the tree, (I think) you always have two children. If you take the upper bound of the right child / lower bound of the left child as your pivot, you can do an O(n) reordering of the list into points that potentially match on the left and right. (If there's a common value shared between right and left, you would have to push those values both ways.) As you go down the tree, you're focused on a sublist of the original list and can reorder that sublist based on the next level's pivot.

The values shared between left and right might mess things up, though, since you can't safely reorder them on the left without potentially messing up the right. Hmm... might still be doable, though.

hanbj:
Thank you for providing ideas. In scenarios with multiple dimensions, the internal nodes in the bkd tree can only be sorted according to a certain dimension. Different internal nodes may have different sorting dimensions, which is indeed difficult to implement.

PR #14357, opened 2025-03-16
BooleanScorer doesn't optimize for TwoPhaseIterator

dsmiley:
Showing a performance problem here in BooleanScorer (used for disjunctions -- "OR").   BS will score all its clauses indepenently, overlapping the same documents, some of which might be expensive with a TwoPhaseIterator.  Wow.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/TestBooleanScorer.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestBooleanScorer.java
@@ -18,6 +18,8 @@ package org.apache.lucene.search;
 
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.IntStream;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field;
 import org.apache.lucene.document.Field.Store;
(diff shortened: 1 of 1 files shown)


jpountz:
+1 let's use `DisjunctionSumScorerwhich` (which already supports two-phase iteration) when one of the clauses exposes a non-null two-phase iterator?

dsmiley:
Thanks for your confirmation of the problem.  The collect-per-clause is surprising to me; like what would benefit from that algorithm?  Wouldn't that _only_ be in fact _needed_ if scores are _needed_?  Since they need to be aggregated across the clauses on a matching document.  If so, this isn't really about TPIs, it's about the need for scores.

jpountz:
BS2 uses a heap to merge multiple `DocIdSetIterator`s. Unfortunately, reordering this heap on every call to `nextDoc()` or `advance(int)` is not completely free and BS1's approach of loading all clauses into a shared bit set and then iterating the bits of this bit set proves to be faster.

Presumably, when two-phase iterators are involved, the cost of reording the heap wouldn't be a big deal compared to the cost of calling `TwoPhaseIterator#matches()`.

dsmiley:
If one or more DISI has a high cost (irrespective of TPIs), thus matching many docs, I could see avoiding BS1 as well.

An aside, if we are going to refer to these as BS1 vs BS2, they should have names more clearly reflecting this.

jpountz:
> If one or more DISI has a high cost (irrespective of TPIs), thus matching many docs, I could see avoiding BS1 as well.

I imagine that your idea is that if most of the cost comes from a single DISI then the heap doesn't need to be reorderer every time so the heap reordering overhead may be less than the bitset overhead. `BooleanScorer` has an optimization for a similar situation where if a single clause matches on a window of 4,096 docs then it will skip the bit set as an intermediate representation of matches and feed this clause directly into the collector.

> An aside, if we are going to refer to these as BS1 vs BS2, they should have names more clearly reflecting this.

Agreed to stop using `BS1`/`BS2`. It's how Lucene's disjunction scorers were historically referred to and these names came naturally after reading your description message that abbreviated `BooleanScorer` to `BS`, but it doesn't properly reflect how Lucene scores disjunctions nowadays.

jpountz:
In case you missed it, `BooleanScorer` had optimizations recently that make it hard to beat by `DisjunctionScorer` when clauses are `PostingsEnum`s:
 - `DocIdSetIterator#intoBitSet` helps impls optimize loading matches into bit sets, in the case of `PostingsEnum` through (auto) vectorization (annotation [`HR`](https://benchmarks.mikemccandless.com/CountOrHighHigh.html))
 - Dense blocks of postings are stored as bit sets, further speeding up OR-ing postings into a bit set ([annotations `HW` and `HX`](https://benchmarks.mikemccandless.com/CountOrHighHigh.html))

dsmiley:
I could imagine improving BooleanScorer so that the TPI clauses are separated and converted to a filter around the collector to try to match docs *not* collected (i.e. test for docs inbetween those collected).  All this logic could be separated to another class such that BooleanScorer is nearly the same or even identical if the caller were to compose it.  DisjunctionDISIApproximation would be instrumental.

dsmiley:
An aside:  `org.apache.lucene.search.DisjunctionScorer.TwoPhase#matches` looks kind of sad, in that each matches() call is going to build a priority queue of "unverified matches" (DisiWrapper holding TwoPhaseIterator).  It seems strange to populate one on visiting each doc instead of maintaining a fixed pre-sorted array of them, since we know which clauses have TPIs.  The DisiWrapper could have a TPI index (by match cost) into an array of TwoPhaseIterators.  The selected unverifiedMatches per matches() call might be noted via a bitmask/bitset that is cheap to set & clear & iterate set bits.  Or could just use an array of DisiWrapper that is cleared & filled.  No matchCost comparisons & heap manipulation.
Not sure if I'm over-optimizing here.  The use-case bringing me here is only one TPI, and it's approximation is all docs.

jpountz:
The current approach is probably not the fastest indeed. We should add a task to nightly benchmarks if we want to optimize this. Something like a disjunction of phrase queries (possibly both for top-n collection and counting)?

dsmiley:
BTW I don't have plans to explore this further.  Anyone should feel free to take over.  Or abandon if nobody cares -- I admit it's very unusual to even have a top level disjunction, let alone one with an expensive clause, as implied by the TPI.  I found this while writing a simple test, so simple that it wasn't real-world representative.  The real-world representative has the disjunction deeper than the top level, and that which doesn't show this performance problem as the bulk scorer is then not used.

PR #14364, opened 2025-03-17
Completion FSTs to be loaded off-heap at all times

javanna:
All the existing completion postings format load their FSTs on-heap. It is possible to customize that behaviour by mainintaing a custom postings format that override the fst load mode.

TestSuggestField attempted to test all modes, but in practice, it can only test the default load mode because the read path relies on the default constructor called via SPI, that does not allow providing the fst load mode.

This commit switches loading of FST to off-heap, and removes the fst load mode argument from all completion postings format classes, effectively making the loading always off-heap.

See related discussion where this change originated at #14275 .

### DIFF
--- a/lucene/suggest/src/java/org/apache/lucene/search/suggest/document/CompletionPostingsFormat.java
+++ b/lucene/suggest/src/java/org/apache/lucene/search/suggest/document/CompletionPostingsFormat.java
@@ -104,9 +104,7 @@ public abstract class CompletionPostingsFormat extends PostingsFormat {
   static final String INDEX_EXTENSION = "cmp";
   static final String DICT_EXTENSION = "lkp";
 
-  /**
-   * Creates a {@link CompletionPostingsFormat} with the given name.
-   */
+  /** Creates a {@link CompletionPostingsFormat} with the given name. */
   public CompletionPostingsFormat(String name) {
     super(name);
   }
--- a/lucene/suggest/src/java/org/apache/lucene/search/suggest/document/CompletionPostingsFormat.java
+++ b/lucene/suggest/src/java/org/apache/lucene/search/suggest/document/CompletionPostingsFormat.java
@@ -105,8 +105,7 @@ public abstract class CompletionPostingsFormat extends PostingsFormat {
   static final String DICT_EXTENSION = "lkp";
 
   /**
-   * Creates a {@link CompletionPostingsFormat} that will use the provided <code>fstLoadMode</code>
-   * to determine if the completion FST should be loaded on or off heap.
+   * Creates a {@link CompletionPostingsFormat} with the given name.
    */
   public CompletionPostingsFormat(String name) {
     super(name);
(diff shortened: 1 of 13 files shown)


lucene/suggest/src/java/org/apache/lucene/search/suggest/document/CompletionPostingsFormat.java
```diff
@@ -122,11 +122,6 @@ public enum FSTLoadMode {
 
   private final FSTLoadMode fstLoadMode;
 
-  /** Used only by core Lucene at read-time via Service Provider instantiation */
-  public CompletionPostingsFormat(String name) {
```
javanna, line 126: This was unused and redundant.

lucene/suggest/src/test/org/apache/lucene/search/suggest/document/TestSuggestField.java
```diff
@@ -951,7 +951,16 @@ static IndexWriterConfig iwcWithSuggestField(Analyzer analyzer, final Set<String
...
+          // FST load mode can only be overridden via a custom completion postings format
+          final PostingsFormat postingsFormat =
+              new CompletionPostingsFormat("Completion101", fstLoadMode) {
+                final CompletionPostingsFormat delegate = new Completion101PostingsFormat();
+
+                @Override
+                protected PostingsFormat delegatePostingsFormat() {
+                  return delegate.delegatePostingsFormat();
+                }
+              };
```
jpountz, line 963: FWIW this may work accidentally on NRT segments if IndexWriter opens them with the codec configured on the IndexWriterConfig, but it you were to open an index from a Directory using this postings format, it would open it using the default FST load mode and ignore the one passed to the constructor here. This is because, Lucene would read "Completion101" from the segment infos and look it up through SPI, which would return the default format that is registered under this name rather than this instance that you are creating here.

I wonder if we should remove FSTLoadMode completely. To your point, it's trappy to think that this thing is actually configurable?

javanna: you are right, I tried to quickly hack this together and I ended up making the same mistake we already had in existing tests. Shame on me! I really do need to register the custom postings format to SPI for this to leverage the random fst load mode. I was also thinking about the option to remove configurability of fst load mode. That would certainly simplify things and avoid this same mistake in the future. Either that, or have a proper example of how the postings format load mode can be customized.

javanna: I updated this PR to remove the option to configure the load mode entirely.

jpountz (approved):
LGTM, don't miss adding a CHANGES entry. :) We may want to add back-compat for `NRTSuggester#load(IndexInput, FstLoadMode)` upon backporting?

jpountz:
I was thinking of keeping the `load(IndexInput, FSTLoadMode)` static method, documenting that the load mode is ignored and deprecating it. Indeed that would require keeping the `FSTLoadMode` enum, which would be deprecated too.

javanna:
Cool then I will target this PR at main only, and open a separate PR for `branch_10x`. Out of curiosity, what are the usecases where you'd expect users to call `NRTSuggester#load` directly?

jpountz:
Good question. The class is public and looked like a user-facing API hence my comment, but you can't serialize a NRTSuggester yourself since `NRTSuggesterBuilder` is pkg-private. So it looks like this load method should really be pkg-private too?

javanna:
I merged main in after merging #14372 and added a changelog entry. I believe this is ready to go, and can now be backported to branch_10x as-is.

jpountz:
Agreed!

javanna:
Thanks @jpountz for all the help!

PR #14368, opened 2025-03-18
Align comments with upgraded postings format

amosbird:
### Description

Update prefetch heuristic comments to reflect that skip data is now inlined into postings lists.


jpountz (approved):

jpountz:
Good catch!

PR #14369, opened 2025-03-19
Fix for changelog verifier and milestone setter automation

pseudo-nymous:
### Description
This pull request contains a fix for changelog automation that has been added recently. We have seen failures where either diff calculation wrt base commit was wrong or base commit was not even available leading to fatal error. 

This happens because we are using `pull_request_target` event type which checks out `merge base commit` where as we want to checkout `head commit` to do proper diff calculation wrt changelog file. My initial testing was based on `pull_request` which worked as expected since it checkouts `merge head commit`.

We tried a fix for this here: https://github.com/apache/lucene/pull/14355. But we didn't fetch the full history appropriately. This PR adds a fix for this by pulling all the associated history using `git pull --unshallow origin ${{ github.event.pull_request.head.ref }}`.

Tested all the changes here: https://github.com/pseudo-nymous/lucene/pull/22

Addresses: https://github.com/apache/lucene/issues/13898 https://github.com/apache/lucene/issues/14190

### DIFF
--- a/.github/workflows/verify-changelog-and-set-milestone.yml
+++ b/.github/workflows/verify-changelog-and-set-milestone.yml
@@ -19,7 +19,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: ${{ github.event.pull_request.head.repo.full_name }}
-          ref: ${{ github.event.pull_request.head.ref }}
+          ref: ${{ github.event.pull_request.head.sha }}
 
       - name: ChangeLog Entry Verifier and Milestone Setter
         run: |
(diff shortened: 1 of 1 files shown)


pseudo-nymous:
We can fetch all history using checkout actions itself using flag `fetch-depth: 0`. But it fetches all the history for all branches and tags which is not required here.

stefanvodita:
I'm happy to try this out, but I have some doubts it addresses the issue we're experiencing now. For example, take the failure [here](https://github.com/apache/lucene/actions/runs/13876712827/job/38829937408).

In step 1, we `Could not resolve to a PullRequest with the number of 14301.` The PR definitely exists.
Then, in step 2 we get `fatal: bad object`. The fix here is meant to fix that, right? But I wonder if the issue goes deeper.

What do you think @pseudo-nymous?

pseudo-nymous:
Yes, fix here would fix the `fatal: bad object` failue. I haven't seen the first failure before, let me address this.

pseudo-nymous:
Moved it to draft state to address all the failures first.

stefanvodita:
@pseudo-nymous, I'm curious if you had any new thoughts on this. I think to test this properly, we would need two GitHub accounts each with their own fork of a repo trying to open a PR against one another.

pseudo-nymous:
I think I have found the cause of the failure. It's related to fork repository checkout. When we checkout a fork's repository, github api gets linked to forked repo while github event payload is liked with apache repo. That's why when we try to fetch labels associated with PR, it fails since PR number does not exist in user's fork but in apache repo.

I'm trying out a fix by using a second fork on my own fork. I'll update the PR with the fixes.

stefanvodita:
I have been looking into this on my end as well. Agree that the issue arises when we have a PR straddling different forks and it sounds like you have a good theory about the underlying mechanism. Let me know about your findings!

stefanvodita:
@pseudo-nymous, I have my proposed fix in #14606. Your thoughts on it are much appreciated!

pseudo-nymous:
Sorry, I didn't get a chance to pick this up.
Thanks for making the changes. It looks good to me.

pseudo-nymous:
My changes are not required now. Thus, closing this PR.

PR #14370, opened 2025-03-19
Add leafReaders() Method to IndexReader and Unit Test

### LINKED ISSUE #14367: IndexReader#leaves method is slightly confusing
### Description

I was slightly confused that the `leaves` method within both `IndexReader` and `IndexReaderContext` returns `List<LeafReaderContext>`. I was expecting `IndexReader#leaves` to be `List<LeafReader>`, and maybe have another method called `IndexReader#leafContexts` to return `List<LeafReaderContext>`

Really minor at this point, and probably not worth going through the pain of deprecating `IndexReader#leaves` and changing at few hundred places

DivyanshIITB:
This PR introduces leafReaders() in IndexReader for direct access to LeafReader instances, improving usability over leaves(). A corresponding unit test ensures correctness by validating retrieval consistency and resource management.

- Provides a convenient way to access leaf readers without manually iterating over leaves().
- Enhances code readability and usability for developers working with Lucene’s indexing system.
- Ensures functionality is tested to prevent regressions.

Testing & Validation:
✅ Successfully retrieves LeafReader instances.
✅ Ensures the size matches reader.leaves().size().
✅ Properly closes resources to avoid memory leaks.

Fixes #14367

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexReader.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexReader.java
@@ -127,6 +127,16 @@ public abstract sealed class IndexReader implements Closeable permits CompositeR
      */
     void addClosedListener(ClosedListener listener);
   }
+  
+  /**
+   * Returns a list of {@link LeafReader} extracted from {@link LeafReaderContext}.
+   * This provides a direct way to get the leaf readers of an index.
+   *
+   * @return List of {@link LeafReader}
+   */
+  public List<LeafReader> leafReaders() {
+    return leaves().stream().map(LeafReaderContext::reader).collect(Collectors.toList());
+  }
 
   /** A cache key identifying a resource that is being cached on. */
   public static final class CacheKey {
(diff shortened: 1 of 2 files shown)


benwtrent:
Eh, I am not sold that this change needs to occur if ever. While, "this is how its always been" isn't a good argument for some things, I think expanding the public, and then backwards compatible API needs careful consideration. 

I am not sure the syntactic sugar is worth the cavities.

If others in the community really want this, I am cool with it. My opinion is that it isn't necessary.

jainankitk:
> Eh, I am not sold that this change needs to occur if ever. While, "this is how its always been" isn't a good argument for some things, I think expanding the public, and then backwards compatible API needs careful consideration.

I agree with @benwtrent. That's why I had below comment in [the original issue](https://github.com/apache/lucene/issues/14367):

_Really minor at this point, and probably not worth going through the pain of deprecating IndexReader#leaves and changing at few hundred places_

Additionally, I don't think this change really addresses the confusion being called out in https://github.com/apache/lucene/issues/14367. Ideally, the default behavior of `IndexReader#leaves` should be what is being achieved here via `IndexReader#leafReaders`, and there should be more explicit method called `IndexReader#leafReaderContexts` for what `IndexReader#leaves` is doing today

jpountz:
Agreed with @benwtrent and @jainankitk. I'm closing this PR.

PR #14373, opened 2025-03-19
Optimize ParallelLeafReader to improve term vector fetching efficienc

### LINKED ISSUE #7926: ParallelLeafReader.getTermVectors can indirectly load TVs multiple times [LUCENE-6868]
ParallelLeafReader has a getTermVectors(docId) implementation that loops over each field it has in a loop and calls getTermVector(docId,fieldName).  But the implementation of that will load all term vectors for all fields in that reader, yet ParallelLeafReader only wants one.  The effect is an O(n^2) where 'n' is the number of fields, when we could get O\(n) if we do it right. PLR should call getTermVectors(docId) (not referring to a specific field) for each of it's readers and then aggregate them.

This wouldn't be such a problem if our term vector API/Codec was improved to not load all term vectors for all fields from disk at once.

Found via randomized-testing of IndexWriter auto-picking ParallelAtomicReader along with a test I have that asserts TVs aren't fetched for a doc more than once.

---
Migrated from [LUCENE-6868](https://issues.apache.org/jira/browse/LUCENE-6868) by David Smiley (@dsmiley)

DivyanshIITB:
This PR optimizes ParallelLeafReader to avoid redundant term vector fetching.
- Replaces per-field term vector fetching with a single call per reader.
- Reduces complexity from O(n^2) to O(n).
- Improves performance when handling large numbers of fields.
- Verified via existing tests.

Closes #7926

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -36,9 +36,6 @@ Optimizations
 * GITHUB#14011: Reduce allocation rate in HNSW concurrent merge. (Viliam Durina)
 * GITHUB#14022: Optimize DFS marking of connected components in HNSW by reducing stack depth, improving performance and reducing allocations. (Viswanath Kuchibhotla)
 * GITHUB#14373: Optimized `ParallelLeafReader` to improve term vector fetching efficiency.
-- Fetches all term vectors once per reader instead of per field.
-- Reduces complexity from **O(n²) to O(n)**.
-- Enhances performance for documents with many fields. (Divyansh Agrawal)
 
 Bug Fixes
 ---------------------
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -35,6 +35,10 @@ Optimizations
 ---------------------
 * GITHUB#14011: Reduce allocation rate in HNSW concurrent merge. (Viliam Durina)
 * GITHUB#14022: Optimize DFS marking of connected components in HNSW by reducing stack depth, improving performance and reducing allocations. (Viswanath Kuchibhotla)
+* GITHUB#14373: Optimized `ParallelLeafReader` to improve term vector fetching efficiency.
+- Fetches all term vectors once per reader instead of per field.
+- Reduces complexity from **O(n²) to O(n)**.
+- Enhances performance for documents with many fields. (Divyansh Agrawal)
 
 Bug Fixes
 ---------------------
--- a/lucene/core/src/java/org/apache/lucene/index/ParallelLeafReader.java
+++ b/lucene/core/src/java/org/apache/lucene/index/ParallelLeafReader.java
@@ -348,15 +348,24 @@ public class ParallelLeafReader extends LeafReader {
       @Override
       public Fields get(int docID) throws IOException {
         ParallelFields fields = null;
-        for (Map.Entry<String, LeafReader> ent : tvFieldToReader.entrySet()) {
-          String fieldName = ent.getKey();
-          TermVectors termVectors = readerToTermVectors.get(ent.getValue());
-          Terms vector = termVectors.get(docID, fieldName);
-          if (vector != null) {
-            if (fields == null) {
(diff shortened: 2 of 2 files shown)


DivyanshIITB:
Just a gentle reminder 
@vigyasharma

lucene/core/src/java/org/apache/lucene/index/ParallelLeafReader.java
```diff
@@ -348,15 +348,24 @@ public void prefetch(int docID) throws IOException {
...
+
+          if (docFields != null) {
+              if (fields == null) {
+                  fields = new ParallelFields();
+              }
+
+              // Step 3: Aggregate only required fields
+              for (String fieldName : docFields) {
+                  Terms vector = docFields.terms(fieldName);
+                  if (vector != null) {
```
vigyasharma, line 365: When would this be null? Since we're going through fields returned by `termVectors.get(docId)`, the field should exist and have terms.

vigyasharma (commented):
Changes look good to me. Can you run `./gradlew tidy` to fix formatting issues, and add a changes entry before we merge this?

lucene/core/src/java/org/apache/lucene/index/ParallelLeafReader.java
DivyanshIITB: `docFields.terms(fieldName)` can be null if the field exists in `docFields` but does not have term vectors stored. This check ensures we only add fields that actually have term vectors

DivyanshIITB:
> Changes look good to me. Can you run `./gradlew tidy` to fix formatting issues, and add a changes entry before we merge this?

I successfully ran `./gradlew tidy` and the built was successful. 
Also added the changes entry in `CHANGES.txt`

lucene/core/src/java/org/apache/lucene/index/ParallelLeafReader.java
vigyasharma: > `docFields.terms(fieldName)` can be null if the field exists in `docFields` but does not have term vectors stored.

Okay. I thought `termVectors.get(docID)` only returns Fields that have terms.

lucene/CHANGES.txt
```diff
@@ -35,6 +35,10 @@ Optimizations
 ---------------------
 * GITHUB#14011: Reduce allocation rate in HNSW concurrent merge. (Viliam Durina)
 * GITHUB#14022: Optimize DFS marking of connected components in HNSW by reducing stack depth, improving performance and reducing allocations. (Viswanath Kuchibhotla)
+* GITHUB#14373: Optimized `ParallelLeafReader` to improve term vector fetching efficiency.
+- Fetches all term vectors once per reader instead of per field.
+- Reduces complexity from **O(n²) to O(n)**.
+- Enhances performance for documents with many fields. (Divyansh Agrawal)
```
vigyasharma, line 41: We generally keep a single bullet per changes entry. Details are already available in the pull request that the entry points too.

vigyasharma:
> I successfully ran `./gradlew tidy` and the built was successful.

Github build is still failing on spotless (formatting). `tidy` will change and reformat offending files for you, you need to commit and push those changes.

lucene/CHANGES.txt
DivyanshIITB: I have modified `CHANGES.txt` as you said.

DivyanshIITB:
> > I successfully ran `./gradlew tidy` and the built was successful.
> 
> Github build is still failing on spotless (formatting). `tidy` will change and reformat offending files for you, you need to commit and push those changes.

Thanks for the review! 🙌 I have run `./gradlew tidy` and pushed the formatting fixes. Let me know if there's anything else needed.

vigyasharma (approved):

vigyasharma:
Changes merged. Thanks @DivyanshIITB !

PR #14377, opened 2025-03-20
Remove nonexistent PackedBlockLength reference in document

amosbird:
### Description

Remove nonexistent `PackedBlockLength` reference in document. This seems to be a documentation artifact from version 912 onward, with no corresponding implementation found in the codebase.


jpountz (approved):
Thanks, you are right. I'll merge.

PR #14381, opened 2025-03-21
RegExp: add CASE_INSENSITIVE_RANGE support

### LINKED ISSUE #14378: Case insensitive regex query with character range
### Description

Hi,
I'm implementing regex search in java app with Elasticsearch 8.* and I've noticed unexpected behaviour with `CASE_INSENSITIVE` flag.

It seems that Lucene ignores the flag in case of character ranges and I'm getting the following results when searching fields analyzed by KeywordAnalyzer:

```
"[abc]+" -> "abc", "ABC"
"[ABC]+" ->  "abc", "ABC"
"[a-c]+" -> "abc"
"[A-C]+" -> "ABC"
```

Reproduced in https://github.com/petrsimon/lucene-regex

_No response_

rmuir:
Add optional flag to support case-insensitive ranges. A minimal DFA is always created. This works with Unicode but may have a performance cost.

Each codepoint in the range must be iterated, and any alternatives added to a set. This can be large if the range spans much of Unicode.

CPU and memory costs are contained within a single function enabled by the optional flag. For example when matching a caseless `/[a-z]/`, 54 codepoints will be accumulated into an `int[]`, which is then compressed to 4 ranges before adding to the parse tree.

Closes #14378

Here's what resulting `/[a-z]/` automaton looks like in case you are curious:
```dot
digraph Automaton {
  rankdir = LR
    node [width=0.2, height=0.2, fontsize=8]
    initial [shape=plaintext,label=""]
    initial -> 0
    0 [shape=circle,label="0"]
    0 -> 1 [label="A-Z"]
    0 -> 1 [label="a-z"]
    0 -> 1 [label="ſ"]
    0 -> 1 [label="K"]
    1 [shape=doublecircle,label="1"]
}

```


lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
```diff
@@ -778,6 +786,53 @@ private int[] toCaseInsensitiveChar(int codepoint) {
...
+          scratch.addTransition(state, state, alt);
+        }
+      } else {
+        int altCase =
+            Character.isLowerCase(i) ? Character.toUpperCase(i) : Character.toLowerCase(i);
+        if (altCase != i) {
+          scratch.addTransition(state, state, altCase);
+        }
+      }
+    }
```
dweiss, line 821: I think this pattern is used in more than one place. Maybe it'd be nicer to make CaseFolding.forAllAlternatives(altCase -> scratch.addTransition(state, state, altCase)) and have all that logic for checking for alternatives embedded in CaseFolding?

dweiss (approved):
Very nice.

lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
rmuir: good call. this is better than returning mutable arrays that could get messed up by bugs, or creating gazillions of arrays.

rmuir: this one is best as a separate PR. I will work it today.

rmuir:
after fixing the turkish here's the (correct) automaton for `/[a-z]/`: the only special cases are long-s and kelvin sign as you expect:

graphviz (6)

rmuir:
@dweiss thanks for the suggestion there, gazillions of array creations avoided. so now this thing will only spike cpu during parsing at worst. I honestly forget you can pass functions to functions in java now :)

PR #14384, opened 2025-03-21
build: generate CaseFolding.java from "gradle regenerate"

rmuir:
Generate this file via ICU with "gradle regenerate". We can control the supported version by upgrading ICU.

Remove turkish mappings, these were wrongly mixed in and only make things worse. No regex engine supports these alt case mappings as an option, including java. We can always add separate support if truly needed, but it should not be mixed in with normal mappings.

Remove random tests comparing Pattern results to Regexp. We don't need this complexity to tell us that icu needs to be upgraded.

Change function signature to use IntFunction so that callers can do whatever they want with the mappings. This can help with range folding.

I'd like to merge this before #14381 as it will improve the efficiency there, too.


rmuir:
This one is pretty easy to understand, the `CaseFolding` class now just gives you `UnicodeSet(ch).closeOver(UnicodeSet.SIMPLE_CASE_INSENSITIVE)` without requiring that you have ICU. 

The generation depends strictly upon ICU version (which I will separately upgrade for unicode 16 now that java 24 has it).

dweiss (approved):
LGTM. That regeneration code is something I'll have to work on, eventually. Maybe it can be made simpler.

rmuir:
It was easy because @uschindler already created a similar groovy script before.

rmuir:
I will followup with an ICU upgrade PR to this one. I don't expect that this file will change except for the version in the comment though.

dweiss:
I mean the entire structure of tasks that are used in regenerate. It's complex. I remember I couldn't do it in any easier way before - maybe something has changed that would allow it to be simpler (I doubt though).

rmuir:
There was something about gradle itself that was upset about dependencies wrt generation tasks, if i recall... cycle detection or something was complaining about it.

uschindler (commented):
Looks good. +1

lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
```diff
@@ -759,23 +759,14 @@ private Automaton toAutomaton(
...
-      if (altCase != codepoint) {
-        return new int[] {altCase, codepoint};
-      } else {
-        return new int[] {codepoint};
-      }
-    }
+    List<Integer> list = new ArrayList<>();
+    CaseFolding.expand(
+        codepoint,
+        (int variant) -> {
```
uschindler, line 765: wouldn't have `list::add` as method reference not have worked?

rmuir: I thought about it: am not sure due to boxing. can try again. it is 2025 they still have not fixed the boxes :)

PR #14386, opened 2025-03-21
upgrade icu dependency from 74.2 -> 77.1

rmuir:
Bump dependency
Regenerate locks
Regenerate everything

ICUFoldingFilter:
Remove special-cases from Diacritic Folding, these now have `:Diacritic=Yes:`:
* U+1734 HANUNOO SIGN PAMUDPOD
* U+10A3F KHAROSHTHI VIRAMA

ICUTokenizer:
Synchronize Default.rbbi to reflect upstream ICU-22941 wordbreak updates to `MidLetter`:
* U+003A COLON
* U+FE55 SMALL COLON
* U+FF1A FULLWIDTH COLON

"These tailorings were introduced in ICU 72, but feedback has been negative, and the UTC declined to adopt these changes."


dweiss:
I know, I know. I don't think we should remove it - I just hope it can be implemented in a less hairy way.

dweiss (approved):

Issue #14382, opened 2025-03-21
Reduce memory usage when merging bkd trees


iverase:
When building BKD trees, we hold two arrays in memory which sizes grows linearly with the number of leaf nodes. One of the array contains the pointer to the start of a leaf node, and the other containing the split value. The number of leaf nodes does not grow with the number of documents but with the number of values, therefore in the case of multi-values, those arrays can grow quite big. 

The situation is particularly inefficient for the `OneDimensionBKDWriter` where we are using a List to hold the split values. I wonder if we can use more efficient data structures to lower the heapusage.

For example, maybe we can use the `FixedLengthBytesRefArray` to hold split values or used some packing algorithm to hold the leaf pointers.

iverase:
We are using more dense data structures now, in particular for the OneDimensionBKDWriter.

PR #14393, opened 2025-03-24
Pack file pointers when merging BKD trees

iverase:
We are currently using long arrays to hold file pointers. These arrays can get pretty big if the number of points is big which seems wasteful, moreover when those file pointers are monotonically increasing. This commit proposes to pack these arrays using existing lucene algorithms to save some heap during merging and to avoid humongous allocations. 

relates https://github.com/apache/lucene/issues/14382

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
@@ -584,7 +584,7 @@ public class BKDWriter implements Closeable {
 
     scratchBytesRef1.length = config.bytesPerDim();
     scratchBytesRef1.bytes = splitPackedValues;
-    final LongValues longValues =
+    final LongValues leafFPLongValues =
         new LongValues() {
           @Override
           public long get(long index) {
@@ -593,7 +593,12 @@ public class BKDWriter implements Closeable {
         };
 
     return makeWriter(
-        metaOut, indexOut, splitDimensionValues, longValues, leafBlockFPs.length, dataStartFP);
+        metaOut,
+        indexOut,
+        splitDimensionValues,
+        leafFPLongValues,
+        leafBlockFPs.length,
+        dataStartFP);
   }
 
   /* In the 1D case, we can simply sort points in ascending order and use the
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
```diff
@@ -582,8 +584,16 @@ private IORunnable writeFieldNDims(
 
     scratchBytesRef1.length = config.bytesPerDim();
     scratchBytesRef1.bytes = splitPackedValues;
+    final LongValues longValues =
```
jpountz, line 587: Can you give it a less generic name?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
```diff
@@ -687,7 +701,11 @@ private class OneDimensionBKDWriter {
     private int leafCount;
     private int leafCardinality;
 
-    OneDimensionBKDWriter(IndexOutput metaOut, IndexOutput indexOut, IndexOutput dataOut) {
+    OneDimensionBKDWriter(
+        IndexOutput metaOut,
+        IndexOutput indexOut,
+        IndexOutput dataOut,
+        LongAccumulator accumulator) {
```
jpountz, line 708: Likewise here, maybe rename to `leafFPAccumulator` or something along these lines?

lucene/core/src/java/org/apache/lucene/util/bkd/BKDWriter.java
```diff
@@ -596,9 +606,12 @@ private IORunnable writeField1Dim(
...
       throws IOException {
     MutablePointTreeReaderUtils.sort(config, maxDoc, reader, 0, Math.toIntExact(reader.size()));
-
+    final int numLeaves =
+        Math.toIntExact(
+            (reader.size() + config.maxPointsInLeafNode() - 1) / config.maxPointsInLeafNode());
+    checkMaxLeafNodeCount(numLeaves);
     final OneDimensionBKDWriter oneDimWriter =
-        new OneDimensionBKDWriter(metaOut, indexOut, dataOut);
+        new OneDimensionBKDWriter(metaOut, indexOut, dataOut, new ArrayLongAccumulator(numLeaves));
```
jpountz, line 614: Could/should we use the "packed" impl all the time?

jpountz: Intuitively, there is ~512x (the leaf node size) more data than file pointers, so the bottleneck of flushing/merging should be on the data rather than on file pointers. So having small overhead on file pointers should be ok (hopefully benchmarks will confirm this).

iverase: Right, let's use it all the time for OneDimensionBKDWriter, it clearly simplifies the implementation

jpountz (approved):

PR #14395, opened 2025-03-24
Disable sort optimization when tracking all docs

bugmakerrrrrr:
### Description

When `totalHitsThreshold` is greater than or equal to number of docs, we can disable the sort optimization.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
+++ b/lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
@@ -61,7 +61,7 @@ public abstract class TopFieldCollector extends TopDocsCollector<Entry> {
       }
       if (countAllDocs == null) {
         final int maxDoc = ReaderUtil.getTopLevelContext(context).reader().maxDoc();
-        if (countAllDocs = totalHitsThreshold >= maxDoc) {
+        if (countAllDocs = (totalHitsThreshold >= maxDoc)) {
           firstComparator.disableSkipping();
         }
       }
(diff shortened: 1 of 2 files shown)


jpountz:
The change looks correct to me. With recent changes to allow clauses that match all docs to remove themselves from a conjunction, it should be possible to achieve something similar by implementing `#docIDRunEnd()` on competitive iterators. I need to think a bit more about the pros and cons of these two approaches.

bugmakerrrrrr:
> The change looks correct to me. With recent changes to allow clauses that match all docs to remove themselves from a conjunction, it should be possible to achieve something similar by implementing `#docIDRunEnd()` on competitive iterators. I need to think a bit more about the pros and cons of these two approaches.

@jpountz If I understand correctly, I think both optimizations can be implemented at the same time, and there is no conflict between the two. If we know we can't skip any docs before collection, then there's no need to maintain competitiveIterator-related data, and it helps for implementations that don't benefit from `docIDRunEnd`. Instead, `docIDRunEnd` can implement the skip logic during runtime.

Of course, the current implementation only informs the comparator once, but if we could inform each segment separately, we could also disable/enable sort on the fly based on the current total hits and max doc of current segment.

lucene/core/src/java/org/apache/lucene/search/TopFieldCollector.java
```diff
@@ -59,6 +59,13 @@ private abstract class TopFieldLeafCollector implements LeafCollector {
           firstComparator.disableSkipping();
         }
       }
+      if (countAllDocs == null) {
```
vibrantvarun, line 62: Can we a comment here on which scenario this optimization will be useful? just like how line 53 has it for the following code.

vibrantvarun:
Overall code looks good to me.

PR #14405, opened 2025-03-26
Adding TestSpanWithinQuery with basic test cases for SpanWithinQuery

slow-J:
TEST: ./gradlew check

### Description

I was looking at an old issue https://github.com/apache/lucene/issues/7145 which talks about unit tests for SpanWithinQuery. I noticed that there was no class for basic unit tests for SpanWithinQuery, while we do this for many other SpanQueries.


vigyasharma (approved):
Looks good, thanks for adding these tests!

PR #14411, opened 2025-03-26
Revert "gh-12627: HnswGraphBuilder connects disconnected HNSW graph components (#13566)"

txwei:
This reverts commit 217828736c41bfc68065ceb3d5b37c47116ea947 because [HNSW connect components can take an inordinate amount of time](https://github.com/apache/lucene/issues/14214).


benwtrent:
@txwei have you seen this behavior in production? I am wondering on the urgency.

txwei:
@benwtrent we haven't released this to prod yet. We spotted a severe perf regression with the degenerate case of all vectors being the same. Since we have enough customers using all kinds of embeddings in production, this is very likely to be an issue.

Issue #14183, opened 2025-01-29
Allow skip_factor to be set dynamically within QueryCache


sgup432:
### Description

I see there have been many discussions around finding the right value for skip_factor ([here](https://issues.apache.org/jira/browse/LUCENE-9002) and https://github.com/apache/lucene/issues/11081), mostly the concern being that caching large query clauses(as per their cost and eventually size) may not make much sense, considering they will make the overall query slow due to cache overhead(locks, eviction).

In follow up to this issue - https://github.com/apache/lucene/issues/14028, I see by relaxing `skip_factor` limits, one was able to utilize query cache in a much better way and improve performance.  As of today, one can only change this during cache creation [here](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java#L124) and not dynamically.

Can we allow this to be set dynamically within QueryCache, enabling users to increase or decrease it as needed during runtime? It seems like a harmless change and could be quite useful.

jpountz:
I am not entirely sure about the "harmless" part: this class is shared by multiple threads so we would need to make sure that the value is updated in a thread-safe way.

sgup432:
@jpountz Yeah I meant we could use something like AtomicReference to set it dynamically in a thread safe way.
I meant harmless in a sense that it should be easy to implement via getter/setter.


Something like below:

```
private final AtomicReference<Float> skipCacheFactor = new AtomicReference<>(10f);

public float getSkipCacheFactor() {
        return skipCacheFactor.get();
    }

    public void setSkipCacheFactor(float value) {
        skipCacheFactor.set(value);
    }
```

sgup432:
@jpountz If you think it is feasible via above approach, I can quickly raise a PR on this with some UTs.

jpountz:
In general I'm not a fan of exposing tuning knobs just because we can expose them. Deciding when a clause is worth caching feels like something that Lucene is the right decision maker for, rather than the application. Maybe you can share more about how you plan on tuning this knob?

sgup432:
We're considering adjusting the skip_factor limits dynamically to optimize query cache usage when needed, especially when it's underutilized. By exposing this as a dynamic cluster setting in OpenSearch, we can adjust the skip_factor value based on query cache utilization. We've received feedback from several users/customers who have expressed concerns that the query cache isn't being fully utilized in their domain, unlike in previous versions where the skip_factor was set to a higher default.

kkewwei:
[...]

Issue #14222, opened 2025-02-12
Refactor QueryCache to improve concurrency and performance


sgup432:
### Description

Given the significant role of [LRUQueryCache](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java) in Lucene, I see opportunities to enhance its performance. Although there have been many discussions on this topic like [here](https://github.com/apache/lucene/issues/10081), they haven’t led to anywhere. 

Currently, QueryCache uses a `Map<CacheKey, LeafCache>`, where:

- **CacheKey** is tied to a specific segment.
- **LeafCache** is per-segment and maintains a Map<Query, CacheAndCount>, where CacheAndCount is essentially a docID set.

### Current Issue:
We use a global read/write lock to put items into the cache. This allows only one writer at a time holding other writers or other readers.

Even if we were to introduce multiple read/write locks(via multiple partitions), it won't help with `CacheKey` being our key. As if there are only 5 segments but 10,000 unique queries, we end up with just 5 unique cache keys. This means a large number of requests compete for the same write lock, significantly degrading performance by either waiting or not using cache (by skipping it if lock not obtained).

### One simple Solution:

- **Introduce a Composite Key**
  - Instead of CacheKey alone, use (CacheKey, Query) as the key.
  - The new structure would be `Map<CompositeKey, CacheCount>`

- **Partition the Cache for Better Concurrency**
  - Create multiple LRU partitions (e.g., 8, 16, or 64 , configurable).
  - Each partition has its own read/write lock to reduce contention.
  - Assign incoming requests to specific partition using:
      `hash(CacheKey, Query) % partition_count` (as an example)

- I also see we use uniqueQueries/mostRecentlyUsedQueries [here](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java#L94-L98), to maintain a LRU list for the queries. I think we can remove it and instead maintain a separate LRU list per partition, using CompositeKey.

- For Eviction & Cleanup, we can collect stale `CacheKey` entries in a `Set<CacheKey>` when segment is closed due to merge. A background thread later iterates over cache entries and clears them, similar to how RequestCache in OpenSearch works.

Let me know what you think. I think it might work pretty well(and simple enough) but I am open to other ideas.

sgup432:
I almost have a prototype ready for above solution, I will add benchmarking results here comparing performance.

sgup432:
[...]

PR #14412, opened 2025-03-27
Allow skip cache factor to be updated dynamically

sgup432:
### Description
Related issue - https://github.com/apache/lucene/issues/14183

This change allows skip cache factor to be updated dynamically within LRU query cache. This can be done via getter/setter

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -47,6 +47,7 @@ Other
 ---------------------
 
 * GITHUB#14229: Bump minimum required Java version to 25
+* GITHUB#14412: Allow skip cache factor to be updated dynamically. (Sagar Upadhyaya)
 
 ======================= Lucene 10.2.0 =======================
 
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -47,7 +47,6 @@ Other
 ---------------------
 
 * GITHUB#14229: Bump minimum required Java version to 25
-* GITHUB#14412: Allow skip cache factor to be updated dynamically. (Sagar Upadhyaya)
 
 ======================= Lucene 10.2.0 =======================
 
@@ -103,6 +102,8 @@ New Features
 
 * GITHUB#13470: Added `TopDocs#rrf` to combine multiple TopDocs instances using
   reciprocal rank fusion. (Haren Lin, Adrien Grand)
+
+* GITHUB#14412: Allow skip cache factor to be updated dynamically. (Sagar Upadhyaya)
 
 Improvements
 ---------------------
--- a/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
+++ b/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
@@ -32,7 +32,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.atomic.LongAdder;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Predicate;
@@ -100,7 +99,7 @@ public class LRUQueryCache implements QueryCache, Accountable {
   private final Map<IndexReader.CacheKey, LeafCache> cache;
   private final ReentrantReadWriteLock.ReadLock readLock;
   private final ReentrantReadWriteLock.WriteLock writeLock;
-  private final AtomicReference<Float> skipCacheFactor;
+  private volatile float skipCacheFactor;
   private final LongAdder hitCount;
   private final LongAdder missCount;
 
@@ -123,30 +122,14 @@ public class LRUQueryCache implements QueryCache, Accountable {
       long maxRamBytesUsed,
       Predicate<LeafReaderContext> leavesToCache,
       float skipCacheFactor) {
-    this(maxSize, maxRamBytesUsed, leavesToCache, new AtomicReference<>(skipCacheFactor));
-  }
-
-  /**
-   * Additionally, allows the ability to pass skipCacheFactor in form of AtomicReference where the
-   * caller can dynamically update(in a thread safe way) its value by calling skipCacheFactor.set()
-   * on their end.
-   */
-  public LRUQueryCache(
-      int maxSize,
-      long maxRamBytesUsed,
-      Predicate<LeafReaderContext> leavesToCache,
-      AtomicReference<Float> skipCacheFactor) {
     this.maxSize = maxSize;
     this.maxRamBytesUsed = maxRamBytesUsed;
     this.leavesToCache = leavesToCache;
-    if (skipCacheFactor == null || skipCacheFactor.get() == null) {
-      throw new IllegalArgumentException("skipCacheFactor should not be null");
-    }
-    if (skipCacheFactor.get() < 1) { //
+    if (skipCacheFactor < 1) { //
       // NaN >= 1
       // evaluates false
       throw new IllegalArgumentException(
-          "skipCacheFactor must be no less than 1, get " + skipCacheFactor.get());
+          "skipCacheFactor must be no less than 1, get " + skipCacheFactor);
     }
     this.skipCacheFactor = skipCacheFactor;
 
@@ -161,8 +144,12 @@ public class LRUQueryCache implements QueryCache, Accountable {
     missCount = new LongAdder();
   }
 
-  AtomicReference<Float> getSkipCacheFactor() {
+  public float getSkipCacheFactor() {
     return skipCacheFactor;
+  }
+
+  public void setSkipCacheFactor(float skipCacheFactor) {
+    this.skipCacheFactor = skipCacheFactor;
   }
 
   /**
@@ -316,8 +303,8 @@ public class LRUQueryCache implements QueryCache, Accountable {
   }
 
   private void putIfAbsent(Query query, CacheAndCount cached, IndexReader.CacheHelper cacheHelper) {
-    assert !(query instanceof BoostQuery);
-    assert !(query instanceof ConstantScoreQuery);
+    assert query instanceof BoostQuery == false;
+    assert query instanceof ConstantScoreQuery == false;
     // under a lock to make sure that mostRecentlyUsedQueries and cache remain sync'ed
     writeLock.lock();
     try {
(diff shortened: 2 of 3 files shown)


sgup432:
@jpountz Might need your review as discussed in https://github.com/apache/lucene/issues/14183

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -122,12 +123,30 @@ public LRUQueryCache(
...
+  /**
+   * Additionally, allows the ability to pass skipCacheFactor in form of AtomicReference where the
+   * caller can dynamically update(in a thread safe way) its value by calling skipCacheFactor.set()
+   * on their end.
+   */
+  public LRUQueryCache(
+      int maxSize,
+      long maxRamBytesUsed,
+      Predicate<LeafReaderContext> leavesToCache,
+      AtomicReference<Float> skipCacheFactor) {
```
jpountz, line 138: Let's avoid exposing the implementation detail of how the skip cache factor is stored internally, ie. skipCacheFactor should be a regular `float` here?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -99,7 +100,7 @@ public class LRUQueryCache implements QueryCache, Accountable {
   private final Map<IndexReader.CacheKey, LeafCache> cache;
   private final ReentrantReadWriteLock.ReadLock readLock;
   private final ReentrantReadWriteLock.WriteLock writeLock;
-  private final float skipCacheFactor;
+  private final AtomicReference<Float> skipCacheFactor;
```
jpountz, line 103: Making it volatile should be enough?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -142,6 +161,10 @@ public LRUQueryCache(
     missCount = new LongAdder();
   }
 
+  AtomicReference<Float> getSkipCacheFactor() {
+    return skipCacheFactor;
+  }
```
jpountz, line 166: Can you add a regular getter/setter instead?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -269,8 +292,8 @@ boolean requiresEviction() {
   }
 
   CacheAndCount get(Query key, IndexReader.CacheHelper cacheHelper) {
-    assert key instanceof BoostQuery == false;
-    assert key instanceof ConstantScoreQuery == false;
+    assert !(key instanceof BoostQuery);
+    assert !(key instanceof ConstantScoreQuery);
```
jpountz, line 296: Please undo these changes, these `== false` are on purpose (you'll find many of them across the code base) to be more readable.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
sgup432: I actually made it a AtomicReference so that a caller can create that on their end and pass this reference here. Via which the caller can change its value dynamically.

Other way to set it dynamically would have been getter/setter like you mentioned below. But then one cannot use an interface `QueryCache` to initialize `LRUQueryCache`, and they are forced to use concrete implementation(to get/set) which I think is not right.

```
QueryCache lruCache = new LRUQueryCache();
// Cannot accesss getter/setter until unless I typecast or use concrete implementation directly.
```

Let me know what you think.

jpountz: I don't think that this is a reason for exposing it via an AtomicReference instead of a getter/setter. TieredMergePolicy, to give an example, is the same: many of its setters are not exposed on MergePolicy.

sgup432: Ok then, I will change it a getter/setter then.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
sgup432: Yeah considering we are just doing set/get, volatile should be enough.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
sgup432: Changed it back.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -125,7 +125,9 @@ public LRUQueryCache(
     this.maxSize = maxSize;
     this.maxRamBytesUsed = maxRamBytesUsed;
     this.leavesToCache = leavesToCache;
-    if (skipCacheFactor >= 1 == false) { // NaN >= 1 evaluates false
+    if (skipCacheFactor < 1) { //
+      // NaN >= 1
+      // evaluates false
```
jpountz, line 130: Please revert so that we keep rejecting skipCacheFactor=Float.NaN?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -142,6 +144,14 @@ public LRUQueryCache(
     missCount = new LongAdder();
   }
 
+  public float getSkipCacheFactor() {
+    return skipCacheFactor;
+  }
+
+  public void setSkipCacheFactor(float skipCacheFactor) {
+    this.skipCacheFactor = skipCacheFactor;
+  }
```
jpountz, line 153: Can you add javadocs?

jpountz (commented):
I left some comments, can you add a CHANGES entry too so that I can merge?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
sgup432: Added

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
sgup432: Reverted

sgup432:
@jpountz Added a CHANGES entry.

lucene/CHANGES.txt
```diff
@@ -47,6 +47,7 @@ Other
 ---------------------
 
 * GITHUB#14229: Bump minimum required Java version to 25
+* GITHUB#14412: Allow skip cache factor to be updated dynamically. (Sagar Upadhyaya)
```
jpountz, line 50: Can you move the CHANGES entry under version 10.2 below, section "New Features"?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -142,6 +142,20 @@ public LRUQueryCache(
     missCount = new LongAdder();
   }
 
+  public float getSkipCacheFactor() {
```
jpountz, line 145: Can you add javadocs here too? Lucene aims at having all public APIs documented to make javadocs nicer to browse by users. In this case it could mostly refer to the setter, something like

```
Get the skip cache factor.
@see #setSkipCacheFactor
```

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -142,6 +142,20 @@ public LRUQueryCache(
...
+    return skipCacheFactor;
+  }
+
+  /**
+   * This setter enables the skipCacheFactor to be updated dynamically.
+   *
+   * @param skipCacheFactor determines whether we need to skip the cache operation based on cost and
+   *     lead cost.
+   */
+  public void setSkipCacheFactor(float skipCacheFactor) {
```
jpountz, line 155: Can you explain what this factor does, e.g. by copying this from the ctor docs: "clauses whose cost is {@code skipCacheFactor} times more than the cost of the top-level query will not be cached in order to not slow down queries too much."

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
sgup432: Sure!

lucene/CHANGES.txt
sgup432: Added in desired section.

jpountz:
> I am wondering if we should also make maxRamBytesUsed dynamic as well.

What is your use-case for tuning it? Related to my comment on the issue linked to this PR, I worry a bit about optimizing too much for query caching vs. making the query cache less relevant.

To expand a bit more on this, for "good" queries, like term queries or FieldExistQuery that produce good iterators with a low up-front cost, memory would be better spent on the filesystem cache than on the query cache. Boolean queries recently had big improvements that makes the query cache less relevant. (annotations HS, HW and HX at https://benchmarks.mikemccandless.com/CountAndHighHigh.html and https://benchmarks.mikemccandless.com/CountOrHighHigh.html). So that makes the query cache mostly relevant for queries that have a high up-front cost, like point queries and multi-term queries. Rather than investing in the query cache for these queries, I would rather invest in making them behave better, e.g. recent work on vectorizing this up-front cost for `PointRangeQuery` so that it's less likely to be the bottleneck of query evaluation, or `IndexOrDocValuesQuery` to skip this up-front cost entirely when it's hurting more than helping.

kkewwei:
Thank you for your reply. I understand your perspective. You aim to optimize execution efficiency rather than focus on cache optimization.

In certain rare scenarios, the querycache memory becomes full. To rectify the querycache situation, we are currently compelled to restart the JVM, which differs from your line of thought.

I'm wondering if it would be feasible to expose all querycache cache APIs and eliminate any cache strategies within Lucene itself. If Lucene already has a querycache, then upper - level services like Elasticsearch or OpenSearch could refrain from implementing their own querycache. 

Please ignore this suggestion if it doesn't seem viable.

jpountz:
Sorry I'm not sure I get your suggestion.

sgup432:
@jpountz Thanks for expanding on the reasoning above w.r.t. query cache usage. I was working on [refactoring](https://github.com/apache/lucene/issues/14222) query cache so that it is not a bottleneck itself when utilized well. Maybe we can an additional caching policy wrt to determine high upfront cost or query latency, so that we only cache queries which will give us the benefit. Let me know if its still worth a shot, meanwhile we can work on optimizing the queries itself. I can open up a PR soon for that.

Issue #14375, opened 2025-03-19
Handling concurrent search in QueryProfiler


jainankitk:
### Description

Based on the discussion from [this email thread](https://lists.apache.org/thread.html/r7957a2d9ca38af45b1c370753b3c10542fd9faaf9bf95944c5224e12%40%3Cdev.lucene.apache.org%3E), https://github.com/apache/lucene/pull/144 added logic for compiling timings for different pieces of a query or multiple queries. The profiler logic doesn't account for multiple slices of concurrent segment search.

We recently introduced [ConcurrentQueryProfiler](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/profile/query/ConcurrentQueryProfiler.java) . It tries to merge the information across multiple slices into avg, min, max for each count/time metric, but the output is rather verbose (pasted at the end). I am wondering if this is something Lucene will benefit from? 

```
"query":
                        [
                            {
                                "type": "TermQuery",
                                "description": "field1:one",
                                "time_in_nanos": 3944209,
                                "max_slice_time_in_nanos": 3944209,
                                "min_slice_time_in_nanos": 3892625,
                                "avg_slice_time_in_nanos": 3918417,
                                "breakdown":
                                {
                                    "max_match": 0,
                                    "set_min_competitive_score_count": 0,
                                    "match_count": 0,
                                    "avg_score_count": 4,
                                    "shallow_advance_count": 0,
                                    "next_doc": 50625,
                                    "min_build_scorer": 1327791,
                                    "score_count": 8,
                                    "compute_max_score_count": 0,
                                    "advance": 96583,
                                    "min_set_min_competitive_score": 0,
                                    "min_advance": 4042,
                                    "score": 96751,
                                    "avg_set_min_competitive_score_count": 0,
                                    "min_match_count": 0,
                                    "avg_score": 65438,
                                    "max_next_doc_count": 7,
                                    "max_compute_max_score_count": 0,
                                    "avg_shallow_advance": 0,
                                    "max_shallow_advance_count": 0,
                                    "set_min_competitive_score": 0,
                                    "min_build_scorer_count": 2,
                                    "next_doc_count": 8,
                                    "min_match": 0,
                                    "avg_next_doc": 26250,
                                    "compute_max_score": 0,
                                    "min_set_min_competitive_score_count": 0,
                                    "max_build_scorer": 1722750,
                                    "avg_match_count": 0,
                                    "avg_advance": 50125,
[...]

PR #14413, opened 2025-03-27
Adding profiling support for concurrent segment search

jainankitk:
### Description

This code change introduces `AbstractQueryProfilerBreakdown` that can be extended by `ConcurrentQueryProfilerBreakdown` to show query profiling information for concurrent search executions

### Issue

Relates to https://github.com/apache/lucene/issues/14375


jpountz:
Can you explain why we need two impls? I would have assumed that the `ConcurrentQueryProfilerBreakdown` could also be used for searches that are not concurrent?

jainankitk:
> Can you explain why we need two impls? I would have assumed that the ConcurrentQueryProfilerBreakdown could also be used for searches that are not concurrent?

`ConcurrentQueryProfilerBreakdown` maintains separate instance of `QueryProfileBreakdown` for each segment. In the context method, `ConcurrentQueryProfilerBreakdown` returns corresponding `QueryProfileBreakdown` object for each segment. 

```
  @Override
  public QueryProfilerBreakdown context(LeafReaderContext context) {
    return this;
  }
```

Hence, I felt we don't need to take the overhead of creating breakdown per segment and then merging it together during response. That being said, eventually we can keep just `ConcurrentQueryProfilerBreakdown` if we prefer that for simplicity.

msfroh:
Does it make sense to create a separate `QueryProfilerBreakDown` per leaf? Or should it create one per slice? 

Can this be implemented as part of the `CollectorManager#newCollector` logic?

jpountz:
@jainankitk OK. In my opinion, it's more important to handle the concurrent and non-concurrent cases consistently than to save some overhead when searches are not concurrent. I'd really like non-concurrent search to look and feel like a concurrent search with a single slice running on a SameThreadExecutorService as far as profiling is concerned. So I wouldn't specialize the class hierarchy for concurrent vs. non-concurrent.

jainankitk:
> Does it make sense to create a separate QueryProfilerBreakDown per leaf? Or should it create one per slice?

Actually, create one per slice makes lot of sense.

> Can this be implemented as part of the ProfilerCollectorManager#newCollector logic? Maybe not, since we would also need to profile the work done by the Weight + Scorer on each slice

We can always use `ThreadId` for correctly tracking the mapping from slice to `QueryProfilerBreakdown` object.

jainankitk:
> In my opinion, it's more important to handle the concurrent and non-concurrent cases consistently than to save some overhead when searches are not concurrent. I'd really like non-concurrent search to look and feel like a concurrent search with a single slice running on a SameThreadExecutorService as far as profiling is concerned.

Let me try and see if we can maintain per slice `QueryProfilerBreakdown` object. With that, both concurrent and non-concurrent paths would be consistent as well as efficient.

jainankitk:
One of the failing check is:

```
----------
1. ERROR in /home/runner/work/lucene/lucene/lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryProfilerBreakdown.java (at line 63)
	currentThreadId, ctx -> new QuerySliceProfilerBreakdown());
	                 ^^^
The value of the lambda parameter ctx is not used
----------
1 problem (1 error)

> Task :lucene:sandbox:ecjLintMain FAILED
```

I am wondering if there is a workaround for this? One option is to use `putIfAbsent` which doesn't require function as input, but then need to explicitly `get` before returning

jpountz:
You just need to replace `ctx` with `_`.

jainankitk:
> You just need to replace ctx with _.

Ah, my bad! I tried `.`, but we can't use that as part of variable name. Thanks for the suggestion @jpountz.

At a high level, I have unified the concurrent/non-concurrent profiling paths as suggested. The `QueryProfilerTree` is shared across slices, and we recursively build the ProfilerTree for each slice for response. There are few kinks that we still need to be iron out. For example:

* `Weight` creation is global across slices. How do we account for its time? Should be have separate global tree with just the weight times? We can't just get away with having weight count at the top as `Weight` is shared for child queries as well, right?
* The new in-memory structure for profiled queries is bit like below (notice additional list for slices):
```
"query": [ <-- for list of slices
               [ <-- for list of root queries
              {
                "type": "TermQuery",
                "description": "foo:bar",
                "time_in_nanos" : 11972972,
                "breakdown" :
                {
```
We can probably have map of slices, with key being the `sliceId`:

```
       "query": {
               "some global information":
               "slices": {
               "slice1": [ <-- for list of root queries
                {
                "type": "TermQuery",
                "description": "foo:bar",
                "time_in_nanos" : 11972972,
                "breakdown" :
                {...}}],
               "slice2": [],
               "slice3": []}
              }
```

jainankitk:
@jpountz - Can you provide your thoughts on above?

jainankitk:
> I'd have a top-level tree for everything related to initializing the search and combining results (rewrite(), createWeight(), CollectorManager#reduce) and then a list of trees for each slice.

While working on the code, I realized it is better to have list of slices within the tree itself at each level, instead of repeating the query structure and information across multiple trees. In this approach, we can easily view the tree for specific `sliceId` using jq or simple script. The structure looks like below:
```
"query": [ <-- for list of root queries
              {
                "type": "TermQuery",
                "description": "foo:bar",
                "startTime" : 11972972,
                "totalTime": 354343,
                "breakdown" :   {.....}, <-- query level breakdown like weight count and time
                "sliceBreakdowns": [
                {.....},    <-- first slice information
                {.....}]    <-- second slice information
                "queryChildren": [
                {......},  <-- recursive repetition of above structure
                {......}]
```


> Related, it'd be nice if each per-slice object could also tell us about the thread that it ran in and its start time so that we could understand how exactly Lucene managed to parallelize the search.

Yes, that would be really useful. I have included the threadId as `sliceId`, `startTime` and `totalTime` information for each slice object at every query level.

jainankitk:
@jpountz - The code changes are ready for review. For now, I have made changes to accommodate all the timers in `QueryProfilerTimingType`.

While this does not modify (`rewrite()`, `CollectorManager#reduce`) called out earlier as it is not part of `QueryProfilerTimingType`, it does lay the groundwork using `createWeight()`. We can take those changes in a followup PR.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryProfilerBreakdown.java
```diff
@@ -17,46 +17,113 @@
...
  * time may be composed of several internal attributes (rewriting, weighting, scoring, etc).
  */
 class QueryProfilerBreakdown {
-
-  /** The accumulated timings for this query node */
-  private final QueryProfilerTimer[] timers;
+  private static final Collection<QueryProfilerTimingType> QUERY_LEVEL_TIMING_TYPE =
+      Arrays.stream(QueryProfilerTimingType.values()).filter(t -> !t.isSliceLevel()).toList();
+  private final Map<QueryProfilerTimingType, QueryProfilerTimer> queryProfilerTimers;
+  private final ConcurrentMap<Long, QuerySliceProfilerBreakdown> threadToSliceBreakdown;
```
jpountz, line 40: There seems to be a general problem with how profiling figures out which slice it runs in. We should look into reliably identifying slices, or breaking down by thread rather than by slice?

jainankitk: > We should look into reliably identifying slices, or breaking down by thread rather than by slice?

The code currently breaks down by thread only, even though it mentions slice. I incorrectly assumed that every slice is processed by unique thread, which need not be true. Thinking bit more on this, I feel breakdown by thread is good enough, as our primary intent is to understand the concurrent processing. If slice 1 and slice 2 are processed sequentially by thread A, it is actually better to aggregate that information as thread A breakdown instead of breakdown for slice 1 and slice 2. Thoughts?

jainankitk: I have added some changes to make it more explicit that the breakdowns are per thread and not per slice. Although, underlying class is still `QuerySliceProfileBreakdown` as it can be used for measuring slice level breakdown:

```
"query": [ <-- for list of root queries
              {
                "type": "TermQuery",
                "description": "foo:bar",
                "startTime" : 11972972,
                "totalTime": 354343,
                "queryLevelBreakdowns" :   {.....}, <-- query level breakdown like weight count and time
                "threadLevelBreakdowns": [
                {.....},    <-- first thread information
                {.....}]    <-- second thread information
                "queryChildren": [
                {......},  <-- recursive repetition of above structure
                {......}]
```

jpountz: > if slice 1 and slice 2 are processed sequentially by thread A, it is actually better to aggregate that information as thread A breakdown instead of breakdown for slice 1 and slice 2. Thoughts?

IMO both have merits, so if breaking down by thread is easier, I'm good with it.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryProfilerBreakdown.java
```diff
@@ -17,46 +17,113 @@
...
+      Arrays.stream(QueryProfilerTimingType.values()).filter(t -> !t.isSliceLevel()).toList();
+  private final Map<QueryProfilerTimingType, QueryProfilerTimer> queryProfilerTimers;
+  private final ConcurrentMap<Long, QuerySliceProfilerBreakdown> queryThreadBreakdowns;
 
   /** Sole constructor. */
   public QueryProfilerBreakdown() {
-    timers = new QueryProfilerTimer[QueryProfilerTimingType.values().length];
-    for (int i = 0; i < timers.length; ++i) {
-      timers[i] = new QueryProfilerTimer();
+    queryProfilerTimers = new HashMap<>();
```
jpountz, line 44: Or use an EnumMap?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryProfilerTimer.java
```diff
@@ -33,9 +33,21 @@
  *  </pre>
  */
 class QueryProfilerTimer {
-
   private boolean doTiming;
-  private long timing, count, lastCount, start;
+  private long timing, count, lastCount, start, earliestTimerStartTime;
```
jpountz, line 37: This `earliestTimerStartTime` was useful when breaking down by slice, I'm not sure it's still useful when breaking down by thread?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryProfilerTimingType.java
```diff
@@ -21,16 +21,29 @@
...
+  SCORE(true),
+  SHALLOW_ADVANCE(true),
+  COMPUTE_MAX_SCORE(true),
+  SET_MIN_COMPETITIVE_SCORE(true),
+
+  // IMPORTANT: Global timer types must be defined after all the
+  // slice level timers to preserve the contiguous enum ordinals
+  CREATE_WEIGHT(false);
+
+  private boolean sliceLevel;
```
jpountz, line 38: We don't need this anymore now that we're no longer breaking down by slice?

jpountz (commented):
It looks like it's trying to aggregate data differently for slice-level operations vs. top-level operations, but we shouldn't do this anymore now that we're breaking down by thread?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryProfilerBreakdown.java
jainankitk: Thanks @jpountz. This is a good suggestion, and I missed that while making the changes. But when I started using it, I was curious about the implementation and saw that it initializes reference for all the Enum values during initialization:

```
public EnumMap(Class<K> keyType) {
        this.keyType = keyType;
        keyUniverse = getKeyUniverse(keyType);
        vals = new Object[keyUniverse.length];
    }
```

Since we are using small subset of `QueryProfilerTimingType`, I am wondering if this is a good idea. We can also consider separating `QueryProfilerTimingType` on the basis of slice level or query level.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryProfilerTimer.java
jainankitk: I feel it is still useful, as it is not necessary that all the threads start execution immediately. Intuitively, if we submit 4 slices to executor service, 2 of them might be picked immediately on Thread A, B. Slice 3 starts on another thread C (that was busy executing slice for different query), before thread A, B finish slice 1,2 and thread A/B eventually pick slice 4. Please correct me if I am missing something specific to concurrent implementation for Lucene.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryProfilerTimingType.java
jainankitk: We still share the `CREATE_WEIGHT` metrics across slice worker threads at every query level?

jainankitk:
> It looks like it's trying to aggregate data differently for slice-level operations vs. top-level operations, but we shouldn't do this anymore now that we're breaking down by thread?

IMO, `QuerySliceProfilerBreakdown` and `sliceLevel` in `QueryProfilerTimingType` just declare which metrics can be collected at `sliceLevel` vs `queryLevel`. The specific implementation of `QueryProfileBreakdown` we have, provides per thread aggregated view for these slices. Tomorrow if we have separate implementation of `QueryProfileBreakdown`, that does not aggregate slices together, that can continue to use the same specifications. Maybe we can add thin layer like below to make things more explicit and clear, not sure if it is needed though:

```
public class QueryThreadProfilerBreakdown extends QuerySliceProfilerBreakdown {}
```

jainankitk:
After giving some more thought, added new interface `QueryLeafProfilerAggregator`, implementation of which allows aggregating leaf level breakdowns using `THREAD`, `SLICE` or `LEAF` level. The profiler result has `aggregationType` field that indicates the level of aggregation, and what `id` in the aggregated result will correspond to. `thread` or `slice` or `leaf`

jainankitk:
@jpountz - Thanks for the review earlier. Will really appreciate your feedback on the updates.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/AggregatedQueryLeafProfilerResult.java
```diff
@@ -0,0 +1,65 @@
...
+  public AggregatedQueryLeafProfilerResult(
+      long id, Map<String, Long> breakdown, long startTime, long totalTime) {
+    this.id = id;
+    this.breakdown = Objects.requireNonNull(breakdown, "required breakdown argument missing");
+    this.startTime = startTime;
+    this.totalTime = totalTime;
+  }
+
+  /** Retrieve the lucene description of this query (e.g. the "explain" text) */
+  public long getId() {
```
jpountz: There seems to be a mismatch between javadocs and the method signature.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/AggregatedQueryLeafProfilerResult.java
```diff
@@ -0,0 +1,65 @@
...
+    return id;
+  }
+
+  /** The timing breakdown for this node. */
+  public Map<String, Long> getTimeBreakdown() {
+    return Collections.unmodifiableMap(breakdown);
```
jpountz: You could do the wrapping only once in the constructor?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryProfilerTimingType.java
```diff
@@ -21,16 +21,29 @@
...
+  // slice level timers to preserve the contiguous enum ordinals
+  CREATE_WEIGHT(false);
+
+  private boolean leafLevel;
+
+  private QueryProfilerTimingType(boolean leafLevel) {
+    this.leafLevel = leafLevel;
+  }
+
+  public boolean isLeafLevel() {
```
jpountz, line 45: Add some javadocs? I believe that this means that this operation runs on a LeafReader as opposed to the top-level IndexReader?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryLeafProfilerAggregator.java
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.lucene.sandbox.search;
+
+import java.util.List;
+
+interface QueryLeafProfilerAggregator {
```
jpountz: I wonder if we actually need this interface since it seems to have a single implementation?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryLeafProfilerBreakdown.java
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * A record of timings for the various operations that may happen during query execution. A node's
+ * time may be composed of several internal attributes (rewriting, weighting, scoring, etc).
+ */
+public class QueryLeafProfilerBreakdown {
+  private static final Collection<QueryProfilerTimingType> LEAF_LEVEL_TIMING_TYPE =
+      Arrays.stream(QueryProfilerTimingType.values()).filter(t -> t.isLeafLevel()).toList();
+
+  /** The accumulated timings for this query node */
+  private final QueryProfilerTimer[] timers;
```
jpountz, line 36: What about using an `EnumMap`, which will be implemented by an array like this one under the hood?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryLeafProfilerBreakdown.java
```diff
@@ -0,0 +1,87 @@
...
+    long sliceEndTime = Long.MIN_VALUE;
+    for (QueryProfilerTimingType type : LEAF_LEVEL_TIMING_TYPE) {
+      final QueryProfilerTimer timer = timers[type.ordinal()];
+      // Consider timer for updating start/total time only
+      // if it was used
+      if (timer.getCount() > 0) {
+        sliceStartTime = Math.min(sliceStartTime, timer.getEarliestTimerStartTime());
+        sliceEndTime =
+            Math.max(
+                sliceEndTime, timer.getEarliestTimerStartTime() + timer.getApproximateTiming());
```
jpountz, line 74: It doesn't feel right the compute the end time as the start time plus the approximate timing, since operations will often be interleaved. What about reporting the sum of the approximate timings across operation types instead, ie. the value of `toTotalTime()`?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryProfilerResult.java
```diff
@@ -31,39 +31,69 @@
...
  */
 public class QueryProfilerResult {
+  enum AggregationType {
+    // Leaf level is most verbose, practically no aggregation
+    LEAF,
+    // Aggregate leaf level breakdowns based on slice
+    SLICE,
+    // Aggregate leaf level breakdowns based on thread execution
+    THREAD
+  }
```
jpountz: Is this actually used? I think that from previous discussions we agreed that `SLICE` wouldn't work since Lucene doesn't tell us what slice it's in. And I don't see LEAF being used in this `PR`, only `THREAD`?

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryLeafProfilerThreadAggregator.java
```diff
@@ -0,0 +1,87 @@
...
+package org.apache.lucene.sandbox.search;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+/** Implementation of QueryLeafProfilerAggregator that aggregates leaf breakdowns at thread level */
+class QueryLeafProfilerThreadAggregator implements QueryLeafProfilerAggregator {
+  private final ConcurrentMap<Long, QueryLeafProfilerBreakdown> queryThreadBreakdowns;
```
jpountz: I wonder if this should use the actual `Thread` object as a key and then include it in the result profile, so that users can better know what thread specifically ran some operations.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryLeafProfilerThreadAggregator.java
```diff
@@ -0,0 +1,87 @@
...
+  public long getQueryStartTime() {
+    return queryStartTime;
+  }
+
+  @Override
+  public long getQueryEndTime() {
+    return queryEndTime;
+  }
+
+  private QueryLeafProfilerBreakdown getQuerySliceProfilerBreakdown() {
```
jpountz: Let's replace `slice` by `thread` in this method name (and a few other places) to better reflect what it actually does?

jpountz (commented):
Sorry, I had lost track of this PR. I think that my only concern left is how it computes end times by adding the approximate timing to the start time. I'd rather not report it since this profiler doesn't actually compute an end time, but I'd be fine with reporting the sum of approximate timings if we think that it helps.

jainankitk:
> Sorry, I had lost track of this PR. I think that my only concern left is how it computes end times by adding the approximate timing to the start time. I'd rather not report it since this profiler doesn't actually compute an end time, but I'd be fine with reporting the sum of approximate timings if we think that it helps.

No worries @jpountz, thanks for the review. Even I was away for a bit, will try to address the comments and push new revision this week.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryLeafProfilerThreadAggregator.java
jainankitk: Sure, I have made the changes

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryLeafProfilerThreadAggregator.java
jainankitk: That makes sense. The profile object can hold the thread object reference, and while exposing this output via api, we can choose how to serialize the thread object

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryProfilerResult.java
jainankitk: Yeah, only `THREAD` is being used for now. I was planning to introduce other implementations like `LEAF` later, but I guess we can add this enum at that point.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryLeafProfilerBreakdown.java
jainankitk: Good catch! That was an oversight from my side. Will change this `toTotalTime()`, which should approximately reflect the consumed cpu time for processing this query

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryLeafProfilerBreakdown.java
jainankitk: Although minor, but `EnumMap` will initialize the timers for all the possible keys `vals = new Object[keyUniverse.length];`. It seemed bit unnecessary to me, but we can change if using `EnumMap` improves the readability.

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryLeafProfilerAggregator.java
jainankitk: I was planning to introduce other implementations later, but I guess we can add this interface at that point. Removing for now!

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/QueryProfilerTimingType.java
jainankitk: Yes, your understanding is correct. Added the javadocs!

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/AggregatedQueryLeafProfilerResult.java
jainankitk: Oh yeah, good point! Since the object is immutable

lucene/sandbox/src/java/org/apache/lucene/sandbox/search/AggregatedQueryLeafProfilerResult.java
jainankitk: I seem to have copied these getters from some existing code and missed editing the javadocs. Let me fix that!

jainankitk:
@jpountz - I have addressed all the comments from earlier review. Are you able to take another look, to help close this out?

jainankitk:
Thanks all for reviewing this PR. Planning to merge this PR by tomorrow, if there is no new feedback. Again, thanks for helping improve this change with your inputs!

dungba88:
Hi @jainankitk , is `QueryProfilerResult` no longer JSON (de)serializable? The reason is the introduction of `AggregatedQueryLeafProfilerResult` which contains `Thread` object. I don't see its Thread object is being used anywhere, can it be removed?

msokolov:
or if we do need to maintain the Thread association, can we store the Thread.id (an int I think) instead of the Thread?

jainankitk:
Thanks @dungba88 for trying this out.

> I don't see its Thread object is being used anywhere, can it be removed?

The `Thread` object is used for maintaining the association (as @msokolov called out) during aggregation of segment level Profiler Result. I was not exactly sure what all fields from `Thread` might be needed/make sense for serialization, so kept `Thread` itself in `AggregatedQueryLeafProfilerResult`. I am wondering if it makes sense to add `@JsonIgnore` / `transient` to not serialize `Thread` (and avoid exception/errors) and instead have limited view of `Thread`, say `SerializableThread` containing all the properties that we want to have in the serialization?

msokolov:
We don't want to be poking around inside `Thread` do we? can we store `Thread.getId()` (a long)?

jainankitk:
> We don't want to be poking around inside Thread do we? can we store Thread.getId() (a long)?

That should also be fine. Just adding some sort of `SerializableThread` gave more flexibility imo, if say, I am interested in `Thread.getName()` as well in addition to `Thread.getId()`?

PR #14416, opened 2025-03-27
skip keyword in German Normalization Filter

xzhang9292:
Current GermanNormalizationFilter tries to normalize special German characters like ä to a, ü to u. For some words it makes sense to do so, äpfel - > apfel is like apples -> apple. But for some words, it doesn't make sense, Bär -> Bar is like Bear -> Bar. Adding KeywordAttribute to allow users to bypass normalization on some specific words.

Update: Adding keyword attribute could help with German Analyzer which puts normalizer and stemmer after SetKeywordMarkerFilter. Putting "Bär" in exclusion set of German Analyzer can not stop normalizer to change it to "bar".

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/de/GermanNormalizationFilter.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/de/GermanNormalizationFilter.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import org.apache.lucene.analysis.TokenFilter;
 import org.apache.lucene.analysis.TokenStream;
 import org.apache.lucene.analysis.tokenattributes.CharTermAttribute;
+import org.apache.lucene.analysis.tokenattributes.KeywordAttribute;
 import org.apache.lucene.analysis.util.StemmerUtil;
 
 /**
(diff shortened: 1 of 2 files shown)


rmuir:
This keyword is legacy, for stemmers not normalizers. Just use ProtectedTermFilter which works with any tokenfilter without requiring modification to its code?

xzhang9292:
> This keyword is legacy, for stemmers not normalizers. Just use ProtectedTermFilter which works with any tokenfilter without requiring modification to its code?

@rmuir Thank you for your comment. Yes, using ProtectedTermFilter could be an option when directly using German Normalization Filter. Our problem comes when we are using German Analyzer. The GermanNormalizer positioned before GermanStemmers, so even we put word like "Bär" in exclusion set, normalizer can still change it to "bar". I think when people put a word in exclusion set, they probably also want the normalizer to keep the word. Would it be beneficial that we let Normalizer also exclude keywords?

rmuir:
i think it will introduce a ton more complexity: that's why we've pushed back on doing this for anything that isn't a stemmer. otherwise people will want LowerCaseFilter to respect it too... and then indexing performance is dead.

If you want to change the behavior of GermanAnalyzer then just make a CustomAnalyzer with ProtectedTermsFilter.

Issue #14362, opened 2025-03-17
Support modifying segmentInfos.counter in IndexWriter


guojialiang92:
### Description

Can we support modifying `segmentInfos.counter` in `IndexWriter`?
This can be used to skip some segment names when writing.
In the scenario of enabling `segment replication` in an Elasticsearch cluster, this control can be implemented during primary shard peer recovery and replica shard promotion to avoid segment counter conflicts.

Please help evaluate, if possible, I will submit a Pull Request.

vigyasharma:
Hi @guojialiang92, Could you elaborate more on how you plan to use this capability? It's not immediately obvious why modifying `segmentInfos.counter` will help with peer recovery or segment replication conflicts.

guojialiang92:
Hi @vigyasharma 

I will explain the usage scenarios.
In the **segment replication** scenario, the file name and content of the replica and the primary shard will be consistent. Usually, the replica shard uses **ReadOnlyEngine** and does not hold IndexWriter. The primary shard uses **ReadWriteEngine** and uses IndexWriter for segment building.

When there is **replica promotion** or **primary shard relocation**, we need to increase the segment count of the **new primary shard** to the specified value according to the algorithm (such as the raft). We need to avoid the new primary shard generating files with the same name but different content as the old primary shard.

In the current situation, we will not be able to increase the segment counter when the primary shard is created.

In the replica promotion scenario, a current alternative is to add `SegmentInfos#counter` when the replica shard is closed, and execute `SegmentInfos#commit` to persist the file to disk. When the new primary shard initializes **ReadWriteEngine**, it reads the updated `segment_N` file from disk.

However, this alternative solution has some drawbacks. For example, when the replica exits abnormally, it cannot guarantee the successful update of the segment counter. At the same time, in the primary shard relocation scenario, it is impossible to modify the segment counter in InterWriter (you can only use **ReadOnlyEngine** like replica promotion, and then switch to **ReadWriteEngine**, which will cause serious write blocking).

 Looking forward to your reply.

vigyasharma:
Thanks @guojialiang92 . Is the plan here to support creating an IndexWriter with a supplied value of `counter`, say `N`, so that all it's commit generations are `>=N` i.e. `segments_N, segments_N+1, ...` and so on ? 

[...]

PR #14417, opened 2025-03-27
Support modifying segmentInfos.counter in IndexWriter

guojialiang92:
### Description

This PR aims to address issue [14362](https://github.com/apache/lucene/issues/14362). This issue includes a discussion of the benefits of this modification.

### Tests

- simple scenario
  - I added test `TestIndexWriter#testAdvanceSegmentInfosCounter`, the writer index a bunch of docs, then advance its counter.
- crash and recovery scenario
  - I added test `TestIndexWriter#testAdvanceSegmentCounterInCrashAndRecoveryScenario`, the writer index a bunch of docs, then close it, start a new writer on the same index, and advance its counter.
- Concurrent writing and modification scenario
  - Modified the test `ThreadedIndexingAndSearchingTestCase`, in the concurrent write thread, increased the logic of randomly modifying the segment counter, and checked after all write threads ended.

### Checklist
- [x] I have reviewed the guidelines for [How to Contribute](https://github.com/apache/lucene/blob/main/CONTRIBUTING.md) and my code conforms to the standards described there to the best of my ability.
- [x] I have given Lucene maintainers [access](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) to contribute to my PR branch. (optional but recommended)
- [x] I have developed this patch against the main branch.
- [x] I have run ./gradlew check.
- [x] I have added tests for my changes.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -1437,10 +1437,11 @@ public class IndexWriter
     if (segmentInfos.counter < newCounter) {
       segmentInfos.counter = newCounter;
     }
+    changed();
   }
 
   /** Returns the {@link SegmentInfos#counter}. */
-  public synchronized long getSegmentInfosCounter() {
+  public long getSegmentInfosCounter() {
     this.ensureOpen();
     return segmentInfos.counter;
   }
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -1428,6 +1428,24 @@ public class IndexWriter
   }
 
   /**
+   * If {@link SegmentInfos#counter} is below {@code newCounter} then update it to this value.
+   *
+   * @lucene.internal
+   */
+  public synchronized void advanceSegmentInfosCounter(long newCounter) {
+    this.ensureOpen();
+    if (segmentInfos.counter < newCounter) {
+      segmentInfos.counter = newCounter;
+    }
+  }
+
+  /** Returns the {@link SegmentInfos#counter}. */
+  public synchronized long getSegmentInfosCounter() {
+    this.ensureOpen();
+    return segmentInfos.counter;
+  }
+
+  /**
    * Returns true if this index has deletions (including buffered deletions). Note that this will
    * return true if there are buffered Term/Query deletions, even if it turns out those buffered
    * deletions don't match any documents.
--- a/lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
@@ -5064,7 +5064,9 @@ public class TestIndexWriter extends LuceneTestCase {
       writer.commit();
     }
 
-    assertEquals(1041, writer.getSegmentInfosCounter());
+    // There may be merge operations in the background, here only verifies that the current segment
+    // counter is greater than 1000.
+    assertTrue(writer.getSegmentInfosCounter() > 1000);
 
     IndexWriter.DocStats docStats = writer.getDocStats();
     assertEquals(50, docStats.maxDoc);
--- a/lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
@@ -5040,10 +5040,8 @@ public class TestIndexWriter extends LuceneTestCase {
 
   public void testAdvanceSegmentInfosCounter() throws IOException {
     Directory dir = newDirectory();
-
     IndexWriter writer;
     IndexReader reader;
-
     writer = new IndexWriter(dir, newIndexWriterConfig(new MockAnalyzer(random())));
 
     // add 10 documents
@@ -5075,6 +5073,51 @@ public class TestIndexWriter extends LuceneTestCase {
     assertEquals(50, reader.maxDoc());
     assertEquals(50, reader.numDocs());
     reader.close();
+    dir.close();
+  }
+
+  public void testAdvanceSegmentCounterInCrashAndRecoveryScenario() throws IOException {
+    Directory dir = newDirectory();
+    IndexWriter writer;
+    IndexReader reader;
+    writer = new IndexWriter(dir, newIndexWriterConfig(new MockAnalyzer(random())));
+
+    // add 100 documents
+    for (int i = 0; i < 100; i++) {
+      addDocWithIndex(writer, i);
+      if (random().nextBoolean()) {
+        writer.commit();
+      }
+    }
+    IndexWriter.DocStats docStats = writer.getDocStats();
+    assertEquals(100, docStats.maxDoc);
+    assertEquals(100, docStats.numDocs);
+    writer.commit();
+    writer.close();
+
+    // recovery and advance segment counter
+    writer = new IndexWriter(dir, newIndexWriterConfig(new MockAnalyzer(random())));
+    assertEquals(100, writer.getDocStats().numDocs);
+    long newSegmentCounter = writer.getSegmentInfosCounter() + 1000;
+    writer.advanceSegmentInfosCounter(newSegmentCounter);
+
+    // add 10 documents
+    for (int i = 0; i < 10; i++) {
+      addDocWithIndex(writer, i);
+      if (random().nextBoolean()) {
+        writer.commit();
+      }
+    }
+
+    assertTrue(writer.getSegmentInfosCounter() >= newSegmentCounter);
+
+    assertEquals(110, writer.getDocStats().numDocs);
+    // check that the index reader gives the same numbers.
+    reader = DirectoryReader.open(dir);
+    assertEquals(110, reader.maxDoc());
+    assertEquals(110, reader.numDocs());
+    reader.close();
+    writer.close();
     dir.close();
   }
 }
--- a/lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
@@ -5037,4 +5037,45 @@ public class TestIndexWriter extends LuceneTestCase {
       }
     }
   }
+
+  public void testAdvanceSegmentInfosCounter() throws IOException {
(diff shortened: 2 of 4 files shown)


lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -5037,4 +5037,45 @@ public void testDocValuesSkippingIndexWithoutDocValues() throws Exception {
...
+    writer.advanceSegmentInfosCounter(1000);
+    assertEquals(1000, writer.getSegmentInfosCounter());
+
+    // add 40 documents
+    for (int i = 10; i < 50; i++) {
+      addDocWithIndex(writer, i);
+      writer.commit();
+    }
+
+    assertEquals(1041, writer.getSegmentInfosCounter());
```
vigyasharma: I think it's sufficient to assert that the counter value is `>=1000` i.e. higher than what you advanced it to. `SegmentInfos.counter` is public and gets incremented each time a segment is named, including background merges. Asserting on an exact value can make the test flaky.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -1427,6 +1427,24 @@ public synchronized void advanceSegmentInfosVersion(long newVersion) {
...
+   */
+  public synchronized void advanceSegmentInfosCounter(long newCounter) {
+    this.ensureOpen();
+    if (segmentInfos.counter < newCounter) {
+      segmentInfos.counter = newCounter;
+    }
+  }
+
+  /** Returns the {@link SegmentInfos#counter}. */
+  public synchronized long getSegmentInfosCounter() {
```
vigyasharma, line 1443: Do we really need this synchronized?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -1427,6 +1427,24 @@ public synchronized void advanceSegmentInfosVersion(long newVersion) {
     changed();
   }
 
+  /**
+   * If {@link SegmentInfos#counter} is below {@code newCounter} then update it to this value.
+   *
+   * @lucene.internal
+   */
+  public synchronized void advanceSegmentInfosCounter(long newCounter) {
```
vigyasharma, line 1435: Do we need to call `changed()` or increment change count here?

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -5037,4 +5037,47 @@ public void testDocValuesSkippingIndexWithoutDocValues() throws Exception {
...
+    writer = new IndexWriter(dir, newIndexWriterConfig(new MockAnalyzer(random())));
+
+    // add 10 documents
+    for (int i = 0; i < 10; i++) {
+      addDocWithIndex(writer, i);
+      writer.commit();
+    }
+    long beforeAdvanceSegmentCounter = writer.getSegmentInfosCounter();
+    writer.advanceSegmentInfosCounter(1);
+    assertEquals(beforeAdvanceSegmentCounter, writer.getSegmentInfosCounter());
```
vigyasharma, line 5056: Let's add a similar `gte` check here as in the assert below, instead of assuming that counter value is 1.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
```diff
@@ -5037,4 +5037,47 @@ public void testDocValuesSkippingIndexWithoutDocValues() throws Exception {
...
+
+    writer.advanceSegmentInfosCounter(1000);
+    assertEquals(1000, writer.getSegmentInfosCounter());
```
vigyasharma, line 5059: Do we need this check? Let's assert with `gte` to avoid flakiness.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
guojialiang92: You're right, I have already made modifications.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
guojialiang92: I have removed the synchronization to avoid affecting performance when it is heavily called in some scenarios.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
guojialiang92: From the `IndexWriter#changed` interface definition, it should indeed be called.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
guojialiang92: I have modified.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriter.java
guojialiang92: I have modified.

guojialiang92:
Thanks for helping with the code review, I have made modifications according to the suggestions

vigyasharma:
I think we can add a couple more tests to make it robust.
1. Some tests around concurrency –  index with multiple threads, then advance the counter in one of the threads, and validate behavior. You can look at `ThreadedIndexingAndSearchingTestCase` and its derived tests for motivation.
2. A test for the crash-recovery scenario, which I suppose it the primary use case. We could make the writer index a bunch of docs, then kill it, start a new writer on the same index, and advance its counter.

vigyasharma:
Also, IIUC `IndexWriter#advanceSegmentInfosVersion()` was added to handle similar scenarios for NRT replication (Lucene's native segment replication implementation). I'm curious why we didn't run into the need to advance `SegmentInfos#counter` at that time. Do you remember, @mikemccand (I know it's been a while! (: )?

guojialiang92:
Thanks, @vigyasharma 
> I think we can add a couple more tests to make it robust.
> 
> 1. Some tests around concurrency –  index with multiple threads, then advance the counter in one of the threads, and validate behavior. You can look at `ThreadedIndexingAndSearchingTestCase` and its derived tests for motivation.
> 2. A test for the crash-recovery scenario, which I suppose it the primary use case. We could make the writer index a bunch of docs, then kill it, start a new writer on the same index, and advance its counter.

I have added the following tests according to the suggestions: 
1. For crash-recovery scenario, I have added `TestIndexWriter#testAdvanceSegmentCounterInCrashAndRecoveryScenario`.
2. For scenarios with multi-threaded concurrent writes, I have modified the test `ThreadedIndexingAndSearchingTestCase`, in the concurrent write thread, increased the logic of randomly modifying the segment counter, and checked after all write threads ended.

guojialiang92:
Thanks, @vigyasharma 
I also looked at Lucene's native segment replication, just sharing my personal opinion.

> Also, IIUC `IndexWriter#advanceSegmentInfosVersion()` was added to handle similar scenarios for NRT replication (Lucene's native segment replication implementation). I'm curious why we didn't run into the need to advance `SegmentInfos#counter` at that time. Do you remember, @mikemccand (I know it's been a while! (: )?

In the code comments of Lucene's native segment replication, the risk of file conflicts is also mentioned, but no additional processing is done. From a robustness perspective, perhaps control should also be carried out. The relevant code is as follows：
ReplicaNode#fileIsIdentical (**Segment name was reused!  This is rare but possible and otherwise devastating**)
```
  private boolean fileIsIdentical(String fileName, FileMetaData srcMetaData) throws IOException {

    FileMetaData destMetaData = readLocalFileMetaData(fileName);
    if (destMetaData == null) {
      // Something went wrong in reading the file (it's corrupt, truncated, does not exist, etc.):
      return false;
    }

    if (Arrays.equals(destMetaData.header(), srcMetaData.header()) == false
        || Arrays.equals(destMetaData.footer(), srcMetaData.footer()) == false) {
      // Segment name was reused!  This is rare but possible and otherwise devastating:
      if (isVerboseFiles()) {
        message("file " + fileName + ": will copy [header/footer is different]");
      }
      return false;
    } else {
      return true;
    }
  }
```

guojialiang92:
Hi, @vigyasharma 
Could you please help me take a look again? Do you have any other suggestions?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -1427,6 +1427,25 @@ public synchronized void advanceSegmentInfosVersion(long newVersion) {
...
+    this.ensureOpen();
+    if (segmentInfos.counter < newCounter) {
+      segmentInfos.counter = newCounter;
+    }
+    changed();
+  }
+
+  /** Returns the {@link SegmentInfos#counter}. */
+  public long getSegmentInfosCounter() {
+    this.ensureOpen();
```
vigyasharma, line 1445: Do we care that the writer is open for this API?

vigyasharma (commented):
Apologies for the delay, these changes look great, @guojialiang92. I love the new tests, and how you're testing multi-threaded scenarios. 

I have some minor comments. Please also add a `CHANGES.txt` entry as well and I'll merge this.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
guojialiang92: Yeah, we currently cannot obtain `IndexWriter#segmentInfos` or `IndexWriter#segmentInfos.counter` from `IndexWriter` and Lucene users cannot perform `IndexWriter#segmentInfos.counter` related verification.

lucene/test-framework/src/java/org/apache/lucene/tests/index/ThreadedIndexingAndSearchingTestCase.java
```diff
@@ -189,6 +190,19 @@ public void run() {
                     addedField = null;
                   }
 
+                  // Maybe advance segment counter
+                  if (random().nextBoolean()) {
```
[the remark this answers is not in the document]
guojialiang92: Thanks, I have modified the test and CHANGES.txt.

guojialiang92:
Thanks, @vigyasharma 
I have made the modifications as suggested, please reivew the code again.

vigyasharma (approved):

PR #14418, opened 2025-03-28
quick exit on filter query matching no docs when rewriting knn query

bugmakerrrrrr:

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -64,7 +64,7 @@ Improvements
 
 Optimizations
 ---------------------
-(No changes)
+* GITHUB#14418: Quick exit on filter query matching no docs when rewriting knn query. (Pan Guixin)
 
 Bug Fixes
 ---------------------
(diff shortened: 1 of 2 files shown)


bugmakerrrrrr:
> Can you help me understand what work this change helps save?

Thanks for your reply. I think if the filter query matches no docs, the knn query must also match no docs. Before this small change, we always executed search tasks for each leaf reader even if the filter query is `MatchNoDocsQuery`.

https://github.com/apache/lucene/blob/a7ff87fdb5479a2f1e89746d88c80023b0aecbb1/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L89-L98

jpountz:
Thank you, that makes sense. So in the case when the filter rewrites to MatchNoDocsQuery, `IndexSearcher#search` would now return immediately, while it may currently need to wait for tasks to be submitted (if the queue is not empty).

Can you add a CHANGES entry?

bugmakerrrrrr:
> So in the case when the filter rewrites to MatchNoDocsQuery, `IndexSearcher#search` would now return immediately, while it may currently need to wait for tasks to be submitted (if the queue is not empty).

Yes, you are right.

> Can you add a CHANGES entry?

Added.

jpountz (approved):

PR #14421, opened 2025-03-28
Fix HistogramCollector to not create zero-count buckets.

jpountz:
If a bucket in the middle of the range doesn't match docs, it would be returned with a count of zero. Better not return it at all.

### DIFF
--- a/lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
+++ b/lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
@@ -279,7 +279,9 @@ final class HistogramCollector implements Collector {
     public void finish() throws IOException {
       // Put counts that we computed in the int[] back into the hash map.
       for (int i = 0; i < counts.length; ++i) {
-        collectorCounts.addTo(leafMinBucket + i, counts[i]);
+        if (counts[i] != 0) {
+          collectorCounts.addTo(leafMinBucket + i, counts[i]);
+        }
       }
       checkMaxBuckets(collectorCounts.size(), maxBuckets);
     }
--- a/lucene/sandbox/src/test/org/apache/lucene/sandbox/facet/plain/histograms/TestHistogramCollectorManager.java
+++ b/lucene/sandbox/src/test/org/apache/lucene/sandbox/facet/plain/histograms/TestHistogramCollectorManager.java
@@ -148,7 +148,7 @@ public class TestHistogramCollectorManager extends LuceneTestCase {
             .add(NumericDocValuesField.newSlowRangeQuery("f", Long.MIN_VALUE, 2), Occur.SHOULD)
             .add(NumericDocValuesField.newSlowRangeQuery("f", 10, Long.MAX_VALUE), Occur.SHOULD)
             .build();
-    actualCounts = searcher.search(query, new HistogramCollectorManager("f", 4));
+    actualCounts = searcher.search(query, new HistogramCollectorManager("f", 4, 3));
     expectedCounts = new LongIntHashMap();
     for (long value : values) {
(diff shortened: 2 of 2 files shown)


gsmiller (approved):
Good catch!

lucene/sandbox/src/test/org/apache/lucene/sandbox/facet/plain/histograms/TestHistogramCollectorManager.java
```diff
@@ -137,6 +140,23 @@ private void doTestSkipIndex(IndexWriterConfig cfg) throws IOException {
...
 
+    // Create a query so that bucket "1" (values from 4 to 8), which is in the middle of the range,
+    // doesn't match any docs. HistogramCollector should not add an entry with a count of 0 in this
+    // case.
+    Query query =
+        new BooleanQuery.Builder()
+            .add(NumericDocValuesField.newSlowRangeQuery("f", Long.MIN_VALUE, 2), Occur.SHOULD)
+            .add(NumericDocValuesField.newSlowRangeQuery("f", 10, Long.MAX_VALUE), Occur.SHOULD)
+            .build();
+    actualCounts = searcher.search(query, new HistogramCollectorManager("f", 4));
```
jainankitk, line 151: Initially, I was wondering if `checkMaxBuckets(collectorCounts.size(), maxBuckets)` might cause some of the tests earlier expecting exception to fail. But, all the tests with `counts[i] == 0`, use `1024` as `maxBuckets` which is comfortably over `collectorCounts.size()`

Maybe, we can specify 3 as `maxBuckets` for even stronger condition:

```
searcher.search(query, new HistogramCollectorManager("f", 4, 3));
```

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
```diff
@@ -279,7 +279,9 @@ public void collect(DocIdStream stream) throws IOException {
     public void finish() throws IOException {
       // Put counts that we computed in the int[] back into the hash map.
       for (int i = 0; i < counts.length; ++i) {
-        collectorCounts.addTo(leafMinBucket + i, counts[i]);
+        if (counts[i] != 0) {
+          collectorCounts.addTo(leafMinBucket + i, counts[i]);
+        }
       }
       checkMaxBuckets(collectorCounts.size(), maxBuckets);
```
jainankitk, line 286: While unrelated to this change, I am wondering if we should check eagerly to prevent unnecessary iterations:

```
if (counts[i] != 0) {
  collectorCounts.addTo(leafMinBucket + i, counts[i]);
  checkMaxBuckets(collectorCounts.size(), maxBuckets);
}
```

We are doing similar validation in other places:

```
if (bucket != prevBucket) {
  counts.addTo(bucket, 1);
  checkMaxBuckets(counts.size(), maxBuckets);
  prevBucket = bucket;
}
```

jainankitk (approved):

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
jpountz: I think it's ok the way it is. The end goal is to prevent unbounded heap allocation. In this case, the amount of excess heap we may allocate is bounded by 1024 entries, so I'd err on the side of simplicity by not checking the number of buckets in the loop?

jainankitk: Sounds fair. I was expecting low bound like 1024, just wanted to confirm!

jpountz:
I backported to branch_10_2 since this is a bugfix cc @iverase

PR #14426, opened 2025-04-01
Add support for determining off-heap memory requirements for KnnVectorsReader

ChrisHegarty:
This PR adds support to `KnnVectorsReader` in order to determine the off-heap memory requirements.

The motivation here is to give better insight into the size of off-heap memory that will be needed, so that deployments can be better scaled so that vector search workloads fit in memory, in order to provide best execution performance.


lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene90/Lucene90HnswVectorsReader.java
```diff
@@ -306,6 +306,16 @@ private HnswGraph getGraphValues(FieldEntry entry) throws IOException {
...
   }
 
+  @Override
+  public long offHeapBytes() {
+    long bytes = 0L;
+    for (var field : fields.values()) {
+      var f = field.value;
+      bytes += f.vectorDataLength() + f.indexDataLength();
+    }
+    return bytes;
```
tteofili, line 316: what about an `util` static method that can be applied here and for other `*VectorsReader` without having to re-implement the same code block multiple times?

tteofili (approved):
LGTM, thanks Chris, this sounds like a nice building block for assessing memory requirements.

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene90/Lucene90HnswVectorsReader.java
ChrisHegarty: That a good point, and one that I did think off too, but I wasn't able to come up with something simple as all these fields are their own individual inner classes, and any refactoring I experimented with just seemed to add complexity. :-(

lucene/core/src/java/org/apache/lucene/util/OffHeapAccountable.java
```diff
@@ -0,0 +1,44 @@
...
+
+package org.apache.lucene.util;
+
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * An object whose off-heap memory requirements can be computed.
+ *
+ * @lucene.internal
```
benwtrent, line 26: Is this really an "internal" API? Maybe I misunderstand its use, but it seems designed for folks to make decisions or provide feedback directly using this API.

So, maybe we call it "experimental" if unsure of its usage/stability right now?

lucene/core/src/java/org/apache/lucene/codecs/lucene102/Lucene102BinaryQuantizedVectorsReader.java
```diff
@@ -257,6 +257,15 @@ public long ramBytesUsed() {
...
 
+  @Override
+  public long offHeapBytes() {
+    long bytes = 0L;
+    for (var field : fields.values()) {
+      bytes += field.vectorDataLength();
+    }
+    return bytes;
+  }
+
```
benwtrent, line 268: Here is where we need to be careful with the API. Technically binary quantized vectors keeps the raw vectors (same with other quantized techniques).

Technically, the add to the overall off-heap size, though they may not need to be loaded into off-heap memory for search to be fast.

For example, if somebody is using `byte` type vectors, they aren't actually quantized, but its searchable. Meaning, they would all need to be in off-heap memory to be fast.

benwtrent (commented):
I like the direct API for gathering "codec off-heap file size". However, I am not sure we are doing that accounting correct in this PR. 

I do not like calling it "Off heap requirements". Really, its just the potential size if all the file must be loaded in off-heap.

lucene/core/src/java/org/apache/lucene/util/OffHeapAccountable.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public interface OffHeapAccountable {
```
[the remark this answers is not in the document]
ChrisHegarty: Yeah, "requirements" is not the best word here. It's just a "size".

lucene/core/src/java/org/apache/lucene/codecs/lucene102/Lucene102BinaryQuantizedVectorsReader.java
benwtrent: So, the user needs to know "when to sum them" and when "not to sum them" without any information as to what the child resources actually provide? 

I guess they can do an `instanceof` and determine what they want to do per codec name :(

seems brittle. I am not sure of a better way.

ChrisHegarty: I don't deny that this is not ideal, but it's not a new model, it already exists - `Accountable`.  One should always consider whether or not children should be counted, and if so how.  Lemme see if/how named  accountable could help here.

ChrisHegarty:
Given the feedback so far, I've pivot this quite a bit to now include per-field metrics. To support this I removed the previously proposed OffHeapAccountable interface and put the accessor on KnnVectorsReader - where other per-field accessors are. I used FieldInfo as the lookup since that it needed to determine if the field is either Byte of Float32, which is required to know in the "richer" codec implementations which provide quantization of floats and passthrough for bytes.

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsReader.java
```diff
@@ -130,4 +134,56 @@ public KnnVectorsReader getMergeInstance() {
...
+   *
+   * <pre>{@code
+   * getOffHeapByteSize(field).values().stream().mapToLong(Long::longValue).sum();
+   * }</pre>
+   *
+   * @param fieldInfo the fieldInfo
+   * @return a map of the desired off-heap memory requirements by category
+   * @lucene.experimental
+   */
+  public abstract Map<String, Long> getOffHeapByteSize(FieldInfo fieldInfo);
```
mayya-sharipova, line 172: Very nice API! +1
But I am also thinking about other use cases: where we need to know the total size across all vector fields and we don't know/don't remember field names. Do you think it is worth to have this API: may be that returns a map of maps?

ChrisHegarty: @mayya-sharipova the expected usage is that the caller would get all the fieldInfos from the reader ( `LeafReader::getFieldInfos` ), and then iterate over them checking for vector info. Which should be straightforward to do without an additional API point.

ChrisHegarty: While it's convenient to reuse Maps, an alternative here is to encapsulate the categorised stats in a container, which we could enrich later with information about whether the index is actually resident in memory. 
 
```
public abstract OffHeapStats getOffHeapByteSize(FieldInfo fieldInfo);

record OffHeapStats(String field, long raw, long hnswGraph, long quantized) { }
```

lucene/core/src/java/org/apache/lucene/codecs/lucene102/Lucene102BinaryQuantizedVectorsReader.java
```diff
@@ -257,6 +259,19 @@ public long ramBytesUsed() {
     return size;
   }
 
+  @Override
+  public Map<String, Long> getOffHeapByteSize(FieldInfo fieldInfo) {
+    Objects.requireNonNull(fieldInfo);
+    var raw = rawVectorsReader.getOffHeapByteSize(fieldInfo);
+    var fieldEntry = fields.get(fieldInfo.name);
+    if (fieldEntry == null) {
+      assert fieldInfo.getVectorEncoding() == VectorEncoding.BYTE;
```
jimczi, line 268: This is not possible, this format doesn't accept raw vector in the byte format.

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsReader.java
```diff
@@ -130,4 +134,56 @@ public KnnVectorsReader getMergeInstance() {
    * <p>The default implementation is empty
    */
   public void finishMerge() throws IOException {}
+
+  /** A string representing the off-heap category for quantized vectors. */
+  public static final String QUANTIZED = "QUANTIZED";
```
jimczi, line 139: nit: I wonder if we should rather reflect the underlying format here. Something like flat_vector_float, flat_vector_byte, flat_vector_bbq?

jimczi (commented):
LGTM on the new API, I left some nit comments on naming and assertions.

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsReader.java
benwtrent: >  I wonder if we should rather reflect the underlying format here. Something like flat_vector_float, flat_vector_byte, flat_vector_bbq?

IMO, this is then transforming to just the codec format name as I would expect different formats to have different characteristics, even between versions....

jimczi: Yep, then maybe using the file name extension is more consistent? That would be `VEC`, `VEX`, `VEQ` and `VEB`

ChrisHegarty: The file extension is quite cute, but discloses more of the implementation details to the caller which seems unnecessary.  

I deliberately chose to represent the categories as a `String` (rather than say an enum), since this offers more flexibility to the implementation while standardising on the three main categories.

ChrisHegarty: I chatted with Jim offline;  the file extension relates nicely with the how preLoading currently works, and makes that connection more concrete, which is a very nice property.  It also reduces the surface of the API change, allowing to remove the constant strings. I'll update the use the file extension.

ChrisHegarty: Done.

mayya-sharipova (approved):
Thanks Chris, organization by file extensions is very good

lucene/core/src/java/org/apache/lucene/codecs/KnnVectorsReader.java
```diff
@@ -130,4 +134,48 @@ public KnnVectorsReader getMergeInstance() {
...
+   *
+   * <pre>{@code
+   * getOffHeapByteSize(field).values().stream().mapToLong(Long::longValue).sum();
+   * }</pre>
+   *
+   * @param fieldInfo the fieldInfo
+   * @return a map of the desired off-heap memory requirements by category
+   * @lucene.experimental
+   */
+  public abstract Map<String, Long> getOffHeapByteSize(FieldInfo fieldInfo);
```
benwtrent, line 164: Doesn't this need to be backwards compatible? Here we are adding a new required base function to be overridden by all extenders of the `KnnVectorsReader`.

Marking it as experimental is good. But maybe it supplies a default implementation that simply returns `null` or `Map.of()`.

benwtrent (commented):
I like this much better. I do think this requires a CHANGES.txt as it modifies the KnnVectorReader object API. And possibly the API shouldn't be required for all vector readers due to BWC concerns?

navneet1v:
@ChrisHegarty this PR provides the info on how much off heap space is needed but this doesn't provide info on how much is loaded into memory correct? and do we have any plans to expose the current usage info?

PR #14428, opened 2025-04-02
Remove slices creation overhead from IndexSearcher constructor

javanna:
We simplified the slices creation in IndexSearcher with #13893. That removed the need for a caching supplier, but had an unexpected effect: we eagerly create slices now for the case where no executor is provided, and that may cause some overhead.

There are situations in which a searcher is created, although search methods that trigger the creation of slices are never called. We can in that case avoid going through that trouble.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
+++ b/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
@@ -220,7 +220,7 @@ public class IndexSearcher {
         : "IndexSearcher's ReaderContext must be topLevel for reader " + context.reader();
     reader = context.reader();
     this.taskExecutor =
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java
```diff
@@ -223,12 +223,6 @@ public IndexSearcher(IndexReaderContext context, Executor executor) {
         executor == null ? new TaskExecutor(Runnable::run) : new TaskExecutor(executor);
```
jainankitk, line 223: We need to use `DIRECT_TASK_EXECUTOR` here, for `taskExecutor == TaskExecutor.DIRECT_TASK_EXECUTOR` in `computeAndCacheSlices` to be true, right?

PR #14435, opened 2025-04-03
KeywordField.newSetQuery() to reuse prefixed terms in IndexOrDocValuesQuery

### LINKED ISSUE #14425: Reuse packedTerms between two TermInSetQuery which are combined by IndexOrDocValuesQuery
### Description

In cases like these 
```
new IndexOrDocValuesQuery(
           new TermInSetQuery(MultiTermQuery.CONSTANT_SCORE_BLENDED_REWRITE, name(), iBytesRefs), 
           new TermInSetQuery(MultiTermQuery.DOC_VALUES_REWRITE, name(), dVByteRefs));
```
I would like to reuse packedTerms between `TermInSetQuery` instances. What's the best approach to do it? Or it isn't worthwhile?

mkhludnev:
fix #14425

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/document/KeywordField.java
+++ b/lucene/core/src/java/org/apache/lucene/document/KeywordField.java
@@ -22,7 +22,6 @@ import org.apache.lucene.index.DocValuesType;
 import org.apache.lucene.index.IndexOptions;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.search.ConstantScoreQuery;
-import org.apache.lucene.search.IndexOrDocValuesQuery;
 import org.apache.lucene.search.MultiTermQuery;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.SortField;
@@ -175,9 +174,8 @@ public class KeywordField extends Field {
   public static Query newSetQuery(String field, Collection<BytesRef> values) {
     Objects.requireNonNull(field, "field must not be null");
     Objects.requireNonNull(values, "values must not be null");
-    Query indexQuery = new TermInSetQuery(field, values);
-    Query dvQuery = new TermInSetQuery(MultiTermQuery.DOC_VALUES_REWRITE, field, values);
-    return new IndexOrDocValuesQuery(indexQuery, dvQuery);
+    return TermInSetQuery.newIndexOrDocValuesQuery(
+        MultiTermQuery.CONSTANT_SCORE_BLENDED_REWRITE, field, values);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


lucene/core/src/java/org/apache/lucene/document/KeywordField.java
```diff
@@ -175,9 +174,8 @@ public static Query newExactQuery(String field, String value) {
   public static Query newSetQuery(String field, Collection<BytesRef> values) {
     Objects.requireNonNull(field, "field must not be null");
     Objects.requireNonNull(values, "values must not be null");
-    Query indexQuery = new TermInSetQuery(field, values);
-    Query dvQuery = new TermInSetQuery(MultiTermQuery.DOC_VALUES_REWRITE, field, values);
-    return new IndexOrDocValuesQuery(indexQuery, dvQuery);
+    return TermInSetQuery.newIndexOrDocValuesQuery(
+        MultiTermQuery.CONSTANT_SCORE_BLENDED_REWRITE, field, values);
```
jainankitk, line 178: Probably we can use `TermInSetQuery.newIndexOrDocValuesQuery(field, values)` here?

mkhludnev: Right. Just want to make it explicit.

jainankitk: Then we don't need that method, right?

jpountz (approved):

lucene/core/src/java/org/apache/lucene/document/KeywordField.java
jainankitk: We should remove the unused method `TermInSetQuery#newIndexOrDocValuesQuery(String field, Collection<BytesRef> terms)` in that case, as anyone invoking `TermInSetQuery#newIndexOrDocValuesQuery` should use the expert method `TermInSetQuery#newIndexOrDocValuesQuery(RewriteMethod indexRewriteMethod, String field, Collection<BytesRef> terms)`?

mkhludnev: ok. Got it. Since we add something, let's add as least as possible.

jpountz (approved):

lucene/core/src/java/org/apache/lucene/document/KeywordField.java
jainankitk: Thanks for making the change. I know its minor, but important for keeping it clean!

jainankitk (approved):

PR #14438, opened 2025-04-03
Let Decompressor implement the Closeable interface.

mulugetam:
### Description

In the current implementation, the [Compressor](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/compressing/Compressor.java) class implements the `Closeable` interface, but the [Decompressor](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/compressing/Decompressor.java) class doesn’t. When using Named SPI to plug in custom compressors—like the ones from [OpenSearch/custom-codecs](https://github.com/opensearch-project/custom-codecs)—it’d be helpful if `Decompressor` also implemented `Closeable` so we can properly release resources (e.g. natively allocated resources). This PR aims to fix that.

### DIFF
--- a/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/compressing/CompressionMode.java
+++ b/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/compressing/CompressionMode.java
@@ -144,6 +144,11 @@ public abstract class CompressionMode {
         public Decompressor clone() {
           return this;
         }
+
+        @Override
+        public void close() throws IOException {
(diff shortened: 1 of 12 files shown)


jpountz:
Unfortunately, you can't easily use close() to release resources from a Decompressor, because `StoredFieldsReader` is cloneable, and close() is never called on the clones. The only workaround that comes to mind would consist of using thread-locals, but I don't think we want to support that.

Issue #13335, opened 2024-05-01
Multi range traversal for numeric range aggregations


jainankitk:
### Description

In OpenSearch, we have introduced multi range traversal for collecting matching **document count** in single tree traversal (https://github.com/opensearch-project/OpenSearch/pull/13317). That has helped improve the performance of numeric aggregations in OpenSearch significantly. I am wondering if there are other use cases that can benefit from this and change should be included in Lucene.

Constraints:

1. The ranges are non-overlapping and in increasing order. For example - In (a1,b1),(a2,b2)...(an,bn), it is assumed ai<aj for all i<j and bi < a(i+1) for every i
2. Field is 1 dimensional guaranteeing the points are stored in increasing order, and segment does not have any deletions

jainankitk:
Looks related to #9814, but there are differences between the two. As discussed offline with @bowenlan and @rishabhmaurya, MultiRangeQuery only tells what matches with your multi ranges, not what matches with each of your multi ranges.

jainankitk:
@jpountz - Thoughts?

mikemccand:
I like this optimization.  Maybe it best fits in Lucene's facet module?  But, I don't think our facet impls today ever use points, directly, to do counting/aggregation -- it's a two step process of first collecting into a bitset holding the matched docs, and, second, iterating those docs and looking doc values or facet ords (also from doc values) and counting/aggregating from there.  But in the browse-only cases where a query just wants counts of ranges across all docs in the index, this opto should be a crazy fast way to achieve it when there are no deletions.  Even when there are deletions, this opto could visit all docs and check the live docs and count/aggregate accordingly?  The time is no longer sub-linear, but it'd still be faster than the two phased approach that Lucene's facets use today?

@stefanvodita / @Shradha26 WDYT?

stefanvodita:
It's true that we have this two-step process for aggregations (incl. counts) and that it's not always the most efficient solution.
+1 to try out this optimisation, sounds promising!

jainankitk:
I guess we can close this issue now - https://github.com/apache/lucene/blob/main/lucene/CHANGES.txt#L97

PR #14439, opened 2025-04-04
Logic for collecting Histogram efficiently using Point Trees

jainankitk:
### Description

This PR adds multi range traversal logic to collect the histogram on numeric field indexed as pointValues for MATCH_ALL cases. Even for non-match all cases like `PointRangeQuery`, if the query field == histogram field, this logic can be used. For the later, need to supply the `PointRangeQuery` bounds for building the appropriate `Ranges` to be collected. Need some inputs from the community on how it can be plugged correctly into the `HistogramCollector`

One of the key assumptions is absence of any deleted documents. Maybe going forward (especially if the deleted documents percentage is low), we can consider correcting the collected `Ranges` by subtracting for deleted documents. Although if I remember correctly, getting doc values for just deleted documents was non-trivial task!

Related issue https://github.com/apache/lucene/issues/13335


jainankitk:
@stefanvodita / @jpountz - Would love to get your thoughts on this optimization, and how we can leverage it in Lucene. In a nutshell, it solves the below problem:

Given a sorted non-overlapping set of intervals (Histogram buckets could be an example), it collects the matching documents count in single travel of `PointsTree` index, by skipping over the `leafBlocks` completely unless the values in `leafBlock` overlap with more than one interval. This ensures that the `# leafBlocks` actually traversed is bounded by the `# buckets` and remaining `leafBlocks` are collected in bulk. Hence it can very efficiently collect the doc counts, especially when the `# documents / # buckets` is pretty high.

stefanvodita (commented):
Sorry, I only had a quick look. Is the opimisation here analogous to the one `HistogramLeafCollector` does with the skipper?

jainankitk:
> Sorry, I only had a quick look. Is the opimisation here analogous to the one `HistogramLeafCollector` does with the skipper?

No, this approach is different from the skipper as it leverages `PointValues` instead of `DocValues` for computing the buckets

stefanvodita:
I didn't mean to imply that the two solutions are the same, apologies if that's how it came across.

> Need some inputs from the community on how it can be plugged correctly into the HistogramCollector

Let me know if this doesn't answer the question @jainankitk, maybe you'd already gone through this and you were looking for a different answer.
I think you could start in `HistogramCollector.getLeafCollector` ([code](https://github.com/apache/lucene/blob/4957766fcee52c534d786e3948fadf6d36c9779f/lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java#L50)). Right now we throw an exception if the field we're using isn't doc values ([code](https://github.com/apache/lucene/blob/4957766fcee52c534d786e3948fadf6d36c9779f/lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java#L59)). You'd need a new branch for the case you want to implement and a new `LeafCollector`, similar to the ones already in the file. Having that would make it easier to think through the next steps.

At a higher level, I'm curious if you had a use-case in mind.

jainankitk:
> I didn't mean to imply that the two solutions are the same, apologies if that's how it came across.

Not at all. Even I was initially confused with skipper logic, only after spending some time realized this approach is slightly different. So, thanks for reiterating the question.

> I think you could start in HistogramCollector.getLeafCollector ([code](https://github.com/apache/lucene/blob/4957766fcee52c534d786e3948fadf6d36c9779f/lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java#L50)). Right now we throw an exception if the field we're using isn't doc values ([code](https://github.com/apache/lucene/blob/4957766fcee52c534d786e3948fadf6d36c9779f/lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java#L59)).

Currently, `Collector` doesn't need to be aware of the `Query` itself. They are designed to collect individual docId or using `DocIdStream` from the scorer. But this `CustomCollector`, does not need the scorer to provide documents, but can `BulkCollect` documents, assuming `MATCH_ALL` or `PointRangeQuery` (where `PointRangeQuery.field == histogram.field`). Otherwise, it should fallback to traditional methods for collecting matching documents.


> At a higher level, I'm curious if you had a use-case in mind.

This optimization can be applied to following use cases:
* Number of sale based on the price range (0-50, 50-100, 100-250,.....)
* Number of visits on website for each day in a month

Just as a data point, this change helped us improve date histogram latency from 5168 ms to 160 ms (~32x!!) for big5 workload in OpenSearch

jainankitk:
I have updated the PR, and the code flow is like below now:

* `HistogramCollector` overrides the `setWeight` for accessing the underlying `Query`
* To keep things simple, just optimizing for `MATCH_ALL_DOCS` query and no deleted documents for now
* Optimized path is enabled only if `pointTree` is built for the field
* There are few other conditions for optimized path being enabled. Being conservative for now, and fallback to original.
* Add small unit test to verify working as expected

Will add few more unit tests, once I get some feedback on the code changes. Can also include small performance unit test that demonstrates, pointTree based collection is faster than the docValues based collector.

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
```diff
@@ -0,0 +1,219 @@
...
+        ArrayUtil.getValue(pointValues.getBytesPerDimension());
+    final long minValue = getLongFromByte(byteToLong, pointValues.getMinPackedValue());
+    final long maxValue = getLongFromByte(byteToLong, pointValues.getMaxPackedValue());
+    long leafMinBucket = Math.floorDiv(minValue, bucketWidth);
+    long leafMaxBucket = Math.floorDiv(maxValue, bucketWidth);
+
+    // We want that # leaf nodes is more than # buckets so that we can completely skip over
+    // some of the leaf nodes. Higher this ratio, more efficient it is than naive approach!
+    if ((pointValues.size() / 512) < (leafMaxBucket - leafMinBucket)) {
+      return false;
```
stefanvodita, line 54: Similar question as above. What will the user get in this case?

stefanvodita (commented):
Thanks for iterating on this @jainankitk!

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
```diff
@@ -0,0 +1,219 @@
...
+
+import java.io.IOException;
+import java.util.function.Function;
+import org.apache.lucene.index.PointValues;
+import org.apache.lucene.internal.hppc.LongIntHashMap;
+import org.apache.lucene.search.CollectionTerminatedException;
+import org.apache.lucene.search.DocIdSetIterator;
+import org.apache.lucene.util.ArrayUtil;
+
+class PointTreeBulkCollector {
```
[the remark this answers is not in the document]
jainankitk: Do we need explicit `@lucene.experimental` even for classes within sandbox? As per my understanding, everything sandbox is experimental by default?

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
jainankitk: The user does not need to get anything. We will just skip the optimized path and fall back to the original execution plan

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
```diff
@@ -0,0 +1,219 @@
...
+      final long bucketWidth,
+      final LongIntHashMap collectorCounts,
+      final int maxBuckets)
+      throws IOException {
+    // TODO: Do we really need pointValues.getDocCount() == pointValues.size()
+    if (pointValues == null
+        || pointValues.getNumDimensions() != 1
+        || pointValues.getDocCount() != pointValues.size()
+        || ArrayUtil.getValue(pointValues.getBytesPerDimension()) == null) {
+      return false;
```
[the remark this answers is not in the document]
jainankitk: The user does not need to get anything. We will just skip the optimized path and fall back to the original execution plan

lucene/sandbox/src/test/org/apache/lucene/sandbox/facet/plain/histograms/TestHistogramCollectorManager.java
```diff
@@ -115,6 +117,38 @@ public void testSkipIndexWithSortAndLowInterval() throws IOException {
             .setCodec(TestUtil.alwaysDocValuesFormat(new Lucene90DocValuesFormat(3))));
   }
 
+  public void testMultiRangePointTreeCollector() throws IOException {
```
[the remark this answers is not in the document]
jainankitk: Okay, ran the test and few times and I understand what the error is.

```
java.lang.AssertionError: expected:<[1=>978, 3=>1004, 4=>1000, 2=>1022, 0=>996]> but was:<[4=>5000, 1=>4890, 2=>5110, 3=>5020, 0=>4980]>
```

The actual values are 5 times of the expected. Essentially, the logic efficiently collects values for entire segment and in case of intra-segment concurrency, it ends up collecting 5 times. I am wondering if the collectors can be aware of intra-segment concurrency, and not use the optimized path when it is enabled?

lucene/core/src/java/org/apache/lucene/util/ArrayUtil.java
```diff
@@ -801,4 +802,16 @@ public static int compareUnsigned4(byte[] a, int aOffset, byte[] b, int bOffset)
     return Integer.compareUnsigned(
         (int) BitUtil.VH_BE_INT.get(a, aOffset), (int) BitUtil.VH_BE_INT.get(b, bOffset));
   }
+
+  public static Function<byte[], Long> getValue(int numBytes) {
```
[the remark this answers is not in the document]
jainankitk: Thanks for suggesting `NumericUtils`. I wasn't aware of the `NumericUtils#sortableBytesToLong`, exactly what I needed to flip the sign bit correctly!

```
// Flip the sign bit back
    v ^= 0x8000000000000000L;
```

I have removed this method from `ArrayUtil` and added in `PointTreeBulkCollector`. Please let me know if you still see any concerns

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
```diff
@@ -53,11 +59,22 @@ public LeafCollector getLeafCollector(LeafReaderContext context) throws IOExcept
       // The segment has no values, nothing to do.
       throw new CollectionTerminatedException();
     }
+
+    // We can use multi range traversal logic to collect the histogram on numeric
+    // field indexed as point range query for MATCH_ALL cases
+    final PointValues pointValues = context.reader().getPointValues(field);
+    if (query instanceof MatchAllDocsQuery && !context.reader().hasDeletions()) {
```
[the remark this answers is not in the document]
jainankitk: Yes, I will open follow-up once we are close to merging this PR. I will keep this conversation open, until the issue is created

lucene/sandbox/src/test/org/apache/lucene/sandbox/facet/plain/histograms/TestHistogramCollectorManager.java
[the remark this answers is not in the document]
jainankitk: Okay, I have added logic for collecting only for one of the partitions if intra-segment concurrency is enabled. Ran the test 50k times after that without any failures:

```
> Task :lucene:core:testClasses
> Task :lucene:core:test
> Task :lucene:core:wipeTaskTemp

> Task :lucene:sandbox:test
:lucene:sandbox:test (SUCCESS): 50000 test(s)

> Task :lucene:sandbox:wipeTaskTemp
The slowest suites (exceeding 1s) during this run:
  580.41s TestHistogramCollectorManager (:lucene:sandbox)

BUILD SUCCESSFUL in 9m 58s
246 actionable tasks: 181 executed, 65 up-to-date
```

jainankitk:
Addressed most of the review comments. Ran small performance benchmark to see the difference, and we can see some even with small number of documents:

Bulk time is new approach, and skip index time is the existing
With 100k docs, and 5k bucketWidth
```
  1> Bulk time was : 10593417 vs skip index time : 12941417
  1> Bulk time was : 1006750 vs skip index time : 3241959
  1> Bulk time was : 770916 vs skip index time : 4690875
  1> Bulk time was : 870917 vs skip index time : 2621291
  1> Bulk time was : 754709 vs skip index time : 3175500
  1> Bulk time was : 1389750 vs skip index time : 2557125
  1> Bulk time was : 788125 vs skip index time : 4633541
  1> Bulk time was : 767125 vs skip index time : 5078917
  1> Bulk time was : 727208 vs skip index time : 6333083
  1> Bulk time was : 755250 vs skip index time : 3179917
```

With 250k documents, and 10k bucketWidth:
```
  1> Bulk time was : 14125250 vs skip index time : 20235666
  1> Bulk time was : 915583 vs skip index time : 15929834
  1> Bulk time was : 886959 vs skip index time : 15842375
  1> Bulk time was : 904667 vs skip index time : 10911500
  1> Bulk time was : 801959 vs skip index time : 10846750
  1> Bulk time was : 844834 vs skip index time : 14980917
  1> Bulk time was : 855209 vs skip index time : 5865500
  1> Bulk time was : 804334 vs skip index time : 15966667
  1> Bulk time was : 805042 vs skip index time : 11613791
  1> Bulk time was : 830958 vs skip index time : 11036833
```

As we can see the difference is bit more pronounced with increasing number of documents and wide enough bucketWidth.

jainankitk:
@stefanvodita - Thanks for a prompt review. Addressed most of the review comments. Adding JMH benchmark instead of the not so useful performance test added earlier. The benchark results demonstrate significant increase in throughput with increasing # documents and bucket width (lesser buckets mean less low level traversal in point range tree and more documents collected in bulk):

```
Benchmark                                     (bucketWidth)  (docCount)  (pointEnabled)   Mode  Cnt     Score      Error  Units
HistogramCollectorBenchmark.collectHistogram           5000      500000            true  thrpt    3  2114.002 ±   46.852  ops/s
HistogramCollectorBenchmark.collectHistogram           5000      500000           false  thrpt    3  1236.532 ±  724.312  ops/s
HistogramCollectorBenchmark.collectHistogram           5000     5000000            true  thrpt    3   220.636 ±   17.688  ops/s
HistogramCollectorBenchmark.collectHistogram           5000     5000000           false  thrpt    3   147.054 ±   37.055  ops/s
HistogramCollectorBenchmark.collectHistogram          25000      500000            true  thrpt    3  9788.181 ± 4388.445  ops/s
HistogramCollectorBenchmark.collectHistogram          25000      500000           false  thrpt    3  1227.537 ±   21.222  ops/s
HistogramCollectorBenchmark.collectHistogram          25000     5000000            true  thrpt    3   856.278 ±  840.625  ops/s
HistogramCollectorBenchmark.collectHistogram          25000     5000000           false  thrpt    3   145.322 ±   86.603  ops/s
```

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
```diff
@@ -53,11 +65,30 @@ public LeafCollector getLeafCollector(LeafReaderContext context) throws IOExcept
       // The segment has no values, nothing to do.
       throw new CollectionTerminatedException();
     }
+
+    // We can use multi range traversal logic to collect the histogram on numeric
+    // field indexed as point for MATCH_ALL cases. In future, this can be extended
+    // for Point Range Query cases as well
+    if (query instanceof MatchAllDocsQuery && context.reader().hasDeletions() == false) {
```
jpountz, line 72: Maybe replace this condition with `weight.count(context) == context.reader().maxDoc()` to make it a bit more generic, e.g. it could be a range query that happens to fully match the segment.

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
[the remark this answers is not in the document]
stefanvodita: There are other files in sandbox with experimental tags, so maybe better safe then sorry? I guess it's easy to forget to add those when moving a piece of code out of sandbox.

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
```diff
@@ -0,0 +1,227 @@
...
+      // Used by IntPoint, FloatPoint, LatLonPoint, LatLonShape
+      return a -> (long) NumericUtils.sortableBytesToInt(a, 0);
+    }
+
+    return null;
+  }
+
+  static boolean canCollectEfficiently(final PointValues pointValues, final long bucketWidth)
+      throws IOException {
+    // TODO: Do we really need pointValues.getDocCount() == pointValues.size() for this optimization
```
jpountz, line 43: I think so, the collector makes sure to count each doc only once, including docs that have two values that fall into the same bucket.

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    long leafMinBucket =
+        Math.floorDiv(byteToLong.apply(pointValues.getMinPackedValue()), bucketWidth);
+    long leafMaxBucket =
+        Math.floorDiv(byteToLong.apply(pointValues.getMaxPackedValue()), bucketWidth);
+
+    // We want that # leaf nodes is more than # buckets so that we can completely skip over
+    // some of the leaf nodes. Higher this ratio, more efficient it is than naive approach!
+    if ((pointValues.size() / 512) < (leafMaxBucket - leafMinBucket)) {
```
jpountz, line 62: I imagine that 512 is the size of leaves with the default codec. It's a bit annoying to rely on a codec parameter. I'm tempted to replace it with a simpler heuristic like the collector does, something like `leafMaxBucket - leafMinBucket <= 1024`.

jpountz (commented):
Interesting idea! I like that you integrated it transparently into the collector, so that users can benefit from it out of the box.

lucene/core/src/java/org/apache/lucene/util/ArrayUtil.java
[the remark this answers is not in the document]
stefanvodita: I'm happy with how it is now. Thank you!

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
jainankitk: Ah, thats a good point! I was thinking of cases, when two values are across two buckets, but two values falling in same bucket makes the condition obvious

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
jainankitk: Yes we can, but I am anyway planning to address `PointRangeQuery` case more generically as fast followup - https://github.com/apache/lucene/issues/14535. Also with intra-segment search, `leafCollector` can be called multiple times, so I want to ensure that in cases we don't use this optimization has minimal overhead. I am assuming `weight.count(context)` will have some overhead?

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
jainankitk: > I imagine that 512 is the size of leaves with the default codec. It's a bit annoying to rely on a codec parameter. I'm tempted to replace it with a simpler heuristic like the collector does, something like leafMaxBucket - leafMinBucket <= 1024.

While I share your annoyance of relying on codec parameter, `leafMaxBucket - leafMinBucket <= 1024` works for prior case as it uses dense array that is just function of this. But here, if we collect billion documents,  even ` leafMaxBucket - leafMinBucket  <= 10^5` is fairly reasonable. Maybe we can use something more generic and reasonable like 1000 instead of sticking exactly to the codec parameter?

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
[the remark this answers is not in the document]
stefanvodita: Let me be more specific. I'm questioning if the `Expected numeric field, but got doc-value type` exception we used to throw still makes sense now or needs to be adjusted since we now have another case for `PointValues` which must have not worked out to reach the exception.

jainankitk:
> Interesting idea! I like that you integrated it transparently into the collector, so that users can benefit from it out of the box.

Thanks @jpountz for the review. Had some challenges with the integration initially, but `seWeight` method was pretty useful!

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
stefanvodita: I could be misunderstanding, so please bear with me. Let's say a user only has `PointValues` indexed and no `DocValues` and they try to use the histogram. Would they be able to do that even if their data doesn't fit the efficiency condition above?

stefanvodita (commented):
Nice results in the benchmark!

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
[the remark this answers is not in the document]
jainankitk: > There are other files in sandbox with experimental tags, so maybe better safe then sorry?

Yup, adding the tag!

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
```diff
@@ -0,0 +1,219 @@
...
+    private void count() {
+      counter++;
+    }
+
+    private void countNode(int count) {
+      counter += count;
+    }
+
+    private void finalizePreviousBucket(byte[] packedValue) {
+      // TODO: Can counter ever be 0?
```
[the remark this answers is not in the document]
jainankitk: I think this specific TODO, we can leave for now. More like a safety guard. I will try to add unit test in future to validate this, and remove if not needed

Others TODOs were important and have been addressed

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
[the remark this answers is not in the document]
jainankitk: I think we can keep this for now as is. The way to read it is, it works same as before including exceptions and everything. There is possibly optimized path if pointValues are indexed. I have created another issue for tackling this specific scenario - https://github.com/apache/lucene/issues/14536. Might need some more changes or slightly different approach to `PointTreeBulkCollector`

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
jainankitk: Ah, not really. As explained in other, we can discuss that as part of https://github.com/apache/lucene/issues/14536?

jainankitk:
> Oooh those JMH benchy results are nice! Though, it's dangerous testing only on random data -- you can draw random conclusions/results. But it's better than no benchmark! Maybe we should add histogram faceting benchy to Lucene's nightly benchmarks?

Thanks @mikemccand for the feedback. Having histogram faceting benchmark in Lucene's nightly will be great! Created issue https://github.com/mikemccand/luceneutil/issues/375 for following up on this.

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
jainankitk: Oh yeah, that's a good point! Hadn't thought about queries on other fields. Changed

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
```diff
@@ -53,11 +63,31 @@ public LeafCollector getLeafCollector(LeafReaderContext context) throws IOExcept
       // The segment has no values, nothing to do.
       throw new CollectionTerminatedException();
     }
+
+    // We can use multi range traversal logic to collect the histogram on numeric
+    // field indexed as point for MATCH_ALL cases. In future, this can be extended
+    // for Point Range Query cases as well
+    if (context.reader().hasDeletions() == false
```
jpountz, line 70: You don't need this check, Weight#count already accounts for deletions.

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
```diff
@@ -53,11 +63,31 @@ public LeafCollector getLeafCollector(LeafReaderContext context) throws IOExcept
...
+        && weight.count(context) == context.reader().maxDoc()) {
```
jpountz, line 71: I would check weight against `null`, applications may have collectors that don't set a weight.

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
jpountz: Sounds good to me.

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
jainankitk: Ah, nice!

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
jainankitk: Added `null` check

jpountz (approved):

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/PointTreeBulkCollector.java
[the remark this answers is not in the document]
stefanvodita: I see your point and I'm happy to have a follow-up issue if necessary for this and the related comment below.

stefanvodita (approved):
Thank you @jainankitk!

lucene/CHANGES.txt
```diff
@@ -78,6 +78,8 @@ Optimizations
 ---------------------
 * GITHUB#14418: Quick exit on filter query matching no docs when rewriting knn query. (Pan Guixin)
 
+* GITHUB#14439: Efficient Histogram Collection using multi range traversal over PointTrees
```
stefanvodita, line 81: You should add your name too!

jainankitk: ah, my bad. Fixed!

PR #14447, opened 2025-04-07
Compute the doc range more efficiently when flushing doc block

bugmakerrrrrr:
### Description

Today, we sum all the doc delta to compute the range when flushing doc block, which is equivalent to `lastDocID - level0LastDocID`.


jainankitk (approved):

stefanvodita (commented):
This might still apply to `Lucene103PostingsWriter`. Should we rebase?

jainankitk:
Thanks for updating the PR to include `Lucene103PostingsWriter`. LGTM!

stefanvodita (changes requested):
Thank you for updating the 10.3 postings writer! I'm happy to merge that change, but I would rather not mix in the 10.1 change since that release has already gone out and we're unlikely to release a follow-up unless we find serious bugs.

Could you also add an entry under 10.3 in CHANGES.txt?

stefanvodita:
Sorry, I don't know if I was clear enough. I was asking if we can also revert the change to `Lucene101PostingsWriter` and only change `Lucene103PostingsWriter` this time.

Thanks for adding the changelog entry!

bugmakerrrrrr:
> Sorry, I don't know if I was clear enough. I was asking if we can also revert the change to `Lucene101PostingsWriter` and only change `Lucene103PostingsWriter` this time.
> 
> Thanks for adding the changelog entry!

My fault, I forgot to revert.

stefanvodita (approved):
Thank you @bugmakerrrrrr!

PR #14448, opened 2025-04-07
Tests: TestLongNormValueSource, AssertingScorable

dsmiley:
TestLongNormValueSource: test specifics
AssertingScorable: don't wrap needlessly

### DIFF
--- a/lucene/queries/src/test/org/apache/lucene/queries/function/TestLongNormValueSource.java
+++ b/lucene/queries/src/test/org/apache/lucene/queries/function/TestLongNormValueSource.java
@@ -16,7 +16,6 @@
  */
 package org.apache.lucene.queries.function;
 
-import java.util.function.Function;
 import org.apache.lucene.document.Document;
 import org.apache.lucene.document.Field.Store;
 import org.apache.lucene.document.StringField;
@@ -65,8 +64,7 @@ public class TestLongNormValueSource extends LuceneTestCase {
         searcher = newSearcher(reader);
         searcher.setSimilarity(sim);
 
-        Query q =
-            new FunctionQuery(((Function<String, ValueSource>) NormValueSource::new).apply(field));
+        Query q = new FunctionQuery(new NormValueSource(field));
         TopDocs docs = searcher.search(q, searcher.getIndexReader().numDocs(), new Sort(), true);
         final var expectedDocIds = new int[] {1, 0};
         final var expectedScoreDocs =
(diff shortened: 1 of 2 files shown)


lucene/test-framework/src/java/org/apache/lucene/tests/search/AssertingScorable.java
```diff
@@ -44,7 +44,7 @@ public void setMinCompetitiveScore(float minScore) throws IOException {
   }
 
   public static Scorable wrap(Scorable in) {
-    if (in instanceof AssertingScorable) {
+    if (in instanceof WrappedScorer || in instanceof AssertingScorable) {
```
jpountz, line 47: I don't understand this part of the change, this method expects the returned `Scorable` to have assertions enabled, but it may not be the case if this is an instance of `WrappedScorer`?

lucene/queries/src/test/org/apache/lucene/queries/function/TestLongNormValueSource.java
```diff
@@ -16,99 +16,70 @@
...
+        searcher.setSimilarity(sim);
 
-  public void testNorm() throws Exception {
-    Similarity saved = searcher.getSimilarity();
-    try {
-      // no norm field (so agnostic to indexed similarity)
-      searcher.setSimilarity(sim);
-      assertHits(new FunctionQuery(new NormValueSource("text")), new float[] {0f, 0f});
+        Query q =
+            new FunctionQuery(((Function<String, ValueSource>) NormValueSource::new).apply(field));
```
jpountz, line 69: Why are you calling the constructor via a method reference rather than directly?

lucene/test-framework/src/java/org/apache/lucene/tests/search/AssertingScorable.java
jpountz: Thanks for explaining, there can be much wrapping indeed. I had missed that `AssertingScorable` would sometimes do this, I opened a PR that hopefully cleans this up a bit. #14452

jpountz (approved):
LGTM

lucene/test-framework/src/java/org/apache/lucene/tests/search/AssertingScorable.java
dsmiley: Nice :-)

PR #14452, opened 2025-04-07
Clean up how the test framework creates asserting scorables.

jpountz:
This removes the `WrappedScorer` class and uses the existing `AssertingScorer` class instead.


lucene/test-framework/src/java/org/apache/lucene/tests/search/AssertingScorer.java
```diff
@@ -80,8 +74,8 @@ boolean iterating() {
 
   @Override
   public void setMinCompetitiveScore(float score) throws IOException {
-    assert scoreMode == ScoreMode.TOP_SCORES;
-    assert canCallMinCompetitiveScore;
+    assert canScore;
+    assert canSetMinCompetitiveScore;
```
jainankitk, line 78: Still trying to understand this code better, but it seems that the assertion statement is weaker now. Earlier, it was specifically `scoreMode == ScoreMode.TOP_SCORES`, and now it is `scoreMode.needsScores()`. Do we allow it for other `scoreMode` that `needsScores` now?

jainankitk (commented):
The refactoring looks good, even cleans up some unused code like `AssertingScorer#random`. Minor comment to understand it better

jainankitk (approved):

PR #14453, opened 2025-04-07
Reduce polymorphism of DocIdSetIterator#docID().

jpountz:
This introduces a new `AbstractDocIdSetIterator` abstract class that tracks the current doc IDs and refactors several `DocIdSetIterator` implementations to extend either this `AbstractDocIdSetIterator` or `FilterDocIdSetIterator`, in order to reduce polymorphism of call sites of `DocIdSetIterator#docID()`.


stefanvodita (approved):
Cool to see how all these implementations had evolved and how much we can simplify the code!

PR #14457, opened 2025-04-09
tests: set a socket timeout on replicator tests

rmuir:
These nightly tests sometimes hang and are killed by timeout, instead of passing or failing. Upon inspection of one of the failures, the hung thread is stuck in `SocketDispatcher.read0()` native code, presumably waiting for bytes that will never appear.

Instead, set a socket timeout for these tests, so that infinite hangs (of this sort) will no longer happen: they may become failures instead. Failures are better than hangs for many reasons.

Relates: #14454


uschindler:
I sit enough to only change that part so it affects all NRT tests?

uschindler (approved):

rmuir:
> Is it enough to only change that part so it affects all NRT tests?

Yes, this is the only place creating Sockets there.
Also, I didn't touch any ServerSocket as that's different and not causing hangs.

rmuir:
Yeah, we can look into it separately. It is different and is timeout for `accept()` calls only IIRC. I haven't looked at the way these tests manage connections, but it is not implicated in any test hangs yet.

Issue #14444, opened 2025-04-05
Add an IndexDeletion policy that retains the last N commits


vigyasharma:
In segment based replication systems, a large replication payload (checkpoint) can induce heavy page faults, cause thrashing for in-flight search requests, and affect overall search performance. We can address this problem by leveraging multiple commit points in the replication payload for the Lucene index. Searcher Managers can then incrementally refresh on the commit points they can safely absorb, instead of always directly on to the latest one.

To support this mechanism, we need to modify the index to retain multiple commit points. It would be convenient to have Lucene provide an `IndexDeletionPolicy` that accepts `N` and always retains the last N commits. Index deletion policy accepts a list of commits that is sorted by commit generation (oldest to latest). The proposed policy, `KeepLastNCommitsDeletionPolicy` can delete all commits except the last N commits in the list.

This could be a good first issue for beginners. Helps #14219

owaiskazi19:
@vigyasharma opened https://github.com/apache/lucene/pull/14458 for the issue.

owaiskazi19:
Can we close this @vigyasharma ?

PR #14458, opened 2025-04-10
Add an IndexDeletion policy that retains the last N commits

owaiskazi19:
### Description
This change introduces a new IndexDeletion policy called as `KeepLastNCommitsDeletionPolicy` that retains the last N commits and removes all prior commits after a new commit is done.

Issue

Relates to https://github.com/apache/lucene/issues/14444

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/KeepLastNCommitsDeletionPolicy.java
+++ b/lucene/core/src/java/org/apache/lucene/index/KeepLastNCommitsDeletionPolicy.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


lucene/core/src/java/org/apache/lucene/index/KeepLastNCommitsDeletionPolicy.java
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Constructor that sets the number of commits to keep.
+   *
+   * @param numCommitsToKeep the number of most recent commits to retain
+   * @throws IllegalArgumentException if numCommitsToKeep is not positive
+   */
+  public KeepLastNCommitsDeletionPolicy(int numCommitsToKeep) {
+    if (numCommitsToKeep <= 0) {
+      throw new IllegalArgumentException("number of recent commits to keep must be positive");
+    }
```
vigyasharma, line 40: Thanks! It is critical that we don't accept `0 numCommitsToKeep`, otherwise the policy will simply delete all commits i.e. the entire index! Let's also add a test to ensure people get warned if they inadvertently remove this check.

vigyasharma (commented):
Changes look great! Please add an entry to 10.3 in CHANGES.txt and I'll merge this.

lucene/core/src/java/org/apache/lucene/index/KeepLastNCommitsDeletionPolicy.java
owaiskazi19: Sounds good. Added a test for the same

vigyasharma (approved):

PR #14459, opened 2025-04-10
Move some DocValuesConsumer merge logic to helper methods.

martijnvg:
Move code that creates merged doc value instances and ordinal map to protected helper methods. This will allow custom DocValuesConsumer implementations to reuse these helper methods while providing custom merging logic.


stefanvodita (approved):
This looks reasonable to me. Curious if you can share more about what you're trying to achieve! We should have a CHANGES entry as well.

martijnvg:
Thanks for taking a look @stefanvodita! I've added the CHANGES entry.

> Curious if you can share more about what you're trying to achieve!

We have [a specialized doc values codec](https://github.com/elastic/elasticsearch/tree/main/server/src/main/java/org/elasticsearch/index/codec/tsdb/es819) for append only use cases that relies on index sorting. In order to speed up merging, we make use of the metadata statistics of the doc value producer in order to avoid pulling doc values instances from the producer and iterating over it multiple times. When indexing sorting is active, then this is more expensive. The consumer of this doc value codec creates a custom producer instance in the `merge*(...)` methods, which the `add*Field(...)` methods use to determine whether metadata statistics need to be computed or not. However the logic that creates the merged doc value instances and ordinal map are identical in the `DocValuesConsumer` class. This PR will allow us the reuse that logic and I think anyone creating a custom doc value codec with custom merge logic can benefit from this as well.

stefanvodita:
Thanks for the explanation! The new revision looks good as well.

dnhatn (approved):
Looks great. Thanks Martijn!

martijnvg:
Thanks for the reviews @dnhatn and @stefanvodita.

PR #14462, opened 2025-04-10
Configure dependabot for python deps, fix linter issues

rmuir:
These dependencies are not pinned which can cause lack of reproducibility. Pin the direct dependencies and configure the dependabot. Clear up lint issues with most recent versions of checkers.

Hopefully, this should help keep the scripts stable and working.


stefanvodita (approved):

rmuir:
I will tune it and lower the frequency after we see it working, at least the type checker releases quite frequently. So I expect a lot of noise at first, but just want to get the kinks ironed out.

PR #14476, opened 2025-04-11
Removing unnecessary splitX in ComponentTree

jainankitk:
### Description
IMO, storing splitX for every `TreeNode` makes sense, only if we are exposing API to partially search `ComponentTree` from arbitrary `TreeNode` which is not allowed today. Hence, makes sense to address the TODO and remove `splitX`.

Code without `splitX` looks reasonably straightforward to me, just follow the same structure as `createTree`. Start with `ROOT_SPLITX` and keep toggling `splitX` as we recurse through the levels

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/geo/ComponentTree.java
+++ b/lucene/core/src/java/org/apache/lucene/geo/ComponentTree.java
@@ -27,11 +27,13 @@ import org.apache.lucene.util.ArrayUtil;
  */
 final class ComponentTree implements Component2D {
 
-  private static final Comparator<Component2D> XComparator =
+  private static final Comparator<Component2D> X_COMPARATOR =
       Comparator.comparingDouble(Component2D::getMinX).thenComparingDouble(Component2D::getMaxX);
 
-  private static final Comparator<Component2D> YComparator =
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/geo/ComponentTree.java
```diff
@@ -48,23 +50,18 @@ final class ComponentTree implements Component2D {
...
   // a consistent bounding box. Internal nodes should not be accessed
   // outside if this class.
-  private Component2D left;
-  private Component2D right;
-
-  /** which dimension was this node split on */
-  // TODO: its implicit based on level, but boolean keeps code simple
-  private final boolean splitX;
+  private ComponentTree left;
+  private ComponentTree right;
```
jainankitk, line 54: left, right should always be `ComponentTree` to indicate the recursive structure. While `Component2D` worked as `ComponentTree` implements the interface, but plugging in arbitrary implementation of `Component2D` here won't work.

rmuir (approved):
Seems fine! I remember storing it this way out of laziness, but it should really be implicit.

rmuir:
Thank you @jainankitk

PR #14492, opened 2025-04-15
ci: pin github actions versions in use

### LINKED ISSUE #14491: better pin github actions versions
### Description

I added dependabot.yml in https://github.com/apache/lucene/pull/14462

Currently it sends us pull requests for:
- github actions
- pip dependencies in dev-tools/

But most github actions we use seem to only pin major version. They are also solid github "official" ones: https://github.com/apache/lucene/network/dependencies?q=ecosystem%3A%22GitHub+Actions%22

For more builds reliability, we could pin them better, e.g. to minor, patch or even hash is probably possible too. Personally I think it would be best to "fully pin" as exactly as we can, and then address update frequency via dependabot.yml (e.g. set the Github Actions ecosystem to monthly PR frequency). Just something never to worry about: code changing underneath us.

Personally I think it is just easier to upgrade iteratively too rather, versus huge major versions. It is also helpful to get digestible-size release notes to take advantage of any new functionality. To get an idea of what I mean, consider the `setup-java` changelogs (we specify `4.*.*` today) https://github.com/actions/setup-java/releases

rmuir:
Currently the github actions are pinned to major versions only, which means the code changes out from under us, without notice.

Instead pin the versions exactly, and have dependabot send us PR updates when things change. This isn't just about breaking changes, it also alerts you to new functionality that may be useful.

Frequency of updates can always be changed, for example, to only send us monthly PRs for the actions changes. But it is best to have no changes without a PR.

Closes #14491

### DIFF
--- a/.github/actions/prepare-for-build/action.yml
+++ b/.github/actions/prepare-for-build/action.yml
@@ -19,7 +19,7 @@ runs:
  using: "composite"
  steps:
     - name: Set up Java (${{ inputs.java-distribution }}, ${{ inputs.java-version }})"
-      uses: actions/setup-java@v4
+      uses: actions/setup-java@c5195efecf7bdfc987ee8bae7a71cb8b11521c00 # v4.7.1
       with:
         distribution: ${{ inputs.java-distribution }}
         java-version: ${{ inputs.java-version }}
--- a/.github/workflows/activity-report.yml
+++ b/.github/workflows/activity-report.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           fetch-depth: 0
 
--- a/.github/workflows/label-pull-request.yml
+++ b/.github/workflows/label-pull-request.yml
@@ -18,6 +18,6 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/labeler@v5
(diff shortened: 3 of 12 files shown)


dweiss:
It would be lovely if we could store these versions in one place somewhere - updating would be easier. In general, I've never had a problem with the official gh actions and keep them at major vx tags. For other actions, I prefer hashes much like here. I don't mind doing it either way.

dweiss (approved):

rmuir:
@dweiss I agree this is overkill for the "official" actions/ stuff. It is solid as I mentioned on the issue. 

Instead this is just about giving ourselves certain guarantees, regardless of the actions in use:
* builds not changing out from underneath us... e.g. hard-to-debug problems
* security guarantees due to using something immutable (git hash) vs mutable (git tag)

I think it is a simple one-time pain, you set it once, then dependabot sends you PRs.

@dweiss to your specific question: I think it is unnecessary. My experience doing this, is that you basically "relinquish manual editing of versions" to automation, so it doesn't matter if you have `foobar@1.2.3` in 87 files. You get one pull request and it updates all 87.

dweiss:
Fair enough. Whenever I've used dependabot I gave up, eventually - I have to deal with javascript and it's a nightmare there.

rmuir:
@dweiss I do like the renovatebot for that use-case, it runs stateless e.g. as a `docker run`. It works great for managing front-end dependencies and runs the real `npm` (for better or worse), or additional commands you ask too. It also has a lot of crowdsourced configuration that's easy to suck in (i think just ask for 'recommended'), that manages stuff like grouping together front-end deps that all depend on each other.

It is written itself in typescript and very friendly to that front-end use-case.

rmuir:
I'll check the outstanding dependabot PR's response for this one, worst-case we can revert it, if something goes wrong.

PR #14493, opened 2025-04-15
Fix RegExp.toString generating invalid RE for CHAR and CHAR_RANGE

Tasssadar:
The `toString` on regexes that contain a CHAR or CHAR_RANGE has been broken since https://github.com/apache/lucene/commit/1efce5444dd40142c55c5a3a30eeebc7b86796c3 , because ASCII character like `s` becomes `\s`, which parses as character class.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
+++ b/lucene/core/src/java/org/apache/lucene/util/automaton/RegExp.java
@@ -854,6 +854,19 @@ public class RegExp {
     return b.toString();
   }
 
+  StringBuilder escapeCharIfNeeded(StringBuilder b, int codePoint) {
+    // From https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#bs
+    // "It is an error to use a backslash prior to any alphabetic character that does not denote
+    // an escaped
+    // construct;"
+    // Plus, ASCII characters conflict with character classes.
+    // Escape only characters which are NOT in [A-Za-z]
+    if (!((codePoint >= 65 && codePoint <= 90) || (codePoint >= 97 && codePoint <= 122))) {
+      b.append("\\");
+    }
+    return b.appendCodePoint(codePoint);
+  }
+
   void toStringBuilder(StringBuilder b) {
     switch (kind) {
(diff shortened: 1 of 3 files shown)


rmuir:
I like the change. We can get fancy and extend it in the future, but the `toString()` today does not do its one job, the output is not readable. This fixes it to be more practical and not try to escape, in unnecessary cases.

Separately: the output of `.toString()` is just that, for humans, and not guaranteed to be reparsable. I can practically guarantee it is a bad idea to do this.

rmuir (approved):
Look at tests in diffs of this change, to see why this is an easy win.

dweiss (approved):

Tasssadar:
> Separately: the output of .toString() is just that, for humans, and not guaranteed to be reparsable. I can practically guarantee it is a bad idea to do this.

What would be the right way to get the original input string then, please? Or do you just declare that `RegExp` does not support that?

Also, it's not just about readability - even as a human, if I see `regex.toString() == "\\s"`, before this fix, I don't know whether it is `s` character or `whitespace` class.

rmuir:
> What would be the right way to get the original input string then, please? Or do you just declare that `RegExp` does not support that?
> 

It doesn't have a method to support this. if the feature is really needed, I would be in favor of doing this in the obvious way if possible: just holding a ref to the original string before any parsing.

> Also, it's not just about readability - even as a human, if I see `regex.toString() == "\\s"`, before this fix, I don't know whether it is `s` character or `whitespace` class.

This PR makes it better though. Personally my opinion: the entire `.toString()` should be rewritten from scratch.  It was never properly designed to be user-friendly in any way.  I still plan to merge your incremental improvement though.

For testing purposes we rely on two things:
- Correct Parse: mostly this is done via `toStringTree()` which returns structured-ish format of the represenation. See TestRegExpParsing.java as example.
- Correct Behavior: mostly this is done via `Automaton` classes which is unambiguous.

Assertions on the `toString()` are mostly just legacy and there's no guarantees about the format of that method AFAIK, so we can change it to be better.

Tasssadar:
Hello, could this be merged?

msfroh (approved):

PR #14494, opened 2025-04-15
Introduce a mapping to map sparse labels to a continuous range

gf2121:
In ASCII encoding, numbers, lowercase letters, uppercase letters, and punctuation have discontinuous value ranges. This PR proposes to use a mapping to map the discontinuous value ranges to a continuous value range starting from 0. This can save some storage space and increase the probability of using the `BITSET` strategy instead of the `ARRAY` strategy, which may also improve some performance.

In luceneutil, this patch reduces ~1.5% size of tip, including ~5% for the `id` field, which encoded with base36 (numbers and lowercase letters). Performance is generally even.


mikemccand:
Whoa, exciting!  I will try to review soon!  Thanks @gf2121.

lucene/core/src/java/org/apache/lucene/codecs/lucene103/blocktree/TrieBuilder.java
```diff
@@ -201,17 +208,53 @@ void save(DataOutput meta, IndexOutput index) throws IOException {
...
+  private int[] saveLabelDictionary(DataOutput out) throws IOException {
+    int[] labels = new int[labelsSeen.cardinality()];
+    int label = -1;
+    for (int i = 0; i < labels.length; i++) {
+      label = labels[i] = labelsSeen.nextSetBit(label + 1);
+    }
+    assert label == labelsSeen.length() - 1
+        || labelsSeen.nextSetBit(label + 1) == DocIdSetIterator.NO_MORE_DOCS;
+    if (labels.length == 0 || labels[labels.length - 1] - labels[0] + 1 == labels.length) {
+      out.writeVInt(0);
```
mikemccand, line 233: Maybe add a comment?

I think what this is doing is writing a `0` byte if there were no labels, or, if the labels were already fully dense/compact to begin with?  And something else (above) will be able to tell the difference?

Edit: OK I see that `0` at read-time means "don't map", i.e. use the int label directly.  And this works for the "no labels at all" case because who cares what the mapping is if you will never use it (sort of like if a tree falls in the woods and nobody hears, did it make any sound?).

lucene/core/src/java/org/apache/lucene/codecs/lucene103/blocktree/TrieReader.java
```diff
@@ -190,9 +218,19 @@ Node lookupChild(int targetLabel, Node parent, Node child) throws IOException {
       return null;
     }
 
+    int lookUpLabel;
+    if (labelMap == null) {
+      lookUpLabel = targetLabel;
+    } else {
+      lookUpLabel = labelMap[targetLabel];
+      if (lookUpLabel == -1) {
```
mikemccand, line 226: Hmm is this paranoia?  Why would we write a label ord that wasn't mapped to a valid label?

lucene/core/src/java/org/apache/lucene/codecs/lucene103/blocktree/TrieBuilder.java
```diff
@@ -201,17 +208,53 @@ void save(DataOutput meta, IndexOutput index) throws IOException {
...
+
+  private void saveNodes(IndexOutput index, int[] labelMap) throws IOException {
     final long startFP = index.getFilePointer();
-    Deque<Node> stack = new ArrayDeque<>();
+    Deque<Node> stack =
+        new ArrayDeque<>() {
+          @Override
+          public void push(Node node) {
+            if (labelMap != null) {
+              node.label = labelMap[node.label];
```
mikemccand, line 253: Maybe `assert node.label != -1`?  This label must've been defined?

lucene/core/src/java/org/apache/lucene/codecs/lucene103/blocktree/TrieReader.java
```diff
@@ -74,14 +77,39 @@ IndexInput floorData(TrieReader r) throws IOException {
...
     this.root = new Node();
     load(root, rootFP);
   }
 
+  private static int[] labelMap(DataInput in) throws IOException {
+    int cnt = in.readVInt();
+    if (cnt == 0) {
+      return null;
+    } else {
+      int[] labelMap = new int[TrieBuilder.BYTE_RANGE];
```
mikemccand, line 104: Could we make this `byte[]` instad?  This added heap could add up for many shards X many indices for multitenant (OpenSearch, Elastcisearch, Solr, ...) cases, and with many separate indexed fields?

lucene/core/src/test/org/apache/lucene/codecs/lucene103/blocktree/TestTrie.java
```diff
@@ -71,6 +74,46 @@ public void testOneByteTerms() throws Exception {
     }
   }
 
+  public void testContinuousValues() throws Exception {
```
mikemccand, line 77: Could we also explicitly test both ends (0, 255) directly?  Maybe 2-3 labels starting with 0, and ending with 255?  Trying to tease out any scary ob1-Kenobi bugs!

lucene/core/src/java/org/apache/lucene/codecs/lucene103/blocktree/TrieBuilder.java
```diff
@@ -63,7 +67,7 @@ private enum Status {
   private static class Node {
 
     // The utf8 digit that leads to this Node, 0 for root node
-    private final int label;
+    private int label;
```
mikemccand, line 70: Hmm why did we need to un-final?  Don't we remap on write, not update each `Node`'s label in place?

lucene/core/src/java/org/apache/lucene/codecs/lucene103/blocktree/TrieReader.java
```diff
@@ -74,14 +77,39 @@ IndexInput floorData(TrieReader r) throws IOException {
   final RandomAccessInput access;
   final IndexInput input;
   final Node root;
+  final int[] labelMap;
```
mikemccand, line 80: Could you add a comment explaining a bit what's happening here?  Explain that it is global across the entire Trie, and it remaps the written labels to compact ordinals for more efficient storage w/ negligible impact on terms lookup performance?

mikemccand (approved):
I like this opto!

This was my first pass through so I left some probably head scratcher / fresh eyes confusing sort of comments :)

Thnks @gf2121 and sorry for the slow review... thank you Stale Bot (hmm we really need a better name for this incredibly important bot -- Lars?).

lucene/core/src/java/org/apache/lucene/codecs/lucene103/blocktree/TrieReader.java
gf2121: This is on the reader side, we could be looking for a label that does not exist in this trie.

lucene/core/src/java/org/apache/lucene/codecs/lucene103/blocktree/TrieBuilder.java
gf2121: We are modifying the label during saving:

https://github.com/apache/lucene/pull/14494/files#diff-833f2a438947805927e571611f25e77ad4be6944fb2ff8816b3168ce2b48182eR253

lucene/core/src/test/org/apache/lucene/codecs/lucene103/blocktree/TestTrie.java
gf2121: I'm not sure i get your point but i added a test `testContinuousValues0To255` :)

lucene/core/src/java/org/apache/lucene/codecs/lucene103/blocktree/TrieReader.java
keikino: If the main reasoning for using an `int[]` is because we want to check for existence, and we'd otherwise get the same functionality by using the `byte[]`, I agree that using something like a `bitset` instead would be better -

256 size `bitset`: 256/64 bytes + overhead = ~48 bytes
256 size `byte[]` : 256 bytes + overhead = ~272 bytes
sum ~320 bytes

vs

256 size `int[]`: 256*4 + overhead = ~1048 bytes

We'd get a ~70% memory footprint reduction by using the `bitset`+`byte[]`, and as this will compound over all of the indexes, I think it's definitely a worthy memory optimization!

gf2121: Thanks for the feedback @keikino , I agree that this does reduce memory, but there is an extra `bitset#get` overhead for each child lookup. I'll see what luceneutil will say.

keikino: Thank you for looking into it @gf2121 !

Issue #14490, opened 2025-04-15
update PRs for java dependencies


rmuir:
### Description

I added dependabot.yml in https://github.com/apache/lucene/pull/14462

Currently it sends us pull requests for:
- github actions
- pip dependencies in `dev-tools/`

But nothing yet for java dependencies. I think it might be enough to rename `versions.toml` to `libs.versions.toml` to get (build-failing) pull requests? From the docs I have read, the filename is not configurable.

If the filename is not so important, it would be nice to rename it, just so that github "understands" our dependency tree and allows for features around that (such as security ones).

In order to make PRs nice, where they stand a chance to pass, it would be more work. Seems the recommended way is to integrate with actions in order to run the "post upgrade commands" and issue a commit with them. (#14506 

Alternatively we could do renovatebot, but we don't have many dependencies and our needs are simple, so it would be cool if we could have dependabot fully working for us.

rmuir:
I'm not actively working this yet, it is just for discussion. But if we can coerce the thing to work well with our project, I think it is beneficial. Better to have a failing PR, well-formatted, with release notes info, etc, to upgrade a dependency, than no PR at all? And managing completely manually?

I do think it would be a little tricky to configure all the automation necessary to create "passing by default" PRs, that dont require additional commits by developers, to be merged. It is more than just regenerating lockfile and licenses/shas, it is also regenerate tasks!

dweiss:
I think the "gradle convention" is to keep libs.version.toml under gradle/ and I think it's what dependabot understands - and I personally hate this, without any specific objective reason... Moving it is easy but, to me, it hides the source of dependencies from its prominent top-level location. See this commit: https://github.com/dweiss/lucene/commit/8b99dab8298761398e62a377c4e3b1af1cda9ad5

If you feel it's ok, feel free to cherry pick it to main.

rmuir:
@dweiss mainly I am curious if we can see the new dependencies under https://github.com/apache/lucene/network/dependencies

I don't know how to test this (i think it only evals default branch). But as you can see, several capabilities such as security alerts, SBOM generation, etc would all suddenly work. Separately we can configure the dependabot for that ecosystem, if we can get it to be recognized.

Not sure where the file has to be for that, maybe the root is ok? `dependabot.yml` lets us specify the root directory, just not the filename to look for. For example see https://github.com/apache/lucene/blob/main/.github/dependabot.yml#L15 where we configure it for python only in dev-tools/

dweiss:
[...]

PR #14495, opened 2025-04-15
Move version catalog and enable dependabot scanning of gradle dependencies

dweiss:
#14490


.github/dependabot.yml
```diff
@@ -20,3 +20,13 @@ updates:
...
     labels:
       - dependencies
+
+  - package-ecosystem: gradle
+    directory: /
+    schedule:
+      interval: weekly
+      day: tuesday
+    commit-message:
+      prefix: java(deps)
```
rmuir, line 30: for a conventional commit:

```suggestion
      prefix: deps(java)
```

rmuir (approved):
push it: Let's iterate, and see if we can make it work. Just merge it and see what happens LOL.

I guess we know it will be a lot of PRs at once. And maybe all builds fail initially because we dont have sophisticated logic. it is super easy to just push commits to the dependabot-opened-branch and get it passing.

We just have to prioritize automation initially, then the deps will be easier over time IMHO. If we invest no effort, they will suck and all fail the build. Lets strive for quality pull requests with (usually) passing builds and the info you need to make decisions.

PR #14501, opened 2025-04-15
Remove dependabot pr limit=1 for java and actions ecosystems

rmuir:
We'll never upgrade anything in this way, you can bottleneck the entire process on a single dependency.


stefanvodita (approved):

rmuir:
this will probably result in a quantity of PRs...

PR #14017, opened 2024-11-25
Stop using `SlowImpactsEnum` for terms whose `docFreq` is less than 128.

jpountz:
We currently use `SlowImpactsEnum` for terms whose `docFreq` is less than 128 because it's convenient as these terms don't have impacts anyway. But a recent slowdown on nightly benchmarks suggests that this contributes to making some hot calls more polymorphic than we'd like, so this PR moves such terms back to the regular impacts enums.

PR #14511, opened 2025-04-16
Provide better impacts for fields indexed with IndexOptions.DOCS

### LINKED ISSUE #14445: Term Query is slower post Lucene 9.12 for fields with IndexOptions.DOCS
### Description

A term query `searcher.search(new TermQuery(new Term(fieldName, fieldValue)), hits)` on a field which is only indexed with `IndexOptions.DOCS` is slower from 9.12.0 onwards. 

Root cause : 

- Collector sets the minimum competitive score on DISI after it has gathered the required hits. [TopScoreDocCollector sets it here](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/TopScoreDocCollector.java#L173). 
It is an indication to the DISI that when scorer calls the `nextDoc` again it should return a document that has a score >= minCompetitiveScore. 

- ImpactsDISI does a [shallowAdvance via MaxScoreCache during nextDoc](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/MaxScoreCache.java#L58) ( doesn't load the docIds into buffer just moves the file pointers and updates other internal variables in the underlying ImpactsEnum ) after the minCompetitiveScore is set by collector. It tries to move to the block in postings that will contain the target ( docId + 1 ) with score >= minCompetitiveScore.

- The issue occurs when we try to [compute the max score at level zero after the shallowAdvance](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/ImpactsDISI.java#L75-L76) This happens because from 9.12.0 the `Impact` returned is always `DUMMY_IMPACTS` for fields with `IndexOptions.DOCS`. This causes the maxScore calculated by [Similarity$SimScorer](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/MaxScoreCache.java#L72-L79) to be always greater than the minimum competitive score which leads to all docIds being treated as competitive even though they aren't. 

Possible Solutions : 

- Store the field with `IndexOptions.DOCS_AND_FREQS` as this doesn't return `DUMMY_IMPACTS` but actually reads the impact value. Nearly doubles the index size.

- If we don't care about the score and are using a TermQuery as a pure filter, then wrapping the query with a ConstantScoreQuery resolves the issue. This works because [ConstantScoreScorer sets it's delegate as empty](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/ConstantScoreScorer.java#L124) once minimum competitive score is set thus early terminating DefaultBulkScorer. Thanks @msfroh for suggesting this alternative.

- Another way of resolving the issue is [returning a frequency of 1](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene101/Lucene101PostingsReader.java#L74) instead of `NO_MORE_DOCS` in DummyImpacts returned by BlockPostingsEnum. Thanks @jpountz for suggesting this alternative.
However, we would need to handle cases like [ExactPhraseMatcher which depend on the frequency of dummy impacts](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/ExactPhraseMatcher.java#L282-L284).

- I also tried a different approach of not returning `DUMMY_IMPACTS` for fields with `IndexOptions.DOCS` since we already have the Impact information required. Below is a diff : 

<details>

<summary>Expand for Git Diff</summary>

```
diff --git a/lucene/core/src/java/org/apache/lucene/codecs/lucene101/Lucene101PostingsReader.java b/lucene/core/src/java/org/apache/lucene/codecs/lucene101/Lucene101PostingsReader.java
index 1efcdf554dd..0a7c048ee95 100644
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene101/Lucene101PostingsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene101/Lucene101PostingsReader.java
@@ -422,7 +422,7 @@ public final class Lucene101PostingsReader extends PostingsReaderBase {
         Arrays.fill(freqBuffer, 1);
       }

-      if (needsFreq && needsImpacts) {
+      if (needsImpacts) {
         level0SerializedImpacts = new BytesRef(maxImpactNumBytesAtLevel0);
         level1SerializedImpacts = new BytesRef(maxImpactNumBytesAtLevel1);
         level0Impacts = new MutableImpactList(maxNumImpactsAtLevel0);
@@ -1107,9 +1107,6 @@ public final class Lucene101PostingsReader extends PostingsReaderBase {

           @Override
           public int getDocIdUpTo(int level) {
-            if (indexHasFreq == false) {
-              return NO_MORE_DOCS;
-            }
             if (level == 0) {
               return level0LastDocID;
             }
@@ -1118,14 +1115,12 @@ public final class Lucene101PostingsReader extends PostingsReaderBase {

           @Override
           public List<Impact> getImpacts(int level) {
-            if (indexHasFreq) {
               if (level == 0 && level0LastDocID != NO_MORE_DOCS) {
                 return readImpacts(level0SerializedImpacts, level0Impacts);
               }
               if (level == 1) {
                 return readImpacts(level1SerializedImpacts, level1Impacts);
               }
-            }
             return DUMMY_IMPACTS;
           }
```

</details>

Will follow up with a PR on the same after discussion b/w approach 3 and 4. 

Create a Big5 index using OSB workload using OpenSearch 2.17 and OS >=2.18 

```
opensearch-benchmark execute-test --target-hosts http://127.0.0.1:9200 --workload big5 --client-options timeout:120 --workload-params="number_of_shards:1,bulk_indexing_clients:1,number_of_replicas:0" --kill-running-processes --include-tasks delete-index,create-index,check-cluster-health,index-append,refresh-after-index,force-merge,refresh-after-force-merge,wait-until-merges-finish
```

Run a term query as below : 

```
curl -X POST "http://localhost:9200/big5/_search" \
-H "Content-Type: application/json" \
-d '{
  "query": {
    "term": {
      "process.name": "kernel"
    }
  }
}'
```

In OS 2.17 ( Lucene 9.11.1 ) the query will complete in <= 5ms whereas in OS 2.18 ( Lucene 9.12.0 ) the query will take >= 200ms 
Issue persists in 10.1.0 as well.

expani:
Postings always return impacts with freq=Integer.MAX_VALUE and norm=1 when frequencies are not indexed (IndexOptions.DOCS). This significantly overestimates the score upper bound of term queries, since the similarity scorer is effectively called with freq=1 all the time in this case (and either norm=1 if norms are not indexed, or the number of terms in the field otherwise).

This updates postings to always return impacts with freq=1 and norm=1 when frequencies are not indexed, which helps compute better score upper bounds, and in-turn makes dynamic pruning perform better.

Closes #14445

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
@@ -44,7 +44,6 @@ import org.apache.lucene.index.IndexFileNames;
 import org.apache.lucene.index.IndexOptions;
 import org.apache.lucene.index.PostingsEnum;
 import org.apache.lucene.index.SegmentReadState;
-import org.apache.lucene.index.SlowImpactsEnum;
 import org.apache.lucene.internal.vectorization.PostingDecodingUtil;
 import org.apache.lucene.internal.vectorization.VectorizationProvider;
 import org.apache.lucene.store.ByteArrayDataInput;
@@ -67,18 +66,15 @@ import org.apache.lucene.util.VectorUtil;
 public final class Lucene103PostingsReader extends PostingsReaderBase {
 
   static final VectorizationProvider VECTORIZATION_PROVIDER = VectorizationProvider.getInstance();
-
   // Dummy impacts, composed of the maximum possible term frequency and the lowest possible
   // (unsigned) norm value. This is typically used on tail blocks, which don't actually record
-  // impacts as the storage overhead would not be worth any query evaluation speedup, since
-  // there's
+  // impacts as the storage overhead would not be worth any query evaluation speedup, since there's
   // less than 128 docs left to evaluate anyway.
   private static final List<Impact> DUMMY_IMPACTS =
       Collections.singletonList(new Impact(Integer.MAX_VALUE, 1L));
 
-  // We stopped storing a placeholder impact with freq=1 for fields with IndexOptions.DOCS after
-  // 9.12.0
-  private static final List<Impact> NON_COMPETITIVE_IMPACTS =
+  // We stopped storing a placeholder impact with freq=1 for fields with DOCS after 9.12.0
+  private static final List<Impact> DUMMY_IMPACTS_NO_FREQS =
       Collections.singletonList(new Impact(1, 1L));
 
   private final IndexInput docIn;
@@ -290,10 +286,6 @@ public final class Lucene103PostingsReader extends PostingsReaderBase {
   @Override
   public ImpactsEnum impacts(FieldInfo fieldInfo, BlockTermState state, int flags)
       throws IOException {
-    if (state.docFreq <= BLOCK_SIZE) {
-      // no skip data
-      return new SlowImpactsEnum(postings(fieldInfo, state, null, flags));
-    }
     return new BlockPostingsEnum(fieldInfo, flags, true).reset((IntBlockTermState) state, flags);
   }
 
@@ -1298,11 +1290,14 @@ public final class Lucene103PostingsReader extends PostingsReaderBase {
 
           @Override
           public int numLevels() {
-            return level1LastDocID == NO_MORE_DOCS ? 1 : 2;
+            return indexHasFreq == false || level1LastDocID == NO_MORE_DOCS ? 1 : 2;
           }
 
           @Override
           public int getDocIdUpTo(int level) {
+            if (indexHasFreq == false) {
+              return NO_MORE_DOCS;
+            }
             if (level == 0) {
               return level0LastDocID;
             }
@@ -1311,16 +1306,16 @@ public final class Lucene103PostingsReader extends PostingsReaderBase {
 
           @Override
           public List<Impact> getImpacts(int level) {
-            if (indexHasFreq) {
-              if (level == 0 && level0LastDocID != NO_MORE_DOCS) {
-                return readImpacts(level0SerializedImpacts, level0Impacts);
-              }
-              if (level == 1) {
-                return readImpacts(level1SerializedImpacts, level1Impacts);
-              }
-              return DUMMY_IMPACTS;
+            if (indexHasFreq == false) {
+              return DUMMY_IMPACTS_NO_FREQS;
             }
-            return NON_COMPETITIVE_IMPACTS;
+            if (level == 0 && level0LastDocID != NO_MORE_DOCS) {
+              return readImpacts(level0SerializedImpacts, level0Impacts);
+            }
+            if (level == 1) {
+              return readImpacts(level1SerializedImpacts, level1Impacts);
+            }
+            return DUMMY_IMPACTS;
           }
 
           private List<Impact> readImpacts(BytesRef serialized, MutableImpactList impactsList) {
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
@@ -44,6 +44,7 @@ import org.apache.lucene.index.IndexFileNames;
 import org.apache.lucene.index.IndexOptions;
 import org.apache.lucene.index.PostingsEnum;
 import org.apache.lucene.index.SegmentReadState;
+import org.apache.lucene.index.SlowImpactsEnum;
 import org.apache.lucene.internal.vectorization.PostingDecodingUtil;
 import org.apache.lucene.internal.vectorization.VectorizationProvider;
 import org.apache.lucene.store.ByteArrayDataInput;
@@ -72,6 +73,11 @@ public final class Lucene103PostingsReader extends PostingsReaderBase {
   // less than 128 docs left to evaluate anyway.
   private static final List<Impact> DUMMY_IMPACTS =
       Collections.singletonList(new Impact(Integer.MAX_VALUE, 1L));
+
+  // We stopped storing a placeholder impact with freq=1 for fields with IndexOptions.DOCS
+  // from 9.12.0 onwards with @org.apache.lucene.backward_codecs.lucene912.Lucene912PostingsReader
+  private static final List<Impact> NON_COMPETITIVE_IMPACTS =
+          Collections.singletonList(new Impact(1, 1L));
 
   private final IndexInput docIn;
   private final IndexInput posIn;
@@ -282,6 +288,10 @@ public final class Lucene103PostingsReader extends PostingsReaderBase {
   @Override
   public ImpactsEnum impacts(FieldInfo fieldInfo, BlockTermState state, int flags)
       throws IOException {
+    if (state.docFreq <= BLOCK_SIZE) {
+      // no skip data
+      return new SlowImpactsEnum(postings(fieldInfo, state, null, flags));
+    }
     return new BlockPostingsEnum(fieldInfo, flags, true).reset((IntBlockTermState) state, flags);
   }
 
@@ -1286,14 +1296,11 @@ public final class Lucene103PostingsReader extends PostingsReaderBase {
 
           @Override
           public int numLevels() {
-            return indexHasFreq == false || level1LastDocID == NO_MORE_DOCS ? 1 : 2;
+            return level1LastDocID == NO_MORE_DOCS ? 1 : 2;
           }
 
           @Override
           public int getDocIdUpTo(int level) {
-            if (indexHasFreq == false) {
-              return NO_MORE_DOCS;
-            }
             if (level == 0) {
               return level0LastDocID;
             }
@@ -1310,7 +1317,7 @@ public final class Lucene103PostingsReader extends PostingsReaderBase {
                 return readImpacts(level1SerializedImpacts, level1Impacts);
               }
             }
-            return DUMMY_IMPACTS;
+            return NON_COMPETITIVE_IMPACTS;
           }
 
           private List<Impact> readImpacts(BytesRef serialized, MutableImpactList impactsList) {
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
@@ -68,6 +68,14 @@ public final class Lucene103PostingsReader extends PostingsReaderBase {
 
   static final VectorizationProvider VECTORIZATION_PROVIDER = VectorizationProvider.getInstance();
 
+  // Dummy impacts, composed of the maximum possible term frequency and the lowest possible
+  // (unsigned) norm value. This is typically used on tail blocks, which don't actually record
+  // impacts as the storage overhead would not be worth any query evaluation speedup, since
+  // there's
+  // less than 128 docs left to evaluate anyway.
+  private static final List<Impact> DUMMY_IMPACTS =
+      Collections.singletonList(new Impact(Integer.MAX_VALUE, 1L));
+
   // We stopped storing a placeholder impact with freq=1 for fields with IndexOptions.DOCS after
   // 9.12.0
   private static final List<Impact> NON_COMPETITIVE_IMPACTS =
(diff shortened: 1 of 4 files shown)


lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
```diff
@@ -1310,7 +1317,7 @@ public List<Impact> getImpacts(int level) {
                 return readImpacts(level1SerializedImpacts, level1Impacts);
               }
             }
-            return DUMMY_IMPACTS;
+            return NON_COMPETITIVE_IMPACTS;
```
msfroh, line 1320: This was the only reference to `DUMMY_IMPACTS`, right? Can we remove it?

lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
```diff
@@ -282,6 +288,10 @@ public PostingsEnum postings(
   @Override
   public ImpactsEnum impacts(FieldInfo fieldInfo, BlockTermState state, int flags)
       throws IOException {
+    if (state.docFreq <= BLOCK_SIZE) {
+      // no skip data
+      return new SlowImpactsEnum(postings(fieldInfo, state, null, flags));
+    }
```
msfroh, line 294: This is essentially taking the place of `DUMMY_IMPACTS`, right? 

It's the thing that kicks in on tail blocks, which is what `DUMMY_IMPACTS` was there for. (I'm trying to make sure I understand the change.)

expani: DUMMY_IMPACTS in PostingsReader was a `List<Impact>` 
whereas 
the DUMMY_IMPACTS returned by SlowImpactsEnum is a `Impacts` which would still apply for cases where using the skip list doesn't make sense `state.docFreq <= BLOCK_SIZE` 

I picked this check from `Lucene99PostingsReader` where we stored a default impact of freq=1 even if the field was indexed with `IndexOptions.DOCS`

lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
expani: Good catch. 
This is unused after the change, so can remove it.

lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
expani: You understanding is spot on.

The bug was this check getting extended to fields with `IndexOptions.DOCS`

lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
expani: Looks like there was a case in CheckIndex failing due to this. 
So added it back for fields that have at least freqs enabled. 

Failing test
```
./gradlew :lucene:benchmark:test --tests "org.apache.lucene.benchmark.quality.TestQualityRun.testTrecQuality" -Ptests.jvms=1 -Ptests.jvmargs= -Ptests.seed=6AD36F6087951803 -Ptests.useSecurityManager=true -Ptests.gui=true -Ptests.file.encoding=ISO-8859-1 -Ptests.vectorsize=256 -Ptests.forceintegervectors=true
```

lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
```diff
@@ -282,6 +290,10 @@ public PostingsEnum postings(
...
```
jpountz, line 296: Let's not use `SlowImpactsEnum` again or it would cancel the speedup from https://github.com/apache/lucene/pull/14017 (annotation HJ at https://benchmarks.mikemccandless.com/OrHighHigh.html).

lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
```diff
@@ -66,13 +67,20 @@
...
+  // impacts as the storage overhead would not be worth any query evaluation speedup, since
+  // there's
   // less than 128 docs left to evaluate anyway.
   private static final List<Impact> DUMMY_IMPACTS =
       Collections.singletonList(new Impact(Integer.MAX_VALUE, 1L));
 
+  // We stopped storing a placeholder impact with freq=1 for fields with IndexOptions.DOCS after
+  // 9.12.0
+  private static final List<Impact> NON_COMPETITIVE_IMPACTS =
+      Collections.singletonList(new Impact(1, 1L));
```
jpountz, line 82: Maybe call it `DUMMY_IMPACTS_NO_FREQS` or something like that to draw a parallel with `DUMMY_IMPACTS` above?

lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
```diff
@@ -1286,14 +1298,11 @@ public long cost() {
 
           @Override
           public int numLevels() {
-            return indexHasFreq == false || level1LastDocID == NO_MORE_DOCS ? 1 : 2;
+            return level1LastDocID == NO_MORE_DOCS ? 1 : 2;
```
jpountz, line 1301: I would keep it the way it was for now. Exposing a single level that covers the whole doc ID space may cause inefficiencies with some scorers, but we should fix these scorers instead of tuning this class. If all docs have the same score, it should be just fine to return a single level of impact that covers the whole doc ID space.

lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
```diff
@@ -1286,14 +1298,11 @@ public long cost() {
...
           }
 
           @Override
           public int getDocIdUpTo(int level) {
-            if (indexHasFreq == false) {
-              return NO_MORE_DOCS;
-            }
```
jpountz, line 1296: Likewise, let's keep this?

lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
```diff
@@ -1309,8 +1318,9 @@ public List<Impact> getImpacts(int level) {
               if (level == 1) {
                 return readImpacts(level1SerializedImpacts, level1Impacts);
               }
+              return DUMMY_IMPACTS;
             }
-            return DUMMY_IMPACTS;
+            return NON_COMPETITIVE_IMPACTS;
```
jpountz, line 1323: I would keep it the way it was before and just add this at the beginning of this method (similar to `getDocIdUpTo`):

```java
            if (indexHasFreq == false) {
              return DUMMY_IMPACTS_NO_FREQS;
            }
```

jpountz (commented):
Sorry for only seeing now, I left some suggestions as to how to update your change.

lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
expani: Done

lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
expani: I had made these changes to bring back the same behavior as 9.11.1

Without these changes, it doesn't read the skip data [ and exits here ](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/ImpactsDISI.java#L85) right after the min competitive score is set.

Whereas in 9.11.1, it reads the skip data [by entering this part of ImpactDISI](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/ImpactsDISI.java#L88-L97) which @msfroh described in the mail chain

```
It was fast because (once the collector has filled its priority queue), we'd check the (constant) impacts to find the first block that's strictly better than the min competitive score. Since all scores are equal, that would quickly skip to the end.
```

Although, not keeping this achieves the same result. Should I add a TODO to remove this later after we fix the scrorers you mentioned ? 

Also, can this give incorrect results when norms are enabled ? Since, we are not reading the impacts.

lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
expani: Done

lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
jpountz: I understand why this change helps, but this problem is not unique to term queries indexed with IndexOptions.DOCS, `ConstantScoreScorer` (used by many queries) causes the same problem. I'd rather fix the root cause than merge this workaround that we may forget to remove later.

> Also, can this give incorrect results when norms are enabled ? Since, we are not reading the impacts.

Impacts help compute upper bounds of the score over ranges of doc IDs. Since scores are required to not increase when the norm increases, a score computed with norm=1 will always be greater than or equal to a score computed with any other norm value. So this is correct, it may just return a score upper bound that is greater than the actual best score from the block.

jpountz (approved):
LGTM. Can you add a CHANGES entry under 10.3 and undo the new line in SlowImpactsEnum?

lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
expani: > Since scores are required to not increase when the norm increases

Wasn't aware of this. Makes sense now.

expani:
Addressed comments.
I want to backport these to `9.12.x` and `10.2.x` as well. Will open separate PRs for the same.

jpountz:
This sounds safe enough for 10.2.1 for me. Can you move the CHANGES entry to 10.2.1 then? cc @ChrisHegarty

jpountz:
I hope you don't mind, I updated this PR title and description to better reflect the change.

expani:
>I hope you don't mind, I updated this PR title and description to better reflect the change.

Not at all. Thanks for taking the time to explain the different pieces of this code. 

It was really fun debugging this and would definitely love to visit this part of the code again.

expani:
@ChrisHegarty @jpountz Moved the change log to 10.2.1

ChrisHegarty:
> @ChrisHegarty @jpountz Moved the change log to 10.2.1

Eh! I think you moved it to 10.2.0, rather than 10.2.1.

expani:
Oops hadn't rebased with main. Fixed it now.

ChrisHegarty:
> This sounds safe enough for 10.2.1 for me. Can you move the CHANGES entry to 10.2.1 then? cc @ChrisHegarty

What am I missing?  This is not applicable to 10.2.1, since the only changed file is Lucene103PostingsReader.java which is not present in 10.2 !

expani:
>What am I missing? This is not applicable to 10.2.1, since the only changed file is Lucene103PostingsReader.java which is not present in 10.2 ! Did the rebase mess something up ?

I had updated 103PostingsReader as the initial plan was not to backport. 

Updated 101PostingsReader which is used in 10.2.1 

Should I also raise against some other branch as well ?

expani:
I made the same change in `Lucene101PostingsReader` as well.

gf2121:
To be clear, i raised https://github.com/apache/lucene/pull/14557 and https://github.com/apache/lucene/pull/14558 for backporting. I plan to merge this now if no one objects.

gf2121:
@expani could you resolve the conflicts so that i can merge?

PR #14523, opened 2025-04-18
Correct TermOrdValComparator competitive iterator intoBitSet implementation

### LINKED ISSUE #14517: Strange stack traces for new bitset focused doc iterators
### Description

With Lucene 10.2, we have seen some exceptions that are rather troubling. It appears that the into bit set code is buggy when utilizing multiple layers of iterators. 

It does seem strange that we have ran into this type of failure on a couple of different paths. This indicates a significant API shift (e.g. how iterators need to be handled is now fundamentally different), or some underlying bugs in this new implementation.

Here is one trace. This one seems possible if the iterator has iterated passed `windowBase` and `windowMax` in: https://github.com/apache/lucene/blob/2772951f5f15ea02e6accd44f6904077b4a09060/lucene/core/src/java/org/apache/lucene/search/DenseConjunctionBulkScorer.java#L238

```
      DocIdSetIterator lead = iterators.get(0);
      if (lead.docID() < windowBase) {
        lead.advance(windowBase);
      }
      lead.intoBitSet(windowMax, windowMatches, windowBase);
```

```
Caused by: java.lang.ArrayIndexOutOfBoundsException: Index -33 out of bounds for length 64
	at org.apache.lucene.core@10.2.0/org.apache.lucene.util.FixedBitSet.set(FixedBitSet.java:283)
	at org.apache.lucene.core@10.2.0/org.apache.lucene.search.DocIdSetIterator.intoBitSet(DocIdSetIterator.java:268)
	at org.apache.lucene.core@10.2.0/org.apache.lucene.search.comparators.TermOrdValComparator$CompetitiveIterator.intoBitSet(TermOrdValComparator.java:539)
	at org.apache.lucene.core@10.2.0/org.apache.lucene.search.DenseConjunctionBulkScorer.scoreWindowUsingBitSet(DenseConjunctionBulkScorer.java:242)
	at org.apache.lucene.core@10.2.0/org.apache.lucene.search.DenseConjunctionBulkScorer.scoreWindow(DenseConjunctionBulkScorer.java:210)
	at org.apache.lucene.core@10.2.0/org.apache.lucene.search.DenseConjunctionBulkScorer.score(DenseConjunctionBulkScorer.java:132)
	at org.elasticsearch.server@9.1.0/org.elasticsearch.search.internal.CancellableBulkScorer.score(CancellableBulkScorer.java:46)
	at org.elasticsearch.server@9.1.0/org.elasticsearch.search.internal.ContextIndexSearcher.searchLeaf(ContextIndexSearcher.java:460)
	at org.apache.lucene.core@10.2.0/org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:809)
	at org.elasticsearch.server@9.1.0/org.elasticsearch.search.internal.ContextIndexSearcher.search(ContextIndexSearcher.java:387)
	at org.elasticsearch.server@9.1.0/org.elasticsearch.search.internal.ContextIndexSearcher.lambda$search$3(ContextIndexSearcher.java:365)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
	at org.apache.lucene.core@10.2.0/org.apache.lucene.search.TaskExecutor$Task.run(TaskExecutor.java:173)
	at org.apache.lucene.core@10.2.0/org.apache.lucene.search.TaskExecutor.invokeAll(TaskExecutor.java:111)
	at org.elasticsearch.server@9.1.0/org.elasticsearch.search.internal.ContextIndexSearcher.search(ContextIndexSearcher.java:369)
	at org.elasticsearch.server@9.1.0/org.elasticsearch.search.internal.ContextIndexSearcher.search(ContextIndexSearcher.java:336)
	at org.elasticsearch.server@9.1.0/org.elasticsearch.search.query.QueryPhase.addCollectorsAndSearch(QueryPhase.java:212)
	... 15 more
```

This following trace seems possible if `DISIDocIdStream#count(upto)` where `upto` is > `max`. Even though we do a check for `iterator.docID() >= upTo` there is no check to verify we are iterating past the configured `max`. Maybe we are handling the iterators incorrectly?

https://github.com/apache/lucene/blob/0018e62cf21fe589633b01812fe9ec49fd42b426/lucene/core/src/java/org/apache/lucene/search/DISIDocIdStream.java#L53

```
	Suppressed: java.lang.IndexOutOfBoundsException: Range [0, -2) out of bounds for length 4096
		at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
		at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckFromToIndex(Preconditions.java:112)
		at java.base/jdk.internal.util.Preconditions.checkFromToIndex(Preconditions.java:349)
		at java.base/java.util.Objects.checkFromToIndex(Objects.java:391)
		at org.apache.lucene.core@10.2.0/org.apache.lucene.util.FixedBitSet.cardinality(FixedBitSet.java:212)
		at org.apache.lucene.core@10.2.0/org.apache.lucene.search.DISIDocIdStream.count(DISIDocIdStream.java:63)
		at org.apache.lucene.core@10.2.0/org.apache.lucene.search.DocIdStream.count(DocIdStream.java:50)
		at org.apache.lucene.core@10.2.0/org.apache.lucene.search.TotalHitCountCollector$1.collect(TotalHitCountCollector.java:69)
		at org.apache.lucene.core@10.2.0/org.apache.lucene.search.DenseConjunctionBulkScorer.scoreWindow(DenseConjunctionBulkScorer.java:208)
		at org.apache.lucene.core@10.2.0/org.apache.lucene.search.DenseConjunctionBulkScorer.score(DenseConjunctionBulkScorer.java:132)
		at org.apache.lucene.core@10.2.0/org.apache.lucene.search.ReqExclBulkScorer.score(ReqExclBulkScorer.java:66)
		at org.elasticsearch.server@9.1.0/org.elasticsearch.search.internal.CancellableBulkScorer.score(CancellableBulkScorer.java:46)
		at org.elasticsearch.server@9.1.0/org.elasticsearch.search.internal.ContextIndexSearcher.searchLeaf(ContextIndexSearcher.java:460)
		at org.apache.lucene.core@10.2.0/org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:809)
		at org.elasticsearch.server@9.1.0/org.elasticsearch.search.internal.ContextIndexSearcher.search(ContextIndexSearcher.java:387)
		at org.elasticsearch.server@9.1.0/org.elasticsearch.search.internal.ContextIndexSearcher.lambda$search$3(ContextIndexSearcher.java:365)
		at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
		at org.apache.lucene.core@10.2.0/org.apache.lucene.search.TaskExecutor$Task.run(TaskExecutor.java:173)
		at org.apache.lucene.core@10.2.0/org.apache.lucene.search.TaskExecutor.lambda$invokeAll$1(TaskExecutor.java:98)
		... 6 more
```

Lucene 10.2

benwtrent:
Similar to @ChrisHegarty 's change for the count fix, this will move up the `max` check to before the into bit set. It seems like we could be calling intobitset erroneously on some edge cases. 

Additionally, TermOrdValComparator's competitive iterator will ensure that the docsWithField iterator (used when not all docs have a given field), is at least the baseline of the scoring window when doing intoBitSet.

closes: https://github.com/apache/lucene/issues/14517


lucene/core/src/java/org/apache/lucene/search/DisjunctionDISIApproximation.java
```diff
@@ -146,7 +146,13 @@ public int advance(int target) throws IOException {
...
   public void intoBitSet(int upTo, FixedBitSet bitSet, int offset) throws IOException {
     while (leadTop.doc < upTo) {
-      leadTop.approximation.intoBitSet(upTo, bitSet, offset);
+      int intoBitSetBase = offset;
+      if (leadTop.approximation.docID() < offset) {
+        intoBitSetBase = leadTop.approximation.advance(offset);
+      }
+      if (intoBitSetBase < upTo) {
+        leadTop.approximation.intoBitSet(upTo, bitSet, intoBitSetBase);
+      }
```
benwtrent, line 155: It seemed weird to me that we have this inconsistency in the API between DCBS and the disjunction checker. So, I added advancement here as well, guarding the `intoBitSet` similarly.

lucene/core/src/java/org/apache/lucene/search/comparators/NumericComparator.java
```diff
@@ -434,6 +434,9 @@ public void intoBitSet(int upTo, FixedBitSet bitSet, int offset) throws IOExcept
           if (competitiveIterator.docID() < doc) {
             competitiveIterator.advance(doc);
           }
+          if (doc >= upTo) {
+            return;
+          }
```
benwtrent, line 439: We have this check in many (if not all?) of the intoBitSet implementations, seemed sane to put it here as well, don't bother calling it if we advanced past `upTo`.

lucene/core/src/java/org/apache/lucene/search/comparators/TermOrdValComparator.java
```diff
@@ -524,6 +524,7 @@ public int advance(int target) throws IOException {
 
     @Override
     public void intoBitSet(int upTo, FixedBitSet bitSet, int offset) throws IOException {
+      upTo = Math.min(upTo, maxDoc);
```
benwtrent, line 527: Similar protection that Chris found. Its possible that iteration past `max` might break things in unexpected ways. Let's protect it.

ChrisHegarty: ++

lucene/core/src/java/org/apache/lucene/search/comparators/NumericComparator.java
ChrisHegarty: ++

lucene/core/src/java/org/apache/lucene/search/DenseConjunctionBulkScorer.java
```diff
@@ -236,10 +236,13 @@ private void scoreWindowUsingBitSet(
...
       DocIdSetIterator lead = iterators.get(0);
+      int intoBitSetBase = windowBase;
       if (lead.docID() < windowBase) {
-        lead.advance(windowBase);
+        intoBitSetBase = lead.advance(windowBase);
+      }
+      if (intoBitSetBase < windowMax) {
+        lead.intoBitSet(windowMax, windowMatches, intoBitSetBase);
       }
-      lead.intoBitSet(windowMax, windowMatches, windowBase);
```
ChrisHegarty, line 242: ++

ChrisHegarty (approved):
LGTM

lucene/core/src/java/org/apache/lucene/search/DenseConjunctionBulkScorer.java
```diff
@@ -236,10 +236,13 @@ private void scoreWindowUsingBitSet(
       windowMatches.set(0, windowMax - windowBase);
     } else {
       DocIdSetIterator lead = iterators.get(0);
+      int intoBitSetBase = windowBase;
       if (lead.docID() < windowBase) {
-        lead.advance(windowBase);
+        intoBitSetBase = lead.advance(windowBase);
+      }
+      if (intoBitSetBase < windowMax) {
+        lead.intoBitSet(windowMax, windowMatches, intoBitSetBase);
```
gf2121, line 244: This does not seem correct to me: lead clause and other clause are intersecting with different offsets?

lucene/core/src/java/org/apache/lucene/search/DisjunctionDISIApproximation.java
```diff
@@ -146,7 +146,13 @@ public int advance(int target) throws IOException {
   @Override
   public void intoBitSet(int upTo, FixedBitSet bitSet, int offset) throws IOException {
     while (leadTop.doc < upTo) {
-      leadTop.approximation.intoBitSet(upTo, bitSet, offset);
+      int intoBitSetBase = offset;
+      if (leadTop.approximation.docID() < offset) {
+        intoBitSetBase = leadTop.approximation.advance(offset);
+      }
+      if (intoBitSetBase < upTo) {
+        leadTop.approximation.intoBitSet(upTo, bitSet, intoBitSetBase);
```
gf2121, line 154: Likewise here, i do not think we should change the offset? it makes the conjunction/disjunction meaningless.

lucene/core/src/java/org/apache/lucene/search/DocIdSetIterator.java
```diff
@@ -203,7 +203,8 @@ protected final int slowAdvance(int target) throws IOException {
    * @lucene.internal
    */
   public void intoBitSet(int upTo, FixedBitSet bitSet, int offset) throws IOException {
-    assert offset <= docID();
+    assert offset <= docID() && offset <= upTo
```
gf2121, line 206: Do we really need to assert offset <= upTo? Like we pass upTo=0 in, we just do nothing and return, which seems correct?

lucene/core/src/java/org/apache/lucene/search/DisjunctionDISIApproximation.java
benwtrent: Maybe I misunderstand how `intoBitSet` works, but wouldn't it always advance to the first available doc anyways, not setting the bits of things that it doesn't actually have?

lucene/core/src/java/org/apache/lucene/search/DenseConjunctionBulkScorer.java
benwtrent: Maybe I misunderstand how intoBitSet works, but wouldn't it always advance to the first available doc anyways, not setting the bits of things that it doesn't actually have?

lucene/core/src/java/org/apache/lucene/search/DocIdSetIterator.java
benwtrent: You would say its valid to provide `offset` that is strictly less than `upTo`?

lucene/core/src/java/org/apache/lucene/search/DenseConjunctionBulkScorer.java
gf2121: If there are 2 iterators intersecting, like:

```
lead:  10000, 10001, 10002
other: 10000, 10002, 10003
```

Then we want to load them into a 4-bit bitset. We should use the same offset for them, like 10000:
```
lead bitset:  0, 1, 2
other bitset: 0, 2, 3
```

So that they can do conjunction.

Your change is use `intoBitSetBase` as the offset of lead, but `windowBase` as the offset of others, so i do not think they can intersect properly?

lucene/core/src/java/org/apache/lucene/search/DocIdSetIterator.java
gf2121: I think it is valid to provide offset bigger than offset, it would be a no-op method, but nothing broken?

lucene/core/src/java/org/apache/lucene/search/DenseConjunctionBulkScorer.java
benwtrent: @gf2121 Ah, I see, ok, my other change (which adjusts a loop where the same baseline is used for multiple iterators), definitely breaks that. 

I am not sure the change here actually breaks that. `intoBitSet` will only set bits for docs that match the iterator, so adjusting `offset` to the first valid value for a given iterator seems ok.

But I agree, this shouldn't change the offset used for ALL iterators in a given query. Let me see about the other change I made (in disjunction).

gf2121: > I am not sure the change here actually breaks that. intoBitSet will only set bits for docs that match the iterator, so adjusting offset to the first valid value for a given iterator seems ok.

I do not think its ok? The offset here it is not where the iterator starts iteration, but a number docs need to minus.

lucene/core/src/java/org/apache/lucene/search/DocIdSetIterator.java
benwtrent:
```
2> abr. 18, 2025 7:45:04 DA TARDE com.carrotsearch.randomizedtesting.RandomizedRunner$QueueUncaughtExceptionsHandler uncaughtException
  2> WARNING: Uncaught exception in thread: Thread[#786,Thread-718,5,TGRP-TestLucene99PostingsFormat]
  2> java.lang.RuntimeException: java.lang.AssertionError: offset=2147483133 docID()=2147483647 upTo=-2147483636
  2> 	at __randomizedtesting.SeedInfo.seed([27691DDFD301323F]:0)
  2> 	at org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.index.RandomPostingsTester$TestThread.run(RandomPostingsTester.java:1448)
  2> Caused by: java.lang.AssertionError: offset=2147483133 docID()=2147483647 upTo=-2147483636
  2> 	at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.search.DocIdSetIterator.intoBitSet(DocIdSetIterator.java:207)
  2> 	at org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.index.RandomPostingsTester.verifyEnum(RandomPostingsTester.java:1399)
  2> 	at org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.index.RandomPostingsTester.testTermsOneThread(RandomPostingsTester.java:1605)
  2> 	at org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.index.RandomPostingsTester$TestThread.run(RandomPostingsTester.java:1445)
  2> 
```

This is some of the test failures. That seems weird to me. Isn't this obvious input garbage indicating some iteration failures?

lucene/core/src/java/org/apache/lucene/search/DenseConjunctionBulkScorer.java
gf2121: Maybe my english is too poor to explaining this, but i think the default impl of `intoBitset` is clear enough? You can see the `doc - offset` there.

https://github.com/apache/lucene/blob/4ac2d145acb150251001b6f8b454bb91b5e18f48/lucene/core/src/java/org/apache/lucene/search/DocIdSetIterator.java#L205-L210

benwtrent: > The offset here is not where the iterator starts iteration, but a number docs need to minus.

Gotcha! I will revert. I better understand now.

benwtrent: I still don't fully understand why we advance iterators to `offset`, but then expect the bit set to be set from that given offset.

The weird bugs we are seeing are all about the interaction between iterators position and the offset requested (see: https://github.com/apache/lucene/issues/14517)

gf2121: > I still don't fully understand why we advance iterators to offset, but then expect the bit set to be set from that given offset.

We advance offset to make sure `doc - offset` is positive, which is required by Bitset.

> The weird bugs we are seeing are all about the interaction between iterators position and the offset requested (see: https://github.com/apache/lucene/issues/14517)

Yes, I saw this when you asked this question. I stared at the code for hours and still don't understand it. It looks like the competitiveIterator has advanced to `windowBase`, but in the next iteration, it encounters a document that is smaller than `windowBase`, which should not happen.

lucene/core/src/java/org/apache/lucene/search/comparators/NumericComparator.java
```diff
@@ -434,6 +434,9 @@ public void intoBitSet(int upTo, FixedBitSet bitSet, int offset) throws IOExcept
           if (competitiveIterator.docID() < doc) {
```
benwtrent, line 434: This part is strange. All other `intoBitSet` impls that I can find call the competitiveIterator to advance to at least `offset`, not to some internal `doc`. I suppose NumericComparator is unique?

lucene/core/src/java/org/apache/lucene/search/DenseConjunctionBulkScorer.java
benwtrent: @gf2121 thank you for your patience, I am a completely ignorant when it comes to this section of the code. 

I reverted many of these offset changes. I did add one more check to numeric comparator and did a (hopefully) redundant advance on `TermOrdValComparator.CompetitiveComparator#intoBitSet` for the `docsWithField`.

lucene/core/src/java/org/apache/lucene/search/DisjunctionDISIApproximation.java
```diff
@@ -146,6 +146,9 @@ public int advance(int target) throws IOException {
   @Override
   public void intoBitSet(int upTo, FixedBitSet bitSet, int offset) throws IOException {
     while (leadTop.doc < upTo) {
+      if (leadTop.approximation.docID() < offset) {
+        leadTop.approximation.advance(offset);
+      }
```
benwtrent, line 151: This follows the pattern laid out everywhere else. It seemed weird to me that we are not advancing to atleast `offset` here. Is this wrong?

@gf2121 what do you think?

gf2121: I think we did not need this because the caller of `DisjunctionDISIApproximation` should make sure `docID >= offset` before calling `intoBitset`. There should be no problem here like the competitiveIterator in `TermOldValComparator` where the doc is inconsistent with the inner structure.

lucene/core/src/java/org/apache/lucene/search/comparators/TermOrdValComparator.java
```diff
@@ -524,17 +524,21 @@ public int advance(int target) throws IOException {
...
       if (upTo <= doc) {
         return;
       }
       // Optimize the case when intersecting the competitive iterator is expensive, which is when it
       // hasn't nailed down a disjunction of competitive terms yet.
       if (disjunction == null) {
         if (docsWithField != null) {
+          // Paranoid check, we need to be absolutely sure that the iterator is at least at offset
+          if (docsWithField.docID() < offset) {
```
gf2121, line 536: +1

lucene/core/src/java/org/apache/lucene/search/DisjunctionDISIApproximation.java
benwtrent: OK! I will revert this piece as well. Just making sure :)

lucene/core/src/java/org/apache/lucene/search/comparators/TermOrdValComparator.java
benwtrent: This is the bug fix, the lazy setting of `docsWithField` was the cause. Thank you for the test case @jpountz :D.

lucene/core/src/test/org/apache/lucene/search/comparators/TestTermOrdValComparator.java
```diff
@@ -0,0 +1,87 @@
...
+        Weight weight =
+            searcher.createWeight(query, ScoreMode.COMPLETE_NO_SCORES, RANDOM_MULTIPLIER);
+        SortField sortField = KeywordField.newSortField("sort", false, SortedSetSelector.Type.MIN);
+        sortField.setMissingValue(SortField.STRING_LAST);
+        Sort sort = new Sort(sortField);
+        Collector collector = new TopFieldCollectorManager(sort, 10, 10).newCollector();
+        LeafCollector leafCollector = collector.getLeafCollector(context);
+        BulkScorer bulkScorer = weight.bulkScorer(context);
+        // We need to split on this specific doc ID so that the current doc of the competitive
+        // iterator
```
gf2121, line 78: Nit: maybe format this single-line word better?

gf2121 (approved):
Thank you

lucene/core/src/java/org/apache/lucene/search/comparators/TermOrdValComparator.java
```diff
@@ -524,17 +524,21 @@ public int advance(int target) throws IOException {
...
+      upTo = Math.min(upTo, maxDoc);
       if (upTo <= doc) {
         return;
       }
       // Optimize the case when intersecting the competitive iterator is expensive, which is when it
       // hasn't nailed down a disjunction of competitive terms yet.
       if (disjunction == null) {
         if (docsWithField != null) {
+          // we need to be absolutely sure that the iterator is at least at offset
+          if (docsWithField.docID() < offset) {
```
jpountz, line 536: The contract from `DocIdSetIterator#intoBitSet` requires starting from the current doc rather than the first doc on or after `offset`, so it would be more correct to advance to `doc` here. In practice, it wouldn't make a difference today since all call sites first advance iterators to `offset` before calling `intoBitSet`, but it may not always be the case.

gf2121: > The contract from DocIdSetIterator#intoBitSet requires starting from the current doc rather than the first doc on or after offset, so it would be more correct to advance to doc here. 

So things might got broken when current doc is after the first doc like

```
CompetitiveIterator competitiveIterator = CompetitiveIterator.of(1, 2, 3, 4);
competitiveIterator.advance(2); // advance current doc to 2
competitiveIterator.update(0, 2000); // update to doc value
final int offset = 1;
if (competitiveIterator.docID() < offset) {
  competitiveIterator.advance(offset);
}
competitiveIterator.intoBitSet(upTo, bitset, offset); // we should start from 2 instead of 1.
```

? Nice catch!

jpountz: Yes, exactly.

PR #13472, opened 2024-06-07
TaskExecutor should not fork unnecessarily

original-brownbear:
When executing N tasks and waiting on the result of all of them, we should only fork N - 1 times and execute one task on the calling thread that is getting blocked anyway. This saves at least one context switch, removes the need for any reentrancy protection, and makes better use of available CPU resources.

PR #13861, opened 2024-10-05
Reduce TaskExecutor overhead

original-brownbear:
The `TaskGroup` class is redundant, the futures list can be a local variable shared by the tasks (this also removes any need for making it read-only). Also, the tasks can be made a little more lightweight too by using a var handle instead of an atomic and being explicit about the fact that they need a reference to the list of other tasks.

Similar to #13860 a small but not insignificant improvement here + reducing the noise in this code now sets up cleaner benchmarking for potential solutions to the small task overhead problems we ran into here and there:

```
                            TaskQPS baseline      StdDevQPS my_modified_version      StdDev                Pct diff p-value
           BrowseMonthTaxoFacets       12.82     (23.1%)       12.65     (29.1%)   -1.3% ( -43% -   66%) 0.819
                         LowTerm      846.91      (5.6%)      846.92      (5.9%)    0.0% ( -10% -   12%) 0.999
               HighTermMonthSort     1525.55      (5.2%)     1529.85      (6.6%)    0.3% ( -10% -   12%) 0.832
                      OrHighHigh      193.59      (9.9%)      194.63      (5.3%)    0.5% ( -13% -   17%) 0.762
                     AndHighHigh      135.09      (9.6%)      135.88      (7.1%)    0.6% ( -14% -   19%) 0.756
                    OrNotHighLow     1787.14      (5.4%)     1798.09      (4.5%)    0.6% (  -8% -   11%) 0.581
                       OrHighLow      714.69      (3.6%)      720.59      (3.9%)    0.8% (  -6% -    8%) 0.325
                        HighTerm      439.29      (9.9%)      443.35      (9.2%)    0.9% ( -16% -   22%) 0.665
            BrowseDateSSDVFacets        1.24      (6.9%)        1.26      (9.2%)    1.0% ( -14% -   18%) 0.585
                         MedTerm      582.73      (7.8%)      588.53      (8.6%)    1.0% ( -14% -   18%) 0.586
                    OrHighNotLow      517.51      (7.3%)      524.34      (5.7%)    1.3% ( -10% -   15%) 0.366
                         Prefix3     1737.96      (5.1%)     1761.79      (5.8%)    1.4% (  -9% -   12%) 0.260
                       MedPhrase      121.93      (8.9%)      123.89      (7.2%)    1.6% ( -13% -   19%) 0.374
                   OrHighNotHigh      601.29      (6.1%)      611.36      (5.1%)    1.7% (  -8% -   13%) 0.180
[...]

PR #14524, opened 2025-04-18
Make task executor non-final

Shibi-bala:
### Description

The new task executor implementation from https://github.com/apache/lucene/pull/13472 means users can't exactly control the parallelism of the tasks being run. I think it is fair to allow task executor to be overridden in cases where users want the old or slightly different behavior.

One example of this change breaking consumers: if I had an executor service that I wanted to use to fairly control the execution of tasks across various queries, the calling thread would bypass this fair task handling completely.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/TaskExecutor.java
+++ b/lucene/core/src/java/org/apache/lucene/search/TaskExecutor.java
@@ -44,7 +44,7 @@ import org.apache.lucene.util.ThreadInterruptedException;
  *
  * @lucene.experimental
  */
-public final class TaskExecutor {
+public class TaskExecutor {
   private final Executor executor;
 
   /**


jainankitk:
While I am not sure about changing to non-final, I am wondering if we should execute task on the current thread? Not sure if we save too much overhead and that makes code less readable.

Although, that was the case even before https://github.com/apache/lucene/pull/13861. So, I am wondering how did that change break the fair task handling?

Shibi-bala:
@jainankitk ah linked the wrong PR. It was https://github.com/apache/lucene/pull/13472 that made this change, and it seems like there were considerable speedups. Though it comes at the cost of making fair task handling very hard to control.

Shibi-bala:
This really only impacts Knn queries where the work is done in rewrite and indexSearcher.getTaskExecutor() is called from the knn query. For all other queries, the caller can control what leaves are provided in indexSearcher.search(...) to control the parallelism. The current workaround is the create a new Knn query that submits tasks into a custom "task executor" instead of using the one provided by indexSearcher

jpountz:
IMO Lucene should own how queries execute concurrently instead of making it pluggable. So I'd rather not allow users to pass a custom `TaskExecutor`.

javanna:
Having introduced TaskExecutor in the first place and worked quite a bit on inter-segment concurrency and other ways to parallelize execution, the general idea was to make concurrency as transparent as possible. If we did mean to make TaskExecutor pluggable, it would not be just a matter of making it non final, but we would rather need to figure out what users need that is not currently available, and how would they plug in their custom behaviour. Could we start from something along these lines? How would you like execution to be parallelized in your example?

Shibi-bala:
@javanna I think what users need (or expect) with a TaskExecutor is to be able to have the created tasks run on the ExecutorService provided by the user. In my example, I'd like to be able to have tasks submitted to the ExecutorService to be run in a "fair" way. Basically, implementing query fairness by grouping tasks by originating user and estimated task cost. So my ideal TaskExecutor ideally just submits tasks and then collects results. The current implementation also would run tasks outside the "fair" ExecutorService which is why I want to plug in my own task executor to give up the optimization of running tasks on the calling thread in exchange for complete control of when and the order tasks are run.

Pluggable TaskExecutor is one solution here. As @jainankitk mentioned `Not sure if we save too much overhead and that makes code less readable.`

javanna:
Thanks for clarifying @Shibi-bala . 

I get a bit confused here with the mention of query fairness. What #13472 does is advance execution on the caller thread, that would otherwise be waiting for the executor service's threads to complete their work, which is wasteful. At the same time it decreases the need for forking which provides better performance.

I find it counter-intuitive that to provide more fairness we would go back to a model where the caller thread waits most of its time, and does a lot more forking than needed. Is there a way to implement the grouping of fairness you have in mind on top of the current TaskExecutor? What I am thinking is that despite it may execute on the caller thread, it does still execute tasks in a certain order.

Shibi-bala:
@javanna  Hey! I kinda disagree with: `general idea was to make concurrency as transparent as possible` since the caller thread is used alongside another exec service. I think trading off slightly better perf in exchange for less transparency from the user pov isn't correct. In the prev case, it was very easy to see how resources were used and control that.

> Is there a way to implement the grouping of fairness you have in mind on top of the current TaskExecutor?
Not that I am aware of, which is why I suggested to make the task executor pluggable.

HUSTERGS:
> @javanna Hey! I kinda disagree with: `general idea was to make concurrency as transparent as possible` since the caller thread is used alongside another exec service. I think trading off slightly better perf in exchange for less transparency from the user pov isn't correct. In the prev case, it was very easy to see how resources were used and control that.
> 
> > Is there a way to implement the grouping of fairness you have in mind on top of the current TaskExecutor?
> 
> Not that I am aware of, which is why I suggested to make the task executor pluggable.


@Shibi-bala I'm wondering whether it is what you want if you put your `searcher.search()` itself directly into the same Executor, and wait for it to complete (although I do admit this might not be best way to achieve your goal):
```java
ExecutorService executor = Executors.newFixedThreadPool(4);
IndexSeacher searcher = new IndexSearcher(indexReader, executor);
executor.submit(() -> searcher.search()).get();
```
Please let me know if I got anything wrong

Issue #14508, opened 2025-04-16
Tone down TestIndexWriterDelete.testDeleteAllRepeated (OOMs sometimes)


dweiss:
### Description

This tests OOMs on nightly runs sometimes - when its directory implementation is picked to be in ram.

![Image](https://github.com/user-attachments/assets/365028a7-237c-426f-aebd-14a046d526a7)

Looking at the test, I wonder if we really want to keep it or if there is any other way to implement it because it writes a gazillion files and takes a looong time.

dweiss:
Originally added in #10657.

dweiss:
@msfroh would you know how to tackle this better than forcing FSDirectory in this test?

uschindler:
> [@msfroh](https://github.com/msfroh) would you know how to tackle this better than forcing FSDirectory in this test?

Policeman Jenkins poor NVME disk!!!!!!! The new server is only online since a few weeks and it already eats up the "percentage_used" smart counter:

```
root@serv1 ~ # nvme smart-log /dev/nvme0
Smart Log for NVME device:nvme0 namespace-id:ffffffff
critical_warning                        : 0
temperature                             : 32 °C (305 K)
available_spare                         : 100%
available_spare_threshold               : 10%
percentage_used                         : 4%
endurance group critical warning summary: 0
Data Units Read                         : 15428184 (7.90 TB)
Data Units Written                      : 46860669 (23.99 TB)
host_read_commands                      : 251666556
host_write_commands                     : 787078812
controller_busy_time                    : 3115
power_cycles                            : 11
power_on_hours                          : 423
unsafe_shutdowns                        : 4
media_errors                            : 0
num_err_log_entries                     : 0
Warning Temperature Time                : 0
Critical Composite Temperature Time     : 0
Temperature Sensor 1           : 32 °C (305 K)
Temperature Sensor 2           : 34 °C (307 K)
Thermal Management T1 Trans Count       : 0
Thermal Management T2 Trans Count       : 0
Thermal Management T1 Total Time        : 0
Thermal Management T2 Total Time        : 0
root@serv1 ~ # nvme smart-log /dev/nvme1
Smart Log for NVME device:nvme1 namespace-id:ffffffff
critical_warning                        : 0
temperature                             : 34 °C (307 K)
available_spare                         : 100%
available_spare_threshold               : 5%
percentage_used                         : 3%
endurance group critical warning summary: 0
Data Units Read                         : 10776282 (5.52 TB)
[...]

PR #14526, opened 2025-04-19
Make TestIndexWriterDelete.testDeleteAllRepeated a monster test and force FSDirectory to prevent OOM.

dweiss:
Workaround for #14508

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/index/TestIndexWriterDelete.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestIndexWriterDelete.java
@@ -42,6 +42,7 @@ import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TermQuery;
 import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.FSDirectory;
 import org.apache.lucene.tests.analysis.MockAnalyzer;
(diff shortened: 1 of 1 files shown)


rmuir (approved):
It should not impact jenkins as it is now labelled `@Monster`. 

To try to address @uschindler concerns about disks, we also have `@SuppressFsync` annotation available in LuceneTestCase, for tests that may need to do tons of commits. real fsyncs are suppressed for a given test class 99% of the time by default, but the annotation can make that 100% of the time.

See logic: https://github.com/apache/lucene/blob/ca6879dff4592411e2180770f160e43517122642/lucene/test-framework/src/java/org/apache/lucene/tests/util/TestRuleTemporaryFilesCleanup.java#L132-L137

This won't help much if the test truly writes tons of actual data that will get synced in background: in such a situation you need tmpfs or nobarrier, or tone down the test :)

rmuir:
fwiw: here is my tmpfs setup, which avoids breaking my SSD. This one is 7 years old now. I've got 5MB of trash from lucene builds after 9 days uptime, plenty of trash from other less-behaved programs:

```fstab
# tmpfs mount with a 3GB limit for /tmp
tmpfs  /tmp  tmpfs  rw,nodev,nosuid,size=3G 0 0
```

```properties
# ~/.gradle/gradle.properties or probably possible via -D or -p or something
# the subdirectory will get created automatically by the build if needed
tests.workDir=/tmp/lucene_gradle
```

Issue #14521, opened 2025-04-18
Speed up soft delete


gf2121:
### Description

Soft deletes consume a lot of CPU when flushing docvalue updates or calculating the `numsToDelete` in `SoftDeleteRetentionMergePolicy`. I was looking for some way to speed up these operations. The new `DocIdSetIterator#intoBitset` interface seems to provide a good approach, which is as follows:

- [ ] Implement `intoBItset` of IndexedDISI.

- [ ] Implement `intoBitset` of `MergedDocValues` and `SingleValueNumericDocValuesFieldUpdates#iterator`, and call intoBitset in `IndexedDISI#writeBitSet`.
- [ ] Count numsToDelete with `intoBitset`, `Bits#applyMask` and `popCnt`.

Another optimization I'm looking for is to expose the fact that soft deleted fields always use a single value, so that we can avoid having to go through the calculations to calculate the min/max/gcd. My current idea of API designing is pretty simple, but I'm not sure if it's good enough.

```
public abstract class NumericDocValues extends DocValuesIterator {

  /**
   * If the impl knows all docs have the same value, return the value, otherwise null. 
   */
  public abstract Long singleValue();

  ...

}
```

PR #14529, opened 2025-04-21
Impl intoBitset for IndexedDISI and Docvalues

gf2121:
Implement `intoBitset` for `IndexedDISI` and Docvalues.

`intoBitset` of Docvalues has already been called in competitive iterators, and can also be used to speed up soft delete operations.

Logic of this impl is different from default impl when calling `intoBitset` after `advanceExact` return false, where default impl starts from a not-existing doc but this impl starts from the next doc. I do not know which one is intended, maybe this should just be undefined.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
@@ -811,8 +811,10 @@ public final class IndexedDISI extends AbstractDocIdSetIterator {
      * method returns true if there are remaining docs (gte upTo) in the block, otherwise false.
      * When false return, fp of {@link IndexedDISI#slice} is at {@link IndexedDISI#blockEnd} and
      * {@link IndexedDISI#index} is correct but other status vars are undefined. Caller should
-     * decode the header of next block by {@link #readBlockHeader()} and make sure {@link
-     * IndexedDISI#doc} greater than or equals to {@link IndexedDISI#block}.
+     * decode the header of next block by {@link #readBlockHeader()}.
+     *
+     * <p>Caller need to make sure {@link IndexedDISI#doc} greater than or equals to {@link
+     * IndexedDISI#block} when calling this.
      */
     abstract boolean intoBitSetWithinBlock(
         IndexedDISI disi, int upTo, FixedBitSet bitSet, int offset) throws IOException;
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
@@ -590,7 +590,6 @@ public final class IndexedDISI extends AbstractDocIdSetIterator {
 
   enum Method {
     SPARSE {
-      private static final int BATCH_SIZE = 128;
 
       @Override
       boolean advanceWithinBlock(IndexedDISI disi, int target) throws IOException {
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
@@ -733,10 +733,10 @@ public final class IndexedDISI extends AbstractDocIdSetIterator {
 
         int sourceFrom = disi.doc & 0xFFFF;
         int sourceTo = Math.min(upTo - disi.block, BLOCK_SIZE);
-        int destFrom = disi.block - offset + sourceFrom;
+        int destFrom = disi.doc - offset;
 
         long fp = disi.slice.getFilePointer();
-        disi.slice.seek(fp - Long.BYTES);
+        disi.slice.seek(fp - Long.BYTES); // seek back a long to include current word (disi.word).
         int numWords = FixedBitSet.bits2words(sourceTo) - disi.wordIndex;
         disi.slice.readLongs(disi.bitSet.getBits(), disi.wordIndex, numWords);
         FixedBitSet.orRange(disi.bitSet, sourceFrom, bitSet, destFrom, sourceTo - sourceFrom);
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
@@ -590,7 +590,6 @@ public final class IndexedDISI extends AbstractDocIdSetIterator {
 
   enum Method {
     SPARSE {
-
       @Override
       boolean advanceWithinBlock(IndexedDISI disi, int target) throws IOException {
         final int targetInBlock = target & 0xFFFF;
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
@@ -804,37 +804,37 @@ public final class IndexedDISI extends AbstractDocIdSetIterator {
      */
     abstract boolean intoBitSetWithinBlock(
         IndexedDISI disi, int upTo, FixedBitSet bitSet, int offset) throws IOException;
+  }
 
-    /**
-     * If the distance between the current position and the target is > 8 words, the rank cache will
-     * be used to guarantee a worst-case of 1 rank-lookup and 7 word-read-and-count-bits operations.
-     * Note: This does not guarantee a skip up to target, only up to nearest rank boundary. It is
-     * the responsibility of the caller to iterate further to reach target.
-     *
-     * @param disi standard DISI.
-     * @param targetInBlock lower 16 bits of the target
-     * @throws IOException if a DISI seek failed.
-     */
-    private static void rankSkip(IndexedDISI disi, int targetInBlock) throws IOException {
-      assert disi.denseRankPower >= 0 : disi.denseRankPower;
-      // Resolve the rank as close to targetInBlock as possible (maximum distance is 8 longs)
-      // Note: rankOrigoOffset is tracked on block open, so it is absolute (e.g. don't add origo)
-      final int rankIndex =
-          targetInBlock >> disi.denseRankPower; // Default is 9 (8 longs: 2^3 * 2^6 = 512 docIDs)
+  /**
+   * If the distance between the current position and the target is > 8 words, the rank cache will
+   * be used to guarantee a worst-case of 1 rank-lookup and 7 word-read-and-count-bits operations.
+   * Note: This does not guarantee a skip up to target, only up to nearest rank boundary. It is the
+   * responsibility of the caller to iterate further to reach target.
+   *
+   * @param disi standard DISI.
+   * @param targetInBlock lower 16 bits of the target
+   * @throws IOException if a DISI seek failed.
+   */
+  private static void rankSkip(IndexedDISI disi, int targetInBlock) throws IOException {
+    assert disi.denseRankPower >= 0 : disi.denseRankPower;
+    // Resolve the rank as close to targetInBlock as possible (maximum distance is 8 longs)
+    // Note: rankOrigoOffset is tracked on block open, so it is absolute (e.g. don't add origo)
+    final int rankIndex =
+        targetInBlock >> disi.denseRankPower; // Default is 9 (8 longs: 2^3 * 2^6 = 512 docIDs)
 
-      final int rank =
-          (disi.denseRankTable[rankIndex << 1] & 0xFF) << 8
-              | (disi.denseRankTable[(rankIndex << 1) + 1] & 0xFF);
+    final int rank =
+        (disi.denseRankTable[rankIndex << 1] & 0xFF) << 8
+            | (disi.denseRankTable[(rankIndex << 1) + 1] & 0xFF);
 
-      // Position the counting logic just after the rank point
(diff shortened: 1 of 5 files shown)


jpountz:
To keep things as simple as possible, I'm wondering about:
 - Using a very simple/robust impl in the `SPARSE` case that moves to the next doc in a loop until either the end of block or `upTo`, since it should never be a bottleneck (especially as `DenseConjunctionBulkScorer` is only used on dense iterators)?
 - Refactoring `DENSE` to load doc IDs into an in-memory bit set all the time, so that `intoBitSet` could directly reuse `FixedBitSet.orRange`.

gf2121:
Thanks for feedback,

> Using a very simple/robust impl in the SPARSE case that moves to the next doc in a loop until either the end of block or upTo, since it should never be a bottleneck (especially as DenseConjunctionBulkScorer is only used on dense iterators)?

I won't assert `intoBitset` will only be called by `DenseConjunctionBulkScorer`. I was also thinking about using this for soft delete (https://github.com/apache/lucene/issues/14521) as well, `IndexedDISI#writeBitset` will always load the whole block of docs into a bitset. 

But I'm not sure how much performance this patch gains, i agree to keep it simple as a first step.

> Refactoring DENSE to load doc IDs into an in-memory bit set all the time, so that intoBitSet could directly reuse FixedBitSet.orRange.

I'm a bit confused on this, current patch has already been loading into a in-memory bitset and reuse `orRange`, could you be more specific?

gf2121:
I'm not sure, would this cause heavy overreading in scenarios where only advance/advanceExact is required a few times in this block?  E.g. we saw aggregation on hundreds of metrics with small total hits, which could be the case harmed by the overreading.

gf2121:
I personally prefer not to eagerly load docs for advance/advanceExact, or maybe leave it to another issue/PR. 

I update code to one of the version during my local iteration, which reduces the complexity by using a single `orRange` instead of multiple round of `orRange`. I wonder if this will make sense to you.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
```diff
@@ -491,6 +492,14 @@ public int advance(int target) throws IOException {
     return doc;
   }
 
+  @Override
+  public void intoBitSet(int upTo, FixedBitSet bitSet, int offset) throws IOException {
+    assert doc >= offset;
```
jpountz, line 497: Can you also assert something like `doc == NOT_MORE_DOCS || advanceExact(doc)`? It shouldn't be legal to call intoBitSet after advanceExact returns false (or maybe a better place for this would be asserting doc values).

It looks like some implementations of `intoBitsetWithinBlock` try to cover the case when the current doc doesn't exist, maybe they can be simplified?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
```diff
@@ -535,6 +544,7 @@ private void readBlockHeader() throws IOException {
       method = Method.SPARSE;
       blockEnd = slice.getFilePointer() + (numValues << 1);
       nextExistDocInBlock = -1;
+      exists = false;
```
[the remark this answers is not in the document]
gf2121: This is to make `intoBitset` work with `advanceExact`, which is not really needed. I removed this.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
```diff
@@ -625,8 +635,36 @@ boolean advanceExactWithinBlock(IndexedDISI disi, int target) throws IOException
...
+      boolean intoBitsetWithinBlock(IndexedDISI disi, int upTo, FixedBitSet bitSet, int offset)
+          throws IOException {
+        if (disi.exists) {
+          if (disi.doc >= upTo) {
+            return true;
+          }
+          bitSet.set(disi.doc - offset);
+        }
+
+        for (int i = disi.index, to = disi.nextBlockIndex; i < to; i++) {
```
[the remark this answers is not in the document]
gf2121: I changed this to make it clearer.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
gf2121: > (or maybe a better place for this would be asserting doc values

+1

> It looks like some implementations of intoBitsetWithinBlock try to cover the case when the current doc doesn't exist, maybe they can be simplified?

Done :)

lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
```diff
@@ -625,6 +634,29 @@ boolean advanceExactWithinBlock(IndexedDISI disi, int target) throws IOException
...
+
+      @Override
+      boolean intoBitsetWithinBlock(IndexedDISI disi, int upTo, FixedBitSet bitSet, int offset)
+          throws IOException {
+        if (disi.doc >= upTo) {
+          return true;
+        }
+        bitSet.set(disi.doc - offset);
+
+        for (int i = disi.index + 1, to = disi.nextBlockIndex; i <= to; i++) {
```
jpountz, line 646: I'm still a bit confused, why is `disi.nextBlockIndex` included in the loop when it's excluded in `advanceWithinBlock`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
```diff
@@ -491,6 +492,14 @@ public int advance(int target) throws IOException {
     return doc;
   }
 
+  @Override
+  public void intoBitSet(int upTo, FixedBitSet bitSet, int offset) throws IOException {
+    assert doc >= offset;
+    while (method.intoBitsetWithinBlock(this, upTo, bitSet, offset) == false) {
+      readBlockHeader();
```
jpountz, line 499: I wonder if we should call something like below after reading the block header (similar to what `advance(target)` does):

```
boolean found = method.advanceWithinBlock(this, block);
assert found;
```
This would guarantee that `disi.doc` is always a doc ID of the block when `intoBitsetWithinBlock` is called, which could help simplify some implementations of `intoBitsetWithinBlock`? (I'm looking at `DENSE` in particular)

lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
```diff
@@ -720,6 +805,15 @@ boolean advanceExactWithinBlock(IndexedDISI disi, int target) {
...
      */
     abstract boolean advanceExactWithinBlock(IndexedDISI disi, int target) throws IOException;
+
+    /**
+     * Similar to {@link DocIdSetIterator#intoBitSet}, load docs in this block into a bitset. This
+     * mehtod return true if there are remaining docs (gte upTo) in the block, otherwise false. When
+     * false return, fp of disi#slice is at blockEnd and disi#index is correct but other status vars
+     * are undefined. Caller should decode the header of next block by {@link #readBlockHeader()}.
+     */
+    abstract boolean intoBitsetWithinBlock(
```
jpountz, line 815: Nit: for consistency with how it's called on `DocIdSetIterator`.

```suggestion
    abstract boolean intoBitSetWithinBlock(
```

lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
gf2121: In `advanceWithinBlock`, `disi.index++` is executed in the loop body before `if (doc >= targetInBlock)` so it is not actually excluded. `advanceWithinBlock` can return true when `disi.index` equals `disi.disi.nextBlockIndex`

lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
gf2121: I like the simplification!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
gf2121: I update to the pattern like `advanceWithinBlock`

lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
```diff
@@ -693,6 +727,44 @@ boolean advanceExactWithinBlock(IndexedDISI disi, int target) throws IOException
         disi.index = disi.numberOfOnes - Long.bitCount(leftBits);
         return (leftBits & 1L) != 0;
       }
+
+      @Override
+      boolean intoBitSetWithinBlock(IndexedDISI disi, int upTo, FixedBitSet bitSet, int offset)
+          throws IOException {
+        if (disi.doc >= upTo) {
+          advanceWithinBlock(disi, disi.doc);
```
jpountz, line 735: Do we really need to call `advanceWithinBlock`? If `disi.doc` is already `>= upTo`, there shouldn't be anything to do?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
```diff
@@ -693,6 +727,44 @@ boolean advanceExactWithinBlock(IndexedDISI disi, int target) throws IOException
...
+          return true;
+        }
+
+        int sourceFrom = disi.doc & 0xFFFF;
+        int sourceStartWord = sourceFrom >>> 6;
+        int sourceTo = Math.min(upTo - disi.block, BLOCK_SIZE);
+        int numWords = FixedBitSet.bits2words(sourceTo) - sourceStartWord;
+
+        if (disi.bitSet == null || disi.bitSet.getBits().length < numWords) {
```
jpountz, line 744: Should we always create a bit set of size `BLOCK_SIZE` when it's null for simplicity? This is what users will get in the worst-case scenario anyway, and it's not crazy (65k bits = 8kB)?

lucene/core/src/test/org/apache/lucene/codecs/lucene90/TestIndexedDISI.java
```diff
@@ -555,6 +565,47 @@ private void assertAdvanceExactRandomized(
...
+      BitSetIterator actual = new BitSetIterator(set1, set1.cardinality());
+      for (int expectedDoc = expected.nextDoc();
+          expectedDoc != DocIdSetIterator.NO_MORE_DOCS;
+          expectedDoc = expected.nextDoc()) {
+        int actualDoc = actual.nextDoc();
+        assertEquals(expectedDoc + offset, actualDoc + offset); // plus offset for better message.
+      }
+      assertEquals(DocIdSetIterator.NO_MORE_DOCS, actual.nextDoc());
+      set1.clear();
+      set2.clear();
```
jpountz, line 605: Can you also check that nextDoc() returns the expected value after `intoBitSet` returns? To make sure that all required state is set correctly?

jpountz (approved):
I left some minor comments, other than that it looks good to me!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
gf2121: This was useful before we advancing `disi.block`, but not now. I like the simplification more :)

lucene/core/src/test/org/apache/lucene/codecs/lucene90/TestIndexedDISI.java
gf2121: This helps me find a bug: `upTo` could be a number in last block when `RANGE` do `intoBitSetWithinBlock`, i'm missing a `doc > upTo` check there.

lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
```diff
@@ -693,6 +723,34 @@ boolean advanceExactWithinBlock(IndexedDISI disi, int target) throws IOException
...
+      @Override
+      boolean intoBitSetWithinBlock(IndexedDISI disi, int upTo, FixedBitSet bitSet, int offset)
+          throws IOException {
+        if (disi.bitSet == null) {
+          disi.bitSet = new FixedBitSet(BLOCK_SIZE);
+        }
+
+        int sourceFrom = disi.doc & 0xFFFF;
+        int sourceTo = Math.min(upTo - disi.block, BLOCK_SIZE);
+        int destFrom = disi.block - offset + sourceFrom;
```
jpountz, line 736: Is this the same as `disi.doc - offset`?

lucene/core/src/java/org/apache/lucene/codecs/lucene90/IndexedDISI.java
```diff
@@ -693,6 +723,34 @@ boolean advanceExactWithinBlock(IndexedDISI disi, int target) throws IOException
...
+
+        long fp = disi.slice.getFilePointer();
+        disi.slice.seek(fp - Long.BYTES);
```
jpountz, line 739: nit: maybe leave a small comment explaining why we're seeking backwards, I believe that it's to include the word that contains the current doc, right?

jpountz (approved):

jpountz (approved):

PR #14530, opened 2025-04-21
Advance doc instead of offset in competitive iterator

gf2121:
https://github.com/apache/lucene/pull/14523#discussion_r2051658845

Should we include this in 10.2.1?


lucene/core/src/java/org/apache/lucene/search/comparators/TermOrdValComparator.java
```diff
@@ -533,8 +533,8 @@ public void intoBitSet(int upTo, FixedBitSet bitSet, int offset) throws IOExcept
       if (disjunction == null) {
         if (docsWithField != null) {
           // we need to be absolutely sure that the iterator is at least at offset
-          if (docsWithField.docID() < offset) {
-            docsWithField.advance(offset);
+          if (docsWithField.docID() < doc) {
+            docsWithField.advance(doc);
```
benwtrent, line 537: I think this is good. It should also go into 10.2, could you augment the changes entry for this bug fix to include this PR and such?

jpountz (approved):

PR #14531, opened 2025-04-21
speed up numDeletesToMerge of SoftDeletesRetentionMergePolicy

gf2121:
This change helps `SoftDeletesRetentionMergePolicy` to get a chance to take advantage of `DenseConjunctionBulkScorer` to speed up count `numDeletesToMerge`.


jpountz (commented):
I wonder if this `Bits` instance could implement `applyMask`, otherwise I worry that this code wouldn't be efficient?

PR #14532, opened 2025-04-21
Make competitive iterators more robust.

jpountz:
As per a recent bug (#14517), competitive iterators are hard to get right given how their state gets updated in place. This commit tries to make them more robust by extracting the logic of updating the state of a `DocIdSetIterator` to a shared class, which can then be tested on its own.

A side-effect is that it implements `#intoBitSet` on the doc competitive iterator and `#docIDRunEnd` on all competitive iterators.


gf2121 (approved):
I like the simplification,

> A side-effect is that it implements #intoBitSet on the doc competitive iterator and #docIDRunEnd on all competitive iterators.

By 'side-effect', do you mean another gain? I thought it means negative effects :)

jpountz:
> By 'side-effect', do you mean another gain? I thought it means negative effects :)

Hopefully a gain, I still need to run benchmarks. This change also increases polymorphism, so I'm not sure if it will make things better or not.

jpountz:
I could confirm that it's due to `#docIdEnd()` now being implemented on the competitive iterators. This triggers usage of `DISIDocIdStream`, which calls `nextDoc()` in a loop, and calling `nextDoc()` in a loop happens to be a slower way of iterating doc IDs than calling `intoBitSet` and then iterating set bits when blocks are encoded as a bit set.

PR #14533, opened 2025-04-22
Upgrade to Gradle 8.14 and Java 24 minimum

### LINKED ISSUE #14379: build support: java 24
### Description

java 23 has disappeared and has been replaced with java 24. 

the build currently requires 23 exactly, which creates a hurdle for users, since it is difficult to get: does not exist in operating system package manager, must be dug out of a graveyard, etc.

yes, i know the problem of gradle here, but that doesn't excuse their behavior. they shouldnt be parsing classfiles of the jvm anyway.

dweiss:
This is an initial set of changes that:
* [x] update gradle to 8.14 (rc2), supporting Java 24,
* [x] bump minimum Java version to Java 24 (#14379)
* [x] upgrade to an ECJ version that supports Java 24 (none yet) and revert the enabled=false from the patch
* [x] upgrade to gradle 8.14 (final)

Showstoppers for now:
* ECJ doesn't have a version that supports Java 24 yet. I disabled ecj checks for now to verify other things work.
* we should probably wait for the final version of gradle 8.14.


rmuir:
> * upgrade to an ECJ version that supports Java 24 (none yet) and revert the enabled=false from the patch

I don't understand this one: their language server works fine with java 24 and i'm able to navigate the new JDK24 classfile api, etc. I guess I haven't used it extensively, due to the gradle issue, and they've got some bugfixes since the last release.

dweiss:
> I don't understand this one: their language server works fine with java 24 and i'm able to navigate the new JDK24 classfile api, etc. I guess I haven't used it extensively, due to the gradle issue, and they've got some bugfixes since the last release.

I don't fully understand it either - the latest artifact for the batch ecj compiler (what we use) seems to be from march:
https://repo1.maven.org/maven2/org/eclipse/jdt/ecj/3.41.0/

this doesn't support Java 24. I've just downloaded the latest Eclipse and it seems to contain the same batch compiler. Their website says this:

image

There are some marketplace extensions to support Java 24 but these seem to be external - so... I don't think Eclipse supports Java 24 yet?

dweiss:
I've no idea either. All I know is that if I download eclipse-jdt, it doesn't have support for Java 24 yet. There is a milestone version which supposedly has it - but not an official stable version, as far as I can see. And the maven repo definitely doesn't have the batch compiler supporting java24. That eclipse-jdls has a fork of jdt... which is also weird.

dweiss:
Eclipse 4.36M1 seems to have Java 24 support in the batch compiler. This is marked as "stable" but artifacts for this release are not on maven central yet.

rmuir:
Yeah I think "big old eclipse" lags behind, whereas the jdtls is "the latest" and sees more usage (e.g. all VSCode users and other editors).

There's even a plugin to allow use of jdtls in "big old eclipse": https://github.com/redhat-developer/eclipseide-jdtls/

very confusing ecosystem, but that's my read on it.

dweiss:
I believe they use p2 repositories for interim releases and only do a maven push once in a quarter for stable final releases. 

If this becomes a problem, we can pull ecj directly from p2 drops, like here:
https://www.eclipse.org/downloads/download.php?file=/eclipse/downloads/drops4/S-4.36M1-202504031800/ecj-4.36M1.jar&r=1

dweiss:
I've replaced ecj batch compiler dependency with a direct download from Eclipse's p2 drop repositories. I don't think anybody will even notice the difference and p2 repositories seem to have a lot eager release cadence compared to maven eclipse artifacts (~3 months).

uschindler:
Please give a short notice to me directly before your merge this to main branch so I can update Jenkins Jobs on Policeman.

uschindler:
I am not really happy about downloading from the non-maven repo, but as this won't affect any release, I am fine. If we release main branch we should rellay switch to an offical Maven artifact for ECJ.

dweiss:
Well, I'm not happy either but their release cadence suggests Java 24 won't be supported in maven ecj until June...

uschindler:
Thats fine, we should just make sure (by comment or wahtever) that we can easily switch back in a documented way to official Maven artifacts before we release Main branch.

lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -365,7 +365,9 @@ private static <A> MMapIndexInputProvider<A> lookupProvider() {
       // access through the lookup:
       final var constr = lookup.findConstructor(cls, MethodType.methodType(void.class, int.class));
       try {
-        return (MMapIndexInputProvider<A>) constr.invoke(maxPermits);
+        @SuppressWarnings("unchecked")
```
uschindler, line 368: LOL, whats the reason for this?

We should anways get rid on the whole code here in main branch and merge all files to the main compilation unit (except vectors, of course).

dweiss: The compiler complains about it. Try reverting and see for yourself.

dweiss: It does look odd, I agree.
```
  error: warnings found and -Werror specified
  /home/.../lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java:368: warning: [unchecked] unchecked cast
          return (MMapIndexInputProvider<A>) constr.invoke(maxPermits);
                                                          ^
    required: MMapIndexInputProvider<A>
    found:    MMapIndexInputProvider
    where A is a type-variable:
      A extends Object declared in method <A>lookupProvider()
  1 error
  1 warning

```

uschindler: I opened https://github.com/apache/lucene/pull/14564 to get rid of all this shit.

At moment the other PR is based on Java 23, but it is easy to update to 24. At moment we forgot to change the code, so Java 21 was still in business in Main!

uschindler: In future, when we update Java version we also need the updates in build system and sourcesets for vectors. So we need to change the mrJarVersions property in Gradle build, too and rename the sourceset folder java23 -> java24

dweiss:
Ok, I've applied this to main. I will backport gradle upgrade selectively to 10x tomorrow (can't be cherry picked automatically).

dweiss:
Ok, I've applied this to main. I will backport gradle upgrade selectively to 10x tomorrow (can't be cherry picked automatically).

dweiss:
Ok, I've applied this to main. I will backport gradle upgrade selectively to 10x tomorrow (can't be cherry picked automatically).

dweiss:
Ok, I've applied this to main. I will backport gradle upgrade selectively to 10x tomorrow (can't be cherry picked automatically).

uschindler:
I disabled Policeman Jenkins Main Jobs, working on Upgrade. Can you take care of ASF Jenkins?

uschindler:
When testing this I figured out the following warning when starting with Java 24:

```
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::load has been called by net.rubygrapefruit.platform.internal.NativeLibraryLoader in an unnamed module (file:/C:/Users/Uwe%20Schindler/.gradle/wrapper/dists/gradle-8.14-bin/38aieal9i53h9rfe7vjup95b9/gradle-8.14/lib/native-platform-0.22-milestone-28.jar)
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled
```

I think we need to change the gradlew / gradlew.cmd files to pass the additional command line option `--enable-native-access=ALL-UNNAMED`

uschindler:
I updated the other PR and generated the APIJAR for 24. For the next update, basically the same linke this commit needs to be done: https://github.com/apache/lucene/pull/14564/commits/2ad382aba9586704166dd85c78ec5e8b2b294679

dweiss:
> I think we need to change the gradlew / gradlew.cmd files to pass the additional command line option --enable-native-access=ALL-UNNAMED

I didn't touch the launcher scripts... could be that the latest from gradle already does it but we have quite a few tweaks in there that'd have to be ported. I'll look into this tomorrow, hopefully.

dweiss:
> I disabled Policeman Jenkins Main Jobs, working on Upgrade. Can you take care of ASF Jenkins?

I've switched all jobs on asf jenkins to Java 24. Hopefully didn't miss anything.

uschindler:
> > I think we need to change the gradlew / gradlew.cmd files to pass the additional command line option --enable-native-access=ALL-UNNAMED
> 
> I didn't touch the launcher scripts... could be that the latest from gradle already does it but we have quite a few tweaks in there that'd have to be ported. I'll look into this tomorrow, hopefully.

I think adding that as command line option should be easy to do. Basically the parameter is available since Java 20 or 21 (since Panama exists), so it should also work with launching from older java versions and it won't fail with error message before ours.

uschindler:
Policeman Jenkins was also updated. I hope the randomization and RUNTIME_JAVA_HOME with 25 works, too. If something breaks, I will keep an eye on it.

uschindler:
I also updated the 10.x release tester job to use Java 24, too. This will affect the next release. Of course Gradle backport is needed.

dweiss:
> I think we need to change the gradlew / gradlew.cmd files to pass the additional command line option `--enable-native-access=ALL-UNNAMED`

gradle is now launched with --jar and contains a manifest entry for this (no need to tweak the scripts). Should be fine with #14592

uschindler:
Fixed: 0550cc6d2d0294641679131b597d591896211fce

uschindler:
One information about Exclipse: With the recent commit all works fine. It just sometimes fails with a compilation error (but that's not new) regarding: "Permitted type LeafReaderContext does not declare org.apache.lucene.index.IndexReaderContext as a direct supertype"

This error is bullshit but something seems to confuse the compile inside the Eclipse IDE.

Anyways: If you want to use exclipse, you need to install the Milestone 1 of 2025-06 release, otherwise it won't work (same reason like discussed above). Luckily on starting Eclipse it asks you if you want to try the milestone release.

dweiss:
Thank you, Uwe.

PR #14543, opened 2025-04-23
Fix leadCost calculation in BooleanScorerSupplier.requiredBulkScorer

### LINKED ISSUE #14542: [Bug] Lead cost in boolean conjunction queries can be miscalculated
### Description

In boolean conjunction queries, iteration is led by the DISI with the lowest `cost()`. To do this we compute the `leadCost` first from the relevant `ScorerSupplier`s, and then get the actual `Scorer`s using `leadCost`. 

In Lucene 10.0 and onwards, there seems to be a bug in `BooleanScorerSupplier.requiredBulkScorer()`. I think the intention is to return the minimum `cost()` of the scorer suppliers for all MUST and FILTER clause, which matches my understanding of how conjunctions should work, and also matches what would happen in [the pre-10.0 code](https://github.com/apache/lucene/blob/main/CONTRIBUTING.md). 

However the [code](https://github.com/apache/lucene/blob/main/CONTRIBUTING.md) now does: 

```
long leadCost =
        subs.get(Occur.MUST).stream().mapToLong(ScorerSupplier::cost).min().orElse(Long.MAX_VALUE);
    leadCost =
        subs.get(Occur.FILTER).stream().mapToLong(ScorerSupplier::cost).min().orElse(leadCost);
```
so that if there is both a MUST and a FILTER clause, leadCost always ends up equaling the minimum FILTER clause's cost, even if this is greater than the minimum MUST cost. 

This can cause a significant performance hit for such queries. See this OpenSearch issue (https://github.com/opensearch-project/OpenSearch/issues/17870) for some more details and some benchmark numbers. 

In affected queries, most of the CPU time is actually spent constructing the BulkScorer, not using it to score documents. This happens because the ScorerSupplier provided by IndexOrDocValuesQuery uses `leadCost` to [choose whether](https://github.com/apache/lucene/blob/a8b503fdf250b9dbd183cf953fead380b3d51d34/lucene/core/src/java/org/apache/lucene/search/IndexOrDocValuesQuery.java#L174) to get the cost from `indexScorerSupplier.get(leadCost)` (in my case from PointRangeQuery), or `dvScorerSupplier.get(leadCost)`. 

When `leadCost` is miscalculated due to the bug, it's higher than it really should have been, and so we erroneously use `indexScorerSupplier.get(leadCost)`. In this case, that requires running a bunch of BKD tree-related code, which slows the query by 40-300% in my tests. 

See this flamegraph for an example:

Image

Tar install of current OpenSearch 3.0, which uses Lucene 10.1, running on an AL2 ec2 instance.

peteralfonsi:
### Description

Fixes https://github.com/apache/lucene/issues/14542 by setting `leadCost` in `BooleanScorerSupplier.requiredBulkScorer` to the minimum of both MUST and FILTER clauses' costs. This bug caused performance regressions in some queries. More details are in the original issue.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/BooleanScorerSupplier.java
+++ b/lucene/core/src/java/org/apache/lucene/search/BooleanScorerSupplier.java
@@ -69,6 +69,14 @@ final class BooleanScorerSupplier extends ScorerSupplier {
     this.maxDoc = maxDoc;
   }
 
+  private long computeShouldCost() {
+    final Collection<ScorerSupplier> optionalScorers = subs.get(Occur.SHOULD);
+    return ScorerUtil.costWithMinShouldMatch(
+        optionalScorers.stream().mapToLong(ScorerSupplier::cost),
+        optionalScorers.size(),
+        minShouldMatch);
+  }
+
   private long computeCost() {
     OptionalLong minRequiredCost =
         Stream.concat(subs.get(Occur.MUST).stream(), subs.get(Occur.FILTER).stream())
@@ -78,11 +86,7 @@ final class BooleanScorerSupplier extends ScorerSupplier {
       return minRequiredCost.getAsLong();
     } else {
       final Collection<ScorerSupplier> optionalScorers = subs.get(Occur.SHOULD);
-      final long shouldCost =
-          ScorerUtil.costWithMinShouldMatch(
-              optionalScorers.stream().mapToLong(ScorerSupplier::cost),
-              optionalScorers.size(),
-              minShouldMatch);
+      final long shouldCost = computeShouldCost();
       return Math.min(minRequiredCost.orElse(Long.MAX_VALUE), shouldCost);
     }
   }
@@ -293,9 +297,10 @@ final class BooleanScorerSupplier extends ScorerSupplier {
       return new MaxScoreBulkScorer(maxDoc, optionalScorers, null);
     }
 
+    long shouldCost = computeShouldCost();
     List<Scorer> optional = new ArrayList<Scorer>();
     for (ScorerSupplier ss : subs.get(Occur.SHOULD)) {
-      optional.add(ss.get(Long.MAX_VALUE));
+      optional.add(ss.get(shouldCost));
     }
 
     return new BooleanScorer(optional, Math.max(1, minShouldMatch), scoreMode.needsScores());
--- a/lucene/core/src/java/org/apache/lucene/search/BooleanScorerSupplier.java
+++ b/lucene/core/src/java/org/apache/lucene/search/BooleanScorerSupplier.java
@@ -359,10 +359,14 @@ final class BooleanScorerSupplier extends ScorerSupplier {
       return scorer;
     }
 
-    long leadCost =
+    long mustLeadCost =
         subs.get(Occur.MUST).stream().mapToLong(ScorerSupplier::cost).min().orElse(Long.MAX_VALUE);
-    leadCost =
-        subs.get(Occur.FILTER).stream().mapToLong(ScorerSupplier::cost).min().orElse(leadCost);
+    long filterLeadCost =
+        subs.get(Occur.FILTER).stream()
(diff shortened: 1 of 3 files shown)


peteralfonsi:
Marking as draft for now, I will add a test case for this soon.

benwtrent:
🤔 seems like this is another candidate for a 10.2.1 bugfix release.

jpountz:
Thanks for catching this! For reference, this was introduced in 10.0 (by me). The change looks good, let me know if you need help writing a test.

jpountz:
@peteralfonsi I pushed tests to your branch so that this change has a chance to make it to 10.2. I hope you don't mind. We already had good tests for `ScorerSupplier#scorer`, I just extended them to cover `ScorerSupplier#bulkScorer` too.

jpountz:
For reference, the new tests found a similar bug with disjunctive queries that configure a minimum number of matching clauses, so I fixed it too.

lucene/core/src/java/org/apache/lucene/search/BooleanScorerSupplier.java
```diff
@@ -78,11 +86,7 @@ private long computeCost() {
       return minRequiredCost.getAsLong();
     } else {
       final Collection<ScorerSupplier> optionalScorers = subs.get(Occur.SHOULD);
```
benwtrent, line 88: THis is unused now. CI is mad :)
```suggestion
```

lucene/core/src/test/org/apache/lucene/search/TestBoolean2ScorerSupplier.java
```diff
@@ -315,6 +318,9 @@ public void testDisjunctionLeadCost() throws IOException {
     new BooleanScorerSupplier(
             new FakeWeight(), subs, RandomPicks.randomFrom(random(), ScoreMode.values()), 0, 100)
         .get(100); // triggers assertions as a side-effect
+    new BooleanScorerSupplier(
+            new FakeWeight(), subs, RandomPicks.randomFrom(random(), ScoreMode.values()), 0, 100)
+        .bulkScorer(); // triggers assertions as a side-effect
```
benwtrent, line 323:
```
TestBoolean2ScorerSupplier > testDisjunctionLeadCost FAILED
    java.lang.AssertionError: FakeLazyScorer(cost=42,leadCost=54) expected:<54> but was:<9223372036854775807>
        at __randomizedtesting.SeedInfo.seed([59321C032B75088A:CB8AA4FC1AF92C0A]:0)
        at org.junit.Assert.assertEquals(Assert.java:647)
        at org.apache.lucene.search.TestBoolean2ScorerSupplier$FakeScorerSupplier.get(TestBoolean2ScorerSupplier.java:108)
        at org.apache.lucene.search.BooleanScorerSupplier.optionalBulkScorer(BooleanScorerSupplier.java:294)
        at org.apache.lucene.search.BooleanScorerSupplier.booleanScorer(BooleanScorerSupplier.java:216)
        at org.apache.lucene.search.BooleanScorerSupplier.bulkScorer(BooleanScorerSupplier.java:178)
        at org.apache.lucene.search.TestBoolean2ScorerSupplier.testDisjunctionLeadCost(TestBoolean2ScorerSupplier.java:323)
```

Is failing?

```
gradlew test --tests TestBoolean2ScorerSupplier.testDisjunctionLeadCost -Dtests.seed=59321C032B75088A -Dtests.locale=yrl-VE -Dtests.timezone=Etc/GMT-14 -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```

benwtrent (approved):

ChrisHegarty (approved):
LGTM.  I'll merge and backport to `branch_10_2`.

peteralfonsi:
@jpountz Thanks for the help with the tests - didn't realize 10.2 was coming soon.

jpountz:
No worries at all, nobody knew it was coming soon until a few days ago. :)

PR #14550, opened 2025-04-24
Remove DISIDocIdStream.

jpountz:
I had initially introduced `DISIDocIdStream` to avoid introducing regressions when `DenseConjunctionBulkScorer` started accepting single clauses. However, benchmarks on #14532 suggested that going through `DISIDocIdStream` is slower than loading docs into a bit set first and then iterating the bit set, when the postings list has many of its blocks encoded as bit sets.

This makes sense, the way how `BitSetDocIdStream` iterates set bits saves a number of operations compared with calling `FixedBitSet#nextSetBit` in a loop.

So I'm suggesting removing `DISIDocIdStream` for now for simplicity.


gf2121 (approved):
Excited to see a combination of simplification and optimization!

ChrisHegarty (approved):
LGTM

PR #14552, opened 2025-04-24
Speed up flush of softdelete by intoBitset

gf2121:
Speed up the flush of softdelete by intoBitset.


gf2121:
I managed to get some numbers on this change:

```
Baseline: Soft delete flush total took: 30311ms, IndexedDISI#writeBitset total took: 8324ms.
Candidate: Soft delete flush total took: 22635ms, IndexedDISI#writeBitset total took: 964ms.
```

BTW, i added some docvalue fields in my benchmark. Then i'm seeing rather huge cost of virtual calls represented by `vtable stub` when computing min/max/gcd.

image

gf2121:
I made some effort to simplify the change, this is ready for review now.

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
```diff
@@ -526,6 +540,60 @@ public int nextDoc() throws IOException {
...
+    public void intoBitSet(int upTo, FixedBitSet bitSet, int offset) throws IOException {
+      if (onDiskDocValues == null) {
+        super.intoBitSet(upTo, bitSet, offset);
+        return;
+      }
+
+      // we need a scratch bitset because the param bitset doesn't allow bits to be cleared.
+      if (scratch == null || scratch.length() < upTo - offset) {
+        // intoBitSet is usually called with fixed window size so we do not do overSize here.
+        scratch = new FixedBitSet(upTo - offset);
```
jpountz, line 554: Maybe use FixedBitSet#ensureCapacity?

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
```diff
@@ -526,6 +540,60 @@ public int nextDoc() throws IOException {
...
+
+    @Override
+    public void intoBitSet(int upTo, FixedBitSet bitSet, int offset) throws IOException {
+      if (onDiskDocValues == null) {
+        super.intoBitSet(upTo, bitSet, offset);
+        return;
+      }
+
+      // we need a scratch bitset because the param bitset doesn't allow bits to be cleared.
+      if (scratch == null || scratch.length() < upTo - offset) {
```
jpountz, line 552: Should we compare scratch.length() with bitSet.length() rather than `upTo - offset`? My concern is that there are multiple call sites that call intoBitSet with upTo=NO_MORE_DOCS, so this could create a giant bit set?

lucene/core/src/java/org/apache/lucene/index/SingletonSortedNumericDocValues.java
```diff
@@ -57,6 +58,16 @@ public int advance(int target) throws IOException {
...
 
+  @Override
+  public void intoBitSet(int upTo, FixedBitSet bitSet, int offset) throws IOException {
+    in.intoBitSet(upTo, bitSet, offset);
+  }
+
+  @Override
+  public int docIDRunEnd() throws IOException {
+    return in.docIDRunEnd();
+  }
```
jpountz, line 69: Maybe extract this change to a separate PR and cover all wrappers?

jpountz (commented):
I left some small comments but the change makes sense to me, nice optimization.

lucene/core/src/java/org/apache/lucene/index/ReadersAndUpdates.java
gf2121: +1, nice catch!

lucene/core/src/java/org/apache/lucene/index/SingletonSortedNumericDocValues.java
gf2121: I filed https://github.com/apache/lucene/pull/14618 to cover `SingletonSortedNumericDocValues` and `SingletonSortedSetDocValues`. I think all these work that implement methods for wrappers could be avoided after https://github.com/apache/lucene/pull/14475.

jpountz (approved):

PR #14559, opened 2025-04-26
Disabling point range collector if doc values not indexed

### LINKED ISSUE #14536: Allow Histogram Collection using PointTree when SortedNumericDocValues is absent
### Description

The HistogramCollector uses `PointTreeBulkCollector` logic only when the PointTree is dense compared to the buckets across which it is getting collected. But for cases that don't have `SortedNumericDocValues` indexed at all, we should allow the collection using PointTree irrespective of how slow it is, instead of throwing `IllegalStateException("Expected numeric field, but got doc-value type: ")`. Related to https://github.com/apache/lucene/pull/14439#discussion_r2054911636

jainankitk:
### Description
Enabling point range collector if doc values are not indexed

Resolves https://github.com/apache/lucene/issues/14536


jainankitk:
> My mental model is that this collector works on doc values in the default case, and can opportunistically take advantage of index statistics or points indexes when it makes things faster. So doc values should be required, and index statistics and point indexes should be optional.

+1, except that it works today even when doc_values are not indexed and that too in some cases depending on the bucket width. Maybe to address both the inconsistencies we have discussed, let us move the below check before using points index for collection?

```
if (fi.getDocValuesType() != DocValuesType.NUMERIC
    && fi.getDocValuesType() != DocValuesType.SORTED_NUMERIC) {
  throw new IllegalStateException(
      "Expected numeric field, but got doc-value type: " + fi.getDocValuesType());
}
```

jainankitk:
Reran the benchmark, just to ensure nothing is regressed, and collector works as expected:

```
  Benchmark                                             (bucketWidth)  (docCount)  (pointEnabled)   Mode  Cnt      Score      Error  Units                               
  HistogramCollectorBenchmark.pointRangeQueryHistogram           5000      500000            true  thrpt    3   6457.918 ± 2480.922  ops/s                               
  HistogramCollectorBenchmark.pointRangeQueryHistogram           5000      500000           false  thrpt    3    119.723 ±   10.479  ops/s                               
  HistogramCollectorBenchmark.pointRangeQueryHistogram           5000     5000000            true  thrpt    3    794.071 ±   78.626  ops/s                               
  HistogramCollectorBenchmark.pointRangeQueryHistogram           5000     5000000           false  thrpt    3     12.416 ±    0.423  ops/s                               
  HistogramCollectorBenchmark.pointRangeQueryHistogram          25000      500000            true  thrpt    3  14587.655 ± 2973.992  ops/s                               
  HistogramCollectorBenchmark.pointRangeQueryHistogram          25000      500000           false  thrpt    3    120.221 ±   10.853  ops/s                               
  HistogramCollectorBenchmark.pointRangeQueryHistogram          25000     5000000            true  thrpt    3   3492.358 ±   92.812  ops/s                               
  HistogramCollectorBenchmark.pointRangeQueryHistogram          25000     5000000           false  thrpt    3     12.418 ±    0.406  ops/s
```

jainankitk:
@jpountz - As per the suggestion, I have changed the PR to disable `HistogramCollection` instead if docValues is not indexed. Please let me know if there is any other feedback.

jpountz (approved):
Thank you!

PR #14560, opened 2025-04-26
Enabling histogram collection for PointRangeQuery

### LINKED ISSUE #14535: Leverage multi range traversal Histogram Collection for PointRangeQuery
### Description

While https://github.com/apache/lucene/pull/14439 adds Histogram Collector optimization for `MATCH_ALL` cases, it does not apply for `PointRangeQuery` filtering out few documents. This issue is for extending the collection logic to account for `PointRangeQuery` cases as well

jainankitk:
### Description
Enables efficient histogram collection using point range query, when querying field is same as the histogram field

Resolves https://github.com/apache/lucene/issues/14535

### Testing
Successfully ran the tests 10k times.

```
> Task :lucene:sandbox:test
:lucene:sandbox:test (SUCCESS): 10000 test(s)

> Task :lucene:sandbox:wipeTaskTemp
The slowest suites (exceeding 1s) during this run:
  109.33s TestHistogramCollectorManager (:lucene:sandbox)

BUILD SUCCESSFUL in 1m 53s
246 actionable tasks: 92 executed, 154 up-to-date
```

Initially they were failing due to first bucket occasionally not having any documents, causing `finalizePreviousBucket` to not move to the next bucket. That also addresses the TODO from earlier
```
java.lang.AssertionError: expected:<[2=>1041, 3=>671, 1=>1007]> but was:<[]>
```

### DIFF
--- a/lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
+++ b/lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
@@ -17,7 +17,10 @@
 package org.apache.lucene.sandbox.facet.plain.histograms;
 
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.ConcurrentMap;
+import java.util.function.Function;
 import org.apache.lucene.index.DocValues;
 import org.apache.lucene.index.DocValuesSkipper;
 import org.apache.lucene.index.DocValuesType;
@@ -27,11 +30,16 @@ import org.apache.lucene.index.NumericDocValues;
 import org.apache.lucene.index.PointValues;
 import org.apache.lucene.index.SortedNumericDocValues;
 import org.apache.lucene.internal.hppc.LongIntHashMap;
+import org.apache.lucene.queries.function.FunctionScoreQuery;
+import org.apache.lucene.search.BoostQuery;
 import org.apache.lucene.search.CollectionTerminatedException;
 import org.apache.lucene.search.Collector;
+import org.apache.lucene.search.ConstantScoreQuery;
 import org.apache.lucene.search.DocIdStream;
+import org.apache.lucene.search.IndexOrDocValuesQuery;
 import org.apache.lucene.search.LeafCollector;
 import org.apache.lucene.search.PointRangeQuery;
+import org.apache.lucene.search.Query;
 import org.apache.lucene.search.Scorable;
 import org.apache.lucene.search.ScoreMode;
 import org.apache.lucene.search.Weight;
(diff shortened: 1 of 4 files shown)


lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
```diff
@@ -330,4 +333,23 @@ static void checkMaxBuckets(int size, int maxBuckets) {
...
+  private boolean isMatchAll(LeafReaderContext context) throws IOException {
+    return weight != null && weight.count(context) == context.reader().maxDoc();
+  }
+
+  private PointRangeQuery getPointRangeQuery(final String field) {
+    if (weight == null || weight.getQuery() == null) {
+      return null;
+    }
+
+    if (weight.getQuery() instanceof PointRangeQuery) {
```
jpountz, line 346: Nit: you could use more modern syntax since `main` requires Java 21 now.

```suggestion
    if (weight.getQuery() instanceof PointRangeQuery prq) {
```

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
```diff
@@ -330,4 +333,23 @@ static void checkMaxBuckets(int size, int maxBuckets) {
...
+
+  private boolean isMatchAll(LeafReaderContext context) throws IOException {
+    return weight != null && weight.count(context) == context.reader().maxDoc();
+  }
+
+  private PointRangeQuery getPointRangeQuery(final String field) {
+    if (weight == null || weight.getQuery() == null) {
+      return null;
+    }
+
```
jpountz, line 345: You'll probably want to remove layers of queries that don't affect matches, e.g. BoostQuery, ConstantScoreQuery, IndexOrDocValuesQuery, before doing the instanceof check. Especially as PointRangeQuery is often wrapped in an IndexOrDocValuesQuery.

jpountz (approved):

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
jainankitk: Thanks for sharing this. Really nice improvement, wasn't aware of it. `instanceof` is bit of a pain otherwise!

lucene/sandbox/src/java/org/apache/lucene/sandbox/facet/plain/histograms/HistogramCollector.java
jainankitk: That's a good point. I have added code for recursively unwrapping `BoostQuery`, `ConstantScoreQuery`, `FunctionScoreQuery` and `IndexOrDocValuesQuery`.

PR #14591, opened 2025-04-30
Upgrade to gradle 8.14 on branch_10x (backport).

dweiss:
This is a limited backport of #14533 bumping gradle to v8.14.


lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java
```diff
@@ -379,7 +379,9 @@ private static <A> MMapIndexInputProvider<A> lookupProvider() {
       // access through the lookup:
       final var constr = lookup.findConstructor(cls, MethodType.methodType(void.class, int.class));
       try {
-        return (MMapIndexInputProvider<A>) constr.invoke(maxPermits);
+        @SuppressWarnings("unchecked")
```
uschindler, line 382: Is this needed here? This was only to workaround a java 24 bug.

uschindler: That's why I asked. The compilation with java 24 on Jenkins worked all the time. 

It is only an issue if release is targeting 24. They had some language updates. But we should open a bug report about this.

I will make a reproducer and open issue for 24.

uschindler:
I got a strange error: 

* What went wrong:
Execution failed for task ':downloadEcjDrop'.
> A failure occurred while executing de.undercouch.gradle.tasks.download.internal.DefaultWorkerExecutorHelper$DefaultWorkAction
   > javax.net.ssl.SSLHandshakeException: (certificate_unknown) PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target

But on second try it worked.

Another problem is that the ECJ compiler config is not using 24 but still 23 in this PR. I will fix this in a commit.

uschindler:
Fixed: 0550cc6d2d0294641679131b597d591896211fce

uschindler:
One information about Exclipse: With the recent commit all works fine. It just sometimes fails with a compilation error (but that's not new) regarding: "Permitted type LeafReaderContext does not declare org.apache.lucene.index.IndexReaderContext as a direct supertype"

This error is bullshit but something seems to confuse the compile inside the Eclipse IDE.

Anyways: If you want to use exclipse, you need to install the Milestone 1 of 2025-06 release, otherwise it won't work (same reason like discussed above). Luckily on starting Eclipse it asks you if you want to try the milestone release.

uschindler:
sorry the messages above are in wrong PR. deleting....

PR #14594, opened 2025-04-30
add comments around how dependencies are used

rmuir:
Add one-liner above comments giving a hint how the dependency is used. This can help make the dependencies easier to maintain: it points you in the right direction and will be shown in dependabot diffs.


stefanvodita (approved):
Thank you for adding those explanations!

PR #14597, opened 2025-04-30
Rewrite expressions module with Java 24 classfile API

uschindler:
First step in getting rid of ASM dependency:

Java 24 added the classfile API (https://openjdk.org/jeps/484) in non-preview mode. This allows to generate class files with a builder-based API (unfurtunately its is not the usual ASM visitor desaster, it is more a lambda-inside-lambda desaster).

This PR removes the license files and all ASM dependencies from runtime code (and module descriptors). The code was rewritten. The rewrite may not yet be the most elegant lambda-way to make it, but the PR is easier to review with not so many changes and reshuffled code.

We still have ASM core as a dependncy for Gradle's APIJAR generator, but this one can also be removed soon.


dweiss (approved):
Pretty cool.

rjernst (approved):
Looks great! It seems like ASM translated relatively well (in this case) to the ClassFile API (much of it was just changing imports/methods to call but looks very similar).

jdconrad (approved):
LGTM! I appreciate the fairly direct replacement as the first step since it made it so easy to review.

rmuir (approved):
looks great! also, I trust the tests here have good coverage. Thank you for doing this.

uschindler:
I did some final cleanup and added changes entry. Will merge soon.

Thanks for the reviews by @rjernst and @jdconrad, very appreciated.

uschindler:
> Looks great! It seems like ASM translated relatively well (in this case) to the ClassFile API (much of it was just changing imports/methods to call but looks very similar).

The hard part was more to understand and figure out how it works. After that transforming the Antlr visitor to Classfile API was trivial.

Very hard was that I was unable to figure out how to get a `ConstantDesc` which is needed for the bootstrapper from a `String` or `Integer`. I had to figure out first that all those classes implement `ConstantDesc` so you can just pass the `String` instance as parameter where type `ConstantDesc` is expected: https://github.com/apache/lucene/blob/938be2a1c313787bb31c6a5b9fde61b1e0025161/lucene/expressions/src/java/org/apache/lucene/expressions/js/JavascriptCompiler.java#L436

image

There are still some nice goodies missing in Classfile API:
- No replacement for `GeneratorAdaptor#loadArgs` and a better named function like `GeneratorAdaptor#loadThis` (as synonmy for `aload(0)`). This would make the typical pass all parameters automatically to the invoke super call in ctors or overridden methods.
- More methods with `TypeKind` to also generate compare code without knowing the exact tyes.

dweiss:
> The hard part was more to understand and figure out how it works. 

For the fun of it, I asked chatgpt to do it. Uwe - you're still much, much better. :)

rmuir:
lol the day we have an army of Uwe AIs, I feel like I've already seen this [movie](https://www.youtube.com/watch?v=wN6F1l4cJ1Y)

uschindler:
I made a small followup commit to make usage more like recommended by Classfile API docs: e2d06855cc09df4b1293817c99d767ef7d8adc6d

uschindler:
I made some followup changes here: #14602

PR #14599, opened 2025-04-30
Switch s2-geometry to google's 'official' version

dweiss:
This isn't complete. There are some extra dependencies I'm not sure are needed. But overall, the transition seems to be simple. Not that the "new" library from Google is that well maintained (there is only one version published).


gradle/validation/dependencies.gradle
```diff
@@ -28,7 +28,6 @@ allprojects {
 
   def testConfigurations = project.configurations.matching {
     it.name in [
-      "annotationProcessor",
```
dweiss, line 31: I've removed annotation processor configuration from version lock computation. I don't think it's relevant to us what's used in annotation processors.

dweiss:
This requires pulling in guava, which in turn has a module-info in a multi-release jar location. ECJ fails on this due to this bug I filed 4 years ago:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=577790
ported to gh as https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2778

dweiss:
I've made it run and compile. It is surprisingly hairy because of modular classpaths, errorprone interactions and extra dependencies brought in. There is a single version of this s2 geometry library ever published. At this point I'm thinking we could just fold whatever is needed into this module (it is asl2) instead of importing the external jar...

rmuir:
I like the idea of reducing the deps: not anything crazy, but just trying to keep the dep tree minimal. We frequently hear from users on decades-old versions on the ML, so it is less problems for end-users as well.

dweiss:
I've already excluded what was possible. Anything more would require copying the code over and manual removal of dependencies (which shouldn't be too difficult). I'm at crossroads here. This guava dependency is going to be a problem, I can sense it.

PR #14602, opened 2025-05-01
Refactor the expressions compiler to use official ClassData BSM with indexed lookup

uschindler:
With more review of the code, I thought that it would be better to use one of the default Bootstrap methods available for ClassData. This allows to get rid of our custom BSM method and just use one of the defaults (`MethodHandles#classDataAt()`). By default it is possible to supply a `List` as ClassData and the bootstrap gets the MethodHandle by index. The code for that index was already there and is now reused. Like for the variables, the index is incremented on each new constant (like on each variable) and the index used for lookup.

As this change is simple it can also be backported, in the backport I just need to adapt the ASM 'Handle` to point to the default ClassData bootstrap method.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -78,6 +78,9 @@ Improvements
 
 * GITHUB#14476: Removing unnecessary splitX in ComponentTree (Ankit Jain)
 
+* GITHUB#14602: Refactor the expressions compiler to use official ClassData BSM with indexed lookup
+  (Uwe Schindler)
+
 Optimizations
 ---------------------
 * GITHUB#14418: Quick exit on filter query matching no docs when rewriting knn query. (Pan Guixin)
--- a/lucene/expressions/src/java/org/apache/lucene/expressions/js/JavascriptCompiler.java
+++ b/lucene/expressions/src/java/org/apache/lucene/expressions/js/JavascriptCompiler.java
@@ -32,7 +32,6 @@ import java.lang.classfile.instruction.BranchInstruction;
 import java.lang.classfile.instruction.OperatorInstruction;
 import java.lang.constant.ClassDesc;
 import java.lang.constant.ConstantDescs;
-import java.lang.constant.DirectMethodHandleDesc;
 import java.lang.constant.DynamicConstantDesc;
 import java.lang.constant.MethodTypeDesc;
 import java.lang.invoke.MethodHandle;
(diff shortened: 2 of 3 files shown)


uschindler:
I also added a new test to make sure that using the same function two times works correct.

```java
  public void testSameFunctionTwoTimes() throws Exception {
    Expression expr = compile("sqrt(20)+abs(-7)+sqrt(30)");
    assertEquals(Math.sqrt(20) + 7 + Math.sqrt(30), expr.evaluate(null), DELTA);
  }
```

I also verified the bytecode, now looks like this:

```
//  (version 24 : 68.0, no super bit)
public final class org.apache.lucene.expressions.js.JavascriptCompiler$CompiledExpression extends org.apache.lucene.expressions.Expression {
  
  // Method descriptor #6 (Ljava/lang/String;[Ljava/lang/String;)V
  // Stack: 3, Locals: 3
  public JavascriptCompiler$CompiledExpression(java.lang.String arg0, java.lang.String[] arg1);
    0  aload_0 [this]
    1  aload_1 [arg0]
    2  aload_2 [arg1]
    3  invokespecial org.apache.lucene.expressions.Expression(java.lang.String, java.lang.String[]) [8]
    6  return

  
  // Method descriptor #10 ([Lorg/apache/lucene/search/DoubleValues;)D
  // Stack: 5, Locals: 2
  public double evaluate(org.apache.lucene.search.DoubleValues[] arg0) throws java.io.IOException;
     0  ldc <Dynamic> 0 _ java.lang.invoke.MethodHandle [22]
     2  ldc2_w <Double 20.0> [23]
     5  invokevirtual java.lang.invoke.MethodHandle.invokeExact(double) : double [30]
     8  ldc <Dynamic> 1 _ java.lang.invoke.MethodHandle [32]
    10  ldc2_w <Double 7.0> [33]
    13  dneg
    14  invokevirtual java.lang.invoke.MethodHandle.invokeExact(double) : double [30]
    17  dadd
    18  ldc <Dynamic> 0 _ java.lang.invoke.MethodHandle [22]
    20  ldc2_w <Double 30.0> [35]
    23  invokevirtual java.lang.invoke.MethodHandle.invokeExact(double) : double [30]
    26  dadd
    27  dreturn
    28  aload_0 [this]
    29  invokestatic org.apache.lucene.expressions.js.JavascriptCompiler.patchStackTrace(java.lang.Throwable, org.apache.lucene.expressions.Expression) : java.lang.Throwable [44]
    32  athrow
      Exception Table:
        [pc: 0, pc: 28] -> 28 when : java.lang.Throwable
      Stack map table: number of frames 1
        [pc: 28, same_locals_1_stack_item, stack: {java.lang.Throwable}]

Bootstrap methods:
  0 : # 18 invokestatic java/lang/invoke/MethodHandles.classDataAt:(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/Class;I)Ljava/lang/Object;
	Method arguments:
		#11 0,
  1 : # 18 invokestatic java/lang/invoke/MethodHandles.classDataAt:(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/Class;I)Ljava/lang/Object;
	Method arguments:
		#31 1
}
```

uschindler:
@s1monw it should be same function, not similar!

uschindler:
The backport is the following commit: 6fb9dee0e2078a6044dd3a5c0020736f14fb2f5a

PR #14607, opened 2025-05-04
Revise strategy for opening an index for reading

rahulgoswami:
### Description

This PR is created after discussing the idea on the lucene dev list.

https://lists.apache.org/thread/gk3kwplon73llz356szz1mn3myn3nnm3

The revised proposal **[Approach 2]** (https://github.com/apache/lucene/pull/14607#issuecomment-3264544357) keeps the SegmentInfos.indexCreatedVersionMajor untouched. Instead the proposal is to introduce a new property currentIndexMajorVersion which can be updated as appropriate to facilitate index back compatibility.

Update (09/19/2025): There is also an **[Approach 3]** presented in https://github.com/apache/lucene/pull/14607#issuecomment-3310474509

**Update (09/23/2025) - Summary:**

**Problem statement**
If all the segments have minVersion and version >= MIN_SUPPORTED_MAJOR , please allow the user to work with the index irrespective of when the index was originally created (i.e. even if SegmentInfos.indexCreatedVersionMajor<MIN_SUPPORTED_MAJOR)

**Example scenario**
User creates an index on Lucene 10 (also, MIN_SUPPORTED_MAJOR=10). Say it stays the same until Lucene 13. In Lucene 13 there is a change which requires reindexing. This means that when Lucene 14 releases, MIN_SUPPORTED_MAJOR moves up to 13.  
However, while on Lucene 13 if user is able to reindex in a way that at the end of the process, for each segment, version=minVersion=13 (refer https://github.com/apache/lucene/pull/14607#issuecomment-3325749486 for how this can be achieved outside of Lucene), then upon moving to Lucene 14, this index should be readable even though index was originally created on 10.

**For more context**
https://github.com/apache/lucene/pull/14607#issuecomment-3325749486
https://github.com/apache/lucene/pull/14607#issuecomment-3264544357

**Proposed approaches**
(All approaches assume the prerequisite that all segments of the index have been converted outside of Lucene to have version=minVersion=LATEST )

**[Approach 1]** Rewrite SegmentInfos.indexCreatedVersionMajor to LATEST upon commit. (PR implementation as of 09/23/2025)

**[[Approach 2]](https://github.com/apache/lucene/pull/14607#issuecomment-3264544357)** Keep indexCreatedVersionMajor untouched. Introduce a mutable parallel property in SegmentInfos (say "currentIndexMajorVersion") and use this for deciding whether an index should be opened. 

**[[Approach 3]](https://github.com/apache/lucene/pull/14607#issuecomment-3310474509)** Look at the list of SegmentInfo (instead of indexCreatedVersionMajor) at read time to determine compatibility.


rmuir:
As states on the dev list, I don't think we should do this.

rahulgoswami:
> As states on the dev list, I don't think we should do this.

Can you please elaborate why? I have tried to address the concern you brought up on the mailing list via the implementation in this PR. 

The primary premise behind the API is that IF all segments of an index are created by the LATEST version, the index in all respects is LATEST. "indexCreatedVersionMajor" should ideally not block a Lucene upgrade in that case. 

Happy to learn if I am missing anything.

rmuir:
Because that's not what this variable means. If you want to change index backwards compatibility policy, lying about the version is not the way. -1

rahulgoswami:
An index where all segments have minVersion=LATEST and version=LATEST is essentially an index with version LATEST.

Please note that this API does NOT require Lucene to perform the upgrade/reindexing part. That will be done by the upper layer (Solr, Elasticsearch, etc) , say, through a custom merge policy. The requirement from the Lucene API is "If all the segments have minVersion=version=Version.LATEST, please allow the user to work with the index irrespective of what the SegmentInfos.indexCreatedVersionMajor says".

**[Approach 2]** 
If the concern is to maintain tracking data regarding the origins of the index, SegmentInfos.indexCreatedVersionMajor could be left untouched. Instead a new property "currentIndexMajorVersion" could be introduced which sits parallel to indexCreatedVersionMajor

For a fresh index, currentIndexMajorVersion=indexCreatedVersionMajor=Version.LATEST

Upon an upgrade, if the upper layer (Solr, Elasticsearch, etc) successfully converts all segments to Version.LATEST, calling the IndexWriter.commitAndUpdateVersionCreatedMajor() (which will now be called commitAndUpdateCurrentMajorVersion() ) should set currentIndexMajorVersion=Version.LATEST, while indexCreatedVersionMajor remains untouched.

SegmentInfos.readCommit()  now relies on SegmentInfos.currentIndexMajorVersion for codec compatibility checks. 

@jpountz It would be valuable to get your thoughts on this too. Thank you.


**Update (09/19/2025)**:
Since https://github.com/apache/lucene/pull/15012 got merged earlier today, rewording certain parts of the requirement..."if all the segments have minVersion and version >= MIN_SUPPORTED_MAJOR , please allow the user to work with the index irrespective of what the SegmentInfos.indexCreatedVersionMajor says"

rahulgoswami:
**[Approach 3]**
This is inspired by @msokolov 's idea at the recently concluded Community Over Code where I gave a talk on this topic ("Achieving zero downtime and zero cost index upgrade in Apache Solr"). 

Do we HAVE to rely on indexCreatedVersionMajor to decide index compatibility? 

If in SegmentInfos.readCommit() there could be a check over the list of SegmentInfo and if all segments have minVersion and version >= MIN_SUPPORTED_MAJOR, it should be good to open. Considering that even for an average large index this would mean a few hundred iterations, would this one time check at startup time be a worthy tradeoff?

msokolov:
Yes, I think your "Approach 3" is nice since it doesn't have us lying about the min created version, and it doesn't require rewriting anything immutable. It just changes our readers so they can  read an index whose segments were written by compatible Lucene versions.  It still prevents from reading indexes that were originally written with an index version < MIN_SUPPORTED_MAJOR.

rahulgoswami:
Appreciate your inputs @msokolov . I will rework this PR with Approach 3 soon.  

>It still prevents from reading indexes that were originally written with an index version < MIN_SUPPORTED_MAJOR.

I guess you meant this already but just in the interest of complete clarity, it DOES allow reading indexes originally written with an index version < MIN_SUPPORTED_MAJOR provided that ALL of the _current_ segments of the index have version and minVersion >= MIN_SUPPORTED_MAJOR (as opposed to relying on "indexCreatedVersionMajor" property in SegmentInfos today).

rmuir:
> I guess you meant this already but just in the interest of complete clarity, it DOES allow reading indexes originally written with an index version < MIN_SUPPORTED_MAJOR provided that ALL of the _current_ segments of the index have version and minVersion >= MIN_SUPPORTED_MAJOR (as opposed to relying on "indexCreatedVersionMajor" property in SegmentInfos today).

This doesnt work. Sometimes we need to force a reindex. That's because the original thing encoded was lossy (e.g. norms) and we need to change the encoding, or because it was indexed incorrectly. all the merging in the world can't fix that.

rahulgoswami:
>This doesnt work. Sometimes we need to force a reindex. That's because the original thing encoded was lossy (e.g. norms) and we need to change the encoding, or because it was indexed incorrectly. all the merging in the world can't fix that.

For each older segment which is >=MIN_SUPPORTED_MAJOR but <LATEST, if it can be excluded from merges  (eg: via a custom merge policy) and, in parallel, can be read and reindexed outside of Lucene (Eg: in Solr, etc ),  the result would be a LOSSLESS index where ALL segments are LATEST version. Of course this only applies to any index where all fields are either stored or docValues enabled. 

Such indexes can benefit from this kind of in-place reindexing giving the ability to keep "upgrading" even in case of any back incompatible changes like the norms change that happened in 7.x (as long as Lucene maintains readability for LATEST-1 version segments). 

I did mention this basic premise for this PR on the mailing list, but in hindsight, could have re-stressed upon it. The Solr side effort is being tracked under https://issues.apache.org/jira/browse/SOLR-17725 . 

Once Solr can achieve the state as mentioned above (aka all segments have been converted to LATEST version), today the only thing preventing the index from opening is the check in SegmentInfos based on indexCreatedVersionMajor (https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java#L349). 

The three approaches outlined in this PR offer alternatives to this check.

msokolov:
Right, I forgot the key fact here that also came out during your presentation at ApacheCoC. Namely in the system where you want to use this, the documents have been *completely reindexed* from source that is stored in the index in a stored field (IIRC) so there can be no question of lingering data that was indexed by a previous version. So all the individual segments are "new version" segments and it is only the index-wide min-segment version that is preventing the reader from opening the index, which is really now just a piece of "ghost metadata" that preserves the fact that at one time there were some segments in this index that had been created by an older version of Lucene.

I think the important restriction to keep in place is that readers will still not be able to open indexes where *any segment* was originally written with an index version < MIN_SUPPORTED_MAJOR.  The change is to have the reader make the decision based on the aggregate of the segment-level metadata (in SegmentInfo) rather than looking at the index-wide metadata (in SegmentInfos, I think).

rahulgoswami:
@msokolov  That's absolutely correct. You summarized every bit perfectly. Thank you.

So the breakdown of tasks for the complete solution looks like this:
1) Reindex all older segments with a custom merge policy (to prevent them from merging with LATEST version segments during the reindexing process) ==> to be done by Solr ==> Results in ALL segments being LATEST version
2) At index read time, perform a check to ensure none of the segments are < MIN_SUPPORTED_MAJOR. Don't rely on the immutable metadata ("indexCreatedVersionMajor") in SegmentInfos ==> to be done by Lucene.

Also, another requirement for the overall solution to work is that the user is not jumping across incompatible major versions. Eg: If an index is created in Lucene 11 (with MIN_SUPPORTED_MAJOR=10) and a breaking change happens in Lucene 13 (while MIN_SUPPORTED_MAJOR is still 10), then before moving to Lucene 14 (with now MIN_SUPPORTED_MAJOR=13), user would need to reindex on Lucene 13 via the above process. They can't jump versions, say from Lucene 12 to Lucene 14 and expect to reindex on 14 (since the index won't open for reading on 14).

rahulgoswami:
@msokolov  @mikemccand Would appreciate if you could take a look please..

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
```diff
@@ -410,6 +398,85 @@ private static void parseSegmentInfos(
       SegmentInfo info =
           codec.segmentInfoFormat().read(directory, segName, segmentID, IOContext.READONCE);
       info.setCodec(codec);
+      Version segMinVersion = info.getMinVersion();
```
msokolov: is it a functional change to move this version checking early in the function? It makes reviewing harder since the diffs look bigger and one has to carefully read and scroll around to compare by eyeball, so if it's not needed, let's leave it where it was.

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
```diff
@@ -410,6 +398,85 @@ private static void parseSegmentInfos(
...
+      Version segmentVersion = info.getVersion();
+
+      if (!segmentVersion.onOrAfter(infos.minSegmentLuceneVersion)) {
```
msokolov: we prefer `== false` to `!` for clarity - it helps to avoid mistakes

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
```diff
@@ -410,6 +398,85 @@ private static void parseSegmentInfos(
...
+              "segments infos must record minVersion with indexCreatedVersionMajor="
+                  + infos.indexCreatedVersionMajor,
+              input);
+        }
+      }
+
+      if (segMinVersion.major < minSupportedMajorVersion) {
+        throw new IndexFormatTooOldException(
+            input,
+            "Index has segment traces from Lucene version "
```
msokolov: not sure what you mean by "traces" here? Maybe like "traces of an older version"? I think it will be clearer to just say "Index segment derived from version "?

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
```diff
@@ -410,6 +398,85 @@ private static void parseSegmentInfos(
...
+                + Version.LATEST
+                + ". This Lucene version only supports indexes with major version "
+                + minSupportedMajorVersion
+                + " or later (found: "
+                + segMinVersion.major
+                + ", minimum supported: "
+                + minSupportedMajorVersion
+                + "). To resolve this issue: (1) Re-index your data using Lucene "
+                + minSupportedMajorVersion
+                + ".x or later (preferably "
```
msokolov: Let's strike "preferably" and say explicitly what is allowed

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
rahulgoswami: Yes that was exactly the intention- to move the check early in the function to fail fast. Agree it makes the reviewing trickier. Will incorporate the suggestion.

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
rahulgoswami: Interesting. I generally find the `!` to be more readable in a natural language flow. Will follow Lucene's convention with `== false`.

rahulgoswami:
@msokolov  Thanks for taking the time to review this. I have refactored the PR with your suggestions.

msokolov (approved):
Thanks, much cleaner!

msokolov:
I just want to change the commit message and CHANGES doc-line to be more descriptive, something like:

> Index open performs version check on each segment, ignores indexCreatedVersionMajor (#14607)

I will do this myself and push

msokolov:
I merged from command-line in order to change make the CHANGES.txt and commit message change. Maybe it would have been cleaner to do it here, sorry

msokolov:
If it would be helpful, I think we could also backport to 10x branch

rahulgoswami:
> If it would be helpful, I think we could also backport to 10x branch

yes pretty please! Solr is in the process of releasing 10x and I was hoping it could have this. Ideally Lucene 9x too if that's a possibility (wishful thinking?)

Thanks for merging this PR!

msokolov:
OK, do you want to try? It's just a question of cherry-picking to `branch_10x`. I doubt there will be any 9x release. but you might be able to cut one for yourself?

rahulgoswami:
Sure I'll cherry pick for 10x. 

> I doubt there will be any 9x release. but you might be able to cut one for yourself?

Is Lucene 9x already marked EOL ? I assumed since 11x is not out yet, 9x must still be receiving fixes \(?\) .

msokolov:
hmm, I had to revert  because of https://github.com/apache/lucene/issues/15374  ... I don't totally grok what is going on there, but it seems that we attempt to read segments with version 7.4 and fail because we don't have the backwards-compat codec

rahulgoswami:
Investigating https://github.com/apache/lucene/issues/15374. As communicated there, seems like this is only failing for the nightly build. gradle check ran fine with the commit.

rahulgoswami:
Yeah seems like the error is since the lucene-backward-codecs.jar doesn't contain codecs for 7.x anymore. The nightly test was succeeding earlier since we were bailing out by looking at indexCreatedVersionMajor in SegmentInfos straightaway. With the new change we are looking into each segment which needs the codec to open the segment. 
 
>         Caused by:
   >         java.lang.IllegalArgumentException: An SPI class of type org.apache.lucene.codecs.Codec with name 'Lucene70' does not exist.  You need to add the corresponding JAR file supporting this SPI to your classpath.  The current classpath supports the following names: [Lucene104, Asserting, CheapBastard, DeflateWithPresetCompressingStoredFieldsData, FastCompressingStoredFieldsData, FastDecompressionCompressingStoredFieldsData, HighCompressionCompressingStoredFieldsData, LZ4WithPresetCompressingStoredFieldsData, DummyCompressingStoredFieldsData, ConfigurableMCodec, SimpleText, Lucene80, Lucene84, Lucene86, Lucene87, Lucene90, Lucene91, Lucene92, Lucene94, Lucene95, Lucene99, Lucene912, Lucene100, Lucene101, Lucene103]
   >             at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.util.NamedSPILoader.lookup(NamedSPILoader.java:113)
   >             at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.codecs.Codec.forName(Codec.java:119)
   >             at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.SegmentInfos.readCodec(SegmentInfos.java:528)
   >             ... 47 more

rahulgoswami:
I am thinking maybe this needs exception handling in the code in case of missing codec (in which case index is too old anyway) and/or corresponding change in the exception message during assert check in the test. I'll revisit this tonight. Thanks.

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
```diff
@@ -402,11 +389,38 @@ private static void parseSegmentInfos(
...
     long totalDocs = 0;
+
     for (int seg = 0; seg < numSegments; seg++) {
       String segName = input.readString();
       byte[] segmentID = new byte[StringHelper.ID_LENGTH];
       input.readBytes(segmentID, 0, segmentID.length);
-      Codec codec = readCodec(input);
+      Codec codec = null;
+      try {
+        codec = readCodec(input);
```
msokolov: readCodec is private in this class so we're free to change it. Let's make it throw `IndexFormatTooOldException` directly; then there is no need to catch and rethrow.

Also -- it is still possible that the user actually *did* forget to load lucene-backward-codecs.jar and the version of the index they are trying to load actually *is* a supported one.

I think what we need to do is to add an explicit check on the index version somewhere so we can distinguish between a supported codec that failed to load and an unsupported (too old) codec.

msokolov: hmm thinking more about this, maybe we are unable to know what version the old index is if we cannot read the segmentinfos file and the best we can do is issue an exception like we did before?  At least we should acknowledge that the problem *might* be caused by missing jars on the classpath rather than a very old index - we can't tell for sure

rahulgoswami: Hmm this is tricky. As such, we are able to read the SegmentInfos file and extract the indexCreatedVersionMajor which is how it was done  previously. However if we base our decision on it, it would defeat the purpose of this PR. 

The idea is that if the reindexing is done in a way as described in my talk and also in the discussion on this PR, then we could end up in a scenario where indexCreatedVersionMajor is pretty old, but all segments have still been reindexed with supported codec. And we should be able to open such an index.

I agree that the best course of action might be to retain the existing behavior of throwing IllegalArgumentException outlining the two possibilities for failure (backward-codecs-jar not being present or index is too old). Unless I can think of anything better, I guess I will just change TestAncientIndicesCompatibility instead.

rahulgoswami: Attempting to solve in the readCodec() private method as below. Reasonable?

```
 catch (IllegalArgumentException e) {
      for (String codecName : Codec.availableCodecs()) {

      // other backward codecs are present in the classpath
        // so if this was a supported index version, we should have found it in the class path too
        // aka this is likely an old index.
        if (<condition to detect whether codecName belongs to to org.apache.lucene.backward_codecs>) {
        
          throw new IndexTooOldException(message);
        }
      }

      //older logic
      if (name.startsWith("Lucene")) {
            "Could not load codec '"
                + name
                + "'. Did you forget to add lucene-backward-codecs.jar?",
            e);
      }
      throw e;
    }
```

msokolov: So ... I am pretty sure that `Codec.availableCodecs` is simply listing what is available on the classpath?  I think the best we can do is attempt to read the segmentinfos file, and if we fail, give a fairly generic exception message.

You seemed to be saying your system would be able to read an index for which we don't have any backward codec available ("As such, we are able to read the SegmentInfos file and extract the indexCreatedVersionMajor which is how it was done previously"), but how can that be?

msokolov: I guess with your change we can still read indexes that were originally created with 8x and 9x (which would not have been possible before) as long as their segments have been rewritten by 10x or 11x, but if the index was created by 7x or earlier we would not.

rahulgoswami: >you seemed to be saying your system would be able to read an index for which we don't have any backward codec available ("As such, we are able to read the SegmentInfos file and extract the indexCreatedVersionMajor which is how it was done previously"), but how can that be?

Not the index, but looks like the nightly test failure suggests we can at least read the 'indexCreatedVersion' from SegmentInfos (aka 'segments_*' file) written in 7x without the codec present. If you see there are only two places at which readCommit() throws IndexFormatTooOldException. The previous nightly failure occurred because the below check was removed
```
if (indexCreatedVersion < minSupportedMajorVersion) {...}
``` 
which led to checking individual segments in parseSegmentInfos() (which is what we intend). I think it is in reading the segment level metadata/data where it needs the codec (`Codec codec = readCodec(input)` which is where the failure occurred)?  
Which also means even the original implementation assumes that we'll always be able to read the "indexCreatedVersion" from SegmentInfos once we make it past the first check in readCommit() for too old index (`if (magic != CodecUtil.CODEC_MAGIC)`)

Exception from nightly for reference:
```
    java.lang.IllegalArgumentException: Could not load codec 'Lucene70'. Did you forget to add lucene-backward-codecs.jar?
        at __randomizedtesting.SeedInfo.seed([14FBA2DCF615D67D:EF0090FF468E7121]:0)
        at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.SegmentInfos.readCodec(SegmentInfos.java:532)
        at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.SegmentInfos.parseSegmentInfos(SegmentInfos.java:397)
        at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.SegmentInfos.readCommit(SegmentInfos.java:354)
```

<br/>

> I guess with your change we can still read indexes that were originally created with 8x and 9x (which would not have been possible before) as long as their segments have been rewritten by 10x or 11x

Exactly

> but if the index was created by 7x or earlier we would not. 

Going by the above explanation, I think at least for index created by 7x, we _will_ be able to read, provided the segments have been rewritten by 10x or 11x.

rahulgoswami:
A little under the weather, but I don't want to lose the momentum on this. Thanks so much for helping navigate this PR so far Mike. I'll have a revision today based on our discussion.

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
msokolov: I tried something different, changing

```
      format = CodecUtil.checkHeaderNoMagic(input, "segments", VERSION_74, VERSION_CURRENT);
```
to
```
      format = CodecUtil.checkHeaderNoMagic(input, "segments", VERSION_86, VERSION_CURRENT);
```

since we can't read 7.4 anyway, why do we pretend to do so here?

and this makes the test pass up until version 8.0 indexes, when it fails in a new way I  haven't understood yet (the below assertion fails for 8.0):

 ```
      if (version.startsWith("8.") || version.startsWith("9.")) {
        assertTrue(indexStatus.clean);
      }
```

Comment (msokolov) [reply lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java]: It seems to me we still need a "min supported major" version check for the index, but it will be for a different version than the per-segment check (ie 8 instead of 10).

**Comment by msokolov** [comment]
sorry, hope you feel better! no pressure, take your time

File: lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java (hunk 5 above)
Comment (rahulgoswami) [reply lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java]: Seems like TestAncientIndicesCompatibility.testUnsupportedOldIndexes is happy as long as it gets an IndexFormatTooOldException (IFTOE). The test fails in the DirectoryReader.open(dir) call with an IllegalArgumentException (which is what the latest commit fixes by intercepting and throwing IFTOE instead during segment level check).

I observe two things...
1) the version values for VERSION_74 and VERSION_86 are different from what the variable names suggest.
 ```
 public static final int VERSION_74 = 9;

 public static final int VERSION_86 = 10;
```
2) CodecUtil.checkHeaderNoMagic(input, "segments", VERSION_74, VERSION_CURRENT) should have thrown an IndexFormatTooOldException for the 7.x index because of `if (actualVersion < minVersion) ` check. I am not sure yet why it didn't . That would have made the test pass. Something in the way the test index has been created? 

Checking further.

rahulgoswami: Please disregard my previous comment since my understanding of the VERSION_74 and VERSION_86 variables was incorrect. They are just versions assigned to the "segments" codec, and the variable names just represent the Lucene version in which the "segments" file version was assigned. I now fully understand what's happening and why the assertion fails when you changed to 

`format = CodecUtil.checkHeaderNoMagic(input, "segments", VERSION_86, VERSION_CURRENT);`

Moving to VERSION_86 means that CodecUtil.checkHeaderNoMagic() now throws IndexFormatTooOldException for indexes before Lucene 8.6.0. The test fails the assertion for version 8.0.0 on the line you mentioned since it now finds indexStatus.clean=false. That is because the below call in  checker.checkIndex() fails:

` infos = SegmentInfos.readCommit(dir, fileName, 0);`

And the reason it passes for >=8.6.0 is because the last param (min supported version) is 0, which means it proceeds to do  an individual segment check down the line, which passes because 8.x codecs are shipped and available in class path! 
**Essentially today if someone just removes the indexCreatedVersionMajor check in SegmentInfos, they should be able to read indexes created in 8.x even with Lucene 11 (!!)** 

By similar reasoning with VERSION_74, the test was only failing for indexes with version 7.4 to < 8.0.0 previously because the codec is not present. Maybe a message in IndexFormatTooOldException representing both possibilities (too old version or absent Lucene-backward-codecs) is a reasonable solution after all?

rahulgoswami:
I bumped up the min version of 'segments' codec to VERSION_86 in CodecUtil.checkHeaderNoMagic(). And updated the test to be more specific. This still maintains backward compatibility with Lucene 10. And even if cherry-picked to 10, will continue to maintain backward compatibility with 9x. 

The failed nightly passes fine now. Thanks for the hint @msokolov !

This helps us avoid the dilemma of a lucene-backward-codecs.jar actually not being present in the class path vs an actual unsupported index. I feel good about this one. 

Tried running "gradlew test" with -Dtests.nightly=true but had to terminate as it was taking awfully long.

dweiss:
You can try to run nightly on just selected subprojects (gradlew -p lucene/core check -Ptests.nightly=true).

rahulgoswami:
Thanks @dweiss . The nightlies for both lucene/core and lucene/backward-codecs pass fine after the latest commit!

lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java
rahulgoswami: I was able to avoid the ambiguity around messaging and resolve the nightly failure more elegantly with the latest commit (Detailed in the main thread below)

gaeljw:
Thanks a lot for all this work and the reviews 👏

If this is not too complex, having it in 9.x would be very welcome as well.

msokolov:
Looks good! Glad you were able to get back to this - I'll merge

rahulgoswami:
Ya had a rough few days healthwise, and hence the dip in activity. But I am back now, and glad we were able to drive this to conclusion. Thank you for the thoughtful reviews Mike! 
I believe https://github.com/apache/lucene/issues/15374 can be closed too ?

msokolov:
Thanks I closed that issue. Would you care to post a PR for the backport to 10.x? This would be a simple cherry-pick to `branch_10x`.  (We won't do 9.x since it isn't being maintained since we won't likely produce any further 9-series releases).

rahulgoswami:
Sure, I'll do that. Thanks.

rahulgoswami:
Created #15431 for back port to branch_10x

PR #14608, opened 2025-05-04
Always check gradle wrapper sha checksum and download if necessary

### LINKED ISSUE #14598: gradle-wrapper.jar will not be updated when its sha/version changes
### Description

Uwe noticed this. After the update to gradle 8.14, I've updated the gradle-wrapper script's version and SHA. The new scripts launch it using -jar (previously it was launched using an explicit main method). This uncovered a potential problem that hasn't manifested itself so far - we only check if gradle-wrapper.jar exists; if it doesn't exist, we invoke WrapperDownloader but if gradle-wrapper.jar already exists, we take no action and use this file (which may be older/different).

We could check the sha of the wrapper script on each invocation but this seems like an overkill. I don't have a fix for this one. I think we're just going to live with this oddity, hoping that it doesn't cause side effects often (what Uwe experienced).

_No response_

dweiss:
Fixes #14598

### DIFF
--- a/gradlew.bat
+++ b/gradlew.bat
@@ -81,11 +81,9 @@ SET DEFAULT_JVM_OPTS=%DEFAULT_JVM_OPTS% "-Djava.io.tmpdir=%GRADLE_TEMPDIR%"
 
 @rem LUCENE-9266: verify and download the gradle wrapper jar if we don't have one.
 set GRADLE_WRAPPER_JAR=%APP_HOME%\gradle\wrapper\gradle-wrapper.jar
-IF NOT EXIST "%GRADLE_WRAPPER_JAR%" (
-    "%JAVA_EXE%" %JAVA_OPTS% "%APP_HOME%/build-tools/build-infra/src/main/java/org/apache/lucene/gradle/WrapperDownloader.java" "%GRADLE_WRAPPER_JAR%"
-    IF %ERRORLEVEL% EQU 1 goto failWithJvmMessage
-    IF %ERRORLEVEL% NEQ 0 goto fail
-)
+"%JAVA_EXE%" -verbose -XX:TieredStopAtLevel=1 %JAVA_OPTS% "%APP_HOME%/build-tools/build-infra/src/main/java/org/apache/lucene/gradle/WrapperDownloader.java" "%GRADLE_WRAPPER_JAR%"
+IF %ERRORLEVEL% EQU 1 goto failWithJvmMessage
+IF %ERRORLEVEL% NEQ 0 goto fail
 
 @rem Generate gradle.properties if they don't exist
 IF NOT EXIST "%APP_HOME%\gradle.properties" (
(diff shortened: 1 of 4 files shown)


gradlew.bat
```diff
@@ -81,10 +81,22 @@ SET DEFAULT_JVM_OPTS=%DEFAULT_JVM_OPTS% "-Djava.io.tmpdir=%GRADLE_TEMPDIR%"
...
+set GRADLE_WRAPPER_CHECKSUM=%APP_HOME%\gradle\wrapper\gradle-wrapper.jar.sha256
+
+@rem Read the expected hash from .sha256 file
+for /f "tokens=1" %%A in (%GRADLE_WRAPPER_CHECKSUM%) do (
+    set "EXPECTED=%%A"
+)
+@rem Get actual SHA-256 hash using certutil
+for /f "tokens=* delims=" %%H in ('certutil -hashfile "%GRADLE_WRAPPER_JAR%" SHA256 ^| findstr /R /B /I /X "[0-9a-f]*"') do (
+    set "ACTUAL=%%H"
+)
```
dweiss, line 93: this computes sha256 using certutil, then tries to pick a line that looks like an sha256 checksum. Again - if something fails, we rely on the Java version to tell us what's wrong.

dweiss:
I've verified this works on Windows, Linux (ubuntu) and Mac.

rmuir (approved):
working for me.

still haven't figured out why eclipse whines about sha every time i open vim, but that's another issue (I don't want it invoking `gradle` in any way and try to disable that, but it still obviously has its own priorities)

Screen_Shot_2025-05-04_at_13 09 26

dweiss:
Interesting. I've got no idea, to be honest. Here is a list of shas for all releases -
https://gradle.org/release-checksums/

v8.14 is there and consistent with what we have.

vigyasharma:
Thanks for fixing this! 

I ran into the same problem. My old `gradle-wrapper.jar` was not getting updated and I kept hitting:
```bash
% ./gradlew clean
no main manifest attribute, in /Users/vigyas/repos/lucene/gradle/wrapper/gradle-wrapper.jar
```

My workaround till last night was to delete the `gradle-wrapper.jar`.  Verified that this change fixes it by putting back the old jar and having gradle find checksum mismatch and redownload.
```bash
% ./gradlew clean
Checksum mismatch, will attempt to re-download gradle-wrapper.jar
/Users/vigyas/repos/lucene/gradle/wrapper/gradle-wrapper.jar
Downloading gradle-wrapper.jar from https://raw.githubusercontent.com/gradle/gradle/v8.14.0/gradle/wrapper/gradle-wrapper.jar
Starting a Gradle Daemon (subsequent builds will be faster)
```

uschindler:
There is a small problem in windows with whitespace in path (missing escapes):

```
C:\Users\Uwe Schindler\Projects\lucene\lucene>gradlew :lucene:core:generateJdkApiJar24
Die Datei "C:\Users\Uwe" kann nicht gefunden werden.
"Checking gradle shas."
```

Can you fix this?

dweiss:
Should work now, @uschindler

uschindler:
Works, thanks!

uschindler:
> As a (former) windows user... Uwe, you're asking for trouble if you have a whitespace in your home dir... I'll take a look.

That's why I am doing this!

uschindler:
You know what I am doing? --> stay tuned, almost working!

> There is a small problem in windows with whitespace in path (missing escapes):
> 
> ```
> C:\Users\Uwe Schindler\Projects\lucene\lucene>gradlew :lucene:core:generateJdkApiJar24
> Die Datei "C:\Users\Uwe" kann nicht gefunden werden.
> "Checking gradle shas."
> ```
> 
> Can you fix this?

PR #14609, opened 2025-05-04
Overrides rewrite in PointRangeQuery to optimize AllDocs/NoDocs cases

ebradshaw:
Overrides rewrite in PointRangeQuery range to handle cases where the query either fully contains or fully excludes all documents within the shard.

Often, particularly when using time based partitioning, range queries may overlap several indexes.  Many of these indexes have timestamp values that are fully contained by the query, in which case the query can be rewritten to a MatchAllDocsQuery or a FieldExistsQuery.  On the other hand, many indexes can be fully excluded if they're outside the requested time range, in which case the query can be rewritten to a MatchNoDocsQuery.

While a similar optimization exists at the leaf level in the createWeight function, rewriting at the shard level enables other optimizations downstream.

Please let me know if this has been ruled out in the past for other reasons or if the implementation misses anything.  Thanks.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
@@ -22,11 +22,21 @@ import java.util.Objects;
 import java.util.function.BiFunction;
 import java.util.function.Predicate;
 import org.apache.lucene.document.IntPoint;
-import org.apache.lucene.index.*;
+import org.apache.lucene.index.DocValuesType;
+import org.apache.lucene.index.FieldInfo;
+import org.apache.lucene.index.IndexReader;
+import org.apache.lucene.index.LeafReader;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.index.PointValues;
 import org.apache.lucene.index.PointValues.IntersectVisitor;
 import org.apache.lucene.index.PointValues.Relation;
-import org.apache.lucene.util.*;
+import org.apache.lucene.util.ArrayUtil;
 import org.apache.lucene.util.ArrayUtil.ByteArrayComparator;
+import org.apache.lucene.util.BitSetIterator;
+import org.apache.lucene.util.BytesRef;
+import org.apache.lucene.util.DocIdSetBuilder;
+import org.apache.lucene.util.FixedBitSet;
+import org.apache.lucene.util.IntsRef;
 
 /**
  * Abstract class for range queries against single or multidimensional points such as {@link
--- a/lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
@@ -22,17 +22,11 @@ import java.util.Objects;
 import java.util.function.BiFunction;
 import java.util.function.Predicate;
 import org.apache.lucene.document.IntPoint;
-import org.apache.lucene.index.LeafReader;
-import org.apache.lucene.index.LeafReaderContext;
-import org.apache.lucene.index.PointValues;
+import org.apache.lucene.index.*;
 import org.apache.lucene.index.PointValues.IntersectVisitor;
 import org.apache.lucene.index.PointValues.Relation;
-import org.apache.lucene.util.ArrayUtil;
+import org.apache.lucene.util.*;
 import org.apache.lucene.util.ArrayUtil.ByteArrayComparator;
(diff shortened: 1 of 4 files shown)


jpountz:
+1 to do this

Feel free to ask me for a review on a PR if you'd live to give it a try.

jpountz (commented):
I left some minor comments but the change looks good to me in general. Thank you!

lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
```diff
@@ -580,4 +557,111 @@ public final String toString(String field) {
...
+      BytesRef qLow = new BytesRef(queryLow, offset, bytesPerDim);
+      BytesRef qHigh = new BytesRef(queryHigh, offset, bytesPerDim);
+      BytesRef gMin = new BytesRef(globalMinPacked, offset, bytesPerDim);
+      BytesRef gMax = new BytesRef(globalMaxPacked, offset, bytesPerDim);
+
+      if (qLow.compareTo(gMin) <= 0 && qHigh.compareTo(gMax) >= 0) {
+        fullyContainedCount++;
+      } else if (qLow.compareTo(gMax) > 0 || qHigh.compareTo(gMin) < 0) {
+        fullyExcludedCount++;
+      }
```
jainankitk, line 596: We can `return super.rewrite(searcher)` if any dimension is not fullyContained / Excluded?

lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
```diff
@@ -580,4 +557,111 @@ public final String toString(String field) {
...
+
+    if (fullyContainedCount == dims) {
+      if (canRewriteToMatchAllQuery(reader)) {
+        return new MatchAllDocsQuery();
+      } else if (canRewriteToFieldExistsQuery(reader)) {
+        return new FieldExistsQuery(field);
+      }
+    } else if (fullyExcludedCount == dims) {
+      return new MatchNoDocsQuery();
+    }
```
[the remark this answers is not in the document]
jainankitk: +1 to using `relate(byte[], byte[])`. Actually looking at `relate(byte[], byte[])` method, I realized that we don't need `fullyExcludedCount == dims` for rewriting into `MatchNoDocsQuery`. Any dimension being completely outside is sufficient condition for the query to be rewritten as `MatchNoDocsQuery`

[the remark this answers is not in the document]
ebradshaw: Agreed.  Missed the relate method on the first run but that logic is nice and clean.

lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
ebradshaw: Since we're using the relate method, we'll get the short circuit for excluded, but it'll still run through every contained check to determine inside vs crosses.  Should be okay since it's just a rewrite call?

ebradshaw:
I think we're good now.  Let me know if you guys see anything else, and thanks for the quick review!

lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
```diff
@@ -580,4 +534,114 @@ public final String toString(String field) {
...
+    for (LeafReaderContext leaf : reader.leaves()) {
+      checkValidPointValues(leaf.reader().getPointValues(field));
+    }
+
+    // fetch the global min/max packed values across all segments
+    byte[] globalMinPacked = PointValues.getMinPackedValue(reader, getField());
+    byte[] globalMaxPacked = PointValues.getMaxPackedValue(reader, getField());
+
+    if (globalMinPacked == null || globalMaxPacked == null) {
+      return super.rewrite(searcher);
```
jpountz, line 551: A null min or max packed value indicates an empty field, so you could rewrite to a `MatchNoDocsQuery` here as well.

jpountz (approved):
LGTM, can you add a CHANGES entry?

lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
jainankitk: Yeah, as you said this is only for rewrite and logic is much cleaner by reusing the relate method

jainankitk (approved):

jainankitk:
Thanks @ebradshaw for this change, and congratulations on your first contribution in lucene! :)

PR #14613, opened 2025-05-05
Rewrite APIJAR extractor to use Java 24 classfile API and kill ASM dependency also for build system

uschindler:
The rewrite was not so easy but I got it working.... (in fact, the Javascript compiler was easier).

The code also adds some more sanity checks about correct Java versions. So it will complain in future when the APIJAR extractor extracts for a older java version than the compile target.


uschindler:
P.S.: The preview flag removal code is not used by verctor API, but I ported it over, too. I tested it with temporrily extracting the Java 24 `ScopedValue` preview code.

uschindler:
I added an additional check missing previously as replacement for the last line in above output: All referenced classes left over after processing need to be publicly available in `java.base` module.

dweiss (approved):
nice

ChrisHegarty (approved):
Awesome! LGTM

uschindler:
Thanks for review, I will merge now!

PR #14614, opened 2025-05-05
Fix FuzzySet#getEstimatedNumberUniqueValuesAllowingForCollisions to account for hashCount

gsmiller:
### Description

Estimating bloom filter cardinality should account for the number of hash functions used. It appears this method assumes one function is used, which isn't correct.

Note: It also looks like the API surface area of `FuzzySet` in general may be due for a cleanup. This method doesn't appear to have any direct usage today, so it's debatable whether-or-not we should deprecate this. I'm in favor of keeping it, but would probably suggest deprecating a few other methods (include the version of this I kept around that assumes one hash function). I'll publish a separate PR where we can discuss this, but I think it's worth fixing this bug for now.

### Testing

I verified that this method dramatically over-estimates bloom filter cardinality in its current state (by a factor approximately equal to the hash count) and verified this change corrects it. This was in ad hoc testing related to something else I'm working on. I notice there are no unit tests for `FuzzySet` so I didn't add any explicit test cases for now. I can create some tests for `FuzzySet` as part of this if folks have a strong opinion on this. I can also add testing as a separate task when looking into cleaning up the API surface area...

### DIFF
--- a/lucene/codecs/src/java/org/apache/lucene/codecs/bloom/FuzzySet.java
+++ b/lucene/codecs/src/java/org/apache/lucene/codecs/bloom/FuzzySet.java
@@ -268,18 +268,31 @@ public class FuzzySet implements Accountable {
   }
 
   public int getEstimatedUniqueValues() {
-    return getEstimatedNumberUniqueValuesAllowingForCollisions(bloomSize, filter.cardinality());
+    return getEstimatedNumberUniqueValuesAllowingForCollisions(
+        bloomSize, filter.cardinality(), hashCount);
   }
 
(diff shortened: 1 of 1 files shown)


gsmiller:
Related: also opened #14615 to dry up the public API a bit.

msokolov (approved):
so this basically extends the API so `hashCount` can be supplied rather than always assumed to be 1?

gsmiller:
> so this basically extends the API so hashCount can be supplied rather than always assumed to be 1?

That's right. (But I think we should get rid of the version that assumes 1... kept it to be back-compat for this PR)

PR #14615, opened 2025-05-06
Shrink FuzzySet public API surface area

gsmiller:
### Description

Proposal to remove some public APIs we don't use internally anymore and likely don't want to maintain for public consumption. (Would back port with deprecation notices if the community agrees with this)

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -54,6 +54,8 @@ API Changes
 
 * GITHUB#15486: Remove deprecated SortField.setMissingValue() method (Alan Woodward)
 
+* GITHUB#14615 : Remove unnecessary public methods in FuzzySet (Greg Miller)
+
 New Features
 ---------------------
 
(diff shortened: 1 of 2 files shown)


gsmiller:
I was looking at Lucene's bloom filter codec again today for another reason, and was reminded of this open PR. I still think it would be nice to shrink the API surface area here. Curious if anyone has thoughts?

msfroh:
> Curious if anyone has thoughts?

For what it's worth, I am onboard. Always nice to shrink the API surface area if we can.

If someone wants a general-purpose Bloom filter in Java with all the bells and whistles, they can just use Guava. (As is, what's left of FuzzySet is a pretty useful Bloom filter.)

gsmiller:
Thanks @msfroh for your thoughts on this! I'll dust this back off and plan to merge it tomorrow if nobody else has objections. I think it's worth tightening this up.

PR #14616, opened 2025-05-06
Fix FuzzySet#createSetBasedOnMaxMemory to honor bytes not bits

gsmiller:
### Description

Looks like a bug to me, but no unit tests for this (should probably create some). I'm also wondering if we can find a better way to avoid needing this method altogether. It's only used for testing. Maybe there's a better way...


gsmiller:
Since this looks like a pretty obvious bug and only impact our unit test setup, I'm planning to merge this fix soon (going back through some older PRs and trying to move them along). Please speak up if you disagree with this change :)

(Separately, I think we can probably create a better way to cap bloom filter memory usage to replace this, but I think it's worth fixing this as-is for now)

PR #14626, opened 2025-05-08
Avoid unnecessary comparison for CELL_CROSSES_QUERY cases

jainankitk:
### Description
Probably this change is not required, as compiler might be doing this optimization implicitly. Will be great if someone can confirm the same.


dweiss:
I think I liked the previous version better - this wasn't unreadable at all to my eyes, typical boolean expression?

jainankitk:
> I think I liked the previous version better - this wasn't unreadable at all to my eyes, typical boolean expression?

You're right. In hindsight that does read better. I probably wanted to insist on the fact that comparison is skipped if crosses = true, and very common coding pattern for that is:

```
if (x == null || x.something)
```

but I guess it is pretty clear otherwise as well.

lucene/core/src/java/org/apache/lucene/search/PointRangeQuery.java
```diff
@@ -154,16 +154,18 @@ private Relation relate(byte[] minPackedValue, byte[] maxPackedValue) {
             return Relation.CELL_OUTSIDE_QUERY;
           }
 
-          crosses |=
-              comparator.compare(minPackedValue, offset, lowerPoint, offset) < 0
-                  || comparator.compare(maxPackedValue, offset, upperPoint, offset) > 0;
+          if (crosses == false) {
```
iverase, line 157: I would expect this comparision is not happening already. The current command is equivalent to:
```
crosses = crosses || omparator.compare(minPackedValue, offset, lowerPoint, offset) < 0
                  || comparator.compare(maxPackedValue, offset, upperPoint, offset) > 0;
```
Which should only be executed is crosses is false?

jainankitk: Unfortunately, the current command is equivalent to:

```
crosses = crosses | (comparator.compare(minPackedValue, offset, lowerPoint, offset) < 0
                  || comparator.compare(maxPackedValue, offset, upperPoint, offset) > 0);
```

In Java, the bitwise OR operator | can be applied to boolean operands. When used with booleans, it performs a logical OR operation, _but it evaluates both operands regardless of the left operand's value_. This is different from the logical OR operator ||, which short-circuits and doesn't evaluate the right operand if the left operand is true.

PR #14634, opened 2025-05-09
Use the preload hint on completion fields and memory terms dictionaries.

jpountz:
This enables the `PreloadHint` introduced in #14604 on completion fields and memory terms dictionaries, which are both expected to fit in the page cache in practice.

I don't have specific interest in these two file formats, I was more interested in having more than one file format that uses `PreloadHint` to make sure it's generally useful and not only to KNN vectors.

### DIFF
--- a/lucene/codecs/src/java/org/apache/lucene/codecs/memory/FSTTermsReader.java
+++ b/lucene/codecs/src/java/org/apache/lucene/codecs/memory/FSTTermsReader.java
@@ -41,6 +41,7 @@ import org.apache.lucene.index.TermsEnum;
 import org.apache.lucene.store.ByteArrayDataInput;
 import org.apache.lucene.store.FileTypeHint;
 import org.apache.lucene.store.IndexInput;
+import org.apache.lucene.store.PreloadHint;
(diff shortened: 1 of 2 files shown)


jpountz:
cc @thecoop @ChrisHegarty

lucene/codecs/src/java/org/apache/lucene/codecs/memory/FSTTermsReader.java
```diff
@@ -75,7 +76,8 @@ public FSTTermsReader(SegmentReadState state, PostingsReaderBase postingsReader)
 
     this.postingsReader = postingsReader;
     this.fstTermsInput =
-        state.directory.openInput(termsFileName, state.context.withHints(FileTypeHint.INDEX));
+        state.directory.openInput(
+            termsFileName, state.context.withHints(FileTypeHint.DATA, PreloadHint.INSTANCE));
```
jpountz, line 80: I also changed the FileTypeHint to DATA since this FST effectively stores all the terms data.

ChrisHegarty (approved):
LGTM

PR #14635, opened 2025-05-09
Clean up FileTypeHint a bit.

jpountz:
- more javadocs to guide usage of `INDEX` vs. `DATA`,
 - the `METADATA` constant is removed, as metadata files should be opened with `Directory#openChecksumIndexInput`, which doesn't take hints,
 - configure `FileTypeHint` on more files of the default codec,
 - remove checks on `FileTypeHint` from `toReadAdvice` - the default impl should only look at `DataAccessHint` to determine the appropriate read advice.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene90/compressing/Lucene90CompressingTermVectorsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene90/compressing/Lucene90CompressingTermVectorsReader.java
@@ -58,7 +58,6 @@ import org.apache.lucene.store.ByteBuffersDataOutput;
 import org.apache.lucene.store.ChecksumIndexInput;
 import org.apache.lucene.store.DataAccessHint;
 import org.apache.lucene.store.Directory;
-import org.apache.lucene.store.FileDataHint;
 import org.apache.lucene.store.FileTypeHint;
 import org.apache.lucene.store.IOContext;
 import org.apache.lucene.store.IndexInput;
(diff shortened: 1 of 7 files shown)


lucene/core/src/java/org/apache/lucene/codecs/lucene90/compressing/Lucene90CompressingTermVectorsReader.java
```diff
@@ -149,10 +148,7 @@ public Lucene90CompressingTermVectorsReader(
       final String vectorsStreamFN =
           IndexFileNames.segmentFileName(segment, segmentSuffix, VECTORS_EXTENSION);
       vectorsStream =
-          d.openInput(
-              vectorsStreamFN,
-              context.withHints(
-                  FileTypeHint.DATA, FileDataHint.KNN_VECTORS, DataAccessHint.RANDOM));
+          d.openInput(vectorsStreamFN, context.withHints(FileTypeHint.DATA, DataAccessHint.RANDOM));
```
jpountz, line 151: Note: I removed KNN_VECTORS since these are term vectors, not KNN vectors.

jpountz:
cc @thecoop @ChrisHegarty

ChrisHegarty (approved):
LGTM

thecoop:
Oops - copy/paste error

bharath-techie:
Hi @thecoop @jpountz ,
I've commented the same in https://github.com/apache/lucene/pull/14510/files#r2265722765

```
lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene101/Lucene101PostingsReader.java
lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java

- (removed)
    // Postings have a forward-only access pattern, so pass ReadAdvice.NORMAL to perform
      // readahead.
      docIn = state.directory.openInput(docName, state.context.withReadAdvice(ReadAdvice.NORMAL))

+ (added) 
docIn =
          state.directory.openInput(
              docName, state.context.withHints(FileTypeHint.DATA, FileDataHint.POSTINGS));
```
> comment : This doesn't set a DataAccessHint as it's not RANDOM or SEQUENTIAL, just NORMAL.

In this PR we have removed the change in MMAPDirectory to apply read advice as normal. 

So if no data access hint is provided, it falls back to default read advice which is "random". 

So read advice for this file has changed from default to random ? 

Can you please confirm if I'm missing something and if not, then should we revert back the read advice to normal ?

thecoop:
The logic in `MMapDirectory.toReadAdvice` sets the read advice to `NORMAL` if `FileTypeHint.DATA` is specified. It is here, so this results in a read advice of `NORMAL` being used here

bharath-techie:
@thecoop But that is reverted in this PR, isn't it ? am i missing something ?

image

thecoop:
Ah, yes it does. However, https://github.com/apache/lucene/pull/15040 (also applied to 10.3) changes the default read advice to NORMAL

bharath-techie:
Yes got it thanks :) I just noted this PR.

PR #14642, opened 2025-05-10
Expressions: Avoid creating a dynamic constant with BSM if MethodHandle descriptor can be used

uschindler:
This is an optimization to avoid using a bootstrap when creting the MethodHandle constant. This won't be backported as it's not easy possible to do with ASM.

Basically, if the `MethodHandle` in our function list is a direct one and reachable from system classloader, we push a reference to it directly to stack. This is the common case for static (named) functions and all methods in the default functions. The classfile API makes it easy, because if a `MethodHandle` is direct and reachable by system classloader, it returns a non-empty `describeConstable()`.

This won't improve performance for the execution (as it is a constant before and after), but it should improve startup time.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -33,6 +33,9 @@ Improvements
   This avoids clashes with other ASM versions on classpath because it removes the dependency.
   (Uwe Schindler)
 
+* GITHUB#14642: Expressions: Avoid creating a dynamic constant with BSM if MethodHandle
+  descriptor can be used.  (Uwe Schindler)
+
 * GITHUB#14226: Use optimistic-with-checking KNN Query execution strategy in place of cross-thread
   global queue min-score checking. Improves performance and consistency. (Mike Sokolov, Dzung Bui,
   Ben Trent)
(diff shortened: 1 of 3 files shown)


rmuir (approved):
nice optimization. I imagine this should help some use-cases that don't reuse expression by reducing warmup hit/costs.

uschindler:
There is a problem with that commit, because it does not work with private lookup. Theres an additional check needed.

The problem is that `describeConstable()` uses a full-trusted lookup and not the one of our code. I have another branch here which does this correct and uses a direct static call instead of a method handle, if possible.

This one here looked more elegant, but I forgot to port my test.

uschindler:
This code is more complex, but does the right thing. Only if the MethodHandle is direct and works witho our own lookup, we translate the method call to a invokestatic directly.

This is basically similar to the code we had in the very first version of the Expressions module, but this one can do type checking better and uses the mechanisms of classfile library to generate the opcodes.

uschindler:
I am not really happy with that PR because it adds complexity for nonsense: The improvements are not really there because we move the MethodHandle access checks and cracking to the compiler while with the BSM it happens at linking time. When we create the expression we create and link anyways, so it is not a big difference. One can only say that we don't use a MethodHandle at all if the call is direct, but this is no longer relevant (it was in Java 7 / Java 8).

So I'd like to stay on safe side and close this. The PR can be used for others to get further inspiration - as playground.

uschindler:
I just added the improved test (with a really private method/class) to main: 04126613864

PR #484, opened 2021-11-28
Speed up ECJ tasks by avoiding --release

rmuir:
LUCENE-10185 caused a large performance regression in ECJ tasks by using the --release flag. See graph: http://people.apache.org/~mikemccand/lucenebench/antcleantest.html

We recovered some, but not all, of the performance with https://github.com/apache/lucene/pull/483, but it is still 50% slower.

Instead of using --release, we can just disable "terminal deprecation", and leave this check to `javac`. That single check makes this tool run 50% slower, so it isn't worth it.

I've confirmed that the change fully recovers the build performance drop that was lost from LUCENE-10185:
```
Trunk:
Aggregate task times (possibly running in parallel!):
 160.88 sec.  ecjLintTest
 149.60 sec.  ecjLintMain

Patch:
Aggregate task times (possibly running in parallel!):
 112.95 sec.  ecjLintMain
 105.10 sec.  ecjLintTest
```

PR #14651, opened 2025-05-12
Build: remove hard-coded Java versions from ecj.javadocs.prefs

dsmiley:
This small improvement to the build eliminates references to the Java version in `ecj.javadocs.prefs`.  They were already being set at the CLI when ECJ Lint was invoked.  What _wasn't_ set that was the Java "compliance level", which is different from source & target.  All 3 (compliance, source, target) can simply be set with a hyphen then the version.


gradle/validation/ecj-lint/ecj.javadocs.prefs
```diff
@@ -151,7 +149,6 @@ org.eclipse.jdt.core.compiler.problem.unusedTypeParameter=error
 org.eclipse.jdt.core.compiler.problem.unusedWarningToken=ignore
 org.eclipse.jdt.core.compiler.problem.varargsArgumentNeedCast=error
 org.eclipse.jdt.core.compiler.release=disabled
-org.eclipse.jdt.core.compiler.source=24
```
rmuir, line 154: These need to be set also for IDE integration. It isn't just used for linting.

gradle/validation/ecj-lint.gradle
```diff
@@ -94,10 +96,10 @@ allprojects {
 
         args += [ "-d", "none" ]
 
+        assert project.java.sourceCompatibility == project.java.targetCompatibility
+
         // Compilation environment.
-        // we use -source/-target as it is significantly faster than --release
-        args += [ "-source", project.java.sourceCompatibility ]
-        args += [ "-target", project.java.targetCompatibility ]
+        args += ["-$project.java.sourceCompatibility"] // "compliance" level
```
uschindler, line 102: is this correct as just a number?

gradle/validation/ecj-lint/ecj.javadocs.prefs
```diff
@@ -15,9 +15,7 @@ org.eclipse.jdt.core.compiler.annotation.owning=org.eclipse.jdt.annotation.Ownin
 org.eclipse.jdt.core.compiler.annotation.resourceanalysis=disabled
 org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
 org.eclipse.jdt.core.compiler.codegen.methodParameters=do not generate
-org.eclipse.jdt.core.compiler.codegen.targetPlatform=24
```
uschindler, line 18: if we remove this from here the Eclipse task has a problem. The source and target and compliance is then missing and Eclipse IDE fails to compile.

If we want to remove this here, we would need to add some task to the Eclipse ICE and append the missing lines.

uschindler (changes requested):
The problem is this breaks Eclipse IDE where the file is also used for setting up project.

gradle/validation/ecj-lint/ecj.javadocs.prefs
uschindler: exactly! see my comment below

gradle/validation/ecj-lint.gradle
dsmiley: yes.  See the link to the documentation that I added

gradle/validation/ecj-lint/ecj.javadocs.prefs
uschindler: My proposal would be: Keep the linter task unmodified and just preprocess the properties file for both Eclipse IDE and ECJ.

dsmiley:
I'm hoping an Eclipse user could improve this PR to address that side of the equation.  I've never used Eclipse.

uschindler:
> I'm hoping an Eclipse user could improve this PR to address that side of the equation. I've never used Eclipse.

Not sure how, but if we remove the java version from the ecj javadocs setting file, we have to add the back in the .settings directory for the Eclipse IDE:

https://github.com/apache/lucene/blob/92e0eb8f71e2eaad8c2d18890ed3e506bbfdc5b8/gradle/ide/eclipse.gradle#L96-L100

At this place it processes all files in the `dot.settings` directory and replaces the tag:

https://github.com/apache/lucene/blob/92e0eb8f71e2eaad8c2d18890ed3e506bbfdc5b8/gradle/ide/eclipse/dot.settings/org.eclipse.jdt.core.prefs#L4

So maybe we just need to add the target version here and set those three properties.

Uwe

uschindler:
Something like:

```groovy
filter(ReplaceTokens, tokens: [ 
         'ecj-lint-config': ecjLintFile.getText('UTF-8').replaceAll(/=error\b/, '=' + errorMode)
           +'org.eclipse.jdt.core.compiler.codegen.targetPlatform='+targetVersion /..... and so on
 ]) 
```

uschindler:
But in general: Iw would keep the properties file as is and inject the Java version there. The command line of ECJ should not be modified.

gradle/validation/ecj-lint.gradle
```diff
@@ -94,10 +96,10 @@ allprojects {
 
         args += [ "-d", "none" ]
 
+        assert project.java.sourceCompatibility == project.java.targetCompatibility
+
         // Compilation environment.
-        // we use -source/-target as it is significantly faster than --release
```
dweiss, line 98: I would be actually more keen on using explicit source/target/release options - much easier to understand when you're debugging options at some point. The major problem is that this "compliance" you switched to is actually using the release flag, which - according to the comment above - used to be much slower than just source/target (we did have it at some point).

dsmiley: I didn't see a CLI based "compliance" explicit switch, although "-$version" is documented to set the compliance.  Would setting this *and* source & target work?

> is actually using the release flag

I suppose you are looking at its source code, maybe, to deduce that?

gradle/validation/ecj-lint/ecj.javadocs.prefs
dsmiley: > If we want to remove this here, we would need to add some task to the Eclipse ICE and append the missing lines.

Sounds better than the current practice of a hard-coded Java version number in 3 places in this file!

dsmiley:
> The command line of ECJ should not be modified.

Why; is it perfection?  It seems the "compliance" wasn't being set.

gradle/validation/ecj-lint.gradle
dweiss: Hmm... I'm looking at the link you included in the patch? And I see this:
image
image

So it looks like -[num] is equivalent to setting --source, --target and --release to the same number? Am I reading it wrong?

dweiss: you didn't see the switch because it was commented out - because it was much much slower than source/target and we use javac with java version compliance settings anyway (so anything wrong there would be detected before it hits ecj). I think this duplicated checking + slow execution of compliance mode in ecj was the motivation to comment it out.

dsmiley: Based on the ECJ documentation I linked with that excerpt, it's not going to set "release", so we're safe.  I should bring back the comment reworded as a warning to *not* set "release" and to refer back to #484 for further info.

dweiss: To me, --release is equivalent to "compliance".

gradle/validation/ecj-lint/ecj.javadocs.prefs
uschindler: See my comments below on how to do it. I will push a change to you branch a bit later. I will make sure it reproduces the same settings file for a eclipse IDE as it existed before.
Give me a few hours, watching TV

gradle/validation/ecj-lint.gradle
dweiss: although I just checked and I see no difference on my machine so maybe they've fixed it.
```
main:
  38.53 sec.  ecjLintMain

-release:
  39.06 sec.  ecjLintMain

-24:
  38.40 sec.  ecjLintMain
```

rmuir: To reproduce the slowdown, you need to have a jdk > releaseTarget. E.g. jdk25 with -release 24 set.

dweiss: ah. will try it tomorrow.

uschindler:
I fixed the eclipse setup in https://github.com/apache/lucene/pull/14651/commits/965a00b006a419288a4284be8ef1d2f365484fee

It basically adds the follwoing to the settingss file generated for eclipse config:
```
org.eclipse.jdt.core.compiler.codegen.targetPlatform=24
org.eclipse.jdt.core.compiler.compliance=24
org.eclipse.jdt.core.compiler.source=24
```

To me all is fine, just try to figure out with @dweiss and @rmuir if the release flag is enabled or not. At least in my eclipse settings the "release" is not checked in the options, only source and compliance level are set. Target does not exist in eclipse config.

gradle/validation/ecj-lint/ecj.javadocs.prefs
uschindler: i fixed this in a later commit

gradle/validation/ecj-lint/ecj.javadocs.prefs
uschindler: I fixed this in 965a00b006a419288a4284be8ef1d2f365484fee

uschindler (approved):

rmuir:
> To me all is fine, just try to figure out with @dweiss and @rmuir if the release flag is enabled or not.

Please don't do it here. Please make separate PR if you want to enable that flag: it needs to be backed by benchmarks on JDK > releaseVersion, as the performance issues with it won't show with JDK == releaseVersion.

uschindler:
> > To me all is fine, just try to figure out with @dweiss and @rmuir if the release flag is enabled or not.
> 
> Please don't do it here. Please make separate PR if you want to enable that flag: it needs to be backed by benchmarks on JDK > releaseVersion, as the performance issues with it won't show with JDK == releaseVersion.

I can just give my observation: When the compliance level is set, the Eclipse IDE does not show the "release" setting enabled:
image

uschindler:
But of course it may be different with the command line parameters. But to me it looks like Eclipse disables the release flag and compliance is just a combination of above pictures "source" and "target" settings and version specific preview flags.

rmuir:
Sorry, I don't have a better idea other than, try it on branch_10x, using a JDK24 where the "target" is 21.

Or maybe "template" the file the same way as is done for `gradle eclipse` and then use that resulting processed file for ecj-linting, rather than CLI flags.

uschindler:
Use Jdk 25 EA to check. Jenkins does this all the time.

dweiss:
I dug deep into this, fascinating. So the problem is indeed in the release check flag. Here is a flame graph from a slowed-down execution of ecj:
image

you'll notice a lot of time spent in ```Files.exists``` there, which in turn is called from ```ClasspathJep247Jdk12```. JEP 247 (https://openjdk.org/jeps/247) introduced the ``--release`` option for the compiler and also introduced an - undocumented? - 'ct.sym' file, which is part of the JDK and is a zip file with class signatures from previous Java version. Couldn't find the docs on this file anywhere and ecj's source is a fun reading too:
```
 * Abstraction to the ct.sym file access (see https://openjdk.java.net/jeps/247). The ct.sym file is required to
 * implement JEP 247 feature (compile with "--release" option against class stubs for older releases) and is currently
 * (Java 15) a jar file with undocumented internal structure, currently existing in at least two different format
 * versions (pre Java 12 and Java 12 and later).
 * <p>
 * The only documentation known seem to be the current implementation of
 * com.sun.tools.javac.platform.JDKPlatformProvider and probably some JDK build tools that construct ct.sym file. Root
 * directories inside the file are somehow related to the Java release number, encoded as single digit or letter (single
 * digits for releases 7 to 9, capital letters for 10 and higher).
 * <p>
 * If a release directory contains "system-modules" file, it is a flag that this release files are not inside ct.sym
 * file because it is the current release, and jrt file system should be used instead.
 * <p>
 * All other release directories contain encoded signature (*.sig) files with class stubs for classes in the release.
 * <p>
 * Some directories contain files that are shared between different releases, exact logic how they are distributed is
 * not known.
 * <p>
 * ...
```

Anyway. The cause of the slowdown in ecj is caused by many repetitive calls to zip Filesystem provider, which doesn't seem to be too efficient. There does seem to be some caches around it in CtSym.java but evidently it depends not only on ecj version but also on the content of ct.sym file - how many patch releases there were for the target version (all patch sub-directories need to be checked).

It's quite insane. I would leave release checking out of ecj.

uschindler:
> the compiler and also introduced an - undocumented? - 'ct.sym' file, which is part of the JDK and is a zip file with class signatures from previous Java version. Couldn't find the docs on this file anywhere and ecj's source is a fun reading too:

This is well documented in the source code of the JDK; I don't remember if there's also a JEP. The ct.sym file is generated by the JDK build and later committed when they start a new release. Actually the way how it is generated was my inspiration for the stub-only APIJAR files we have for compilation of MR-JARs. Basically it is more or less exactly the same: The ct.sym file is just a ZIP file with empty stubs and (in addition to Lucene's approach) with some additional logic to only record the differences between API versions. Because of this it is accessing the filesystem provider many times, because it has to load a symbol from many files until it has gotten all symbols by adding up all incremental changes. This is the reason for the file exists. It looks up for every previous Java relaese if a sym file exists. And as we are at 24 already it goes down to 7 or 8. This does not scale, so they should improve the lookup structure by having a pointer to previous release of a specific file in each class stub.

The most important question here: Did you find out if "compliance" flag uses release or not?

uschindler:
P.S.: It could be that the newest version of the ct.sym generator uses the Classfile API in a similar way like I refactored it last week. Have not checked source code, the last time I looked at it, it was NOT using ASM, but the compiler classfile parser. Nowadays all is merged together, so chances are high that it uses Classfile API nowadays.

dweiss:
> The most important question here: Did you find out if "compliance" flag uses release or not?

Looking at the code, it seems to be equivalent to setting the release flag. 

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/Main.java#L1985-L1997

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/Main.java#L2579-L2599

dweiss:
There is a lot of complexity involved in checking this release compatibility. I'd leave it out to javac and just use source/target here. No need to double-check the same stuff twice and slow down the build.

dweiss:
Yep, I'd stick to using what was there before. Not a bad idea to replace java properties if you've already done it though.

uschindler:
> There is a lot of complexity involved in checking this release compatibility. I'd leave it out to javac and just use source/target here. No need to double-check the same stuff twice and slow down the build.

This is not true, I checked the code. When you specify `-release` on command line (and only then), it sets relaese version:

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/85f1797ab9f379eeffd203d94522fbf3c6a4e7ce/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/Main.java#L2588

If you just set compliance level it does not do this,

Later in the code it compares the current compliance level and release version via `(this.releaseVersion != null && this.complianceLevel < jdkLevel)` -- At this place it sets up the "Horrible slowness":

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/85f1797ab9f379eeffd203d94522fbf3c6a4e7ce/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/Main.java#L5191-L5198

So basically, the code in this PR is fine as it does not set the release version unless you use "release". If you just give compliance level it is identical to setting source/target.

uschindler:
But at end I don't care if we set source+target or just compliance. It is effectively doing the same at end. We should just make sure to not set "release".

Just to confirm that my observations are correct: Did you measure slowness with this PR if executed in Java 25? If yes, I am wrong with my analysis.

dweiss:
Ok, I think the second snippet of yours is spot-on - yes. Still, I'd prefer using javac options. I understand them. Everyone who uses javac will. ecj's "numeric" compliance options are a shortcut that save like 10 characters but they're rather obscure. Why bother?

dweiss:
> Just to confirm that my observations are correct: Did you measure slowness with this PR if executed in Java 25? 

I don't see a lot of slowdown at all when I run on Java 21+. It is slightly slower but nowhere near what it used to be. I still don't see the point of using this entire back-compat infrastructure if it's duplicated in javac anyway.

rmuir:
Please don't enable this slowness, there is zero benefit as javac already checks it.

We don't need it. Just avoid it.

gradle/validation/ecj-lint.gradle
rmuir, line 102: Please don't change commandline flags here.

Somehow this issue has turned into... let's slow down the build.

It should be possible to template a string without changing stuff like this and slowing down the build.

rmuir (changes requested):

gradle/validation/ecj-lint.gradle
uschindler: This line does not slow down, so it does not matter here. Unless you use "release" flag. Please read Dawid and my analysis above.

uschindler: Compliance level is translated to source and target. If you pass source and or traget it is vice versa converted to compliance level. Unless you set release, the releaseVersion property is NOT set and the code does not slow down. Please see my comment.

uschindler: Actually the two lines above do the same. The change can be applied or not, the result is identical. At end it sets compliance level.

I agree with Dawid, we can keep source/target as it is easier to understand for javac people, but the whole discussion here is really nonsense. We never wanted to make it slow and we havce heard you all the time complaining often enough: "YES I HEARD YOU! I have said multiple times, the code above won't make it slow. We also tested it with Java 25 and a backported version". So please stop arguing. Thanks!

rmuir: It isn't nonsense: the title of the PR is to "remove hard-coded versions" not "Change ECJ CLI flags".

If gradle can't "template" a file and replace some simple variables without ALSO changing CLI flags, perhaps we should look at a different build tool

uschindler: We already agreed to keep the command line options as is. I just want you to calm down and argue about something we agreed several times.

Sometimes you can do a cleanup or not, what @dsmiley dis was just to replace a command line flag to make the command line shorter.

Actually puting target and source on the command line will set compliance level, too. So it won't need to be included in the config file. It was always obsolete and was only included for the Eclipse config. So see this two lines just as a cleanup. We can argue several months about it, but I would no longer listen to you. Sorry. Complete nonsense!

Therefore I added a commit yesterday to set the compliance level and source/target level in the eclipse configuration. The ECJ linter uses the property from command line, which is cheaper to do than patching a file (I don't want to create random temp files for nonsense as sthis slows down on Windows!).

rmuir: i dont think we need to do any CLI flags change: That's exactly what I'm requesting to revert here. And thats why the PR has 40+ comments.

If we were doing this with `make` we'd just `envsubst` and the hardcoded value would be replaced by variable: problem solved. 

I don't understand why it should be any more difficult than that with gradle.

uschindler: Man man. The change here is only cleanup. There is no need to change any files. The removal of the propertie sin the file are not needed here. It was already fixed by my commit.

The change here is..... ONLY COSMETIC. It does not change anything in functionality. So we can leave it out. Please stop arguing. This is annoying to me. Sorry.

uschindler:
I am complete pissed off, bye! Do whatever you want. Damn!

gradle/validation/ecj-lint.gradle
uschindler: reverted. Man man.

uschindler (approved):

dweiss (approved):

rmuir (approved):

dsmiley:
Thanks Dawid for the deep dive and Uwe for helping with the Eclipse support!  
For this minor build improvement, I wasn't going to update CHANGES.txt but let me know if you think I should add it.  I plan to back-port to 10x and also take it to Solr.  I'll merge later tonight (like in ~8hrs).

PR #14653, opened 2025-05-12
build: configure 3 retries for all gradle Download tasks

### LINKED ISSUE #14652: improve reliability of de.undercouch downloads
### Description

Downloads using the tool have been failing in Jenkins rarely.
The tool supports retries, but they are disabled by default:
```
retries
    Specifies the maximum number of retry attempts if a request has failed. By default, requests are never retried and the task fails immediately if the first request does not succeed. If the value is greater than 0, failed requests are retried regardless of the actual error. This includes failed connection attempts and file-not-found errors (404). A negative value means infinite retries. (default: 0)
```

I think if we configure this to a reasonable value such as 2 or 3, it may really help.

https://github.com/michel-kraemer/gradle-download-task

rmuir:
The default value of retries is 0. Set it to 3, to improve reliability of downloads during builds.

Closes #14652 

Please review, and make sure I didn't miss any `Download` tasks. The goal was just to add this to all of them, in the hopes that if someone adds a new Download task they will continue to follow the pattern.

### DIFF
--- a/gradle/hacks/downloader-retries.gradle
+++ b/gradle/hacks/downloader-retries.gradle
@@ -1,0 +1,28 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


dweiss:
This can be done by applying a single configuration to a task type, Rob (tasks.withType(Download).configureEach...). I'm away from the computer but will tweak it tomorrow.

dweiss:
something like this -
https://github.com/apache/lucene/blob/main/gradle/hacks/gradle-archives.gradle

gradle/hacks/downloader-retries.gradle
```diff
@@ -0,0 +1,27 @@
...
+allprojects {project ->
+    // Limit configuration to just those projects that actually have the plugin enabled.
+    plugins.matching { Plugin plugin ->
+        return plugin.class.name == "de.undercouch.gradle.tasks.download.DownloadTaskPlugin"
+    }.configureEach {
+        project.tasks.withType(Download).configureEach {
+            it.retries 3
+        }
+    }
+}
```
rmuir, line 27: Thank you @dweiss. Much cleaner. Hopefully it stops the annoying build failures!

PR #14654, opened 2025-05-13
Override ValueSource.FromDoubleValuesSource.getSortField

dsmiley:
Thus reducing indirection and supporting needsScores correctly (fixes a bug).

The added test, without the fix, would throw an assertion inside DoubleValuesSource [here](https://github.com/apache/lucene/blob/304d4e7855deb39b4650d954d027ce8697873056/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSource.java#L329) because the required scores were not provided to the DoubleValuesSource.getValues implementation.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -113,7 +113,8 @@ Optimizations
 
 Bug Fixes
 ---------------------
-(No changes)
+* GITHUB#14654: ValueSource.fromDoubleValuesSource(dvs).getSortField() would throw errors when
+  used if the DoubleValuesSource needed scores. (David Smiley)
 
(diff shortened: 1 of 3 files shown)


mkhludnev (approved):

dsmiley:
Thanks for the review.I added a CHANGES.txt.  I'll merge in another day or two.  I may backport to 9.x as well -- why not.

ChrisHegarty:
@dsmiley Given that this change has been backported to 9.12.2, which will release in tandem with 10.2.2, do you want to port this change to `branch_10_2` also?

dsmiley:
Done.

PR #14661, opened 2025-05-13
Updating skip-changelog label

jainankitk:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,7 +7,7 @@ updates:
       interval: monthly
     commit-message:
       prefix: ci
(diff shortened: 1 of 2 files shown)


msfroh:
While you're at it, you may want to add the label to the `dependabot.yml` config, so that the label automatically gets added to dependabot PRs.

jainankitk:
> While you're at it, you may want to add the label to the `dependabot.yml` config, so that the label automatically gets added to dependabot PRs.

Good point, done!

dweiss (approved):

stefanvodita:
I created the new label and added it to PRs that were using the old label.

PR #14665, opened 2025-05-14
Fix termination condition in TestStressNRTReplication.

### LINKED ISSUE #14664: TestStressNRTReplication may never terminate (exceed suite timeout)
### Description

It hangs in the 'restarter' thread on this condition:
```
        while (startupThreads.size() > 0) {
          Thread.sleep(10);
        }
```
the main thread just joins the restarted and never ends. 

The bug is easy to show if you add a sleep before startupThreads.add(t); it is currently invoked after the sub-thread has been started - this means that it is possible the sub-thread removes itself from the startupThreads list before it's added to it, leaving the list in an inconsistent state (and never reach an empty state).

_No response_

dweiss:
Cleans up the code a bit too. Fixes #14664

### DIFF
--- a/lucene/replicator/src/test/org/apache/lucene/replicator/nrt/TestStressNRTReplication.java
+++ b/lucene/replicator/src/test/org/apache/lucene/replicator/nrt/TestStressNRTReplication.java
@@ -269,7 +269,7 @@ public class TestStressNRTReplication extends LuceneTestCase {
     while (failed.get() == false && System.nanoTime() < endTime) {
 
       // Wait a bit:
-      sleep(TestUtil.nextInt(random(), Math.min(runTimeSec * 4, 200), runTimeSec * 4));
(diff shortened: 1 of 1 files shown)


lucene/replicator/src/test/org/apache/lucene/replicator/nrt/TestStressNRTReplication.java
```diff
@@ -994,26 +998,26 @@ public void run() {
...
                           startupThreads.remove(Thread.currentThread());
+                          message("N" + idx + ": top: removed thread");
                         }
                       }
                     };
+                startupThreads.add(t);
                 t.setName("start R" + idx);
                 t.start();
-                startupThreads.add(t);
               }
```
dweiss, line 1008: This is the actual fix.

benwtrent (approved):

dweiss:
Thanks, @benwtrent

PR #14667, opened 2025-05-14
Fix documentation regarding benchmark running

viliam-durina:
`task.alg` was probably renamed to `taskAlg` some time ago. `task.alg` is ignored and `micro-standard.alg` is run instead of the desired benchmark.

This PR edits only documentation, no code edited.


jainankitk (approved):
Thanks @viliam-durina for this change. Can you add changelog entry as well, before this PR is merged?

viliam-durina:
@jainankitk Do you really want changelogs for this kind of changes? Nevertheless, I added it.

PR #14672, opened 2025-05-15
Dynamic pruning with DocValueSkipper

### LINKED ISSUE #14666: Dynamic pruning with DocValue skipper
### Description

We are requiring point tree index to get the dynamic pruning optimization. Can we take advantage of sparse index to do this as well? Several benefits i can think of:

* Work without point tree index.
* Save big bitset allocations.
* Respect `maxDocVisited` naturally.

gf2121:
closes #14666


iverase (approved):
Lgtm

lucene/core/src/java/org/apache/lucene/search/comparators/NumericComparator.java
```diff
@@ -328,120 +507,47 @@ private void updateSkipInterval(boolean success) {
...
     }
 
-    protected abstract long bottomAsComparableLong();
-
-    protected abstract long topAsComparableLong();
+    @Override
+    void doUpdateCompetitiveIterator() {
+      TwoPhaseIterator twoPhaseIterator =
+          new DocValuesRangeIterator(innerTwoPhase, skipper, minValueAsLong, maxValueAsLong, false);
+      // TODO this twoPhaseIterator is wrapped by #competitiveIterator() so caller can not use
```
martijnvg: I don't think this can be addressed, as long as `LeafFieldComparator#competitiveIterator()` has `DocIdSetIterator` as return type?

martijnvg (approved):
LGTM

lucene/core/src/java/org/apache/lucene/search/comparators/NumericComparator.java
gf2121: Thanks for review!

Before is possible as caller may use `TwoPhaseIterator#unwrap` but it is impossible now as we are moving to the new `UpdateableDocIdSetIterator` fashion. I removed this TODO.

PR #14674, opened 2025-05-15
Optimize AbstractKnnVectorQuery#createBitSet with intoBitset

gf2121:
This tries to implement `intoBitset` to speed up the building of BitSet.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
@@ -42,6 +42,7 @@ import org.apache.lucene.search.knn.TopKnnCollectorManager;
 import org.apache.lucene.util.BitSet;
 import org.apache.lucene.util.BitSetIterator;
 import org.apache.lucene.util.Bits;
+import org.apache.lucene.util.FixedBitSet;
 
 /**
  * Uses {@link KnnVectorsReader#search} to perform nearest neighbour search.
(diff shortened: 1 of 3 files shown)


lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -229,6 +230,16 @@ private BitSet createBitSet(DocIdSetIterator iterator, Bits liveDocs, int maxDoc
...
+            public void intoBitSet(int upTo, FixedBitSet bitSet, int offset) throws IOException {
+              assert bitSet.scanIsEmpty()
+                  : "By contract we should not clear bits in this bitset, but it should be empty here.";
+              iterator.intoBitSet(upTo, bitSet, offset);
+              if (liveDocs != null) {
+                liveDocs.applyMask(bitSet, offset);
+              }
+            }
           };
       return BitSet.of(filterIterator, maxDoc);
```
jpountz, line 244: Instead of implementing `intoBitSet` on the iterator, we could do `BitSet bitSet = BitSet.of(iterator, maxDoc);` and then clear deleted docs with `liveDocs.applyMask(bitSet, 0);`? Ie. there wouldn't be a `FilterDocIdSetIterator` anymore?

jpountz:
It looks like a similar change can be made to `AbstractVectorSimilarityQuery`.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
gf2121: Thanks for feedback! 

This is what i had in mind initially. The problem here is `applyMask` is requiring a `FixedBitSet` while `BitSet#of` may return a `SparseFixedBitSet`. So i have to fork the `Bits#applyMask` code for the sparse case, which could look like:

```
BitSet bitSet = BitSet.of(iterator, maxDoc);
if (liveDocs != null) {
  if (liveDocs instanceof FixedBitSet fixedBitSet) {
    liveDocs.applyMask(fixedBitSet, 0);
  } else {
    for (int i = bitSet.nextSetBit(0);
         i != DocIdSetIterator.NO_MORE_DOCS;
         i = i + 1 >= bitSet.length() ? DocIdSetIterator.NO_MORE_DOCS : bitSet.nextSetBit(i + 1)) {
      if (bitSet.get(i) == false) {
        bitSet.clear(i);
      }
    }
  }
}
return bitSet;
```

Or another potential way i'm thinking is making `Bits#applyMask` accept a `BitSet`. Default impl works fine but  specialized impls need an `instanceof FixedBitset` check. 

What do you think?

jpountz: Oh, this is annoying... It's not great either but I guess I'd like to do the SparseFixedBitSet/FixedBitSet decision up-front, something like:

```java
int threshold = maxDoc >> 7; // like BitSet#of
if (it.cost() >= threshold) {
  // Create a FixedBitSet in an optimized way
} else {
  // existing code
}
```

jpountz (approved):

PR #14676, opened 2025-05-15
Added toString() method to BytesRefBuilder

shubhamsrkdev:
## Problem
- The BytesRefBuilder was missing toString() override. This was leading to UnsupportedOperationException instead of IllegalArgumentException in PointInSetQuery's constructor as described here - https://github.com/apache/lucene/issues/14161

## Solution
- Added toString() method in BytesRefBuilder.
- Added UT for verifying the IllegalArgumentException exception.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -49,6 +49,9 @@ Bug Fixes
 
 * GITHUB#14075: Remove duplicate and add missing entry on brazilian portuguese stopwords list. (Arthur Caccavo)
 
+* GITHUB#14161: PointInSetQuery's constructor now throws IllegalArgumentException
+  instead of UnsupportedOperationException when values are out of order. (Shubham Sharma)
+
 Changes in Runtime Behavior
 ---------------------
 * GITHUB#14187: The query cache is now disabled by default. (Adrien Grand)
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -90,6 +90,8 @@ Improvements
 * GITHUB#14602: Refactor the expressions compiler to use official ClassData BSM with indexed lookup
   (Uwe Schindler)
 
+  * GITHUB#14443: Support adaptive refresh in Searcher Managers (Vigya Sharma)
+
 Optimizations
 ---------------------
 * GITHUB#14418: Quick exit on filter query matching no docs when rewriting knn query. (Pan Guixin)
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -49,9 +49,6 @@ Bug Fixes
 
 * GITHUB#14075: Remove duplicate and add missing entry on brazilian portuguese stopwords list. (Arthur Caccavo)
 
-* GITHUB#14161: PointInSetQuery's constructor now throws IllegalArgumentException
-  instead of UnsupportedOperationException when values are out of order. (Shubham Sharma)
-
 Changes in Runtime Behavior
 ---------------------
 * GITHUB#14187: The query cache is now disabled by default. (Adrien Grand)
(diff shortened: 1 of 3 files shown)


msfroh (approved):
Nice, thank you @shubhamsrkdev!

msfroh:
Oh, @shubhamsrkdev, could you please add an entry to lucene/CHANGES.txt? This is a fix that people will be able to experience.

shubhamsrkdev:
Thanks @msfroh! I have added the change log entry.

jpountz (approved):

lucene/CHANGES.txt
```diff
@@ -49,6 +49,9 @@ Bug Fixes
 
 * GITHUB#14075: Remove duplicate and add missing entry on brazilian portuguese stopwords list. (Arthur Caccavo)
 
+* GITHUB#14161: PointInSetQuery's constructor now throws IllegalArgumentException
```
vigyasharma, line 52: We can put this in 10.3, doesn't need to wait for 11.0

lucene/core/src/test/org/apache/lucene/search/TestPointQueries.java
```diff
@@ -2599,4 +2599,33 @@ public void testPointInSetQuerySkipsNonMatchingSegments() throws IOException {
...
+                    };
+                    int index = 0;
+
+                    @Override
+                    public BytesRef next() {
+                      return index < values.length ? values[index++] : null;
+                    }
+                  }) {
+                @Override
+                protected String toString(byte[] point) {
```
vigyasharma, line 2624: Why did we need this override?

lucene/core/src/java/org/apache/lucene/util/BytesRefBuilder.java
```diff
@@ -171,4 +171,9 @@ public boolean equals(Object obj) {
   public int hashCode() {
     throw new UnsupportedOperationException();
   }
+
+  @Override
+  public String toString() {
+    return this.get().toString();
```
vigyasharma, line 177: This will return hex encoded bytes. Is that okay for the exception message? Should we use `utf8ToString` instread?

lucene/core/src/test/org/apache/lucene/search/TestPointQueries.java
```diff
@@ -2599,4 +2599,33 @@ public void testPointInSetQuerySkipsNonMatchingSegments() throws IOException {
     w.close();
     dir.close();
   }
+
+  public void testOutOfOrderValuesInPointInSetQuery() throws Exception {
```
vigyasharma, line 2603: Thanks for adding this test!

lucene/core/src/java/org/apache/lucene/util/BytesRefBuilder.java
dweiss: I don't think we can assume it's always utf-8, let's stick to hex?

lucene/core/src/test/org/apache/lucene/search/TestPointQueries.java
shubhamsrkdev: The `PointInSetQuery` is an abstract class. Following the pattern from [here](https://github.com/apache/lucene/blob/b2547588f1672db162158161cf00085300371c5e/lucene/core/src/test/org/apache/lucene/search/TestPointQueries.java#L2064)

msfroh (approved):

jpountz:
@msfroh Should this be backported to branch_10x?

msfroh:
> @msfroh Should this be backported to branch_10x?

Oh -- it probably should. I'll take care of that.

jpountz:
@msfroh It looks like the CHANGES entry needs to move to 10.4 as well?

msfroh:
Okay -- backported this and moved the CHANGES entry. Thanks, @jpountz!

Issue #12071, opened 2023-01-10
Can we better take advantage of compact strings?


jpountz:
### Description

There's a non-negligible time that we spend on UTF-16 / UTF-8 conversions using our own `UnicodeUtil`, e.g. via the `BytesRef(String)` constructor. But since the introduction of compact strings, `String#getBytes(StandardCharsets.UTF_8)` has some fast tracks, e.g. if neither of the bytes has its highest bit set, then the string is an ASCII string that is the same as the UTF-8 representation.

I ran a quick microbenchmark that suggests that `String#getBytes` can indeed be significantly faster on ASCII strings:
 - `charsetEncoder` leverages `StandardCharsets.UTF_8.encode(state.input)`
 - `stringGetBytes` leverages `String#getBytes(StandardCharsets.UTF_8)`.
 - `unicodeUtil` leverages `UnicodeUtil#UTF16toUTF8`.

```
Benchmark                                    (input)   Mode  Cnt    Score    Error   Units
ConversionBenchmark.charsetEncoder                 a  thrpt    5   36.890 ±  0.572  ops/us
ConversionBenchmark.charsetEncoder  abcdefghijklmnop  thrpt    5   18.717 ±  1.367  ops/us
ConversionBenchmark.charsetEncoder         recherché  thrpt    5   10.098 ±  0.328  ops/us
ConversionBenchmark.stringGetBytes                 a  thrpt    5  142.186 ± 18.849  ops/us
ConversionBenchmark.stringGetBytes  abcdefghijklmnop  thrpt    5  111.259 ±  2.203  ops/us
ConversionBenchmark.stringGetBytes         recherché  thrpt    5   53.565 ±  0.483  ops/us
ConversionBenchmark.unicodeUtil                    a  thrpt    5  103.123 ±  3.970  ops/us
ConversionBenchmark.unicodeUtil     abcdefghijklmnop  thrpt    5   54.223 ±  1.342  ops/us
ConversionBenchmark.unicodeUtil            recherché  thrpt    5   58.166 ±  1.504  ops/us
```

[...]

PR #14678, opened 2025-05-16
Improve BytesRef creation from String

schlosna:
### Description

When creating a `BytesRef` from a String, we can leverage `String::getBytes(StandardCharset.UTF_8)` to avoid 3x memory allocation overhead when the `String` is already `UTF-8` and when the [JDK provides compact `String` representation as `byte[]`](https://openjdk.org/jeps/254) this becomes a single array copy rather than more complex
`org.apache.lucene.util.UnicodeUtil#UTF16toUTF8` logic. Benchmark results for the added `BytesRefBenchmark` on MacBookPro M2 Max show specialized `BytesRef(String)` constructor provides 2x to 6x throughput improvement compared to `BytesRef(CharSequence)` constructor, and 3x less allocations for ASCII Strings.

```
# Apple MacBookPro M2 Max
# JMH version: 1.37
# VM version: JDK 24.0.1, OpenJDK 64-Bit Server VM, 24.0.1+9-FR
# VM invoker: /Library/Java/JavaVirtualMachines/amazon-corretto-24.jdk/Contents/Home/bin/java
# VM options: -XX:+UseParallelGC -XX:TieredStopAtLevel=1 -XX:ActiveProcessorCount=1 -Dfile.encoding=UTF-8 -Duser.country=US -Duser.language=en -Duser.variant
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 3 iterations, 3 s each
# Measurement: 5 iterations, 3 s each

                                                                         (before)              (after)
                                                                       bytesRefCharSequence   bytesRefString
Benchmark                             (length)      (type)   Mode  Cnt      Score     Error      Score     Error   Units
BytesRefBenchmark                         1000       ASCII  thrpt    5      1.213 ±   0.022      6.090 ±   0.145  ops/us
BytesRefBenchmark:gc.alloc.rate.norm      1000       ASCII  thrpt    5   3040.002 ±   0.001   1040.000 ±   0.001    B/op
BytesRefBenchmark                         1000  ISO_8859_1  thrpt    5      1.123 ±   0.022      3.334 ±   0.116  ops/us
BytesRefBenchmark:gc.alloc.rate.norm      1000  ISO_8859_1  thrpt    5   3040.002 ±   0.001   3552.001 ±   0.001    B/op
BytesRefBenchmark                         1000   UTF_8_BMP  thrpt    5      0.660 ±   0.017      1.548 ±   0.092  ops/us
BytesRefBenchmark:gc.alloc.rate.norm      1000   UTF_8_BMP  thrpt    5   3040.004 ±   0.001   5968.002 ±   0.001    B/op
BytesRefBenchmark                         1000      UTF_16  thrpt    5      0.435 ±   0.009      0.811 ±   0.068  ops/us
BytesRefBenchmark:gc.alloc.rate.norm      1000      UTF_16  thrpt    5   5888.006 ±   0.001   9848.003 ±   0.001    B/op
```

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -41,6 +41,7 @@ Optimizations
 ---------------------
 * GITHUB#14011: Reduce allocation rate in HNSW concurrent merge. (Viliam Durina)
 * GITHUB#14022: Optimize DFS marking of connected components in HNSW by reducing stack depth, improving performance and reducing allocations. (Viswanath Kuchibhotla)
+* GITHUB#14678: Optimize BytesRef creation from Strings, improving throughput and reducing allocations. (David Schlosnagle)
 
 Bug Fixes
 ---------------------
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/StemmerOverrideFilter.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/StemmerOverrideFilter.java
@@ -220,7 +220,7 @@ public final class StemmerOverrideFilter extends TokenFilter {
         int id = sort[i];
         BytesRef bytesRef = hash.get(id, spare);
         intsSpare.copyUTF8Bytes(bytesRef);
-        fstCompiler.add(intsSpare.get(), new BytesRef(outputValues.get(id)));
+        fstCompiler.add(intsSpare.get(), BytesRef.of(outputValues.get(id)));
       }
       return new StemmerOverrideMap(
           FST.fromFSTReader(fstCompiler.compile(), fstCompiler.getFSTReader()), ignoreCase);
--- a/lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/BytesRefBenchmark.java
+++ b/lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/BytesRefBenchmark.java
@@ -1,0 +1,103 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.benchmark.jmh;
+
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+import org.apache.lucene.util.BytesRef;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.profile.GCProfiler;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+@State(Scope.Thread)
+@Fork(1)
+@Threads(4)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@BenchmarkMode(Mode.Throughput)
+@Warmup(iterations = 3, time = 3)
+@Measurement(iterations = 5, time = 3)
+public class BytesRefBenchmark {
+
+  @Param({"10", "100", "1000"})
+  private int length;
+
+  @Param({"ASCII", "ISO_8859_1", "UTF_8_BMP", "UTF_16"})
+  private Type type;
+
+  private String string;
+  private CharSequence charSequence;
+
+  @Setup
+  public void before() {
+    string = generateCodepoints(length, type.maxCodePoint);
+    charSequence = string;
+  }
+
+  @Benchmark
+  public BytesRef bytesRefCharSequence() {
+    return new BytesRef(charSequence);
+  }
+
+  @Benchmark
+  public BytesRef bytesRefString() {
+    return new BytesRef(string);
+  }
+
+  private static String generateCodepoints(int size, int maxCodePoint) {
+    Random r = new Random(0);
+    StringBuilder sb = new StringBuilder(size);
+    for (int i = 0; i < size; i++) {
+      sb.appendCodePoint(r.nextInt(maxCodePoint));
+    }
+    return sb.toString();
+  }
+
+  public enum Type {
+    ASCII(128),
+    ISO_8859_1(256),
+    UTF_8_BMP(0xFFFF),
+    UTF_16(0x10FFFF),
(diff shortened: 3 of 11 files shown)


msfroh:
That's a nice improvement! 

Presumably this kicks in when we have [compact strings](https://openjdk.org/jeps/254)? Do you have a JMH benchmark with non-ISO-8859-1/Latin-1 characters?

schlosna:
> Presumably this kicks in when we have [compact strings](https://openjdk.org/jeps/254)? Do you have a JMH benchmark with non-ISO-8859-1/Latin-1 characters?

Yes, this leverages compact strings for Latin-1 for the best case scenario here. I'll throw together a more rigorous benchmark including non-Latin1.

schlosna:
Added `BytesRefBenchmark` demonstrating existing `new BytesRef(CharSequence)` vs. `new BytesRef(String)` demonstrating 2x to 6x throughput improvement on AMD EPYC 7R13 Processor:

```
# AMD EPYC 7R13 Processor
# JMH version: 1.37
# VM version: JDK 24.0.1, OpenJDK 64-Bit Server VM, 24.0.1+9-FR
# VM invoker: /home/dev/.gradle/jdks/amazon-corretto-24.0.1.9.1-linux-x64/bin/java
# VM options: -XX:+UseParallelGC -XX:TieredStopAtLevel=1 -XX:ActiveProcessorCount=1 -Dfile.encoding=UTF-8 -Duser.country=US -Duser.language=en -Duser.variant
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 3 iterations, 3 s each
# Measurement: 5 iterations, 3 s each

                                                      (before)             (after)
                                                    bytesRefCharSequence   bytesRefString
Benchmark          (length)      (type)   Mode  Cnt      Score     Error     Score     Error   Units
BytesRefBenchmark        10       ASCII  thrpt    5     75.650 ±   4.078   180.382 ±   7.672  ops/us
BytesRefBenchmark        10  ISO_8859_1  thrpt    5     66.688 ±   3.957   114.655 ±   7.510  ops/us
BytesRefBenchmark        10   UTF_8_BMP  thrpt    5     46.678 ±   2.387    77.036 ±   2.484  ops/us
BytesRefBenchmark        10      UTF_16  thrpt    5     26.289 ±   1.376    49.021 ±   1.740  ops/us
BytesRefBenchmark       100       ASCII  thrpt    5      9.106 ±   0.487    47.246 ±   1.218  ops/us
BytesRefBenchmark       100  ISO_8859_1  thrpt    5      8.447 ±   0.628    23.809 ±   0.488  ops/us
BytesRefBenchmark       100   UTF_8_BMP  thrpt    5      5.129 ±   0.141    10.431 ±   0.336  ops/us
BytesRefBenchmark       100      UTF_16  thrpt    5      2.963 ±   0.127     5.904 ±   0.271  ops/us
BytesRefBenchmark      1000       ASCII  thrpt    5      0.917 ±   0.038     5.891 ±   0.428  ops/us
BytesRefBenchmark      1000  ISO_8859_1  thrpt    5      0.830 ±   0.024     1.981 ±   0.077  ops/us
BytesRefBenchmark      1000   UTF_8_BMP  thrpt    5      0.468 ±   0.014     0.950 ±   0.057  ops/us
BytesRefBenchmark      1000      UTF_16  thrpt    5      0.298 ±   0.011     0.537 ±   0.021  ops/us
```

lucene/CHANGES.txt
```diff
@@ -41,6 +41,7 @@ Optimizations
 ---------------------
 * GITHUB#14011: Reduce allocation rate in HNSW concurrent merge. (Viliam Durina)
 * GITHUB#14022: Optimize DFS marking of connected components in HNSW by reducing stack depth, improving performance and reducing allocations. (Viswanath Kuchibhotla)
+* GITHUB#14678: Optimize BytesRef creation from Strings, improving throughput and reducing allocations. (David Schlosnagle)
```
vigyasharma, line 45: This could go in 10.3, any reason to keep this for 11.0?

vigyasharma (commented):
Nice, impressive ~6x improvement with ASCII strings! 

Is the ~1.8x improvement with `UTF_16` strings also expected? I thought this worked on compact strings that need 1 byte per char?

rmuir:
There's intentionally not a `BytesRef(String)` constructor, as there have been issues before.

schlosna:
> There's intentionally not a `BytesRef(String)` constructor, as there have been issues before.

@rmuir Do you happen to recall what issues a `BytesRef(String)` constructor create and any history there? 

I imagine you're referring to the case where the provided text is not "well-formed unicode text, with no unpaired surrogates" as defined in the `BytesRef(CharSequence)` constructor, and where `org.apache.lucene.util.UnicodeUtil#validUTF16String(java.lang.CharSequence)` returns false. `UnicodeUtil.UTF16toUTF8` used by `BytesRef(CharSequence)` replaces unpaired surrogates with `�` bytes `0xEF 0xBF 0xBD` while the `BytesRef(String)` constructor as proposed does no replacement or validation.

https://github.com/apache/lucene/blob/b0f992369ba967341d9b51f05f41ab25eb177e1d/lucene/core/src/java/org/apache/lucene/util/BytesRef.java#L77-L81

https://github.com/apache/lucene/blob/b0f992369ba967341d9b51f05f41ab25eb177e1d/lucene/core/src/java/org/apache/lucene/util/UnicodeUtil.java#L168-L172


In https://github.com/apache/lucene/issues/12071 these is mention https://github.com/apache/lucene/issues/12071#issuecomment-1379313710 of using the vector APIs to speed up UnicodeUtil conversions. Has any of that been approached in lucene yet? I imagine a SIMD or SWAR approach like https://lemire.me/blog/2018/05/16/validating-utf-8-strings-using-as-little-as-0-7-cycles-per-byte/ & https://lemire.me/blog/2023/02/16/computing-the-utf-8-size-of-a-latin-1-string-quickly-avx-edition/ would be worth investigating.

rmuir:
git log tells the story. A quick summary:

* conflicting use cases such as usages as a reference vs stringbuilder/string type uses
* ownership issues and API problems resulting from that.

The use-cases were torn apart and the code cleaned up and bytesrefbuilder made separate. we shouldn't go backwards.

this is basically a reference class, intended to support pointing to byte[]s held elsewhere. like a pointer in c lol.

rmuir:
> In #12071 these is mention [#12071 (comment)](https://github.com/apache/lucene/issues/12071#issuecomment-1379313710) of using the vector APIs to speed up UnicodeUtil conversions. Has any of that been approached in lucene yet? I imagine a SIMD or SWAR approach like https://lemire.me/blog/2018/05/16/validating-utf-8-strings-using-as-little-as-0-7-cycles-per-byte/ & https://lemire.me/blog/2023/02/16/computing-the-utf-8-size-of-a-latin-1-string-quickly-avx-edition/ would be worth investigating.

If you want to speed something up, pay no mind to java.lang.String as it isn't involved in the indexing anyway. Look here which is what the indexer calls jazillions of times https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/analysis/tokenattributes/CharTermAttributeImpl.java#L91-L95

it is just char[]->byte[] conversion, with the same reused char[] and byte[], there aren't allocations or strings or any of that.

schlosna:
> > In #12071 these is mention [#12071 (comment)](https://github.com/apache/lucene/issues/12071#issuecomment-1379313710) of using the vector APIs to speed up UnicodeUtil conversions. Has any of that been approached in lucene yet? I imagine a SIMD or SWAR approach like https://lemire.me/blog/2018/05/16/validating-utf-8-strings-using-as-little-as-0-7-cycles-per-byte/ & https://lemire.me/blog/2023/02/16/computing-the-utf-8-size-of-a-latin-1-string-quickly-avx-edition/ would be worth investigating.
> 
> If you want to speed something up, pay no mind to java.lang.String as it isn't involved in the indexing anyway. Look here which is what the indexer calls jazillions of times https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/analysis/tokenattributes/CharTermAttributeImpl.java#L91-L95
> 
> it is just char[]->byte[] conversion, with the same reused char[] and byte[], there aren't allocations or strings or any of that.

Thanks, I'm going to close out this PR for now. When I have a chance I will take a look at optimizing the `char[]` to `byte[]` conversions via `jdk.incubator.vector.ShortVector.fromCharArray`.

rmuir:
yeah it is tricky, since the 'strings' indexed in search are usually small: words. for a lot of natural languages average word length is already small (e.g. english: ~5), and often in search people are shortening that even more (with stemmers etc).

PR #14688, opened 2025-05-20
Fix patience knn queries to work with seeded knn queries

tteofili:
This fixes `PatienceKnnVectorQuery` to work better with `SeededKnnVectorQuery`, especially given the changes in #14226.
This makes sure those queries are tested together and query rewriting works properly.


ChrisHegarty (approved):
LGTM

Issue #7865, opened 2015-09-15
AbstractRangeQueryNode toQueryString not working as intended [LUCENE-6807]


asfimport:
It is my understanding that for a given `QueryNode` node, `parse(node.toQueryString());` should return a `QueryNode` which is functionally identical to the original node.

That is not the case with AbstractQueryNode:
if we have a range query on FIELD from "A" to "B" (`node = parse("FIELD:[A B]")`), then `node.toQueryString()` will return "[FIELD:A FIELD:B]" which, in turn, is parsed as a range query on the default field from "FIELD:A" to "FIELD:B".

As far as I know, this affects all versions of lucene.

I believe I have the knowledge to provide the patch, so I will be working on that today.

As of now, I have thought of two options to implement this fix, both of which involve modifying the `ValueQueryNode` interface to include a method which returns `value` as a `CharSequence`. 

The first option is to add a new method to the interface which returns (formatted if necessary) the value as a `CharSequence` and implement it in all implementing classes (`FieldQueryNode` and `NumericQueryNode`). Then in `AbstractQueryNode#toQueryString()` we will call that method and escape the values using the provided `EscapeQuerySyntax`.

The second option is to make the protected method `getTermEscaped()`, which is already present in all implementing classes, public, and add it to the interface.

While I think that the second option is certainly cleaner, I do not know why this method is protected in the first place, so I will proceed with the first option until someone who is more familiar with the lucene project than I am can comment on the matter.

As I am writing this, it occurs to me that implementing the first option is essentially just bypassing the protected scope on `getTermEscapeed()`, so maybe it is correct to just review whether or not that method needs to be protected. It also occurred to me to use the existing `getValue()` method, and format and escape the `toString()` of that, but that depends on a generic class having implemented `toString()` in a way that plays nicely with the query parser, so, to me, this is below the two previously mentioned options.

---
Migrated from [LUCENE-6807](https://issues.apache.org/jira/browse/LUCENE-6807) by Peter Barna, 1 vote, updated May 17 2020
[...]

PR #14697, opened 2025-05-21
Fix AbstractRangeQueryNode#toQueryString

phb-ig:
Re: #7865 

It now returns a string which is valid Lucene range query syntax and can be parsed back into the original node.

Added public method `getTermEscaped(EscapeQuerySyntax)` to `ValueQueryNode` and made existing methods in `Field/PointQueryNode` public.

Added tests to verify.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -862,7 +862,7 @@ Improvements
 
 * GITHUB#13285: Early terminate graph searches of AbstractVectorSimilarityQuery to follow timeout set from
   IndexSearcher#setTimeout(QueryTimeout). (Kaival Parikh)
-
+  
 * GITHUB#13633: Add ability to read/write knn vector values to a MemoryIndex. (Ben Trent)
 
 * GITHUB#12627: patch HNSW graphs to improve reachability of all nodes from entry points
@@ -1779,7 +1779,7 @@ New Features
   closed while queries are running can no longer crash the JVM. To disable this feature,
   pass the following sysprop on Java command line:
   "-Dorg.apache.lucene.store.MMapDirectory.enableMemorySegments=false" (Uwe Schindler)
-
+  
 * GITHUB#12252 Add function queries for computing similarity scores between knn vectors. (Elia Porciani, Alessandro Benedetti)
 
 Improvements
@@ -2458,7 +2458,7 @@ New Features
 * LUCENE-10385: Implement Weight#count on IndexSortSortedNumericDocValuesRangeQuery
   to speed up computing the number of hits when possible. (Lu Xugang, Luca Cavanna, Adrien Grand)
 
-* LUCENE-10422: Monitor Improvements: `Monitor` can use a custom `Directory`
+* LUCENE-10422: Monitor Improvements: `Monitor` can use a custom `Directory` 
   implementation. `Monitor` can be created with a readonly `QueryIndex` in order to
   have readonly `Monitor` instances. (Niko Usai)
 
@@ -2517,7 +2517,7 @@ Optimizations
   term of each block as a dictionary when compressing suffixes of the other 63
   terms of the block. (Adrien Grand)
 
-* LUCENE-10411: Add nearest neighbors vectors support to ExitableDirectoryReader.
+* LUCENE-10411: Add nearest neighbors vectors support to ExitableDirectoryReader. 
   (Zach Chen, Adrien Grand, Julie Tibshirani, Tomoko Uchida)
 
 * LUCENE-10542: FieldSource exists implementations can avoid value retrieval (Kevin Risden)
@@ -2682,7 +2682,7 @@ New Features
   points are indexed.
   (Quentin Pradet, Adrien Grand)
 
-* LUCENE-10263: Added Weight#count to NormsFieldExistsQuery to speed up the query if all
+* LUCENE-10263: Added Weight#count to NormsFieldExistsQuery to speed up the query if all 
   documents have the field.. (Alan Woodward)
 
 * LUCENE-10248: Add SpanishPluralStemFilter, for precise stemming of Spanish plurals.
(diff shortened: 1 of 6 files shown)


lucene/CHANGES.txt
```diff
@@ -854,7 +862,7 @@ Improvements
 
 * GITHUB#13285: Early terminate graph searches of AbstractVectorSimilarityQuery to follow timeout set from
   IndexSearcher#setTimeout(QueryTimeout). (Kaival Parikh)
-
```
stefanvodita, line 857: Let's leave this file as it was, except for the new entries.

lucene/queryparser/src/test/org/apache/lucene/queryparser/flexible/standard/nodes/TestAbstractRangeQueryNode.java
```diff
@@ -0,0 +1,69 @@
...
+  public void testTermRangeQueryNode() throws Exception {
+    FieldQueryNode lower = new FieldQueryNode("FIELD", "aaa", 0, 0);
+    FieldQueryNode upper = new FieldQueryNode("FIELD", "zzz", 0, 0);
+    TermRangeQueryNode origNode = new TermRangeQueryNode(lower, upper, true, true);
+    CharSequence queryString = origNode.toQueryString(escaper);
+
+    // query string should have expected format and parse into a valid query
+    assertThat(queryString, is("FIELD:[aaa TO zzz]"));
+    Query parsedQuery = parser.parse(queryString.toString(), "");
+    assertThat(parsedQuery.toString(), is(queryString));
```
stefanvodita, line 46: I like that you've added this to what the original patches tested!

lucene/queryparser/src/test/org/apache/lucene/queryparser/flexible/standard/nodes/TestAbstractRangeQueryNode.java
```diff
@@ -0,0 +1,69 @@
...
+      PointRangeQueryNode origNode =
+          new PointRangeQueryNode(
+              lower, upper, true, true, new PointsConfig(format, Integer.class));
+      CharSequence queryString = origNode.toQueryString(escaper);
+
+      // query string should have expected format and parse into a valid query
+      assertThat(queryString, is("FIELD:[1 TO 999]"));
+      Query parsedQuery = parser.parse(queryString.toString(), "");
+      assertThat(parsedQuery.toString(), is(queryString));
+    } catch (QueryNodeException e) {
```
stefanvodita, line 65: Why do we want to have this catch statement?

stefanvodita (changes requested):
Thank you for picking up this long-standing issue @phb-ig. I saw your comment on it and had been looking into it, but it's helpful that you've put out a PR. How did you decide among the three proposed solutions? I'm also curious if you've run into this issue in the wild.

phb-ig:
@stefanvodita I am the one who originally submitted the patches all that time ago because the work I was doing at the time required inspecting the query string and altering, then re-parsing it under certain circumstances. Unfortunately I don't remember any specifics at this point, and I haven't used Lucene day-to-day since then so I really couldn't tell you much more than that 😅 

As to how I chose, I figured I would just pick the most simple solution. If people who know more about what's going on with the code than I do prefer a different way, then I am happy to make those changes 🙂

lucene/CHANGES.txt
phb-ig: Whoops, I have an IDE setting to kill whitespace at the end of lines

lucene/queryparser/src/test/org/apache/lucene/queryparser/flexible/standard/nodes/TestAbstractRangeQueryNode.java
phb-ig: No idea. Perhaps in the time before `LuceneTestCase` existed, there was some sort of complaint about it being uncaught?

lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/core/nodes/ValueQueryNode.java
```diff
@@ -16,10 +16,14 @@
...
+import org.apache.lucene.queryparser.flexible.core.parser.EscapeQuerySyntax;
+
 /** This interface should be implemented by {@link QueryNode} that holds an arbitrary value. */
 public interface ValueQueryNode<T extends Object> extends QueryNode {
 
   public void setValue(T value);
 
   public T getValue();
+
+  public CharSequence getTermEscaped(EscapeQuerySyntax escaper);
```
stefanvodita, line 28: Could we add a comment explaining what this method should do with an example?

lucene/CHANGES.txt
```diff
@@ -123,6 +127,10 @@ Bug Fixes
 * GITHUB#14161: PointInSetQuery's constructor now throws IllegalArgumentException
   instead of UnsupportedOperationException when values are out of order. (Shubham Sharma)
 
+* GITHUB#7865 AbstractRangeQueryNode#toQueryString(EscapeQuerySyntax) now
```
stefanvodita, line 130: I'm conflicted because we're changing behaviour that is public and has been around a long time, but it's also a bug fix. I think I'd be more comfortable if we moved these entries under 11.0.0 though.

stefanvodita (approved):
Oh wow, thank you for coming back to this issue after 10 years @phb-ig, that's serious perseverance! Sorry it's taken so long!

I only had a couple more comments. I'm happy with the solution, but I want to leave the review open a little longer, since I wasn't around 10 years ago to know if there were objections, nor am I the most knowledgeable in this area of the code. If we address the comments and if no one else reviews, I'll merge this in a few days.

stefanvodita (approved):

PR #14698, opened 2025-05-22
An alternate implementation for RerankVectorQuery

vigyasharma:
This PR presents an alternate to #14009 for `RerankVectorQuery`. It takes an input query, a target vector and a field to use for vector values, and rescores the output of wrapped query with full precision vector values from provided field.

**Main differences with #14009 :**
1. While the primary use case would be to rerank results from a quantized `KnnFloatVectorQuery` against full precision vectors, this implementation has no tight coupling. Users could provide any inner query and rerank against vectors stored in any field.
2. This can, in future, be extended to support reranking using Late Interaction models stored in a different field. We just need to override add some checks to ensure multi-vectors in provided `field` are compatible with provided `target`. This idea has been floated [here](https://github.com/apache/lucene/pull/13525#issuecomment-2445295372) earlier
3. Maintains a regular query flow of keeping heavier computation in Weight and Scorer, instead of keeping heavy matching and scoring logic in rewrite.
4. This should implicitly use any slicing or per leaf parallel execution provided at searchers. We don't need to rewrite logic for per leaf threads.

**Pending Tasks:**
I plan to get some benchmark results next. Would also like to clean up the test and add a few more, if we chose to go this route. Sharing an early impl. to get feedback on the approach.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/RerankFloatVectorQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/RerankFloatVectorQuery.java
@@ -36,7 +36,7 @@ import org.apache.lucene.util.VectorUtil;
  */
 public class RerankFloatVectorQuery extends Query {
 
-  private final Query in;
+  private Query in;
   private final String rerankField;
   private final float[] target;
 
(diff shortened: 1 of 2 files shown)


vigyasharma:
Ideas put forth in this PR along with additional improvements have been addressed in #14009 . Closing this PR.

PR #14700, opened 2025-05-22
Return MatchNoDocsQuery when IndexOrDocValuesQuery::rewrite does not match

ChrisHegarty:
In a similar way to how `IndexOrDocValuesQuery` may rewrite to a `MatchAllDocsQuery`, if either of the queries rewrites to match no docs, then return a `MatchNoDocsQuery`.


jpountz (approved):
LGTM

PR #14708, opened 2025-05-24
Add a DoubleValuesSource for scoring full precision vector similarity

vigyasharma:
A `DoubleValuesSource` that scores on full precision vectors can be used on top of quantized knn vector queries to rerank with full precision similarity scores. 

This change adds a 'full precision' similarity mode to existing `VectorSimilarityValuesSource`s. 
Supports #14009

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/ByteVectorSimilarityValuesSource.java
+++ b/lucene/core/src/java/org/apache/lucene/search/ByteVectorSimilarityValuesSource.java
@@ -33,10 +33,13 @@ import org.apache.lucene.index.VectorSimilarityFunction;
  */
 class ByteVectorSimilarityValuesSource extends VectorSimilarityValuesSource {
 
-  /** Creates a {@link ByteVectorSimilarityValuesSource} that scores on full precision vector values */
+  /**
+   * Creates a {@link ByteVectorSimilarityValuesSource} that scores on full precision vector values
+   */
   public static DoubleValues fullPrecisionScores(
       LeafReaderContext ctx, byte[] queryVector, String vectorField) throws IOException {
-    return new ByteVectorSimilarityValuesSource(queryVector, vectorField, true).getValues(ctx, null);
+    return new ByteVectorSimilarityValuesSource(queryVector, vectorField, true)
+        .getValues(ctx, null);
   }
 
   private final byte[] queryVector;
@@ -62,7 +65,8 @@ class ByteVectorSimilarityValuesSource extends VectorSimilarityValuesSource {
    * @param useFullPrecision uses full precision raw vectors for similarity computation if true,
    *     otherwise the configured vectors reader is used, which may be quantized or full precision.
    */
-  public ByteVectorSimilarityValuesSource(byte[] vector, String fieldName, boolean useFullPrecision) {
+  public ByteVectorSimilarityValuesSource(
+      byte[] vector, String fieldName, boolean useFullPrecision) {
     super(fieldName);
     this.queryVector = vector;
     this.useFullPrecision = useFullPrecision;
--- a/lucene/core/src/java/org/apache/lucene/search/ByteVectorSimilarityValuesSource.java
+++ b/lucene/core/src/java/org/apache/lucene/search/ByteVectorSimilarityValuesSource.java
@@ -24,7 +24,6 @@ import org.apache.lucene.index.ByteVectorValues;
 import org.apache.lucene.index.FieldInfo;
 import org.apache.lucene.index.KnnVectorValues;
 import org.apache.lucene.index.LeafReaderContext;
-import org.apache.lucene.index.VectorEncoding;
 import org.apache.lucene.index.VectorSimilarityFunction;
 
 /**
--- a/lucene/core/src/java/org/apache/lucene/search/ByteVectorSimilarityValuesSource.java
+++ b/lucene/core/src/java/org/apache/lucene/search/ByteVectorSimilarityValuesSource.java
@@ -21,18 +21,51 @@ import java.io.IOException;
 import java.util.Arrays;
 import java.util.Objects;
 import org.apache.lucene.index.ByteVectorValues;
+import org.apache.lucene.index.FieldInfo;
+import org.apache.lucene.index.KnnVectorValues;
 import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.index.VectorEncoding;
+import org.apache.lucene.index.VectorSimilarityFunction;
 
 /**
  * A {@link DoubleValuesSource} which computes the vector similarity scores between the query vector
  * and the {@link org.apache.lucene.document.KnnByteVectorField} for documents.
  */
 class ByteVectorSimilarityValuesSource extends VectorSimilarityValuesSource {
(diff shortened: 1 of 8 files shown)


vigyasharma:
Thanks for the review folks! I like the idea of a separate class and a custom vector comparator, will make these changes.

vigyasharma:
I'm not sure about the byte vector case myself. Do we see a viable need for it in FullPrecisionVectorSimilaritySource ?

lucene/core/src/java/org/apache/lucene/search/ByteVectorSimilarityValuesSource.java
```diff
@@ -42,7 +78,35 @@ public VectorScorer getScorer(LeafReaderContext ctx) throws IOException {
...
+    }
+
+    final VectorSimilarityFunction vectorSimilarityFunction = fi.getVectorSimilarityFunction();
+    return new VectorScorer() {
+      final KnnVectorValues.DocIndexIterator iterator = vectorValues.iterator();
+
+      @Override
+      public float score() throws IOException {
+        return vectorSimilarityFunction.compare(
+            queryVector, vectorValues.vectorValue(iterator.index()));
```
dungba88, line 102: One of the possible concern for the reranking is that calling this will page in the full vector data (or what is being used in the iterator), and can compete the RAM with normal vectors used in HNSW graph search. As HNSW will suffer the performance if the vectors are not in RAM, I'm wondering if we can restrict the memory used by the re-ranking phase. Or alternatively, using `mlock` (could be tricky in Java) to lock in the normal vectors page.

dungba88: This can be done separately too. I'm just curious what you think about this.

vigyasharma: It's a valid concern for setups with limited memory.

> As HNSW will suffer the performance if the vectors are not in RAM, I'm wondering if we can restrict the memory used by the re-ranking phase.

Maybe.. I wonder how we decide that the pages used for HNSW search are more important than pages used for FP reranking. For an application which does KNN search and reranks via full precision vectors, a query doesn't really complete until both phases are done. Wouldn't thrashing queries during reranking add to overall query latency. Maybe this is okay if you were reranking for only a subset of queries, and the vast majority is still only HNSW search, but that seems very use-case specific. Might be best to let OS page cache handle this?

Anyway, I think this deserves it's own discussion, perhaps in a separate issue? And as you and others have already mentioned, it can be handled independent of this PR.

vigyasharma: I do remember reading some results on the DiskANN issue where benchmarks indicated that having the vectors needed for ANN graph search in memory (the quantized vectors in this case), does lead to better performance. So maybe, an option to use only DIRECT_IO for this makes sense.

vigyasharma:
Moved full precision scores logic to a separate `FullPrecisionFloatVectorSimilarityValuesSource` that can take a custom vector similarity function.

lucene/core/src/java/org/apache/lucene/search/ByteVectorSimilarityValuesSource.java
dungba88: I'm putting https://github.com/apache/lucene/issues/14746 for further discussion on this topic.

dungba88 (approved):
LGTM, thank you

lucene/core/src/test/org/apache/lucene/search/TestQuantizedVectorSimilarityValueSource.java
```diff
@@ -0,0 +1,217 @@
...
+import org.apache.lucene.index.StoredFields;
+import org.apache.lucene.index.VectorSimilarityFunction;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.tests.util.LuceneTestCase;
+import org.apache.lucene.tests.util.TestUtil;
+import org.apache.lucene.util.TestVectorUtil;
+import org.junit.Before;
+import org.junit.Test;
+
+public class TestQuantizedVectorSimilarityValueSource extends LuceneTestCase {
```
dungba88: nit: should it be `TestFullPrecisionFloatVectorSimilarityValueSource`?

vigyasharma: right, I forgot to rename the test class. fixed it now.

vigyasharma:
Thanks all for reviewing this PR. I've addressed the comments above. Will merge this early next week, unless there is any new feedback.

vigyasharma:
I added some benchmarks to measure NDCG for knn search with 4 bit quantized vectors, reranked with full precision vector similarity scores here - https://github.com/mikemccand/luceneutil/pull/435

Pasting results below as well.

+++

#### Benchmark Results
I see an `8% - 14%` improvement in NDCG@10 and a `6% - 8%` improvement in NDCG@K for 4 bit quantized knn search with full precision reranking. Improvement increases with index size. Latency impact doesn't seem significant?

```ruby
recall  ndcg@10  ndcg@K  rerank  latency(ms)  netCPU  avgCpuCount      nDoc  topK  fanout  maxConn  beamWidth  quantized  index(s)  index_docs/s  num_segments  index_size(MB)  vec_disk(MB)  vec_RAM(MB)  indexType
 0.524    0.920   0.600   false        2.235   2.216        0.991    100000   100      20       32         50     4 bits     16.96       5896.23             4          333.95       329.971       37.003       HNSW
 0.524    0.999   0.637    true        2.279   2.261        0.992    100000   100      20       32         50     4 bits      0.00      Infinity             4          333.95       329.971       37.003       HNSW
 0.490    0.901   0.568   false        3.199   3.183        0.995    500000   100      20       32         50     4 bits     70.18       7124.23             3         1670.44      1649.857      185.013       HNSW
 0.490    0.999   0.607    true        3.201   3.181        0.994    500000   100      20       32         50     4 bits      0.00      Infinity             3         1670.44      1649.857      185.013       HNSW
 0.480    0.891   0.557   false        4.774   4.756        0.996   1000000   100      20       32         50     4 bits    134.12       7456.07             6         3341.86      3299.713      370.026       HNSW
 0.480    0.999   0.598    true        4.697   4.675        0.995   1000000   100      20       32         50     4 bits      0.00      Infinity             6         3341.86      3299.713      370.026       HNSW
 0.462    0.883   0.541   false        5.081   5.035        0.991   2000000   100      20       32         50     4 bits    465.46       4296.82             7         6688.32      6599.426      740.051       HNSW
 0.462    0.998   0.583    true        4.885   4.863        0.995   2000000   100      20       32         50     4 bits      0.00      Infinity             7         6688.32      6599.426      740.051       HNSW
 0.447    0.871   0.526   false       11.633  11.495        0.988  10000000   100      20       32         50     4 bits   2110.54       4738.12            14        33489.11     32997.131     3700.256       HNSW
 0.447    0.998   0.569    true       11.037  10.984        0.995  10000000   100      20       32         50     4 bits      0.00      Infinity            14        33489.11     32997.131     3700.256       HNSW

```

PR #14716, opened 2025-05-26
Merge PostingsEnum and ImpactsEnum.

jpountz:
When impacts were introduced, I tried to avoid touching `PostingsEnum`, which had a relatively stable API while impacts were rather experimental. Impacts have been quite successful, so I think it's now time to make things simpler by adding support to impacts directly into `PostingsEnum`?

More specifically:
 - `PostingsEnum` now has `advanceShallow` and `getImpacts` methods.
 - You can request impacts on a `PostingsEnum` by using the `IMPACTS` flag.
 - When the `IMPACTS` flag is not set, `PostingsEnum` may return dummy impacts (as for term frequencies when not requested).


lucene/core/src/java/org/apache/lucene/index/TermsEnum.java
```diff
@@ -170,11 +170,15 @@ public final PostingsEnum postings(PostingsEnum reuse) throws IOException {
...
 
   /**
-   * Return a {@link ImpactsEnum}.
+   * Return a {@link PostingsEnum} that exposes impacts.
    *
    * @see #postings(PostingsEnum, int)
+   * @deprecated Use {@link #postings(PostingsEnum)} with {@link PostingsEnum#IMPACTS}
    */
-  public abstract ImpactsEnum impacts(int flags) throws IOException;
+  @Deprecated
```
gf2121, line 178: Should we remove this and deprecate in 10x?

lucene/core/src/java/org/apache/lucene/index/PostingsEnum.java
```diff
@@ -63,6 +87,12 @@ public abstract class PostingsEnum extends DocIdSetIterator {
    */
   public static final short ALL = OFFSETS | PAYLOADS;
 
+  /**
+   * Flag to pass to {@link TermsEnum#postings(PostingsEnum, int)} to get impacts in the returned
+   * enum.
+   */
+  public static final short IMPACTS = FREQS | 1 << 7;
```
gf2121, line 94: Just wild thought: can we make impacts work on `PostingEnum.FREQ` instead introducing this flag?  We can delay the reading of impact bytes (where is `needsImpacts` mostly used today)  to `getImpacts`, so that it won't cause any noticeable overhead.

gf2121 (approved):
Nice simplification!

lucene/core/src/java/org/apache/lucene/index/TermsEnum.java
jpountz: Yes! I was hoping to remove in a subsequent PR.

lucene/core/src/java/org/apache/lucene/index/PostingsEnum.java
jpountz: Interesting idea. Let me check the performance impact.

jpountz: After a few iterations on wikibigall, it seems to work fine with the current codec. However it looks challenging to apply correctly to old codecs, which use a different PostingsEnum impl when impacts are requested, and that I'd like to avoid touching too much.

jpountz: I imagine that having a flag also has some benefits, e.g. some impls could decide to store impacts in a separate file in the future and only open this file if impacts are requested.

gf2121: Thanks for trying!

gf2121 (approved):

PR #14720, opened 2025-05-27
Cache high-order bits of hashcode to speed up BytesRefHash

bugmakerrrrrr:
### Description

This PR tries to utilize the unused part of the id to cache the high-order bits of the hashcode to speed up `BytesRefHash`. I used 1 million 16-byte UUIDs to [benchmark this change](https://github.com/bugmakerrrrrr/lucene/commit/43d2945be75acb2464c36ca1eac6067445687fe2), and the results are as follows.
image

The `baselineXXX` version is the current implementation, the `cachedXXX` version uses a separate array of ints to cache hash codes, and the candidate version is the implementation of this PR.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -128,6 +128,8 @@ Optimizations
 
 * GITHUB#14674: Optimize AbstractKnnVectorQuery#createBitSet with intoBitset. (Guo Feng)
 
+* GITHUB#14720: Cache high-order bits of hashcode to speed up BytesRefHash. (Pan Guixin)
+
 Bug Fixes
 ---------------------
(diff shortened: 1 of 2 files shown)


jpountz (commented):
Interesting trick, it makes sense to me that this helps. Can you add some code comments and/or javadocs to explain this trick and what the `ids` array stores?

jpountz (approved):
I got a bit confused at first due to some bits of the change being related to `addByPoolOffset`, which is a bit special, but the change looks good.

jpountz:
Oh, and great javadocs!

dweiss:
> Oh, and great javadocs!

Yes, agree 100% - thank you for adding a human-palatable explanation of what's going on there.

PR #14725, opened 2025-05-28
No ruff violation

ghostbranch-33:
### Description

Didn't find these ruff rules in the code


dweiss (approved):
Seems ok to me, although I'm not a python expert.

PR #14729, opened 2025-05-29
Support for Re-Ranking Queries using Late Interaction Model Multi-Vectors.

vigyasharma:
Late Interaction models, like [ColBERT](https://arxiv.org/abs/2004.12832) and [ColPali](https://arxiv.org/html/2407.01449v2), capture rich semantic interaction between documents and queries, and have been shown to outperform single-vector (no-interaction) models on search relevance. These models operate by using multi-vector representations for query (and document) embeddings. 

One challenge with including late interaction models in search, has been working with multi-vectors at scale. This change provides an efficient workaround, by adding support to rerank results of a query using late interaction multi-vectors.

Typical envisioned use-case is to do the full corpus search using ANN search on single-valued vectors, followed by a second pass that reranks results using late-interaction multi-vector scores. This PR creates:
1. A LateInteractionField that stores multi-vectors in BinaryDocValues
2. A DoubleValuesSource to scores query and document multi-vectors.
3. A FunctionScore query that wraps a provided query and reranks its result with late-interaction model scores.

Note: This first approach does not add additional metadata to `FieldInfo`. As a result, we are unable to ensure consistency in shape for multi-vector indexed in the same field across documents.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/document/LateInteractionField.java
+++ b/lucene/core/src/java/org/apache/lucene/document/LateInteractionField.java
@@ -20,7 +20,6 @@ package org.apache.lucene.document;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.nio.FloatBuffer;
-
 import org.apache.lucene.util.BytesRef;
 
 /**
@@ -76,7 +75,7 @@ public class LateInteractionField extends BinaryDocValuesField {
     if (value == null || value.length == 0) {
       throw new IllegalArgumentException("Value should not be null or empty");
     }
-    if (value[0] == null || value[0].length== 0) {
+    if (value[0] == null || value[0].length == 0) {
       throw new IllegalArgumentException("Composing token vectors should not be null or empty");
     }
     final int tokenVectorDimension = value[0].length;
--- a/lucene/core/src/java/org/apache/lucene/document/LateInteractionField.java
+++ b/lucene/core/src/java/org/apache/lucene/document/LateInteractionField.java
@@ -109,7 +109,7 @@ public class LateInteractionField extends BinaryDocValuesField {
    * LateInteractionField#encode(float[][])}
    *
    * @param payload to decode into multi-vector value
-   * @return
+   * @return decoded multi-vector value
    */
   public static float[][] decode(BytesRef payload) {
     final ByteBuffer buffer = ByteBuffer.wrap(payload.bytes, payload.offset, payload.length);
--- a/lucene/core/src/java/org/apache/lucene/document/LateInteractionField.java
+++ b/lucene/core/src/java/org/apache/lucene/document/LateInteractionField.java
@@ -19,6 +19,8 @@ package org.apache.lucene.document;
 
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
+import java.nio.FloatBuffer;
+
 import org.apache.lucene.util.BytesRef;
 
 /**
@@ -74,12 +76,16 @@ public class LateInteractionField extends BinaryDocValuesField {
     if (value == null || value.length == 0) {
       throw new IllegalArgumentException("Value should not be null or empty");
     }
+    if (value[0] == null || value[0].length== 0) {
+      throw new IllegalArgumentException("Composing token vectors should not be null or empty");
+    }
     final int tokenVectorDimension = value[0].length;
     final ByteBuffer buffer =
         ByteBuffer.allocate(Integer.BYTES + value.length * tokenVectorDimension * Float.BYTES)
             .order(ByteOrder.LITTLE_ENDIAN);
     // TODO: Should we store dimension in FieldType to ensure consistency across all documents?
     buffer.putInt(tokenVectorDimension);
+    FloatBuffer floatBuffer = buffer.asFloatBuffer();
     for (int i = 0; i < value.length; i++) {
       if (value[i].length != tokenVectorDimension) {
         throw new IllegalArgumentException(
@@ -91,7 +97,7 @@ public class LateInteractionField extends BinaryDocValuesField {
                 + " != "
                 + value[i].length);
       }
-      buffer.asFloatBuffer().put(value[i]);
+      floatBuffer.put(value[i]);
     }
     return new BytesRef(buffer.array());
   }
@@ -110,8 +116,8 @@ public class LateInteractionField extends BinaryDocValuesField {
     final ByteBuffer buffer = ByteBuffer.wrap(payload.bytes, payload.offset, payload.length);
     buffer.order(ByteOrder.LITTLE_ENDIAN);
     final int tokenVectorDimension = buffer.getInt();
-    int numVectors = (payload.length - 4) / tokenVectorDimension;
-    if (numVectors * tokenVectorDimension + 4 != payload.length) {
+    int numVectors = (payload.length - Integer.BYTES) / (tokenVectorDimension * Float.BYTES);
+    if (numVectors * tokenVectorDimension * Float.BYTES + Integer.BYTES != payload.length) {
       throw new IllegalArgumentException(
           "Provided payload does not appear to have been encoded via LateInteractionField.encode. "
               + "Payload length should be equal to 4 + numVectors * tokenVectorDimension, "
--- a/lucene/core/src/java/org/apache/lucene/document/LateInteractionField.java
+++ b/lucene/core/src/java/org/apache/lucene/document/LateInteractionField.java
@@ -45,6 +45,21 @@ public class LateInteractionField extends BinaryDocValuesField {
   }
 
   /**
+   * Set multi-vector value for the field.
+   *
+   * <p>Value should not be null or empty. Composing token vectors for provided multi-vector value
+   * should have the same dimension.
+   */
+  public void setValue(float[][] value) {
+    this.fieldsData = encode(value);
+  }
+
+  /** Returns the multi-vector value stored in this field */
+  public float[][] getValue() {
+    return decode((BytesRef) fieldsData);
+  }
+
+  /**
    * Encodes provided multi-vector matrix into a {@link BytesRef} that can be stored in the {@link
    * LateInteractionField}.
    *
@@ -56,8 +71,8 @@ public class LateInteractionField extends BinaryDocValuesField {
    * @return BytesRef representation for provided multi-vector
    */
   public static BytesRef encode(float[][] value) {
-    if (value.length == 0) {
-      throw new IllegalArgumentException("Provided value is empty");
+    if (value == null || value.length == 0) {
+      throw new IllegalArgumentException("Value should not be null or empty");
     }
     final int tokenVectorDimension = value[0].length;
     final ByteBuffer buffer =
--- a/lucene/core/src/java/org/apache/lucene/document/LateInteractionField.java
+++ b/lucene/core/src/java/org/apache/lucene/document/LateInteractionField.java
@@ -1,0 +1,118 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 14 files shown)


vigyasharma:
This change builds on the work shared [here](https://github.com/apache/lucene/pull/13525#issuecomment-2445295372) by @jimczi, thanks Jim!

benwtrent:
I like this idea and its a good starting point for allowing late-interaction brute-force ranking (e.g reranking).

romseygeek:
> Typical envisioned use-case is to do the full corpus search using ANN search on single-valued vectors, followed by a second pass that reranks results using late-interaction multi-vector scores.

This sounds like it would fit nicely with the `Rescorer` infrastructure?  Looking more closely, it seems that Rescorers don't currently support adjusting scores via a DoubleValuesSource, but this shouldn't be too tricky to add (and indeed, the existing Rescorer implementations could quite easily be reworked to use DoubleValuesSource everywhere which would cut down on a lot of code duplication).

vigyasharma:
> This sounds like it would fit nicely with the `Rescorer` infrastructure?

I wasn't aware of Query Rescorers, thanks @romseygeek for pointing me in that direction. At the outset, it seems like we could create a `DoubleValuesSourceRescorer extends Rescorer`  – that rescores based on the `DoubleValuesSource` instead of a second query.

However, this is exactly what the `FunctionScoreQuery` does, and the benefits of one over the other are not immediately obvious to me. I see`FunctionScoreQuery` being a `Query`, exposes a `SegmentCacheable` weight, which could be useful? Though I believe we only cache hits (disi) and not scores, so maybe it doesn't matter for this use case?

Do you see any benefits to using `Rescorer` over `FunctionScoreQuery` ?

romseygeek:
The advantage of a `Rescorer` is that is is explicitly only run over the hits in a `TopDocs` instance, whereas `FunctionScoreQuery` will run over the entire docid space if you let it.  So it's a natural fit for a late-interaction search process - run your first query over the whole document set to get a preliminary top-k, and then pass the resulting `TopDocs` to your rescorer.

vigyasharma:
> The advantage of a `Rescorer` is that is is explicitly only run over the hits in a `TopDocs` instance, whereas `FunctionScoreQuery` will run over the entire docid space if you let it.

Makes sense. I was thinking of only the knn vector queries that return top-N matches only, but reranking can apply more generally to any query. 

I've raised #14776 to add support for `DoubleValuesSource` based rescorers. Once that is merged, I'll modify this PR to use `DoubleValuesSourceRescorer` instead of a `FunctionScoreQuery`.

benwtrent:
@vigyasharma I really like how this is evolving and how we are unifying these vector scoring providers. Great stuff!

lucene/core/src/java/org/apache/lucene/search/LateInteractionFloatValuesSource.java
```diff
@@ -0,0 +1,194 @@
...
+        + ")";
+  }
+
+  @Override
+  public boolean isCacheable(LeafReaderContext ctx) {
+    return true;
+  }
+
+  /** Defines the function to compute similarity score between query and document multi-vectors */
+  public enum ScoreFunction {
```
dungba88: I think this follows the VectorSimilarityFunction convention to define function as enum. But have we thought about allowing a `@FunctionalInternal` as parameter to allow custom scoring without having to modify Lucene core? It's not a strong opinion, as I guess people can still define their own `DoubleValueSource` (but maybe will have to duplicate this code).

mingshl:
This is an exciting feature!! I think it's great idea to create a `LateInteractionField` that can store multi-vector values in the documents, but I have a question in search, when I try to look at the Function score query, it seems that for search request, it first runs a knn query, then we fetch the top N documents for maxSim reranker. 

In this case, we need to use a single vector for knn query, and then use multi-vectors for reranking. 

If we want to use late interaction model for search, after we get the multi-vectors from the model and try to construct the function score query, 

first, we need a way to pool the multi-vector into single vector, and put into knn query
second, we would put the multi-vector into the lateInteractionFloatRerankQuery 

I can see this way, we can save a lot of computing for too many MaxSim calculation, when the array of vector is big. 

I am thinking if that worths a new query type that can handle both pooling and rerank, basically the above two steps together. 

https://github.com/apache/lucene/blob/8dc05f47e0c0bc1c66d530fa6525b96e536a39d0/lucene/queries/src/test/org/apache/lucene/queries/function/TestFunctionScoreQuery.java#L447-L452

lucene/core/src/java/org/apache/lucene/search/LateInteractionFloatValuesSource.java
```diff
@@ -0,0 +1,194 @@
...
+          return Float.MIN_VALUE;
+        }
+        float result = 0f;
+        for (float[] q : queryVector) {
+          float maxSim = Float.MIN_VALUE;
+          for (float[] d : docVector) {
+            if (q.length != d.length) {
+              throw new IllegalArgumentException(
+                  "Provided multi-vectors are incompatible. "
+                      + "Their composing token vectors should have the same dimension, got "
```
dungba88, line 170: minor: maybe we can say got query dimension = ..., document dimension = ...

lucene/core/src/java/org/apache/lucene/search/LateInteractionFloatValuesSource.java
```diff
@@ -0,0 +1,194 @@
...
+      throw new IllegalArgumentException(
+          "composing token vectors in provided query vector should not be null or empty");
+    }
+    for (int i = 1; i < queryVector.length; i++) {
+      if (queryVector[i] == null || queryVector[i].length != queryVector[0].length) {
+        throw new IllegalArgumentException(
+            "all composing token vectors in provided query vector should have the same length");
+      }
+    }
+    return queryVector;
```
dungba88, line 79: We do not change this float vector. I suppose this is just a sugar syntactic to keep L58 to be a single line?

dungba88 (commented):
I like this implementation! It's much simpler and decomposable than https://github.com/apache/lucene/pull/13525.

vigyasharma:
@mingshl you're right, this only adds support to _rerank_ the results of an initial query using late interaction multi-vectors. We would typically need to first run a knn vector query on some indexed vector field, then rerank results using the late interaction field.

...

> I am thinking if that worths a new query type that can handle both pooling and rerank, basically the above two steps together.

This is an interesting idea. We would also have to index the pooled vector values. So essentially have two fields in Lucene, a `KnnFloatVectorField` that indexes the pooled single-vector value into an hnsw graph for searching, and another `LateInteractionField` that stores the un-pooled multi-vectors for reranking.

It seems useful to provide convenience wrappers such that users only provide the late-interaction multi-vectors, and they are internally pooled, indexed, stored and queried through the two fields. I'm not sure if this should be in Lucene, or in higher level layers like OpenSearch/Elasticsearch/Solr. We should pick this up in a spin-off issue.

lucene/core/src/java/org/apache/lucene/search/LateInteractionFloatValuesSource.java
vigyasharma: Yes, that is correct.

mingshl:
>@mingshl you're right, this only adds support to rerank the results of an initial query using late interaction multi-vectors. We would typically need to first run a knn vector query on some indexed vector field, then rerank results using the late interaction field.



>> I am thinking if that worths a new query type that can handle both pooling and rerank, basically the above two steps together.


> This is an interesting idea. We would also have to index the pooled vector values. So essentially have two fields in Lucene, a KnnFloatVectorField that indexes the pooled single-vector value into an hnsw graph for searching, and another LateInteractionField that stores the un-pooled multi-vectors for reranking.

> It seems useful to provide convenience wrappers such that users only provide the late-interaction multi-vectors, and they are internally pooled, indexed, stored and queried through the two fields. I'm not sure if this should be in Lucene, or in higher level layers like OpenSearch/Elasticsearch/Solr. We should pick this up in a spin-off issue.

Agree that the process of the pooled, indexed, stored and queried job should be the work of higher level layer. A new query type and an ingest processor would help this case.  Let's move the discussion to separate RFC https://github.com/opensearch-project/OpenSearch/issues/18091

lucene/core/src/java/org/apache/lucene/search/LateInteractionFloatValuesSource.java
vigyasharma: This may get reused during graph building for multi-vector ANN, so I'm keeping it as a generic comparison between two multi-vectors.

lucene/core/src/java/org/apache/lucene/search/LateInteractionFloatValuesSource.java
vigyasharma: I added a `MultiVectorSimilarity` interface to allow users to provide their custom comparison functions.

vigyasharma:
Thanks for the reviews and suggestions, everyone. The latest revision incorporates all feedback so far. 

It includes a `LateInteractionRescorer` for reranking top-N hits from provided TopDocs, as well as a late interaction `RescoreTopNQuery`, rewrites the provided query to keep only top N hits based on late interaction scores.

vigyasharma:
Planning to merge this PR by tomorrow, if there is no new feedback. Again, thanks everyone in helping shape this change with your inputs!

lucene/core/src/java/org/apache/lucene/document/LateInteractionField.java
```diff
@@ -0,0 +1,138 @@
...
+ */
+public class LateInteractionField extends BinaryDocValuesField {
+
+  /**
+   * Creates a new {@link LateInteractionField} from the provided multi-vector matrix.
+   *
+   * @param name field name
+   * @param value multi-vector value
+   */
+  public LateInteractionField(String name, float[][] value) {
```
mingshl, line 44: I have a question about this `LateInteractionField` which stores the value with an nested array of float. When it stores, will it be the same as ?

```
PUT test_float_array/_doc/1
{
  "review_colbert_vectors": [
    [
      0.123,
      0.234,
      0.345,
      0.456,
      0.567,
      0.678,
      0.789,
      0.123
    ],
    [
      0.123,
      0.234,
      0.345,
      0.456,
      0.567,
      0.678,
      0.789,
      0.123
    ],
    [
      0.234,
      0.234,
      0.345,
      0.456,
      0.567,
      0.678,
      0.789,
      0.123
    ],
    [
      0.345,
      0.345,
      0.345,
      0.456,
      0.567,
      0.678,
      0.789,
      0.123
    ]
  ]
}

GET test_float_array/_doc/1
{
  "_index": "test_float_array",
  "_id": "1",
  "_version": 1,
  "_seq_no": 0,
  "_primary_term": 1,
  "found": true,
  "_source": {
    "review_colbert_vectors": [
      [
        0.123,
        0.234,
        0.345,
        0.456,
        0.567,
        0.678,
        0.789,
        0.123
      ],
      [
        0.123,
        0.234,
        0.345,
        0.456,
        0.567,
        0.678,
        0.789,
        0.123
      ],
      [
        0.234,
        0.234,
        0.345,
        0.456,
        0.567,
        0.678,
        0.789,
        0.123
      ],
      [
        0.345,
        0.345,
        0.345,
        0.456,
        0.567,
        0.678,
        0.789,
        0.123
      ]
    ]
  }
}
```

vigyasharma: Yes, it expects the value to be a 2D float array. This [lucene test](https://github.com/apache/lucene/pull/14729/files) has a java example for the same. 

Is the http request above from OpenSearch? I suppose you need to add a corresponding field there first, that maps to the `LateInteractionField`?

mingshl: yes, if using OpenSearch, we don't need to add a new field though, it seems it's currently supported as float type

PR #14736, opened 2025-05-30
Fix IndexSortSortedNumericDocValuesRangeQuery for int sort (#14732)

mayya-sharipova:
This query assumed that missing value is always of type long. 
This modifies it to allow type int as well.
The test is added that fails without this change.

Backport for #14732


dsmiley:
The CHANGES.txt addition should go under 9.12.2, not 9.12.0

PR #14739, opened 2025-05-31
Keep evaluating conjunction one doc-at-a-time until dynamic pruning kicks in.

jpountz:
This essentially reverts the change from #14701 for conjunctive queries that have not reached their `totalHitsThreshold` yet. This should speed up queries whose total number of matches is in the order of `totalHitsThreshold` or less, such as filtered conjunctions on nightly benchmarks.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/BlockMaxConjunctionBulkScorer.java
+++ b/lucene/core/src/java/org/apache/lucene/search/BlockMaxConjunctionBulkScorer.java
@@ -87,18 +87,64 @@ final class BlockMaxConjunctionBulkScorer extends BulkScorer {
       // NOTE: windowMax is inclusive
       int windowMax = Math.min(scorers[0].advanceShallow(windowMin), max - 1);
 
-      float maxWindowScore = Float.POSITIVE_INFINITY;
       if (0 < scorable.minCompetitiveScore) {
-        maxWindowScore = computeMaxScore(windowMin, windowMax);
+        float maxWindowScore = computeMaxScore(windowMin, windowMax);
+        scoreWindowScoreFirst(collector, acceptDocs, windowMin, windowMax + 1, maxWindowScore);
+      } else {
+        scoreWindowDocFirst(collector, acceptDocs, windowMin, windowMax + 1);
       }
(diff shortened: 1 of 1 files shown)


lucene/core/src/java/org/apache/lucene/search/BlockMaxConjunctionBulkScorer.java
```diff
@@ -87,18 +87,64 @@ public int score(LeafCollector collector, Bits acceptDocs, int min, int max) thr
       // NOTE: windowMax is inclusive
       int windowMax = Math.min(scorers[0].advanceShallow(windowMin), max - 1);
 
-      float maxWindowScore = Float.POSITIVE_INFINITY;
       if (0 < scorable.minCompetitiveScore) {
-        maxWindowScore = computeMaxScore(windowMin, windowMax);
+        float maxWindowScore = computeMaxScore(windowMin, windowMax);
+        scoreWindowScoreFirst(collector, acceptDocs, windowMin, windowMax + 1, maxWindowScore);
+      } else {
+        scoreWindowDocFirst(collector, acceptDocs, windowMin, windowMax + 1);
```
gf2121, line 94: So `minCompetitiveScore` won't get a chance to be respected when filter clause leads the query because `windowMax` is `DocIdSetIterator#NO_MORE_DOCS`, could this cause regression?

jpountz: I believe we've always had this problem? I remember trying to make things better but it didn't look great or caused performance regressions with term queries, the case I care about the most.

gf2121: > I believe we've always had this problem?

I agree that the previous version could not skip windows, but within window, it only needs to do conjunction with the competitive docs, while this PR could evaluate more.

I'm not sure how much this will affect though. `FilteredAndHighHigh` tasks should provide similar case and numbers not look bad. Let's move on.

jpountz:
This restored performance of filtered queries on nightly benchmarks: https://benchmarks.mikemccandless.com/FilteredAnd3Terms.html. And also improved combined conjunctions: https://benchmarks.mikemccandless.com/CombinedAndHighHigh.html. I'll push an annotation.

PR #14740, opened 2025-05-31
.editorconfig

dsmiley:
Complementary to Spotless/tidy

### Description

A while back, I imported a "Google Java Style" into my IntelliJ from some internet source and have been using it in my IDE.  I exported that to editorconfig format.  Then I made some tweaks:
* added `root = true`
* removed entire file format sections that our repo doesn't even have.  IntelliJ was able to point this out easily.
* \*.properties: kept blank lines
* \*.gradle: added as a groovy source file pattern
* `tab_width` removed as generally redundant
* `insert_final_newline = true` because spotless.gradle configures this.

QA: I told IntelliJ to reformat and optimize imports across all of lucene-core.  I saw no changes to java sources, which was correct except for a generated source file.


.editorconfig
```diff
@@ -0,0 +1,939 @@
...
+# for consistent code style configuration across editors/IDEs
+
+# top-most EditorConfig file
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf # matches spotless.gradle insistence on this
+indent_size = 2
+indent_style = space
```
rmuir, line 11: Setting this within `[*]` will mess up some editors if they try to touch a Makefile, we need something like this to prevent editors from doing the wrong thing:
```
[Makefile]
indent_style = tab
indent_size 4
```

.editorconfig
```diff
@@ -0,0 +1,939 @@
...
+ij_groovy_use_fq_class_names_in_javadoc = true
+ij_groovy_use_relative_indents = false
+ij_groovy_use_single_class_imports = true
+ij_groovy_variable_annotation_wrap = off
+ij_groovy_while_brace_force = never
+ij_groovy_while_on_new_line = false
+ij_groovy_wrap_chain_calls_after_dot = false
+ij_groovy_wrap_long_lines = false
+
+[{*.har,*.jsb2,*.jsb3,*.json,*.jsonc,*.postman_collection,*.postman_collection.json,*.postman_environment,*.postman_environment.json,.babelrc,.eslintrc,.prettierrc,.stylelintrc,.ws-context,jest.config}]
```
rmuir, line 764: can be reduced to just .json as the other filetypes don't exist. same goes for the other blocks, it would be good to minimize to prevent confusion

vigyasharma:
Nice, this would help avoid a bunch of "tidy" commits in my git history!

I tried pulling the `.editorconfig` locally and reformatted a couple of existing files via `Cmd + Alt + L` on IntelliJ. It did make a bunch of changes, mostly on docStrings. Attaching the git diff here, in case this is reconcilable with some simple editorconfig commands.  – [(diff_with_editorconfig_on_main.txt)](https://github.com/user-attachments/files/20538133/diff_with_editorconfig_on_main.txt)

An alternate would be to change reformatting scope to only uncommitted changes. This can be done via a setting in Intellij but might be easy to miss?

dweiss (commented):
I'm fine with adding these hints, although I don't use this convention myself so Robert's comments should probably be reviewed and applied first - especially if adding this file may result in corruping certain file types (Makefile).

vigyasharma:
> I'm fine with adding these hints, although I don't use this convention myself

Oh wait, I thought these changes were coming from the `editorconfig` in this PR. But looks like it's adding some stuff from my local IDE setup, duh! Surprising because I don't use this convention either!



> Robert's comments should probably be reviewed and applied first - especially if adding this file may result in corruping certain file types (Makefile).

100% agree. We also don't need to add any extra hints, as long as we have parity with the tidy bot, and reformatting with this config doesn't change existing committed code.

dsmiley:
I simplified the glob patterns.

I realized my QA was quite flawed because I had the "Google Java Format" IntelliJ plugin installed, which overrides the style settings.  I disabled it and spent hours wrestling with IntelliJ's settings to try to get it to match as close as possible.  Alas... there remains irreconcilable differences, especially in line wrapping of Javadoc but also in line wrapping of long assignments.  This is the best I could do.  I updated the Groovy settings to match because of it's syntactical closeness to Java.

A reminder:  these settings are applied to certain lines of code or files when the user takes certain actions.  Yes it can be done automatically on commit but that's opt-in and I'm not sure I'd condone that in a project that uses a non-IDE solution as we do.

.editorconfig
```diff
@@ -0,0 +1,918 @@
+# EditorConfig: https://editorconfig.org
+# for consistent code style configuration across editors/IDEs
+
+# top-most EditorConfig file
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf # matches spotless.gradle insistence on this
```
dweiss, line 9: Will editing of gradlew.bat be affected by this? I feel tempted to only list the files we explicitly know the formatting of - java, txt, md, etc... I don't know how this works with all sorts of text editors but I'm afraid of screwing up something by just opening a file with these settings defined.

.editorconfig
```diff
@@ -0,0 +1,918 @@
...
+ij_markdown_wrap_text_if_long = true
+ij_markdown_wrap_text_inside_blockquotes = true
+
+[*.properties]
+ij_properties_align_group_field_declarations = false
+ij_properties_keep_blank_lines = true
+ij_properties_key_value_delimiter = equals
+ij_properties_spaces_around_key_value_delimiter = false
+
+[*.py]
```
rmuir: remove this as it disagrees with the ruff configuration (which autoformats via LSP). I'd remove other languages such as javascript too as suggested before. 1000 line .editorconfig is too much.

dsmiley: What's LSP?
Should I at least keep the 80 char line length, which is what Google's style guide [specifies for Python](https://google.github.io/styleguide/pyguide.html#32-line-length)?

.editorconfig
dsmiley: Opening a file shouldn't have any effect.  *Editing* a file would.  Since you have Eclipse and Windows, might you try please?

I can understand your suggestion though... I'm ambivalent since I think it's nice to declare a default.

dweiss: Ok, although I've seen a fair share of editors and some of them are... eager to convert stuff.  We'll see if this happens, I guess.

dweiss: (I neither have Eclipse nor Windows... I have been going back and forth over the years but am on Linux and MacOS for quite some time now; my Windows is on VM only).

dsmiley: Well if we find this to be annoying, we can always change it.  Progress not perfection.

dsmiley:
Are there any concerns with merging this in its current state?  It's trimmed down from the original.

dweiss:
Most of those settings apply to intellij, right? They are not some "standard" .editorconfig things. I'm fine with adding this - if something clashes with other settings, we can always tweak this file's configuration (for example, gradle and groovy files are auto-formatted on #14764).

dsmiley:
Yes; _most_ lines are IntelliJ specific.  Thanks for pointing to your recent build work... I'll want to adjust this this configuration to align with the Groovy style from there.

dsmiley:
I plan to merge this tonight (~10pm EST) if I don't hear advise to the contrary.

.editorconfig
rmuir: please, just remove the python completely. dont care about google style, its not relevant here.

rmuir:
please remove the python completely as we already have a python autoformatter and verifier (like spotless, except not slow as hell). We don't need conflicting editor configuration around it.

dsmiley:
> please remove the python completely as we already have a python autoformatter and verifier

I think you're missing a point of the value of this.  It's _complementary_ with Spotless (or similar).  It helps us write code formatted according to the project's standards during the writing/editing process, especially when utilizing IDE features that manipulate code (e.g. refactorings, perhaps code suggestion completions too).  Consequently, forgetting to run tidy/Spotless can be less of an annoyance... pushing more formatting earlier.

dweiss:
Just FYI - I noticed we already have an editorconfig file in scripts - 
https://github.com/apache/lucene/blob/main/dev-tools/scripts/.editorconfig

is this complementary? Could it be moved to the top level as part of this patch?

dsmiley:
OMG that's ironic!  @rmuir, you added it (in March), and it only configures Python :-)   LOL

Okay... well I think that file should be removed and it's python section integrated into the top-level file coming with this PR.  As to the very specifics of what it configures, I don't care (whatever you guys say) as Python isn't my thing.

dsmiley:
The existing editorconfig was redundant with information in this editorconfig, so I simply removed it.  I'll merge this tonight.

rmuir:
> The existing editorconfig was redundant with information in this editorconfig, so I simply removed it. I'll merge this tonight.

That's not true, now a maximum line length is set, inconsistent with the linter. I will fix it.

rmuir:
> I think you're missing a point of the value of this. It's _complementary_ with Spotless (or similar). It helps us write code formatted according to the project's standards during the writing/editing process, especially when utilizing IDE features that manipulate code (e.g. refactorings, perhaps code suggestion completions too). Consequently, forgetting to run tidy/Spotless can be less of an annoyance... pushing more formatting earlier.

@dsmiley .editorconfig has nothing to do with spotless or IDEs or anything like that. It configures editors... all editors.

If affects us all, for all files: even those people not using an IDE. When this PR was merged my editor (vim) was throwing errors left and right!

So I'd ask if we want to tweak intellij formatting, that if possible, we configure that intellij-specific stuff elsewhere in the project.

This file is very important, will be interpreted by all kinds of editors (not just IDE) so its best to be on the minimal side and be extremely careful around it.

msokolov:
This got me worried so I checked emacs editorconfig support https://www.gnu.org/software/emacs/manual/html_node/emacs/EditorConfig-support.html and it turns out it is off by default, phew.

msokolov:
This got me worried so I checked emacs editorconfig support https://www.gnu.org/software/emacs/manual/html_node/emacs/EditorConfig-support.html and it turns out it is off by default, phew.

dsmiley:
I'm confused how the presence of an `.editorconfig` resulted in errors; can you elaborate how that could happen?  Judging from your commit here just recently, was this a minor glitch due to where a comment was in this file?  Thanks for fixing that, and hopefully it's a non-issue now.  @msokolov I would appreciate it if you enable emacs's support for editorconfig and let us know if it was terrible/bad; you seem worried about it for some reason.

And how could the presence of IntelliJ specific matters interfere with someone not using IntelliJ such as yourself Rob?

rmuir:
.editorconfig is not an intellij specific file. It is parsed and respected by many editors, even ones you may not consider, such as github repository editor in the browser UI. So the file should really be simple and minimal: it is not the place to put a bunch of intellij-specific stuff. This is not the file to have complexity.

my editor complained loudly (rightfully) because the file wasn't syntactically valid.

msokolov:
> I would appreciate it if you enable emacs's support for editorconfig and let us know if it was terrible/bad; you seem worried about it for some reason.

I would rather do my manual formatting while I am working and let tidy take care of the rest. Or is this covering something tidy does not? If it is, could tidy be made to handle?

dweiss:
> I would rather do my manual formatting while I am working and let tidy take care of the rest.

I'm from the same camp. This is one of the reasons I like gjf - I can move things around and they half-magically revert back to the previous state. No need to think about formatting that much.

rmuir:
Personally I think there's just too much going on here. 
The file was rushed in too quickly and is far too aggressive (e.g. applying too many aggressive settings to ALL files).

It is better to proceed with caution around this file. An example would be the python file deleted here: https://github.com/apache/lucene/blob/f83587416ecaa7b5fb201dd3938fd06e2af95aba/dev-tools/scripts/.editorconfig

Because there was inconsistent formatting of the python files in that folder, it was introduced JUST for very specific filetype in JUST that specific folder. 

Trying to apply such settings to ALL files of ALL types in repo is too much, you need to deal with wacky file types (e.g. make with tabs), windows file types that should have CRLF, generated files, test files, files that aren't valid utf-8, and the list goes on and on.

I'm trying to cope regardless, but it isn't easy and I still think there's too much going on here. See #14819

dsmiley:
I definitely disagree on this PR being rushed.  You aren't a fan of it but that doesn't make it rushed.

> Trying to apply such settings to ALL files of ALL types in repo is too much, you need to deal with wacky file types (e.g. make with tabs), windows file types that should have CRLF, generated files, test files, files that aren't valid utf-8, and the list goes on and on.

I mostly sympathize.

As you & Dawid suggested, I can remove the initial everything glob, and instead copy the rules there now to the files types it lists.

rmuir:
I'm actually a huge fan of having editorconfig, 

I just know it can be tricky if you want to get it right (e.g. not have random style changes in PRs). Similar process to when we moved to spotless, fixing all sources to conform, marking the git commits special so they don't mess up `git blame`, dealing with the long tail issues described above.

rmuir:
If you don't mind, i'd like to get the CI piece working before making any big changes to this file. Then we can tone it down, fix it, etc, but changes to it will be validated rather than eyeballed. Just gimme a bit to get a PR up

Issue #14741, opened 2025-05-31
Failing test: TestSpellChecking.testGeneratedSuggestions — ComparisonFailure with expected suggestion list


N624-debu:
### Description

I'm encountering a consistent test failure when running the following test:

```
org.apache.lucene.analysis.hunspell.TestSpellChecking.testGeneratedSuggestions
```

**Git hash**: `ca427071708e3b7d17bc654a31712a5cdbb1bd7d`

**System Info**:
- OS: Windows 10 10.0 amd64
- Java: Eclipse Adoptium 24.0.1 (64-bit)
- Locale: `kam`, Timezone: `Europe/Uzhgorod`
- CPUs: 1, Threads: 1

**Error**:
```
org.junit.ComparisonFailure: Suggester=default expected:<kennen, kenne[ >     Theater, Heather, T...> but was:<kennen, kenne[] >     Theater, Heather, T...>
```

**Reproduction Command**:  
(Note: the original line printed had issues with quoting. Below is the corrected form)

```bash
gradlew :lucene:analysis:common:test --tests "org.apache.lucene.analysis.hunspell.TestSpellChecking.testGeneratedSuggestions" -Ptests.jvms=2 "-Ptests.jvmargs=-XX:+UseParallelGC -XX:ActiveProcessorCount=1" -Ptests.seed=320EBC31DC094689 -Ptests.useSecurityManager=false -Ptests.gui=false -Ptests.file.encoding=UTF-8 -Ptests.vectorsize=512 -Ptests.forceintegervectors=true
```

The printed `Reproduce with:` line had two bugs:
- JVM args weren’t quoted
- It was missing `-Ptests.jvmargs=` prefix

Let me know if I can help further debug this.

### Version and environment details

Lucene commit hash: ca427071708e3b7d17bc654a31712a5cdbb1bd7d
OS: Windows 10 10.0 amd64
JVM: Eclipse Temurin OpenJDK 24.0.1+9 (64-bit)
Locale: kam
Timezone: Europe/Uzhgorod
CPUs: 1

vigyasharma:
> The printed `Reproduce with:` line had two bugs:
> 
> * JVM args weren’t quoted
> * It was missing `-Ptests.jvmargs=` prefix

Independent of the test failure, which we still need to debug, this sounds like an opportunity to fix our test-failure messages. Would you like to open a PR for the same, @N624-debu ?

mhk2-cmu:
[...]

PR #14742, opened 2025-06-01
Fix: Include tests.jvmargs in "Reproduce with" output

mhk2-cmu:
Appended tests.jvmargs from system properties to support accurate reproduction lines.

### Description

This patch fixes an issue where the "Reproduce with" line generated by the test runner omits `tests.jvmargs`. This leads to loss of custom JVM flags (e.g., `-XX:+UseParallelGC`, `-XX:ActiveProcessorCount=1`) when attempting to reproduce test failures.

The fix appends the system property `tests.jvmargs` using `addVmOpt`, ensuring the JVM args used during test execution are preserved in the reproduction line.

Fix for #14741.


dweiss:
I think this is ok, although some of these flags are derived from the main randomization seed so - in theory - they should be identical if your seed is the same.

dweiss (approved):

PR #14748, opened 2025-06-02
Merge DocAndFreqBuffer and DocAndScoreBuffer.

jpountz:
I initially kept those separated but I'm now coming to the conclusion that they should be merged:
 - While frequencies are ints today, some queries produce float frequencies (`SloppyPhraseQuery`, `CombinedFieldQuery`).
 - `TermScorer` no longer needs to keep track of a separate buffer for frequencies.


lucene/core/src/java/org/apache/lucene/search/DocAndFloatFeatureBuffer.java
```diff
@@ -21,29 +21,32 @@
...
+public final class DocAndFloatFeatureBuffer {
+
+  private static final float[] EMPTY_FLOATS = new float[0];
 
   /** Doc IDs */
   public int[] docs = IntsRef.EMPTY_INTS;
 
-  /** Frequencies */
-  public int[] freqs = IntsRef.EMPTY_INTS;
+  /** Scores */
```
romseygeek:
```suggestion
  /** Float-valued features */
```

lucene/core/src/java/org/apache/lucene/search/DocAndFloatFeatureBuffer.java
```diff
@@ -21,29 +21,32 @@
...
+  public float[] features = EMPTY_FLOATS;
 
-  /** Number of valid entries in the doc ID and frequency arrays. */
+  /** Number of valid entries in the doc ID and score arrays. */
```
romseygeek:
```suggestion
  /** Number of valid entries in the doc ID and float-valued feature arrays. */
```

romseygeek (approved):
LGTM

jpountz:
It looks like the simplification in `TermScorer` helped speed things up a bit: https://benchmarks.mikemccandless.com/Term.html. I'll push an annotation.

PR #14749, opened 2025-06-02
Add code comment about use of Collections#synchronizedMap in LRUQueryCache.

### LINKED ISSUE #14677: Does LRUQueryCache#uniqueQueries really needs wrapped by Collections.synchronizedMap?
### Description

`LRUQueryCache#uniqueQueries` is wrapped by `Collections.synchronizedMap`, but we use it under `readLock/writeLock`.

jpountz:
It may not be totally obvious why this map needs to be synchronized since reads and writes are performed under a read-write lock. Except that unlike most other collections, reads are not thread-safe on a `LinkedHashMap`.

Closes #14677


benwtrent (approved):
Thanks @jpountz !

vsop-479:
Thanks @jpountz !

PR #14753, opened 2025-06-04
Implement IndexedDISI#docIDRunEnd

HUSTERGS:
### Description
This try to implement `IndexedDISI#docIDRunEnd`

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -140,6 +140,8 @@ Optimizations
 
 * GITHUB#14720: Cache high-order bits of hashcode to speed up BytesRefHash. (Pan Guixin)
 
+* GITHUB#14753: Implement IndexedDISI#docIDRunEnd. (Ge Song)
+
 Bug Fixes
 ---------------------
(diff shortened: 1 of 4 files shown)


gf2121:
Thanks for contribution! I wonder if we should also implement `SparsexxxDocValues#DocIdRunEnd` in `Lucene90DocValuesProducer` so that this can be actually used in queries like `FieldExistsQuery`?

HUSTERGS:
> Thanks for contribution! I wonder if we should also implement `SparsexxxDocValues#DocIdRunEnd` in `Lucene90DocValuesProducer` so that this can be actually used in queries like `FieldExistsQuery`?

@gf2121 Of Course, just pushed another commit containing corresponding implementation

gf2121 (approved):
LGTM

gf2121:
Could you add a CHANGE entry under 10.3.0?

HUSTERGS:
> Could you add a CHANGE entry under 9.3.0?

Yes, I forgot that, pushed another commit

PR #14757, opened 2025-06-04
Fix too many documents collected when only bool-filter condition is present

kkewwei:
### Description

New logic is consistent with the implementation of 9.8
https://github.com/apache/lucene/blob/releases/lucene/9.8.0/lucene/core/src/java/org/apache/lucene/search/Boolean2ScorerSupplier.java#L112

Closes: #14755

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -57,6 +57,8 @@ Bug Fixes
   returns a string which can be parsed back into the original node.
   (Peter Barna, Adam Schwartz)
 
+* GITHUB#14755: Fix too many documents collected when only bool-filter condition is present. (Ke Wei)
+
 Changes in Runtime Behavior
 ---------------------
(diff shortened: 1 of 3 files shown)


jpountz (approved):

jpountz:
Thank you!

jpountz:
FYI I backported to branch_10x so that it's included in 10.3.

uschindler:
The changes entry needs to be moved.

kkewwei:
> The changes entry needs to be moved.

@uschindler Kindly advise whether it is should be placed under the "Improvements" or `Optimizations`?

uschindler:
It's only about the version number. It was a message for @jpountz who merged this.

uschindler:
He did that already.

jpountz:
Yup, I didn't want to bother @kkewwei with this so I did it myself, it's easy enough. :)

PR #14759, opened 2025-06-05
IndexOrDocValuesQuery and IndexSortSortedNumericDocValuesRangeQuery should only be counted once when computing maxClauseCount

### LINKED ISSUE #14756: IndexOrDocValues query is counted twice when computing `maxClauseCount`
I noticed that if I change a query from a PointRangeQuery to an IndexOrDocValuesQuery, then the change might make previous valid boolean queries to fail because IndexOrDocValuesQuery counts twice when computing `maxClauseCount.

I think this is wrong as IndexOrDocValues is just providing lucene with the choice of two execution paths and should not count twice for that limit.

This test proves the issue:

```
 public void testIndexOrDocValues() throws IOException {
    RandomIndexWriter writer = new RandomIndexWriter(random(), dir);
    Document d = new Document();
    d.add(new LongField("foo", 0L, LongField.Store.NO));
    writer.addDocument(d);
    d = new Document();
    d.add(new LongField("foo", Long.MAX_VALUE, LongField.Store.NO));
    writer.addDocument(d);

    IndexReader reader = writer.getReader();
    IndexSearcher searcher = newSearcher(reader);
    writer.close();
    int maxClauseCount = IndexSearcher.getMaxClauseCount();
    BooleanQuery.Builder qb = new BooleanQuery.Builder();

    for (int i = 0; i < maxClauseCount; i++) {
      qb.add(LongPoint.newRangeQuery("foo", 0, i), BooleanClause.Occur.SHOULD);
    }
    // should not throw an exception, because  it is below the limit
    searcher.rewrite(qb.build());

    qb = new BooleanQuery.Builder();
    for (int i = 0; i < maxClauseCount; i++) {
      qb.add(LongField.newRangeQuery("foo", 0, i), BooleanClause.Occur.SHOULD);
    }
    // should not throw an exception, because  it is below the limit
    searcher.rewrite(qb.build());
  }
```

It fails in the second rewrite with the exception:

```
org.apache.lucene.search.IndexSearcher$TooManyNestedClauses: Query contains too many nested clauses; maxClauseCount is set to 1024
	at __randomizedtesting.SeedInfo.seed([1DBE2F81F126780A:AB3F95EEC9898FF7]:0)
	at org.apache.lucene.search.IndexSearcher$1.visitLeaf(IndexSearcher.java:905)
	at org.apache.lucene.document.SortedNumericDocValuesRangeQuery.visit(SortedNumericDocValuesRangeQuery.java:74)
	at org.apache.lucene.search.IndexOrDocValuesQuery.visit(IndexOrDocValuesQuery.java:134)
	at org.apache.lucene.search.BooleanQuery.visit(BooleanQuery.java:661)
```

iverase:
see https://github.com/apache/lucene/issues/14756

The idea proposed here is that both queries are just leaf queries and should only call once the method QueryVisitor#visitLeaf. For IndexOrDocValuesQuery is a bit more complicated because the query does not contain the field that is being queried, therefore it delegates the registration of the query to the visitor to the index query.

closes https://github.com/apache/lucene/issues/14756

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/IndexOrDocValuesQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/IndexOrDocValuesQuery.java
@@ -130,8 +130,9 @@ public final class IndexOrDocValuesQuery extends Query {
   @Override
   public void visit(QueryVisitor visitor) {
     QueryVisitor v = visitor.getSubVisitor(BooleanClause.Occur.MUST, this);
+    // we cannot register this query in the visitor, since we don't have the field name,
+    // so we just visit one of the subqueries.
     indexQuery.visit(v);
-    dvQuery.visit(v);
   }
 
   @Override
--- a/lucene/core/src/java/org/apache/lucene/search/IndexSortSortedNumericDocValuesRangeQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/IndexSortSortedNumericDocValuesRangeQuery.java
@@ -113,7 +113,6 @@ public class IndexSortSortedNumericDocValuesRangeQuery extends Query {
   public void visit(QueryVisitor visitor) {
     if (visitor.acceptField(field)) {
       visitor.visitLeaf(this);
-      fallbackQuery.visit(visitor);
     }
   }
 
--- a/lucene/core/src/test/org/apache/lucene/search/TestMaxClauseLimit.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestMaxClauseLimit.java
@@ -19,8 +19,15 @@ package org.apache.lucene.search;
 
 import java.io.IOException;
 import java.util.Arrays;
+import org.apache.lucene.document.Document;
+import org.apache.lucene.document.LongField;
+import org.apache.lucene.document.LongPoint;
+import org.apache.lucene.document.SortedNumericDocValuesField;
+import org.apache.lucene.index.IndexReader;
 import org.apache.lucene.index.MultiReader;
 import org.apache.lucene.index.Term;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.tests.index.RandomIndexWriter;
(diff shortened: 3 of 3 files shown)


iverase:
@romseygeek sorry for the direct ping but as you were involved in the original development of the query visitor API, do you have any opinions here? what is a leave in the query tree?

romseygeek:
QueryVisitor is supposed to be a query introspection API, so I wouldn't change its behaviour here.  If the problem is that we want to be able to query multiple ranges without creating giant boolean queries, then I'd suggest instead looking at adding a `newSetRangeQuery` method to `LongField`, and maybe promoting `LongMultiRangeBuilder` from the sandbox?

iverase:
My problem is that I want to use IndexOrDocValues query without changing the behaviour of the queries. This is a bit weird as the benefit of that query happens when it used on boolean queries but with the current implementation it is bogus as the behaviour change, e.g. it is not an optimisation anymore. 

For what you say, then the way we compute maxClauseCount is bogus and we should not be using the QueryVisitor?

romseygeek:
Well it's still an optimization, but just at the cost of a bigger query footprint.

I think there are a couple of things pulling in different directions here:
* IndexOrDocValuesQuery trades off in-memory complexity for better query-time performance
* We have the too many clauses check to push people towards using queries that use less memory, eg TermInSetQuery instead of very large disjunctions of TermQuery

So what you're running into there is the limit of that trade-off.  I don't think anything is wrong here, I think this is telling us that if you want to do a disjunction over ranges then past a certain point just building a boolean query is going to be very inefficient and you should use a different type of query.  Maybe one that we don't actually provide yet!

iverase:
I think the current behaviour is trappy, I would expect the test I post on the issue to pass. I agree though that building those big boolean queries is not optimal, but that's what users do sometimes.

iverase:
Maybe we should remove the concept of  `maxClauseCount` as we are not counting those but something else, query complexity?

PR #14762, opened 2025-06-09
Add non-committer activity to activity report

dweiss:
This adds a section to display non-committer activity in our git repository. This is useful in preparing board reports but also considering committer nominations, I think. Looks something like this:
```
    19 Simon Cooper | simon.cooper@elastic.co
      6 panguixin | panguixin@bytedance.com
      4 zhouhui | vsop_479@163.com
      4 Viliam Durina | viliam-durina@users.noreply.github.com
      4 hanbj | hanbj0707@163.com
      4 guojialiang | guojialiang.2012@bytedance.com
      4 Amos Bird | amosbird@gmail.com
```

I had to add existing committer names (and a few aliases sprinkled in) as this information is not readily available from the ASF roster (committer names are public but not emails or gh aliases). I don't think this is a privacy violation since these names and handles are part of the git commit history anyway.

### DIFF
--- a/.github/workflows/activity-report.sh
+++ b/.github/workflows/activity-report.sh
@@ -47,6 +47,12 @@ git log --all --pretty='format:%an' --since="$SINCE" --before="$UNTIL" | sort | 
 echo '```'
 
 echo
+echo "## Top non-committer contributors in the given time period (all commits, any branch)"
+echo '```'
+git log --all --pretty='format:%an | %ae' --since="$SINCE" --before="$UNTIL" | sort | uniq -c | sort -r -n | grep -v -f .github/workflows/activity-report-known-committers.txt
+echo '```'
+
+echo
 echo "## All pull requests:"
 echo '```'
 gh pr list --state all --search "created:$SINCE_TS..$UNTIL_TS" --repo $REPO --limit 1000 \
--- a/.github/workflows/activity-report-known-committers.txt
+++ b/.github/workflows/activity-report-known-committers.txt
@@ -1,0 +1,110 @@
+# a list of known committer names and their aliases, to highlight 
+# commits from non-committers in the activity-report.
+dependabot
+guofeng.my
+gf2121
+Benjamin Trent
+ChrisHegarty
+Andrzej Bialecki
+Alessandro Benedetti
+Armin Braun
+Adriano Crestani Campos
+Ramkumar Aiyengar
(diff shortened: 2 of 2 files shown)


stefanvodita (approved):
This is great for making committership more transparent and consistent! Do we have a mechanism to update this list? Maybe when we send instructions to the new committer to update the who-are-we webpage, we should also mention this file?

dweiss:
> Do we have a mechanism to update this list? Maybe when we send instructions to the new committer to update the who-are-we webpage, we should also mention this file?

I don't think there is a need, really - the chair should, periodically, generate the report; if there is anybody standing out, it'll take 2 seconds to add a name if it's a committer's name. It's not really mechanical since people use different e-mails, names, etc. So I don't think it's worth automating, to be honest - it'll take more time to dodge around exceptional cases than it'll take to add them manually for the next 50 years. After which, I don't think I'll care that much. :)

.github/workflows/activity-report.sh
```diff
@@ -46,6 +46,12 @@ echo '```'
 git log --all --pretty='format:%an' --since="$SINCE" --before="$UNTIL" | sort | uniq -c | sort -r -n
 echo '```'
 
+echo
+echo "## Top non-committer contributors in the given time period (all commits, any branch)"
+echo '```'
+git log --all --pretty='format:%an | %ae' --since="$SINCE" --before="$UNTIL" | sort | uniq -c | sort -r -n | grep -v -f .github/workflows/activity-report-known-committers.txt
```
mikemccand, line 52: Go Unix pipes!

mikemccand (approved):
Very cool!

mikemccand:
We might also similarly (separately!  no need to block this awesome tool) track committers who are still active but not yet included on the Lucene PMC.

dweiss:
> We might also similarly (separately! no need to block this awesome tool) track committers who are still active but not yet included on the Lucene PMC.

Yes, sure. Just wanted to point out that git activity isn't the only way to contribute (and be a contributor). Addressing mailing list questions or improving the documentation could also merit committer status. This said, Lucene is a bit low-level in this regard in that most activity takes place in git[hub] anyway.

dweiss:
I've merged this in. Here is an example output for the last 9 months.
https://github.com/apache/lucene/actions/runs/15533320029

image

Issue #14763, opened 2025-06-09
Are we spending too much CPU in HNSW `ramBytesUsed`, and clearing bitsets?


mikemccand:
### Description

Peeking at the last nightly benchmark data point, I see the top CPU hotspots during indexing.

That 2nd one (13.12% in `ramBytesUsed`) is concerning ... I know we need to account properly for RAM so IW can flush when RAM exceeds its allowance ... but maybe we can optimize how we do that for HNSW?

8.63% spent clearing bitsets for HNSW searching is also scary -- that likely impacts search performance too (since building an HNSW graph is done by doing a search for each inserted vector)?

Also what exactly is `reduceLanesTemplate`?  I find this name very non-intuitive :)  Is it essentially a cast (like `long` -> `int`) for a vector?

```
Profiler for cpu:
WARNING: Using incubator modules: jdk.incubator.vector
PROFILE SUMMARY from 4433882 events (total: 4M)
  tests.profile.mode=cpu
  tests.profile.count=50
  tests.profile.stacksize=4
  tests.profile.linenumbers=false
PERCENT       CPU SAMPLES   STACK
23.57%        1M            jdk.incubator.vector.FloatVector#reduceLanesTemplate() [Inlined code]
                              at jdk.incubator.vector.Float256Vector#reduceLanes() [Inlined code]
                              at org.apache.lucene.internal.vectorization.PanamaVectorUtilSupport#dotProductBody() [JIT compiled code]
                              at org.apache.lucene.internal.vectorization.PanamaVectorUtilSupport#dotProduct() [Inlined code]
13.12%        581719        org.apache.lucene.util.hnsw.NeighborArray#ramBytesUsed() [Inlined code]
                              at org.apache.lucene.util.hnsw.OnHeapHnswGraph#updateGraphRamBytesUsed() [JIT compiled code]
                              at org.apache.lucene.util.hnsw.OnHeapHnswGraph#addNode() [JIT compiled code]
                              at org.apache.lucene.util.hnsw.HnswGraphBuilder#addGraphNodeInternal() [JIT compiled code]
9.51%         421729        org.apache.lucene.index.FloatVectorValues$1#vectorValue() [Inlined code]
                              at org.apache.lucene.codecs.hnsw.DefaultFlatVectorScorer$FloatScoringSupplier$1#score() [Inlined code]
                              at org.apache.lucene.util.hnsw.HnswGraphSearcher#searchLevel() [JIT compiled code]
[...]

PR #14765, opened 2025-06-09
Improve hnsw on heap ram est

benwtrent:
This makes OnHeapGraph builder RAM estimates much cheaper by incrementally updating the estimate as nodes are added to the NeighborArray.

Local benchmarking shows almost no performance impact, but I can no longer see any ram estimate methods in the top methods given the CPU samples. 

One outstanding bit here is that `OnHeapGraph#graphRamBytesUsed` is only volatile. So, its not really threadsafe for estimates. However, from what I can tell, the only time we actually use this is during initial flush, and that is within a single thread anyways, right?

Follow on from: https://github.com/apache/lucene/pull/14527

related: https://github.com/apache/lucene/issues/14763


mikemccand:
> Local benchmarking shows almost no performance impact, but I can no longer see any ram estimate methods in the top methods given the CPU samples.

Whoa, thank you for the quick PR @benwtrent!  That's disappointing that there's no speedup on local benchmark, odd.  I wonder if the nightly profiler (JFR) is somehow giving ghosts ... it should be async (no safepoint bias)?  Or maybe it's an env/benchy difference and once we merge this we'll maybe see gains in the nightly benchy.

mikemccand:
Hmm, not sure it matters, but the JFR result I linked to in #14763 was from the single-threaded (deterministic, `SerialMergeScheduler`, takes forever) indexing run.  I don't think it should matter ... that shouldn't cause unusual profiler ghosts.

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
```diff
@@ -444,7 +444,6 @@ void finish() throws IOException {
     //  see: https://github.com/apache/lucene/issues/14214
     // connectComponents();
     frozen = true;
-    hnsw.finishBuild();
```
mikemccand, line 447: Hmm, why remove this?

lucene/core/src/java/org/apache/lucene/util/hnsw/OnHeapHnswGraph.java
```diff
@@ -32,6 +32,15 @@
  */
 public final class OnHeapHnswGraph extends HnswGraph implements Accountable {
 
+  private static final long RAM_BYTES_USED =
+      4L * Integer.BYTES // all int fields
+          + 1 // field: noGrowth
+          + RamUsageEstimator.NUM_BYTES_OBJECT_REF
```
mikemccand: Is this for the `graph` pointer we hold?

lucene/core/src/java/org/apache/lucene/util/hnsw/OnHeapHnswGraph.java
```diff
@@ -32,6 +32,15 @@
...
+          + RamUsageEstimator.NUM_BYTES_OBJECT_HEADER
+          + 2 * Integer.BYTES // field: entryNode
+          + 3L * (Integer.BYTES + RamUsageEstimator.NUM_BYTES_OBJECT_HEADER) // 3 AtomicInteger
```
mikemccand: Also add in `RamUsageEstimator.NUM_BYTES_OBJECT_REF` for the pointer to these `AtomicInteger`?

mikemccand: Maybe ask Claude or Q or CoPiolot or ChatGPT or Gemini or so to generate this big long `RAM_BYTES_USED`?  It's so error proned and time consuming for we humans ... maybe genai can do it one-time quickly / more reliably for this PR.

lucene/core/src/java/org/apache/lucene/util/hnsw/OnHeapHnswGraph.java
```diff
@@ -158,21 +168,30 @@ public void addNode(int level, int node) {
...
     if (level == 0) {
-      graph[node][level] = new NeighborArray(nsize0, true);
+      graph[node][level] =
+          new NeighborArray(
+              nsize0,
+              true,
+              l -> {
+                long bytesUsed = graphRamBytesUsed;
+                graphRamBytesUsed = bytesUsed + l;
+                assert l > 0;
```
mikemccand: Why both asserts?  Maybe move this `assert` up to the top of the lambda?

mikemccand (commented):
I like this incremental approach instead!  I was confused on whether we are missing some tracking in that base `RAM_BYTES_USED`.

I think your reasoning about thread safety is correct (we only track RAM during initial flush, and that is single threaded).  Let's at least add comment / javadoc somewhere that explains this precarious thread safety situation?  Who knows maybe someday `IndexWriter` will need to account for RAM consumed by merging...

lucene/core/src/java/org/apache/lucene/util/hnsw/HnswGraphBuilder.java
benwtrent: @mikemccand its not used. It was only added for the RAM estimate added in the previous PR. Since we are doing incremental updates, I just removed it.

lucene/core/src/java/org/apache/lucene/util/hnsw/OnHeapHnswGraph.java
benwtrent: I am switching all of this to a simple `RamUsageEstimator.shallowSizeOfInstance(OnHeapHnswGraph.class);` attempting to account for fields directly is just too complicated

lucene/core/src/java/org/apache/lucene/util/hnsw/OnHeapHnswGraph.java
benwtrent: ++ Yep! I will simplify, I was testing somethings out and left the other assert :/

lucene/core/src/test/org/apache/lucene/util/hnsw/HnswGraphTestCase.java
```diff
@@ -786,6 +786,7 @@ public void testHnswGraphBuilderInvalid() throws IOException {
         IllegalArgumentException.class, () -> HnswGraphBuilder.create(scorerSupplier, 10, 0, 0));
   }
 
+  @com.carrotsearch.randomizedtesting.annotations.Repeat(iterations = 100)
```
ChrisHegarty: leftover, or intentional ?

ChrisHegarty (approved):
LGTM

lucene/core/src/test/org/apache/lucene/util/hnsw/HnswGraphTestCase.java
benwtrent: LOL, of course, I want CI to REALLY exercise this test ;)

weizijun:
@benwtrent Thanks for helping improve `OnHeapGraph#graphRamBytesUsed`!

lucene/core/src/test/org/apache/lucene/util/hnsw/HnswGraphTestCase.java
dweiss: you can always override this locally by adding "-Ptests.iters=100".

PR #14770, opened 2025-06-11
Fix map size in CustomAnalyzer#paramsToMap.

vsop-479:

### DIFF
--- a/lucene/analysis/common/src/java/org/apache/lucene/analysis/custom/CustomAnalyzer.java
+++ b/lucene/analysis/common/src/java/org/apache/lucene/analysis/custom/CustomAnalyzer.java
@@ -606,7 +606,7 @@ public final class CustomAnalyzer extends Analyzer {
         throw new IllegalArgumentException(
             "Key-value pairs expected, so the number of params must be even.");
       }
-      final Map<String, String> map = CollectionUtil.newHashMap(params.length);
+      final Map<String, String> map = CollectionUtil.newHashMap(params.length / 2);
       for (int i = 0; i < params.length; i += 2) {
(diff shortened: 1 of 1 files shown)


uschindler (approved):
Looks fine, I think this was an oversight.

Anyways we should remove and deprecated CollectionUtil, see comment here:
https://github.com/apache/lucene/blob/a55be5cb368cba698114f63bc8e8be2a2a55b089/lucene/core/src/java/org/apache/lucene/util/CollectionUtil.java#L41-L44

We're on Java 19 since long time! :-) So both Lucene 10 and Main can use the JDK provided method. Let's open a PR!

uschindler:
Let's backport this to 10.x

vsop-479:
> We're on Java 19 since long time! :-) So both Lucene 10 and Main can use the JDK provided method. Let's open a PR!

OK, I will do it.

uschindler:
Hi,
I think we should add an extra buffer of 1 entries to the has map, because the code later adds 1 extra default entries to the map, so it will be resized.

https://github.com/vsop-479/lucene-fork/blob/f0be57db06f3ba821dd12815c8a55e3ea2757888/lucene/analysis/common/src/java/org/apache/lucene/analysis/custom/CustomAnalyzer.java#L596

vsop-479:
> I think we should add an extra buffer of 1 entries to the has map, because the code later adds 1 extra default entries to the map, so it will be resized.

OK, I will fix it.

vsop-479:
> I think we should add an extra buffer of 1 entries to the has map, because the code later adds 1 extra default entries to the map, so it will be resized.

Added with https://github.com/apache/lucene/pull/15517.

PR #14771, opened 2025-06-12
Optimize filtered kNN search with FixedBitSet intersections

mrkm4ntr:
### Description
If a filter query in kNN is in query cache, it can be saved as a `BitSetIterator` wrapping a `FixedBitSet`. But we are creating new `FixedBitSet` combined with liveDocs for every query. This is really high cost because it needs to iterate over long BitSet in `applyMask`.
Instead of creating new BitSet by iteration, I will create new BitSet implementing only reading methods. By using this, we don't need to iterate over BitSet. We can leverage `intersectionCount` of FixedBitSet that is super fast.

This is the flamegraph when I ran filtered kNN queries (k=100, num_candidates=100, term filter that filters out about 50% of documents).
スクリーンショット 2025-06-12 16 21 14

After applying this, `applyMask` disappeared completely.
スクリーンショット 2025-06-12 16 24 43

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
@@ -42,8 +42,8 @@ import org.apache.lucene.search.knn.TopKnnCollectorManager;
 import org.apache.lucene.util.BitSet;
 import org.apache.lucene.util.BitSetIterator;
 import org.apache.lucene.util.Bits;
-import org.apache.lucene.util.FixedBits;
 import org.apache.lucene.util.FixedBitSet;
+import org.apache.lucene.util.FixedBits;
 
 /**
  * Uses {@link KnnVectorsReader#search} to perform nearest neighbour search.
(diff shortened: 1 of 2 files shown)


benwtrent:
This is interesting indeed!

I wonder how it performs with a bunch of random deletions (say 20%) and combined with filters (30+% filtered out).

@gf2121 is the bitset guru ;) Maybe they have some ideas here.

mrkm4ntr:
I forgot to mention but during taking above profile, I was running continuous document updates. Documents deletion ratio is constantly about 20%.

benwtrent:
> I forgot to mention but during taking above profile, I was running continuous document updates. Documents deletion ratio is constantly about 20%.

Ah, fabulous @mrkm4ntr !!!

Did you notice an improvement in latency/throughput for the queries? 

I suppose the latency is still dominated by the vector ops later. But would be interested to know if there was improvements at all :)

gf2121:
Thanks for contribution, I'm actually a bit confused on your description.

According to the flame graph your provide, it seems like it is running default implementation of `Bits#applyMask`. But as far as i know `liveDocs` is mostly `FixedBits` (And your change is optimizing this case as well), and `FixedBits` has override `applyMask` to run bitset intersection?

https://github.com/apache/lucene/blob/686ef91e15f6be22aa6d2d55335d0e5160636c6a/lucene/core/src/java/org/apache/lucene/util/FixedBits.java#L34-L36

Could you provide more context on how did you catch this flame graph and why your change helps?

mrkm4ntr:
Ah, I patched some changes to Lucene 9 by myself. But I missed this override...
After overriding this, the flame graph became like this. So my change is not needed at all. Sorry for bothering you.
スクリーンショット 2025-06-13 23 49 49

PR #14773, opened 2025-06-12
Add prefetching for terms dict in doc values

easyice:
This follows a similar approach as doc values and only prefetches the first page of data. Perhaps these were missed  at the time?


jpountz:
Doing this is a bit less obvious in my mind since terms dictionaris are allowed to have a random access pattern, when doc-value iterators are required to be consumed in order?

Somewhat separately, I've been wondering if we need to APIs, one for saying "I'm going to use this field" and pass the full range of data, and another one to say "I'm going to use this specific range of bytes", which would be what `IndexInput#prefetch` does today.

easyice:
Thanks for the explanation, You're right, I'll close this PR. I feel an API to prefetch full range of data could be useful, e.g. for files like `.tip`, which are relatively small but have a significant impact on read performance?

PR #14774, opened 2025-06-12
Use IntArrayList/IntHashSet to replace usages of List/Set of Integer

easyice:
No functional changes — only optimization to reduce auto-boxing. However, this involves a public API change in `UpdateGraphsUtils#computeJoinSet`. not sure whether we should touch this public API?


dweiss (approved):
LGTM. I think it's fine to change the signature of those low-level methods.

lucene/core/src/java/org/apache/lucene/util/hnsw/UpdateGraphsUtils.java
```diff
@@ -37,11 +36,11 @@ public class UpdateGraphsUtils {
    *
    * @return a set of nodes that best cover the graph
    */
-  public static Set<Integer> computeJoinSet(HnswGraph graph) throws IOException {
+  public static IntHashSet computeJoinSet(HnswGraph graph) throws IOException {
```
benwtrent, line 41: Honestly, this whole class should be marked as internal. 

I think making this change is perfectly fine.

benwtrent (approved):
Could you add a change log note? I realize that this is more of a refactoring/performance thing.

lucene/core/src/java/org/apache/lucene/util/hnsw/UpdateGraphsUtils.java
easyice: Thanks for the guidance. I’ve added `@lucene.internal` for this class.

easyice:
@dweiss @benwtrent Thanks for the quick review, I have added the CHANGES entry under 10.3.0

PR #14775, opened 2025-06-12
Add the ability to inverse a Sort

HoustonPutman:
### Description

Currently there is no easy way to reverse a Sort unless you know how that sort was created. Adding an `inverse` option gives users an easy way to reverse a given sort.

In Solr, I plan on using this to implement reverse pagination (cursorMark).

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/document/FeatureSortField.java
+++ b/lucene/core/src/java/org/apache/lucene/document/FeatureSortField.java
@@ -45,6 +45,11 @@ final class FeatureSortField extends SortField {
   }
 
   @Override
+  public SortField inverseSort() {
+    throw new UnsupportedOperationException("Inverse sort not supported on FeatureSortField");
+  }
+
+  @Override
   public FieldComparator<?> getComparator(int numHits, Pruning pruning) {
     return new FeatureComparator(numHits, getField(), featureName);
   }
--- a/lucene/core/src/java/org/apache/lucene/document/LatLonPointSortField.java
+++ b/lucene/core/src/java/org/apache/lucene/document/LatLonPointSortField.java
@@ -39,6 +39,11 @@ final class LatLonPointSortField extends SortField {
   }
 
   @Override
+  public SortField inverseSort() {
+    throw new UnsupportedOperationException("Inverse sort not supported on LatLonPointSortField");
+  }
+
+  @Override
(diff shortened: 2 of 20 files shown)


jpountz:
I'm unsure about this PR for a few reasons:

1. It sounds relatively easy to do on the application side by using `Sort` factories that can either produce a `Sort` object in the natural or inverse order and creating this `Sort` object at the last moment when we know what order is needed.

2. It doesn't seem to handle ties correctly, you probably need to append a `SortField` that sorts by descending doc ID, since Lucene tie-breaks by ascending doc ID by default. But then it's a bit annoying that the inverse `Sort` object produces different sort values compared with the original `Sort` object?

3. The handling of missing values probably needs changing, inversing a `SortField` on a String field that it configured to sort missing values last should probably sort missing values first and vice-versa?

I wonder if there's a better way of doing this, maybe something like adding support for `searchBefore`, which would do the opposite of `searchAfter` (I'm thinking out loud, I having given much thoughts to this).

HoustonPutman:
> I wonder if there's a better way of doing this, maybe something like adding support for `searchBefore`, which would do the opposite of `searchAfter` (I'm thinking out loud, I having given much thoughts to this).

I would be all for this, but it seemed like a harder/more intrusive change.

> It doesn't seem to handle ties correctly, you probably need to append a SortField that sorts by descending doc ID, since Lucene tie-breaks by ascending doc ID by default. But then it's a bit annoying that the inverse Sort object produces different sort values compared with the original Sort object?

That's fair, it wasn't something I was considering since in our use case, the uniqueId must be a part of the sort. (No ties)

> The handling of missing values probably needs changing, inversing a SortField on a String field that it configured to sort missing values last should probably sort missing values first and vice-versa?

Again, fair. In my mind, Lucene only had a "missing value" not a "sort missing last" option, like Solr does. But you are right, there would be no way to fix that when using this `inverse()` method.

Thanks for the feedback, I think it'll probably be best to do this in the Application, as you mentioned. (Or have a searchBefore() option, which would be great!)

jpountz:
Thinking a bit more about it, adding `searchBefore` sounds like it could work. Would you like to give it a try?

msokolov:
Should we refuse to allow inverse of Sort.SCORE? or RELEVANCE?

PR #14776, opened 2025-06-12
Add a rescorer that uses DoubleValuesSource values to re-score first pass hits

vigyasharma:
A `Rescorer` that uses values from provided `DoubleValuesSource` to re-score top N hits of a query.

Enables us to rescore hits from ANN vector queries using full-precision or late-interaction DoubleValuesSources (#14729 and #14708).

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSourceRescorer.java
+++ b/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSourceRescorer.java
@@ -77,12 +77,12 @@ public abstract class DoubleValuesSourceRescorer extends Rescorer {
     }
 
     if (topN < hits.length) {
-      ArrayUtil.select(hits, 0, hits.length, topN, ScoreDoc.scoreDocComparator);
+      ArrayUtil.select(hits, 0, hits.length, topN, ScoreDoc.COMPARATOR);
       ScoreDoc[] subset = new ScoreDoc[topN];
       System.arraycopy(hits, 0, subset, 0, topN);
       hits = subset;
     }
-    Arrays.sort(hits, ScoreDoc.scoreDocComparator);
+    Arrays.sort(hits, ScoreDoc.COMPARATOR);
 
     return new TopDocs(firstPassTopDocs.totalHits, hits);
   }
--- a/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSourceRescorer.java
+++ b/lucene/core/src/java/org/apache/lucene/search/DoubleValuesSourceRescorer.java
@@ -1,0 +1,93 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


lucene/core/src/java/org/apache/lucene/search/DoubleValuesSourceRescorer.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+    Arrays.sort(hits, ScoreDoc.scoreDocComparator);
+
+    return new TopDocs(firstPassTopDocs.totalHits, hits);
+  }
+
+  @Override
+  public Explanation explain(IndexSearcher searcher, Explanation firstPassExplanation, int docID)
+      throws IOException {
+    Explanation first =
```
dungba88, line 93: Maybe it should be moved to when it is needed, such as L120?

lucene/core/src/java/org/apache/lucene/search/DoubleValuesSourceRescorer.java
```diff
@@ -0,0 +1,133 @@
...
+
+    LeafReaderContext leafWithDoc = null;
+    for (LeafReaderContext ctx : searcher.getIndexReader().leaves()) {
+      if (docID >= ctx.docBase && docID < (ctx.docBase + ctx.reader().maxDoc())) {
+        leafWithDoc = ctx;
+        break;
+      }
+    }
+    if (leafWithDoc == null) {
+      throw new IllegalArgumentException(
```
dungba88, line 105: Why we are throwing exception here instead of Explanation.noMatch?

vigyasharma: Yeah, this is a tricky one. Since this explain API can take any docId, I can see the argument of surfacing a "noMatch" here. However, my thinking is that `Rescorer#explain` API is meant to explain how the _Rescorer_ would score this docId for given searcher, when `rescore()` is called.

If we hit null here, it means that the docId was not matched by the provided searcher in the first-pass itself. This is likely because a different searcher was passed than the one that produced provided docId, which is an error that we should surface.

Suppose we say `noMatch` for first pass, and the docId does have a value in DoubleValuesSource. Then the explain API will return a score for the docId, However, the actual rescore API will never return a score for this (searcher, docId) combination because they don't match.

lucene/core/src/java/org/apache/lucene/search/DoubleValuesSourceRescorer.java
vigyasharma: My flow was to produce first pass explanation, then second explanation, then combined score and final explanation.

lucene/core/src/java/org/apache/lucene/search/DoubleValuesSourceRescorer.java
dungba88: Gotcha, thanks for the explanation. Maybe we can add some nuance you mentioned above to the error message, if the only possible case is a different IndexSearcher was passed? Anyhow, no strong opinion here.

lucene/core/src/java/org/apache/lucene/search/ScoreDoc.java
```diff
@@ -51,4 +52,16 @@ public ScoreDoc(int doc, float score, int shardIndex) {
   public String toString() {
     return "doc=" + doc + " score=" + score + " shardIndex=" + shardIndex;
   }
+
+  /** Utility comparator that sorts by score descending, then by docId ascending */
+  public static final Comparator<ScoreDoc> scoreDocComparator =
```
romseygeek: Nit: constants are usually in uppercase, so I'd rename this to just `COMPARATOR` as it will always be accessed like `ScoreDoc.COMPARATOR`.

romseygeek (approved):
Nice!

As a follow up, I think we should be able to re-implement QueryRescorer using this class and `DoubleValuesSource#fromQuery(Query)`

PR #14777, opened 2025-06-13
Add list initial capacity in FirstPassGroupingCollector#getTopGroups.

vsop-479:


vigyasharma (approved):
Looks good. Out of curiosity, what motivated this change? Was it performance related or general code hygiene?

vsop-479:
> what motivated this change? Was it performance related or general code hygiene?

Thanks @vigyasharma , Since we use grouping function (self-developed base on `FirstPassGroupingCollector`)  in our application , I am trying to improve the performance (https://github.com/apache/lucene/pull/14813). This change just related it.

PR #14780, opened 2025-06-14
Fix deprecated Gradle groovy dsl syntax on property assignments

breskeby:
### Description

Replaces usages of deprecated Gradle DSL syntax of using "propName value". This will be removed in future Gradle versions and should not be used anymore.


dweiss (approved):
These are very tedious to correct - thank you for going through them!

PR #14783, opened 2025-06-15
[Build] Fix more gradle deprecation warnings scheduled to be removed in 9.0

breskeby:
### Description

This fixes three more types of deprecations seen in the build:

1. Fix file permission setup using non deprecated Gradle API
2. Using Configuration#getFirstLevelModuleDependencies().findAll instead of deprecated  getFirstLevelModuleDependencies(Spec) 
3. Fix deprecated usage of other projects configurations
Resolution of the configuration :icu_current was done from a context different than the root project.
This is problematic and deprecated behavior scheduled to be removed in Gradle 9.0.

See https://docs.gradle.org/8.14/userguide/how_to_share_outputs_between_projects.html\#variant-aware-sharing for details how dependencies should be shared accross projects.

### DIFF
--- a/build-tools/build-infra/src/main/groovy/lucene.regenerate.icu.gradle
+++ b/build-tools/build-infra/src/main/groovy/lucene.regenerate.icu.gradle
@@ -30,14 +30,15 @@ def resources = rootProject.file("gradle/regenerate/icu")
  * download and compile a matching icu4c version automatically.
  */
 
-// Configure different icu4j dependencies.
-configure(rootProject) {
-  configurations {
+
+def setupIcuDependencies = { proj ->
+  // Configure different icu4j dependencies.
+  proj.configurations {
     icu_current
     groovy
   }
 
-  dependencies {
+  proj.dependencies {
     icu_current deps.icu4j
     // Use a newer groovy that doesn't have illegal reflective accesses.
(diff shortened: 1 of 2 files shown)


build-tools/build-infra/src/main/groovy/lucene.regenerate.icu.gradle
```diff
@@ -30,14 +30,15 @@ def resources = rootProject.file("gradle/regenerate/icu")
  * download and compile a matching icu4c version automatically.
  */
 
-// Configure different icu4j dependencies.
-configure(rootProject) {
```
breskeby, line 34: Declaring a configuration in one project and use it in another as done here before was one. of those things that had been technically possible forever, never recommended and now been deprecated. A good example of the problem with the historical approach in Gradle that in build logic there hasn't been any boundaries accross projects and the whole global model has been available from everywhere.

breskeby: IMO this plugin is a good example of something that can benefit from been ported into a binary plugin and only keep its actual implementation in this convention plugin rather all the nitty bits of implementation details.

dweiss (approved):

PR #14784, opened 2025-06-15
Make `pack` methods public for `BigIntegerPoint` and `HalfFloatPoint`

prudhvigodithi:
### Description

Following this commit https://github.com/apache/lucene/commit/94c76c790cd0b8b1bca384c5de18ec5c7957d5d9 make `pack` methods public for `BigIntegerPoint` and `HalfFloatPoint`.

Please let me know if there’s any restriction in sandbox packages that requires methods to be non-public.

Example (with IntPoint) with public `pack` we can directly use as follows
```
IntPoint.pack(new int[] { l }).bytes,
IntPoint.pack(new int[] { u }).bytes,
```

Without pack 
```
// Create byte array
byte[] lowerPoint = new byte[HalfFloatPoint.BYTES];
byte[] upperPoint = new byte[HalfFloatPoint.BYTES];
// Encode value into bytes
HalfFloatPoint.encodeDimension(l, lowerPoint, 0);
HalfFloatPoint.encodeDimension(u, upperPoint, 0);
```

Once this PR is merged we can:
```
HalfFloatPoint.pack(new int[] { l }).bytes,
HalfFloatPoint.pack(new int[] { u }).bytes,
```


prudhvigodithi:
I have just updated the CHANGES.txt adding to 10.2.2, please let me know if this is ok else I can change back to 10.3.0.

jpountz:
I'm good with the change, but I'd put the change in 10.3 instead of 10.2.2 since it's a new feature rather than a bug fix.

prudhvigodithi:
Thanks Adrien, updated the CHANGES.txt moving to 10.3.0.
@getsaurabh02

prudhvigodithi:
Just pushed a commit to fix the conflicts. @jpountz a gentle follow up to see if we are ok to merge this change.
Thanks

PR #14788, opened 2025-06-16
Remove duplicate -Xlint:options flags.

dweiss:
Fixes #14782 by not emitting duplicate lint flags to javac.

### DIFF
--- a/build-tools/build-infra/src/main/groovy/lucene.java.core.mrjar.gradle
+++ b/build-tools/build-infra/src/main/groovy/lucene.java.core.mrjar.gradle
@@ -32,9 +32,14 @@ configure(project(":lucene:core")) {
       tasks.named("compileMain${jdkVersion}Java").configure {
         def apijar = apijars.file("jdk${jdkVersion}.apijar")
 
+        // TODO: this depends on the order of argument configuration...
         int releaseIndex = options.compilerArgs.indexOf("--release")
         options.compilerArgs.removeAt(releaseIndex)
         options.compilerArgs.removeAt(releaseIndex)
+
+        // Remove conflicting options for the linter. #14782
+        options.compilerArgs.removeAll("-Xlint:options")
+
         options.compilerArgs += [
           "-Xlint:-options",
           "--add-exports",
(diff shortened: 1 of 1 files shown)


build-tools/build-infra/src/main/groovy/lucene.java.core.mrjar.gradle
```diff
@@ -32,9 +32,14 @@ configure(project(":lucene:core")) {
       tasks.named("compileMain${jdkVersion}Java").configure {
         def apijar = apijars.file("jdk${jdkVersion}.apijar")
 
+        // TODO: this depends on the order of argument configuration...
         int releaseIndex = options.compilerArgs.indexOf("--release")
         options.compilerArgs.removeAt(releaseIndex)
         options.compilerArgs.removeAt(releaseIndex)
+
+        // Remove conflicting options for the linter. #14782
+        options.compilerArgs.removeAll("-Xlint:options")
```
dweiss, line 41: @breskeby is there any nicer way to do this kind of compilerArgs manipulation that doesn't
depend on the ordering of those configuration callbacks? I can think of a custom argument provider but this would be even more confusing than what it is now, I think.

uschindler (approved):
If this works it is ok as a first step. I had same idea.

Indeed the problem with configuration order is hard to handle. Currently we rely on order of configuration.

Maybe @breskeby has an idea.

I will comment on JDK issue that there is kind of a regression because the order how the options are applied changed. Maybe they should have a defined order.

I was not able to test it yesterday in train, I just opened issue. My bad.

Greetings from bbuzz2025.

uschindler:
Can you backport this change?

dweiss:
I will apply and backport, no worries, Uwe. Enjoy a beer in Berlin!

uschindler:
If I remember: the order of configuration is still predictable by the order of including them into the main build script. I always felt bad with that, but it worked. Just want to confirm it also applies to changes in main.

dweiss:
Yes, I think this can be relied upon. I don't think it's mentioned anywhere, officially, but I think it's the case.

build-tools/build-infra/src/main/groovy/lucene.java.core.mrjar.gradle
breskeby: I don't have a better idea than you. I can think of a smarter  ArgumentProvider but no easy fix

dweiss: Thank you!

Issue #13158, opened 2024-03-05
Create a read-only index that drops index files not needed for searching 


msokolov:
### Description

Now that we have vector quantization we face the possibility of writing an index that is 5 times bigger than is needed for searching. If the index is primarily vectors and they are quantized, we will save the full-precision vectors but they may not be required at all for searching. In an architecture where indexes are written on one set of hosts and replicated to another set of hosts for searching, it is wasteful to copy all of the full-precision vectors to the searcher nodes. But Lucene doesn't have any way of distinguishing. I wonder if we could create a "write read-only index" operation that would effectively clone the existing index, dropping any data required only for indexing, and mark the index as read-only so it could never be opened for writing.  This might be useful in some way for version upgrades as well?

mikemccand:
This is a cool idea @msokolov.  It is wasteful to lug around those `float32` precision vectors out to the searchers in an NRT segment replication architecture.  In practice, they would consume disk space on the searchers, and waste time copying them out, but since the OS would never load them at search time, their bytes would remain cold on disk and not put much pressure on OS free RAM?  The OS would only cache the disk pages in the index that are actually needed at search time.  It would be nice not to copy all that deadweight around ...

Probably the solution would have to be something like segment to segment?  I.e. for each segment in the index, we would make a corresponding "read only" segment (stripped of the `float32` vectors).  This way, as the normal index changes (gets new flushed/merged segments), we could also incrementally/NRT maintain the shadow read-only index.

I wonder if there are other things in a Lucene index today that are needed only during indexing?

benwtrent:
Don't "deletes" require "writes"? Meaning, if enough docs get deleted in a segment, it will ultimately require to be merged, which then is a "write"?

For a quicker win in scalar quantization, it could be cheaper or easier to have a configured threshold where we throw away the floating point as we know the distribution of the values won't change significantly from that point on. Then on merges, if adjustments are required, we assume the cost of de-quantizing and re-quantizing. This can help relevancy more than you would expect, but obviously not as much as having access to the raw values.

msokolov:
> Don't "deletes" require "writes"? Meaning, if enough docs get deleted in a segment, it will ultimately require to be merged, which then is a "write"?

The goal of segment-replication is to completely separate searching from writing, so in that world, no merging is done by searchers -- it happens upstream in a writer/indexer, or perhaps in a dedicated merger (we have both setups going on).

benwtrent:
[...]

PR #14792, opened 2025-06-16
Introduced OffHeapQuantizedFloatVectorValues class to access float vectors when only quantized byte vectors are available in the index

Pulkitg64:
### Description

Introduce `getQuantizedVectorValues` method in `LeafReader` to access `QuantizedVectorValues`.

In a search architecture where searchers and writer runs on separate machine, it is wasteful to have raw float vectors on machine when vector quantization enabled. This PR is adding getQuantizedVectorValues in LeafReader which will help to read QuantizedByteVectors directly without need of reading raw float vectors.

Partially solving #13158

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene95/OffHeapFloatVectorValues.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene95/OffHeapFloatVectorValues.java
@@ -270,8 +270,7 @@ public abstract class OffHeapFloatVectorValues extends FloatVectorValues impleme
     }
   }
 
-  /** Empty vector values when there are no vectors stored. */
-  public static class EmptyOffHeapVectorValues extends OffHeapFloatVectorValues {
+  private static class EmptyOffHeapVectorValues extends OffHeapFloatVectorValues {
 
     public EmptyOffHeapVectorValues(
         int dimension,
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene95/OffHeapFloatVectorValues.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene95/OffHeapFloatVectorValues.java
@@ -270,7 +270,8 @@ public abstract class OffHeapFloatVectorValues extends FloatVectorValues impleme
     }
   }
 
-  private static class EmptyOffHeapVectorValues extends OffHeapFloatVectorValues {
+  /** Empty vector values when there are no vectors stored. */
+  public static class EmptyOffHeapVectorValues extends OffHeapFloatVectorValues {
 
     public EmptyOffHeapVectorValues(
         int dimension,
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99ScalarQuantizedVectorsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99ScalarQuantizedVectorsReader.java
@@ -27,7 +27,6 @@ import org.apache.lucene.codecs.CodecUtil;
 import org.apache.lucene.codecs.KnnVectorsReader;
 import org.apache.lucene.codecs.hnsw.FlatVectorsReader;
 import org.apache.lucene.codecs.hnsw.FlatVectorsScorer;
-import org.apache.lucene.codecs.lucene95.OffHeapFloatVectorValues;
 import org.apache.lucene.codecs.lucene95.OrdToDocDISIReaderConfiguration;
 import org.apache.lucene.index.ByteVectorValues;
 import org.apache.lucene.index.CorruptIndexException;
@@ -190,8 +189,7 @@ public final class Lucene99ScalarQuantizedVectorsReader extends FlatVectorsReade
   public FloatVectorValues getFloatVectorValues(String field) throws IOException {
     final FieldEntry fieldEntry = getFieldEntry(field);
     final FloatVectorValues rawVectorValues = rawVectorsReader.getFloatVectorValues(field);
-    if (rawVectorValues instanceof OffHeapFloatVectorValues.EmptyOffHeapVectorValues
-        && fieldEntry.size != 0) {
+    if (rawVectorValues.size() == 0) {
       return OffHeapQuantizedFloatVectorValues.load(
           fieldEntry.ordToDoc,
           fieldEntry.dimension,
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99ScalarQuantizedVectorsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99ScalarQuantizedVectorsReader.java
@@ -27,6 +27,7 @@ import org.apache.lucene.codecs.CodecUtil;
 import org.apache.lucene.codecs.KnnVectorsReader;
 import org.apache.lucene.codecs.hnsw.FlatVectorsReader;
 import org.apache.lucene.codecs.hnsw.FlatVectorsScorer;
+import org.apache.lucene.codecs.lucene95.OffHeapFloatVectorValues;
 import org.apache.lucene.codecs.lucene95.OrdToDocDISIReaderConfiguration;
 import org.apache.lucene.index.ByteVectorValues;
 import org.apache.lucene.index.CorruptIndexException;
@@ -189,6 +190,21 @@ public final class Lucene99ScalarQuantizedVectorsReader extends FlatVectorsReade
   public FloatVectorValues getFloatVectorValues(String field) throws IOException {
     final FieldEntry fieldEntry = getFieldEntry(field);
     final FloatVectorValues rawVectorValues = rawVectorsReader.getFloatVectorValues(field);
+    if (rawVectorValues instanceof OffHeapFloatVectorValues.EmptyOffHeapVectorValues
+        && fieldEntry.size != 0) {
+      return OffHeapQuantizedFloatVectorValues.load(
+          fieldEntry.ordToDoc,
+          fieldEntry.dimension,
+          fieldEntry.size,
+          fieldEntry.scalarQuantizer,
+          fieldEntry.similarityFunction,
+          vectorScorer,
+          fieldEntry.compress,
+          fieldEntry.vectorDataOffset,
+          fieldEntry.vectorDataLength,
+          quantizedVectorData);
+    }
+
     OffHeapQuantizedByteVectorValues quantizedByteVectorValues =
         OffHeapQuantizedByteVectorValues.load(
             fieldEntry.ordToDoc,
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene99/OffHeapQuantizedFloatVectorValues.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene99/OffHeapQuantizedFloatVectorValues.java
@@ -1,0 +1,365 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.lucene.codecs.lucene99;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import org.apache.lucene.codecs.hnsw.FlatVectorsScorer;
+import org.apache.lucene.codecs.lucene90.IndexedDISI;
+import org.apache.lucene.codecs.lucene95.HasIndexSlice;
+import org.apache.lucene.codecs.lucene95.OrdToDocDISIReaderConfiguration;
+import org.apache.lucene.index.FloatVectorValues;
+import org.apache.lucene.index.VectorSimilarityFunction;
+import org.apache.lucene.search.DocIdSetIterator;
+import org.apache.lucene.search.VectorScorer;
+import org.apache.lucene.store.IndexInput;
+import org.apache.lucene.util.Bits;
+import org.apache.lucene.util.hnsw.RandomVectorScorer;
+import org.apache.lucene.util.packed.DirectMonotonicReader;
+import org.apache.lucene.util.quantization.ScalarQuantizer;
+
+/** Read the vector values from the index input. This supports both iterated and random access. */
+public abstract class OffHeapQuantizedFloatVectorValues extends FloatVectorValues
+    implements HasIndexSlice {
+
+  protected final int dimension;
+  protected final int size;
+  protected final int numBytes;
+  protected final ScalarQuantizer scalarQuantizer;
+  protected final VectorSimilarityFunction similarityFunction;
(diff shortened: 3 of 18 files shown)


benwtrent:
@Pulkitg64 I don't understand how this is part of https://github.com/apache/lucene/issues/13158

I would have thought the APIs stay the same. Quantization should be able to "rehydrate" the quantized vectors into floating point (or whatever the original values).

So, the segment, depending on what data it has access to, will:

 - Return the original doc value floating point vectors
 - Rehydrate the quantized values.

Either way, users should still be able to call `float[] vectorValue(int ord)`. 

I would think there is a sub-class called `QuantizedFloatVectorValues`, that satisfies the `FloatVectorValues` interface. 

But maybe we add an `isApproximate()` or a `extractQuantizedValues()` that returns null, or the `QuantizedFloatVectorValues` interface.

But it is likely useless for the user to have access to the quantized bytes directly as they don't provide much value without knowing how to use them.

Pulkitg64:
Thanks @benwtrent for the quick review and comments.

Regarding your comment about how this relates to issue #13158 - I agree in a way this PR doesn't directly help create a "read-only" index as mentioned in the issue. Let me clarify the motivation:

This PR addresses a scenario where:
* Raw (unquantized) vectors are removed from the index since they aren't needed for searching
* The architecture has searcher and writer running on separate machines

Currently, there's no way to directly access quantized vectors - we can only access raw vectors. But if raw vectors are dropped, this causes errors. This PR adds methods to access ByteQuantizedVectors in such cases.

As for the usefulness of accessing quantized bytes directly - we have specific use cases, such as returning the vectors themselves when requested in a query.

Please let me know your thoughts.

Regarding accessing quantized vectors directly - we could also consider using the [QuantizedVectorValues](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99ScalarQuantizedVectorsReader.java#L408) class, which is currently returned by the [getFloatVectorValues](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99ScalarQuantizedVectorsReader.java#L189) method. While this class wraps both raw and quantized vectors, its members are private, preventing direct access to the quantized vectors like we're doing in this PR.

Would it make more sense to make the relevant members public in QuantizedVectorValues rather than adding getQuantizedVectorValues to LeafReader?

benwtrent:
> As for the usefulness of accessing quantized bytes directly - we have specific use cases, such as returning the vectors themselves when requested in a query.

I would assume the caller would want something akin to the `float` values. What would a caller be expected to do with the quantized bytes directly?


I am saying that returning the quantized bytes, without knowing all the other information (quantized technique, the technique's parameters, etc.) is pretty useless.

Pulkitg64:
We are experimenting with large vector indexes, and since (raw unquantized) vectors consume significant disk space (4x more than quantized vectors), we want to drop the raw vectors from searcher machines. We are currently using vector values for below use cases:
1. Calculating the dot-product scores and return them in search results
2. Returning the vectors in search results
3. Vector counting for metrics

For use case 1 we have started to use [vectorScorer](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99ScalarQuantizedVectorsReader.java#L444) which use quantized vectors for computing score so we are good there. For use cases 2 and 3, we currently use floatVectorValues using [getFloatVectorValues](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99ScalarQuantizedVectorsReader.java#L189) but need to switch to quantizedVectorValues since searchers won't have float vectors anymore and we are okay in accepting the accuracy loss from float-to-byte quantization.

To address these use cases, we have two options:
* Introduce a new API: getQuantizedVectorValues to access quantizedByteVector OR
* Use our local workaround: Make the [QuantizedVectorValues](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99ScalarQuantizedVectorsReader.java#L408) class and its members public to directly access quantized vectors

I would like to know your thoughts on whether we should create such an API, and if you think the above use cases don't justify a new API, what are your thoughts on implementing the workaround solution and pushing it upstream?

benwtrent:
> Returning the vectors in search results

Why would you need to do this?

Generally, I would assume that any access to the vector would be "Give me what I gave you", and the best we can do with quantized vectors is the dequantized vector. 

I don't fully understand how serializing a read-only segment that is missing files (e.g. missing the "vec" file), but the format should do the right thing and see that the file isn't there and provide an approximate view of the floating point vectors.

> Vector counting for metrics

I don't understand what this means really. Just counting how many vectors there are? This should be doable via the `FloatVectorValues` interface.

> but need to switch to quantizedVectorValues since searchers won't have float vectors anymore and we are okay in accepting the accuracy loss from float-to-byte quantization.

Again, I think we should do the nice thing, de-quantize the vectors as the user asks for them. 

It should fully satisfy the `FloatVectorValues` API, de-quantizing the vectors and indicate that the vector returned is an approximation.

Getting access to the raw quantized bytes is basically useless without all the other parameters  that were used to quantized the vector.

benwtrent:
@Pulkitg64 

Basically, I don't think callers should know if they are hitting quantized vectors or raw. Or at least have to make that decision up front.

Requiring the user to pick the right thing seems unnecessary when we have the appropriate interfaces already. Its just all about determining how the format itself knows that its missing the `vec` file.

msokolov:
For the return values use case, another choice is to disable it in the case the original vectors were not "stored" in the searchable index. Otherwise, I agree with Ben that we could support "rehydration" in the codec. For example, suppose we see that we have zero full-precision vectors, but nonzero quantized vectors; then we could fall back to "rehydration". 

For the counting case (get total number of vectors), should we always use the quantized count where today we use the full-precision count?

Pulkitg64:
Thanks @benwtrent @msokolov 

> Again, I think we should do the nice thing, de-quantize the vectors as the user asks for them.

Sorry, I am surely missing something. But if we can't access the quantized vector then how can we do this? We still need byte vector to rehydrate float vectors, right?

benwtrent:
@Pulkitg64 I think the format has access to the quantized vectors. I am saying we shouldn't add a new `getQuantizedVEctors` to the leaf or kNN APIs. 

When the `vec` file isn't present, the format should still return a `FloatVectorValues`, and when folks call `float[] getVector(int ord)`, they get the de-quantized representation of the vector.

This should require no new public facing APIs anywhere.

Here is the idea.

```
public FloatVectorValues getFloatVectorValues(String field) throws IOException {
    return OffHeapQuantizedFloatVectorValues.load(...);
}

...

static class OffHeapQuantizedFloatVectorValues extends FloatVectorValues {
  
  final float[] scratch;
  int curOrd = -1;

  ...

  public  float[] vectorValue(int ord) throws IOException {
    if (ord == curOrd) return scratch;
    dequantize(ord);
    curOrd = ord;
    return scratch;
  }

  private void dequantize(int ord) throws IOException {
    //read quantized values and parameters
    // dequantize into scratch
  }
  
}

```

Pulkitg64:
Oh I understand now what you meant in your comments. This approach is much more cleaner and doesn't require any new API addition. Will raise a new revision in some time.

Pulkitg64:
I have made the new revision to introduce a new class ```OffHeapQuantizedFloatVectorValues``` which will be returned during `getFloatVectorValues()` call and when there are no raw vectors present but quantized vectors still present in the index.

lucene/core/src/java/org/apache/lucene/codecs/lucene95/OffHeapFloatVectorValues.java
```diff
@@ -270,7 +270,8 @@ public DocIdSetIterator iterator() {
     }
   }
 
-  private static class EmptyOffHeapVectorValues extends OffHeapFloatVectorValues {
+  /** Empty vector values when there are no vectors stored. */
```
msokolov: "values when" should read "values to be returned when"

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99ScalarQuantizedVectorsReader.java
```diff
@@ -189,6 +190,21 @@ private FieldEntry getFieldEntry(String field) {
   public FloatVectorValues getFloatVectorValues(String field) throws IOException {
     final FieldEntry fieldEntry = getFieldEntry(field);
     final FloatVectorValues rawVectorValues = rawVectorsReader.getFloatVectorValues(field);
+    if (rawVectorValues instanceof OffHeapFloatVectorValues.EmptyOffHeapVectorValues
```
msokolov: I think I'd like this better if we checked `rawVectorValues.size() == 0` instead of having an `instanceof` check.  We might change the class, or replace it with an anonymous class or something, but we will always have size=0.

msokolov (commented):
Could you say where the implementations came from ... was it copied from OffHeapFloatVectorValues and then changed?

Also: this is kind of interesting because it provides the ability to read an index that we don't actually provide any way to create! Maybe add some comments explaining how it is intended to be used -- ie like the `simulate` method in the test?

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99ScalarQuantizedVectorsReader.java
benwtrent: I agree with @msokolov here. Checking size simplifies things.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/OffHeapQuantizedFloatVectorValues.java
```diff
@@ -0,0 +1,365 @@
...
+    byteBuffer = ByteBuffer.allocate(dimension);
+    binaryValue = byteBuffer.array();
+    floatValue = new float[dimension];
+    this.similarityFunction = similarityFunction;
+    this.vectorsScorer = vectorsScorer;
+  }
+
+  public ScalarQuantizer getScalarQuantizer() {
+    return scalarQuantizer;
+  }
```
benwtrent: this isn't strictly necessary

lucene/core/src/java/org/apache/lucene/codecs/lucene99/OffHeapQuantizedFloatVectorValues.java
```diff
@@ -0,0 +1,365 @@
...
+import org.apache.lucene.search.DocIdSetIterator;
+import org.apache.lucene.search.VectorScorer;
+import org.apache.lucene.store.IndexInput;
+import org.apache.lucene.util.Bits;
+import org.apache.lucene.util.hnsw.RandomVectorScorer;
+import org.apache.lucene.util.packed.DirectMonotonicReader;
+import org.apache.lucene.util.quantization.ScalarQuantizer;
+
+/** Read the vector values from the index input. This supports both iterated and random access. */
+public abstract class OffHeapQuantizedFloatVectorValues extends FloatVectorValues
```
benwtrent: If this is gonna be public, it should likely have a `internal` annotation to make folks aware that its interface can break whenever we want.

lucene/core/src/java/org/apache/lucene/codecs/lucene95/OffHeapFloatVectorValues.java
```diff
@@ -270,7 +270,8 @@ public DocIdSetIterator iterator() {
     }
   }
 
-  private static class EmptyOffHeapVectorValues extends OffHeapFloatVectorValues {
```
benwtrent: if switching to `size() == 0` for float vector existence, I think this change is unnecessary.

benwtrent (commented):
thanks @Pulkitg64, this is pretty cool :)

lucene/core/src/java/org/apache/lucene/codecs/lucene95/OffHeapFloatVectorValues.java
Pulkitg64: Removed this change, since we will check size==0 instead of making the class public.

lucene/core/src/java/org/apache/lucene/codecs/lucene95/OffHeapFloatVectorValues.java
Pulkitg64: Removed this change.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/OffHeapQuantizedFloatVectorValues.java
Pulkitg64: Removed from new revision.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99ScalarQuantizedVectorsReader.java
Pulkitg64: Yes, that simplifies things. Fixed in new revision.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/OffHeapQuantizedFloatVectorValues.java
Pulkitg64: Make sense.

Pulkitg64:
> Could you say where the implementations came from ... was it copied from OffHeapFloatVectorValues and then changed?

The implementation is coming from OffHeapQuantizedByteVectorValues with change in vectorValue() function which return float vectors after byte vector is dequantized

> Also: this is kind of interesting because it provides the ability to read an index that we don't actually provide any way to create! Maybe add some comments explaining how it is intended to be used -- ie like the `simulate` method in the test?

Sure, added in next revision

lucene/core/src/java/org/apache/lucene/codecs/lucene99/OffHeapQuantizedFloatVectorValues.java
msokolov: does it need to be public though? I think callers will merely need to access the public methods on `FloatVectorValues`.

msokolov (commented):
Do we need to change the title of the PR? If only so we will have a better commit message. I think if we do that, and add a usage comment on the class (something like -- full precision vectors can be removed from an index using a method like shown in the tests, but after this, the index can no longer safely be updated and should be treated as a read-only index.), and I think make the new class private or package-private if needed in a test then it's ready

lucene/core/src/java/org/apache/lucene/codecs/lucene99/OffHeapQuantizedFloatVectorValues.java
Pulkitg64: Yes, you are right, we can make it package-private. Changing this in next revision.

msokolov:
Thanks for addressing the feedback; I'll push this now. FYI - we generally want to avoid force pushes since github loses track of the changes then.  Usually it's possible to do so by using merge commits, which will get squashed away when we eventually squash and merge to main branch.

Pulkitg64:
Thanks Mike, Ben for all your feedback and comments. I will keep these feedbacks in mind for my next PRs

Pulkitg64:
Should we backport this PR to 10.x branch? I am trying to backport this change about [enforcing fallback support for float vector retrieval in quantized KNN vector formats](https://github.com/apache/lucene/pull/15476) but the test case is failing there because the fallback support is not present in the Lucene99ScalarQuantizedVectorsReader.java on 10x branch.

Pulkitg64:
Hey @benwtrent @msokolov, is it good idea to back port this change to 10.x branch regarding [this](https://github.com/apache/lucene/pull/14792#issuecomment-3745277101) comment?

benwtrent:
@Pulkitg64 I would say backport & override/mute for the unsupported old format.

mikemccand:
Yeah +1 to backport this original PR (produce de-quantized vectors if full-precision vector isn't present but quantized vector is), and also https://github.com/apache/lucene/pull/15476 (the test infra to confirm future KNN Codecs on 10.x also preserve this capability).

PR #14797, opened 2025-06-16
Adjust base knn format assert assertOffHeapByteSize

benwtrent:
assertOffHeapByteSize makes a ton of assumptions and these don't really work well for custom formats. 

Changing to `protected` instead of `static` to allow simple overriding by custom format testers.


ChrisHegarty (approved):
LGTM.

Issue #14786, opened 2025-06-16
build and push release regression


dweiss:
### Description

Something stopped working here -
https://github.com/apache/lucene/actions/runs/15670563859

This depends on #14796

PR #14798, opened 2025-06-16
Correct python release scripts for the new location of base version

dweiss:
#14786

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ gradle/wrapper/gradle-wrapper.jar
 # Python
 .pydevproject
 __pycache__
+.venv
 
 # Emacs backup
 *~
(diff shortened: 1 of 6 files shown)


rmuir:
Try a `make reformat`. Not sure what editor you use, but eg for vscode if you can install ruff and basedpyright extensions, and enable format-on-save and organize-imports-on-save too, you will pretty much never need to deal with `make`. Lmk, I can try to document or add eg vscode settings file or whatever helps.

dweiss:
Thanks. I think I'll go back to the regexp-scan instead of using the javaproperties module. I don't think we import the requirements in github workflow that runs the tests... Will return to it tomorrow, I'm done for today.

rmuir:
For now you can try `make autofix` which is like `make reformat` but will also (safely) fix any linter issues it can too.

rmuir:
#14799 to make the messaging better when `make lint` fails.

rmuir (approved):

PR #14800, opened 2025-06-17
Fix assemble source release

dweiss:
Fixes #14796


dweiss:
I'll let myself merge this in to proceed with other things. If anybody would like to change anything here, please comment on the patch and I'll follow-up.

PR #14801, opened 2025-06-17
Remove all security manager and java security references

dweiss:
these are no-ops in JDK24+.


build-tools/build-infra/src/main/groovy/lucene.validation.ecj-lint.gradle
```diff
@@ -74,6 +76,8 @@ def lintTasks = sourceSets.collect { SourceSet sourceSet ->
     dependsOn sourceSet.compileClasspath
     dependsOn ecjConfiguration
 
+    mustRunAfter tasks.withType(SpotlessApply)
```
dweiss, line 79: piggybacking this so that 'gradlew tidy check' works properly...

PR #14804, opened 2025-06-17
detect and ban wildcard imports in Java

### LINKED ISSUE #14553: Fail spotless check for wildcard imports
### Description

Currently, several files in the codebase contain wildcard imports ([example](https://github.com/apache/lucene/blob/main/lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/VectorUtilBenchmark.java#L22)). 

Wildcard imports can:
- Make it harder to track dependencies
- Lead to naming conflicts
- Reduce code readability
- Make it difficult to identify unused imports

OpenSearch has already implemented a check to prevent wildcard imports ([see](https://github.com/opensearch-project/OpenSearch/pull/3939)). We can adopt a similar mechanism for Lucene to enforce explicit imports.

Proposed changes:
Add a spotless rule to detect wildcard imports to avoid committing wildcard imports. Replace existing wildcard with explicit imports.

dweiss:
Fixes #14553.

I'm not completely happy with this. For some reason, the custom formatting step always triggers full spotless run - incremental mode doesn't work.

```
> ./gradlew -p lucene/grouping/ spotlessCheck  --info
> ./gradlew -p lucene/grouping/ spotlessCheck  --info
...
> Task :lucene:grouping:spotlessJava
Caching disabled for task ':lucene:grouping:spotlessJava' because:
  Build cache is disabled
Task ':lucene:grouping:spotlessJava' is not up-to-date because:
  Value of input property 'stepsInternalEquality' has changed for task ':lucene:grouping:spotlessJava'
The input changes require a full rebuild for incremental task ':lucene:grouping:spotlessJava'.
Not incremental: removing prior outputs
Resolve mutations for :lucene:grouping:spotlessJavaCheck (Thread[#9108,Execution worker Thread 4,5,main]) started.
:lucene:grouping:spotlessJavaCheck (Thread[#9108,Execution worker Thread 4,5,main]) started.
```

The detection is also costly (regexp over the entire codebase); this could be probably simplified to line-by-line scanning and a heuristic to short-circuit early when import statements are no longer possible...


rmuir:
Can we consider ast-grep for this? it is really fast and doesn't require regular expressions, has plugins for editors. I wrote a rule for this in less than a minute:
```yaml
id: wildcard-import-not-allowed
language: java
rule:
  kind: asterisk
  inside:
    kind: import_declaration
severity: error
message: don't use wildcard imports
note: please use full import instead
url: https://whatever/explanation
```

dweiss:
I'll take a look, sure. I'll also try to figure out why the heck spotless isn't behaving the way it should.

dweiss:
The reason it's never up to date with a custom formatting step:
https://github.com/diffplug/spotless/issues/2516

dweiss:
Error prone has a rule for this, actually -
```
diff --git a/build-tools/build-infra/src/main/groovy/lucene.validation.error-prone.gradle b/build-tools/build-infra/src/main/groovy/lucene.validation.error-prone.gradle
index 49a4a067af6..1070346a19f 100644
--- a/build-tools/build-infra/src/main/groovy/lucene.validation.error-prone.gradle
+++ b/build-tools/build-infra/src/main/groovy/lucene.validation.error-prone.gradle
@@ -331,6 +331,7 @@ allprojects { prj ->
           '-Xep:UnsafeWildcard:ERROR',
           '-Xep:UnusedAnonymousClass:ERROR',
           '-Xep:UnusedCollectionModifiedInPlace:ERROR',
+          "-Xep:WildcardImport:ERROR",
           '-Xep:VarTypeName:ERROR',
           // '-Xep:WrongOneof:OFF', // we don't use protobuf
           '-Xep:XorPower:ERROR',
```
I know we only run it on a CI but ast-grep requires an external tool, so maybe it's a simpler/ better alternative? This is what the output looks like:

```
/home/dweiss/mounts/ocean/work/apache/lucene/lucene/core/src/java/org/apache/lucene/internal/hppc/IntIntHashMap.java:20: error: [WildcardImport] Wildcard imports, static or otherwise, should not be used
import static org.apache.lucene.internal.hppc.HashContainers.*;
^
    (see https://google.github.io/styleguide/javaguide.html?cl=head#s3.3.1-wildcard-imports)
```

rmuir:
yeah, that's fine.

we can discuss it separately maybe. Elsewhere I use the tool a lot, replace entire slow linters with it! It is wicked fast...

there's a strange cultural thing at play here, with all the other programming languages, "external tools" written in rust are heavily adopted, makes everything so fast on the dev side. It isn't just python here but e.g. microsoft rewriting typescript in golang, bun, etc.

with java, and only java, there's this strange NIH cultural thing where all tooling must itself also be written in java. I don't understand it, but it makes everything really slow :)

The recent "git" is a great example, insanely fast tool designed by literally, the best, yet there's pressure to use a crappy java substitute.

dweiss:
I have noticed this as well - in the js world most (fast) things use a rust backend.

Anyway, I'll fiddle with this a bit, it's not a high priority thing. Maybe make it optionally available, if the tool is present. I'll see.

dweiss:
It is wicked fast indeed (uses all cores).
```
$ time ast-grep scan  . -r /home/dweiss/tmp/no-wildcard-imports.sg.yaml
...
real	0m0.738s
user	0m6.995s
sys	0m0.269s
```

rmuir:
In my builds, I install such tools via `pip` or `npm` package managers. So it's no different than downloading jar and using that, just faster. Just food for thought.

rmuir:
The python linting, formatting, type checking in dev tools is an example. None of those are written in the python programming language.

dweiss:
I've added a task that runs ast-grep on the CI and locally, if you define ```lucene.tool.ast-grep``` property (for example, ```lucene.tool.ast-grep=ast-grep``` or permanently, in local build options or gradle properties).

image

Opens up the possibility to add new rules too (gradle/validation/ast-grep).

dweiss:
I'll try to finalize this patch later today, have to leave now.

rmuir:
Thanks for playing with it. I would be happy to try to write ast-grep rules to try to replace some of the slower error-prone checks, or whatever else (nocommit checks etc).

We could install the package via pip or npm in the GH actions to support the checker. At just a glance, many of the rules are basically trivial e.g. https://errorprone.info/bugpattern/EqualsNull. 

From what I remember, each check has a high overhead for errorprone which I don't believe is the case for ast-grep. I know I'm oversimplifying it, but I assume each file is parsed only once and then all rules run similar to https://tree-sitter.github.io/tree-sitter/using-parsers/queries/1-syntax.html, but without the s-expressions :)

dweiss:
Well, if this patch goes in, all it takes to add a rule is to create a new file in gradle/validation/ast-grep - it should be picked up automatically. We can also make ast-grep a requirement (much like python, perl and git) in the long term if it proves to be beneficial.

I've taken a look at the tree sitter too, very interesting stuff. Gotta love how with all the cool languages out there, it's still C that prevails. :)

rmuir:
Another related idea here is to prevent any IDE from automatically adding the *'s we don't want.

Eclipse's default limit is 99 imports before it starts doing this... of course user can change, and some rare files might exceed this. there's a way to bump it to huge value (e.g. Integer.MAX_VALUE) in our eclipse settings, i think, but i don't have a concrete solution: https://github.com/eclipse-jdtls/eclipse.jdt.ls/blob/0b3651f030e2e68e43019268f35880a07aa7a8c5/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/preferences/Preferences.java#L345-L367

I don't know how intellij behaves here or if we can prevent similar behavior. All this stuff can be followups, really, I'm just brainstorming.

Thanks for the ast-grep though!!! Gives me a way to help speed up the checks :) I will leave a few comments, not review ones intended to block the change. I can help out with followup PRs.

gradle/validation/ast-grep/sgconfig.yml
```diff
@@ -0,0 +1,2 @@
+ruleDirs:
+  - ./rules
```
rmuir: I've found the built-in self-test to be essential for having and keeping correct rules. It takes milliseconds and they are easy to write.

```suggestion
  - ./rules
testConfigs:
- testDir: ./tests
```

I run these with `ast-grep test --skip-snapshot-tests` before the `ast-grep scan`. 

For a rule like this one, I would add tests (untested!) like this: `tests/errorprone.yml` (which would test rules in `rules/errorprone.yml`. I organize them this way with `---` yaml doc separator between the rules in both files. For multiline tests, I use yaml literal style (`|`): https://yaml.org/spec/1.2-old/spec.html#id2795688

```yaml
---
# test wildcard import detector
id: wildcard-import-not-allowed
valid:
  - import foo.bar.Baz;
  - import static foo.bar.Baz;
invalid:
  - import foo.bar.*;
  - import static foo.bar.*;
---
# test for another error prone rule
id: some-other-errorprone-rule
invalid: 
  - |
    if (x.equals(null) {
    }
  - |
    foo = x.equals(null) ? 1 : 2;
# ...
```

dweiss:
I've forked the ast-grep thing into a separate PR - https://github.com/apache/lucene/pull/14808. I think we can merge that one in, then I can apply a smaller patch for wildcards, using your rule.

gradle/validation/ast-grep/sgconfig.yml
dweiss: > I run these with ast-grep test --skip-snapshot-tests before the ast-grep scan.

We can integrate this check into gradle as well. These are nice indeed - I've seen the docs on the tool.

gradle/validation/ast-grep/rules/no-wildcard-imports.sg.yaml
```diff
@@ -0,0 +1,15 @@
+id: wildcard-import-not-allowed
```
rmuir, line 2:
```suggestion
# yaml-language-server: $schema=https://raw.githubusercontent.com/ast-grep/ast-grep/refs/heads/main/schemas/java_rule.json
id: wildcard-import-not-allowed
```

I add the comment above each rule, users with https://github.com/redhat-developer/yaml-language-server get autocomplete / validation / hover documentation in editors.

msfroh:
> I don't know how intellij behaves here or if we can prevent similar behavior. All this stuff can be followups, really, I'm just brainstorming.

We should be able to make it easier for IntelliJ to do the right thing without too much trouble. One option is to use `gradle-idea-ext`, which is what we do on OpenSearch ([ref](https://github.com/opensearch-project/OpenSearch/blob/6ff44d9b5e1f1f6f22b9935d21e408957383e940/gradle/ide.gradle#L67)). Another option is to use a `.editorconfig` file, where there are some IntelliJ-specific extensions (like `ij_java_names_count_to_use_import_on_demand`) that can crank up the threshold for wildcard imports.

I can take a shot at a separate PR to help out there.

dweiss:
> We should be able to make it easier for IntelliJ to do the right thing without too much trouble. One option is to use `gradle-idea-ext`, 

I've never used this plugin - have it on my "to-check" list (!). It would be awesome to provide more reasonable IntelliJ defaults. My experience with using google-java-format plugin is that once you enable it, the import section formatting settings are completely ignored...

gradle/validation/ast-grep/sgconfig.yml
dweiss: I've added it, nice.

dweiss:
Ok. I think this one is ready.

gradle/validation/ast-grep/rules/no-wildcard-imports.sg.yaml
```diff
@@ -0,0 +1,16 @@
...
+language: java
+rule:
+  kind: asterisk
+  inside:
+    kind: import_declaration
+severity: error
+message: don't use wildcard imports
+note: please use full imports instead
+url: https://google.github.io/styleguide/javaguide.html?cl=head#s3.3.1-wildcard-imports
```
rmuir: I can see pointing at the style guide, but as far as being practical, i wonder if we can name the file errorprone.yml and name the checks accordingly (e.g. `WildcardImport` and link to their https://errorprone.info/bugpattern/WildcardImport url. 

It is just how i've been organizing them, after seeing others use similar approach. I also think the errorprone URL is more helpful to end-user in fixing such issues (specs are more difficult to read), but that's just me.

dweiss: Up to you, entirely. I just pasted... something. :)

dweiss: Done.

rmuir: we can always change it. that particular description is really cool too, because of how meta this PR is. By banning these wildcards, we can do more with these simple rules :)

Just as an example, you could write unused import rule which would work correctly (if you wanted to), whereas with wildcards its not reasonable.

dweiss: Oh, if you'd like to rename the rules - feel free to just commit whatever it is your gut tells you. I think I'm done here - once you think it's ready, just squash to main, please. Thanks!

rmuir (approved):

gradle/validation/ast-grep/rules/no-wildcard-imports.sg.yaml
rmuir: let's just commit yours, i'll try to see if i can find some easy-wins in the errorprone rule list and organize while I'm there.

dweiss: Thanks for coming up with this great idea - it seems like it opens up a lot of new possibilities. Definitely better than the custom formatter step I initially added.

dweiss:
FYI - spotless runs google java format in a single-thread, that's part of the reason it's sluggish. But it's still an order of magnitude slower than the c parser. I ran this, for the fun of it (this uses multiple threads for parsing):
```
> find . -name "*.java" > files.in
> wc -l files.in
5973 files.in
> time java -jar ~/tmp/google-java-format-1.27.0-all-deps.jar  -n --skip-reflowing-long-strings --skip-javadoc-formatting --skip-removing-unused-imports @files.in
real	0m5.338s
user	2m7.927s
sys	0m6.549s
```
The user time - compared to tree-sitter's 7 seconds - is rather pale.

rmuir:
@dweiss yeah, the parsers are generally fast. and the java one in particular is very nice. Of course they are doing a lot less, too. 

parser test framework tracks the performance in terms of MB/s and warns if the speed gets too low:
```
  custom_tag:
      5. ✓ custom tag
      6. ✓ custom tag: inline
      7. ✓ custom tag: missing description
      8. ✓ custom tag: inline, missing description -- Warning: Slow parse rate (1671.346 bytes/ms)
```

owaiskazi19:
Spotless now supports removing wildcard https://github.com/diffplug/spotless/pull/2517 from 7.1.0. We can use the same

rmuir:
@owaiskazi19 thank you for the heads up. I opened https://github.com/apache/lucene/issues/14960 for this.

I saw the spotless 7.1 PR go through but unfortunately nothing in the release notes in the dependabot PR about a new wildcard-remover feature: https://github.com/apache/lucene/pull/14956

dweiss:
Except we're now using custom code to call into google java format... Eh.

dweiss:
Hmm... looking at the diff, the added step just removes wildcard imports - it doesn't resolve them (like intellij or eclipse would)? I don't know if this is that helpful.

rmuir:
Yeah, just removing them isn't "fixing" :)

PR #14806, opened 2025-06-18
Utilize `docIdRunEnd` on `ReqExclBulkScorer`

HUSTERGS:
### Description
This PR propose to utilize `docIdRunEnd` on `ReqExclBulkScorer`, so we can jump faster on `MUST_NOT` clause


lucene/core/src/java/org/apache/lucene/search/ReqExclBulkScorer.java
```diff
@@ -57,7 +57,10 @@ public int score(LeafCollector collector, Bits acceptDocs, int min, int max) thr
         exclDoc = exclApproximation.advance(upTo);
       }
       if (exclDoc == upTo) {
-        if (exclTwoPhase == null || exclTwoPhase.matches()) {
+        if (exclTwoPhase == null) {
+          // from upTo to docIdRunEnd() are excluded, so we scored up to docIdRunEnd()
+          upTo = exclApproximation.docIDRunEnd();
+        } else if (exclTwoPhase.matches()) {
```
gf2121, line 63: Is `upTo = Math.max(upTo + 1, exclTwoPhase.docIdRunEnd())` correct here?

lucene/core/src/java/org/apache/lucene/search/ReqExclBulkScorer.java
```diff
@@ -57,7 +57,10 @@ public int score(LeafCollector collector, Bits acceptDocs, int min, int max) thr
         exclDoc = exclApproximation.advance(upTo);
       }
       if (exclDoc == upTo) {
-        if (exclTwoPhase == null || exclTwoPhase.matches()) {
+        if (exclTwoPhase == null) {
+          // from upTo to docIdRunEnd() are excluded, so we scored up to docIdRunEnd()
+          upTo = exclApproximation.docIDRunEnd();
```
gf2121, line 62: Do we need to check `exclApproximation.docID() != DocIdSetIterator#NO_MORE_DOCS`?

HUSTERGS: If I understand correctly, `exclApproximation.docID()` should equals `exclDoc`, which is equal to `upTo` (under the `if` clause ), and `upTo` is less than `max`, so `exclApproximation.docID()` should never be `DocIdSetIterator#NO_MORE_DOCS`?

lucene/core/src/java/org/apache/lucene/search/ReqExclBulkScorer.java
HUSTERGS: I think it's correct here, and I run lucene test locally, everything looks fine. What I'm concerned here is that only `DocValuesRangeIterator` implement it's own `docIdRunEnd` method for now, I'm not sure whether adding this will hurt the performance or not

lucene/core/src/java/org/apache/lucene/search/ReqExclBulkScorer.java
gf2121: Oh yes, You are right. Thanks for explanation!

gf2121 (approved):
Lgtm

dweiss:
Unfortunately this commit broke some of the tests. Here is a repro:
```
./gradlew :lucene:core:test --tests "org.apache.lucene.search.TestLRUQueryCache.testRandom" -Ptests.file.encoding=US-ASCII -Ptests.gui=true -Ptests.haltonfailure=false -Ptests.jvmargs= -Ptests.jvms=6 -Ptests.multiplier=3 -Ptests.nightly=true -Ptests.seed=7600B69A193F746E -Ptests.vectorsize=256
```
the builds mailing list is showing more failures that are triggered from the asserting bulk scorer.

dweiss:
I will allow myself to revert this on main so that the builds pass, hope it's ok.

HUSTERGS:
> Unfortunately this commit broke some of the tests. Here is a repro:

Sorry about that, it seems the `docIDRunEnd()` can sometimes exceed the `max`, thus causes `AssertingBulkScorer` failed at line 83, then causes the thread leak:
https://github.com/apache/lucene/blob/973183258d1e11fee88a84b2a40b5fe02f0c477d/lucene/test-framework/src/java/org/apache/lucene/tests/search/AssertingBulkScorer.java#L82-L84

I changed 
```diff
- upTo = exclApproximation.docIDRunEnd();
+ upTo = Math.min(max, exclApproximation.docIDRunEnd());
```
And the test can pass properly locally

dweiss:
Could you file another PR that brings this back? We'd probably need to beast this more - you can do this locally too, see:
https://github.com/apache/lucene/blob/main/help/tests.txt#L93-L127

HUSTERGS:
> Could you file another PR that brings this back? We'd probably need to beast this more - you can do this locally too, see: https://github.com/apache/lucene/blob/main/help/tests.txt#L93-L127

Thanks for your guidance ! Will run the test locally to verify the change before I raise a new PR.

PR #14807, opened 2025-06-18
Remove unused delGen() method from FieldTermIterator

romseygeek:
This was used in the past to correctly order terms in MergedPrefixCodedTermsIterator, 
but that class was removed long ago so the method is no longer needed.


stefanvodita (commented):
Shall we add this to the 11.0 changelist?

romseygeek:
It's a package-private class so not part of the public API - I don't think I would normally add a CHANGES entry here?

stefanvodita:
Sorry, my bad, you're right!

stefanvodita (approved):

PR #14813, opened 2025-06-19
Use PriorityQueue instead of TreeMap in FirstPassGroupingCollector.

vsop-479:


vsop-479:
Benchmark: 
```
Benchmark                                   Mode  Cnt   Score    Error   Units
PriorityQueueBenchmark.getTopNWithHeap     thrpt   15  ≈ 10⁻³           ops/us
PriorityQueueBenchmark.getTopNWithTreeMap  thrpt   15  ≈ 10⁻⁴           ops/us
```

vsop-479:
I added a new benchmark (`FirstPassGroupingBenchmark`) to simulate `FirstPassGroupingCollector` 's implementation.
It seems use heap can gain a better performance:
```
Benchmark                                       Mode  Cnt  Score   Error   Units
FirstPassGroupingBenchmark.getTopNWithHeap     thrpt   15  0.010 ± 0.002  ops/us
FirstPassGroupingBenchmark.getTopNWithTreeMap  thrpt   15  0.005 ± 0.001  ops/us
```

vsop-479:
Emm~, Use heap will remove(pop) element in orderedGroups, which will result in `TestGrouping#testRandom` fail, since this test case `getTopGroups` many times.

PR #14814, opened 2025-06-19
Add isEmpty in PriorityQueue.

vsop-479:

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/PriorityQueue.java
+++ b/lucene/core/src/java/org/apache/lucene/util/PriorityQueue.java
@@ -343,7 +343,7 @@ public abstract class PriorityQueue<T> implements Iterable<T> {
     };
   }
 
-  /** Returns true if this list contains no elements */
+  /** Returns true if this heap contains no elements */
   public boolean isEmpty() {
     return size == 0;
   }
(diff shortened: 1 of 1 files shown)


lucene/core/src/java/org/apache/lucene/util/PriorityQueue.java
```diff
@@ -342,4 +342,13 @@ public T next() {
       }
     };
   }
+
+  /**
+   * Returns {@code true} if this list contains no elements.
```
stefanvodita: Should we just keep the other line in the comment?

stefanvodita (commented):
If we expand public API, can we have a good justification, e.g. use this somewhere in Lucene or explain how you would like to use it in your application?

lucene/core/src/java/org/apache/lucene/util/PriorityQueue.java
vsop-479: Yes, I removed it.

vsop-479: > e.g. use this somewhere in Lucene or explain how you would like to use it in your application?

I think it is useful when calling `PriorityQueue#pop` to fetch topN, just like a `Collection`, e.g. `Stack`, `ArrayList`, etc. 
Although `PriorityQueue` is not a `Collection`, and we can do this by judging `PriorityQueue#size`.

msokolov:
I believe the reason Collection has isEmpty() is there may be a way to more efficiently know if size > 0? But that doesn't apply here, and I don't think we ought to be adding sugar methods to conform to an interface we don't even implement

vsop-479:
Thanks for explaining @msokolov, closed.

PR #14816, opened 2025-06-19
Expose search strategy in KNN query

tteofili:
This simply exposes `AbstractKnnQuery#searchStrategy`. 
This might be useful to debug/inspect and perform optimizations based on different such strategies for consumers of KNN queries.


ChrisHegarty (approved):
LGTM

PR #14818, opened 2025-06-20
docs: fix invalid html

rmuir:
These html files (whether manually or automatically generated) have inconsistent usage of 'gt' and 'amp' html entities: sometimes they do it, sometimes they neglect to use it.

I added rules to try to keep these kinds of problems at bay: not really ideal but it works in lieu of some better linter for the long tail of file types, looks like this:

```
error[json-syntax]: Parse Error
  ┌─ lucene/test-framework/src/generated/checksums/generateEmojiTokenizationTest.json:4:38
  │
4 │     "property:unicodeVersion": "12.1",
  │                                      ^
  │
  = Treesitter parse error: editors and tools may treat it incorrectly.
    Check for trailing commas and comments, which JSON forbids.

error[css-syntax]: Parse Error
    ┌─ lucene/analysis/common/src/test/org/apache/lucene/analysis/charfilter/MS-Word 14 generated.htm:333:1
    │
333 │ <!--
    │ ^^^^
    │
    = Treesitter parse error: editors and tools may treat it incorrectly.
      Disable false-positives with /* ast-grep-ignore */
```


lucene/queryparser/docs/xml/LuceneContribQuery.dtd.html
```diff
@@ -27,7 +27,7 @@
 </td></tr></table>
 <p>BooleanQuerys implement Boolean logic which controls how multiple Clauses should be interpreted.
 Some clauses may represent optional Query criteria while others represent mandatory criteria.</p><p><span class='inTextTitle'>Example:</span> <em>Find articles about banks, preferably talking about mergers but nothing to do with "sumitomo"</em>
-</p><pre>
```
rmuir: I opened #14819 to address all these unrelated changes. if we can merge that one first, the diff here will become readable for review.

Issue #14819, opened 2025-06-20
fix sources to conform to .editorconfig


rmuir:
### Description

editorconfig with aggressive settings was applied to all files, but without fixing any existing problems or preventing new ones.

It causes the issue that when editing any impacted file, random unrelated formatting changes will appear. For example trailing whitespace is disabled, so if you edit any files that have it, your editor will correct that, leading to unrelated changes.

validation/autofix is tough: especially since indentation and line-length settings are applied to ALL files: these need ast-awareness to validating and fix correctly (e.g. spotless, prettier, etc). On the other hand, editors tend to not aggressively autofix these for the same reason.

We can start by fixing the most basic problems: e.g. trailing whitespace, missing newlines, etc in one fell swoop with low-tech tools, and add the commit to `.git-blame-ignore-revs`. These are the corrections that editors tend to do automatically across the entire individual file when saving it.

rmuir:
It looks to me like https://github.com/ec4j/editorconfig-gradle-plugin might be exactly what is wanted here for the build to keep this tidy.

its TextLinter will address the common ones that cause this kind of noise in PRs:
end_of_line, trim_trailing_whitespace, insert_final_newline. We can disable its XMLLinter which will do indentation as well for .xml and .xsl files, or just keep it, doesn't matter to me. I'm mostly concerned about the issues detected/fixed by TextLinter because editors will automatically introduce this noise into files otherwise.

cc: @dweiss

rmuir:
I'm having decent luck with the fast https://gitlab.com/greut/eclint, which allows disabling some of the problems in the .editorconfig with `eclint_` prefix. There are many categories of problems, e.g. invalid encoding: as we now say all non-binary files are UTF-8. I'm iterating with it, I've had to regenerate that huge DFA several times.

dweiss:
> It causes the issue that when editing any impacted file, random unrelated formatting changes will appear.

This is what I was worried about.

rmuir:
next step up, is the CI linter task. From the gradle side I will look to do it consistent with the ast-grep. 

because elements in .editorconfig, configure the editor, I think its ok to just invoke it in CI (or if the user opts-in via a property). Newly introduced problems going forwards should be rare.

rmuir:
closing this out: we're now linting in CI, space and indentation is fixed.

One thing that happened was disabling the `max_line_length`. Personally I'm ok with that, but if you don't like it, feel free to make a PR as long as sources are fixed to match :)

[...]

PR #14820, opened 2025-06-20
style: fix sources to conform to .editorconfig (basic fixes only)

rmuir:
Rather than have editors introduce these changes incrementally in unrelated PRs over time, fix them in one commit which may be added to .git-blame-ignore-revs as a followup commit.

Steps performed:
1. comment out indentation and line-length settings from .editorconfig (fixes are not "basic")
2. ran eclint -fix
3. ran gradlew regenerate

Relates: #14819


rmuir:
all done, took a few iterations due to that gnarly encoding error and dealing with a long-tail of weird stuff.

Checker passes now:
```
$ time eclint

real	0m3.708s
user	0m6.757s
sys	0m0.464s
```

rmuir:
Several options if we want to use this linter in CI (i'm concerned the java-gradle based one may be slow):
* get binaries from releases: https://gitlab.com/greut/eclint/-/releases
* build+cache with `go install gitlab.com/greut/eclint/cmd/eclint`
* use their container/action

Avoid the javascript based tool with the same name which can be confusing...

build-tools/build-infra/src/main/groovy/lucene.regenerate.gradle
```diff
@@ -219,7 +219,8 @@ configure([
           // Recompute checksums after the task has completed and write them.
           def updatedChecksums = computeChecksummedEntries(sourceTask)
           checksumsFile.setText(
-              JsonOutput.prettyPrint(JsonOutput.toJson(new TreeMap<String, String>(updatedChecksums))), "UTF-8")
+              JsonOutput.prettyPrint(JsonOutput.toJson(new TreeMap<String,
+              String>(updatedChecksums))) + '\n', "UTF-8")
```
dweiss, line 224: This is mildly surprising. Passes with greclipse and both versions are "ok" for it but I wonder if conflicts between the rules for different systems (editorconfig, spotless) won't cause headaches.

dweiss (approved):
I also ran ```gradlew regenerate --rerun-tasks```, no changes. lgtm.

dweiss:
If we want to run it on a CI, I think the gh action seems to be the easiest. Alternatively, running docker directly but it requires setting up git safe.directory because eclint uses git to list the files to validate.

rmuir:
I'm guessing its not on the approved list at apache though. The action just uses their docker image.

build-tools/build-infra/src/main/groovy/lucene.regenerate.gradle
rmuir: This is why I think it is bad to have inconsistent max_line_length set in the editorconfig. Any time I edit gradle you are going to see a change like this.

rmuir:
I'm gonna land this one (it was extremely annoying to get to this stage with the large DFA regeneration etc), and followup with tooling on a separate PR. If some problems sneak in between those times, I'll fix them.

PR #14822, opened 2025-06-20
style: disable max-line-length in editorconfig

rmuir:
This is set for all files and causes issues where editors will wrap inconsistently with the formatter. It will happen even with java code: because spotless ignores this restriction for comments, as an example.

Best to leave such settings to tools like spotless which are AST-aware.

Relates: #14819


rmuir:
Full list of [violations.txt](https://github.com/user-attachments/files/20835991/violations.txt)

dweiss (approved):

PR #14824, opened 2025-06-20
Custom google java format tasks to replace spotless

dweiss:
Spotless's support for google java format is kind of slow - I don't think it implements gradle's workers api so concurrency is limited to different projects. 

This issue aims to provide two custom tasks for applying and checking google java format, then use them as a substitute for those found in spotless.

This is in draft state and not fully working but it seems faster for me already (although there are open questions I've been struggling with).

Remaining:
* [x] improve display of diff'ed changes.
* [x] add eol normalization and trailing-line tweaks, which allows us to bypass spotless-java entirely,
* [x] correct any the hooks that currently link to spotless tasks (regenerate?)
* [x] benchmark differences in speed (cold start, warm start, incremental)
* [x] maybe consider an isolated runner (it'd allow us to remove jvm flags passed to gradle) [can't do it - too many forked jvms on high-end machines]
* [x] consider larger batches of files passed to each action: ```lucene.gjf.batchSize``` build option.


build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/spotless/CheckGoogleJavaFormatTask.java
```diff
@@ -0,0 +1,195 @@
...
+      StringBuilder sb = new StringBuilder();
+      for (Path changeset : allChanges) {
+        if (sb.length() > 1024 * 1024 * 2) {
+          sb.append("(...and more files, omitted)");
+          break;
+        }
+
+        sb.append(Files.readString(changeset));
+        sb.append("\n");
+      }
```
dweiss: @breskeby - how do you deal with worker api actions that produce something which the task should check/aggregate? The above solution seems... crude at best. It works but I am not happy with it. I also couldn't find any examples of worker api actions returning or producing something the source task should consume.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/spotless/ParentGoogleJavaFormatTask.java
```diff
@@ -0,0 +1,69 @@
...
+                })
+                .map(FileChange::getFile)
+                .toList();
+    }
+
+    @Internal
+    protected WorkQueue getWorkQueue() {
+        // TODO: maybe fork a separate jvm so that we can pass open-module settings there and fine-tune the jvm for the task?
+        return getWorkerExecutor().noIsolation();
+    }
```
dweiss: I'd love to fork a separate jvm here - where we can set some options that open up modules, etc. The problem is - there is no way to control how many isolated processes this forks. On high-end multi-core machines I see dozens of forked java processes - this doesn't make sense and is super heavy. Is there a way to control the concurrency of the isolated worker executor, @breskeby ?

dweiss:
```
> time ./gradlew clean checkGoogleJavaFormat  --no-daemon
real    0m13.884s
user    1m14.370s
sys     0m8.233s

> time ./gradlew clean spotlessJavaCheck --no-daemon
real    0m29.668s
user    1m27.121s
sys     0m8.733s
```

I had to remove ```-XX:ActiveProcessorCount=1``` from gradle.properties for the above. I think this setting for some reason makes gradle sluggish and unpredictable (the timings vary a lot).

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/spotless/ApplyGoogleJavaFormatTask.java
```diff
@@ -0,0 +1,77 @@
...
+
+            for (File inputFile : getParameters().getTargetFiles().getFiles()) {
+                var inputPath = inputFile.toPath();
+                try {
+                    String input = Files.readString(inputPath);
+                    String output = formatter.formatSourceAndFixImports(input);
+
+                    if (!input.equals(output)) {
+                        Files.writeString(inputPath, output);
+                    }
```
dweiss: @breskeby this is one example where I think gradle build model doesn't shine - for this task, the outputs and inputs overlap. It takes input files, formats them and writes them back to the same location. I know Ned Twigg (spotless) had similar problems.

rmuir:
> I had to remove `-XX:ActiveProcessorCount=1` from gradle.properties for the above. I think this setting for some reason makes gradle sluggish and unpredictable (the timings vary a lot).

For just the `org.gradle.jvmargs`, correct? I think it makes sense to nuke it here. It makes better sense for forked processes (e.g. `tests.jvmargs`), where, on an 8 processor machine, you want to avoid fork()'ing 8 JVMS, each of which is sized for 8 CPU machine (8 * 8 = 64).

dweiss:
Right - it's gradle.properties. I'll be experimenting and will try to push this forward - it's a promising direction to cut some time from one of the most expensive tasks at the moment.

image

rmuir:
You could also consider cheating for the local developer. For a formatting task, it is safe to only format changed files (eg based on git status or gradle cache or whatever). For formatting there aren't dependencies between files. As long as user can do a `CI=true ./gradlew tidy` to force it happening the slow way, corner cases such as formatter upgrades can all be handled.

dweiss:
This is already part of this patch, Robert - it is incremental so it should only apply to files that have changed since the last run. I don't think it's smart enough to do ratchet-style updates based on git history or anything like that.

rmuir:
Sorry for the noise, just my lack of gradle knowledge there.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/spotless/ParentGoogleJavaFormatTask.java
breskeby: forking the jvm is still expensive to some degree. I usually try to get away with isolated first if I need separate classpaths. 

you should be able to do configure the jvm options like this:

```
return getWorkerExecutor().processIsolation(spec -> {
      spec.forkOptions(javaForkOptions -> {
        javaForkOptions.jvmArgs("-Xmx2g", "-Xms2g");
      });
    });
```

There is no build-in mechanism to restrict the parallization for a specific task implementation. gradle just has a global property of `org.gradle.workers.max` that defaults to #cpus which is used when forking jvms for workers, test tasks, etc etc.
Is the point here to parallise the handling of sources in a single project and speed up format checks by this?

If forking a process is needed I would actually rework the implementation of the worker api to not submit one file after another but operate on the list of sources within the worker. This would basically mean having one worker per task. It'll still really improve the paralleiism as you can run on different sourcesets per project.

breskeby:
This is pretty cool. In elasticsearch we're also struggling with spotless and this looks like a good way forward, also reducing third party plugin dependencies which have proven to be brittle in the past.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/spotless/ParentGoogleJavaFormatTask.java
dweiss: I have already experimented with this (multiple files per worker). The problem is that they're not balanced - lucene/core has a lot more files than anything else, it's the long tail.

I also used process isolation but this won't fly - I have 24 workers level on my machine and this results in dozens of forked jvms - it costs memory and overhead of booting up those additional JVMs. I'll experiment some more though. Thanks for your feedback.

breskeby: Thinking about this more I wonder if you can restrict the parallization by using a BuildService. Again this is not exactly the simple api for restricting parallization but it should work. 
Here's an example on how to use it with WorkerAPI: https://docs.gradle.org/current/userguide/build_services.html#using_a_build_service_with_the_worker_api 

Then you would register the service on project level and set a maxParallel usage. Here's an example on how we throttle things this way in the elasticsearch build: https://github.com/elastic/elasticsearch/blob/7b6bdfa323fcc7460dc252b3dbc84c6b1314fb66/build-tools/src/main/java/org/elasticsearch/gradle/testclusters/TestClustersPlugin.java#L126-L133

dweiss: Ah, that's an idea! This is pretty complex. Let's leave it for a follow-up, if it's needed. I'm quite happy with the current state - no isolation it is but it seems a lot faster than going through spotless. I should finish this today.

dweiss:
Fun. https://github.com/google/google-java-format/issues/1260

dweiss:
Here are some benchmarks from my Linux system. 

This runs the patch with varying number of workers and the default
```-XX:ActiveProcessorCount=1``` in gradle.properties. This is also the worst-case scenario of reformatting
all files from scratch, no incremental information. My system is an Ubuntu AMD
Ryzen Threadripper 3970X, 32 core.

```
echo "With bg daemon: './gradlew clean checkGoogleJavaFormat'."
./gradlew -q clean 
./gradlew -q --stop
for workers in 1 2 4 8 16 32; do
  echo "max-workers: $workers"
  (for i in `seq 1 3`; do time ./gradlew clean checkGoogleJavaFormat --max-workers $workers ; done ) 2>&1 | grep "real"
done
```
results:
```
max-workers: 1
real	0m51.099s
real	0m42.128s
real	0m41.845s
max-workers: 2
real	0m23.382s
real	0m23.619s
real	0m23.563s
max-workers: 4
real	0m15.160s
real	0m15.877s
real	0m15.169s
max-workers: 8
real	0m13.651s
real	0m13.534s
real	0m13.468s
max-workers: 16
real	0m18.783s
real	0m18.273s
real	0m18.813s
max-workers: 32
real	0m26.930s
real	0m26.884s
real	0m27.259s
```

The CPU is mostly idle during most of these runs. Weird. If I remove ```-XX:ActiveProcessorCount=1```
from gradle.properties, I get these results:
```
max-workers: 1
real	0m43.419s
real	0m41.613s
real	0m41.829s
max-workers: 2
real	0m22.878s
real	0m22.745s
real	0m22.604s
max-workers: 4
real	0m13.625s
real	0m13.452s
real	0m13.608s
max-workers: 8
real	0m8.392s
real	0m8.490s
real	0m8.319s
max-workers: 16
real	0m6.225s
real	0m6.929s
real	0m6.356s
max-workers: 32
real	0m6.052s
real	0m6.863s
real	0m6.204s
```
so it clearly is a benefit if you have higher core counts. It's also close to the lower limit
of manually running google-java-format on all source files (they do have concurrent processing 
inside).

For the incremental case... it's fast enough (well, it doesn't do anything), even from a
"cold" start, without any daemon in the background (the first call will show configuration
time):
```
echo "With bg daemon, from cold-start, incremental: './gradlew checkGoogleJavaFormat'."
./gradlew checkGoogleJavaFormat
for workers in 2 4 8; do
  echo "max-workers: $workers"
  ./gradlew --stop -q
  (for i in `seq 1 3`; do time ./gradlew checkGoogleJavaFormat --max-workers $workers ; done ) 2>&1 | grep "real"
done
```
results:
```
max-workers: 2
real	0m8.953s
real	0m2.105s
real	0m2.002s
max-workers: 4
real	0m8.746s
real	0m2.061s
real	0m2.037s
max-workers: 8
real	0m8.856s
real	0m2.054s
real	0m1.978s
```

Rather internal detail but shows different batch sizes of input files 
for a constant number of workers (the default is 5):
```
for batchSize in 1 2 4 8 16 32 64; do
  echo "batch size: $batchSize"
  (for i in `seq 1 3`; do time ./gradlew clean checkGoogleJavaFormat -Plucene.gjf.batchSize=$batchSize --max-workers 8 ; done ) 2>&1 | grep "real"
done
```
results:
```
batch size: 1
real	0m8.958s
real	0m8.760s
real	0m8.585s
batch size: 2
real	0m8.542s
real	0m8.401s
real	0m8.481s
batch size: 4
real	0m8.341s
real	0m8.378s
real	0m8.469s
batch size: 8
real	0m8.671s
real	0m8.494s
real	0m8.458s
batch size: 16
real	0m8.363s
real	0m8.408s
real	0m8.395s
batch size: 32
real	0m8.384s
real	0m8.320s
real	0m8.423s
batch size: 64
real	0m8.578s
real	0m8.622s
real	0m8.699s
```

Finally, the same check for the previous, spotless-based implementation (main branch).
```
./gradlew --stop
./gradlew clean
for workers in 1 2 4 8 16 32; do
  echo "max-workers: $workers"
  (for i in `seq 1 3`; do time ./gradlew clean spotlessJavaCheck --max-workers $workers ; done ) 2>&1 | grep "real"
done
```
results:
```
max-workers: 1
real	0m49.843s
real	0m47.934s
real	0m48.256s
max-workers: 2
real	0m28.170s
real	0m27.980s
real	0m27.851s
max-workers: 4
real	0m21.620s
real	0m21.475s
real	0m21.503s
max-workers: 8
real	0m21.192s
real	0m20.962s
real	0m20.895s
max-workers: 16
real	0m20.984s
real	0m20.783s
real	0m20.773s
max-workers: 32
real	0m21.290s
real	0m21.077s
real	0m21.037s
```

Faster. Note I didn't do anything here - all the heavy lifting is done by the same implementation
in google-java-format. The difference is in the long-tail of the longest operation (formatting 
lucene/core), which is now parallel.

I also toyed with removing "-XX:TieredStopAtLevel=1" from gradle.properties, then re-ran the benchmark:
```
./gradlew -q clean 
./gradlew -q --stop
for workers in 8 16 32; do
  echo "max-workers: $workers"
  (for i in `seq 1 3`; do time ./gradlew clean checkGoogleJavaFormat --max-workers $workers ; done ) 2>&1 | grep "real"
done
```
results:
```
max-workers: 8
real	0m16.147s
real	0m6.014s
real	0m5.939s
max-workers: 16
real	0m4.830s
real	0m4.700s
real	0m4.716s
max-workers: 32
real	0m5.271s
real	0m5.228s
real	0m5.243s
```

So you get that initial "hit" when hotspot compiles all that code in the daemon, then it's a bit faster
compared to what we currently have as the default. I don't know what's better.

**Comment by rmuir** [comment]
@dweiss maybe remove the parallelgc along with if you are testing. It's only needed because with 1 CPU ergonomics will default you to SerialGC which is too slow.

**Comment by dweiss** [comment]
With:
```
-Xmx1g -XX:ReservedCodeCacheSize=256m  -XX:TieredStopAtLevel=1
```
I get:
```
max-workers: 8
real	0m9.768s
real	0m9.486s
real	0m9.501s
max-workers: 16
real	0m6.540s
real	0m6.374s
real	0m6.367s
max-workers: 32
real	0m6.166s
real	0m6.379s
real	0m6.477s
```

With:
```
-Xmx1g -XX:ReservedCodeCacheSize=256m
```
I get:
```
max-workers: 8
real	0m6173s
real	0m6.335s
real	0m6.074s
max-workers: 16
real	0m4.750s
real	0m4.610s
real	0m4.498s
max-workers: 32
real	0m4.917s
real	0m5.099s
real	0m5.702s
```

With:
```
-Xmx1g -XX:ReservedCodeCacheSize=256m -XX:+UseParallelGC
```
I get:
```
max-workers: 8
real	0m6.042s
real	0m6.052s
real	0m5.873s
max-workers: 16
real	0m4.899s
real	0m4.879s
real	0m4.735s
max-workers: 32
real	0m5.137s
real	0m5.070s
real	0m5.250s
```

So not that much of a difference.

dweiss:
On gh runners changing these options seems to have no effect. 
* before removing ```-XX:TieredStopAtLevel=1 -XX:ActiveProcessorCount=1```:
image
* after removing ```-XX:TieredStopAtLevel=1 -XX:ActiveProcessorCount=1```:
image

* It is faster than on main though, so at least something. 
image

rmuir:
The runners in use might not be beefy enough for larger sizing (gc threads, thread pools, etc) to help?

PR #14825, opened 2025-06-21
build: add eclint to verify editorconfig in CI

rmuir:
You generally shouldn't need the check, as your editor should follow the editorconfig and not introduce the errors checked-for here.

But it helps maintain the .editorconfig itself and ensure that files in the repository actually match whatever is configured there.

If you want to enable the check locally for debugging:
```
GOBIN=${HOME}/.local/bin go install gitlab.com/greut/eclint/cmd/eclint@latest
echo "lucene.tool.eclint=eclint" >> ~/.gradle/gradle.properties
```

The eclint action is not used, as it is not on the apache approved list. Getting the 3MB binary in CI takes 2s (cached) and 9s (uncached, using setup-go).

The check itself runs fast:
```
real	0m3.647s
user	0m6.702s
sys	0m0.474s
```

@dweiss, I imagine you want to keep things a certain way, please feel free to push commits here. I just hack when it comes to the gradle side of things, but the actions part is decent I think.

### DIFF
--- a/.github/actions/eclint/action.yml
+++ b/.github/actions/eclint/action.yml
@@ -1,0 +1,33 @@
+name: Install eclint
+description: Installs eclint from cache, or builds it
+
+inputs:
+  eclint-version:
+    required: false
+    description: "eclint version to use"
+    default: "v0.5.1"
+
+runs:
+ using: "composite"
+ steps:
+  - name: Cache eclint binary
+    id: cache-eclint
+    uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+    with:
+      path: ~/.local/bin/eclint
+      key: eclint-${{ inputs.eclint-version }}-${{ runner.os }}-${{ runner.arch }}
+  - name: Setup go
+    if: steps.cache-eclint.outputs.cache-hit != 'true'
+    uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
+    with:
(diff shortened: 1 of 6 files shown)


.github/actions/eclint/action.yml
```diff
@@ -0,0 +1,33 @@
...
+
+runs:
+ using: "composite"
+ steps:
+  - name: Cache eclint binary
+    id: cache-eclint
+    uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+    with:
+      path: ~/.local/bin/eclint
+      key: eclint-${{ inputs.eclint-version }}-${{ runner.os }}-${{ runner.arch }}
```
rmuir, line 18: May seem overkill to only save single-digit seconds, but the purpose is to avoid the network requests (similar to gradle-wrapper).

.github/actions/eclint/action.yml
```diff
@@ -0,0 +1,33 @@
+name: Install eclint
+description: Installs eclint from cache, or builds it
+
+inputs:
+  eclint-version:
+    required: false
+    description: "eclint version to use"
+    default: "v0.5.1"
```
rmuir, line 8: we could alternatively do go.mod/go.sum to let dependabot take care, but its a bit more awkward and I'm afraid of tons of noise from the indirect deps. Plus those files use tab indentation :)

dweiss (approved):
This looks perfect to me. Thank you.

PR #14826, opened 2025-06-21
build: fix incorrect gradle dependency

rmuir:
applyAstGrep doesn't run in CI, because it is set to depend on "tidy" instead of "check"


dweiss (approved):
Of course, sorry, don't know what I was thinking here. Perhaps I wasn't thinking at all.

PR #14828, opened 2025-06-21
[Build] Fix deprecation warning about automatic loading of test framework

breskeby:
### Description

This fixes deprecation warning that will result in a failure with Gradle 9.0: 

"The automatic loading of test framework implementation dependencies has been deprecated." 

The gradle test tasks relies on having the test framework on its classpath as the TestWorker has dependencies on this.

In the existing setup we had only module classpaths declared in a jvm arg provider which is not taken into account by Gradle when setting up the test worker jvm.

### DIFF
--- a/build-tools/build-infra/src/main/groovy/lucene.java.modules.gradle
+++ b/build-tools/build-infra/src/main/groovy/lucene.java.modules.gradle
@@ -188,7 +188,7 @@ allprojects {
         jvmArgumentProviders.add(modularPaths.runtimeArguments)
 
         // Modify the default classpath.
-        classpath = modularPaths.runtimeClasspath
+        classpath = modularPaths.testRuntimeClasspath
 
         doFirst {
           modularPaths.logRuntimePaths(logger)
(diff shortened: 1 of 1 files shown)


dweiss (approved):
I think they're doing too much at gradle. They shouldn't be messing with those parameters - it reminds me of the javadoc task which is the same way (there is an issue I filed about it at some point, somewhere).

breskeby:
it will be actually interesting how this will behave in gradle 9.0. Ideally it would just work as junit would be found on the module path in this case here. But I'm afraid they keep the check and just fail. 🤞 . I'll will do some testing against the 9.0-rc-1 at one point.

dweiss:
> But I'm afraid they keep the check and just fail. 

I think they shouldn't mess with extra arguments - maybe you can convince them so. It's similar to this issue -
https://github.com/gradle/gradle/issues/11898 Messing with arguments behind the dev's back is crazy and not intuitive at all. Maybe the intentions are good but it's an evil feature.

PR #14831, opened 2025-06-22
build: replace six simple error prone checks

rmuir:
Error-prone has many checks, but the tool doesn't scale well with respect to many checks: replace six (arbitrary) easy ones.

ast-grep:
* https://errorprone.info/bugpattern/SubstringOfZero
* https://errorprone.info/bugpattern/JUnit4ClassAnnotationNonStatic
* https://errorprone.info/bugpattern/JUnit4EmptyMethods
* https://errorprone.info/bugpattern/PackageInfo
* https://errorprone.info/bugpattern/IncorrectMainMethod

forbidden-apis:
* https://errorprone.info/bugpattern/ICCProfileGetInstance


lucene/test-framework/src/test/org/apache/lucene/tests/util/TestBeforeAfterOverrides.java
```diff
@@ -30,7 +30,9 @@ public TestBeforeAfterOverrides() {
 
   public static class Before1 extends WithNestedTests.AbstractNestedTest {
     @Before
-    public void before() {}
+    public void before() {
+      /* intentionally left blank */
+    }
```
rmuir, line 35: their checker is kinda wimpy on these and wasn't catching these. I added a comment node inside each one so they wouldn't be empty anymore: hope this is OK. 

alternatively we could add ignore entire file or try to use an `ast-grep-ignore:` comment.

dweiss: It's clearer now.

dweiss (approved):

PR #14832, opened 2025-06-22
style: fix indents inconsistent with editorconfig and validate in CI

rmuir:
editorconfig indents were set across the board for all files, but nothing was done with existing files. For example, yaml was set to have a 2-space indent, yet it is the wild-west out there: 1-space indents, 3-space indents, you name it.

* Validate any indent settings specified in editorconfig
* Fix all existing violations (html, xml, etc were especially painful)
* Disable indent settings where it isn't appropriate or reasonable (e.g. .txt files)


rmuir:
The validation does make `eclint` run slower, but it still burns relative to other tasks in the build:
```
$ time eclint

real	0m5.289s
user	0m7.823s
sys	0m0.601s
```

dweiss (approved):
Interesting how much comment garbage there was in java files!

rmuir:
will also push a commit adding this one to `.git-blame-ignore-revs`.

PR #14837, opened 2025-06-23
Small enhancements to IndexWriter's InfoStream to support segment tracing

mikemccand:
Some small fixes to IndexWriter's InfoStream logging, uncovered when working on the new segment tracing tool from luceneutil (example: https://githubsearch.mikemccandless.com/segments_15.html).  I've been sitting on these changes in my main dev area and finally got around to making a PR before I lose track of them, phew!

Details:
  * KNN merging was logging "start/done" multiple times and then I realized it was for a different field each time, so I added field name
  * Thread name was duplicated in a single line at end of full flush
  * Defensive change: don't invoke merge-on-commit logic if merge policy returns empty list of merges
  * I enhanced merge-on-commit logging to state how many merges it will do, and whether all merges completed in the timeout window or not
  * TMP now also logs its deletesPctAllowed and forceMerteDeletesPctAllowed, and more details when it kicks off merge selection

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -3724,7 +3724,7 @@ public class IndexWriter
         maybeCloseOnTragicEvent();
       }
 
-      if (pointInTimeMerges != null && pointInTimeMerges.merges.size() > 0) {
+      if (pointInTimeMerges != null) {
         if (infoStream.isEnabled("IW")) {
           infoStream.message(
               "IW", "now run merges during commit: " + pointInTimeMerges.segString(directory));
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -3724,7 +3724,7 @@ public class IndexWriter
         maybeCloseOnTragicEvent();
       }
 
-      if (pointInTimeMerges != null) {
+      if (pointInTimeMerges != null && pointInTimeMerges.merges.size() > 0) {
         if (infoStream.isEnabled("IW")) {
           infoStream.message(
               "IW", "now run merges during commit: " + pointInTimeMerges.segString(directory));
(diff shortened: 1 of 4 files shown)


msokolov:
Seems a bit weird to include a merge policy change in a PR that is mostly about changing logging, but OK. I think I would be happier if the PR was entitled "Don't invoke merge-on-commit logic if merge policy returns empty list of merges" and secondarily patched infoStream logging.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3724,18 +3724,25 @@ private long prepareCommitInternal() throws IOException {
         maybeCloseOnTragicEvent();
       }
 
-      if (pointInTimeMerges != null) {
+      if (pointInTimeMerges != null && pointInTimeMerges.merges.size() > 0) {
         if (infoStream.isEnabled("IW")) {
           infoStream.message(
               "IW", "now run merges during commit: " + pointInTimeMerges.segString(directory));
```
vigyasharma, line 3730: There's a TODO to deprecate `segString(Directory dir)`, and indeed the directory is not used anywhere. We can simply use `toString()` instead. I just happened to find it as I wanted to make sure we log the number of submitted merges. We don't need to address it in this PR, in fact, it could be a good first issue.

vigyasharma (approved):

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
vigyasharma: created https://github.com/apache/lucene/issues/14899 as a "good first issue"

mikemccand: Thanks @vigyasharma -- good idea!

mikemccand:
> Seems a bit weird to include a merge policy change in a PR that is mostly about changing logging, but OK. I think I would be happier if the PR was entitled "Don't invoke merge-on-commit logic if merge policy returns empty list of merges" and secondarily patched infoStream logging.

Oh I can revert that part if you want?  And break it into its own PR ...

My primary purpose here was to fix confusing issues with the `InfoStream` as I was trying to get segment tracing working, and then when I saw MOC logging a pointless/confusing message (`now run merges during commit:` followed by nothing) I wanted to fix that.  But you're right -- it does a lot of addition (pointless) things (besides logging) inside that `if`...

mikemccand:
OK I restored that `if` @msokolov -- so now if the `MergePolicy` returns empty list of merges it'll still run through all the MOC motions, logging empty messages, etc.  I'll open a separate PR for that change ... now this PR is really just fixing up `InfoStream` logging.  I plan to merge soon ...

PR #14838, opened 2025-06-23
Make it possible to extend Patience/Seeded knn queries

tteofili:
Currently `PatienceKnnQuery` extends and is constructed by an `AbstractKnnQuery`, which is not public and therefore avoids other classes to extend it (as the related ctor is also package private).
This simply makes `PatienceKnnQuery` constructors public and strict about the type of the wrapped query (float, byte, seeded knn queries),

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/PatienceKnnVectorQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/PatienceKnnVectorQuery.java
@@ -285,7 +285,11 @@ public class PatienceKnnVectorQuery extends AbstractKnnVectorQuery {
           new SeededKnnVectorQuery(
               seededKnnVectorQuery.delegate,
               seededKnnVectorQuery.seed,
-              seededKnnVectorQuery.createSeedWeight(indexSearcher));
+              seededKnnVectorQuery.createSeedWeight(indexSearcher),
+              delegate.field,
+              delegate.k,
+              delegate.filter,
+              delegate.searchStrategy);
     }
     return super.rewrite(indexSearcher);
   }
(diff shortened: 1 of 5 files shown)


tteofili:
nit: this also fixes a minor javadoc issue

lucene/core/src/java/org/apache/lucene/search/PatienceKnnVectorQuery.java
```diff
@@ -123,8 +123,24 @@ public static PatienceKnnVectorQuery fromSeededQuery(SeededKnnVectorQuery knnQue
...
 
-  PatienceKnnVectorQuery(
-      AbstractKnnVectorQuery knnQuery, double saturationThreshold, int patience) {
+  public PatienceKnnVectorQuery(
+      SeededKnnVectorQuery knnQuery, double saturationThreshold, int patience) {
+    super(knnQuery.field, knnQuery.k, knnQuery.filter, knnQuery.searchStrategy);
+    this.delegate = knnQuery;
+    this.saturationThreshold = saturationThreshold;
+    this.patience = patience;
+  }
```
benwtrent: We should have one package private ctor for `AbstractKnnVectorQuery` that the public ones reference.

benwtrent (approved):
Would be good to do this for `SeededKnnVectorQuery` as well while we are at it :)

It follows the same pattern.

tteofili:
@benwtrent extended this to `SeededKnnVectorQuery ` in [ecc37de](https://github.com/apache/lucene/pull/14838/commits/ecc37de92b9c78fecfdf5214ba219692c873c4ad) (changed also PR description and changelog entry accordingly).

PR #14839, opened 2025-06-23
assertDocValuesEquals should support sparse sorted doc_values

parkertimmins:
In LuceneTestCase, assertDocValuesEquals compares the doc_values in two indices. For SortedSetDocValues, as well as other doc value types, it does not assume that every document has a value. But for SortedDocValues, it requires that every doc have a value. This restricts what indices `assertDocValuesEquals` can be used to test. So this PR removes this restriction and does not assume that every doc has a value.


lucene/core/src/test/org/apache/lucene/index/TestDuelingCodecs.java
```diff
@@ -113,6 +114,11 @@ public static void createRandomIndex(int numdocs, RandomIndexWriter writer, long
       for (String trash : split) {
         document.add(new SortedSetDocValuesField("sortedset", new BytesRef(trash)));
       }
+      // add a sorted dv field sometimes
+      document.removeFields("sparseset");
+      if (random.nextBoolean()) {
+        document.add(new SortedDocValuesField("sparseset", new BytesRef(title)));
+      }
```
parkertimmins, line 121: This function is also used in `TestDuelingCodecsAtNight`. This comment: https://github.com/apache/lucene/blob/a94fbb79ac6e60c1ebfa76b24cab23fabcb60bbf/lucene/core/src/test/org/apache/lucene/index/TestDuelingCodecsAtNight.java#L24 made me a bit concerned that adding an additional field will slow down the nightlies. But I ran several tests with the same seed both with and without the above block, and saw no appreciable difference. (In fact, the tests ran slightly faster on average with new field, though this must be noise.)

jpountz (approved):

jpountz:
Can you add a CHANGES entry under 10.3 before I merge?

parkertimmins:
@jpountz Whoops, I missed that it needed a change log entry. Thanks for the heads up. Add an entry to the 10.3 section. Thanks!

Issue #14835, opened 2025-06-23
A workflow "bot" that would apply formatting fixes and create a PR against a PR?


dweiss:
### Description

There are more different tools now used to harness code quality. I wonder if it'd be possible to create a workflow that would:
* run a set of designated formatting-validation tasks,
* if these fail, apply automatic fixes (like the tidy task), try to rerun formatting validation tasks again,
* if the fixes are successful, file a pull request against the original repository/branch (a PR against a PR).

We could even make this workflow an opt-in - using "commands" like the ones dependabot uses. 

This is just an idea for anybody willing to play with it - I have no experience with such workflows and I probably won't have the time to dig deep into this. It seems like a fun task though.

rmuir:
We could also try e.g. `--format github` option of tools to get better integration. I think they typically use "annotations" and might be much more friendly I'm not sure how autofixes look but it might be an easy win, and I think many tools support "reporters" that are compatible with this.

screenshot of unrelated tool so you know what i mean by annotation: 

![Image](https://github.com/user-attachments/assets/90566233-f9b6-4850-9413-aba78050acdb)

Currently all tools just print to the console in CI.

We can try it out with ast-grep as an experiment

dweiss:
This looks nice.

> Currently all tools just print to the console in CI.

Not all of them - this one creates a report that's part of the workflow's result -
https://github.com/apache/lucene/actions/runs/15533320029

georgereuben:
Hi @rmuir @dweiss , this seems like an interesting task that I'd like to work on. Can I take this up?

dweiss:
Absolutely, if you have the time - go ahead and explore what's possible out there./

PR #14840, opened 2025-06-24
ci: enable gh annotations with ast-grep

rmuir:
Error is still printed to console, but in a special structured format recognized by actions to highlight and annotate the problem in your PR.

When clicking the build failure, annotations can be seen at the very top of action link (no need to look at logs). Github does highlight the failures in red in the logs as well.

You can see the problems annotated inline in the sources, when reviewing the changes.

Relates: #14835

@dweiss, I know it doesn't solve that issue, nor does it even make use of the autofix I wrote, but I still think it is an improvement to usability.

Screen_Shot_2025-06-23_at_21 27 33


dweiss (approved):
This is awesome!

rmuir:
ill try to look if it makes sense to configure any other checkers... if you have ideas let me know. the EditorConfigLint would be my first idea to look at next otherwise, due to the fact it is golang tool and not "gradle native".

if linters doesn't support the actions format, we can alternatively put json files in .github directory that specify vim efm-like patterns: https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md

PR #14849, opened 2025-06-26
improve 'javadoc-style-license-header' and 'license-after-package'

rmuir:
these two checks are currently implemented with regex rules looking for comments, then calling into rat API to check if it is a license. It means parsing every java file with rat just for this purpose, and the regexes are not perfect.

The new rules are very simple but find some inconsistencies that weren't found before: they aren't stricter, the previous code just misses some.

We now only call into rat here for xml files, separately we should decide if that is worth it. If not, we can reduce the gradle code further and avoid challenges with Rat API on upgrades.

To me, checking this for the 29 XML files is not worth the hassle of installing additional C library / configuring `customLanguages` to check the xml with this tool, I add xml parser in my own personal config but I think its too much of a hassle for anyone else / CI.


gradle/validation/ast-grep/rules/java-patterns.yml
```diff
@@ -21,3 +21,25 @@ rule:
...
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ast-grep/ast-grep/refs/heads/main/schemas/java_rule.json
+id: license-after-package
+language: java
+rule:
+  kind: package_declaration
+  precedes:
+    matches: java-license
+ignores:
+  # TODO: something not right with regeneration of this lone file
+  - "**/StandardSyntaxParser.java"
```
rmuir: I TODO'd this one because I feel like something didn't work correctly with `regenerate` ?

As you can see, it has a `package` statement before the license header. But it is the only JavaCC file like this, so something seems off:

https://github.com/apache/lucene/blob/4b47fb1a3113d22bca6cd8c1664529ef2d7f4877/lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/standard/parser/StandardSyntaxParser.java#L3-L20

gradle/validation/ast-grep/rules/java-patterns.yml
```diff
@@ -21,3 +21,28 @@ rule:
...
+rule:
+  kind: package_declaration
+  precedes:
+    matches: java-license
+    # allow anything in between (e.g. import statements)
+    stopBy: end
+ignores:
+  # TODO: something not right with regeneration of this lone parser
+  - "**/StandardSyntaxParser.java"
+  - "**/StandardSyntaxParserTokenManager.java"
```
rmuir: Beefed up the check a bit and the same parser's tokenmanager has similar issues. See comment on the commit, I think maybe this parser hasn't been regenerated in a while? Maybe it is just missing from the gradle task...

https://github.com/apache/lucene/blob/4b47fb1a3113d22bca6cd8c1664529ef2d7f4877/lucene/queryparser/src/java/org/apache/lucene/queryparser/flexible/standard/parser/StandardSyntaxParserTokenManager.java#L3-L20

gradle/validation/ast-grep/rules/java-patterns.yml
dweiss: It works fine - the license just wasn't in the right place in the template. I moved it and removed the exclusion.

dweiss:
I've corrected the location of the ASF header in that one file you mentioned but I think there is a larger issue here - some other generated files don't have the license header at all. These are currently "accepted" by this fragment:
```
      approvedLicense(familyName: "Generated")
```
in ```lucene.validation.rat-sources.gradle```. I wonder if we should remove this and just glue the ASF header on all generated files, for consistency.

dweiss:
I've added ASL headers to antlr sources for symmetry with javacc. Please feel free to revert if you think this is too much for this patch.

rmuir:
Thanks for the help with the generated files: Consistency is good, I think.

PR #14851, opened 2025-06-26
Fewer virtual calls to enhance performance

kkewwei:

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -47,7 +47,7 @@ Optimizations
 ---------------------
 * GITHUB#14011: Reduce allocation rate in HNSW concurrent merge. (Viliam Durina)
 * GITHUB#14022: Optimize DFS marking of connected components in HNSW by reducing stack depth, improving performance and reducing allocations. (Viswanath Kuchibhotla)
-* GITHUB#14851: Fewer virtual calls to enhance performance. (Ke Wei)
+* GITHUB#14851: Fewer virtual calls to enhance performance of point queries. (Ke Wei)
 
 Bug Fixes
 ---------------------
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -47,6 +47,7 @@ Optimizations
 ---------------------
 * GITHUB#14011: Reduce allocation rate in HNSW concurrent merge. (Viliam Durina)
 * GITHUB#14022: Optimize DFS marking of connected components in HNSW by reducing stack depth, improving performance and reducing allocations. (Viswanath Kuchibhotla)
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
```diff
@@ -246,7 +246,11 @@ private static class BKDPointTree implements PointTree {
     private final DocIdsWriter docIdsWriter;
     // if true the tree is balanced, otherwise unbalanced
     private final boolean isTreeBalanced;
-
+    private final IntsRef scratchIntsRef = new IntsRef();
+    {
+      // This is here to not rely on the default constructor of IntsRef to set offset to 0
+      scratchIntsRef.offset = 0;
```
dweiss: Maybe change to an assertion then? I think it's safe to assume new IntsRef is going to have offset set to 0: we can just assert that it does (in tests) to prevent regressions.

lucene/CHANGES.txt
```diff
@@ -47,6 +47,7 @@ Optimizations
 ---------------------
 * GITHUB#14011: Reduce allocation rate in HNSW concurrent merge. (Viliam Durina)
 * GITHUB#14022: Optimize DFS marking of connected components in HNSW by reducing stack depth, improving performance and reducing allocations. (Viswanath Kuchibhotla)
+* GITHUB#14851: Fewer virtual calls to enhance performance. (Ke Wei)
```
jpountz:
```suggestion
* GITHUB#14851: Fewer virtual calls to enhance performance of point queries. (Ke Wei)
```

jpountz (approved):

lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
```diff
@@ -247,10 +247,7 @@ private static class BKDPointTree implements PointTree {
     // if true the tree is balanced, otherwise unbalanced
     private final boolean isTreeBalanced;
     private final IntsRef scratchIntsRef = new IntsRef();
-    {
-      // This is here to not rely on the default constructor of IntsRef to set offset to 0
-      scratchIntsRef.offset = 0;
```
dweiss: I didn't mean to remove it but perhaps add ``` assert scratchIntsRef.offset == 0```. Fine either way though.

kkewwei: Sorry for my misunderstood, I have reverted the change.

jpountz:
Thank you!

jpountz:
I backported and moved the changes entry to 10.3.

PR #14853, opened 2025-06-27
source patterns: @author/nocommit/TOOD/tabs

rmuir:
The tab rule is handled by eclint for all filetypes. @author/nocommit/TOOD moved to ast-grep rules: they are easier when they only match inside comment nodes to begin with, and can give better feedback from the CLI, editor, and GH actions.

This mornings sudoku puzzles to make more progress on the file.

### DIFF
--- a/build-tools/build-infra/src/main/groovy/lucene.validation.source-patterns.gradle
+++ b/build-tools/build-infra/src/main/groovy/lucene.validation.source-patterns.gradle
@@ -124,10 +124,6 @@ class ValidateSourcePatternsTask extends DefaultTask {
   @TaskAction
   public void check() {
     def invalidPatterns = [
-      (~$/@author\b/$) : '@author javadoc tag',
-      (~$/(?i)\bno(n|)commit\b/$) : 'nocommit',
-      (~$/\bTOOD:/$) : 'TOOD instead TODO',
-      (~$/\t/$) : 'tabs instead spaces',
       (~$/[\u202A-\u202E\u2066-\u2069]/$) : 'misuse of RTL/LTR (https://trojansource.codes)',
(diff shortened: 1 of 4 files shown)


dweiss (approved):
The only downside I see with moving everything to ast-grep is that it makes it slowly a requirement for people to install ast-grep locally... Not that it's such a big burden but it's an additional step to take. I'm sure Uwe won't be happy.

rmuir:
@dweiss I think most contributors familiar with the project are aware of stuff such as `@author` conventions and don't often hit these checks, I don't expect it to be a problem?

If you want it locally, you can install the tool in various ways (e.g. via `npm` / `pip` / `brew`) and can integrate with editor via `ast-grep lsp` (or e.g. vscode extension) so it lints-as-you-type and gives you lightbulbs with the autofixes.

The errors for new contributors on PRs are much better for the various strange checks.

We also have your GH issue to make `autofix` bot process, which I do think is a good one. I write the fixes when its easy and safe, but we could improve it too.

The CLI output is more clear than java regexps too:
Screen_Shot_2025-06-27_at_09 22 35

rmuir:
When we look at the precommit task times in CI, these are the ones I'm trying to incrementally make progress on right now:

```
Aggregate task times (possibly running in parallel!):
  85.78 sec.  compileTestJava <-- includes slow error-prone stuff
  83.79 sec.  compileJava <-- includes slow error-prone stuff
  83.00 sec.  renderSiteJavadoc
  60.26 sec.  renderJavadoc
  55.29 sec.  ecjLintMain
  50.46 sec.  ecjLintTest
  40.93 sec.  checkGoogleJavaFormat
  34.04 sec.  checkBrokenLinks
  13.92 sec.  validateSourcePatterns <-- includes slow regexps on every file
```

`checkBrokenLinks` also seems excessive, but that's for another day.

PR #14856, opened 2025-06-27
Only score doc first when minCompetitiveScore is zero

gf2121:
`minCompetitiveScore` could be set when `setScorer` so this proposes only score doc first when minCompetitiveScore is zero. Luceneutil does not show a noticeable improvement but this might be useful in some case.


jpountz (approved):

jpountz:
@gf2121 Let's merge this PR?

Issue #14857, opened 2025-06-27
AbstractKnnVectorQuery breaks shallowAdvance contract, causing chaos


benwtrent:
### Description

When utilizing knn vector queries in Lucene v9 in a context that requires advanceShallow, it can break higher level logic, ending up causing infinite loops. 

Specifically, if an upstream caller passes in `NO_MORE_DOCS` to `bstractKnnVectorQuery.DocAndScoreQuery#advanceShallow` it will blindly add `context.docBase` and execute a binary search, resulting in, potentially a doc ID other than `NO_MORE_DOCS`.

This issue appeared in a rather complex query mixing kNN results with required clauses in a `ReqOptSumScorer`

### Version and environment details

Lucene v9

Bug does not exist in v10.

softwaredoug:
Not sure it matters given Lucene 10, but I vaguely remember Lucene having fairly robust tests for whether a query / scorer etc behaved. So this might be a good case to add a test for (pass NO_MORE_DOCS, expect to get back NO_MORE_DOCS).

benwtrent:
@softwaredoug for sure, I added a test case for that for vectors

benwtrent:
This has been patched in lucene 9.12.x and will be available if/when another bugfix is released from that branch.

PR #12146, opened 2023-02-13
Simplify max score for kNN vector queries

jtibshirani:
The helper class DocAndScoreQuery implements advanceShallow to help skip
non-competitive documents. This method doesn't actually keep track of where it
has advanced, so it might do extra work on each call.

Overall the complexity here didn't seem worth it, given the low cost of
collecting matching kNN docs. This PR switches to a simple approach, which uses
a fixed upper bound on the max score. This is low overhead, while still
allowing for skipping in some cases.

PR #14858, opened 2025-06-27
Ensure vector queries handle advanceShallow correctly

benwtrent:
closes: https://github.com/apache/lucene/issues/14857

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
@@ -437,6 +437,9 @@ abstract class AbstractKnnVectorQuery extends Query {
 
             @Override
             public int advanceShallow(int docid) {
+              if (docid == NO_MORE_DOCS) {
+                return NO_MORE_DOCS;
+              }
               int start = Math.max(upTo, lower);
               int docidIndex = Arrays.binarySearch(docs, start, upper, docid + context.docBase);
               if (docidIndex < 0) {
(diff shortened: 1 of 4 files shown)


lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -437,6 +437,9 @@ public float score() {
 
             @Override
             public int advanceShallow(int docid) {
+              if (docid == NO_MORE_DOCS) {
+                return NO_MORE_DOCS;
+              }
```
jpountz: Should it instead be something like below?

```java
if (docid >= context.reader.maxDoc()) { // out of range
  return NO_MORE_DOCS;
} 
```

Otherwise we may be computing blocks based on hits that belong to other segments?

jpountz (commented):
does `getMaxScore` need fixing as well? I see that it adds `context.docBase` too.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
benwtrent: Ah yeah! Let me get that

vigyasharma:
> does `getMaxScore` need fixing as well? I see that it adds `context.docBase` too.

I suppose this gets handled by the `idx < upper` check [on L427](https://github.com/apache/lucene/blob/branch_9_12/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L427), which would ensure we stop at segment boundary?

vigyasharma:
Sneaky bug! Thanks for fixing @benwtrent

benwtrent:
@jpountz what do you think if we just backport: https://github.com/apache/lucene/pull/12146 ?

That handles the issue and greatly simplifies things. I am actually not sure why it wasn't backported to begin with.

jpountz:
This looks like a safe change to backport to me.

benwtrent:
@jpountz @vigyasharma I changed this PR to just be a backport of https://github.com/apache/lucene/pull/12146

It fixes both maxscore & advshallow bugs while greatly simplifying the code.

jpountz (approved):

PR #14859, opened 2025-06-28
chore(java): fix typos

jbampton:
Just one typo fixed in many places

### DIFF
--- a/lucene/highlighter/src/test/org/apache/lucene/search/vectorhighlight/TestFastVectorHighlighter.java
+++ b/lucene/highlighter/src/test/org/apache/lucene/search/vectorhighlight/TestFastVectorHighlighter.java
@@ -70,7 +70,7 @@ public class TestFastVectorHighlighter extends LuceneTestCase {
     type.freeze();
     Field field =
         new Field(
-            "field", "This is a test where foo is highlighed and should be highlighted", type);
+            "field", "This is a test where foo is highlighted and should be highlighted", type);
 
     doc.add(field);
     writer.addDocument(doc);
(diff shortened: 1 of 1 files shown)


dweiss:
Thank you!

dweiss:
I had to revert this, unfortunately - the builds started to go crazy after this change. Something depends on this spelling mistake. Here is an example:

https://jenkins.thetaphi.de/job/Lucene-MMAPv2-Windows/2354/

dweiss:
If you can look into this again and correct this behavior - please file another PR. Thank you and sorry!

uschindler:
Did you figure out why this was not detected by the Github runners?

Actually all Jenkins builds fail. I was thinking it is maybe the length of the text string which causes the highlighter to produce a different fragment.

dweiss:
No, I'm on holidays and only occasionally checking email. It is surprising that it failed on jenkins but not on the CI - let me take a look.

dweiss:
Something is definitely not right because that test... failed in the CI but the build passed. I'm looking at it.

dweiss:
It's a build system regression that caused the CI to pass: #14944

dweiss:
I've reapplied this patch here:
https://github.com/apache/lucene/pull/14945

msokolov:
This is bewildering ... I guess there is some kind of megaflux capacitor quantum entanglement between  this typo fix and the halt-on-failure fix?

dweiss:
Good one. :) This patch showed a green CI status so I merged it in. But if you go and check the logs, it actually failed the test -
image

This can also be the case with other PRs against main, sadly. I switched the default on one of the test-task control flags while rewriting the code. Errare humanum est.

dweiss:
As for explaining why CI jobs on jenkins failed - this is due to the fact that Jenkins parses test output XMLs (I believe) and it detects test failures in addition to looking at the overall build status.

uschindler:
It also does not use all of the system properties you changed, but passes cmd line option "--continue" to gorefully proceed with all tasks.

/home/jenkins/workspace/Lucene-main-Linux/gradlew -Dtests.haltonfailure=false -Dtests.badapples=false --continue --stacktrace check

And yes, at end it finds test XML files and reports failures. There's also a difference. If exit code is non success build gets marked as failed, otherwise it gets yellow and reports only test failures (see above).

uschindler:
The latter is nice, because one can quickly see that compilation failed vs. only tests failed on the indicator.

msokolov:
ah thanks for explaining; we were passing builds even though there were failing tests, so the highlighting test typo fix PR looked green and we merged it, phew

PR #14862, opened 2025-06-29
Gradle plugins to java, batch 1

dweiss:
This is a small effort to move some of the trivial plugins to Java code. It should be faster to compile and apply, no functional changes are expected.

I'll port maybe two or three more, then apply.


build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/java/JavaFolderLayoutPlugin.java
```diff
@@ -0,0 +1,82 @@
...
+    SourceSetContainer sourceSets = javaExt.getSourceSets();
+
+    SourceSet mainSrcSet = sourceSets.named("main").get();
+    mainSrcSet.java(srcSetDir -> srcSetDir.setSrcDirs(List.of("src/java")));
+    mainSrcSet.resources(resources -> resources.setSrcDirs(List.of("src/resources")));
+
+    SourceSet testSrcSet = sourceSets.named("test").get();
+    testSrcSet.java(srcSetDir -> srcSetDir.setSrcDirs(List.of("src/test")));
+    testSrcSet.resources(resources -> resources.setSrcDirs(List.of("src/test-files")));
+
```
dweiss, line 46: This is way more verbose than groovy code but it has advantages - type checking among them.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/spotless/ParentGoogleJavaFormatTask.java
```diff
@@ -114,6 +114,9 @@ protected static String applyFormatter(Formatter formatter, String input)
     input = ImportOrderer.reorderImports(input, JavaFormatterOptions.Style.GOOGLE);
     input = RemoveUnusedImports.removeUnusedImports(input);
     input = formatter.formatSource(input);
+
+    // input = StringWrapper.wrap(input, formatter);
+
     return input;
```
dweiss, line 120: This would add long string reflowing. I didn't enable it because it entails reformatting quite a number of sources. Let's leave it for later to clean up.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/java/JavaFolderLayoutPlugin.java
rmuir: Thank you for doing this: personally, these .java files are much more "accessible", work in the editor with proper tooling support, etc.

dweiss: I agree. This has become such a black magic that anything that lowers the entry bar is worth the effort. Using java may be more verbose but at least it's understandable what's going on. Groovy is great but - as anywhere - dynamic properties and no types make it a headache to figure out what's going on.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/globals/LuceneBuildGlobalsExtension.java
```diff
@@ -0,0 +1,53 @@
...
+  public static final String NAME = "buildGlobals";
+
+  /** Base Lucene version ({@code x.y.z}). */
+  public String baseVersion;
+
+  /** Major Lucene version ({@code x} in {@code x.y.z}). */
+  public String majorVersion;
+
+  /** {@code true} if this build is a snapshot build. */
+  public boolean snapshotBuild;
```
dweiss, line 30: @breskeby do these need to be defined as proper properties (fields with accessors)? Or maybe full Property<type> fields (this seems like an overkill and a downer to me).

Just wondering.

dweiss:
I think this is large enough to allow merging. I've pulled some ext property refs to an extension and moved some convention plugins to java code. Like I said - it's more verbose but far less cryptic, especially if you're not used to groovy.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/globals/LuceneBuildGlobalsExtension.java
breskeby: these days, you would declare them via

```

  @Input
  abstract Property<String> getBaseVersion();


  @Input
  abstract Property<String> getMajorVersion();

  @Input
  abstract Property<Boolean> getSnapshotBuild();
```

breskeby: the benefit is the wiring the Provider API allows these days. especially in combination with task avoidance api.

dweiss: Yeah... I thought about that. But then everything (reading and writing those properties from Java code) bloats 2x. I'll never understand this design direction, I think. Eh.

breskeby: Especially when builds get bigger it used to be a huge hassle to track task properties and when they've been set if you relied on them in other places of the build. This usually ended up in a ordering problem of your plugin and build script code finding your code polluted with "afterEvaluate" and "taskgraph.whenReady" and all that type of hooks to defer execution. The idea of provider api is to be able to wire properties in your build without the need for know when and where a property has been actually configured making these deferred hooks needless. See https://docs.gradle.org/current/userguide/lazy_configuration.html#lazy_properties

dweiss: Oh, I understand the motivation and theory. I just don't think it helps. My experience says that moving everything to properties doesn't solve anything - it just moves the lazy evaluation to another place but often you have places which can't or don't accept lazy providers and then you need to access the property explicitly using get anyway. Thanks for explaining though.

PR #14867, opened 2025-06-29
build: download eclipse source jars and specify in .classpath

### LINKED ISSUE #14865: configure third-party javadocs via gradlew eclipse
### Description

Currently, gradlew eclipse creates a `.classpath` file, but it only contains jars of `kind="lib"`. 

This means there's e.g. no documentation for third-party libs. Not much of a problem considering lucene doesn't interact with third-party libraries, it only impacts a small percentage of the code.

But there's cases where we attach to third-party APIs and the improved editor support would really help: particularly the gradle api. The gradle tasks in java work almost perfectly with eclipse/vscode/vim/etc, except for the docs.

Last time I looked into this, my theory was that gradlew eclipse task actually adds this stuff, but we might be accidentally filtering it out with our logic. I'm gonna take another round at it.

rmuir:
hackishly applies "eclipse" plugin to all subprojects. this forces it to download the source jars, which doesn't happen today.

pokes around in the gradle cache to find corresponding source jars and specifies them in the .classpath properly.

Still needs to remove .classpath/.project/.settings that is being generated in each subproject as a side-effect.

Closes #14865

I need help as I spent hours trying to remove .classpath/.project/.settings from the subprojects, but can't figure it out. But if you run `gradlew eclipse` from this branch and then run `git clean -df` afterwards, it will take care of them and you will see that it works:

Screen_Shot_2025-06-29_at_11 37 15

### DIFF
--- a/gradlew
+++ b/gradlew
@@ -243,6 +243,8 @@ if [ ! -e "$APP_HOME/gradle.properties" ]; then
         exit $GENERATOR_STATUS
     fi
 fi
+# hack
+buildarg="$1"
 # END OF LUCENE CUSTOMIZATION
 
 # Collect all arguments for the java command:
@@ -289,4 +291,12 @@ eval "set -- $(
         tr '\n' ' '
     )" '"$@"'
 
-exec "$JAVACMD" "$@"
+if [[ "$buildarg" == "eclipse" ]]; then
+  "$JAVACMD" "$@"
+  # clean up unwanted .classpath/.project/.settings in subprojects
+  # i'm sure there's a better way, but various functionality of gradle such as doLast doesnt work
+  # use what works.
+  git clean -df
+else
+  exec "$JAVACMD" "$@"
+fi
--- a/build-tools/build-infra/src/main/groovy/lucene.ide.eclipse.gradle
+++ b/build-tools/build-infra/src/main/groovy/lucene.ide.eclipse.gradle
@@ -27,6 +27,11 @@ Provider<String> eclipseJavaVersionOption = buildOptions.addOption("eclipse.java
 Provider<String> eclipseErrorsOption = buildOptions.addOption("eclipse.errors", "Sets eclipse IDE lint level (ignore/warning/error)", "warning")
 
 if (gradle.startParameter.taskNames.contains("eclipse")) {
+  // force it to download source jars
+  subprojects {
+    apply plugin: 'eclipse'
+  }
+
   project.pluginManager.apply("java-base")
   project.pluginManager.apply("eclipse")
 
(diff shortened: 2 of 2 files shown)


dweiss:
Can't do it today, Rob - away from my computer. I'll try to see what it does tomorrow though. If I remember correctly, we used a single top-level Eclipse project, without separating subprojects. This one-blob approach was needed because things didn't work correctly back then. Maybe something has changed.

rmuir:
Thanks for looking, no rush. It's still the same approach. High level I just want `gradle eclipse` to download source jars and attach the location as XML attribute. 

Applying the plugin to all the subprojects was the only way I could find to force the source jars to be downloaded. Hopefully there's a better way, as it leaves a mess (unwanted .classpath, .settings, .project files in each subproject that we don't want)

rmuir:
I added a thermonuclear hack on top of the other hacks, at least so it has correct behavior.

All I want here is, when you run `./gradlew eclipse`, that it does exactly what it does today, and makes exactly the same files, except (example icu4j)

1. `icu4j-77.1-sources.jar` is downloaded to gradle cache.
2. `.classpath` entry contains `sourcepath` pointing to it:
```xml
<classpathentry
path="/home/rmuir/.gradle/caches/modules-2/files-2.1/com.ibm.icu/icu4j/77.1/38693cf0b1d7362a8b726af74dc06026a7c23809/icu4j-77.1.jar"
kind="lib"
sourcepath="/home/rmuir/.gradle/caches/modules-2/files-2.1/com.ibm.icu/icu4j/77.1/3816e175e91475575e91393835140f34a739dbc7/icu4j-77.1-sources.jar"/>
```

gradlew
```diff
@@ -289,4 +291,12 @@ eval "set -- $(
         tr '\n' ' '
     )" '"$@"'
 
-exec "$JAVACMD" "$@"
+if [[ "$buildarg" == "eclipse" ]]; then
+  "$JAVACMD" "$@"
+  # clean up unwanted .classpath/.project/.settings in subprojects
+  # i'm sure there's a better way, but various functionality of gradle such as doLast doesnt work
+  # use what works.
```
rmuir, line 298: I think some of this is related to #14868, the existing logic isn't even executing at all, which explains why I had so much trouble.

dweiss:
Take a look at this - https://github.com/apache/lucene/pull/14870 - maybe it can be achieved in a simpler way? There are still gotchas there but I think it reduces the amount of our code in favor of what's built-in into gradle eclipse plugin.

I didn't want to push it to this PR because it's a completely different take.

rmuir:
More than happy to decline this one! Thank you for looking at this.

PR #14870, opened 2025-06-30
An attempt to use gradle's eclipse plugin to provide us with src paths for libs

dweiss:
This is an alternative take to provide src paths for lib entries in Eclipse. It leverages gradle's build-in utilities -

https://docs.gradle.org/current/dsl/org.gradle.plugins.ide.eclipse.model.EclipseClasspath.html

I'm sure it doesn't follow gradle's conventions (we collect configurations from subprojects) - maybe @breskeby will know how to do it in a nicer way.

I also had to turn off split package checks in eclipse because the generated gradle api jar is full of jdk classes: https://github.com/gradle/gradle/issues/19577

The generated file seems identical to what we used to have and include src refs. When I open the project in eclipse it doesn't compile because of this - I'm not sure how to fix it because it seems like valid code:

image

### DIFF
--- a/build-tools/build-infra/src/main/groovy/lucene.ide.eclipse.gradle
+++ b/build-tools/build-infra/src/main/groovy/lucene.ide.eclipse.gradle
@@ -26,6 +26,15 @@ if (project != project.rootProject) {
 Provider<String> eclipseJavaVersionOption = buildOptions.addOption("eclipse.javaVersion", "Set Java version for Eclipse IDE", deps.versions.minJava.get())
 Provider<String> eclipseErrorsOption = buildOptions.addOption("eclipse.errors", "Sets eclipse IDE lint level (ignore/warning/error)", "warning")
 
+configurations {
+  foo
+}
+
+dependencies {
+  foo deps.zstd
+}
+
+
 if (gradle.startParameter.taskNames.contains("eclipse")) {
   project.pluginManager.apply("java-base")
   project.pluginManager.apply("eclipse")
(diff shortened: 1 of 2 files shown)


dweiss:
This must be some kind of Eclipse bug because editing IndexReaderContext and saving it again makes the bug report disappear.

rmuir:
@dweiss I don't think the error is new with this PR. I've always seen "Found 1 error" when it builds the project for some time now. 

I never looked into what the "1 error" is, because I don't use "workspace diagnostics" and look at all the errors across the project: just the file that I've got open.

Screen_Shot_2025-06-30_at_08 56 00

dweiss:
Ok. Let me know if this helps then - if it does, I can merge it in.

rmuir:
This PR works correctly for me! I did the full test:

```
$ git clean -xdf # nuke any repo state such as .classpath/.settings/.project
$ rm -rf ~/.local/share/jdtls # nuke any eclipse state
$ rm -rf ~/.gradle/caches # nuke any gradle state
$ ./gradlew eclipse
```

As a followup, I may see what it would take to get sources for that `gradle-api.jar`. It is an unrelated thing: this PR is good to go as-is, nothing wrong with the approach here. Problem may impact intellij, too. I think this is not a "normal" jar but autogenerated with fancier logic and requires jumping thru more hoops to get sources attached. But it would make working with our gradle sources in these editors much easier!

rmuir (approved):

dweiss:
> As a followup, I may see what it would take to get sources for that gradle-api.jar.

this thing is - I believe - generated dynamically. I don't think there will be a single "source" for it. Don't look inside or take some tranquilizing pills before you do.

dweiss:
I've merged this in. If we can improve it (eclipse's compilation failure, gradle sources, perhaps multi-module setup?), it's going to go in a follow-up issue.

PR #14880, opened 2025-07-01
Null exception occured when click on Luke desktop browser button

wuth:
Null exception occured when click on Luke desktop browser button 
[apache/lucene/issues/13345]


lucene/luke/src/java/org/apache/lucene/luke/app/desktop/components/dialog/menubar/OpenIndexDialogFactory.java
```diff
@@ -293,7 +293,11 @@ private File getLastOpenedDirectory() {
       if (!history.isEmpty()) {
         Path path = Paths.get(history.get(0));
         if (Files.exists(path)) {
-          return path.getParent().toAbsolutePath().toFile();
+          if(path.getParent() != null) {
+            return path.getParent().toAbsolutePath().toFile();
+          }else {
+            return path.toAbsolutePath().toFile();
+          }
```
dweiss: Run ./gradlew tidy to clean up code formatting. When is this situation possible, by the way (a file without a parent directory)?

dweiss (changes requested):
Please take a look at my comment.

lucene/luke/src/java/org/apache/lucene/luke/app/desktop/components/dialog/menubar/OpenIndexDialogFactory.java
wuth: Sorry ,I am a newer for github.  And I just commit the code which I had tested last year. So I have no local develop environment.  How can I Run ./gradlew tidy to clean up code formatting?   This situation occured when I open a directory from another window pc by share folder.

dweiss:
It's all explained here -
https://github.com/apache/lucene/blob/main/CONTRIBUTING.md

but I see you've managed to fix it, great. I am still curious how you managed to hit that code corner - will give it a try in a spare moment using windows shares.

wuth:
> It's all explained here - https://github.com/apache/lucene/blob/main/CONTRIBUTING.md
> 
> but I see you've managed to fix it, great. I am still curious how you managed to hit that code corner - will give it a try in a spare moment using windows shares.

Yes, I test to run ./gradlew tidy by adding a workflow yml directely on github web, but it didn't work.  Then I fixed the code formatting manually.  About this bug, at that time, I was short of local disk space, so I utilized the shared directory.

reproduce this problem :
1.Clear the content in .luke.d/history
2.Double click lucene-x.x.x\luke\luke.bat
3.Select an share folder like as \\xxx.xxx.x.x\luceneDataFolder
Then you would find you can not select another luceneDataFolder, nothing response on the select button being clicked.

dweiss:
Thank you. I could reproduce the issue.

PR #14884, opened 2025-07-01
tighten java-lang-import to avoid false-positive on subpackage wildcard

rmuir:
This is being a bit anal, but this rule shouldn't emit a false positive if you do a wildcard import from a subpackage of java.lang, e.g.

    import java.lang.classfile.*;

The issue is, it can be a bit confusing as you get two errors instead of just the `WildcardImport` error. Anchor it.

Found by testing rules on openjdk sources.

### DIFF
--- a/gradle/validation/ast-grep/rules/java-patterns.yml
+++ b/gradle/validation/ast-grep/rules/java-patterns.yml
@@ -5,7 +5,7 @@
 id: java-lang-import
 language: java
 rule:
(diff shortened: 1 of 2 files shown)


dweiss (approved):

rmuir:
I'll test future ones against openjdk too, should have done this to begin with: it is a great (larger) corpus for this kind of thing. I found it very useful when iterating through javadocs grammar.

PR #14885, opened 2025-07-01
WritableQueryIndex Relies On Hashcode Invariance Across JVMs

kotman12:
### Description

The full details can be found [here](https://github.com/apache/lucene/issues/14886).


lucene/core/src/java/org/apache/lucene/search/DisjunctionMaxQuery.java
```diff
@@ -85,6 +85,13 @@ public Collection<Query> getDisjuncts() {
     return Collections.unmodifiableCollection(disjuncts);
   }
 
+  /**
+   * @return the disjuncts in the same order as the input Query collection.
+   */
+  public Collection<Query> getDisjunctsInOriginalOrder() {
```
romseygeek: I think we can change `getDisjuncts()` to return things in order rather than adding an additional method here.  We don't make any guarantees about the ordering at present (and indeed we changed it before in https://issues.apache.org/jira/browse/LUCENE-9940).

lucene/monitor/src/test/org/apache/lucene/monitor/TestMonitorPersistence.java
```diff
@@ -96,4 +109,88 @@ public void testEphemeralMonitorDoesNotStoreQueries() throws IOException {
...
+
+    @Override
+    public boolean equals(Object o) {
+      if (!(o instanceof SimulateUpgradeQuery that)) return false;
+      return Objects.equals(innerQuery, that.innerQuery);
+    }
+
+    @Override
+    public int hashCode() {
+      return innerQuery.hashCode() * HASHCODE_FACTOR;
```
romseygeek, line 238: This is a nice test!

lucene/core/src/java/org/apache/lucene/search/DisjunctionMaxQuery.java
romseygeek: Maybe also add a test that asserts that the output of `getDisjuncts()` is ordered correctly so we don't inadvertently break this again (as I did in the above mentioned JIRA issue!).

romseygeek (changes requested):
Hi @kotman12, thanks for creating these fixes (and apologies for the delay in reviewing them).  I think I prefer this one as being less fragile than the other QueryVisitor option.

lucene/core/src/java/org/apache/lucene/search/DisjunctionMaxQuery.java
kotman12: Thanks for the feedback! I think I addressed your comments with the latest commit by removing this method and adding a check to the renamed `testCasesWhenDisjunctOrderMatters`. I've also beefed up `TestQueryDecomposer` to validate order since the `QueryDecomposer` is especially sensitive to this behavior. In theory you could also make sure any Query parsers we know about don't break this upstream of the Disjunction/BooleanQueries but in practice this is not maintainable.

dweiss:
> This PR does not have an entry in lucene/CHANGES.txt. Consider adding one.

Please, do. Seems like a good way to credit your work, @kotman12 .

lucene/CHANGES.txt
```diff
@@ -281,6 +281,8 @@ Bug Fixes
 
 * GITHUB#15074: Fix to use READONCE when reading segment infos (Chris Hegarty)
 
+* GITHUB#14886: Fix for lucene-monitor to prevent missed matches caused by inconsistent cache and index.
```
romseygeek: Add your name here too so we know who to credit? :)

romseygeek: Maybe leave a bit more detail around what the actual fix is, so that people reading CHANGES can know if this is likely to affect them?  Something like:

```
Ensure iteration of disjunctions from DisjunctionMaxQuery is stable.  This fixes a 
bug in the monitor where matching branches of a query could be wrongly ignored
if the QueryIndex was built by a different JVM.
```

kotman12: > Add your name here too so we know who to credit? :)

Woops, how did I miss that?!

> Maybe leave a bit more detail

I wanted to stress the impact in case someone searched for "missed matches" in the repo, they could hopefully  find the fix in 10.X. But I'll add more detail on "how" 😃

romseygeek:
Merged and backported to 10x - thanks @kotman12!

kotman12:
thanks @romseygeek! I see sadly the build failed ... I don't think it makes a big difference but I made it an instance member to achieve the same effect to avoid noise https://github.com/apache/lucene/pull/15087

Issue #11761, opened 2022-09-09
Expand TieredMergePolicy deletePctAllowed limits


mdmarshmallow:
### Description

I'm an engineer at Amazon Search and we have been experimenting with more aggressively getting rid of deleted documents. We use TieredMergePolicy and we would like to set `TieredMergePolicy#deletesPctAllowed` to be [lower than the current limit of 20%](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/TieredMergePolicy.java#L157). 

I was wondering why this limit was set in place. I'm sure I could be missing some context here. Maybe we could keep the limits in place but allow users to explicitly remove the checks? Any information would be much appreciated, thanks!

jpountz:
Historically this was not configurable and Lucene would allow up to 50% deleted documents. When we introduced an option, we made sure to introduce a lower bound on the value because a value of zero would essentially require Lucene to rewrite every segment that has a deletion after every update operation, which is certainly undesirable. Allowing users to go from 50% to 20% felt like a significant improvement already.

We could discuss lowering the limit if we feel like this could lead to merging patterns that are still acceptable. E.g. I used `BaseMergePolicyTestCase#doTestSimulateUpdates` in the past to get a sense of how this option would influence write amplification.

mdmarshmallow:
Hi, thanks for the response! Your explanation of 0% not being allowed makes complete sense. For some context though, using our own forked version of `TieredMergePolicy`, we have tested with down to 2% allowable deletion and still see that behavior is desirable for us (more specifically much lower index sizes than at 20% deletes).

Maybe if we want to maintain those limits, we could create a `setDeletesPctAllowedUnsafe()` or something like that which has no limits on it (or at least drops the lower bound to being > 0 instead of > 20)?

jpountz:
I got some numbers for write amplification for the case tested in `TestTieredMergePolicy#testSimulateUpdates`:

| Allowed percentage of deletes | Write amplification |
| ------------- | ------------- |
| 50 (max) | 4.34 |
| 33 (default) | 4.34 |
| 20 (min) | 4.68 |
| 10 | 6.13 |
| 5 | 8.76 |
| 4 | 10.31 |
| 3 | 12.97 |
| 2 | 18.76 |
| 1 | 37.89 |
| 0 | 10779.78 |

[...]

PR #14893, opened 2025-07-02
Decrease minimum deletes percentage in TMP

stefanvodita:
### Description

This will allow users to set lower deletes percentages on their merge policies, which can be used to create indexes that are more efficient to search.

Previous discussion: #11761


lucene/core/src/java/org/apache/lucene/index/TieredMergePolicy.java
```diff
@@ -139,9 +139,9 @@ public double getMaxMergedSegmentMB() {
    * amplification factor will lead to higher CPU and I/O activity as indicated above.
    */
   public TieredMergePolicy setDeletesPctAllowed(double v) {
-    if (v < 5 || v > 50) {
+    if (v <= 0 || v > 50) {
```
msokolov, line 149: I like that we still enforce > 0. It makes it more apparent that you are approaching a singular pathological case as you go to 0.001 or whatever

lucene/core/src/java/org/apache/lucene/index/TieredMergePolicy.java
```diff
@@ -130,7 +130,7 @@ public double getMaxMergedSegmentMB() {
   /**
    * Sets the maximum percentage of doc id space taken by deleted docs. The denominator includes
    * both active and deleted documents. Lower values make the index more space efficient at the
-   * expense of increased CPU and I/O activity. Values must be between 5 and 50. Default value is
+   * expense of increased CPU and I/O activity. Values must be between 0 and 50. Default value is
```
mikemccand, line 133: Can we enhance this javadoc to note the dangers of very low (< 5%) target deletions?  Something like:

```
Values below 5% can lead to exceptionally high merge cost where indexing will continuously
merge nearly all segments, and select newly merged segments immediately for merging again,
often forcing degenerate merge selection like singleton merges.  If you venture into this dark
forest, consider limiting the maximum number of concurrent merges and threads (link to
ConcurrentMergeScheduler's setMaxMergesAndThreads) as a coarse attempt to bound the
otherwise pathological indexing behavior.
```

mikemccand (commented):
I think it's OK to take the training wheels off here, but, we should at least add a stern warning.  Values below 5% are really quite insane -- O(N^2) merging cost, where the future cost of merging is `O(N^2)` when the initial indexing cost was `N`.

lucene/core/src/java/org/apache/lucene/index/TieredMergePolicy.java
stefanvodita: Thank you for the suggestion @mikemccand! I added the paragraph almost exactly as you wrote it.

stefanvodita:
Thank you for the reviews! Should we try to get this finished before 10.3? Sounds like that will be in early September.

mikemccand (approved):
Thanks @stefanvodita -- this removes the training wheels but notes the resulting dangers well.  +1 to merge now (for 10.3.0).

PR #14897, opened 2025-07-03
Moved lucene.java.tests-and-randomization.gradle and more scripts to java

dweiss:
- Convert the related lucene.java.beasting.gradle too.
- remove root ext props (rootSeedUserProvided) and use build globals extension.


dweiss:
I think this is already large enough and fairly self-isolated. I'd appreciate if you could take a look, @breskeby, although I'm sure it's not light reading.

dweiss:
I think this is already large enough and fairly self-isolated. I'd appreciate if you could take a look, @breskeby, although I'm sure it's not light reading.

dweiss:
I think this is already large enough and fairly self-isolated. I'd appreciate if you could take a look, @breskeby, although I'm sure it's not light reading.

build-tools/build-infra/src/main/groovy/lucene.documentation.render-javadoc.gradle
```diff
@@ -329,16 +330,16 @@ class RenderJavadocTask extends DefaultTask {
...
-  @Input
-  final Property<String> executable = project.objects.property(String).convention(
-  project.provider { Jvm.current().javadocExecutable.toString() })
-
   @InputDirectory
   @PathSensitive(PathSensitivity.RELATIVE)
   @IgnoreEmptyDirectories
   File taskResources
 
+  @Inject
```
breskeby: I think you can remove this and just use getProject() when you need a project reference

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/LuceneGradlePlugin.java
```diff
@@ -62,4 +63,24 @@ protected void applicableToRootProjectOnly(Project project) {
           "This plugin is applicable to the rootProject only: " + getClass().getSimpleName());
     }
   }
+
+  /**
+   * Returns a filesystem path to a given resource that the plugin uses. At the moment, these
+   * resources are located under the top-level {@code gradle/} folder.
+   */
+  protected Path gradlePluginResource(Project project, String relativePath) {
+    return project
```
breskeby, line 72: Instead of project.rootProject.layout.projectDir you can use project.layout. getSettingsDirectory() 

This avoids crossreferencing root project here.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/java/AlternativeJdkSupportPlugin.java
```diff
@@ -0,0 +1,176 @@
...
+            return String.format(
+                Locale.ROOT,
+                "%s (%s %s, home at: %s)",
+                jvmMetadata.getLanguageVersion(),
+                jvmMetadata.getDisplayName(),
+                jvmMetadata.getRuntimeVersion(),
+                jvmMetadata.getJavaHome());
+          };
+
+      project
```
breskeby, line 115: Not sure I would create a task just for a logger message

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/java/AlternativeJdkSupportPlugin.java
```diff
@@ -0,0 +1,176 @@
...
+
+    var altJvmExt = project.getRootProject().getExtensions().getByType(AltJvmExtension.class);
+    if (altJvmExt.getAltJvmUsed().get()) {
+      // Set up custom JVM for tests
+      project
+          .getTasks()
+          .withType(Test.class)
+          .configureEach(
+              task -> {
+                task.dependsOn(":altJvmWarning");
```
breskeby, line 147: Can't you just put the logging here?

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/java/RenderJavadocTaskBase.java
```diff
@@ -0,0 +1,39 @@
...
+public abstract class RenderJavadocTaskBase extends DefaultTask {
+  @Optional
+  @Input
+  public abstract Property<String> getExecutable();
+
+  @Inject
+  public RenderJavadocTaskBase(Project project) {
+    getExecutable()
+        .convention(
+            project.getProviders().provider(() -> Jvm.current().getJavadocExecutable().toString()));
```
breskeby: s said above. You should be able to just use getProject() without the need of injecting project via the ctor

breskeby (commented):
Just a few minor remarks. I think k this makes things definitely better.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/LuceneGradlePlugin.java
dweiss: I do remember your comment from before but getSettingsDirectory pops up as "unstable" API though... is it ok?

build-tools/build-infra/src/main/groovy/lucene.documentation.render-javadoc.gradle
dweiss: Done.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/java/AlternativeJdkSupportPlugin.java
dweiss: The reason for it is that I only want to get it printed once. I couldn't figure out any other clean way of doing it. Also, if you log during configuration time, the log message is stripped of the "context" (task where it comes from) and I think it's nice to be able to track it back to something more concrete. If you know how to implement this in a different way, let me know.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/java/AlternativeJdkSupportPlugin.java
dweiss: Explained above - it'd print the log message multiple times. Not ideal, I think.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/LuceneGradlePlugin.java
dweiss: I've changed it to getSettingsDirectory, even though it's marked as unstable API (incubating).https://docs.gradle.org/current/javadoc/org/gradle/api/file/ProjectLayout.html#getSettingsDirectory()

dweiss:
I am going to merge this one in, but I'd love to see what you have to say with regard to the comments above. If there is a better way to achieve the goal (one-time log message), I'll follow up.

rmuir:
Thank you for another of these: I just had to mod these sources on another PR and it is really nice to do it with more confidence, compile-time checks, etc.

dweiss:
I agree. It's more verbose but much easier to write/ modify/ understand. Happy 4th of July!

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/java/AlternativeJdkSupportPlugin.java
breskeby: Ultimately the way for those loggings would be to use the problems api noawadays but again that's an overload to just introduce this for this warning. I have used that while working on the license checks pr. I'd suggest to leave it as is for now and we can look into leverage the problems api at another time Dir more placing to have a central place for these warnings

dweiss: ok

PR #14898, opened 2025-07-04
Fix typo in Circle2D.java

nehemiaharchives:
### Description

Fixes geX() to getX()

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -242,6 +242,7 @@ Other
 
 * GITHUB#14667: Fix documentation regarding benchmark running. (Viliam Durina)
 
+* GITHUB#14898: Fix typo in Circle2D.java (Hokuto Joel Ide)
 
 ======================= Lucene 10.2.2 =======================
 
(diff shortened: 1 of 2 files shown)


vigyasharma (approved):
Thanks for fixing this! The change looks good, and being only used in private interfaces, we don't have to worry about API changes.

Would you like to an entry in CHANGES.txt for this? It can go into 10.3 under the "Other" section?

nehemiaharchives:
@vigyasharma 
>CHANGES.txt for this

Oh, yes if its ok. Should I add an entry in CHANGES.txt ?

nehemiaharchives:
@vigyasharma By the way, I found another typo. 

1) Should I commit it here?
2) Should I create another PR?

mikemccand:
> @vigyasharma
> 
> > CHANGES.txt for this
> 
> Oh, yes if its ok. Should I add an entry in CHANGES.txt ?

+1 -- under the Other section for 10.3 makes sense?

Thank you for the attention to detail!

PR #14901, opened 2025-07-06
remove HAS_FAST_INTEGER_VECTORS checks from panama support.

rmuir:
For any integer vectorization code, the developer must remember to make this check, or suffer a 10x+ slowdown if AVX2 is unavailable. This can happen in virtual environments (default QEMU, virtualbox, etc).

It isn't worth the benefit of supporting floating point vectors on such machines, just remove this trap completely.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/internal/vectorization/VectorizationProvider.java
+++ b/lucene/core/src/java/org/apache/lucene/internal/vectorization/VectorizationProvider.java
@@ -46,7 +46,6 @@ import org.apache.lucene.util.VectorUtil;
 public abstract class VectorizationProvider {
 
   static final OptionalInt TESTS_VECTOR_SIZE;
-  static final boolean TESTS_FORCE_INTEGER_VECTORS;
   static final int UPPER_JAVA_FEATURE_VERSION = getUpperJavaFeatureVersion();
 
   static {
@@ -63,16 +62,6 @@ public abstract class VectorizationProvider {
       // ignored
     }
     TESTS_VECTOR_SIZE = vs;
-
-    boolean enforce = false;
-    try {
-      enforce = Boolean.getBoolean("tests.forceintegervectors");
-    } catch (
-        @SuppressWarnings("unused")
-        SecurityException se) {
-      // ignored
-    }
-    TESTS_FORCE_INTEGER_VECTORS = enforce;
   }
 
   private static final String UPPER_JAVA_FEATURE_VERSION_SYSPROP =
@@ -157,9 +146,9 @@ public abstract class VectorizationProvider {
       vectorMod.ifPresent(VectorizationProvider.class.getModule()::addReads);
       // check for testMode and otherwise fallback to default if slowness could happen
       if (!testMode) {
-        if (TESTS_VECTOR_SIZE.isPresent() || TESTS_FORCE_INTEGER_VECTORS) {
+        if (TESTS_VECTOR_SIZE.isPresent()) {
           LOG.warning(
-              "Vector bitsize and/or integer vectors enforcement; using default vectorization provider outside of testMode");
+              "Vector bitsize enforcement; using default vectorization provider outside of testMode");
           return new DefaultVectorizationProvider();
         }
         if (Constants.IS_CLIENT_VM) {
--- a/lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorConstants.java
+++ b/lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorConstants.java
@@ -42,8 +42,7 @@ final class PanamaVectorConstants {
     // to be fair, they do document this thing only works well with AVX2/AVX3 and Neon
     boolean isAMD64withoutAVX2 =
         Constants.OS_ARCH.equals("amd64") && PREFERRED_VECTOR_BITSIZE < 256;
-    HAS_FAST_INTEGER_VECTORS =
-        VectorizationProvider.TESTS_FORCE_INTEGER_VECTORS || (isAMD64withoutAVX2 == false);
+    HAS_FAST_INTEGER_VECTORS = isAMD64withoutAVX2 == false;
 
     PRERERRED_LONG_SPECIES =
         VectorSpecies.of(long.class, VectorShape.forBitSize(PREFERRED_VECTOR_BITSIZE));
--- a/lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorConstants.java
+++ b/lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorConstants.java
@@ -29,7 +29,6 @@ final class PanamaVectorConstants {
   /** Whether integer vectors can be trusted to actually be fast. */
   static final boolean ENABLE_INTEGER_VECTORS;
 
-  static final VectorSpecies<Long> PRERERRED_LONG_SPECIES;
   static final VectorSpecies<Integer> PRERERRED_INT_SPECIES;
 
   static {
@@ -45,8 +44,6 @@ final class PanamaVectorConstants {
     ENABLE_INTEGER_VECTORS =
         (isAMD64withoutAVX2 == false) || VectorizationProvider.TESTS_VECTOR_SIZE.isPresent();
 
-    PRERERRED_LONG_SPECIES =
-        VectorSpecies.of(long.class, VectorShape.forBitSize(PREFERRED_VECTOR_BITSIZE));
     PRERERRED_INT_SPECIES =
         VectorSpecies.of(int.class, VectorShape.forBitSize(PREFERRED_VECTOR_BITSIZE));
   }
--- a/lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorConstants.java
+++ b/lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorConstants.java
@@ -27,7 +27,7 @@ final class PanamaVectorConstants {
   static final int PREFERRED_VECTOR_BITSIZE;
 
   /** Whether integer vectors can be trusted to actually be fast. */
-  static final boolean HAS_FAST_INTEGER_VECTORS;
+  static final boolean ENABLE_INTEGER_VECTORS;
 
   static final VectorSpecies<Long> PRERERRED_LONG_SPECIES;
   static final VectorSpecies<Integer> PRERERRED_INT_SPECIES;
@@ -42,7 +42,8 @@ final class PanamaVectorConstants {
     // to be fair, they do document this thing only works well with AVX2/AVX3 and Neon
     boolean isAMD64withoutAVX2 =
         Constants.OS_ARCH.equals("amd64") && PREFERRED_VECTOR_BITSIZE < 256;
-    HAS_FAST_INTEGER_VECTORS = isAMD64withoutAVX2 == false;
+    ENABLE_INTEGER_VECTORS =
+        (isAMD64withoutAVX2 == false) || VectorizationProvider.TESTS_VECTOR_SIZE.isPresent();
 
     PRERERRED_LONG_SPECIES =
         VectorSpecies.of(long.class, VectorShape.forBitSize(PREFERRED_VECTOR_BITSIZE));
--- a/lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorizationProvider.java
+++ b/lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorizationProvider.java
@@ -45,6 +45,10 @@ final class PanamaVectorizationProvider extends VectorizationProvider {
           "Vector bit size is less than 128: " + PanamaVectorConstants.PREFERRED_VECTOR_BITSIZE);
     }
 
+    if (PanamaVectorConstants.HAS_FAST_INTEGER_VECTORS == false) {
+      throw new UnsupportedOperationException("No integer vector support");
+    }
+
     this.vectorUtilSupport = new PanamaVectorUtilSupport();
 
     var log = Logger.getLogger(getClass().getName());
@@ -53,8 +57,7 @@ final class PanamaVectorizationProvider extends VectorizationProvider {
             Locale.ENGLISH,
             "Java vector incubator API enabled; uses preferredBitSize=%d%s%s",
             PanamaVectorConstants.PREFERRED_VECTOR_BITSIZE,
-            Constants.HAS_FAST_VECTOR_FMA ? "; FMA enabled" : "",
-            PanamaVectorConstants.HAS_FAST_INTEGER_VECTORS ? "" : "; floating-point vectors only"));
+            Constants.HAS_FAST_VECTOR_FMA ? "; FMA enabled" : ""));
   }
 
   @Override
--- a/lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorizationProvider.java
+++ b/lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorizationProvider.java
@@ -45,8 +45,9 @@ final class PanamaVectorizationProvider extends VectorizationProvider {
           "Vector bit size is less than 128: " + PanamaVectorConstants.PREFERRED_VECTOR_BITSIZE);
     }
 
-    if (PanamaVectorConstants.HAS_FAST_INTEGER_VECTORS == false) {
-      throw new UnsupportedOperationException("No integer vector support");
+    if (PanamaVectorConstants.ENABLE_INTEGER_VECTORS == false) {
+      throw new UnsupportedOperationException(
+          "CPU type or flags do not guarantee support for fast integer vectorization");
     }
 
     this.vectorUtilSupport = new PanamaVectorUtilSupport();
@@ -57,7 +58,8 @@ final class PanamaVectorizationProvider extends VectorizationProvider {
             Locale.ENGLISH,
             "Java vector incubator API enabled; uses preferredBitSize=%d%s%s",
             PanamaVectorConstants.PREFERRED_VECTOR_BITSIZE,
-            Constants.HAS_FAST_VECTOR_FMA ? "; FMA enabled" : ""));
+            Constants.HAS_FAST_VECTOR_FMA ? "; FMA enabled" : "",
+            VectorizationProvider.TESTS_VECTOR_SIZE.isPresent() ? "; testMode enabled" : ""));
   }
 
   @Override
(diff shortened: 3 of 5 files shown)


rmuir:
@uschindler this one may impact jenkins randomization (simplify it, I think).

ChrisHegarty (approved):
LGTM

jpountz (approved):
I like the simplification.

msokolov:
so now we throw an exception instead to guard the unwary from falling into the trap? Makes sense to me.

rmuir:
@msokolov previously, on x86 machines without AVX2 (e.g. AVX1 or SSSE), the vector API was still "enabled". But it is a trap, as many operations are not supported, and run 10x slower (or more).

With this change, we just disable vectors completely on such setups (floating point too). It means developers don't need to guard all the integer functions with `if (HAS_FAST_INTEGER_VECTORS)`. 

Goal is to remove the trap: these days you are most likely to encounter such a setup via virtualization, that isn't setup in a fully optimized way (e.g. casual desktop user with virtualbox or qemu or something).

lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorizationProvider.java
```diff
@@ -53,8 +57,7 @@ final class PanamaVectorizationProvider extends VectorizationProvider {
             Locale.ENGLISH,
             "Java vector incubator API enabled; uses preferredBitSize=%d%s%s",
```
uschindler, line 59: I think the second `%s` needs to go away? Have you checked what's logged?

rmuir: @uschindler THANK YOU. Totally missed this, I fear I've become lazy because I've gotten too accustomed to having `-Wformat` or equivalent checks in all other languages...

rmuir:
The bug Uwe found is truly crazy. Need to step back here and fix SOMETHING to fail on it...

No java linter finds the issue at compile time, which is really surprising to me. With clang the problem is caught immediately by `-Wformat-insufficient-args`, it is the kind of silly mistake I'd expect to be caught at compile time.

Screen_Shot_2025-07-07_at_09 15 24

At runtime, the test passes, UNLESS you run it with `-Dtests.verbose`, then it fails.

Screen_Shot_2025-07-07_at_09 20 49

In the short-term, I think this is the easiest to address. I know we don't want tests to print but maybe we can hack the test so it fails without `-Dtests.verbose`

rmuir:
Theres a subfolder of related checks in error-prone to investigate: https://github.com/google/error-prone/tree/d17e312d982badf2070404dde11b88c2538a5222/core/src/main/java/com/google/errorprone/bugpatterns/formatstring

Need to reaudit all the available checks anyway (as clearly these were missed by me before), but this might be a good compile-time way to find it.

uschindler:
I think, this chekc can now also be removed:
https://github.com/apache/lucene/pull/14901/files#diff-f58a1168777f5e9afd4c4bb868d74c619abafd0ed8a03065823d2ce1727ed43aR75

uschindler:
> At runtime, the test passes, UNLESS you run it with `-Dtests.verbose`, then it fails.

Actually it doe snot throw the exception visible to the enduser. Just the provider interface's ctor throws it so the SPI implementation falls back to default implementation. This disables Panama Vectors completely.

Previously the downstream code in the Panama SPI had to do the constant check at many places.

uschindler:
> At runtime, the test passes, UNLESS you run it with `-Dtests.verbose`, then it fails.

I still don't understand why it does not fail the tests. The info loggig is always executed. Maybe it fails before due to randomization sometimes?

We want to test Panama Vector code also on non-supported hardware, so it creates a "Test instance" of the provider which is slow, but not on production.

I will check close later, just update code first. It is a good thing to remove complexity for "seldom" CPU flag combinations. Actually in KVM w/ QEMU (when using libvirt) it never disables if you have selected CPU type "Host Passthrough" (like Policeman Jenkins).

uschindler:
> @uschindler this one may impact jenkins randomization (simplify it, I think).

Should not affect Jenkins, we don't randomize that flag, it is/was done by Gradle itsself.

dweiss:
> At runtime, the test passes, UNLESS you run it with `-Dtests.verbose`, then it fails.

What's the repro line for this? I don't think this should matter (gradle-wise); LuceneTestCase uses this flag - to set the verbose constant, maybe something depends on this.
```
public static final boolean VERBOSE = systemPropertyAsBoolean("tests.verbose", false);
```

lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorizationProvider.java
```diff
@@ -45,6 +45,10 @@ final class PanamaVectorizationProvider extends VectorizationProvider {
           "Vector bit size is less than 128: " + PanamaVectorConstants.PREFERRED_VECTOR_BITSIZE);
     }
 
+    if (PanamaVectorConstants.HAS_FAST_INTEGER_VECTORS == false) {
+      throw new UnsupportedOperationException("No integer vector support");
```
uschindler: I'd change this message a bit more verbose as this is logged to enduser in a warning (see `VectorizationProvider` around line 169).

Maybe "CPU flags do not guarantee fast integer vectors".

uschindler:
> > At runtime, the test passes, UNLESS you run it with `-Dtests.verbose`, then it fails.
> 
> What's the repro line for this? I don't think this should matter (gradle-wise); LuceneTestCase uses this flag - to set the verbose constant, maybe something depends on this.
> 
> ```
> public static final boolean VERBOSE = systemPropertyAsBoolean("tests.verbose", false);
> ```

I think the test should fail due to the IllegalArgumentException anyways. There is some stupid bullshit going on. Verbose or not should both create the same result.

As said before, I think this did not fail on the github runner, because the virtualization does not enable panama at all. And a "test only" enforcement of bitsize is done randomly, so it won't fail all the time.

The problem with the above `String#format()` is that it is only executed if everything is sane with your CPU/virtualization and Hotspot flags.

uschindler:
Hi for me the test failed without verbose:

```
$ gradlew :lucene:core:test --tests TestVectorUtil

> Task :lucene:core:test

TestVectorUtil > classMethod FAILED
    java.util.MissingFormatArgumentException: Format specifier '%s'
        at java.base/java.util.Formatter.format(Formatter.java:2760)
        at java.base/java.util.Formatter.format(Formatter.java:2698)
        at java.base/java.lang.String.format(String.java:4509)
        at org.apache.lucene.internal.vectorization.PanamaVectorizationProvider.<init>(PanamaVectorizationProvider.java:56)
        at org.apache.lucene.internal.vectorization.VectorizationProvider.lookup(VectorizationProvider.java:168)
        at org.apache.lucene.internal.vectorization.BaseVectorizationTestCase.maybePanamaProvider(BaseVectorizationTestCase.java:39)
        at org.apache.lucene.internal.vectorization.BaseVectorizationTestCase.<clinit>(BaseVectorizationTestCase.java:25)
        at org.apache.lucene.util.TestVectorUtil.<clinit>(TestVectorUtil.java:404)
        at java.base/java.lang.Class.forName0(Native Method)
        at java.base/java.lang.Class.forName(Class.java:543)
        at com.carrotsearch.randomizedtesting.RandomizedRunner$2.run(RandomizedRunner.java:631)

org.apache.lucene.util.TestVectorUtil > test suite's output saved to C:\Users\Uwe Schindler\Projects\lucene\lucene\lucene\core\build\test-results\test\outputs\OUTPUT-org.apache.lucene.util.TestVectorUtil.txt, copied below:
   >     java.util.MissingFormatArgumentException: Format specifier '%s'
   >         at java.base/java.util.Formatter.format(Formatter.java:2760)
   >         at java.base/java.util.Formatter.format(Formatter.java:2698)
   >         at java.base/java.lang.String.format(String.java:4509)
   >         at org.apache.lucene.internal.vectorization.PanamaVectorizationProvider.<init>(PanamaVectorizationProvider.java:56)
   >         at org.apache.lucene.internal.vectorization.VectorizationProvider.lookup(VectorizationProvider.java:168)
   >         at org.apache.lucene.internal.vectorization.BaseVectorizationTestCase.maybePanamaProvider(BaseVectorizationTestCase.java:39)
   >         at org.apache.lucene.internal.vectorization.BaseVectorizationTestCase.<clinit>(BaseVectorizationTestCase.java:25)
   >         at org.apache.lucene.util.TestVectorUtil.<clinit>(TestVectorUtil.java:404)
   >         at java.base/java.lang.Class.forName0(Native Method)
   >         at java.base/java.lang.Class.forName(Class.java:543)
   >         at com.carrotsearch.randomizedtesting.RandomizedRunner$2.run(RandomizedRunner.java:631)

:lucene:core:test (FAILURE): 1 test(s), 1 failure(s)

**Comment by dweiss** [comment]
> I think the test should fail due to the IllegalArgumentException anyways. There is some stupid bullshit going on. Verbose or not should both create the same result.

It's not about tests.verbose - it's about the seed. This code is either hit or not, depending on the main tests.seed - some other randomized, derived, parameters are affecting it.

**Comment by dweiss** [comment]
This:
```
    // hotspot misses some SSE intrinsics, workaround it
    // to be fair, they do document this thing only works well with AVX2/AVX3 and Neon
    boolean isAMD64withoutAVX2 =
        Constants.OS_ARCH.equals("amd64") && PREFERRED_VECTOR_BITSIZE < 256;
    HAS_FAST_INTEGER_VECTORS = isAMD64withoutAVX2 == false;
```
the vector size is randomized from this pool:
```
    String randomVectorSize =
        RandomPicks.randomFrom(
            new Random(buildGlobals.getProjectSeedAsLong().get()),
            List.of("default", "128", "256", "512"));
```
if it's 128, that code path is never reached.

**Comment by uschindler** [comment]
> > I think the test should fail due to the IllegalArgumentException anyways. There is some stupid bullshit going on. Verbose or not should both create the same result.
> 
> It's not about tests.verbose - it's about the seed. This code is either hit or not, depending on the main tests.seed - some other randomized, derived, parameters are affecting it.

Exactly! When running it multiple times it fails 1/3 of it.... If the randomly assigned vector bitsize does not match expectations it fails.

The problem with this PR is: We don't test the integer vectors anymore because Robert removed the "forceIntegerVectors" flag. I think it should stay alive (so please revert removal of the sysprop).

**Comment by dweiss** [comment]
So this will never fail:
```
./gradlew -p lucene/core test  --tests TestVectorUtil -Ptests.seed=6EA802BC0CAA73E2
```
but this will always fail:
```
./gradlew -p lucene/core test  --tests TestVectorUtil -Ptests.seed=6EA802BC0CAA73E2 -Ptests.vectorsize=256
```

**Comment by uschindler** [comment]
Ok, so the vectorSize is missing in the test reproducer line.

Anyways, thats an old issue. My problem with Robert's patch is that he removed the second system property `tests.forceintegervectors`. This allowed us to test integer vectors correctness.

To keep the sysprop away, my proposal would be: **Whenever somebody sets the "bitsize" we should NOT bail out and execute the tests in TestVectorUtil.**

**Comment by uschindler** [comment]
We should only apply this PR on main branch, because I don't want people to get sudden slowdowns in 10.x branch.

**Comment by rmuir** [comment]
> We should only apply this PR on main branch, because I don't want people to get sudden slowdowns in 10.x branch.

Who would get a slowdown? If they are impacted by this PR, they've probably got a 10x slowdown happening somewhere already :)

I think the ongoing risk would be, that developer adds a new function to PanamaVectorUtilSupport.java, backports it, and then it causes a new 10x-20x slowdown.

That's the motivation behind this PR anyway. It shouldn't slow anyone down, just requires AVX2 on amd64 machines before we try to vectorize. IMO best to disable vectorization completely in such cases, it will actually prevent a 10x slowdown. "supporting" some SSSE-based 2x speedup for a couple floating point methods isn't worth it?

File: lucene/core/src/java/org/apache/lucene/internal/vectorization/VectorizationProvider.java [hunk 3]
```diff
@@ -63,16 +62,6 @@ public abstract class VectorizationProvider {
       // ignored
     }
     TESTS_VECTOR_SIZE = vs;
-
-    boolean enforce = false;
```
uschindler, line 67: I would keep this in, because it allows us to still test correctness of the integer and float vectors. The TestVeczorUtil test enforces a fixed bitsize to test all variants of our code (also the slow ones on the test-runner CPU).

As we no bail out completely if we have no fast integer vectors, we no longer test it on all CPUs.

So we should keep the boolean or alternatively use `TESTS_VECTOR_SIZE.isPresent()` to check for test mode like:

```java
TESTS_FORCE_INTEGER_VECTORS = TESTS_VECTOR_SIZE.isPresent();
```

The complete removal of any testing is not a good idea.

lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorConstants.java
```diff
@@ -42,8 +42,7 @@ final class PanamaVectorConstants {
     // to be fair, they do document this thing only works well with AVX2/AVX3 and Neon
     boolean isAMD64withoutAVX2 =
         Constants.OS_ARCH.equals("amd64") && PREFERRED_VECTOR_BITSIZE < 256;
-    HAS_FAST_INTEGER_VECTORS =
-        VectorizationProvider.TESTS_FORCE_INTEGER_VECTORS || (isAMD64withoutAVX2 == false);
+    HAS_FAST_INTEGER_VECTORS = isAMD64withoutAVX2 == false;
```
uschindler: See above and let's do it like that:

```java
HAS_FAST_INTEGER_VECTORS = isAMD64withoutAVX2 == false || TESTS_VECTOR_SIZE.isPresent()

uschindler: In the special test mode I'd like to test our code, although CPU is wrong.

rmuir: if we can better contain this constant we can also give it a better name. The idea is: if you are on x86, you need AVX2 as a minimum baseline for us to support vectors.

uschindler (changes requested):
see my comments

lucene/core/src/java/org/apache/lucene/internal/vectorization/VectorizationProvider.java
rmuir: it is already tested, you hit the same condition by passing vector size 128. 

I think the trap is just the way we randomize, you need to run the tests a couple times to exercise all the cases? Apparently we got "lucky" (or unlucky depending on how you see it) here.

uschindler:
> if it's 128, that code path is never reached.

That's the reason why I want the test mode keep testing integer vectors also on Intel CPUs without AVX2.

lucene/core/src/java/org/apache/lucene/internal/vectorization/VectorizationProvider.java
uschindler: My comment is not to fix the problem with the `String#format`. I want randomly the testing code to also test integer vectors if the CPU is wrong.

We can keep all code, but I want the "throw new UnsuppoortedOperationException()" only happen on production. That's all. At moment when you explicitely set the bitsize you enable test mode and then I'd like to test also integers.

lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorConstants.java
uschindler: That's OK. I just want the condition to stay alive here.

rmuir:
There's nothing to test. This change makes it so developers only support avx2 and avx512. That's the idea. 

I will fix the name of the confusing constant and consider inverting the logic to call it something like "ancientIntel". It's not related to anything about integers anymore. If you want vectors on x86 you need AVX256 or AVX512.

uschindler:
> There's nothing to test. This change makes it so developers only support avx2 and avx512. That's the idea.
> 
> I will fix the name of the confusing constant and consider inverting the logic to call it something like "ancientIntel". It's not related to anything about integers anymore. If you want vectors on x86 you need AVX256 or AVX512.

Please read again what I said: I want the test mode (where we randomize also preferred bitsize) to also always execute the correctness of the integer code. Testing can be slow, but the integer code has to be tested. Therefor please remove the UOE in test mode. That's plain simple.

If you don't agree with that let's remove all test-mode code. Now it is incomplete.

rmuir:
I understand but the terminology is killing us. There's no more integer-specific stuff anymore, just 128bit case. And we should test it... Slowly, and maybe throw in 64 too!

But the purpose of this change is to remove the confusion around integers and floats. Instead it's just 128, 256, 512. If you are on Intel and only have 128, you get no special sauce.

uschindler:
Thats all fine, then remove the constant. It is no longer risky because the bail out is at one single place: In production

Maybe add a constant called "PANAMA_TEST_MODE" that is enabled once you set bitsize. Let's remove the second sysprop, just keep the "bitsize".

Once any bitsize is enabled, lets stay in test mode and therefor the tests for integers should run. That is a one-line change, so it does not add confusion, just rename the constant.

I am just waiting for you to finish and then add my changes, I did not want to interfere with you.

Sorry: I wrote the original code with the systemprops. It's a bit complicated, so I only try to help. And because of the problem with the formatter we noticed the problem at all: Github runner never executes Panama Tests anymore, because it has no AVX2.

rmuir:
Its helpful, it points out the confusion. The constant needs renaming, to really improve this.

Separately testing needs "backdoor" (which may not be working correctly with this PR in its current state), otherwise nothing will reproduce for anyone, because everyone has different hardware.

uschindler:
I can apply my modifications. I'd suggest to not add a constant and remove the property as you did.

I will just change the code to test integer vectors when a enforced bitsize is also given by `-Ptests.vectorsize=...`

uschindler:
I would also remove the constant so people don't even think about checking `HAS_FAST_INTEGER_VECTORS`.

rmuir:
if you have the time, please feel free. I won't get to this until later.

uschindler:
OK changes pushed.

lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorizationProvider.java
uschindler: it is working again, because i added the testMode back into the game

uschindler:
When verifying the PanamaVectorConstants file I found out that `PRERERRED_LONG_SPECIES` is unused field. Should we remove?

uschindler (approved):

rmuir:
> When verifying the PanamaVectorConstants file I found out that `PRERERRED_LONG_SPECIES` is unused field. Should we remove?

+1 to nuke it. this is the last place we want to have dead code...

uschindler:
> > When verifying the PanamaVectorConstants file I found out that `PRERERRED_LONG_SPECIES` is unused field. Should we remove?
> 
> +1 to nuke it. this is the last place we want to have dead code...

done.

rmuir:
I merged this (hopefully not prematurely) because I'm looking into options to enforce correct checks are happening and this makes the code confusing.

PR #14903, opened 2025-07-06
build: include overview.html into javadocs again

rmuir:
Previously invalid filepaths were being passed to `javadoc` tool:

```
-overview /home/rmuir/workspace/lucene/lucene/core/src/java/overview.html/overview.html
```

Fix the paths so that overview.html content is included. This also means the content is being validated by the build again.

### DIFF
--- a/build-tools/build-infra/src/main/groovy/lucene.documentation.render-javadoc.gradle
+++ b/build-tools/build-infra/src/main/groovy/lucene.documentation.render-javadoc.gradle
@@ -364,13 +364,11 @@ abstract class RenderJavadocTask extends RenderJavadocTaskBase {
         .collect {  dir -> project.file("${dir}/overview.html") }
         .findAll { overviewFile -> overviewFile.exists() }
 
-    if (!overviewFiles.isEmpty()) {
-      assert overviewFiles.size() == 1 : "More than one overview file?: " + overviewFiles
-      opts << [
-        '-overview',
-        project.file("${overviewFiles.getFirst()}/overview.html")
-      ]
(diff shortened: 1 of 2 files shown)


build-tools/build-infra/src/main/groovy/lucene.documentation.render-javadoc.gradle
```diff
@@ -364,13 +364,11 @@ abstract class RenderJavadocTask extends RenderJavadocTaskBase {
...
         .findAll { overviewFile -> overviewFile.exists() }
 
-    if (!overviewFiles.isEmpty()) {
-      assert overviewFiles.size() == 1 : "More than one overview file?: " + overviewFiles
-      opts << [
-        '-overview',
-        project.file("${overviewFiles.getFirst()}/overview.html")
-      ]
-    }
+    assert overviewFiles.size() == 1 : "Must be exactly one overview.html file: " + overviewFiles
```
rmuir, line 367: I think I got too greedy here, as this trips in CI, but didn't locally, will fix.

rmuir: This problem was just me not running all the checks. I added the single missing file: I think it is easier on our reasoning to just require it.

dweiss (approved):
Yes, looks like a bug to me, sorry.

rmuir:
@dweiss it is pretty crazy that `javadoc` doesn't complain about the invalid filename passed as CLI arg: that is what allowed this one.

PR #14905, opened 2025-07-06
build: modernize the javadoc validation, part 1

rmuir:
Previously, we validated javadocs by generating 8,000+ html files, then recursively crawling them with a python script and looked for broken links. This script will no longer find problems, due to how javadocs works.

Instead, DocLint in javac, especially `reference` checks, will fail at compile-time if an invalid reference is created, rather than create a broken link. This is nice because it does not require creating thousands of files to validate what we want.

```
* What went wrong:
  /home/rmuir/workspace/lucene/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java:174: error: reference not found
   * This is a broken link: {@link #bogusReference}
                                   ^
  1 error
```

Valid, but not-visible references from javadocs also no longer create broken links: instead the javadoc tool treats them as if they were `{@linkplain}`.

```html
This is a broken link: <code>deleterIsPrivate</code>
```

We did hit a rare bug before, where javadocs generated an `invalid-tag` and we hacked the script to detect it, but it was just a bug: https://bugs.openjdk.org/browse/JDK-8202617

No longer crawling thousands of html files has a great benefit for `gradle check` development workflow, it means we no longer have to generate them, either.

This saves large amounts of time from `gradle check`, which is wasted: we could be using it instead to find actual problems instead of long-tail problems.

To be especially cautious, this PR still causes CI builds `check` to depend upon `documentation` in order to detect rare corner cases:
* invalid references in overview.html (only 'javadoc' knows these)
* problems in build machinery itself (e.g. gradle). There's some comfort in knowing the thousands of files get created, I guess.

On my computer, doing `gradlew check -x test`:
Before: `BUILD SUCCESSFUL in 4m 33s`
After: `BUILD SUCCESSFUL in 2m 41s`


rmuir:
fwiw, I also experimented with different approaches here, such as disabling the doclint in some tasks.

it may seem wasteful at a glance that we:
1. pass -Xdoclint to javac
2. pass -Xdoclint to renderJavadoc
3. pass -Xdoclint to renderSiteJavadoc
4. lint javadocs via ecj tasks

But it does not matter: Xdoclint doesn't seem to slow anything down. if you take it away from javac, javac doesn't get any faster. The costly part is instead the reading and writing of thousands of files. 

Separately, when debugging the tasks, the ecj tasks have also gotten insanely costly, i'm not sure what happened, will need to look into it at another time.

build-tools/build-infra/src/main/groovy/lucene.documentation.gradle
```diff
@@ -70,6 +70,15 @@ configure(project(':lucene:documentation')) {
...
   })
 
+  // in CI, fully build documentation. this is very costly, way too
+  // much for a developer workflow, but it validates build machinery,
+  // and link references from overview.html, that javac won't catch
+  if (System.getenv("CI") != null) {
+    tasks.named("check").configure {
+      dependsOn 'documentation'
+    }
+  }
```
rmuir, line 81: I still feel dirty about adding this. It costs minutes in the CI build to do this. Maybe we should nuke this, as its a ton of waste to find very rare stuff:

nightly-smoke job will find them, maybe it is a better way.

dweiss: I agree.

dweiss: I replaced the CI check with a global extension flag.

dweiss (approved):

build-tools/build-infra/src/main/groovy/lucene.documentation.gradle
dweiss: But I'm also ok to move it to nightly only.

PR #14907, opened 2025-07-07
build: enable more of javac 24's lints, fix some issues

rmuir:
Let's maximize what javac gives. I revisited all the disabled ones too, the only ones still disabled are the ones that are a real PITA to fix.


dweiss (approved):

dweiss:
Java 25 and on complain with this:
```
> Task :lucene:core:compileMain24Java FAILED
warning: [path] Unexpected extension for archive file: /home/jenkins/workspace/Lucene-main-Linux/lucene/core/src/generated/jdk/jdk24.apijar
error: warnings found and -Werror specified
1 error
1 warning
```

rmuir:
@dweiss thank you, let's just disable that one for now then.

dweiss:
No, wait - I think it's better to just rename those apijar files to something more palatable. Working on it.

rmuir:
even better. I honestly haven't dug into this check, but it seems like a really great sanity check for the build system. As we've seen, the tools don't always fail clearly if you pass them crazy inputs.

dweiss:
#14909 should do it.

uschindler (commented):
OK. Thanks!

PR #14916, opened 2025-07-07
build: enable error-prone's FormatString check

rmuir:
This check will find some printf-style format problems at compile-time instead of runtime: enable it.

Please note: this won't find my original bug yet on #14901, maybe because the file is in `src/java24`. But I tested it works by adding an extra `%s` to a format string in IndexWriter.java and it properly fails the build. I will dig on the `src/java24` issue separately... I suspect we're not running error-prone on these sources.

We have hundreds of such format strings in the codebase, many of which might not have perfect test coverage (they are printing), so it is a good check to have.

### DIFF
--- a/build-tools/build-infra/src/main/groovy/lucene.validation.error-prone.gradle
+++ b/build-tools/build-infra/src/main/groovy/lucene.validation.error-prone.gradle
@@ -199,7 +199,7 @@ allprojects { prj ->
           // '-Xep:FloggerLogVarargs:OFF', // we don't use flogger
           // '-Xep:FloggerSplitLogStatement:OFF', // we don't use flogger
           // '-Xep:ForOverride:OFF', // we don't use this annotation
-          // '-Xep:FormatString:OFF',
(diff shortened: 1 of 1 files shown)


dweiss (approved):

uschindler (approved):

uschindler:
I think we can backport this.

rmuir:
@uschindler It is a good idea. I will take care of it shortly.

rmuir:
Sorry: I came back to this and its conflict-central because build sources are moved around. I'm gonna stay focused on improving `main`, but of course don't mind if anyone wants to backport PRs of interest.

dweiss:
The build on main is far from branch_10x. I've been wondering if backporting everything is a good idea and when to do it (now, later). I have a feeling that - until main reaches some kind of maturity - backporting will involve a lot of extra work. Maybe we should hold off and do it as a one-step later, don't know.

rmuir:
it is also ok to make new major releases when there is lots of change like this :)

PR #14918, opened 2025-07-08
build: move missing-spi-block-tag out of source-patterns.gradle

rmuir:
Making progress on this file, let's do the fun one.

Screen_Shot_2025-07-07_at_22 52 45


dweiss (commented):
Nice. This is similar in style to what you can do with archunit -
https://www.archunit.org/

ast-grep is likely a lot faster but it's the same kind of pattern matching on top of the ast/type tree. I like it.

dweiss (approved):

rmuir:
@dweiss thanks for the link to the archunit, have not heard of this one before. Looks cool, sorta like a forbidden-apis++ in some ways. I imagine it could be fast: forbidden-apis is really fast. Bytecode is better for the kind of analysis those tools do, but sometimes you need to match code comments :)

dweiss:
I have been using this tool for some time. It is... interesting. There are some interesting things you can do with it, especially if you go into custom rules. It isn't particularly fast but it runs slower checks on the CI for me, so it doesn't really bother me. Definitely worth a bookmark though - interesting tool.

PR #14919, opened 2025-07-08
build: use a faster and more robust source-patterns

rmuir:
scans all text files for these patterns. respects .gitignore files so we don't need to configure special exclusions.

uses all cores, scans in like 100ms instead of 20-30 seconds. no need to exclude reuters files or anything like that.

Will take care of the remaining patterns as a separate PR, this gets the majority of them.

### DIFF
--- a/build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/gitgrep/GitGrepPlugin.java
+++ b/build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/gitgrep/GitGrepPlugin.java
@@ -21,23 +21,19 @@ import java.io.UncheckedIOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
-import org.apache.tools.ant.taskdefs.condition.Os;
+import org.apache.lucene.gradle.plugins.LuceneGradlePlugin;
 import org.gradle.api.GradleException;
-import org.gradle.api.Plugin;
 import org.gradle.api.Project;
+import org.gradle.api.provider.Provider;
 import org.gradle.api.tasks.Exec;
 import org.gradle.api.tasks.TaskContainer;
 import org.gradle.api.tasks.TaskProvider;
 
 /** Uses {@code git-grep(1)} to find text files matching a list of patterns. */
-public class GitGrepPlugin implements Plugin<Project> {
+public class GitGrepPlugin extends LuceneGradlePlugin {
   @Override
   public void apply(Project project) {
-    if (project != project.getRootProject()) {
-      throw new GradleException("This plugin can be applied to the root project only.");
-    }
-
-    TaskContainer tasks = project.getTasks();
+    applicableToRootProjectOnly(project);
 
     // currently configured as basic regular expressions...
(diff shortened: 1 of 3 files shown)


rmuir:
From the CI check:
 
    Aggregate task times (possibly running in parallel!):
       0.43 sec.  applyGitGrepRules

dweiss:
I've changed how git executable name is looked up to be consistent with other places. I also confirmed it works on Windows.

dweiss (approved):

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/gitgrep/GitGrepPlugin.java
```diff
@@ -0,0 +1,124 @@
...
+        tasks.register(
+            "applyGitGrepRules",
+            Exec.class,
+            (Exec task) -> {
+              Provider<String> gitExecName =
+                  getBuildOptions(project).getOption("lucene.tool.git").asStringProvider();
+              if (gitExecName.isPresent()) {
+                task.setExecutable(gitExecName.get());
+              } else {
+                task.setEnabled(false);
```
uschindler, line 78: we should log a warning!?

dweiss: I do have a plan to look at this globally, eventually - so that one can run a build without git and without .git (so the source package, for example). We could warn here too, of course.

rmuir: if you want it to work well without `.git` we can also support doing this via ripgrep, which will still respect the .gitignore patterns. That's the single most important thing for performance.

Lemme know, we'd need to tweak the parameters for such a use-case, but both tools will scream.

rmuir: even simpler, we can pass `--no-index` to existing `git grep`, to solve the case where there's no .git.

rmuir: i tried it out, if i use `--no-index` then the `git grep` doesn't process .gitignore, unfortunately. ripgrep is the better solution for this case.

uschindler: Anyways: If there is no git available, the check should be disabled. This may happen on many machines (e.g. Windows).

uschindler: We should still allow users to build Lucene from the Source TGZ without any tools except Java available.

dweiss: Yes, I agree - I do have this planned, it's just low-priority at the moment.

PR #14920, opened 2025-07-08
Avoid allocations in IndexFileNames.parseGeneration method

doom369:
`IndexFileNames.parseGeneration` allocates multiple unnecessary strings (3-6 strings per call) + string[] objects:

image (17)

With often reindexing this allocations become notable during profiling. These allocations could be easily eliminated with a manual underscore search. After the fix only 1 string will be allocated in the worst case.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexFileNames.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexFileNames.java
@@ -145,14 +145,28 @@ public final class IndexFileNames {
   /** Returns the generation from this file name, or 0 if there is no generation. */
   public static long parseGeneration(String filename) {
     assert filename.startsWith("_");
-    String[] parts = stripExtension(filename).substring(1).split("_");
+    int dot = filename.indexOf('.');
+    int end = (dot != -1) ? dot : filename.length();
+    int start = 1; // skip initial '_'
+
+    int first = filename.indexOf('_', start);
+    if (first == -1 || first >= end) {
+      return 0;
+    }
+
+    int second = filename.indexOf('_', first + 1);
+    int third = (second != -1) ? filename.indexOf('_', second + 1) : -1;
+
+    int parts = (second == -1 || second >= end) ? 2 : (third == -1 || third >= end) ? 3 : 4;
+
     // 4 cases:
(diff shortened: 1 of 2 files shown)


rmuir:
> With often reindexing this allocations become notable during profiling.

Your profiler is lying to you. Toss it and get a better one :)

lucene/core/src/java/org/apache/lucene/index/IndexFileNames.java
```diff
@@ -145,14 +145,28 @@ public static String stripSegmentName(String filename) {
...
+    int first = filename.indexOf('_', start);
+    if (first == -1 || first >= end) {
+      return 0;
+    }
+
+    int second = filename.indexOf('_', first + 1);
+    int third = (second != -1) ? filename.indexOf('_', second + 1) : -1;
+
+    int parts = (second == -1 || second >= end) ? 2 : (third == -1 || third >= end) ? 3 : 4;
+
```
rmuir, line 161: This is a massive increase in complexity of the code: the tradeoff isn't worth it.

It's a simple string method called once, doesn't even approach the cost of the system call to actually read the file.

rmuir (changes requested):
Sorry, this doesn't look like the right tradeoff to me. I'd recommend using a better profiler.

I don't see IndexFileNames anywhere in the CPU nor heap profiling of our benchmarks:

* https://benchmarks.mikemccandless.com/2025.07.06.18.05.17.html#profiler_1kb_indexing_1_cpu
* https://benchmarks.mikemccandless.com/2025.07.06.18.05.17.html#profiler_1kb_indexing_1_heap

lucene/core/src/java/org/apache/lucene/index/IndexFileNames.java
doom369: In our scenario, index updates (and we have ~6 different indexes) can happen every 5 seconds. It's profiling info from a 5-minute production run with a async-profiler (one of the best at the moment, imho). So, if these allocations were captured, they're not rare. Agree about complexity. But this method now should be also 5-10x faster if that matters for you. If you want, I can provide you with a microbenchmark.

PR #14924, opened 2025-07-08
build: move forbidden-apis and more to java

dweiss:
This still isn't complete - I don't think rules are applied to build-infra. Just wanted to signal that I'm working on it. 

I've modernized the script a bit and moved signature files around so that it's cleaner which are applied where.  Basically, it's three folders under gradle/validation/forbidden-apis:
```
all/
main/
test/
```
and signatures in each of these are applied to the main source set(s), test source sets or everything.


rmuir:
> This still isn't complete - I don't think rules are applied to build-infra. Just wanted to signal that I'm working on it.

I was going to open an issue about this, maybe it is more than just this task, but let's eventually try to maximize the value of the java sources and crank up any static analysis.

It was one of my biggest frustrations with the groovy that the java solves: on my slow machine, with groovy I'd have to iterate for hours and hours and then ultimately fail. With java, it is less frustrating because of compile-time safety.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/java/ApplyForbiddenApisPlugin.java
```diff
@@ -0,0 +1,293 @@
...
+                            })
+                        .map(
+                            moduleVersion ->
+                                moduleVersion.getName() + "-unsafe-" + moduleVersion.getVersion())
+                        .sorted()
+                        .toList();
+                  });
+          task.getBundledSignatures().addAll(builtInSignatureProvider.get());
+        });
+  }
```
dweiss, line 255: Don't know how to solve this in a better way.

lucene/analysis/common/src/tools/java/org/apache/lucene/analysis/standard/GenerateJflexTLDMacros.java
```diff
@@ -49,11 +51,23 @@
...
+
+  @SuppressForbidden(reason = "Uses System.err, which is fine here.")
+  private static PrintStream getSysErr() {
+    return System.err;
+  }
+
+  @SuppressForbidden(reason = "Uses System.out, which is fine here.")
+  private static PrintStream getSysOut() {
+    return System.err;
+  }
```
dweiss: I pulled out those sysout uses so that other checks are applied. I wish SuppressForbidden had some more granularity to it (think of labels on signature sets or something).

lucene/spatial-extras/src/test/org/apache/lucene/spatial/prefix/TestHeatmapFacetCounter.java
```diff
@@ -66,8 +66,8 @@ public void setUp() throws Exception {
 
   @After
   public void after() {
-    log.info(
-        "Validated " + cellsValidated + " cells, " + cellValidatedNonZero + " non-zero"); // nowarn
+    System.out.println(
+        "Validated " + cellsValidated + " cells, " + cellValidatedNonZero + " non-zero");
```
dweiss, line 70: There were three java util logging calls across the entire codebase (in tests). I nuked them to simplify the build.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/java/ApplyForbiddenApisPlugin.java
```diff
@@ -0,0 +1,293 @@
...
+    }
+  }
+
+  /**
+   * This is hacky; certain signatures are built into forbidden-api checker. Unfortunately, the
+   * task's setBundledSignatures doesn't accept providers and if we try to instantiate the
+   * dependency graph early, it breaks the build... I don't know how to fix this better than by
+   * moving to execution stage.
+   */
+  private static void addBuiltInSignatures(
```
uschindler, line 229: sorry for this....

The alternative would be to hardcode this for the groovy modules using "commons-io" with correct version. We should have a version variable somewhere with the commons-io version?

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/java/ApplyForbiddenApisPlugin.java
uschindler: is this about the commons-io desaster?

uschindler (commented):
Just some comments/suggestions, forbiddenapis looks fine.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/java/ApplyForbiddenApisPlugin.java
dweiss: Don't worry - I still think it's gradle being gradle... It really should be simpler (and it used to be). It works just fine, even though it's not compatible with some gradle's new features (conf. cache).

I think it would be good to have the "@include" tag work for public signature files as well though :) I think it'd be beneficial. It's a great tool, nonetheless.

dweiss: Just to clarify - it's needed if you want to use built-in signatures for commons-io... we could externalize them and put them in Lucene's sources - then it wouldn't be needed at all.

uschindler: Anyways: I thought we have a variable or property with the version number. This would solve the version issue not known without resolve.
The second feature here is to have commons-io automatically applied. This would require resolution and causes the issue here. Could we not simply add the forbidden apis statically to all modules that use forbidden apis?

Anyways, best would be when the Gradle Plugin provided by forbidden-apis could have a bundled signature named "auto-commons-io" that detects version and dependency at runtime and automatically applied. Basically port the code from here over to forbidden-apis.

uschindler: > Don't worry - I still think it's gradle being gradle... It really should be simpler (and it used to be). It works just fine, even though it's not compatible with some gradle's new features (conf. cache).
> 
> I think it would be good to have the "@include" tag work for public signature files as well though :) I think it'd be beneficial. It's a great tool, nonetheless.

No idea what `@Include` tag is. Is it an annotation of newer Gradle?

dweiss: No, I'm takling about this:
https://github.com/policeman-tools/forbidden-apis/blob/main/src/main/resources/de/thetaphi/forbiddenapis/signatures/commons-io-unsafe-2.8.0.txt#L17

it would be nice to be able to include one signature file into another - much like you do, internally.

msokolov:
This broke my local build:

```

* Where:
Build file '/home/ANT.AMAZON.COM/sokolovm/workspace/lucene/build.gradle' line: 30

* What went wrong:
An exception occurred applying plugin request [id: 'lucene.java-projects.conventions']
> Failed to apply plugin class 'org.apache.lucene.gradle.plugins.java.JarManifestConfigurationPlugin'.
   > Extension of type 'LuceneBuildGlobalsExtension' does not exist. Currently registered extension types: [ExtraPropertiesExtension, LibrariesForDeps, LibrariesForLibs, VersionCatalogsExtension, BuildOptionsExtension, BasePluginExtension, DefaultArtifactPublicationSet, GitInfoExtension, AlternativeJdkSupportPlugin.AltJvmExtension]

```

Is there some setup required?  I tried letting the build regenerate gradle.properties but got the same error

uschindler:
I'd start with `gradlew clean` and possibly kill any still running daemons. I had something similar two days ago.

msokolov:
Thanks! Before, gradlew clean would not work either, but it is working now. I think possibly I just waited long enough and the daemons died?

dweiss:
> I'd start with gradlew clean and possibly kill any still running daemons. I had something similar two days ago.

I'm sorry to hear this. Never happened to me and I mess around having the daemon in the background all the time. Eh.

PR #14926, opened 2025-07-09
Add eclint instructions to formatting.txt and 'eclint -fix' to run as part of tidy

dweiss:


rmuir (approved):
hopefully it is rare that someone hits this, as most editors respect editorconfig.

dweiss:
> hopefully it is rare that someone hits this, as most editors respect editorconfig.

I just hit this with sublime text on ubuntu. Don't know - maybe it doesn't care about the editor config or maybe I have set up something incorrectly.

Regardless, I think tidy should do whatever it takes to do most cleanups automatically. To me it's a routine to run tidy and feel like it really does tidy up whatever it needs to. In yet other words: the build system should be your friend, not your enemy.

rmuir:
@dweiss I fully agree. I don't want developers to have "friction" with build.

I think the root cause here is that sublime text needs a plugin: https://github.com/sindresorhus/editorconfig-sublime#readme

PR #14927, opened 2025-07-09
Feat: Add auto formatting bot

### LINKED ISSUE #14835: A workflow "bot" that would apply formatting fixes and create a PR against a PR?

georgereuben:
### Description

Added a workflow bot that automatically formats a PR's file changes and raises a separate PR against that PR

### How it works :
When someone comments ```/format-fix apply``` on a PR, the bot will:
- Check your code using ./gradlew check -x test to find any formatting/style issues
- Fix what it can by running ./gradlew tidy (which includes Google Java Format, EditorConfig fixes, and other style tools)
- Verify everything looks good by running the checks again
- Create a separate PR with just the formatting fixes, so you can review and merge them without mixing formatting changes with your actual code changes

### Supported files:
Java, Gradle, Groovy, Markdown, properties, XML, Python, shell scripts, and batch files.
Just comment "/format-fix apply" on any PR to trigger it

Closes: [Issue #14835](https://github.com/apache/lucene/issues/14835) 

Test Run : [Test PR](https://github.com/georgereuben/lucene/pull/2)

### DIFF
--- a/.github/workflows/auto-format.yml
+++ b/.github/workflows/auto-format.yml
@@ -35,7 +35,7 @@ jobs:
         run: |
           gh api repos/${{ github.repository }}/issues/comments/${{ github.event.comment.id }}/reactions \
             --method POST \
-            --field content="eyes"
+            --input - <<< '{"content":"eyes"}'
 
       - name: Get PR details and checkout
         run: |
@@ -172,10 +172,10 @@ jobs:
       - name: Add success reaction
         if: success()
         run: |
-          # Remove eyes reaction and add success reaction
+          # Add success reaction
           gh api repos/${{ github.repository }}/issues/comments/${{ github.event.comment.id }}/reactions \
             --method POST \
-            --field content="+1"
+            --input - <<< '{"content":"+1"}'
 
       - name: Comment on PR with success (fixes applied)
         if: success() && steps.initial-validation.outcome == 'failure' && steps.check_changes.outputs.changes_made == 'true'
@@ -233,7 +233,7 @@ jobs:
           # Add failure reaction
           gh api repos/${{ github.repository }}/issues/comments/${{ github.event.comment.id }}/reactions \
             --method POST \
-            --field content="-1"
+            --input - <<< '{"content":"-1"}'
 
       - name: Comment on PR with failure
         if: failure()
--- a/.github/workflows/auto-format.yml
+++ b/.github/workflows/auto-format.yml
@@ -1,0 +1,292 @@
+name: Lucene Auto Format Bot
+
+on:
+  issue_comment:
+    types: [created]
+
+env:
+  DEVELOCITY_ACCESS_KEY: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
+
+jobs:
+  format-check:
+    # Only run on pull requests when the bot is mentioned
+    if: |
+      github.event.issue.pull_request &&
+      contains(github.event.comment.body, '/format-fix apply')
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    permissions:
+      contents: write
+      pull-requests: write
+      issues: write
+
+    steps:
+      - name: Check permissions
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const actorPermission = await github.rest.repos.getCollaboratorPermissionLevel({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              username: context.actor
+            });
+
+            const permission = actorPermission.data.permission;
+            const hasPermission = ['admin', 'write'].includes(permission);
+
+            if (!hasPermission) {
+              core.setFailed(`User @${context.actor} does not have permission to run the format bot. Required: admin or write access.`);
+              return;
+            }
+
+      - name: Get PR details
+        id: pr_details
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const prNumber = context.issue.number;
+            const { data: pr } = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: prNumber
+            });
+
+            return {
+              head_ref: pr.head.ref,
+              head_sha: pr.head.sha,
+              base_ref: pr.base.ref,
+              base_sha: pr.base.sha,
+              clone_url: pr.head.repo.clone_url,
+              ssh_url: pr.head.repo.ssh_url,
+              repo_full_name: pr.head.repo.full_name
+            };
+
+      - name: Checkout PR code
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ fromJson(steps.pr_details.outputs.result).repo_full_name }}
+          ref: ${{ fromJson(steps.pr_details.outputs.result).head_ref }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+          fetch-depth: 0
+
+      - name: Set up JDK 21
+        uses: actions/setup-java@v4
+        with:
+          java-version: '21'
+          distribution: 'temurin'
+
+      - uses: ./.github/actions/prepare-for-build
+
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v46
+        with:
+          files: |
+            **/*.java
+            **/*.gradle
+            **/*.groovy
+            **/*.md
+            **/*.properties
+            **/*.xml
+            **/*.py
+            **/*.sh
+            **/*.bat
+            **/*.cmd
+
+      - name: Initial validation
+        id: initial-validation
+        continue-on-error: true
+        run: |
+          echo "Running initial validation..."
+          ./gradlew check -x test
+
+      - name: Fix formatting issues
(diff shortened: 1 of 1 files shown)


georgereuben:
Hi @dweiss , added a workflow bot to apply formatting fixes and create a PR against a PR. Could you please review and suggest changes.

dweiss:
This looks very interesting... and way beyond my knowledge of gh actions. What are the security implications for having these permissions in the workflow? Can somebody craft a PR that would somehow alter Lucene's main repo?

georgereuben:
> This looks very interesting... and way beyond my knowledge of gh actions. What are the security implications for having these permissions in the workflow? Can somebody craft a PR that would somehow alter Lucene's main repo?

There is a check before the workflow runs so that only users with admin or write repository permissions can trigger the bot. I have tested it in the test PR and verified that the workflow rejects the trigger command from an external actor

Screenshot 2025-07-09 at 5 23 25 PM

.github/workflows/auto-format.yml
```diff
@@ -0,0 +1,359 @@
...
+        uses: actions/setup-java@v4
+        with:
+          java-version: '21'
+          distribution: 'temurin'
+
+      - uses: ./.github/actions/prepare-for-build
+
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v46
```
rmuir: This one needs to use a different action, one approved by apache. list:

https://github.com/apache/infrastructure-actions/blob/main/actions.yml

.github/workflows/auto-format.yml
```diff
@@ -0,0 +1,359 @@
...
+    timeout-minutes: 30
+
+    permissions:
+      contents: write
+      pull-requests: write
+      issues: write
+
+    steps:
+      - name: Check permissions
+        uses: actions/github-script@v7
```
rmuir: can we avoid the use of all the javascript and just use `gh` cli instead?

rmuir: if we need js, then we need it, but such large scripts should be factored into separate .js files e.g. in a scripts/ directory.

inline is not good for managing large amounts of logic.

.github/workflows/auto-format.yml
```diff
@@ -0,0 +1,359 @@
...
+
+            ---
+            *This was performed automatically by the Auto Format Bot*`;
+
+            await github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: comment
+            });
```
rmuir: for example, for this entire set of javascript, i'd just use `gh pr comment`

dweiss:
Also - this should be primarily for PR authors who, after they submit the PR, find out something didn't pass the check. So it's a convenience to people who don't have permissions to the main Lucene repository. Is this what the permission check is actually verifying (sorry, it's a bit confusing with those permissions)?

georgereuben:
> Also - this should be primarily for PR authors who, after they submit the PR, find out something didn't pass the check. So it's a convenience to people who don't have permissions to the main Lucene repository. Is this what the permission check is actually verifying (sorry, it's a bit confusing with those permissions)?

Understood, thanks for the insight. I see the issue. Should I go ahead and implement a check where it will allow the PR author along with folks with admin and write access to trigger the format bot? 
I think this should be fine security wise since the automated PR raised would only suggest edits for the PR raised by the person triggering it and not to the main branch. What would you suggest?

Working on the other suggestions meanwhile, I will update here soon

dweiss:
> Should I go ahead and implement a check where it will allow the PR author along with folks with admin and write access to trigger the format bot?

Yes, I think this makes sense. It's a PR against their own repository (against their PR), so I think it's fine (?).

dweiss:
Right - thank you for working on this. It seems way more complex than I thought. Also, it entails some worrying uncertainty with respect to what can go wrong. Here's an example - I received an email today with this:

image

but I can't locate this "comment" anywhere on this issue. Yet it did send out an e-mail, somehow. It's kind of scary to me.

georgereuben:
> Right - thank you for working on this. It seems way more complex than I thought. Also, it entails some worrying uncertainty with respect to what can go wrong. Here's an example - I received an email today with this:
> but I can't locate this "comment" anywhere on this issue. Yet it did send out an e-mail, somehow. It's kind of scary to me.

Yes sorry, I was testing the gh cli comment command locally and accidentally added this PR instead of my forked repo's PR as the target. So this test comment was posted from my id on that PR. I deleted it later, my apologies.

.github/workflows/auto-format.yml
```diff
@@ -0,0 +1,268 @@
...
+    if: |
+      github.event.issue.pull_request &&
+      contains(github.event.comment.body, '/format-fix apply')
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    permissions:
+      contents: write
+      pull-requests: write
+      issues: write
```
dweiss, line 23: Why is the contents permission needed? Just wondering if we could omit it.

.github/workflows/auto-format.yml
```diff
@@ -0,0 +1,268 @@
...
+          repository: ${{ env.head_repo_full }}
+          ref: ${{ env.head_ref }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+          fetch-depth: 0
+
+      - name: Set up JDK 21
+        uses: actions/setup-java@v4
+        with:
+          java-version: '21'
+          distribution: 'temurin'
```
dweiss: this should be java 24 for the main branch. Ideally, you'd reuse Lucene's own prepare-for-build action, much like other workflows do.

.github/workflows/auto-format.yml
georgereuben: It will be needed for pushing formatting commits

.github/workflows/auto-format.yml
georgereuben: Got it thanks, fixing this

georgereuben:
Hi @dweiss, I now see that it might not be possible to raise a PR against a PR raised from a branch in a forked repo since the github actions workflow will not have access to that forked repo. 
The other viable solution I can see is to add a commit to the author's PR itself as this will be possible if the PR author has allowed edits by maintainers. I have been working on this but I wanted to check with you if this would be viable

dweiss:
Yes, this could be an alternative. I initially wanted a PR because they you'd see what's changed but a commit is also fine if the change is triggered explicitly.

.github/workflows/auto-format.yml
georgereuben: I tried using ana06/get-changed-files@v2.3.0, but it doesn't support issue comment triggers. So I have replaced this action with a gh cli script that would perform the same action when triggered by a comment. Please let me know if this works.

.github/workflows/auto-format.yml
georgereuben: Thanks I see the issue, I have replaced all with gh cli scripts.

.github/workflows/auto-format.yml
georgereuben: Updated with the existing prepare-for-build action

georgereuben:
Hi @dweiss @rmuir, I have updated the workflow. If the PR is raised in the same repo by a maintainer, it will raise a PR with formatting fixes, and if the PR is raised from a forked repo, it will directly commit the formatting fixes to the author's PR and the author can check the commit changes.

I have updated the workflow with the rest of your suggestions as well, please review and let me know if any other improvements can be made, thanks.

dweiss (approved):
Thank you for your effort, @georgereuben 

It's a hell of a lot of a complex workflow... but I think we should give it a try. If it's used - fine, if it's not used after some time, we can drop it. 

I'm going to merge it and see what happens.

dweiss:
Hi @georgereuben ! So... I finally have a chance to try this new shiny thing but it doesn't seem to be doing anything - https://github.com/apache/lucene/pull/15052#issuecomment-3182536372 - I wonder if I'm doing something wrong.

dweiss:
Uh, sorry - it does work, it just didn't show progress while working!

uschindler:
Hi,

Because of the impact shown in https://github.com/apache/lucene/pull/16124, I deleted this workflow. More details were reported privately to the Lucene PMC.

The workflow is way too complex and needs write rights in the main ASF repository which should not be executed as PR actions. I don't understand why that workflow needs to create branches and new PRs. If it could be rewritten to only reformat an single PR in isolation and needs no extra permissions, then I am fine to give it a second chance. Another option would be that the workflow only makes suggestions.

Uwe

PR #14929, opened 2025-07-10
Fix TestLateInteractionField.testInputValidation

HUSTERGS:
### Description

Currently the `value.length` can be very small, and there is a chance that all vector have same dim, thus makes the last Assertion fail, this can be stably reproduced on my machine with `gradlew test --tests TestLateInteractionField.testInputValidation -Dtests.seed=92E678DBD2F6CD0A -Dtests.locale=en-DE -Dtests.timezone=Etc/GMT+7 -Dtests.asserts=true -Dtests.file.encoding=UTF-8`

https://github.com/apache/lucene/blob/c5a802521fe51536da9d951dba782c96608ee03c/lucene/core/src/test/org/apache/lucene/document/TestLateInteractionField.java#L67-L75

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/document/TestLateInteractionField.java
+++ b/lucene/core/src/test/org/apache/lucene/document/TestLateInteractionField.java
@@ -64,14 +64,16 @@ public class TestLateInteractionField extends LuceneTestCase {
     expectThrows(IllegalArgumentException.class, () -> LateInteractionField.encode(emptyTokens));
 
     final int dim = 128;
-    float[][] value = new float[random().nextInt(20, 30)][];
-    for (int i = 0; i < value.length; i++) {
+    final int numVectors = random().nextInt(3, 12);
+    float[][] value = new float[numVectors][];
+    for (int i = 0; i < numVectors - 1; i++) {
       if (random().nextBoolean()) {
         value[i] = TestVectorUtil.randomVector(dim);
       } else {
         value[i] = TestVectorUtil.randomVector(dim + 1);
       }
     }
+    value[numVectors - 1] = TestVectorUtil.randomVector(dim + 2);
     expectThrows(IllegalArgumentException.class, () -> LateInteractionField.encode(value));
   }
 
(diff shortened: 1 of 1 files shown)


lucene/core/src/test/org/apache/lucene/document/TestLateInteractionField.java
```diff
@@ -64,7 +64,7 @@ public void testInputValidation() {
     expectThrows(IllegalArgumentException.class, () -> LateInteractionField.encode(emptyTokens));
 
     final int dim = 128;
-    float[][] value = new float[random().nextInt(3, 12)][];
+    float[][] value = new float[random().nextInt(20, 30)][];
```
vigyasharma: It happens if `random().nextBoolean()` in the test always returns the same value, thus generating all token vectors with the same dimensions. Instead of increasing the number of vectors this way, let's make the test more deterministic by ensuring that the last vector has different dimensions. Something like:

```java
    final int dim = 128;
    final int numVectors = random().nextInt(3, 12);
    float[][] value = new float[numVectors][];
    for (int i = 0; i < numVectors - 1; i++) {
      if (random().nextBoolean()) {
        value[i] = TestVectorUtil.randomVector(dim);
      } else {
        value[i] = TestVectorUtil.randomVector(dim + 1);
      }
    }
    value[numVectors - 1] = TestVectorUtil.randomVector(dim + 2);
    expectThrows(IllegalArgumentException.class, () -> LateInteractionField.encode(value));
```

vigyasharma (changes requested):
Whoa, Thanks for spotting this! Appreciate the quick fix.

lucene/core/src/test/org/apache/lucene/document/TestLateInteractionField.java
HUSTERGS: Thanks for your reply! I've pushed a new commit according to your suggetions

vigyasharma (approved):

PR #14933, opened 2025-07-10
Move more gradle scripts to java

dweiss:
Just another iteration. Draft until it reaches a reasonable size.


dweiss:
Just a note that I may be less responsive this week due to holidays. I'll try to push this patch forward in the mean time but it is still in draft mode (so lots of changes occurring here and there).

build-tools/build-infra/src/main/groovy/lucene.regenerate.gradle
```diff
@@ -135,14 +195,17 @@ configure([
     //   ignoreWithSource: any other tasks that should be ignored if the internal task is ignored
     //     (checksums are identical)
     //   mustRunBefore: any tasks which should be scheduled to run after the internal task.
-    wrapWithPersistentChecksums = { Task sourceTask, Map<String, Object> extraConfig = [:] ->
+    wrapWithPersistentChecksums2 = {TaskProvider sourceTaskProvider, Map<String, Object> extraConfig = [:] ->
+      wrapWithPersistentChecksums(sourceTaskProvider.get(), extraConfig)
+    }
+    wrapWithPersistentChecksums = {Task sourceTask, Map<String, Object> extraConfig = [:] ->
```
dweiss, line 201: This is something I'm leaving for later - it currently instantiates the task but there is no simple way around it.

gradle/regenerate/snowball/snowball.sh
```diff
@@ -47,7 +47,7 @@ for file in ${SRCDIR}/java/org/tartarus/snowball/ext/*.java; do
   # title-case the classes (fooStemmer -> FooStemmer) so they obey normal java conventions
   base=$(basename $file)
   oldclazz="${base%.*}"
-  newclazz=${oldclazz^}
+  newclazz=$(echo "$oldclazz" | perl -pe 's/^(.)/\U$1/')
```
dweiss, line 50: this hack is needed for macos because it has an old bash by default and doesn't support the fancy uppercasing syntax. Hope it's not a problem.

rmuir: long-term i think the best fix would be to submit snowball a patch to address the naming to remove the hack: they were welcoming of such improvements the last time I tried to work some of these hacks off.

rmuir: we could also try to stop doing this, and add exceptions to any checkers around the naming in use.

I think I previously explored that option, but java tools were extremely angry about class named `fooStemmer`... they complained louder than you would imagine.

dweiss:
I think this is already large enough and fixes enough things that we may merge it as is. I'll create another PR with more progress.

PR #14938, opened 2025-07-12
add illegal-incubator-usage lint check

### LINKED ISSUE #14937: move all vector api kernels to PanamaVectorUtilSupport.java
### Description

Ban access to the vector api from any other file with ast-grep check.

Currently invocations are spread needlessly across multiple files, which makes static analysis more difficult than it needs to be.

rmuir:
Prevent proliferation of vectors api code outside of PanamaVectorUtilSupport.

It is already happening: example is the memory segment abstraction, which I've left as a TODO to fix.

Let's keep all vectors code into a single file for easy static analysis: we can add simple rules to prevent performance traps instead of constantly staring at assembly, looking at openjdk sources, debugging 10x slowdowns.

But we can't do it, if kernels spread via complex java abstractions, because it makes the analysis too difficult.

Closes #14937

This is just the first step. We may have to adopt some different patterns and so on with our vectors code to help prevent the traps. But I think we must use automation and not do it manually on every PR.

### DIFF
--- a/gradle/validation/ast-grep/rules/vectors.yml
+++ b/gradle/validation/ast-grep/rules/vectors.yml
@@ -1,0 +1,20 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ast-grep/ast-grep/refs/heads/main/schemas/java_rule.json
+id: illegal-incubator-usage
+language: java
+ignores:
+  # the only place vector api code should sit.
+  - "**/internal/vectorization/PanamaVector{Constants,UtilSupport}.java"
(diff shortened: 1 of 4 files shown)


rmuir:
Screen_Shot_2025-07-11_at_20 07 41

I hope we can address this one as a followup.
The other usage fixed here was a jdk bug, that's been fixed in JDK22, is for security manager only, and doesn't impact us on JDK24

rmuir:
declined in favor of #14942

PR #14896, opened 2025-07-03
Vectorize `filterCompetitiveHits`

HUSTERGS:
### Description

This PR is a follow-up of the [comment](https://github.com/apache/lucene/pull/14827#issuecomment-3018852667) from #14827 , trying to vectorize the `filterCompetitiveHits` function by utilizing `(Int|Float)Vector#compress`.

I'm still working on it, tests are not added yet, nor is the code stable , comments and suggestions are welcomed !

But I do did a quick run of luceneutil based on `62e0276032189deee9559327cc53ac3f59f354a9` with `wikimediumall` with `searchConcurrency=0, taskCountPerCat=1, taskRepeatCount=20`, here is the result after 20 iterations, which seems to be promising (hope I didn't get anything wrong). Will do another run with different setup

```
                            TaskQPS baseline      StdDevQPS my_modified_version      StdDev                Pct diff p-value
                   TermMonthSort     1715.92      (5.2%)     1677.87      (6.7%)   -2.2% ( -13% -   10%) 0.245
                      DismaxTerm      687.90      (3.6%)      672.66      (4.8%)   -2.2% ( -10% -    6%) 0.099
                 FilteredPrefix3      145.56      (3.6%)      142.51      (5.1%)   -2.1% ( -10% -    6%) 0.134
                       OrHighMed      124.23      (5.3%)      122.03     (12.4%)   -1.8% ( -18% -   16%) 0.556
                        Wildcard       89.60      (2.7%)       88.11      (3.4%)   -1.7% (  -7% -    4%) 0.087
                          Fuzzy2       25.52      (3.2%)       25.24      (4.1%)   -1.1% (  -8% -    6%) 0.342
                         Respell       38.38      (2.6%)       37.99      (2.1%)   -1.0% (  -5% -    3%) 0.172
               TermDayOfYearSort      252.39      (3.3%)      250.17      (4.3%)   -0.9% (  -8% -    6%) 0.463
                          Phrase        4.58      (2.5%)        4.54      (3.3%)   -0.8% (  -6% -    5%) 0.378
                     CountPhrase        3.21      (2.1%)        3.19      (2.7%)   -0.7% (  -5% -    4%) 0.352
                 DismaxOrHighMed       65.05      (3.6%)       64.59      (7.7%)   -0.7% ( -11% -   11%) 0.709
             FilteredOrStopWords        8.51      (2.8%)        8.46      (2.6%)   -0.6% (  -5% -    4%) 0.507
[...]

PR #14939, opened 2025-07-12
fail 'gradle check' on unchecked slow vector api feature usage 

rmuir:
slow vector features: operations that might be slow on particular hardware. For now the list is `fma`, `compress`, and `expand`.

unchecked: not guarded by a `if (Constants.HAS_XXX)` for use of the potentially slow features.

check is not perfect, because i'm tired. I implemented some of the chasing (e.g. imports) just to make progress though.

I don't want to overly restrict the idioms in use, but use of fancy java features in the vectorization code such as `var` and `import static` equate to more work, if we want to avoid false positives.

I imagine this thing might graduate from incubating at some point, so it'd be nice to run it across all the code without annoying people.

### DIFF
--- a/gradle/validation/ast-grep/rules/vectors.yml
+++ b/gradle/validation/ast-grep/rules/vectors.yml
@@ -1,3 +1,5 @@
+# Special policing of vectorization sources to prevent performance traps!
+---
 # yaml-language-server: $schema=https://raw.githubusercontent.com/ast-grep/ast-grep/refs/heads/main/schemas/java_rule.json
 id: illegal-incubator-usage
 language: java
@@ -18,3 +20,19 @@ rule:
 severity: error
 message: Incubator APIs must not be used
 note: Move logic to `PanamaVectorUtilSupport`
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ast-grep/ast-grep/refs/heads/main/schemas/java_rule.json
+id: unchecked-slow-vector-feature
+language: java
+files:
+  - "**/internal/vectorization/**.java"
+rule:
+  all:
+    - matches: slow-vector-feature
+    - not:
+        inside:
+          matches: vector-feature-check
+          stopBy: end
+severity: error
+message: "Vector API feature `$IDENTIFIER` may be slow on some hardware"
+note: guard usage inside `if` with `Constants.HAS_<FEATURE>` check
--- a/gradle/validation/ast-grep/rules/vectors.yml
+++ b/gradle/validation/ast-grep/rules/vectors.yml
@@ -1,0 +1,20 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ast-grep/ast-grep/refs/heads/main/schemas/java_rule.json
+id: illegal-incubator-usage
+language: java
+ignores:
+  # the only place vector api code should sit.
+  - "**/internal/vectorization/PanamaVector{Constants,UtilSupport}.java"
+  # TODO: move this vector logic to PanamaVectorUtilSupport!
+  - "**/internal/vectorization/MemorySegmentPostingDecodingUtil.java"
+rule:
+  kind: scoped_identifier
+  all:
+    - has:
+        field: scope
+        pattern: jdk
+    - has:
+        field: name
+        pattern: incubator
+severity: error
+message: Incubator APIs must not be used
+note: Move logic to `PanamaVectorUtilSupport`
--- a/gradle/validation/ast-grep/utils/slow-vector-feature.yml
+++ b/gradle/validation/ast-grep/utils/slow-vector-feature.yml
@@ -1,0 +1,70 @@
+# Matches a vector API feature which may be slow on some hardware
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ast-grep/ast-grep/refs/heads/main/schemas/java_rule.json
+id: slow-vector-feature
+language: java
+utils:
+  # imported as $IMPORT from the vector api
+  has-import:
+    pattern: $$$
+    inside:
+      kind: program
+      stopBy: end
+      has:
+        kind: import_declaration
+        any:
+          - pattern: import static $PACKAGE.$IMPORT;
+          - pattern: import $PACKAGE.$IMPORT;
+    # don't match identifiers inside import statements themselves
+    not:
+      inside:
+        kind: import_declaration
+        stopBy: end
+rule:
+  any:
+    # unqualified access (field or method) with matching static import
+    - kind: identifier
+      all:
+        # matches name on bad list
+        - pattern: $IDENTIFIER
+        # also matches vector api import
+        - pattern: $IMPORT
+        - matches: has-import
+    # qualified constant field access with matching import
+    - kind: field_access
+      all:
+        - has:
+            field: field
+            pattern: $IDENTIFIER
+        - has:
+            field: object
+            pattern: $IMPORT
+            matches: has-import
+    # qualified method call (not a helper function with same name)
+    - kind: method_invocation
+      all:
+        # matches name on bad list
+        - has:
+            field: name
+            pattern: $IDENTIFIER
+        # any receiver, but it must have one
+        # TODO: refine if people complain :)
+        - has:
+            field: object
+            pattern: $$$
+constraints:
+  PACKAGE:
+    regex: ^jdk.incubator.vector
+  # slow methods and constants in the vector API
+  # please send a pull request if you find others!
+  IDENTIFIER:
+    any:
+      # falls back to BigInteger
+      - pattern: fma
+      - pattern: FMA
+      # not supported on NEON
+      - pattern: compress
+      - pattern: COMPRESS_BITS
+      # not supported on NEON
+      - pattern: expand
+      - pattern: EXPAND_BITS
--- a/lucene/core/src/java24/org/apache/lucene/internal/vectorization/MemorySegmentPostingDecodingUtil.java
+++ b/lucene/core/src/java24/org/apache/lucene/internal/vectorization/MemorySegmentPostingDecodingUtil.java
@@ -36,6 +36,8 @@ final class MemorySegmentPostingDecodingUtil extends PostingDecodingUtil {
     this.memorySegment = memorySegment;
   }
 
+  // FIXME: move this logic to PanamaVectorUtilSupport.
+
   private static void shift(
       IntVector vector, int bShift, int dec, int maxIter, int bMask, int[] b, int count, int i) {
     for (int j = 0; j <= maxIter; ++j) {
--- a/gradle/validation/ast-grep/tests/vectors.yml
+++ b/gradle/validation/ast-grep/tests/vectors.yml
@@ -1,0 +1,8 @@
+# Prevent vector api calls from spreading haphazardly without checks
+---
+id: illegal-incubator-usage
+invalid:
+  - import jdk.incubator.vector.FloatVector
+  - import static jdk.incubator.vector.VectorOperators.ADD
+valid:
+  - import java.util.List
--- a/gradle/validation/ast-grep/tests/vectors.yml
+++ b/gradle/validation/ast-grep/tests/vectors.yml
@@ -6,3 +6,64 @@ invalid:
   - import static jdk.incubator.vector.VectorOperators.ADD
 valid:
   - import java.util.List
+---
+id: unchecked-slow-vector-feature
+invalid:
+  - import static jdk.incubator.vector.VectorOperators.FMA
+    a.lanewise(FMA, b, c)
+  - import static jdk.incubator.vector.VectorOperators.FMA
+    if (other) {
+      a.lanewise(FMA, b, c)
+    }
+  - import static jdk.incubator.vector.VectorOperators.FMA
+    if (Constants.HAS_FAST_VECTOR_FMA || other) {
+      a.lanewise(FMA, b, c)
+    }
+  - import static jdk.incubator.vector.VectorOperators.FMA
+    if (!Constants.HAS_FAST_VECTOR_FMA) {
+      a.lanewise(FMA, b, c)
+    }
+  - import static jdk.incubator.vector.VectorOperators.FMA
+    if (Constants.HAS_FAST_VECTOR_FMA) {
+      somethingElse()
+    } else {
+      a.lanewise(FMA, b, c)
+    }
+  - import jdk.incubator.vector.VectorOperators
+    a.lanewise(VectorOperators.FMA, b, c)
+  # note: doesn't actually chase types yet
+  - import jdk.incubator.vector.ByteVector
+    ByteVector x = ByteVector.fromMemorySegment()
+    x.fma(other, other)
+valid:
+  - |
+    import static jdk.incubator.vector.VectorOperators.FMA
+    if (Constants.HAS_FAST_VECTOR_FMA) {
+      a.lanewise(FMA, b, c)
+    }
+  - |
+    import static jdk.incubator.vector.VectorOperators.FMA
+    if (Constants.HAS_FAST_VECTOR_FMA && other) {
+      a.lanewise(FMA, b, c)
+    }
+  - |
+    import static jdk.incubator.vector.VectorOperators.FMA
+    if (other && Constants.HAS_FAST_VECTOR_FMA) {
+      a.lanewise(FMA, b, c)
+    }
+  - |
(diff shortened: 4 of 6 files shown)


rmuir:
another really practical way to help contain this stuff would be to "factor out the unsafeness" into helpers, like this one we have for FMA:

https://github.com/apache/lucene/blob/41a4a1d1ea999446cd29dc4a0f4e71641187fc18/lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java#L81-L87

rmuir:
and if we factor out to different file of small helpers, it would be great to just use forbidden-apis and `@SuppressWarnings` the file. It would be a simpler solution than this one.

I don't know if forbidden will support us banning the incubating APIs, but I support it supporting that, then we wouldnt need this PR at all.

rmuir:
converted to draft: want to chase forbidden-apis based solution. If helpers with `@SuppressWarnings` are simple if-statements then it should work.

rmuir:
@uschindler here is the error from forbidden-apis I get when I try to ban vector api methods:

```
* What went wrong:
Execution failed for task ':lucene:core.tests:forbiddenApisMain'.
> Parsing signatures failed: Class 'jdk.incubator.vector.FloatVector' not found on classpath while parsing signature: jdk.incubator.vector.FloatVector#fma(**)
```

If we can solve this somehow, this PR may become a 4-liner.

uschindler:
Sorry this can't be solved as forbidden-apis runs inside Gradle VM and can't see any classes from API jars or incubator modules. This is why the whole vector stuff is a separate compilation unit.

Please stop trying to do anything like that.

dweiss:
I can't remember - why is it that we exclude vector module from forbidden apis (but pass the apijar stubs for compilation)?

uschindler:
the problem is that forbidden-apis runs in the gradle VM and not in a separate JVM configured to expose the module. It is impossible to do, sorry. This is similar to why ECJ when running in the gradle VM can't correctly lint stuff.

uschindler:
We may try to pass the apijar to forbidden apis as normal classpath entry. I haven't tried.

gradle/validation/ast-grep/rules/vectors.yml
```diff
@@ -0,0 +1,38 @@
+# Special policing of vectorization sources to prevent performance traps!
+---
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ast-grep/ast-grep/refs/heads/main/schemas/java_rule.json
+id: illegal-incubator-usage
+language: java
+ignores:
+  # the only place vector api code should sit.
+  - "**/internal/vectorization/PanamaVector{Constants,UtilSupport}.java"
+  # TODO: move this vector logic to PanamaVectorUtilSupport!
+  - "**/internal/vectorization/MemorySegmentPostingDecodingUtil.java"
```
uschindler, line 10: this is perfectly fine and needed to be there.

lucene/core/src/java24/org/apache/lucene/internal/vectorization/MemorySegmentPostingDecodingUtil.java
```diff
@@ -36,6 +36,8 @@ final class MemorySegmentPostingDecodingUtil extends PostingDecodingUtil {
     this.memorySegment = memorySegment;
   }
 
+  // FIXME: move this logic to PanamaVectorUtilSupport.
```
uschindler, line 39: I disagree here. This is in the internal/vectorization module, no need to move it.

gradle/validation/ast-grep/rules/vectors.yml
uschindler, line 1: Basically the whole chekc here is not needed. Because (as you see with forbiddenapis) in the main sources is imposisble because thanks to the module system the incubator classes are not available.

So a class in normal java/main there is no risk to introduce vector classes (only if somebody adds jdk.incubator) to the `module-info.java`.

So please remove this whole file. Thanks.

uschindler (changes requested):
Hi,
in fact some of the checks are nor usefull because the module system already prevents usage of vector APIs.

The way how it is setup in our build system makes it impossible to call any jdk.incubator.vector APIs form normal souce code in the main sourceSet.

Inside the java24 module there is only vectorization code, so we don't need that check (we compile against the apijar).

You may only add one special check like: don't allow to add `jdk.incubator.vector` to `module-info.java`. That would be enough.

uschindler:
I added a review. The forbiddenapi checks are not needed because noone can call vector API from our main sourceSet. It is limited to the java24 sourceSet. I think you're hunting ghosts - sorry!

uschindler:
FYI, I did a quick test:

```diff
 lucene/core/src/java/org/apache/lucene/util/Bits.java | 2 ++
 1 file changed, 2 insertions(+)

diff --git a/lucene/core/src/java/org/apache/lucene/util/Bits.java b/lucene/core/src/java/org/apache/lucene/util/Bits.java
index 61757a1a34e..79b078f9e73 100644
--- a/lucene/core/src/java/org/apache/lucene/util/Bits.java
+++ b/lucene/core/src/java/org/apache/lucene/util/Bits.java
@@ -18,6 +18,8 @@ package org.apache.lucene.util;
 
 import org.apache.lucene.search.DocIdSetIterator;
 
+import static jdk.incubator.vector.VectorOperators.ADD;
+
 /**
  * Interface for Bitset-like structures.
  *
```

So I added a fake import to a class from the main sourceSet.

If you invoke gradle it fails after a very short time:

```
  C:\Users\Uwe Schindler\Projects\lucene\lucene\lucene\core\src\java\org\apache\lucene\util\Bits.java:21: Fehler: Statischer Import nur aus Klassen und Schnittstellen
  import static jdk.incubator.vector.VectorOperators.ADD;
  ^
  Hinweis: Wiederholen Sie die Kompilierung mit -Xlint:deprecation, um Details zu erhalten.
  C:\Users\Uwe Schindler\Projects\lucene\lucene\lucene\core\src\java\org\apache\lucene\util\Bits.java:21: Fehler: Package jdk.incubator.vector ist nicht sichtbar
  import static jdk.incubator.vector.VectorOperators.ADD;
                             ^
    (Package jdk.incubator.vector wird in Modul jdk.incubator.vector deklariert, aber nicht von Modul org.apache.lucene.core gelesen)
  Hinweis: Einige Eingabedateien verwenden oder ├╝berschreiben eine veraltete API.
  2 Fehler
```

(sorry for being German).

So basically it is already impossible to use incubator code anywhere except in java24 sourceSet, where we have all vector code encapsulated already. That is the main reason why I introduced the separate sourceSet (next to the compilation against apijars).

The checks about the `if (HAS_XY....)` may be usefull, but the rest of this PR and also forbiddenapis is not needed. The type system is strong enough.

And `MemorySegmentPostingDecodingUtil` is perfectly fine that it uses vector code because it is an implementation of the vector API. It is made for that. Please keep the code at the place where it is used.

-1 to merge this PR!

uschindler:
P.S.: The above javac checks is also the reason why forbiddenapis can't find the vector classes!

Another thing: Forbiddenapis prevents usage of any jdk.XXXXX anyways due to the "portable" check (only allow public APIs from JDK from officially exported packages). We only removed the check in the java24 sourceset!

uschindler:
Basically we should add a simple git-grep rule only that checks: Fail build if any `module-info.java` contains 'jdk.incubator.vector':

```
$ git grep jdk.incubator.vector **/module-info.java
```

this creates no positives yet. I know sometimes some devs on PRs try to do that in the benchmark-jmh module, so I'd add this simple grep. This is best to prevent any incubator usage anywhere (and if not forbiddenapis will do it anyways).

dweiss:
thanks.

rmuir:
Uwe I don't think you understand what this PR does.

lucene/core/src/java24/org/apache/lucene/internal/vectorization/MemorySegmentPostingDecodingUtil.java
rmuir: Then please fix the bug in forbidden APIs so that I can use it to ban slow methods such as `compress`.

rmuir:
I think Uwe needs his espresso first. This is what this PR does. It prevents UNCHECKED (meaning if-statement) use of specific vector APi constructs.

fails:
```
byteVector.compress(mask);
```

passes:
```java
if (Constants.HAS_FAST_COMPRESS) {
   byteVector.compress(mask);
}
```

Screen_Shot_2025-07-12_at_09 02 52

gradle/validation/ast-grep/rules/vectors.yml
rmuir: I know you are defensive about the memory segment code, but the vectors logic needs to be encapsulated into static method in the utils class, to make static analysis easier. This way no cross-file analysis is needed and all the kernels are in a single file.

gradle/validation/ast-grep/rules/vectors.yml
rmuir: You don't understand what the check is doing. I'm trying to prevent the spread of vectors code across multiple files in src/java24 itself.

It is a problem: if our tools such as forbidden-apis are broken and buggy on src/java24, then ast-grep is the only weapon I have: means please keep the logic into a single file and don't add vector logic to tons of classes in src/java24

rmuir:
> It is limited to the java24 sourceSet. I think you're hunting ghosts - sorry!

You misunderstand the check. it is about checking java24 itself. It is about keeping vector api code contained into single files there, with simple functions, that can be traced in simple ways for analysis.

What the memory segment code is doing (inlining its own vector code, instead of factoring that out into a separate static method in PanamaVectorUtils) is a perfect example of bad coding practices.

It is no different than containing the unsafeness of FAISS to single file.

With vector kernels only in a single file, then all the accesses can be traced and analyzed with simple tools, especially when tools such as forbidden apis aren't working at all on the sources (which would make solving these problems easier).

rmuir:
> FYI, I did a quick test:

That's not a test of what this PR is doing. Please properly try to understand what the check is trying to do :) I left LOTS of comments and explanations. 

It isn't about src/java at all.

rmuir:
I will commit later if nobody objects (I mean really objects, and actually understands what checks are actually doing). 

If we want to fix forbidden apis then we can simplify some logic.

uschindler:
Sorry I disagree with this. My strong -1.

The whole code in src/java24 ist solely for vector code so I expect and allow vector code everywhere. There is no need to mit this on one single file. VectorUtilSupport is only an implementation class (it is not static!!!!) that implements the interface from main code.

If you want to separate all code using vector incubator do it in a real static class. The current code is a mix of static and non static stuff. I agree to move all code to a single class bit then it needs to be a real static one without instances. The VectorUtilSupport is just implementation of an interface. Therefore the other code should not be moved there. What you do is mixing up different stuff. This is not valid to do.

I will immediately revert any merge of this branch, this is all not needed. If you want control of vector code then add GitHub requirement to have a review by yourself.

Static analysis is overkill.

Forbidden-apis is not broken it can't just work in this setup.

rmuir:
Uwe you miss that people are trying to use specific functions in vectors code such as `compress` that only work on certain cpus (e.g. require ARM SVE), and cause massive slowdowns if used without a guard.

failing any improvements to forbidden-apis, we need to manage a list of functions (such as `fma` and `compress`: only two I specified in the list here) that can be slow without being wrapped in a "feature check". Thats what this is all about.

currently its being done manually, on every single vectors PR. Then someone discovers on XYZ hardware that they have a slowdown.

with this tooling (or alternatively forbidden apis), we can manage these in a better way other than "manually".

gradle/validation/ast-grep/utils/slow-vector-feature.yml
```diff
@@ -0,0 +1,70 @@
...
+    any:
+      # falls back to BigInteger
+      - pattern: fma
+      - pattern: FMA
+      # not supported on NEON
+      - pattern: compress
+      - pattern: COMPRESS_BITS
+      # not supported on NEON
+      - pattern: expand
+      - pattern: EXPAND_BITS
```
rmuir, line 70: Uwe look right here: this is the part you are missing I think.

rmuir:
> The whole code in src/java24 ist solely for vector code so I expect and allow vector code everywhere. 

We literally just hit this exact problem on #14896 using compress() without a check that the cpu has the feature. This PR detects the problem.

rmuir:
Here's how i'd do this with forbidden-apis:

```
@defaultMessage("potentially slow on some cpus, please check the cpu has the feature")
# falls back to BigDecimal
jdk.incubator.vector.Float16#fma(**)
jdk.incubator.vector.FloatVector#fma(**)
jdk.incubator.vector.DoubleVector#fma(**)
jdk.incubator.vector.VectorOperators.FMA
# unsupported on neon
jdk.incubator.vector.ByteVector#compress(**)
jdk.incubator.vector.IntVector#compress(**)
jdk.incubator.vector.ShortVector#compress(**)
jdk.incubator.vector.LongVector#compress(**)
jdk.incubator.vector.VectorOperators.COMPRESS_BITS
# unsupported on neon
jdk.incubator.vector.ByteVector#expand(**)
jdk.incubator.vector.IntVector#expand(**)
jdk.incubator.vector.ShortVector#expand(**)
jdk.incubator.vector.LongVector#expand(**)
jdk.incubator.vector.VectorOperators.EXPAND_BITS
```

We'd have to define helpers (like we have for FMA) for compress() and expand() that have the `@SuppressWarnings`, but it would serve the same goal.

Goal is simple: if the code calls IntVector.compress() or similar, I want us to first check that the CPU supports that capability.

uschindler:
Hi you are doing multiple things in this PR. Mainly I don't want the first file in the diff. This one is my main problem. Don't concentrate all code on one place and mix static and non static stuff in this interface implementation PanmaVectorUtilSupport. This class is only to implement this interface and not static utility methods for other code on this package. The Memory segment code should be where it belongs! We may refactor all this I to a single private static class but I don't think that's needed. It would just add additional indirection into the VectorUtil class. So please implement the code where it belongs. That proposal is my big -1.

To make forbidden apis work: just add the API jars file to forbidden-apis class path (with some Gradle magic) then it would work. I can't find this anymore because the new Java code is too crazy to me. I would have liked to stay forbidden-apis using groovy code. Maybe @dweiss  can help. Basically the forbidden-apis call needs to get an advanced configuration with the API jars provided by the core-mrjar code. Maybe define a custom class path there that includes only the API jars. Do it at the place where it disables the jdk-nonportable bundled signature for the Mrjar code.

Then you can forbid those methods.

rmuir:
> Hi you are doing multiple things in this PR. Mainly I don't want the first file in the diff. This one is my main problem.

OK, let's just remove that here. Pretend it isn't here. 

My goal is to ban trappy methods that require specific hardware support. If I can do it with forbidden-apis, then we'll add helpers like we do for fma() for each one of them, and check for the hardware feature in the helper: problem solved. I will dig more into the forbidden apis logic in the gradle. I think it is in java now, so I should be able to solve it with the api jar.

If we can't check for the trappy methods with forbidden-apis, then only other idea I have is to fallback to hackier solutions as seen here, it is difficult to "ban methods" from source code perspective: you've gotta deal with types and imports and all the crazy logic you see in this PR. Bytecode approach with forbidden apis is simpler.

uschindler:
I will try to fix the Gradle build so it passes correct class path to forbidden apis.
A similar problem exists with ecj, which is disabled at moment (🤪). In future Ecj should add the API jars also to class path / module patching to work correct.
Errorprone works because it is part of java compiler, thanks for adding it in a previous PR.

uschindler:
It's too late for me today. Im really tired and I am a bit frustrated. Maybe tomorrow.

uschindler:
I think easiest is to add the APIJAR to the standard compile class path of the source set. It still needs to be given as module patch for the compiler but adding it to the normal compile class path should work.

Maybe we can also run ecj with similar tricks. Will look into this separately.

rmuir:
Yeah if we can get it working we can just cancel this one.

I want to start with banning just the fma and compress, then do an "audit" of the vector api to look for more issues. 

I don't see the issue going away, I don't think performance is a top priority with the API designs in java. Evidence: Math.fma

dweiss:
Let me know if you figure out what's wrong with the build, Uwe. I'm on holidays but if it's problematic I'll have a long flight home next weekend and can certainly take a look then.

dweiss:
I think this is what you need?
https://github.com/apache/lucene/pull/14942

Feel free to take anything or modify anything there - I won't be able to help out tomorrow.

uschindler:
New PR: #14942 

I will close this one.

PR #14942, opened 2025-07-13
Fail 'gradle check' on unchecked slow vector api feature usage (by using forbiddenapis)

dweiss:
Followup to implement the checks of #14939  with forbiddenapis.

### DIFF
--- a/build-tools/build-infra/src/main/groovy/lucene.java.core.mrjar.gradle
+++ b/build-tools/build-infra/src/main/groovy/lucene.java.core.mrjar.gradle
@@ -1,5 +1,3 @@
-import de.thetaphi.forbiddenapis.gradle.CheckForbiddenApis
-
 /*
  * Licensed to the Apache Software Foundation (ASF) under one or more
  * contributor license agreements.  See the NOTICE file distributed with
@@ -16,6 +14,8 @@ import de.thetaphi.forbiddenapis.gradle.CheckForbiddenApis
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+import de.thetaphi.forbiddenapis.gradle.CheckForbiddenApis
 
 // Produce an MR-JAR for panama vector implementations
 
--- a/build-tools/build-infra/src/main/groovy/lucene.java.core.mrjar.gradle
+++ b/build-tools/build-infra/src/main/groovy/lucene.java.core.mrjar.gradle
@@ -1,3 +1,5 @@
+import de.thetaphi.forbiddenapis.gradle.CheckForbiddenApis
+
 /*
  * Licensed to the Apache Software Foundation (ASF) under one or more
  * contributor license agreements.  See the NOTICE file distributed with
@@ -29,9 +31,13 @@ configure(project(":lucene:core")) {
       configurations["main${jdkVersion}Implementation"].extendsFrom(configurations.maybeCreate("implementation"))
       dependencies.add("main${jdkVersion}Implementation", sourceSets.main.output)
 
-      tasks.named("compileMain${jdkVersion}Java").configure {
-        def apijar = apijars.file("jdk${jdkVersion}-api.jar")
+      def apijar = apijars.file("jdk${jdkVersion}-api.jar")
 
+      tasks.withType(CheckForbiddenApis).named("forbiddenApisMain${jdkVersion}").configure {
+        it.classpath = it.classpath + files(apijar)
+      }
+
+      tasks.named("compileMain${jdkVersion}Java").configure {
         // TODO: this depends on the order of argument configuration...
         int releaseIndex = options.compilerArgs.indexOf("--release")
         options.compilerArgs.removeAt(releaseIndex)
--- a/gradle/validation/ast-grep/tests/java-patterns.yml
+++ b/gradle/validation/ast-grep/tests/java-patterns.yml
@@ -102,3 +102,15 @@ valid:
     public abstract class X extends TokenizerFactory {}
   - |
     private class X extends TokenizerFactory {}
+---
+id: incubator-module-dependency
+invalid:
+  - |
+    module org.apache.lucene.foo {
+      requires jdk.incubator.something;
+    }
+valid:
+  - |
+    module org.apache.lucene.foo {
+      requires org.apache.lucene.bar;
+    }
--- a/gradle/validation/ast-grep/tests/java-patterns.yml
+++ b/gradle/validation/ast-grep/tests/java-patterns.yml
@@ -109,6 +109,7 @@ invalid:
     module org.apache.lucene.foo {
       requires jdk.incubator.something;
     }
+  - import jdk.incubator.something.SomethingElse;
 valid:
   - |
     module org.apache.lucene.foo {
--- a/gradle/validation/forbidden-apis/non-standard/incubator-vector.txt
+++ b/gradle/validation/forbidden-apis/non-standard/incubator-vector.txt
@@ -1,4 +1,7 @@
-@defaultMessage Potentially slow on some CPUs, please check the CPU has feature: FMA may fallback to BigDecimal
+# Some vector APIs are only fast on specific hardware, and fallback to very slow
+# pure-java implementations. List them here, to prevent traps.
+
+@defaultMessage Potentially slow on some CPUs, please check Constants.HAS_FAST_VECTOR_FMA: FMA may fallback to BigDecimal
 jdk.incubator.vector.Float16#fma(**)
 jdk.incubator.vector.FloatVector#fma(**)
 jdk.incubator.vector.DoubleVector#fma(**)
--- a/gradle/validation/forbidden-apis/non-standard/incubator-vector.txt
+++ b/gradle/validation/forbidden-apis/non-standard/incubator-vector.txt
@@ -1,0 +1,17 @@
+@defaultMessage Potentially slow on some CPUs, please check the CPU has feature: FMA may fallback to BigDecimal
+jdk.incubator.vector.Float16#fma(**)
+jdk.incubator.vector.FloatVector#fma(**)
+jdk.incubator.vector.DoubleVector#fma(**)
+jdk.incubator.vector.VectorOperators#FMA
+
+@defaultMessage Potentially slow on some CPUs, please check the CPU has feature: Unsupported on NEON
+jdk.incubator.vector.ByteVector#compress(**)
+jdk.incubator.vector.IntVector#compress(**)
+jdk.incubator.vector.ShortVector#compress(**)
+jdk.incubator.vector.LongVector#compress(**)
+jdk.incubator.vector.VectorOperators#COMPRESS_BITS
+jdk.incubator.vector.ByteVector#expand(**)
+jdk.incubator.vector.IntVector#expand(**)
+jdk.incubator.vector.ShortVector#expand(**)
+jdk.incubator.vector.LongVector#expand(**)
+jdk.incubator.vector.VectorOperators#EXPAND_BITS
--- a/lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java
+++ b/lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java
@@ -78,6 +78,7 @@ final class PanamaVectorUtilSupport implements VectorUtilSupport {
   }
 
   // the way FMA should work! if available use it, otherwise fall back to mul/add
+  @SuppressForbidden(reason = "Uses FMA only where fast and carefully contained")
   private static FloatVector fma(FloatVector a, FloatVector b, FloatVector c) {
     if (Constants.HAS_FAST_VECTOR_FMA) {
       return a.fma(b, c);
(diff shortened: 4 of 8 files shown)


lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java
```diff
@@ -50,6 +50,7 @@
  *
  * Setting these properties will make this code run EXTREMELY slow!
  */
+@SuppressForbidden(reason = "vectors module is allowed here")
```
rmuir: thanks, now that we're scanning (thank you!), I will see if I can tweak this, to try to use a different set of rules somehow for the java24?

I'd like to avoid suppressing entire classes, instead i'd like to disable specific trappy methods such as `VectorOperators.FMA`, and only `@SuppressForbidden` the `fma()` helper method that properly checks the CPU capabilities...

The idea would be to just ban `VectorOperators.FMA` with a message that instructs you it might be slow on some hardware, please use the `fma()` helper method instead.

Same goes for compress/expand.

Hope this makes sense, I'm not trying to police `src/java`, I'm trying to police `src/java24`

uschindler:
Looks fine. Maybe let me try the other variant: add APIJAR to normal class path of whole source set instead. This variant works. Yesterday I was a bit afraid if it works at all die to lazy evaluation of the class path.

lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java
uschindler: I wonder why this complains at all. Did you commit any vector forbids?

uschindler (approved):

uschindler:
I have to look into this and fix the whole setup...
We can enable separate signatures files easily.

build-tools/build-infra/src/main/groovy/lucene.java.core.mrjar.gradle
```diff
@@ -29,9 +31,13 @@ configure(project(":lucene:core")) {
       configurations["main${jdkVersion}Implementation"].extendsFrom(configurations.maybeCreate("implementation"))
       dependencies.add("main${jdkVersion}Implementation", sourceSets.main.output)
 
-      tasks.named("compileMain${jdkVersion}Java").configure {
-        def apijar = apijars.file("jdk${jdkVersion}-api.jar")
+      def apijar = apijars.file("jdk${jdkVersion}-api.jar")
 
+      tasks.withType(CheckForbiddenApis).named("forbiddenApisMain${jdkVersion}").configure {
+        it.classpath = it.classpath + files(apijar)
```
uschindler, line 37: I would exchange order to habe the incubator API first on class path. But that's just nitpicking.

lucene/core/src/java24/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java
uschindler: Maybe I know, have to figure out after breakfast: I thought we disable the "jdk-nonportable" bundled signature here: https://github.com/apache/lucene/blob/ac45d992c2f6189d2a0bffc2c0c6effc87c5d370/build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/java/ApplyForbiddenApisPlugin.java#L117

But we don't do yet. I have the feeling that forbidden now complains because it sees blacklisted classes....

uschindler:
I figured out what's wrong:
```
> Task :lucene:core:forbiddenApisMain24 FAILED
Forbidden class use: jdk.internal.vm.vector.VectorSupport$Vector [non-portable or internal runtime class]
  in org.apache.lucene.internal.vectorization.MemorySegmentPostingDecodingUtil (MemorySegmentPostingDecodingUtil.java, method declaration of 'shift(jdk.incubator.vector.IntVector,int,int,int,int,int[],int,int)')
Forbidden class use: jdk.internal.vm.vector.VectorSupport$Vector [non-portable or internal runtime class]
Forbidden class use: jdk.internal.vm.vector.VectorSupport$Vector [non-portable or internal runtime class]
Forbidden class use: jdk.internal.vm.vector.VectorSupport$Vector [non-portable or internal runtime class]
Forbidden class use: jdk.internal.vm.vector.VectorSupport$Vector [non-portable or internal runtime class]
  in org.apache.lucene.internal.vectorization.PanamaVectorUtilSupport (PanamaVectorUtilSupport.java, method declaration of 'fma(jdk.incubator.vector.FloatVector,jdk.incubator.vector.FloatVector,jdk.incubator.vector.FloatVector)')
Scanned 21 class file(s) for forbidden API invocations (in 0.25s), 2 error(s).
```

The reason for this is because it treats the "apijar" file as part of the user APIs and therefor it can't detect the boundary between Java-Internal classes and Classpath classes. "jdk.internal is on the auto-blacklist). Because the superclass of Vector is somewhere behind the scenes extended from `jdk.internal.vm.vector.VectorSupport$Vector`.

The only workaround that exists is to remove the "jdk-nonportable" forbiddenapis signature. This would make the java24 code not checked againt internal APIs. But we have to live with that.

I will commit a change here and add the signatures @rmuir mentioned in the other issue.

So we have two possibilities:
- don't check vector API for forbidden apis at all
- remove "jdk-nonportable" bundled sigantures
- have allowlists in forbiddenapis, which is possible since forbiddenapis 3.9: `setSignaturesWithSeveritySuppress()` - but unfortunately I have to figure out of if that works with the non-portable bundled signatures.

uschindler:
> The solution for this would be to have allowlists in forbiddenapis, which is possible in forbiddenapis 3.9: `setSignaturesWithSeveritySuppress()` - but unfortunately I have to figure out of if that works with the non-portable bundled signatures.

Unfortunately this does not work with the auto-reported violations. I can maybe fix this in forbiddenapis, but at moment we have to remove the "non-portable" bundled signatures to get rid of above error.

I will try something else before that maybe I can add a workaround....

uschindler:
I have a workaround, the trick is to explicitely add a signature for the above reported "auto-failures" due to `jdk-nonportable`. After that set the signatures to "SUPPRESS". Will commit the fix.

After that we can add separate signatures copied from Robert's other PR.

I will still open an Issue on forbiddenapis that supressing signatures does not work in the current framework. Basically the setup of the whole new infratcructure is not that good. The "shitelist" signatures should be listed as normal signatures and whenever you have a violation, the violation is checked against the whitelist in the same way. Currently you can only suppress signatures which are also added before "as-is". That's the trick I use.

uschindler:
@dweiss: It would have been better to not add the SupressWarnings in this PR, because it would have been better to see the errors in the github actions. That would have given me the possibility to idetify the issues without laptop....

build-tools/build-infra/src/main/groovy/lucene.java.core.mrjar.gradle
uschindler: we can keep it as is...

uschindler:
Robert, I added your signatures. The build fails now:

```
> Task :lucene:core:forbiddenApisMain24 FAILED
Forbidden method invocation: jdk.incubator.vector.FloatVector#fma(**) [Potentially slow on some CPUs, please check the CPU has feature: FMA may fallback to BigDecimal]
  in org.apache.lucene.internal.vectorization.PanamaVectorUtilSupport (PanamaVectorUtilSupport.java:83)
Scanned 21 class file(s) for forbidden API invocations (in 0.30s), 1 error(s).
```

All fine now. It is up to you now to add suppress and check code is faine. But I think it now behaves as we want it to behave: Only one place with FMA is reported which is a static method with a check already!

uschindler (approved):
we just need to followup with ode changes and suppressions

uschindler:
I closed the old PR by Robert: #14939

uschindler:
I was also able to remove the "fail on missing classes = false" setting in forbiddenapis on java24. The classes are now provided by the apijar, so forbiddenapis should be happy!

dweiss:
I only added the configuration example and suppressions on everything so that you can see where to put the stick in to change it. Thanks for following up.

dweiss:
Also - feel free to take over and undraft/merge when you're ready. I won't be able to do it today.

uschindler:
> Also - feel free to take over and undraft/merge when you're ready. I won't be able to do it today.

I hand over to Robert. I think the only thing todo is to add a SuppressForbidden to the single static method affected. I just did not do this to see the message and that it works at all.

uschindler:
> I only added the configuration example and suppressions on everything so that you can see where to put the stick in to change it. Thanks for following up.

All fine! I just said, if you want to show a failure for somebody just watching the issue with webrowser, it is better to let the build system fail build and all is open on table :-)

One annoying thing: I spend several minutes here all the time when I changed a single line of code in the plugin code. Because when the build plugin changes the WHOLE downstream build needs to be repeated. This is very annoying. Why the hell does it need to compile core classes again when just the build plugin changes? I only changed one piece not affecting compilation. 

SO with new gradle build infrastrcuture it got much more annoying than before. Previously it was able to track back a change in one of the gradle files and not rebuild everything that was not touched at all. I have the feeling the issue is now that when it compiles one Java file or Gradle file form the build-plugin, the final JAR file changes so the dependency on the build on the plugin causes all task outputs to get invalidated. Very annoying!

uschindler:
Forbiddenapis Bug: https://github.com/policeman-tools/forbidden-apis/issues/267

uschindler:
As mentioned in the othe rPR, we should still ad a "git grep" pattern to fail if anybody adds some `jdk.incubator.*` module import into any `module-info.java`:

```
$ git grep -e '\bjdk\.incubator\b' **/module-info.java
```

this creates no positives yet. I know sometimes some devs on PRs try to do that in the benchmark-jmh module, so I'd add this simple grep. This is best to prevent any incubator usage anywhere (and if not, forbiddenapis will do it anyways).

rmuir:
> this creates no positives yet. I know sometimes some devs on PRs try to do that in the benchmark-jmh module, so I'd add this simple grep. This is best to prevent any incubator usage anywhere (and if not, forbiddenapis will do it anyways).

 https://github.com/apache/lucene/pull/14942/commits/7704155774d8f7722246bfd0610f2e2d824d5a5f

rmuir (approved):
Thank you! This is exactly what is needed to prevent the crazy 10x slowdowns.

As a followup we can try to "audit" the vector API to look for more trappy things to add to this list. 

Or at least as we encounter such traps, let's add them here to prevent groundhog day on every PR.

gradle/validation/ast-grep/tests/java-patterns.yml
```diff
@@ -102,3 +102,15 @@ valid:
     public abstract class X extends TokenizerFactory {}
   - |
     private class X extends TokenizerFactory {}
+---
+id: incubator-module-dependency
+invalid:
```
uschindler, line 107: Maybe add a check that it also triggers on import

gradle/validation/forbidden-apis/non-standard/incubator-vector.txt
```diff
@@ -0,0 +1,20 @@
+# Some vector APIs are only fast on specific hardware, and fallback to very slow
+# pure-java implementations. List them here, to prevent traps.
+
+@defaultMessage Potentially slow on some CPUs, please check Constants.HAS_FAST_VECTOR_FMA: FMA may fallback to BigDecimal
+jdk.incubator.vector.Float16#fma(**)
+jdk.incubator.vector.FloatVector#fma(**)
+jdk.incubator.vector.DoubleVector#fma(**)
+jdk.incubator.vector.VectorOperators#FMA
+
+@defaultMessage Potentially slow on some CPUs, please check the CPU has feature: Unsupported on NEON
```
uschindler, line 10: Maybe adapt message later

uschindler (approved):

gradle/validation/ast-grep/tests/java-patterns.yml
rmuir: It will work, but we can add a test, `scoped_identifier` is used anywhere you have `x.y.z` identifiers in the grammar.

https://github.com/tree-sitter/tree-sitter-java/blob/488557612b44bde9c5e8c3011138992a8e500c90/grammar.js#L841-L847
https://github.com/tree-sitter/tree-sitter-java/blob/488557612b44bde9c5e8c3011138992a8e500c90/grammar.js#L1003-L1007

gradle/validation/forbidden-apis/non-standard/incubator-vector.txt
rmuir: I think the message is ok for now: we don't yet have a field in Constants for this one, if you want to use the feature you'll need to add that.

uschindler (approved):

dweiss:
> SO with new gradle build infrastrcuture it got much more annoying than before. Previously it was able to track back a change in one of the gradle files and not rebuild everything that was not touched at all. I have the feeling the issue is now that when it compiles one Java file or Gradle file form the build-plugin, the final JAR file changes so the dependency on the build on the plugin causes all task outputs to get invalidated. Very annoying!

Yes, could be... I've noticed it too. I don't know the answer though. We have no control over this, unfortunately. We should try to get the build system to a stable state and then it should be a rare event. For now, we'll have to suffer.

PR #14944, opened 2025-07-14
Enable tests.haltonfailure (regression on main)

dweiss:
There are two confusing options that the gradle test task has: failfast and ignorefailures. One stops the build as soon as the first test failed (failfast), the other causes the test task to not fail the build, even if tests failed. I introduced a regression where the second was set to true, causing tests to never failed the build.

I don't see the benefit of having these two options. Rather, it should be that you fail the build early or you fail the build at the very end of processing if any test failed. I may clean this up later - for now, re-enabling failures on failed tests.


dweiss:
I'll allow myself to commit it in without a review.

PR #14946, opened 2025-07-14
build: fix some eclipse errors in build-infra

rmuir:
I don't think we are running ecj on these sources yet, but just clearing out the problems from my editor: very minor stuff such as missing @override and use of unnamed variables when lambda parameter is not used


rmuir:
I can try to investigate the ecj-lint as a followup. It is fast and if we can do it in `gradle check`, we can have some improved checks for the build sources.

dweiss:
It's a chicken and egg problem with these sources - it's inconceivably difficult to apply the build plugin to itself... That's why there is a "shadow" project in which I've tried to use the same set of source files. I'll take a look at why ecj isn't applied - I need to take a look.

dweiss:
https://github.com/apache/lucene/pull/14949 this should help with ecj-linting and possibly other verification tasks.

dweiss (approved):

PR #14949, opened 2025-07-15
Move srcdir redirection for build-infra-shadow earlier in the build

dweiss:
So that tasks that eagerly do something with the sourceset have a chance to work with the redirected path.

This makes ecj-lint run properly.


dweiss:
This fails on errors fixed in https://github.com/apache/lucene/pull/14946 so I'll wait for that to be merged in first.

PR #14953, opened 2025-07-15
[Draft] Multi-Tenant CMS Manager

yaser-aj:
A proposed solution to #13883.

Currently, each `ConcurrentMergeScheduler` manages its own threads independently, assuming no other merge schedulers are running in the JVM. This leads to potential over-allocation of system resources when many `IndexWriter`s are active. This draft PR demonstrates a simple approach to multi-tenancy by introducing a global manager that divides a global thread budget among all registered CMS instances.

### Key Changes

- New Class: `MultiTenantCMSManager`
    - Singleton manager that tracks all active `ConcurrentMergeScheduler` instances.
    - Divides a global thread budget (based on available CPU cores) equally among all registered CMSs.
    - Dynamically updates each CMS’s `maxThreadCount` and `maxMergeCount` as CMSs are registered or unregistered.
    - Includes testing hooks for registration and budget management.
- Integration with `ConcurrentMergeScheduler`
    - CMS now registers itself with the manager upon construction and unregisters on close.
    - CMS thread/merge limits are automatically managed by the global manager.
- Tests
    - Added unit tests for registration, dynamic rebalancing, live `IndexWriter` integration, and auto-unregistering behavior.

### Important Notes

- This is a demonstration/first step toward multi-tenant merge scheduling. The logic can be extended to support more advanced scheduling and dynamic resource allocation in the future.
- The manager currently uses a simple equal-share policy for thread allocation (total thread count divided by merge scheduler count).

Will update this draft PR with more points + to-dos soon.


lukewilner:
Continuing where @yaser-aj left off, here's what I plan to tackle next:

- Improve the thread sharing logic: Right now it's a simple equal split, but I want to explore smarter ways to assign threads based on actual activity or load.
- Stress test
- Clean up unregister behavior: Making sure CMS instances get removed safely even if they’re GC’d or closed suddenly.
- luceneutil benchmarks 

Please let us know if we are missing anything important. Thanks!

lucene/core/src/java/org/apache/lucene/index/MultiTenantCMSManager.java
```diff
@@ -0,0 +1,86 @@
...
+    schedulers.add(cms);
+    updateBudgets();
+  }
+
+  /**
+   * Unregisters a ConcurrentMergeScheduler from the global manager.
+   *
+   * @param cms the merge scheduler to unregister
+   */
+  public void unregister(ConcurrentMergeScheduler cms) {
```
vigyasharma, line 58: nit: should we call it "deregister"?

lucene/core/src/java/org/apache/lucene/index/MultiTenantCMSManager.java
```diff
@@ -0,0 +1,86 @@
...
+    schedulers.remove(cms);
+    updateBudgets();
+  }
+
+  private void updateBudgets() {
+    int count = schedulers.size();
+    if (count == 0) return;
+
+    int share = Math.max(1, maxThreadCount / count);
```
vigyasharma, line 67: This should be `Math.min()`, otherwise, we'll set some schedulers to 0 maxMergeThreads.

lucene/core/src/test/org/apache/lucene/index/TestMultiTenantCMSManager.java
```diff
@@ -0,0 +1,158 @@
...
+  /** Test basic registration and budgeting across multiple schedulers. */
+  public void testRegistrationAndThreadBudgeting() {
+    MultiTenantCMSManager manager = MultiTenantCMSManager.getInstance();
+    manager.unregisterAllForTest();
+
+    final int numSchedulers = 4;
+    ConcurrentMergeScheduler[] cmsArr = new ConcurrentMergeScheduler[numSchedulers];
+    for (int i = 0; i < numSchedulers; i++) {
+      cmsArr[i] = new ConcurrentMergeScheduler();
+      manager.register(cmsArr[i]);
```
vigyasharma, line 60: We already register in the CMS ctor. Do we need to register again?

lucene/core/src/java/org/apache/lucene/index/MultiTenantCMSManager.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  // -----------------------------------------
+  // 🧪 TESTING HOOKS
+  // -----------------------------------------
+
+  /** Used in tests to read the current registered CMS set */
+  synchronized Set<ConcurrentMergeScheduler> getRegisteredSchedulersForTest() {
+    return new HashSet<>(schedulers);
```
vigyasharma, line 79: Maybe just return a 1Collections.unmodifableSet()` ?  and avoid allocating a new set and copy all the entries into it..

lucene/core/src/java/org/apache/lucene/index/MultiTenantCMSManager.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  /** Used in tests to clear all registered CMS instances */
+  synchronized void unregisterAllForTest() {
```
vigyasharma, line 83: Why does this method need to be `synchronized` ?

vigyasharma (commented):
Thanks Yaser, and Luke, for sharing a prototype of this approach. The simplicity of tweaking things from outside while leaving CMS untouched, is indeed tempting!

I do agree with @jpountz however that 

1. The simplest version (equal sharing) is too limited.
2. As we start building more elaborate logic, things get very complex very quickly. 
   * As a small example, an equal sharing strategy doesn't account for some schedulers getting more merges than others. Once you start doing that, `updateBudget()` must be called each time a merge is submitted or completed, which needs deeply integrated hooks.
    * The real complexity in designing and maintaining, lies in figuring out the right policy budget allocation policy.
5. It doesn't provide a sure way of controlling the total no. of threads consumed by indexing+merging, which is a [primary motivation](https://github.com/apache/lucene/pull/14900#issuecomment-3084596936) for such a scheduler.

Hence, I'm leaning more in the direction of #14900, at least as of now. Happy to hear more thoughts on this.

Regardless, there's a lot of good work here, especially in the tests, which can be leveraged in any impl. we pursue for this change.

PR #14963, opened 2025-07-17
Bypass HNSW graph building for tiny segments

### LINKED ISSUE #13447: Explore bypassing HNSW graph building for tiny segments

shubhamvishu:
### Description

This change avoids creating a HNSW graph if the segment is small (here we have taken the thresholdfor number of vectors  as `10000` based on the conversation [here](https://github.com/apache/lucene/issues/13447#issuecomment-3004565232) for now). 

Some of the points I'm not sure how we would want to go about :
- All the tests passes currently since the option to enable the optimization is `false` by default but setting it to `true` reveals some failing unit tests which inherently assumes that the HNSW graph is created and KNN search is triggered (do we have some idea of how to bypass those in some good clean way?)
- I understand we might want to always keep this optimization on (also less invasive change), but for now in this PR, I made it configurable and enabled it on the KNN format - just to be cautious (wasn't sure if it would not affect back-compact in some unknown way), but happy to make it as default behaviour

&nbsp;
**TODOs:**
- Add specific unit tests
- Benchmarks (luceneutil)

&nbsp;
Closes https://github.com/apache/lucene/issues/13447


lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java
```diff
@@ -137,9 +144,16 @@ public final class Lucene99HnswVectorsFormat extends KnnVectorsFormat {
   private final int numMergeWorkers;
   private final TaskExecutor mergeExec;
 
+  /**
+   * Whether to bypass HNSW graph building for tiny segments (below {@link #HNSW_GRAPH_THRESHOLD}).
+   * When enabled, segments with fewer than the threshold number of vectors will store only flat
+   * vectors, significantly improving indexing performance for workloads with frequent flushes.
+   */
+  private final boolean bypassTinySegments;
```
benwtrent: If we allow this to be a parameter, it should be a threshold that refers to the typical `k` used when querying.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java
```diff
@@ -326,16 +335,19 @@ private void search(
     final KnnCollector collector =
         new OrdinalTranslatedKnnCollector(knnCollector, scorer::ordToDoc);
     final Bits acceptedOrds = scorer.getAcceptOrds(acceptDocs);
-    HnswGraph graph = getGraph(fieldEntry);
-    boolean doHnsw = knnCollector.k() < scorer.maxOrd();
+    boolean doHnsw =
+        knnCollector.k() < scorer.maxOrd()
+            && (bypassTinySegments == false
+                || fieldEntry.size() > Lucene99HnswVectorsFormat.HNSW_GRAPH_THRESHOLD);
```
benwtrent: The reader should just look to see if there is a graph.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -632,8 +695,23 @@ public void addValue(int docID, T vectorValue) throws IOException {
                 + "\" appears more than once in this document (only one value is allowed per field)");
       }
       flatFieldVectorsWriter.addValue(docID, vectorValue);
-      scorer.setScoringOrdinal(node);
-      hnswGraphBuilder.addGraphNode(node, scorer);
+      // Check if we need to initialize graph builder for tiny segment optimization
+      if (bypassTinySegments
+          && graphBuilderInitialized == false
+          && node >= Lucene99HnswVectorsFormat.HNSW_GRAPH_THRESHOLD) {
```
benwtrent: I am thinking we should just use the `expectedNodeVisited` logic for the given node count (treating like its a "graph") vs. the tiny segment threshold number (which should be used as a `k`).

benwtrent (commented):
Some minor ideas.

benwtrent:
I think 10K is likely way too large 10k vector ops vs. `10 * log(10k)` ops is a huge difference. 

If a user typical searches for 10 nearest neighbors, the graph should be built at around 90 vectors

msokolov:
I can think of a circumstance where we might create small segments that will probably never get searched at all, but will very quickly be merged. In that case we might want to allow a larger threshold?

msokolov:
As far as the tests are concerned I'm confused, wouldn't they fail with a high threshold since we wouldn't build a graph until there are many documents? Maybe I didn't understand the meaning of the threshold though.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java
shubhamvishu: Makes sense

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
shubhamvishu: I like the idea, I'll change the predicate to use that logic

shubhamvishu:
@msokolov Actually they do, I had missed setting `byPassTinySegments=true` locally in one of the constructors so tests didn't exercise that path. Setting it to true does reveal the failing unit tests. 

I'll try if there is some clean way to override those checks in the failing unit tests.

jpountz:
I wonder how this interacts with how `AbstractKnnVectorQuery` does pre-filtering by first passing the filter to `KnnVectorsReader#search`, and then falling back to an exact search. If the segment doesn't have a HNSW graph, this may effectively start an exact search (via  `KnnVectorsReader#search`) and then abort it to do an exact search again? Or am I missing something?

benwtrent:
> I can think of a circumstance where we might create small segments that will probably never get searched at all, but will very quickly be merged. In that case we might want to allow a larger threshold?

I think that is fine. I am thinking of semi-nrt with lots of updates. In cases like that 10k is way too big a default. I think the value should be used as an input to expectedVisitedNodes that takes into account the potential graph size. 

Additionally, I would assume users would want to scale quantized formats vs. non-quantized differently (as their vector ops can be much cheaper than floating point ops).

> and then abort it to do an exact search again? Or am I missing something?

I would hope the format just does the right thing, and searches everything, knowing that there isn't a graph.

shubhamvishu:
@jpountz Ahh, I see what you are pointing towards and here is what think we could try maybe :

- We currently also fallback to exact search after the visitedLimit is breached in HNSW search, so now that same visited limit would be applicable when we are iterating over the docs i.e. net-net `approximateKnn (visit V nodes) + exactSearch` ~== `exactSearch (visit V nodes linearly) + exactSearch` which I might not impact the search time?. So one way is to gulp this since we will visit small no. of docs but I agree we can further optimize this path (more on this below points)
- We could completely remove the fallback to exactSearch in `AbstractKnnVectorQuery` and we could relax the check from
     - `if (knnCollector.earlyTerminated())` to 
     - `if (knnCollector instanceof TimeLimitingKnnCollectorManager.TimeLimitingKnnCollector && 
              ((TimeLimitingKnnCollectorManager.TimeLimitingKnnCollector)knnCollector).shouldExit())` after making `TimeLimitingKnnCollector` public and exposing `shouldExit()`
 
  This would ensure we continue the exact search `VectorsReader` and don't fallback to exactSearch in `AbstractKnnVectorQuery`. (we can do better maybe, more on it below)
- [PROPOSED] Though I think `AbstractKnnVectorQuery#exactSearch` is better with exact search since it uses a conjunctive `DocIdSetIterator` rather than iterating on all the docs?. If yes, then for this we could maybe simply add an `else if` condition in VectorsReader to straightaway overwhelm the collector (forcing its `earlyTerminated` to return true) and return so it automatically fallsback to best exactSearch impl (I hope that gives us best of both worlds?)
```
    else if (getGraph(fieldEntry).equals(HnswGraph.EMPTY)) {
      // MakesFallback to exactSearch directly
      knnCollector.incVisitedCount((int) knnCollector.visitLimit() + 1);
    }
```

Let me know your thoughts or if I'm missing something here. Thanks!

jpountz:
My recommendation would be to move the logic of switching to an exact search when the filter is selective to `KnnVectorsReader#search` (in a separate PR) so that the file format can make the right decision depending on whether it only has a flat index or something more sophisticated such as a HNSW index. (It doesn't feel completely straight forward since `KnnVectorsReader#search` may not know how to pull an efficient iterator that matches the same docs at the `Bits acceptDocs`).

benwtrent:
This makes me wonder if the knn search method should accept a ScorerSupplier and the  live docs Bits instead of fully realized bit set that represent both the filter and live docs....

jpountz:
Or some higher-level abstraction that can either be consumed in a random-access fashion (`Bits`) or sequential (`DocIdSetIterator`)?

```java
class AcceptDocs {

  /** Random access to the accepted documents. */
  Bits getBits();

  /** Get an iterator of accepted docs. */
  DocIdSetIterator getIterator();

  /** Return an approximation of the number of accepted documents. */
  long cost();
}
```

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -598,10 +629,21 @@ static FieldWriter<?> create(
...
         throws IOException {
       this.fieldInfo = fieldInfo;
-      RandomVectorScorerSupplier scorerSupplier =
+      this.M = M;
+      this.beamWidth = beamWidth;
+      this.infoStream = infoStream;
+      this.bypassTinySegments = bypassTinySegments;
+      this.flatFieldVectorsWriter = Objects.requireNonNull(flatFieldVectorsWriter);
+      if (bypassTinySegments) {
+        this.bufferedVectors = new ArrayList<>();
```
vigyasharma: Since we only store upto `HNSW_GRAPH_THRESHOLD` no. of vectors, beyond which we resume the regular flow of adding them to the graph, we could use an array here instead of an ArrayList?

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -632,8 +695,23 @@ public void addValue(int docID, T vectorValue) throws IOException {
...
+      // Check if we need to initialize graph builder for tiny segment optimization
+      if (bypassTinySegments
+          && graphBuilderInitialized == false
+          && node >= Lucene99HnswVectorsFormat.HNSW_GRAPH_THRESHOLD) {
+        initializeGraphBuilder();
+        // Replay buffered vectors
+        replayBufferedVectors();
+        bufferedVectors.clear();
+      }
+      if (hnswGraphBuilder != null) {
```
vigyasharma: Does `hnswGraphBuilder != null` do the same thing as `graphBuilderInisialized` ? if so, do we need `graphBuilderInisialized` ?

vigyasharma (commented):
How do we handle backward compatibility in this change? I noticed we don't write any metadata (e.g. in `FieldEntry`) about `bypassTinySegments` or whether a graph was built or not. The flag gets configured when the format is initialized from the codec.

What happens if I create an index with `bypassTinySegments=true`, but later read it in an application with the flag set to false? I think we need to persist information about whether graph was built for the segment.

vigyasharma:
> I think we need to persist information about whether graph was built for the segment.

Maybe we could use one of the existing fields that describe the [graph](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java#L436). Like set `numLevels=0` when there is no graph (otherwise it would at least be 1)?

shubhamvishu:
> Or some higher-level abstraction that can either be consumed in a random-access fashion (Bits) or sequential (DocIdSetIterator)?

Thanks @jpountz. I opened https://github.com/apache/lucene/pull/15011 for adding such abstraction

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
shubhamvishu: Makes sense

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
shubhamvishu: Oh yes, we can drop this check

shubhamvishu:
Thanks for the review @vigyasharma!

> What happens if I create an index with bypassTinySegments=true, but later read it in an application with the flag set to false? I think we need to persist information about whether graph was built for the segment.

> Maybe we could use one of the existing fields that describe the graph. Like set numLevels=0 when there is no graph (otherwise it would at least be 1)?

We write the [vectorIndexLength here](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java#L469) as the meta info which would be 0 in the case there is no HNSW graph so we could use this to determine if we want to do graph search or not (update the `doHnsw` check with this). So, I think there shouldn't be any issue with the backward compatibility and need to store this bypassing info explicitly? 

Let me know if missing something here maybe.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java
msokolov: I do wonder if we would want to expose as a parameter though?  Maybe it should just be a fixed value?  I would have thought about setting it based on a threshold where exhaustive search is no-or-only-slightly more expensive than hnsw search?  I would expect this to be related to the M of the graph maybe?

jpountz: I'd like to have it at least as a parameter to a pkg-protected constructor, so that we can pass random values in `TestLucene99HnswVectorsFormat` to make sure that we exercise properly both the case when there is a graph and the case when there is no graph.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java
```diff
@@ -115,6 +115,13 @@ public final class Lucene99HnswVectorsFormat extends KnnVectorsFormat {
   /** Default to use single thread merge */
   public static final int DEFAULT_NUM_MERGE_WORKER = 1;
 
+  /**
+   * Threshold below which HNSW graph building is bypassed for tiny segments. Segments with fewer
+   * vectors will use flat storage only, improving indexing performance when having frequent
+   * flushes.
+   */
+  public static final int HNSW_GRAPH_THRESHOLD = 10_000;
```
jpountz: I think that the comment should try to expand a bit more on this value to help future readers think through whether it's still right or whether it should be updated.

One thing we discussed on the linked issue is that the number of visited nodes is in the order of `log(size) * k`. So having a graph only helps if `log(size) * k << size` <=> `size / log(size) >> k`. If we arbitrarily choose k = 100, 10,000 is the first power of 10 so that `size / log(size)` is one order of magnitude greater than k (10/log(10) ~= 4.3, 100/log(100) ~= 22, 1000/log(1000) ~= 144, 10000 / log(10000) ~= 1085).

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java
jpountz: It may be good to be able to pass random values in RandomCodec as well to help with the test coverage of the approximate case (we have very few tests that index more than 10k vectors).

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java
shubhamvishu: Sure, I'll elaborate on this.

msokolov:
ooh, it's nice to avoid force-push -- recommended way to keep up with main is `git merge main` into your branch

shubhamvishu:
> ooh, it's nice to avoid force-push -- recommended way to keep up with main is git merge main into your branch

Agreed, sometimes I end up doing `git commit --amend --no-edit` to merge staged changes with the top commit. It changes the commit id and doesn't allow to push cleanly (I should probably train my muscle mem to stop doing that)

shubhamvishu:
OK, I ran the `luceneutil` benchmarks and I see a huge improvement in the indexing throughput with this PR compared to baseline(without this change). I see an almost **`4x`** improvement in the indexing rate/time here. Looking forward to your thoughts. Thanks!

*Note : The improvement in the latency or CPUTime seems to be driven by the less no. of segments which also very slightly impact the recall as we know.*

CC - @benwtrent @msokolov @jpountz @vigyasharma 

#### With `HNSW_GRAPH_THRESHOLD` = 10

**Baseline**
```
Results:
 0.515       11.084  11.071        0.999  500000   100      50       64        250     4 bits    120.51       4149.14             7         1690.12      1649.857      185.013       HNSW
 0.875        9.965   9.946        0.998  500000   100      50       64        250     7 bits    119.18       4195.23             3         1871.00      1832.962      368.118       HNSW
 0.978       19.637  19.621        0.999  500000   100      50       64        250         no    127.63       3917.51             8         1501.98      1464.844     1464.844       HNSW
 ```
 
**Candidate**
```
Results:
 0.517        5.933   5.914        0.997  500000   100      50       64        250     4 bits     30.34      16482.07             2         1694.34      1649.857      185.013       HNSW
 0.871        9.655   9.635        0.998  500000   100      50       64        250     7 bits     27.86      17945.59             3         1869.52      1832.962      368.118       HNSW
 0.961       11.280  11.269        0.999  500000   100      50       64        250         no     31.16      16046.73             3         1503.46      1464.844     1464.844       HNSW
 ```
 
 --------------------------
 
 #### With `HNSW_GRAPH_THRESHOLD` = 100
 
**Baseline**
```
Results:
 0.515       11.101  11.074        0.998  500000   100      50       64        250     4 bits    118.33       4225.40             7         1690.02      1649.857      185.013       HNSW
 0.874       10.199  10.176        0.998  500000   100      50       64        250     7 bits    118.18       4230.83             3         1871.16      1832.962      368.118       HNSW
 0.977       19.990  19.979        0.999  500000   100      50       64        250         no    126.92       3939.61             8         1501.95      1464.844     1464.844       HNSW
```
**Candidate**
```
Results:
 0.517        5.900   5.882        0.997  500000   100      50       64        250     4 bits     30.13      16596.96             2         1694.14      1649.857      185.013       HNSW
 0.872        9.680   9.660        0.998  500000   100      50       64        250     7 bits     28.58      17495.98             3         1869.05      1832.962      368.118       HNSW
 0.964       11.640  11.619        0.998  500000   100      50       64        250         no     27.25      18349.97             3         1502.64      1464.844     1464.844       HNSW
 ```

**Comment by shubhamvishu** [comment]
To verify that this isn't a red herring, I deliberately increased the `HNSW_GRAPH_THRESHOLD` to `1000000`, effectively preventing the creation of any HNSW graphs (as confirmed by the very low indexing time and the absence of graph layers in the structure). As expected, latency increased significantly due to a fallback to exact search. This validates that the earlier results with `HNSW_GRAPH_THRESHOLD` set to `10` or `100` represent a win-win situation i.e. we achieve ~4x faster indexing without compromising on latency (in fact, latency improves due to fewer segments).

#### Candidate (with `HNSW_GRAPH_THRESHOLD` = 1000000)
```
:
.
.
:
Leaf 0 has 0 layers
Leaf 0 has 309324 documents
Leaf 1 has 0 layers
Leaf 1 has 154373 documents
Leaf 2 has 0 layers
Leaf 2 has 36303 documents

Results:
recall  latency(ms)   netCPU  avgCpuCount    nDoc  topK  fanout  maxConn  beamWidth  quantized  index(s)  index_docs/s  num_segments  index_size(MB)  vec_disk(MB)  vec_RAM(MB)  indexType
 0.514       89.174   89.153        1.000  500000   100      50       64        250     4 bits     26.93      18565.96             3         1651.81      1649.857      185.013       HNSW
 0.890      223.908  223.798        1.000  500000   100      50       64        250     7 bits     26.93      18569.41             3         1834.91      1832.962      368.118       HNSW
 1.000      140.186  140.176        1.000  500000   100      50       64        250         no     24.19      20667.99             3         1466.79      1464.844     1464.844       HNSW
 ```

mikemccand:
These look like amazing results!  It's surprising that we see such a gain in indexing throughput?  Maybe `knnPerfTest.py` in luceneutil is waiting for all merges to finish in the end?  But, I thought we recently added an option to not wait.

It might also be that faster merging (not creating HNSW graphs for smallish segments) means more CPU to spend indexing instead of merging, or possibly it changes the backpressure (Lucene will pause indexing threads if merging is falling behind because it sees the indexing threads as evil since they keep producing new segments).

Are there any issues we still need to resolve here?

benwtrent:
Hey @shubhamvishu sorry, this fell off my radar!

Those benchmark numbers are pretty crazy!

I can give a fuller review if you deem it ready :). 

Some initial feed back:

 - readers shouldn't ever know about this new parameter. They should just check to see if there is a graph. If not, be smart about it. I am not 100% sure how this will behave with BWC and not a new format. But intuitively it SHOULD work as the reader just needs to check if the graph is there or not. The on disk format isn't really changing...
 - I am confused by this `String.length()*10` thing. Why not just use numbers directly?

All the conflicts need updating :/

I would also like a benchmark with many random commits (e.g. simulating random commits and searches) to see how it all behaves. I still slightly worry about NRT and tiny segments. Intuitively, they should be fine as we would likely bypass the graph anyways on search, but I have learned never to trust my intuition when I can benchmark it :)

benwtrent:
> in luceneutil is waiting for all merges to finish in the end?

It used to for sure. I think its waiting for the merges so it doesn't kick off the searches while merges are occurring. Did that behavior change recently?

I saw a recent PR about removing "merge time" from "index time". I think we are displaying both merge vs. index time separately now right?

ChrisHegarty:
>readers shouldn't ever know about this new parameter. They should just check to see if there is a graph. If not, be smart about it. I am not 100% sure how this will behave with BWC and not a new format. But intuitively it SHOULD work as the reader just needs to check if the graph is there or not. The on disk format isn't really changing...

This is my understanding also. For example, it is possible for a graph.size to be 0 today and for the index to still have vectors. So I think that this is compatible.

benwtrent:
I have handled the merge conflicts, and adjusted somethings locally. I am gonna test to see if I messed anythign up.

Would you mind me pushing the changes @shubhamvishu ?

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java
```diff
@@ -326,16 +335,16 @@ private void search(
...
     final Bits acceptedOrds = scorer.getAcceptOrds(acceptDocs);
-    HnswGraph graph = getGraph(fieldEntry);
-    boolean doHnsw = knnCollector.k() < scorer.maxOrd();
+    boolean doHnsw = knnCollector.k() < scorer.maxOrd() && fieldEntry.vectorIndexLength > 0;
     // Take into account if quantized? E.g. some scorer cost?
     int filteredDocCount = 0;
     // The approximate number of vectors that would be visited if we did not filter
-    int unfilteredVisit = HnswGraphSearcher.expectedVisitedNodes(knnCollector.k(), graph.size());
+    int unfilteredVisit =
+        HnswGraphSearcher.expectedVisitedNodes(knnCollector.k(), fieldEntry.size());
```
ChrisHegarty: This is a necessary change. I think that this is correct, but someone should verify that graph.size was not actually explicitly intended here. Or if it should be fieldEntry.size if graph.size is 0 (EMPTY)

ChrisHegarty (commented):
Generally, I think that this PR is going in the right direction. Let's get the remaining outstanding things addressed and move it forward.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -550,17 +573,28 @@ static int distFuncToOrd(VectorSimilarityFunction func) {
...
 
     private final FieldInfo fieldInfo;
-    private final HnswGraphBuilder hnswGraphBuilder;
+    private HnswGraphBuilder hnswGraphBuilder; // only created when needed
     private int lastDocID = -1;
     private int node = 0;
     private final FlatFieldVectorsWriter<T> flatFieldVectorsWriter;
     private UpdateableRandomVectorScorer scorer;
+    private final int graphThreshold;
+    private final List<T> bufferedVectors;
```
kaivalnp: I wonder if we can avoid this duplicate list of buffered vectors? The local `FlatFieldVectorsWriter` already stores these vectors, and exposes them via [`getVectors()`](https://github.com/apache/lucene/blob/e706267b893576cd334a783e6dfa8b4008cdc7b2/lucene/core/src/java/org/apache/lucene/codecs/hnsw/FlatFieldVectorsWriter.java#L35)

shubhamvishu:
> readers shouldn't ever know about this new parameter. They should just check to see if there is a graph. If not, be smart about it. I am not 100% sure how this will behave with BWC and not a new format. But intuitively it SHOULD work as the reader just needs to check if the graph is there or not. The on disk format isn't really changing...

@benwtrent Makes sense, we can remove it. We are anyways not using in anywhere on the reader. I also think there should be no issue with BWC. 


> I am confused by this String.length()*10 thing. Why not just use numbers directly?

This was based on the [above comment](https://github.com/apache/lucene/pull/14963#discussion_r2248862786) from Adrien about choosing the graph creation threshold from `k` as where 1 order of magnitude higher.


> All the conflicts need updating :/

Yes, I was occupied with relocation. I'll spend sometime to fix those and have new revision.

> I saw a recent PR about removing "merge time" from "index time". I think we are displaying both merge vs. index time separately now right?

We started reporting the index time without including the merge time now but haven't added the merge time reporting. Thats was on the cards, I'll push a PR also start emitting merge time numbers,

benwtrent:
@shubhamvishu I got it all fixed locally and can push if you like

shubhamvishu:
@benwtrent Ohh wonderful, please feel free. I'm more than happy.

benwtrent:
@shubhamvishu I pushed up a commit:

 - I removed the string logic thing...we can adjust that back if you want to. But, we should likely just do `Math.log10` if that is what we care about.
 - I removed the buffered vectors thing (just using the underlying vectors)
 - adjusted the api and removed the input to the reader (I think)
 - fixed the merges, it compiles and runs, maybe its all good? Will need to double check everything

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -575,16 +612,27 @@ static int distFuncToOrd(VectorSimilarityFunction func) {
     throw new IllegalArgumentException("invalid distance function: " + func);
   }
 
+  private static boolean shouldCreateGraph(int k, int numNodes) {
+    int expectedVisitedNodes =
+        expectedVisitedNodes(k, numNodes); // k is typically small, so this is cheap
+    return numNodes > expectedVisitedNodes;
```
shubhamvishu: One thing to note: the nice improvement we saw in benchmarks was with `expectedVisitedNodes(k, numNodes) * 10`, so that the nodes threshold is at least 1 order higher than what `expectedVisitedNode` returns. I don't expect this to change the luceneutil numbers but still something to keep in mind. We could confirm with another run if this is not having any negative impact on performance numbers.

shubhamvishu:
@benwtrent  Thank you Ben, really appreciate you jumping in to make those changes.

> I removed the string logic thing...we can adjust that back if you want to. But, we should likely just do Math.log10 if that is what we care about.

Yes, I agree. We can confirm once if that is not having any negative impact.

> fixed the merges, it compiles and runs, maybe its all good? Will need to double check everything

Yeah, only other thing I think left is fixing the tests. They all assume that there will a HNSW graph. I don't know if there is any cleaner way to fix these (other than plumbing each of those) but I'll look into fixing the tests from here.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
benwtrent: >  We could confirm with another run if this is not having any negative impact on performance numbers.

Yeah, its definitely changing performance. It removes the magnitude requirement. Also, I noticed it has a bug, I am fixing that and will push shortly.

shubhamvishu:
@benwtrent I fixed the failing tests in the latest commit. They were making assumptions that we always do HNSW and since many of those don't create high cardinality graphs it was causing failures. One way would have been to just increase the number of docs for some of them which create many vectors(but that might have increased the time those take to test; but did that with just one test I think). Others I mostly updated to check if there would be graph present or not. There are some refactoring opportunities in the testing I could think of like to test more random behaviors like when this optimization OFF vs ON (default is on) and some more simplifications but for now I didn't focus much on that and kept it mostly to make all tests happy in the most straightforward way. Maybe we can make these tests more random in a followup change? Looking for your feedback. I'll also post the latest upstream numbers shortly with the latest changes.

lucene/test-framework/src/java/org/apache/lucene/tests/util/LuceneTestCase.java
```diff
@@ -3277,4 +3278,9 @@ public Executor getIntraMergeExecutor(MergePolicy.OneMerge merge) {
       return intraMergeExecutor;
     }
   }
+
+  public static boolean hasGraphPresent(int k, int size) {
+    int expectedVisitedNodes = expectedVisitedNodes(k, size);
+    return size > expectedVisitedNodes && expectedVisitedNodes > 0;
+  }
```
benwtrent: This is very trappy. This sort of assumes that the test case knows the logic that makes a graph exist or not.

Something better might be adjust `TestUtil.getDefaultCodec()` to set the threshold for building the graph to `0` for now.

lucene/CHANGES.txt
```diff
@@ -140,7 +140,9 @@ New Features
 * GITHUB#15176: Add `[Float|Byte]VectorValues#rescorer(element[])` interface to allow optimized rescoring of vectors.
   (Ben Trent)
 
-* GITHUB#15169: Add codecs for 4 and 8 bit Optimized Scalar Quantization vectors (Trevor McCulloch)
+* GITHUB#15169, GITHUB#15223: Add codecs for 4 and 8 bit Optimized Scalar Quantization vectors. The new format
+  `Lucene104HnswScalarQuantizedVectorsFormat` replaces the now legacy `Lucene99HnswScalarQuantizedVectorsFormat`
```
benwtrent: there was a bit of an issue with main branch recently and it needed to be restored. So the commits might be out of alignment. I think your branch needs to have main merged back in or something to fix the changes.

benwtrent (commented):
@shubhamvishu could you confirm the benchmark results?

When I tried benchmarking, I didn't get anywhere near the numbers you got. Would be good to know the exact settings, etc. tested so we can replicate and see where the benefits are for this.

lucene/test-framework/src/java/org/apache/lucene/tests/util/LuceneTestCase.java
shubhamvishu: In that case we will be testing not with this optimization disabled but in general it would be on by default. Maybe we can add a separate test case to test this and leave others to be disabled on the codec as you mentioned? I'll make those changes.

shubhamvishu:
> could you confirm the benchmark results?

@benwtrent Sure, I'll share the results with the latest changes soon.

&nbsp;
> When I tried benchmarking, I didn't get anywhere near the numbers you got. Would be good to know the exact settings, etc. tested so we can replicate and see where the benefits are for this.

Below are the settings I tried for the above results if I'm not mistaken. Also as we discussed earlier, I tried with a little higher threshold(mentioned below) so that is also one thing I would like to confirm the impact on numbers as we have changed it now. Let me know if there is anything else I might be missing. Thanks!

Dataset : Cohere
```
dim = 768
doc_vectors = f"{constants.BASE_DIR}/data/cohere-wikipedia-docs-5M-{dim}d.vec"
query_vectors = f"{constants.BASE_DIR}/data/cohere-wikipedia-queries-{dim}d.vec"
```

Graph Threshold 
```
private static int graphCreationThreshold(int k, int numNodes) {
  return (int)
      Math.pow(10, String.valueOf(HnswGraphSearcher.expectedVisitedNodes(k, numNodes)).length());
}
```

```
PARAMS = {
  "ndoc": (500_000,),
  "maxConn": (64,),
  "beamWidthIndex": (250,),
  "fanout": (50,),
  "numMergeWorker": (12,),
  "numMergeThread": (4,),
  "numSearchThread": (0,),
  "encoding": ("float32",),
  "quantizeBits": (
    4,
    7,
    32,
  ),
  "topK": (100,),
  "quantizeCompress": (True,),
  "queryStartIndex": (0,),
  "forceMerge": (False,),
}
```

lucene/test-framework/src/java/org/apache/lucene/tests/util/LuceneTestCase.java
benwtrent: > Maybe we can add a separate test case to test this and leave others to be disabled on the codec as you mentioned? I'll make those changes.

I think maybe for tests that assume there MUST be a graph, they have codec settings that enforce such. For all other tests, it can be the default.

ChrisHegarty (approved):

shubhamvishu:
@benwtrent I reran the benchmarks and somehow I don't see the old performance difference like earlier and rather similar to baseline. Maybe the recent developments had some indexing improvements (like eg I see 4 bit recall is much better now due to OSQ?). Net-net, against the current main we are not creating those many tiny segments in our luceneutil benchmarks by default. Maybe we see still see higher benefits in older Lucene version probably?

However, to simulate a more realtime update scenario by changing luceneutil `KnnIndexer` to use [setMaxBufferedDocs(100)](https://github.com/mikemccand/luceneutil/blob/main/src/main/knn/KnnIndexer.java#L102) to trigger more flushes and re-ran the benchmarks and I see the below results where indexing gets **~30%** faster with 1000 as the threshold(configurable) in that particular case. So it seems the user could configure a value(1 order higher than topk at query time?) accordingly that helps avoid the overhead of creating tiny graphs. @benwtrent @ChrisHegarty I patched the test related changes in the latest commit, let me know if this looks good or not. Thanks!

**Baseline :**
```
Results:
recall  latency(ms)  netCPU  avgCpuCount    nDoc  topK  fanout  maxConn  beamWidth  quantized  visited  index(s)  index_docs/s  num_segments  index_size(MB)  vec_disk(MB)  vec_RAM(MB)  indexType
 0.901        5.941   5.923        0.997  500000   100      50       64        250     4 bits    31624     84.64       5907.09             4         1690.88      1655.579      190.735       HNSW
 0.965       13.023  13.012        0.999  500000   100      50       64        250     7 bits    40750    127.19       3931.00             7         1871.98      1838.684      373.840       HNSW
 0.971        9.997   9.986        0.999  500000   100      50       64        250         no    40086    103.35       4837.93             7         1498.13      1464.844     1464.844       HNSW
```
**Candidate (HNSW_GRAPH_THRESHOLD = 100) [DEFAULT]:**
```
Results:
recall  latency(ms)  netCPU  avgCpuCount    nDoc  topK  fanout  maxConn  beamWidth  quantized  visited  index(s)  index_docs/s  num_segments  index_size(MB)  vec_disk(MB)  vec_RAM(MB)  indexType
 0.898        4.598   4.587        0.998  500000   100      50       64        250     4 bits    23503     88.53       5647.55             2         1692.61      1655.579      190.735       HNSW
 0.964       13.309  13.291        0.999  500000   100      50       64        250     7 bits    41490    122.04       4097.02             8         1871.81      1838.684      373.840       HNSW
 0.964        7.809   7.788        0.997  500000   100      50       64        250         no    27908    104.95       4764.08             3         1501.41      1464.844     1464.844       HNSW
```

**Candidate (HNSW_GRAPH_THRESHOLD = 1000) :**
```
Results:
recall  latency(ms)  netCPU  avgCpuCount    nDoc  topK  fanout  maxConn  beamWidth  quantized  visited  index(s)  index_docs/s  num_segments  index_size(MB)  vec_disk(MB)  vec_RAM(MB)  indexType
 0.901        6.737   6.720        0.997  500000   100      50       64        250     4 bits    37284     58.96       8480.61             6         1689.44      1655.579      190.735       HNSW
 0.959       10.027  10.015        0.999  500000   100      50       64        250     7 bits    31131     86.87       5755.46             4         1874.15      1838.684      373.840       HNSW
 0.963        7.463   7.452        0.999  500000   100      50       64        250         no    27939     72.11       6933.85             3         1501.16      1464.844     1464.844       HNSW
```

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene99/Lucene99HnswScalarQuantizedVectorsFormat.java
```diff
@@ -63,9 +64,24 @@ public class Lucene99HnswScalarQuantizedVectorsFormat extends KnnVectorsFormat {
   /** The format for storing, reading, merging vectors on disk */
   private final FlatVectorsFormat flatVectorsFormat;
 
+  /**
```
benwtrent: I don't think the backward_codecs need changing. This value is only applied at `write` time, and consequently will never be used in backwards codecs.

lucene/core/src/java/org/apache/lucene/codecs/lucene102/Lucene102HnswBinaryQuantizedVectorsFormat.java
```diff
@@ -64,9 +65,17 @@ public class Lucene102HnswBinaryQuantizedVectorsFormat extends KnnVectorsFormat
   private final int numMergeWorkers;
   private final TaskExecutor mergeExec;
 
+  /**
```
benwtrent: We should also move this one to backwards codecs (I didn't do that yet, my bad...). All HNSW quantized logic is now handled in the single 104 format. Could you just update the 104 format and leave this one alone?

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104HnswScalarQuantizedVectorsFormat.java
```diff
@@ -117,6 +152,7 @@ public Lucene104HnswScalarQuantizedVectorsFormat(
     }
     this.maxConn = maxConn;
     this.beamWidth = beamWidth;
+    this.tinySegmentsThreshold = tinySegmentsThreshold;
```
benwtrent, line 163: Could you apply this value to the writer? Seems like you missed that part here.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java
```diff
@@ -332,18 +332,18 @@ private void search(
...
+    HnswGraph graph = getGraph(fieldEntry);
     int filteredDocCount = Math.min(acceptDocs.cost(), graph.size());
     Bits accepted = acceptDocs.bits();
     final Bits acceptedOrds = scorer.getAcceptOrds(accepted);
     int numVectors = scorer.maxOrd();
     boolean doHnsw = knnCollector.k() < numVectors;
     // The approximate number of vectors that would be visited if we did not filter
     int unfilteredVisit = HnswGraphSearcher.expectedVisitedNodes(knnCollector.k(), graph.size());
-    if (unfilteredVisit >= filteredDocCount) {
+    if (unfilteredVisit >= filteredDocCount || graph.size() == 0) {
```
benwtrent, line 346: I love how simple this is!

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -575,16 +612,26 @@ static int distFuncToOrd(VectorSimilarityFunction func) {
     throw new IllegalArgumentException("invalid distance function: " + func);
   }
 
+  private static boolean shouldCreateGraph(int k, int numNodes) {
+    int expectedVisitedNodes = expectedVisitedNodes(k, numNodes);
+    return (numNodes > expectedVisitedNodes && expectedVisitedNodes > 0) || k < 0;
```
benwtrent: I think the `|| k < 0` is unnecessary. But if you want to exit early, I would simply do an `if (k == 0) return true;`

It seems nicer to me for our "always build graph" to be `0`. That makes more sense than allowing a negative number.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java
```diff
@@ -152,7 +177,20 @@ public Lucene99HnswVectorsFormat() {
...
+    this(maxConn, beamWidth, DEFAULT_NUM_MERGE_WORKER, null, HNSW_GRAPH_THRESHOLD, VERSION_CURRENT);
+  }
+
+  /**
+   * Constructs a format using the given graph construction parameters.
+   *
+   * @param maxConn the maximum number of connections to a node in the HNSW graph
+   * @param beamWidth the size of the queue maintained during graph construction.
+   * @param tinySegmentsThreshold the value of k for the expectedVisitedNodes heuristic, used to
+   *     determine the minimum required graph nodes
```
benwtrent: Can we be clearer about the allowed values and what they mean.

It seems to me that:

 - `0` indicates that the graph is always built
 - `> 0` indicates that the graph requires more nodes to be added before its built
 - Negative values aren't allowed.

Also, let's not say "expectedVisitedNodes" heuristic, I am not sure that is actually documented anywhere. Let's simply say "the expected number of vector operations to return `k` nearest neighbors of the current graph size".

benwtrent (commented):
This is looking really good.

Thank you for doing the more active flushing test, that is exactly what I am looking for!

I think keeping it simple and allowing users to set `k` to their "expected values" or indicating that they may want it to be higher (to be more lazy). 

Could you add a CHANGES.txt entry for 10.4 Optimizations?

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene99/Lucene99HnswScalarQuantizedVectorsFormat.java
shubhamvishu: Sounds good!

lucene/core/src/java/org/apache/lucene/codecs/lucene102/Lucene102HnswBinaryQuantizedVectorsFormat.java
shubhamvishu: Sure!

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104HnswScalarQuantizedVectorsFormat.java
shubhamvishu: Oops, I'll add that!

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
shubhamvishu: `|| k <0` was required since when passing `-1` the `expectedVisitedNodes` was returning a negative value making this function to eventually return false with -1 threshold which was exactly opposite of what we intended for when using -1. But just allowing 0 or higher would make this simpler as you said below.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java
shubhamvishu: The wording makes sense to me, I'll rephrase

shubhamvishu:
Thanks for the review @benwtrent. I addressed the comments and added the CHANGES entry now.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java
ChrisHegarty: ++

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsFormat.java
```diff
@@ -116,6 +116,18 @@ public final class Lucene99HnswVectorsFormat extends KnnVectorsFormat {
...
+   * Threshold which HnswGraphSearcher#expectedVisitedNodes uses as k to determine when HNSW graph
+   * building is bypassed (useful in case if frequent flushes). It is in terms of k for a graph i.e.
+   * number of docs to match for the query. So having a graph only helps if,
+   *
+   * <pre> k &lt;&lt; size / log(size) </pre>
+   *
+   * i.e. k is at least 1 order less than size / log(size) where size if the number of nodes in the
+   * graph
+   */
+  public static final int HNSW_GRAPH_THRESHOLD = 100;
```
ChrisHegarty, line 134: I struggled a little with this description. And also would find it hugely beneficial to have a centralised location describing this optimisation. I'll make a concrete suggestion, but please take it as just that and reword/rephrase as you see fit. And please add corrections if I missed something.

```
 /**
   * Minimum estimated search effort (in terms of expected visited nodes)
   * required before building an HNSW graph for a segment.
   *
   * <p>This threshold is compared against the value produced by
   * {@link HnswGraphSearcher#expectedVisitedNodes(int, int)}, which estimates
   * how many nodes would be visited during a vector search based on the current
   * graph size and {@code k} (neighbours to find).
   *
   * <p>If the estimated number of visited nodes falls below this threshold,
   * HNSW graph construction is skipped for that segment - typically for small
   * flushes or low document count segments - since the overhead of building
   * the graph would outweigh its search benefits.
   *
   * <p>Conceptually, this corresponds to cases where:
   *
   * <pre>{@code
   *   k << size / log(size)
   * }</pre>
   *
   * meaning that the target {@code k} (number of nearest neighbors) is at least
   * an order of magnitude smaller than {@code size / log(size)}, where
   * {@code size} is the number of nodes (vectors) in the segment.
   *
   * <p>Default: {@code 100}
   */
  public static final int HNSW_GRAPH_THRESHOLD = 100;
```

benwtrent: I am fine with this description, however I don't think we do this magnitude check anymore. That is up to the user to just apply a larger k.

shubhamvishu: Yeah, we don't have this check anymore but in general the math is still correct. But if adds any confusion, we can omit this part instead?

ChrisHegarty: D'oh! I see that now. Thanks. Proceed as you see fit.

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104HnswScalarQuantizedVectorsFormat.java
```diff
@@ -60,6 +61,19 @@ public class Lucene104HnswScalarQuantizedVectorsFormat extends KnnVectorsFormat
...
   private final Lucene104ScalarQuantizedVectorsFormat flatVectorsFormat;
 
+  /**
+   * The threshold to use to bypass HNSW graph building for tiny segments in terms of k for a graph
+   * i.e. number of docs to match the query (default is {@link
+   * Lucene99HnswVectorsFormat#HNSW_GRAPH_THRESHOLD}).
+   *
+   * <ul>
+   *   <li>0 indicates that the graph is always built.
+   *   <li>0 indicates that the graph needs certain or more nodes before it starts building.
```
benwtrent:
```suggestion
   *   <li>greater than 0 indicates that the graph needs a certain number of nodes before it starts building. See {@link Lucene99HnswVectorsFormat#HNSW_GRAPH_THRESHOLD} for details.
```

benwtrent (approved):

ChrisHegarty (approved):
LGTM

Issue #14148, opened 2025-01-17
Add an optional bandwidth cap to `TieredMergePolicy`?


mikemccand:
### Description

TL;DR: `TieredMergePolicy` can create massive snapshots if you configure it for aggressive `deletesPctAllowed`, which can hurt searchers (cause page fault storms) in a near-real-time replication world.  Maybe we could add an optional (off by default) "rate limit" on how many amortized bytes/sec TMP is merging?  This is just an idea / brainstorming / design discussion so far ... no PR.

Full context:

At Amazon (Product Search team) we use [near-real-time segment replication](https://blog.mikemccandless.com/2017/09/lucenes-near-real-time-segment-index.html) to efficiently distribute index updates to all searchers/replicas.

Since we have many searchers per indexer shard, to scale to very high QPS, we intentionally tune `TieredMergePolicy` (TMP) to very aggressively reclaim deletions.  Burning extra CPU / bandwidth during indexing to save even a little bit of CPU during searching is a good tradeoff for us (and in general, for Lucene users with high QPS requirements).

But we have a "fun" problem occasionally: sometimes we have an update storm (an upstream team reindexes large-ish portions of Amazon's catalog through the real-time indexing stream), and this leads to lots and lots of merging and many large (max-sized 5 GB) segments being replicated out to searchers in short order, sometimes over links (e.g. cross-region) that are not as crazy-fast as within-region networking fabric, and our searchers fall behind a bit.

Falling behind is not the end of the world: the searchers simply skip some point-in-time snapshots and jump to the latest one,  effectively sub-sampling checkpoints as best they can given the bandwidth constraints.  Index-to-search latency is hurt a bit, but recovers once the indexer catches up on the update storm.

The bigger problem for us is that we size our shards, roughly, so that the important parts of the index (the parts hit often by query traffic) are fully "hot".  I.e. so the OS has enough RAM to hold the hot parts of the index.  But when it takes too long to copy and light (switch over to the new segments for searching) a given snapshot, and we skip the next one or two snapshots, the followon snapshot that we finally do load may have a sizable part of the index rewritten, and the snapshot size maybe a big percentage of the overall index, and copying/lighting it will stress the OS into a paging storm, hurting our long-pole latencies.

[...]

PR #14964, opened 2025-07-17
Add BandwidthCappedMergeScheduler for enforcing a global merge bandwidth cap

nipunbatra8:
# Add BandwidthCappedMergeScheduler for enforcing a global merge bandwidth cap

This draft PR introduces a prototype `BandwidthCappedMergeScheduler`, which extends `ConcurrentMergeScheduler` to enforce a global bandwidth cap on merge operations across all active merges within an `IndexWriter`. The scheduler is inspired by the issue/discussion in [lucene#14148](https://github.com/apache/lucene/issues/14148) and feedback from @mikemccand and others. The motivation is to provide a simple, global bandwidth cap for merge operations, especially useful during "update storms" or "war time" scenarios where aggressive merging can cause page faults.

## Implementation

- **Introduces `BandwidthCappedMergeScheduler` in `lucene.index`**
- **Extends `ConcurrentMergeScheduler` to reuse merge management and threading**
- **Implements a global bandwidth cap by dividing a configurable MB/s limit among all active merges**
- **Overrides `updateMergeThreads()` to dynamically adjust per-merge IO rates**

Initially, I experimented with skipping merges in the scheduler (by aborting or refusing to run merges that would exceed the bandwidth cap). However, this approach proved problematic:
- **Resource Leaks:** By the time the scheduler can skip a merge, resources/files may already be allocated, leading to leaks or inconsistent state.
- **Merge Policy Coordination:** Skipping merges late in the scheduler can result in better merges being blocked or waiting indefinitely, and it’s too late to make an informed decision without communicating with the `IndexWriter` or the `MergePolicy`.
- **Complexity:** Trying to abort merges at the scheduler level introduces complexity and potential for subtle bugs, especially around resource cleanup and merge queue management.

## CMS IO Rate Limiter?

- **ConcurrentMergeScheduler (CMS):** Applies adaptive IO throttling to each merge thread individually, so total bandwidth may exceed a set limit when multiple merges run.
- **BandwidthCappedMergeScheduler:** Divides a single global bandwidth cap among all active merges, ensuring their combined IO rate never exceeds the configured maximum.

### Should We Integrate with CMS IO Rate Limiter?

One open question: should we integrate CMS’s adaptive IO rate logic into this scheduler as well, or something like switch to the global cap only during "war time", such as:
- Use CMS’s adaptive per-merge throttling during "peace time" (normal operation).
- Switch to the global bandwidth cap when the system detects "war time" (e.g., max merges/threads reached, or backlog detected).

## Future Improvements

For now, the implementation is intentionally simple. However, there are several ways to make it more efficient and fair:
1. **Sort merges by size:** Prioritize smaller merges or allocate bandwidth proportionally to merge size.
2. **Priority-based throttling:** Give more bandwidth to urgent merges (e.g., forceMerge, high delete reclaim).
3. **Dynamic feedback:** Adjust cap/rates based on system load actually consumed by merges
4. **Merge policy integration:** Communicate bandwidth usage back to the `MergePolicy` to influence which merges are selected, or even skip merges earlier in the process.

## Testing

- I have tested this scheduler by replacing CMS in `LuceneTestCase.java` and running the full test suite.
- Next, I plan to edit `NRTPerfTest` to simulate update storms and then generate segment traces to visualize bandwidth usage and merge behavior. Hopefully see a less spiky segment tracing graph.
- **Request for feedback:** Are there other suggestions for tests or benchmarks to monitor this?

## Next Steps

- **Discuss with the community**
- **Testing to visualize results as mentioned in the Testing section**
- **Efficiency improvements:** Explore more advanced bandwidth allocation strategies as described above in future improvements.
- **Merge policy integration:** Investigate ways to skip or prioritize merges earlier, possibly by editing `IndexWriter` or implementing a `BandwidthAwareTieredMergePolicy` that can respond to bandwidth constraints in real time. Thoughts on this?

---

**Thanks for reviewing! Looking forward to feedback, suggestions, and further discussion. This PR is opened to propose and evaluate a bandwidth-capped merge scheduler design, and to gather feedback for further development.**

Related to #14148

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
+++ b/lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
@@ -99,7 +99,7 @@ public class BandwidthCappedMergeScheduler extends ConcurrentMergeScheduler {
         activeMerges++;
       }
     }
-    
+
     // Use the effective max thread count to avoid counting threads that CMS has paused
     int effectiveMaxThreads = getMaxThreadCount();
     if (effectiveMaxThreads == ConcurrentMergeScheduler.AUTO_DETECT_MERGES_AND_THREADS) {
@@ -107,7 +107,7 @@ public class BandwidthCappedMergeScheduler extends ConcurrentMergeScheduler {
       effectiveMaxThreads = Math.max(1, coreCount / 2);
     }
     int divisor = Math.min(effectiveMaxThreads, activeMerges);
-    
+
     double perMergeRate;
     if (divisor > 0) {
       perMergeRate = Math.max(0.1, bandwidthMbPerSec / divisor); // Use 0.1 MB/s minimum
--- a/lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
+++ b/lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
@@ -21,10 +21,11 @@ import org.apache.lucene.index.MergePolicy.OneMerge;
 
 /**
  * A {@link MergeScheduler} that extends {@link ConcurrentMergeScheduler} with bandwidth tracking
- * and limiting capabilities. This scheduler maintains a bandwidth rate bucket that is divided
- * among active merges. When bandwidth is limited, merges are throttled to preserve system resources.
+ * and limiting capabilities. This scheduler maintains a bandwidth rate bucket that is divided among
+ * active merges. When bandwidth is limited, merges are throttled to preserve system resources.
  *
- * <p>Key features: Global bandwidth rate bucket with configurable capacity, dynamic per-merge throttling.
+ * Key features: Global bandwidth rate bucket with configurable capacity, dynamic per-merge
+ * throttling.
  *
  * @lucene.experimental
  */
@@ -48,9 +49,7 @@ public class BandwidthCappedMergeScheduler extends ConcurrentMergeScheduler {
     super();
   }
 
-  /**
-   * Set the global bandwidth rate bucket in MB/s (default 1000 MB/s)
-   */
+  /** Set the global bandwidth rate bucket in MB/s (default 1000 MB/s) */
   public void setBandwidthRateBucket(double mbPerSec) {
     if (mbPerSec < MIN_MERGE_MB_PER_SEC || mbPerSec > MAX_MERGE_MB_PER_SEC) {
       throw new IllegalArgumentException(
@@ -64,13 +63,12 @@ public class BandwidthCappedMergeScheduler extends ConcurrentMergeScheduler {
     updateMergeThreads();
   }
 
-  /**
-   * Get the global bandwidth rate bucket in MB/s
-   */
+  /** Get the global bandwidth rate bucket in MB/s */
   public double getBandwidthRateBucket() {
     return bandwidthRateBucket;
   }
 
+  /** Distributes the global bandwidth rate bucket evenly among all active merge threads. */
   @Override
   protected synchronized void updateMergeThreads() {
     super.updateMergeThreads();
@@ -86,6 +84,8 @@ public class BandwidthCappedMergeScheduler extends ConcurrentMergeScheduler {
                 MIN_MERGE_MB_PER_SEC,
                 Math.min(MAX_MERGE_MB_PER_SEC, bandwidthRateBucket / activeMerges))
             : Double.POSITIVE_INFINITY;
+
+    // Apply the calculated rate limit to each active merge thread
     for (MergeThread mergeThread : mergeThreads) {
       if (mergeThread.isAlive()) {
         mergeThread.rateLimiter.setMBPerSec(perMergeRate);
@@ -93,12 +93,14 @@ public class BandwidthCappedMergeScheduler extends ConcurrentMergeScheduler {
     }
   }
 
+  /** Creates a custom merge thread with bandwidth tracking capabilities. */
   @Override
   protected synchronized MergeThread getMergeThread(MergeSource mergeSource, OneMerge merge)
       throws IOException {
     return new BandwidthTrackingMergeThread(mergeSource, merge);
   }
 
+  /** Returns a string representation including the current bandwidth rate bucket setting. */
   @Override
   public String toString() {
     return getClass().getSimpleName()
@@ -112,10 +114,12 @@ public class BandwidthCappedMergeScheduler extends ConcurrentMergeScheduler {
   /** Merge thread that logs the rate limiter value after merge completes. */
   protected class BandwidthTrackingMergeThread extends MergeThread {
     private final double mergeBandwidthMB;
+
     public BandwidthTrackingMergeThread(MergeSource mergeSource, OneMerge merge) {
       super(mergeSource, merge);
       this.mergeBandwidthMB = merge.estimatedMergeBytes / (1024.0 * 1024.0);
     }
+
     @Override
     public void run() {
       long startTime = System.currentTimeMillis();
@@ -129,12 +133,6 @@ public class BandwidthCappedMergeScheduler extends ConcurrentMergeScheduler {
                   + " MB)");
         }
         super.run(); // IO throttling is handled by the RateLimiter
-        // System.out.println(
-        //     "[BandwidthCappedMergeScheduler] Merge thread for segment "
-        //         + getSegmentName(merge)
-        //         + " ran at "
-        //         + String.format(java.util.Locale.US, "%.2f", this.rateLimiter.getMBPerSec())
-        //         + " MB/s");
       } finally {
         long duration = System.currentTimeMillis() - startTime;
         if (verbose()) {
--- a/lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
+++ b/lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
@@ -115,9 +115,7 @@ public class BandwidthCappedMergeScheduler extends ConcurrentMergeScheduler {
   protected class BandwidthTrackingMergeThread extends MergeThread {
     private final double mergeBandwidthMB;
 
-    /**
-     * Creates a new BandwidthTrackingMergeThread for the given merge.
-     */
+    /** Creates a new BandwidthTrackingMergeThread for the given merge. */
     public BandwidthTrackingMergeThread(MergeSource mergeSource, OneMerge merge) {
       super(mergeSource, merge);
       this.mergeBandwidthMB = merge.estimatedMergeBytes / (1024.0 * 1024.0);
--- a/lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
+++ b/lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
@@ -1,0 +1,160 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.index;
+
+import java.io.IOException;
+import org.apache.lucene.index.MergePolicy.OneMerge;
+
+/**
+ * A {@link MergeScheduler} that extends {@link ConcurrentMergeScheduler} with bandwidth tracking
+ * and limiting capabilities. This scheduler maintains a bandwidth rate bucket that is divided
+ * among active merges. When bandwidth is limited, merges are throttled to preserve system resources.
+ *
+ * <p>Key features: Global bandwidth rate bucket with configurable capacity, dynamic per-merge throttling.
+ *
+ * @lucene.experimental
+ */
+// nocommit
+public class BandwidthCappedMergeScheduler extends ConcurrentMergeScheduler {
+
+  /** Floor for IO write rate limit (we will never go any lower than this) */
+  private static final double MIN_MERGE_MB_PER_SEC = 5.0;
+
+  /** Ceiling for IO write rate limit (we will never go any higher than this) */
+  private static final double MAX_MERGE_MB_PER_SEC = 10240.0;
+
+  /** Initial value for IO write rate limit */
+  private static final double START_MB_PER_SEC = 1000.0;
+
+  /** Global bandwidth rate bucket in MB/s */
+  private double bandwidthRateBucket = START_MB_PER_SEC;
+
+  /** Default constructor with 1000 MB/s bandwidth rate bucket */
+  public BandwidthCappedMergeScheduler() {
+    super();
+  }
+
+  /**
+   * Set the global bandwidth rate bucket in MB/s (default 1000 MB/s)
+   */
+  public void setBandwidthRateBucket(double mbPerSec) {
+    if (mbPerSec < MIN_MERGE_MB_PER_SEC || mbPerSec > MAX_MERGE_MB_PER_SEC) {
+      throw new IllegalArgumentException(
+          "Bandwidth rate must be between "
+              + MIN_MERGE_MB_PER_SEC
+              + " and "
+              + MAX_MERGE_MB_PER_SEC
+              + " MB/s");
+    }
+    this.bandwidthRateBucket = mbPerSec;
+    updateMergeThreads();
+  }
+
+  /**
+   * Get the global bandwidth rate bucket in MB/s
+   */
+  public double getBandwidthRateBucket() {
+    return bandwidthRateBucket;
+  }
+
+  @Override
+  protected synchronized void updateMergeThreads() {
+    super.updateMergeThreads();
+    int activeMerges = 0;
+    for (MergeThread mergeThread : mergeThreads) {
+      if (mergeThread.isAlive()) {
+        activeMerges++;
+      }
+    }
+    double perMergeRate =
+        activeMerges > 0
+            ? Math.max(
+                MIN_MERGE_MB_PER_SEC,
+                Math.min(MAX_MERGE_MB_PER_SEC, bandwidthRateBucket / activeMerges))
+            : Double.POSITIVE_INFINITY;
+    for (MergeThread mergeThread : mergeThreads) {
+      if (mergeThread.isAlive()) {
+        mergeThread.rateLimiter.setMBPerSec(perMergeRate);
+      }
+    }
+  }
+
+  @Override
+  protected synchronized MergeThread getMergeThread(MergeSource mergeSource, OneMerge merge)
+      throws IOException {
+    return new BandwidthTrackingMergeThread(mergeSource, merge);
+  }
+
+  @Override
+  public String toString() {
+    return getClass().getSimpleName()
+        + ": "
+        + super.toString()
+        + ", bandwidthRateBucket="
+        + bandwidthRateBucket
+        + " MB/s";
+  }
+
+  /** Merge thread that logs the rate limiter value after merge completes. */
+  protected class BandwidthTrackingMergeThread extends MergeThread {
+    private final double mergeBandwidthMB;
+    public BandwidthTrackingMergeThread(MergeSource mergeSource, OneMerge merge) {
+      super(mergeSource, merge);
+      this.mergeBandwidthMB = merge.estimatedMergeBytes / (1024.0 * 1024.0);
+    }
+    @Override
+    public void run() {
+      long startTime = System.currentTimeMillis();
+      try {
+        if (verbose()) {
+          message(
+              "Starting bandwidth-capped merge: "
+                  + getSegmentName(merge)
+                  + " (estimated="
+                  + mergeBandwidthMB
+                  + " MB)");
+        }
+        super.run(); // IO throttling is handled by the RateLimiter
+        // System.out.println(
+        //     "[BandwidthCappedMergeScheduler] Merge thread for segment "
+        //         + getSegmentName(merge)
+        //         + " ran at "
+        //         + String.format(java.util.Locale.US, "%.2f", this.rateLimiter.getMBPerSec())
+        //         + " MB/s");
+      } finally {
+        long duration = System.currentTimeMillis() - startTime;
+        if (verbose()) {
+          double mbPerSec = mergeBandwidthMB / Math.max(duration / 1000.0, 0.001);
+          message(
+              "Merge completed: "
+                  + getSegmentName(merge)
+                  + " "
+                  + mergeBandwidthMB
+                  + " MB in "
+                  + duration
+                  + "ms ("
+                  + String.format(java.util.Locale.US, "%.2f", mbPerSec)
+                  + " MB/s)");
+        }
+      }
+    }
+  }
+
+  private static String getSegmentName(OneMerge merge) {
+    return merge.info != null ? merge.info.info.name : "_na_";
+  }
+}
--- a/lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
+++ b/lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
@@ -24,7 +24,7 @@ import org.apache.lucene.index.MergePolicy.OneMerge;
  * and limiting capabilities. This scheduler maintains a bandwidth rate bucket that is divided among
  * active merges. When bandwidth is limited, merges are throttled to preserve system resources.
  *
- * Key features: Global bandwidth rate bucket with configurable capacity, dynamic per-merge
+ * <p>Key features: Global bandwidth rate bucket with configurable capacity, dynamic per-merge
  * throttling.
  *
  * @lucene.experimental
--- a/lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
+++ b/lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
@@ -20,52 +20,73 @@ import java.io.IOException;
 import org.apache.lucene.index.MergePolicy.OneMerge;
 
 /**
- * A {@link MergeScheduler} that extends {@link ConcurrentMergeScheduler} with bandwidth tracking
- * and limiting capabilities. This scheduler maintains a bandwidth rate bucket that is divided among
- * active merges. When bandwidth is limited, merges are throttled to preserve system resources.
- *
- * <p>Key features: Global bandwidth rate bucket with configurable capacity, dynamic per-merge
- * throttling.
+ * A {@link MergeScheduler} that caps total IO write bandwidth across all running merges to a
+ * specified max MB/sec bandwidth.
  *
  * @lucene.experimental
  */
 // nocommit
 public class BandwidthCappedMergeScheduler extends ConcurrentMergeScheduler {
 
-  /** Floor for IO write rate limit (we will never go any lower than this) */
-  private static final double MIN_MERGE_MB_PER_SEC = 5.0;
+  /** Global bandwidth cap in MB/s. Mutable so that updates are applied live. */
+  private double bandwidthMbPerSec;
 
-  /** Ceiling for IO write rate limit (we will never go any higher than this) */
-  private static final double MAX_MERGE_MB_PER_SEC = 10240.0;
-
-  /** Initial value for IO write rate limit */
-  private static final double START_MB_PER_SEC = 1000.0;
-
-  /** Global bandwidth rate bucket in MB/s */
-  private double bandwidthRateBucket = START_MB_PER_SEC;
-
-  /** Default constructor with 1000 MB/s bandwidth rate bucket */
-  public BandwidthCappedMergeScheduler() {
-    super();
-  }
-
-  /** Set the global bandwidth rate bucket in MB/s (default 1000 MB/s) */
-  public void setBandwidthRateBucket(double mbPerSec) {
-    if (mbPerSec < MIN_MERGE_MB_PER_SEC || mbPerSec > MAX_MERGE_MB_PER_SEC) {
-      throw new IllegalArgumentException(
-          "Bandwidth rate must be between "
-              + MIN_MERGE_MB_PER_SEC
-              + " and "
-              + MAX_MERGE_MB_PER_SEC
-              + " MB/s");
+  /** Create a scheduler with a required global bandwidth cap in MB/s. */
+  public BandwidthCappedMergeScheduler(double bandwidthMbPerSec) {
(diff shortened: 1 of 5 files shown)


lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * <p>Key features: Global bandwidth rate bucket with configurable capacity, dynamic per-merge
+ * throttling.
+ *
+ * @lucene.experimental
+ */
+// nocommit
+public class BandwidthCappedMergeScheduler extends ConcurrentMergeScheduler {
+
+  /** Floor for IO write rate limit (we will never go any lower than this) */
```
stefanvodita: It will be interesting to see how we tweak these constants based on the tests we do. Would it make sense to make them configurable?

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+        activeMerges > 0
+            ? Math.max(
+                MIN_MERGE_MB_PER_SEC,
+                Math.min(MAX_MERGE_MB_PER_SEC, bandwidthRateBucket / activeMerges))
+            : Double.POSITIVE_INFINITY;
+
+    // Apply the calculated rate limit to each active merge thread
+    for (MergeThread mergeThread : mergeThreads) {
+      if (mergeThread.isAlive()) {
+        mergeThread.rateLimiter.setMBPerSec(perMergeRate);
```
stefanvodita: I'm curious how this behaves in performance tests. I'm guessing work isn't generally evenly distributed among threads.

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+                  + getSegmentName(merge)
+                  + " (estimated="
+                  + mergeBandwidthMB
+                  + " MB)");
+        }
+        super.run(); // IO throttling is handled by the RateLimiter
+      } finally {
+        long duration = System.currentTimeMillis() - startTime;
+        if (verbose()) {
+          double mbPerSec = mergeBandwidthMB / Math.max(duration / 1000.0, 0.001);
```
stefanvodita: The way we compute this looks a bit funny. Do we need it?

lucene/core/src/test/org/apache/lucene/index/TestBandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,706 @@
...
+
+    startLatch.countDown();
+    assertTrue("Threads should complete within timeout", doneLatch.await(10, TimeUnit.SECONDS));
+    assertFalse("No thread should have failed", failed.get());
+
+    scheduler.close();
+    dir.close();
+  }
+
+  public void testMergeSchedulerInheritance() throws IOException {
```
stefanvodita: I appreciate that we're being thorough, but I wonder if we risk distracting from more valuable tests with tests like this one. Can we go through this file and reconsider if we want to include all these tests? I think we can cut a lot.

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+  private static final double MIN_MERGE_MB_PER_SEC = 5.0;
+
+  /** Ceiling for IO write rate limit (we will never go any higher than this) */
+  private static final double MAX_MERGE_MB_PER_SEC = 10240.0;
+
+  /** Initial value for IO write rate limit */
+  private static final double START_MB_PER_SEC = 1000.0;
+
+  /** Global bandwidth rate bucket in MB/s */
+  private double bandwidthRateBucket = START_MB_PER_SEC;
```
stefanvodita: Should the name have the unit of measurement?

lucene/core/src/test/org/apache/lucene/index/TestBandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,706 @@
...
+        if (i % 5 == 0) {
+          writer.commit(); // Force segments
+        }
+      }
+
+      writer.forceMerge(1); // This should trigger merges
+    }
+
+    // The scheduler should have been used
+    assertTrue("Scheduler should have been initialized", scheduler.getBandwidthRateBucket() > 0);
```
stefanvodita: Doesn't this just check line 125 without the lines in between having an impact?

stefanvodita (commented):
I wonder if we're addressing the original issue. Maybe I need to spend more time reading though it, but my understanding is that `MergeRateLimiter`, which we use here, is meant to control the amount of resources we put into the merging process, while the idea we were considering was to limit the size of the output of merges. Please correct me if I got that wrong.

lucene/core/src/test/org/apache/lucene/index/TestBandwidthCappedMergeScheduler.java
nipunbatra8: Yes, just wanted to make sure to be through in case but I agree. Will cut down on test cases.

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
nipunbatra8: Good point, I think something closer to START_BANDWIDTH_RATE would be better.

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
nipunbatra8: I have the 0.001 to avoid division by zero problems in case the duration was computed as 0.

nipunbatra8:
@stefanvodita The original issue is that update storms replicate many large segments out to searchers rapidly due to a lot of merging causing paging storms when they finally load a massive snapshot. So rather than preventing large segments by skipping merges, I control the rate at which merges are being run by throttling merge I/O bandwidth, spreading out segment creation over time to prevent the delta of the snapshot "storm" of large segments that overwhelms searchers.

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
mikemccand: +1 to include `MB_PER_SEC` in whichever name we settle on (maybe `bandwidthMbPerSec`?).  Too many times I've seen "off by 1000" ish bugs happen because of units confusion ...

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+ */
+package org.apache.lucene.index;
+
+import java.io.IOException;
+import org.apache.lucene.index.MergePolicy.OneMerge;
+
+/**
+ * A {@link MergeScheduler} that extends {@link ConcurrentMergeScheduler} with bandwidth tracking
+ * and limiting capabilities. This scheduler maintains a bandwidth rate bucket that is divided among
+ * active merges. When bandwidth is limited, merges are throttled to preserve system resources.
```
mikemccand: Maybe state this more succinctly: `A {@link MergeScheduler} that caps total IO write bandwidth across all running merges to a specified max MB/sec bandwidth.`?

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * <p>Key features: Global bandwidth rate bucket with configurable capacity, dynamic per-merge
```
mikemccand: Maybe don't talk about buckets?  I'm not sure what they are in this context.

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+ * throttling.
+ *
+ * @lucene.experimental
+ */
+// nocommit
```
mikemccand: Aha, thank you!  This is because PR is still draft right?

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+public class BandwidthCappedMergeScheduler extends ConcurrentMergeScheduler {
```
mikemccand: Maybe move this to Lucene's sandbox module?  It's where we land more experimental thingies and give them a chance to be used / thrive / iterate and eventually graduate to core.

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+
+  /** Floor for IO write rate limit (we will never go any lower than this) */
+  private static final double MIN_MERGE_MB_PER_SEC = 5.0;
+
+  /** Ceiling for IO write rate limit (we will never go any higher than this) */
+  private static final double MAX_MERGE_MB_PER_SEC = 10240.0;
```
mikemccand: Do we really need the `min` and `max`?

I.e. the min could be 1 [ULP](https://en.wikipedia.org/wiki/Unit_in_the_last_place) above 0.0, and the max is unbounded (`Double.MAX_VALUE`).  In practice very large values of the cap won't matter because merges will become CPU bound at that point and cannot use so much bandwidth.

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
mikemccand: Hmm why do we even have the `START_MB_PER_SEC`?  Could we rather make the ctor take a required `mbPerSec` and use that as the starting value?

Does this class support updating the cap even while merges are running?  It looks like it does, yay!  (Maybe apps have time-dependent bandwidth availability).

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+
+  /** Initial value for IO write rate limit */
+  private static final double START_MB_PER_SEC = 1000.0;
+
+  /** Global bandwidth rate bucket in MB/s */
+  private double bandwidthRateBucket = START_MB_PER_SEC;
+
+  /** Default constructor with 1000 MB/s bandwidth rate bucket */
+  public BandwidthCappedMergeScheduler() {
+    super();
```
mikemccand: I think this is implicitly done in java?  Even if we add the starting `mbPerSec` to the ctor?

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  /** Set the global bandwidth rate bucket in MB/s (default 1000 MB/s) */
```
mikemccand: Could you add to the javadoc that this setting is "live", meaning any running merges running now will be updated to the new cap too?

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+  public void setBandwidthRateBucket(double mbPerSec) {
```
mikemccand: Let's include units in the method name, e.g. `setMaxMbPerSec`?

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+              + " and "
+              + MAX_MERGE_MB_PER_SEC
+              + " MB/s");
+    }
+    this.bandwidthRateBucket = mbPerSec;
+    updateMergeThreads();
+  }
+
+  /** Get the global bandwidth rate bucket in MB/s */
+  public double getBandwidthRateBucket() {
```
mikemccand: `getMaxMbPerSec`?

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+  @Override
+  protected synchronized void updateMergeThreads() {
+    super.updateMergeThreads();
+    int activeMerges = 0;
+    for (MergeThread mergeThread : mergeThreads) {
+      if (mergeThread.isAlive()) {
+        activeMerges++;
+      }
+    }
+    double perMergeRate =
```
mikemccand: Could we expand this into true `if` (don't use the ternary operator)?

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+        activeMerges > 0
+            ? Math.max(
+                MIN_MERGE_MB_PER_SEC,
+                Math.min(MAX_MERGE_MB_PER_SEC, bandwidthRateBucket / activeMerges))
```
mikemccand: `activeMerges` isn't right because it can include merge threads that are created and alive (in the JVM sense) but are paused by CMS parent (in the "soft merge limit" case).  I think it suffices to take `min(maxThreadCount, activeMerges)` instead.  But you'll need to make `maxThreadCount` member protected, or use `getMaxThreadCount()` instead, but also take steps to protect against the `AUTO_DETECT_MERGES_AND_THREADS` in case it didn't get default'd yet.

CMS has a soft limit (`maxThreadCount`) and a hard limit (`maxMergeCount`) on the number of merges requested by the `MergePolicy` that it will tolerate.  Below or equal to the soft limit, everything is normal.  When more merges are needed than the soft limit, the largest N merges are paused so that the remaining soft limit merges can run.  If more merges than the hard limit are requested, CMS will apply backpressure by hijacking the incoming indexing threads (those threads creating new segments and requesting new merges), pausing them, until the merging can catch up and the number of merges requested is again <= the hard limit.

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
mikemccand: This may indeed prove too simplistic (statically dividing bandwidth into the N running merges), but it's good for starters / MLP ("minimum lovable PR") to unlock fun testing.

Actually, instead of statically dividing the cap across N rate limiters ... couldn't we create a single `MergeRateLimiter` shared across all merge threads, set to the cap?  This way, if one merge is doing CPU heavy part of merging and not able to use much IO bandwidth, other merges could surge over their 1/Nth static share?  But this is kinda a big change... CMS seems to quietly assume/expect that each merge thread has its own rate limiter.  It might freak out if you use the same rate limiter across all threads.

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+
+  /** Get the global bandwidth rate bucket in MB/s */
+  public double getBandwidthRateBucket() {
+    return bandwidthRateBucket;
+  }
+
+  /** Distributes the global bandwidth rate bucket evenly among all active merge threads. */
+  @Override
+  protected synchronized void updateMergeThreads() {
+    super.updateMergeThreads();
```
mikemccand: Let's detect if `doAutioIOThrottle` is enabled and throw some exception somewhere?  That CMS feature is incompatible with this class, I think, so it should be clear to the user.  Oh, maybe simply override `enable/disableAutoIOThrottle`?

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+
+    /** Creates a new BandwidthTrackingMergeThread for the given merge. */
+    public BandwidthTrackingMergeThread(MergeSource mergeSource, OneMerge merge) {
+      super(mergeSource, merge);
+      this.mergeBandwidthMB = merge.estimatedMergeBytes / (1024.0 * 1024.0);
+    }
+
+    @Override
+    public void run() {
+      long startTime = System.currentTimeMillis();
```
mikemccand: Let's use `System.nanoTime()` instead?  It is guaranteed monotonic, and has no relation to the odd human labeling of time (which we do not need since we are just computing duration).

Also let's add the units in the variable name (`startTimeNS`).

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+      try {
+        if (verbose()) {
+          message(
+              "Starting bandwidth-capped merge: "
+                  + getSegmentName(merge)
+                  + " (estimated="
```
mikemccand: `(estimatedMbPerSec=` instead?

Also, this is just the value when this merge kicks off?  So maybe change to `startingMbPerSec=`?

Edit: oh this is the estimated size of the final merged segment in MB?  Maybe just change to `merge.estimatedMergeBytes/1024./1024.` and no need for member variable?

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  /** Merge thread that logs the rate limiter value after merge completes. */
+  protected class BandwidthTrackingMergeThread extends MergeThread {
+    private final double mergeBandwidthMB;
+
+    /** Creates a new BandwidthTrackingMergeThread for the given merge. */
+    public BandwidthTrackingMergeThread(MergeSource mergeSource, OneMerge merge) {
+      super(mergeSource, merge);
+      this.mergeBandwidthMB = merge.estimatedMergeBytes / (1024.0 * 1024.0);
```
mikemccand: Oh, this is not actually a bandwidth right?  It's just the estimated final segment size, in MB?

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+          message(
+              "Starting bandwidth-capped merge: "
+                  + getSegmentName(merge)
+                  + " (estimated="
+                  + mergeBandwidthMB
+                  + " MB)");
+        }
+        super.run(); // IO throttling is handled by the RateLimiter
+      } finally {
+        long duration = System.currentTimeMillis() - startTime;
```
mikemccand: `durationNS`?

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
mikemccand: It's OK to just let the math produce `Double.POSITIVE_INFINITY`?  It's only `int`/`long` etc. that cannot divide-by-zero?

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,159 @@
...
+                  + "ms ("
+                  + String.format(java.util.Locale.US, "%.2f", mbPerSec)
+                  + " MB/s)");
+        }
+      }
+    }
+  }
+
+  private static String getSegmentName(OneMerge merge) {
+    return merge.info != null ? merge.info.info.name : "_na_";
```
mikemccand: Hmm is `merge.info` sometimes `null`?  It shouldn't ever be, I think?

mikemccand (commented):
Thank you @nipunbatra8 -- it'll be awesome to have a `MergeScheduler` that can cap aggregate merge bandwidth.

mikemccand:
> I wonder if we're addressing the original issue. Maybe I need to spend more time reading though it, but my understanding is that `MergeRateLimiter`, which we use here, is meant to control the amount of resources we put into the merging process, while the idea we were considering was to limit the size of the output of merges. Please correct me if I got that wrong.

I like this simplicity of this approach, vs the other ideas on https://github.com/apache/lucene/issues/14148 ... but I do agree that having the MP and/or MS aware that they are operating in a resource-scarce environment might do better.  E.g. TMP might delay picking merges if it's already spent its short-term IO budget, and be forced to allow more deletions to accumulate.

But with this simplistic approach, we might emergently see similar behavior.  Throttling a lerge merge that is trying to reclaim deletes effectively stretches out the lifetime of the segments being merged, allowing more deletions to accumulate that would normally, and delaying the appearance of the newly merged segment so that MP/MS cannot pick it for merging until itself is done being merged.  We need to test this approach to see...

@nipunbatra8 `RateLimiter` (as hairy as it is!!) might be overly simplistic for your usage here, especially if we pivot the approach to share a single rate limiter across all merge threads  ... it does not implement the "[leaky bucket](https://en.wikipedia.org/wiki/Leaky_bucket)" model to smooth out bursty IO.  This is another reason to not pivot to this new approach just yet ...

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
nipunbatra8: Yes

lucene/core/src/java/org/apache/lucene/index/BandwidthCappedMergeScheduler.java
nipunbatra8: Yes, made it configurable.

lucene/sandbox/src/java/org/apache/lucene/sandbox/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,185 @@
...
+ * @lucene.experimental
+ */
+public class BandwidthCappedMergeScheduler extends ConcurrentMergeScheduler {
+
+  /** Global bandwidth cap in MB/s. Mutable so that updates are applied live. */
+  private double bandwidthMbPerSec;
+
+  /** Create a scheduler with a required global bandwidth cap in MB/s. */
+  public BandwidthCappedMergeScheduler(double bandwidthMbPerSec) {
+    if (!(bandwidthMbPerSec > 0.0)
```
shubhamsrkdev: Imo, this could just be : 

`bandwidthMbPerSec <= 0.0`

This would be more readable.

lucene/sandbox/src/java/org/apache/lucene/sandbox/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,185 @@
...
+    if (verbose()) {
+      message("Ignoring enableAutoIOThrottle; using bandwidth cap instead");
+    }
+    // Intentionally no-op
+  }
+
+  /** Ensure auto IO throttling remains disabled. */
+  @Override
+  public synchronized void disableAutoIOThrottle() {
+    // Make sure parent state is disabled if it was somehow enabled earlier
```
shubhamsrkdev: Maybe I am missing something here - What would be the scenario when/how this would be enabled?

lucene/sandbox/src/java/org/apache/lucene/sandbox/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,185 @@
...
+  /**
+   * Set the global bandwidth cap in MB/s.
+   *
+   * <p>This setting is live: merges that are already running will be adjusted to the new per-merge
+   * rate derived from this cap, without requiring a restart.
+   *
+   * @param bandwidthMbPerSec the new global bandwidth cap in MB/s; must be finite and > 0
+   */
+  public synchronized void setMaxMbPerSec(double bandwidthMbPerSec) {
+    if (!(bandwidthMbPerSec > 0.0)
```
shubhamsrkdev: I see we are checking the same thing in constructor also, maybe worth creating a separate private method for this?

lucene/sandbox/src/java/org/apache/lucene/sandbox/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,185 @@
...
+      } finally {
+        long durationNS = System.nanoTime() - startTimeNS;
+        if (verbose()) {
+          double durationMS = durationNS / 1_000_000.0;
+          double mbPerSec = merge.estimatedMergeBytes / (1024.0 * 1024.0) / (durationMS / 1000.0);
+          message(
+              "Merge completed: "
+                  + merge.estimatedMergeBytes / (1024.0 * 1024.0)
+                  + " MB in "
+                  + String.format(java.util.Locale.US, "%.1f", durationMS)
```
shubhamsrkdev: Is there a reason this is `Locale.US` and not `Locale.ROOT`? Same for below.

lucene/sandbox/src/java/org/apache/lucene/sandbox/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,185 @@
...
+      super(mergeSource, merge);
+    }
+
+    @Override
+    public void run() {
+      long startTimeNS = System.nanoTime();
+      OneMerge merge = getMerge();
+      try {
+        if (verbose()) {
+          message(
```
shubhamsrkdev, line 150: nit: Shall we use string.format here?

lucene/sandbox/src/test/org/apache/lucene/sandbox/index/TestBandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,382 @@
...
+
+/** Tests for {@link BandwidthCappedMergeScheduler}. */
+public class TestBandwidthCappedMergeScheduler extends LuceneTestCase {
+
+  public void testBasicFunctionality() throws Exception {
+    Directory dir = newDirectory();
+    BandwidthCappedMergeScheduler scheduler = new BandwidthCappedMergeScheduler(100.0);
+
+    // Test initial value
+    assertEquals(100.0, scheduler.getMaxMbPerSec(), 0.001);
```
shubhamsrkdev: Do we need this delta? Is `assertEquals` not working without it?

lucene/sandbox/src/test/org/apache/lucene/sandbox/index/TestBandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,382 @@
...
+
+    scheduler.close();
+    dir.close();
+  }
+
+  public void testBandwidthConfiguration() throws Exception {
+    Directory dir = newDirectory();
```
shubhamsrkdev: I am not sure about the usefulness of checking setters and getters functionalities. Same goes for previous test.

lucene/sandbox/src/test/org/apache/lucene/sandbox/index/TestBandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,382 @@
...
+
+    assertTrue(str.contains("BandwidthCappedMergeScheduler"));
+    assertTrue(str.contains("bandwidthMbPerSec"));
+    assertTrue(str.contains("MB/s"));
+    assertTrue(str.contains("50.0"));
+
+    scheduler.close();
+  }
+
+  public void testWithIndexWriter() throws IOException {
```
shubhamsrkdev: I am not sure about the usability of this test; we are checking in assertion `scheduler.getMaxMbPerSec()` which returns `bandwidthMbPerSec` which was set in the constructor.  The rest of the logic (document creation, write, commit, merges) don't influence the `bandwidthMbPerSec` value in any way. Wouldn't this make this test very similar to the first test (setting and getting `bandwidthMbPerSec`)?

lucene/sandbox/src/test/org/apache/lucene/sandbox/index/TestBandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    // The scheduler should have been used
+    assertTrue("Scheduler should have been initialized", scheduler.getMaxMbPerSec() > 0);
+
+    scheduler.close();
+    dir.close();
+  }
+
+  public void testBandwidthDistributionAmongMerges() throws IOException {
```
shubhamsrkdev: This test has a missing assertion, what are we asserting here?

lucene/sandbox/src/java/org/apache/lucene/sandbox/index/BandwidthCappedMergeScheduler.java
nipunbatra8: Good point

lucene/sandbox/src/java/org/apache/lucene/sandbox/index/BandwidthCappedMergeScheduler.java
nipunbatra8: The parent class has a IOThrottle that can be enabled for more efficiency. Realized it is defaulted as false in the parent class so this method and the one below are not needed because it will always be false since we disabled the enableAutoIOThrottle function.

lucene/sandbox/src/java/org/apache/lucene/sandbox/index/BandwidthCappedMergeScheduler.java
nipunbatra8: Good idea, makes it more readable too

lucene/sandbox/src/java/org/apache/lucene/sandbox/index/BandwidthCappedMergeScheduler.java
nipunbatra8: No, will make it more universal with Locale.ROOT

lucene/sandbox/src/java/org/apache/lucene/sandbox/index/BandwidthCappedMergeScheduler.java
```diff
@@ -0,0 +1,174 @@
...
+      }
+    }
+
+    // Use the effective max thread count to avoid counting threads that CMS has paused
+    int effectiveMaxThreads = getMaxThreadCount();
+    if (effectiveMaxThreads == ConcurrentMergeScheduler.AUTO_DETECT_MERGES_AND_THREADS) {
+      int coreCount = Runtime.getRuntime().availableProcessors();
+      effectiveMaxThreads = Math.max(1, coreCount / 2);
+    }
+    int divisor = Math.min(effectiveMaxThreads, activeMerges);
```
ytgu, line 98: I saw your discussion with @mikemccand about using `min(maxThreadCount, activeMerges)` here to make sure the paused (even though they're alive) threads are not being counted. Just curious - can't we just use 
```
    for (MergeThread mergeThread : mergeThreads) {
      if (mergeThread.isAlive() && mergeThread.getRateLimiter().getMBPerSec() > 0.0) {
        activeMerges++;
      }
    }
```
to count the active merging threads?

PR #14965, opened 2025-07-17
ecj: enable redundantSpecificationOfTypeArguments and fix issues

rmuir:
One of the easier ECJ checks to enable and fix: although a bit cosmetic it is also the least risky. Makes the code look a little less java 5 and a little more java 24, especially if we address other ECJ issues.


jpountz (approved):

uschindler (approved):
I did not check every line changed, but general setup looks OK.

For me it is only important that we are able to fix issues with newer compiler by being explicit (if type injection fails for some reason). I imagine a simple `@SuppressWarnings` helps.

PR #14966, opened 2025-07-18
ecj: enable possibleAccidentalBooleanAssignment and fix issues

rmuir:
This check trips if you use an assignment as a boolean conditional.

You can avoid it either by not using assignments as boolean conditionals, or by using an explicit comparison such as:

```java
// instead of:
if (x = y) {}

// change to:
if ((x = y) == true) {}
```


jpountz (approved):
The code is clearer this way!

dweiss (approved):

PR #14974, opened 2025-07-21
Tessellator should fail if it cannot process a hole

iverase:
There are some situations when tIllegal polygons are incorrectly processed by the tessellator. This happens when a polygon is provided with a hole that is outside of the polygon. In some cases we cannot find a bridge and we actually just ignored the hole. I think this is not the right behaviour because the polygon is illegal and we should fail instead,


lucene/CHANGES.txt
```diff
@@ -243,6 +243,9 @@ Bug Fixes
 * GITHUB#14922: Fix for IndexedDISI.Dense#intoBitSetWithinBlock to take into account the provided bitset size
   and avoid throwing IndexOutOfBoundsException (Panagiotis Bailis)
 
+ * GITHUB#14974: Lucene tessellator always fails if it cannot process a hole instrad of just ignore it.
```
craigtaverner:
```suggestion
 * GITHUB#14974: Lucene tessellator now always fails if it cannot process a hole instead of just ignoring it.
```

lucene/core/src/java/org/apache/lucene/geo/Tessellator.java
```diff
@@ -339,16 +340,30 @@ private static Node eliminateHoles(
...
         holeMaxY = holePoly.maxY;
       }
-      eliminateHole(holeNode, outerNode, holeMinX, holeMaxX, holeMinY, holeMaxY);
-      // Filter the new polygon.
-      outerNode = filterPoints(outerNode, outerNode.next);
+      if (eliminateHole(holeNode, outerNode, holeMinX, holeMaxX, holeMinY, holeMaxY)) {
+        // Filter the new polygon.
+        outerNode = filterPoints(outerNode, outerNode.next);
+      } else {
+        // cannot process the hole, it is not inside the polygond.
```
craigtaverner:
```suggestion
        // cannot process the hole, it is not inside the polygon.
```

lucene/core/src/java/org/apache/lucene/geo/Tessellator.java
```diff
@@ -374,6 +389,9 @@ private static void eliminateHole(
       Node node = splitPolygon(outerNode, holeNode, fromPolygon);
       // Filter the split nodes.
       filterPoints(node, node.next);
+      return true;
+    } else {
+      return false;
```
craigtaverner, line 395: Are we sure that the inability to find a bridge is always due to an invalid hole?

lucene/core/src/java/org/apache/lucene/geo/XYPolygon.java
```diff
@@ -184,6 +184,19 @@ public boolean equals(Object obj) {
     return true;
   }
 
+  public static String verticesToGeoJSON(final float[] xs, final float[] ys) {
+    StringBuilder sb = new StringBuilder();
+    sb.append('[');
+    for (int i = 0; i < xs.length; i++) {
+      sb.append("[").append(xs[i]).append(", ").append(ys[i]).append("]");
+      if (i != xs.length - 1) {
```
craigtaverner, line 192: I normally use `if(i>0)` and do this at the top of the block. I think it reads slightly better.

craigtaverner (approved):
Looks like a good fix.

lucene/core/src/java/org/apache/lucene/geo/Tessellator.java
iverase: yes, this means we could not find a polygon point to the left of the leftmost point of the hole, therefore that point must be outside of the polygon. I will add a comment.

PR #14976, opened 2025-07-21
Utilize docIDRunEnd on ReqExclBulkScorer

HUSTERGS:
### Description
Since #14806 is reverted because it causes some test broken (sorry about that). This PR tries to fix it and re-raise the same logic. This [comment](https://github.com/apache/lucene/pull/14806#issuecomment-3095570822) explains why it causes problem. I think this might be helpful when some must not clauses implements its own `docIDRunEnd`, although currently there is no such task in luceneutil.

I've test the specific failed test locally with command below:
```
./gradlew -p lucene/core beast -Ptests.dups=500 --tests TestLRUQueryCache.testRandom
```


dweiss:
Thank you. I've applied this to main and branch_10x. Please keep an eye on:
https://lists.apache.org/list.html?builds@lucene.apache.org

Tests are randomized so sometimes what passes on github's PR workflows breaks the build in a different configuration (nothing wrong with that - it's how it's designed to work).

PR #14981, opened 2025-07-22
Convert more gradle convention plugins to java

dweiss:
This is work in progress until it gains some weight again.


dweiss:
I plan to merge this shortly. It's large enough.

PR #14988, opened 2025-07-23
Move isolated-inheritdoc detection to an ast-grep rule

dweiss:
This means the set of patterns in source-patterns validation is empty... easier to port to Java or simplify (?).

### DIFF
--- a/build-tools/build-infra/src/main/groovy/lucene.validation.source-patterns.gradle
+++ b/build-tools/build-infra/src/main/groovy/lucene.validation.source-patterns.gradle
@@ -97,9 +97,7 @@ class ValidateSourcePatternsTask extends DefaultTask {
 
   @TaskAction
   public void check() {
-    def invalidPatterns = [
-      (~$/\Q/**\E((?:\s)|(?:\*))*\Q{@inheritDoc}\E((?:\s)|(?:\*))*\Q*/\E/$) : '{@inheritDoc} on its own is unnecessary',
-    ]
+    def invalidPatterns = []
 
     def violations = new TreeSet();
     def reportViolation = { f, name ->
--- a/gradle/validation/ast-grep/rules/java-patterns.yml
+++ b/gradle/validation/ast-grep/rules/java-patterns.yml
@@ -140,3 +140,12 @@ rule:
         pattern: incubator
 severity: error
 message: Module dependency on incubator APIs
+---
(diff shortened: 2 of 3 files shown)


dweiss:
It's my feeble attempt at an ast-grep rule, I'd appreciate you taking a look, @rmuir , whenever you have time.

rmuir (approved):
Nice solution.

rmuir:
There's no better way without adding a ton of complex. I'm happy you found a solution that doesn't require `npm install tree-sitter-javadoc`. If you're curious, this how that looks: but adding custom injected language makes it too hard on developers and also makes things quite a bit slower.

```yaml
# sgconfig.yml
customLanguages:
  javadoc:
    # works cross platform actually, but requires additional `npm install`
    libraryPath: /path/to/javadoc.so
    extensions: [ javadoc ]
    expandoChar: _
languageInjections:
  - hostLanguage: java
    rule:
      kind: block_comment
      pattern: $CONTENT
      regex: "^/[*][*][^*]"
    injected: javadoc

# now javadoc rules can be written naturally like so:
# javadoc.yml
id: no-javadoc-author-tags
language: javadoc
rule:
  kind: author_tag
message: javadoc `@author` tags should not be used
fix: ""
severity: error
```

build-tools/build-infra/src/main/groovy/lucene.validation.source-patterns.gradle
```diff
@@ -97,9 +97,7 @@ class ValidateSourcePatternsTask extends DefaultTask {
 
   @TaskAction
   public void check() {
-    def invalidPatterns = [
-      (~$/\Q/**\E((?:\s)|(?:\*))*\Q{@inheritDoc}\E((?:\s)|(?:\*))*\Q*/\E/$) : '{@inheritDoc} on its own is unnecessary',
-    ]
+    def invalidPatterns = []
```
rmuir, line 100: Actually i'd remove `java` extension from line 34:
> List<String> extensions = ['java', 'xml',]

This should mean this task now only scans xml files. Thousands of less files. The UTF-8 validation is already happening by `eclint` because editorconfig says utf-8, so it validates that.

dweiss: Ah... if so then indeed the check becomes much, much simpler. Thank you. I'll strip the unused parts when converting it to java (which I'm doing in another pr).

PR #14989, opened 2025-07-24
Add accessors for MergeState and SegmentCommitInfo's public fields

chishui:
### Description
Found some classes don't have good encapsulation and doesn't offer accessors for its fields. It makes mocking MergeState in unit tests difficult.

We can't modify the access level for these fields, so I add getter functions for all these public fields.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/MergeState.java
+++ b/lucene/core/src/java/org/apache/lucene/index/MergeState.java
@@ -91,7 +91,86 @@ public class MergeState {
   /** Indicates if the index needs to be sorted * */
   public boolean needsIndexSort;
 
-  /** Sole constructor. */
+  /** Returns the document ID maps. */
+  public DocMap[] getDocMaps() {
+    return docMaps;
+  }
+
+  /** Returns the segment info of the newly merged segment. */
+  public SegmentInfo getSegmentInfo() {
+    return segmentInfo;
+  }
+
+  /** Returns the field infos of the newly merged segment. */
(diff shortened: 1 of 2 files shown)


lucene/core/src/java/org/apache/lucene/index/MergeState.java
```diff
@@ -91,7 +91,86 @@ public class MergeState {
...
+
+  /** Returns the executor for intra merge activity. */
+  public Executor getIntraMergeTaskExecutor() {
+    return intraMergeTaskExecutor;
+  }
+
+  /** Returns whether the index needs to be sorted. */
+  public boolean getNeedsIndexSort() {
+    return needsIndexSort;
+  }
```
rmuir, line 172: Adding a getter to a public field is senseless java verbosity: it adds no value. The field is public: just access the field directly.

rmuir (changes requested):
Sorry, I can't agree with adding public getters for public fields.

lucene/core/src/java/org/apache/lucene/index/MergeState.java
chishui: Thanks for reviewing. I fully understand that you can just use public fields but that's not the point. Like what I mentioned in the description, it provides better testability, we can easily mock a member function but we can't mock a field access. 

Apart from that, it's not good practice to have public fields as it doesn't provide good encapsulation. It's just the fields are already public and for backward compatibility, we can't change them to private.

PR #14996, opened 2025-07-28
Make BitSet no longer implement Bits.

jpountz:
Lucene has call sites of `Bits#get` that are more polymorphic than they should because they sometimes receive a `FixedBitSet` as `Bits`, and sometimes a `FixedBits` (the return type of `FixedBitSet#asReadOnlyBits`). This makes some call sites bimorphic when they could be monomorphic, of polymorphic when they could be bimorphic (e.g. if `SparseFixedBitSet` also gets used).

Given how polymorphism affects performance, I would like to fix this performance trap. One idea would be to make `BitSet` no longer implement `Bits`, ie. you could no longer pass a `FixedBitSet` when `Bits` are expected, you would need to pass the result of `FixedBitSet#asReadOnlyBits`. This is what this PR does.

I had to move/copy some APIs around between `Bits`, `BitSet` and `FixedBitSet` to make it work, but it's not as invasive as I initially worried.

The API change is a 11.0 change, but we could backport the bits that reduce polymorphism to 10.x, ie. making sure to pass the result of `FixedBitSet#asReadOnlyBits` rather than the `FixedBitSet` directly when `Bits` are expected.


jpountz:
@msokolov @benwtrent You may want to take a look since vector search is probably the most performance-critical code that is touched by this change due to `AbstractKnnVectorQuery`, which passes the following implementing classes for the `Bits` acceptDocs:
  - `FixedBits` for unfiltered searches,
  - `FixedBitSet` for dense filters,
  - `SparseFixedBitSet` for sparse filters.

With this PR, dense filters and unfiltered searches both use a `FixedBits` instance, so call sites of `Bits#get` within the vector search code become bimorphic. I don't expect nightly benchmarks to see a difference though since we don't run queries that are filtered by sparse filters.

lucene/core/src/java/org/apache/lucene/index/SoftDeletesDirectoryReaderWrapper.java
```diff
@@ -188,11 +189,11 @@ private static boolean assertDocCounts(
 
   static final class SoftDeletesFilterLeafReader extends FilterLeafReader {
     private final LeafReader reader;
-    private final FixedBitSet bits;
+    private final Bits bits;
     private final int numDocs;
     private final CacheHelper readerCacheHelper;
 
-    private SoftDeletesFilterLeafReader(LeafReader reader, FixedBitSet bits, int numDocs) {
+    private SoftDeletesFilterLeafReader(LeafReader reader, Bits bits, int numDocs) {
```
msokolov, line 196: curious why we are changing these signatures - it seems the call sites still always use FixedBitSet?

lucene/core/src/java/org/apache/lucene/util/FixedBits.java
```diff
@@ -31,8 +33,17 @@ public boolean get(int index) {
...
+  public void applyMask(FixedBitSet bitSet, int offset) {
+    // Note: Some scorers don't track maxDoc and may thus call this method with an offset that is
+    // beyond bitSet.length()
+    int length = Math.min(bitSet.length(), length() - offset);
+    if (length >= 0) {
+      FixedBitSet.andRange(this.bitSet, offset, bitSet, 0, length);
+    }
+    if (length < bitSet.length()
+        && bitSet.nextSetBit(Math.max(0, length)) != DocIdSetIterator.NO_MORE_DOCS) {
+      throw new IllegalArgumentException("Some bits are set beyond the end of live docs");
```
msokolov, line 45: this comment refers to "live docs" but the method is generic and could apply to other kinds of bit set applications - maybe change to "beyond the end of the bit set"?

msokolov (commented):
Sounds good! I left a few small comments/questions

lucene/core/src/java/org/apache/lucene/index/SoftDeletesDirectoryReaderWrapper.java
jpountz: The `Bits` are indeed always created bia `FixedBitSet#asReadOnlyBits` but I liked doing it on the call site better than in this constructor. Since it's a private constructor, it doesn't break users.

lucene/core/src/java/org/apache/lucene/util/FixedBits.java
jpountz: Right, I'll fix.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -214,7 +214,9 @@ private TopDocs getLeafResults(
 
     // Perform the approximate kNN search
     // We pass cost + 1 here to account for the edge case when we explore exactly cost vectors
-    TopDocs results = approximateSearch(ctx, acceptDocs, cost + 1, timeLimitingKnnCollectorManager);
+    TopDocs results =
+        approximateSearch(
+            ctx, acceptDocs.asReadOnlyBits(), cost + 1, timeLimitingKnnCollectorManager);
```
benwtrent: Within the HNSW format, we do this:
```
 if (acceptDocs instanceof BitSet bitSet) {
      // Use approximate cardinality as this is good enough, but ensure we don't exceed the graph
      // size as that is illogical
      filteredDocCount = Math.min(bitSet.approximateCardinality(), graph.size());
      if (unfilteredVisit >= filteredDocCount) {
        doHnsw = false;
      }
    }
```
We determine the filter size via casting the `Bits` to `BitSet`. So, I am not sure we can make this change without progressing: https://github.com/apache/lucene/pull/15011

Or `bits` should give cardinality information...

PR #14998, opened 2025-07-28
Speed up flushing of live docs.

jpountz:
This uses `Bits#applyMask` (introduced in #14134) to speed up flushing live docs, instead of checking bits one by one.


jpountz:
I'm merging since I don't think it's controversial. Feel free to ask me to revert if you have concerns.

PR #15001, opened 2025-07-29
Remove full integrity check from SortingStoredFieldsConsumer

### LINKED ISSUE #14881: Should SortingStoredFieldsConsumer do a full integrity check?
### Description

Today when index sorting is enabled and stored fields get flushed then the `SortingStoredFieldsConsumer` gets used in order to store stored fields in the order in which index sorting is configured. This class writes temp files to disk that then get read completely twice. The first time is to do an integrity check and the second time the temp files are read in random order. This to write stored fields in the right order (defined by index sorting) in the new segment. 

During heavy indexing the fact that the stored field temp files are read twice is expensive. Especially given that these temp files will be removed after flushing has completed. In other formats (postings, bkd tree, quantized vectors), tmp files that get created during writing seem to be read only once. During reading either integrity is check using `Directory#openChecksumInput()` (only possible if temp file is read from beginning to end) or there is a footer check (reads and validates CRC, footer magic and algorithm id).

I wonder whether it makes sense to remove the full separate [integrity check](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/SortingStoredFieldsConsumer.java#L106) in `SortingStoredFieldsConsumer`? This can be quite costly, especially the integrity check for the temp fdt file and also there is already some light integrity checking via `CodecUtil.checkFooter(...)` in `Lucene90CompressingStoredFieldsReader`.

martijnvg:
In write-heavy scenarios with significant stored field usage, the full integrity check that happens during flushing stored fields to disk when index sorting is active can be very expensive. 
Instead, rely on the light integrity check that happens in `Lucene90CompressingStoredFieldsReader`.

Note that at some point when newly flushed segments are getting merged, a full integrity check will happen.

Closes #14881

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -220,7 +220,7 @@ Optimizations
 
 * GITHUB#14991: Refactor for loop at PointRangeQuery hot path. (Ge Song)
 
-* GITHUB#: Remove full integrity check from SortingStoredFieldsConsumer (Martijn van Groningen)
+* GITHUB#15001: Remove full integrity check from SortingStoredFieldsConsumer (Martijn van Groningen)
 
 Changes in Runtime Behavior
 ---------------------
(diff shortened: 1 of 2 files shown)


ChrisHegarty (approved):
LGTM

PR #15002, opened 2025-07-29
remove IndexReader.registerParentReader()

### LINKED ISSUE #14999: Performance bottleneck in IndexReader.registerParentReader with many parentReaders
### Description

During a recent issue in OpenSearch, we discovered high wait time(>50ms) for `IndexReader.registerParentReader` for both query and fetch phase. During this time,  OpenSearchLeafReader was having more than >30k parentReader, and with many leafReaders(aka segments) per lucene index, this caused significant latency for a search request(especially tail latencies).

Each IndexReader appears to tracks its parents using a synchronized set that holds weak references. These weak references are cleared during subsequent synchronizedSet.add calls if the GC has already reclaimed the associated objects. During the time of this issue, the set had grown beyond  > 30k entries, and one thread was already holding a lock to clean up stale references. As a result, other threads attempting to call  `SynchronizetSet.add` were blocked, leading to contention and delay.

A sample lock profile looked like below. ~33% of the time being spent in trying to add elements to the set.

Image

**More context**:
In OpenSearch, each query and fetch phase maintains a per-request context by wrapping the underlying reader in a NonClosingReaderWrapper. This ensures that plugins can safely wrap the reader without risking accidental closure of the underlying resource. For every request on a shard, multiple short-lived NonClosingReaderWrapper instances (along with other plugin-provided IndexReaderWrappers) are created and registered as parents of the OpenSearchLeafReader.

This leads to lock contention in `synchronizedSet.add`, especially when stale entries in the parentReaders set are being cleared. Under such conditions, lock wait times can exceed 100ms, depending on the size of the parentReaders set. The issue becomes more pronounced under high throughput per shard, where the frequency of reader registration increases significantly.

I understand that one could argue this is primarily an OpenSearch issue, specifically how it creates and manages per-request reader wrappers. However, I believe Lucene should offer a basic mechanism at the IndexReader level to exclude certain instances from being tracked as parents, or alternatively, improve its current approach. The existing use of a synchronizedSet with weak references can lead to performance bottlenecks in such cases. Curious to know what lucene folks think about this.

**Some solutions I can think of are below** 
Open to any other ideas.

1.  Have a way to not track certain index readers as parents. Any short lived readerWrapper can then have trackAsParent as false and avoid being tracked. I did a simple implementation of this, and it improved p99 upto ~30-40% in some cases.
```
class IndexReader {
  public boolean trackAsParent() {
     return true;
  }

   public final void registerParentReader(IndexReader reader) {
      ensureOpen();
       if (reader.trackAsParent()) {
           parentReaders.add(reader);
       }
     }
}
```

2. If not above, consider making `registerParentReader` non-final so its behavior can be overriden?

3. Or maybe provide an explicit way to remove parentReader from the set. So that once short loved readerWrapper are closed, we can explicitly remove them immediately and not allow synchronizedSet to grow in size.  This still might have some contention, but will have to test it out.

rmuir:
Remove this function and associated WeakhashMap: it is not needed any more. The tests still pass, user still gets ACE (just slightly different exception message).

Closes #14999

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/index/TestReaderClosed.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestReaderClosed.java
@@ -81,37 +81,6 @@ public class TestReaderClosed extends LuceneTestCase {
     }
   }
 
-  // LUCENE-3800
-  public void testReaderChaining() throws Exception {
-    assertTrue(reader.getRefCount() > 0);
-    LeafReader wrappedReader = new ParallelLeafReader(getOnlyLeafReader(reader));
-
-    // We wrap with a OwnCacheKeyMultiReader so that closing the underlying reader
-    // does not terminate the threadpool (if that index searcher uses one)
-    IndexSearcher searcher = newSearcher(new OwnCacheKeyMultiReader(wrappedReader));
-
-    TermRangeQuery query = TermRangeQuery.newStringRange("field", "a", "z", true, true);
-    searcher.search(query, 5);
-    reader.close(); // close original child reader
-    try {
-      searcher.search(query, 5);
-    } catch (Exception e) {
-      AlreadyClosedException ace = null;
-      for (Throwable t = e; t != null; t = t.getCause()) {
-        if (t instanceof AlreadyClosedException) {
-          ace = (AlreadyClosedException) t;
-        }
-      }
-      if (ace == null) {
-        throw new AssertionError("Query failed, but not due to an AlreadyClosedException", e);
-      }
-      assertEquals("this IndexReader is closed", ace.getMessage());
-    } finally {
-      // close executor: in case of wrap-wrap-wrapping
-      searcher.getIndexReader().close();
-    }
-  }
-
   @Override
   public void tearDown() throws Exception {
     dir.close();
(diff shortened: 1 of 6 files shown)


lucene/core/src/test/org/apache/lucene/index/TestReaderClosed.java
```diff
@@ -105,9 +105,7 @@ public void testReaderChaining() throws Exception {
       if (ace == null) {
         throw new AssertionError("Query failed, but not due to an AlreadyClosedException", e);
       }
-      assertEquals(
-          "this IndexReader cannot be used anymore as one of its child readers was closed",
-          ace.getMessage());
+      assertEquals("this IndexReader is closed", ace.getMessage());
```
uschindler: I am not sure if this exact message is thrown by all directory implementations. Maybe just check for the type of exection and ignore message. So just remove this line.

See the test name "testReaderChaining", maybe remove it completely?

uschindler (approved):
Looks OK.

IMHO we should not backport this and let it live for some time on main branch. I think the whole thing was only invented for MMapDirectory when closing child readers may sigsegv parent reader.

If we find this useful in main, we can backport it (in 10.x we have no MappedByteBuffer usage anymore). If we do this, maybe deprecate and make the public API | https://github.com/apache/lucene/pull/15002/files#diff-2f71663645c05d1d7a61c17ed8928715a4ef6c01b47fd19d488a51ade76bce1aL162 a noop.

uschindler:
I marked this for 11.0. Please add a changes.txt.

PR #15003, opened 2025-07-29
Avoid reconstructing HNSW graphs during segment merging.

Pulkitg64:
### Description

This is a draft PR to optimize HNSW graph merging during singleton merges. When merging a single segment with deletions, the current implementation reconstructs the entire graph with only live nodes, which is a time-consuming process. This PR avoids full graph reconstruction by dropping deleted nodes and renumbering the remaining live nodes.

TODOs:

Add specific unit tests
Benchmarks (luceneutil)


jpountz:
I don't feel qualified to do the review, but I agree with the motivation. I wonder if this optimization could be applied when there are more than 1 segment to merge by first applying deletions on the bigger segment to merge and then adding vectors from other segments?

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -69,6 +70,8 @@ public final class Lucene99HnswVectorsWriter extends KnnVectorsWriter {
 
   private static final long SHALLOW_RAM_BYTES_USED =
       RamUsageEstimator.shallowSizeOfInstance(Lucene99HnswVectorsWriter.class);
+  static final int DELETE_THRESHOLD_PERCENT = 30;
```
msokolov: I'm curious if we have done any testing to motivate this choice?  I guess as the number of gaps in the neighborhoods left behind by removing the deleted nodes in the graph increases we would expect to see a drop-off in recall, or maybe performance? but I don't have a good intuition about whether there is a knee in the curve, or how strong the effect is

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -347,12 +350,195 @@ private void reconstructAndWriteNeighbours(
...
+   *
+   * @param graph The off-heap graph to process
+   * @param docMap Mapping from old document IDs to new document IDs
+   * @param graphSize The size of the graph (number of vectors)
+   * @param offsets Array to store the byte offsets for each node at each level
+   * @return A mock HnswGraph implementation that provides access to the graph structure
+   * @throws IOException If an error occurs while writing to the vector index
+   */
+  private HnswGraph deleteNodesWriteGraph(
+      Lucene99HnswVectorsReader.OffHeapHnswGraph graph,
```
msokolov: Could we change the signature to accept an `HnswGraph`?

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -347,12 +350,195 @@ private void reconstructAndWriteNeighbours(
...
+    final int[][] validNodesPerLevel = new int[numLevels][];
+
+    // Process all levels
+    for (int level = 0; level < numLevels; level++) {
+      int[] sortedNodes = NodesIterator.getSortedNodes(graph.getNodesOnLevel(level));
+
+      // Count and collect valid nodes
+      int validNodeCount = 0;
+      for (int node : sortedNodes) {
+        if (docMap.get(node) != -1) {
```
msokolov: we might be able to pass in the size of the new graph? At least in the main case of merging we should know (I think?)

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -347,12 +350,195 @@ private void reconstructAndWriteNeighbours(
...
+          validNodeCount++;
+        }
+      }
+
+      // Special case for top level with no valid nodes
+      if (level == numLevels - 1 && validNodeCount == 0 && level > 0) {
```
msokolov: if level ==0 and validNodeCount == 0 the new graph should be empty. I'm not sure how that case will get handled here?

msokolov: In this case though (the top level would be empty) -- isn't it also possible that a lower level is empty?

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -347,12 +350,195 @@ private void reconstructAndWriteNeighbours(
...
+        validNodeCount = 1; // We'll create one connection to lower level
+      }
+
+      validNodesPerLevel[level] = new int[validNodeCount];
```
msokolov: I wonder if we could avoid the up-front counting, allocate a full-sized array and then use only the part of it that we fill up

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -347,12 +350,195 @@ private void reconstructAndWriteNeighbours(
...
+        long offsetStart = vectorIndex.getFilePointer();
+
+        // Process and write neighbors with delta encoding
+        writeNeighbors(graph, docMap, scratch, level == 0 ? graphSize : 0);
+
+        offsets[level][nodeOffsetId++] =
+            Math.toIntExact(vectorIndex.getFilePointer() - offsetStart);
+      }
+
+      // Special case for empty top level
```
msokolov: Perhaps we should special case the first empty level we findand make that the top level, unless it is the bottom level in which case the whole graph is empty

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -347,12 +350,195 @@ private void reconstructAndWriteNeighbours(
...
+      @Override
+      public NodesIterator getNodesOnLevel(int level) {
+        return new ArrayNodesIterator(validNodesPerLevel[level], validNodesPerLevel[level].length);
+      }
+    };
+  }
+
+  /** Writes neighbors with delta encoding to the vector index. */
+  private void writeNeighbors(
+      Lucene99HnswVectorsReader.OffHeapHnswGraph graph,
```
msokolov: can we delegate to an existing method (maybe with a refactor) to ensure we write in the same format?  EG what if we switch to GroupVarInt encoding - we want to make sure this method tracks that change

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -347,12 +350,195 @@ private void reconstructAndWriteNeighbours(
...
+  }
+
   @Override
   public void mergeOneField(FieldInfo fieldInfo, MergeState mergeState) throws IOException {
     CloseableRandomVectorScorerSupplier scorerSupplier =
         flatVectorWriter.mergeOneFieldToIndex(fieldInfo, mergeState);
     try {
       long vectorIndexOffset = vectorIndex.getFilePointer();
+
+      if (mergeState.liveDocs.length == 1
```
msokolov: have you seen IncrementalHnswGraphMerge and MergingHnswGraphBuilder? They select the biggest graph with no deletions and merge the other segments' graphs into it.  Could we expose a utility method here for rewriting a graph (in memory) to drop deletions, and then use it there?

Here we are somewhat mixing the on-disk graph format with the logic of dropping deleted nodes, which I think we could abstract out intoi the util.hnsw realm?

msokolov (commented):
this seems like a promising direction! I left a bunch of comments. My main one is about whether we should do this on-heap to make it more flexible (eg so we could use it when merging multiple graphs, too).

Pulkitg64:
>  I wonder if this optimization could be applied when there are more than 1 segment to merge by first applying deletions on the bigger segment to merge and then adding vectors from other segments?

@jpountz Yes good idea, let me try doing that only in this PR

> this seems like a promising direction! I left a bunch of comments. My main one is about whether we should do this on-heap to make it more flexible (eg so we could use it when merging multiple graphs, too).

Thanks @msokolov . Yes I think that would be best way forward for this optimization. Working on it.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
Pulkitg64: Just saw that class. I think this is a good idea. Will do it in next revision.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
Pulkitg64: Sure.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
Pulkitg64: Sure

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
```diff
@@ -347,12 +350,195 @@ private void reconstructAndWriteNeighbours(
...
+
+      validNodesPerLevel[level] = new int[validNodeCount];
+      offsets[level] = new int[validNodeCount];
+
+      int validNodeIndex = 0;
+      int nodeOffsetId = 0;
+
+      // Process nodes at this level
+      for (int node : sortedNodes) {
+        if (docMap.get(node) == -1) {
```
Pulkitg64: This is incorrect. Graph does not store docIDs but instead they store ordinal. Whereas docMap maps oldDocIds to new DocIDs. 
The correct implementation is to create a map which maps old ords to new ords.

Will fix this in next revision.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsWriter.java
Pulkitg64: > if level ==0 and validNodeCount == 0 the new graph should be empty. I'm not sure how that case will get handled here?

This means 100% nodes are deleted, right? I think we will never reach this case as entry condition to this function is checking if deletes are less than 30%.

Pulkitg64:
The failing test is running fine on my macOS desktop and I have not changed anything in the related classes. Even with the same failing seed  I am unable to reproduce the issue. Not sure why this test in failing in check.

```
TestBPReorderingMergePolicy > testReorderOnAddIndexes FAILED
    java.lang.AssertionError: Called on the wrong instance
        at org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.codecs.asserting.AssertingKnnVectorsFormat$AssertingKnnVectorsReader.getFloatVectorValues(AssertingKnnVectorsFormat.java:140)
        at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.codecs.perfield.PerFieldKnnVectorsFormat$FieldsReader.getFloatVectorValues(PerFieldKnnVectorsFormat.java:289)
        at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CodecReader.getFloatVectorValues(CodecReader.java:244)
        at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.SlowCompositeCodecReaderWrapper$SlowCompositeKnnVectorsReaderWrapper.getFloatVectorValues(SlowCompositeCodecReaderWrapper.java:842)
        at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.CodecReader.getFloatVectorValues(CodecReader.java:244)
        at org.apache.lucene.misc.index.BpVectorReorderer.computeDocMap(BpVectorReorderer.java:590)
        at org.apache.lucene.misc.index.BPReorderingMergePolicy$1.reorder(BPReorderingMergePolicy.java:138)
        at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.IndexWriter.addIndexesReaderMerge(IndexWriter.java:3426)
        at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.IndexWriter$AddIndexesMergeSource.merge(IndexWriter.java:3334)
        at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.ConcurrentMergeScheduler.doMerge(ConcurrentMergeScheduler.java:664)
        at org.apache.lucene.core@11.0.0-SNAPSHOT/org.apache.lucene.index.ConcurrentMergeScheduler$MergeThread.run(ConcurrentMergeScheduler.java:726)
```

Pulkitg64:
I created a pull request to my own repository, there tests are working fine: https://github.com/Pulkitg64/lucene/pull/1

This issue looks to be transient and next commit should fix this.

Pulkitg64:
Adding some KnnPerfTestResults where I tried to simulate deletes while indexing docs. We are seeing consistent improvement in Indexing Time and Indexing Rate (except one weird case when we deleted 40% docs) without impacting recall.

Num Docs: 1MM
Max-Conn: 32
Beam-Width: 250
Quantize Bits: 32
Topk: 100

| Experiment | Baseline |               |               | Candidate |               |               | % Change      |               |
|------------|----------|---------------|---------------|-----------|---------------|---------------|---------------|---------------|
| % Deletes  | Recall   | Indexing Time | Indexing Rate | Recall    | Indexing Time | Indexing Rate | Indexing Time | Indexing Rate |
| 25         | 0.952    | 692           | 1443          | 0.955     | 576           | 1734          | -17%          | 20%           |
| 30         | 0.952    | 581           | 1719          | 0.958     | 517           | 1932          | -11%          | 12%           |
| 40         | 0.951    | 560           | 1782          | 0.945     | 553           | 1805          | -1%           | 1%            |
| 50         | 0.96     | 446           | 2241          | 0.953     | 421           | 2371          | -6%           | 6%            |
| 60         | 0.974    | 234           | 4265          | 0.972     | 208           | 4804          | -11%          | 13%           |

msokolov:
I am confused! This PR suddenly got so much simpler, which is great, but I feel like it dropped a few things that seemed important. EG we are no longer checking the largest graph to see if its delete % is below a threshold?  Also I think we are now ignoring the various edge cases around upper-level graph layers possibly becoming empty?

Pulkitg64:
With MaxConn = 16, I am seeing much better results. But on a weird case with 25% delete I am seeing regression in indexing rate. Trying maxConn=8 in next benchmark run

| Experiment | Baseline |                   |                        | Candidate |                   |                        | % Change      |               |
|------------|----------|-------------------|------------------------|-----------|-------------------|------------------------|---------------|---------------|
| % Deletes  | Recall   | Indexing Time (s) | Indexing Rate (docs/s) | Recall    | Indexing Time (s) | Indexing Rate (docs/s) | Indexing Time | Indexing Rate |
| 25         | 0.922    | 453               | 2205                   | 0.914     | 484               | 2063                   | 7%            | -6%           |
| 30         | 0.918    | 470               | 2125                   | 0.94      | 279               | 3581                   | -41%          | 69%           |
| 40         | 0.903    | 494               | 2022                   | 0.942     | 258               | 3867                   | -48%          | 91%           |
| 50         | 0.915    | 421               | 2372                   | 0.946     | 223               | 4466                   | -47%          | 88%           |
| 60         | 0.934    | 301               | 3303                   | 0.947     | 214               | 4658                   | -29%          | 41%           |

benwtrent:
@Pulkitg64 what exactly are you benchmarking? It seems like the latest version of this PR does nothing to actually correct the graph nodes?

We should handle:

 - If layers get completely removed (do we promote new nodes?)
 - Removing deleted nodes and reconnecting the neighbors to their nearest non-deleted
 - Completely throwing away the graph if deletion percentage is above a certain threshold (the first commit of hits PR had that at 30%, I think it can maybe be as high as 50%).

benwtrent:
Ah, maybe I don't fully grok the current impl. It seems like its doing the "largest graph" thing, but now its more clever and doing the initialized graph thing and that is where the deletes are being removed?

Pulkitg64:
> I am confused! This PR suddenly got so much simpler, which is great, 


Yeah, the ```initGraph``` implementation in `InitializedHnswGraphBuilder.java` simplifies lot of things for us as it is already providing support of creating OnHeapHnswGraph by passing OffHeapGraph from the older segment.

> we are no longer checking the largest graph to see if its delete % is below a threshold?

Yes, in the first revision I added the arbitrary percentage without doing any testing. But this time, I wanted to see the impact of mergePolicy that kicks in when delete % is higher than certain threshold. I thought we may not need to add explicit check of checking delete % of largest graph because merge policy will automatically take care of this.

> Also I think we are now ignoring the various edge cases around upper-level graph layers possibly becoming empty?

```initGraph``` implementation takes care of it. In the implementation we start with top level and if there is no live node in that level the new entry node is never set and when we will iterate to next level with some live nodes there we will set the new entry node. Hence in this way we remove the risk of empty upper layer in the graph. But on the other there is still risk of completely deleting middle layer which we need to take care I believe.

**Comment by Pulkitg64** [comment]
> Ah, maybe I don't fully grok the current impl. It seems like its doing the "largest graph" thing, but now its more clever and doing the initialized graph thing and that is where the deletes are being removed?

That's right @benwtrent, we are skipping deleted nodes from the largest graph in the ```initGraph``` implementation.

**Comment by benwtrent** [comment]
@Pulkitg64 pretty damn clever ;). I gotta think through this. Intuitively, it SHOULD work, even for singleton merges

**Comment by msokolov** [comment]
It's fascinating that we actually see recall *improving* in many cases! Intuitively, I think when we merge more segments in we have an opportunity to patch up the holes left by the deleted docs, and maybe we somehow end up doing that in an even better way the second time around?

I do wonder what recall will look like for graphs with high deletion rates that are singleton-merged only?  I wonder if we could test that with `luceneutil` by creating a single-segment index (with force-merge), deleting 50% of the docs, and then force-merging again?

**Comment by Pulkitg64** [comment]
Based on @msokolov  suggestion, I ran the benchmarks by simulating singleton merging. For this I  indexed 1M docs and then force merge the segments then delete documents and then again force merge the segment.

I am seeing consistent improvement (about 50x speedup) in force merge time after deletes but also degradation in recall numbers (about 10%). It's probably because of disconnectedness issue (Let me try to find connectedness number of these graphs as well.)


| Experiment | Baseline |                      | Candidate |                  | Change  |                  |
|------------|----------|----------------------|-----------|------------------|---------|------------------|
| Delete Pct | Recall   | Force Merge Time (s) | Recall    | Force Merge Time | Recall  | Force Merge Time |
| 50% delete | 0.892    | 417.52               | 0.763     | 8.43             | -14%    | 50x              |
| 40% delete | 0.887    | 505.74               | 0.799     | 9.91             | -10%    | 50x              |
| 30% delete | 0.88     | 585                  | 0.822     | 10.98            | -7%     | 53x              |
| 20% delete | 0.878    | 677                  | 0.802     | 12.4             | -9%     | 54x              |
| 10% delete |  0.874 | 772.42               | 0.856     | 13.5            |  -2%      |  59x               |

**Comment by benwtrent** [comment]
> It's probably because of disconnectedness issue (Let me try to find connectedness number of these graphs as well.)

I would think so. My gut is that we don't actually go through and "fixup" anything when there is just one graph. We just pick the biggest one, and since there are no more vectors to add, we just drop connections on the ground. 

I would expect us to have to iterate through the graph and for every vector that is significantly disconnected, attempt to reconnect it with NNDescent starting at is original place in the graph (initializing with neighbor's neighbors if all its connections were removed).

**Comment by Pulkitg64** [comment]
Thanks @benwtrent for the suggestion. For now, I am thinking that we can keep threshold of 10% deletes i.e. we will consider only those segments for merging without building graph from scratch for which delete % is less than or equal to 10%.

I can create a separate issue/PR for fixing the graph (reconnecting nodes) and try to increase delete threshold from 10%. Please let me know your thoughts


I re-ran the benchmark again with varying delete % till 15% and results are similar only.

| Experiment | Baseline |                      | Candidate |                  | Change  |                  |
|------------|----------|----------------------|-----------|------------------|---------|------------------|
| Delete Pct | Recall   | Force Merge Time (s) | Recall    | Force Merge Time | Recall  | Force Merge Time |
| 0% delete  | 0.872    | 0                    | 0.873     | 0                |         |                  |
| 2% delete  | 0.871    | 831                  | 0.866     | 13               | -1%     | 64x              |
| 5% delete  | 0.873    | 810                  | 0.863     | 13               | -1%     | 62x              |
| 8% delete  | 0.874    | 783                  | 0.861     | 13               | -1%     | 60x              |
| 10% delete | 0.874    | 773                  | 0.857     | 13               | -2%     | 60x              |
| 15% delete | 0.876    | 730                  | 0.848     | 12               | -3%     | 60x              |

Also ran with different max-conn by keeping the delete % threshold as 10%:

| Experiment |            | Baseline |                      | Candidate |                  | Change  |                  |
|------------|------------|----------|----------------------|-----------|------------------|---------|------------------|
| Max Con    | Delete Pct | Recall   | Force Merge Time (s) | Recall    | Force Merge Time | Recall  | Force Merge Time |
| 32         | 10% delete | 0.874    | 773                  | 0.857     | 13               | -2%     | 60x              |
| 16         | 10% delete | 0.811    | 550                  | 0.793     | 12               | -2%     | 45x              |
| 8          | 10% delete | 0.696    | 360                  | 0.675     | 12               | -3%     | 30x              |

Raising a new revision with the threshold limit.

**Comment by msokolov** [comment]
This looks good to me -- except -- I think there is the possibility of creeping graph rot where we continually erode the graph through repeated merges, and each time the deletion % gets reset to 0 so we think everything looks fine?  I wonder if we could instead examine some metric that persists across merges, like graph connectivity. We could (for example) check for nodes whose outdegree drops below 50% of maxconn, and in those cases, use searches to "patch up"
 the graph? We had something similar in place when we were trying to fix disconnected graphs

**Comment by benwtrent** [comment]
The nice thing about our bot going haywire, is that it reminded me of this PR. 

It seems like its dying on the vine. We got some good numbers. Is our concern that we get a disconnected graph?

I like @msokolov's suggestion, where clearing out the deleted nodes deletes a percentage, we force it to do some reconnections.

**Comment by Pulkitg64** [comment]
Hi @benwtrent 

Sorry, I was on long vacation in September, so couldn't spend time on this. But this is my main priority right now.

>  Is our concern that we get a disconnected graph?

Yes, that's right. If I keep dropping nodes from the graph without reconnecting them, then I am seeing big drop in recall numbers. In one of the knnPerfTest run, I kept deleting 10% of nodes from a segment/graph continuously and after 5 iteration, recall for that segment/graph dropped by 10%. 

For single segment graph with 1MM docs/vectors

| Iteration | Number of Nodes After Deletion | Recall % |
|-----------|--------------------------------|----------|
| 0         | 1000000                        | 81       |
| 1         | 900027                         | 79.4     |
| 2         | 810296                         | 77.9     |
| 3         | 729025                         | 75.8     |
| 4         | 656285                         | 73.9     |
| 5         | 590681                         | 71.6     |


So, as next step currently I am trying to reconnect those nodes for which outdegree is less than some threshold as pointed out by @msokolov. Will post the numbers as soon as I get some success.

**Comment by Pulkitg64** [comment]
I've raised a new [DRAFT] revision to address the graph disconnectedness concern. Here's the approach:

* Graph Selection: We select the largest graph, regardless of whether it contains deleted nodes.
* Node Pruning: We remove all deleted nodes and their associated edges from the selected graph.
* Connectivity Restoration: We identify nodes whose connectivity has dropped below 50% of maxConn. For each of these under-connected nodes, we:
    * Perform a graph search using its existing neighbors as entry points
    * Connect the search results to the node using diversity checks
* Level Rebalancing: We rebalance each level to match its expected node count by randomly promoting nodes from the level below.

I will try to share benchmark results with this approach soon. Looking forward for feedback on this approach on how should I refine this.

**Comment by Pulkitg64** [comment]
Ran `knnPerfTest` to compare baseline (current main branch) against candidate (this PR). The candidate implementation includes graph reconnection and rebalancing after deleted nodes are dropped. Benchmarks were performed for 100k docs with various deletion percentages and `maxConn` values using a single merge thread (no concurrency).

### Summary

The candidate shows better `forceMerge` performance than baseline up to `maxConn=32` without significant recall degradation. However, at `maxConn=64`, performance is comparable to baseline, likely because:

1. A high number of nodes require reconnection
2. The dense graph structure increases candidate search time

**Current Reconnection Strategy:**

We reconnect nodes when their connection count falls below 50% of the maximum allowed:
- **Level > 0**: Reconnect if connections < `maxConn / 2`
- **Level 0**: Reconnect if connections < `(2 × maxConn) / 2` = `maxConn`

**Observation at maxConn=64:**

At level 0, more than 90% of nodes (p90) have fewer than 64 connections, triggering reconnection for nearly all of them. (** I think we can try relaxing 50% threshold**)

```
Node connections stats:
Graph level=2 size=21, avg=4.67, p5=2, p10=2, p20=3, p30=3, p50=4, p90=8, p99=9, max=9
Graph level=1 size=1407, avg=27.91, p5=15, p10=17, p20=20, p30=22, p50=27, p90=40, p99=54, max=61
Graph level=0 size=90050, avg=33.03, p5=11, p10=13, p20=17, p30=20, p50=27, p90=61, p99=112, max=128
```

| Experiment |         | Baseline |                | Candidate |                | Change        |                        |
|------------|---------|----------|----------------|-----------|----------------|---------------|------------------------|
| Delete %   | MaxConn | Recall   | ForceMergeTime (sec) | Recall    | ForceMergeTime (sec) | Recall Change | ForceMerge Time Change |
| 10         | 8       | 0.796    | 73.97          | 0.783     | 18.78          | -1.63%        | 3.9x            |
| 10         | 16      | 0.89     | 109.05         | 0.884     | 46.07          | -0.67%        | 2.3x            |
| 10         | 32      | 0.926    | 136.24         | 0.923     | 101.07         | -0.32%        | 1.35x             |
| 10         | 64      | 0.935    | 148.34         | 0.935     | 155.45         | 0.00%         | 0.95x            |
| 20         | 8       | 0.8      | 63.94          | 0.783     | 18.29          | -2.13%        | 3.5x            |
| 20         | 16      | 0.895    | 95.34          | 0.881     | 44.89          | -1.56%        | 2.1x             |
| 20         | 32      | 0.931    | 116.78         | 0.924     | 91.88          | -0.75%        | 1.28x             |
| 20         | 64      | 0.938    | 130.33         | 0.935     | 135.48         | -0.32%        | 0.96x            |
| 30         | 8       | 0.808    | 55.11          | 0.784     | 18.66          | -2.97%        | 2.9x            |
| 30         | 16      | 0.9      | 83.11          | 0.885     | 42.96          | -1.67%        | 1.9x            |
| 30         | 32      | 0.935    | 100.64         | 0.927     | 81.03          | -0.86%        | 1.24x            |
| 30         | 64      | 0.942    | 111.51         | 0.938     | 111.25         | -0.42%        | 1.002x            |
| 50         | 8       | 0.822    | 38.69          | 0.798     | 23.39          | -2.92%        | 1.65x            |
| 50         | 16      | 0.914    | 56.18          | 0.87      | 42.85          | -4.81%        | 1.3x            |
| 50         | 32      | 0.943    | 70.05          | 0.935     | 61.89          | -0.85%        | 1.13x            |
| 50         | 64      | 0.951    | 72.26          | 0.942     | 71.55          | -0.95%        | 1.00x            |

**Comment by Pulkitg64** [comment]
Ran another experiment where I keep deleting 10% nodes from the graph. Started from 100k docs and ended with 49963 docs at the end of 5th iteration. This  benchmark run is similar to [this one](https://github.com/apache/lucene/pull/15003#issuecomment-3412537982), where we saw 10% degradation in recall till 5th iteration but that's not the case anymore. For maxCon=8,16, we are seeing little degradation but maxConn=32,64 is performing as good as baseline.

| MaxConn | Experiment | Iteration 1 | Iteration 2 | Iteration 3 | Iteration 4 | Iteration 5 |
|---------|------------|-------------|-------------|-------------|-------------|-------------|
| 8       | Baseline   | 0.796       | 0.800  | 0.806      | 0.808       | 0.812       | 
|         | Candidate  | 0.783       | 0.767       | 0.743       | 0.753       | 0.753       | 
| 16      | Baseline   | 0.89        |0.895    | 0.898      | 0.902       | 0.906       | 
|         | Candidate  | 0.884       | 0.877       | 0.871       | 0.868       | 0.868       | 
| 32      | Baseline   | 0.926       | 0.930   |0.933   | 0.936       | 0.939       | 
|         | Candidate  | 0.923       | 0.923       | 0.923       | 0.926       | 0.929       |
| 64      | Baseline   | 0.935       | 0.937    |0.94      | 0.945       | 0.946       | 
|         | Candidate  | 0.935       | 0.936       | 0.94        | 0.943       | 0.943       |

**Comment by benwtrent** [comment]
Love the results!

I bet the issue with higher maxconn is that the layer is sparsely connected already, meaning very few deletes throw you under the `50%` threshold. I am not sure how to calculate the "expected percentage connected", but I suspect is under 50% for high max conn.

I am guessing its too tricky given the current APIs to make that 50% relative to the connections the graph had BEFORE the deletes?

Meaning, if node `A` only had `16` connections (though maxconn was 64), we would only push it to be reconnected once it reached `8` (half its non-deleted) connections?

**Comment by Pulkitg64** [comment]
Thanks @benwtrent,

> I am guessing its too tricky given the current APIs to make that 50% relative to the connections the graph had BEFORE the deletes?

> Meaning, if node A only had 16 connections (though maxconn was 64), we would only push it to be reconnected once it reached 8 (half its non-deleted) connections?

I think this should be possible, since we are iterating over old neighbours for adding newNeighbours in InititalizedGraphBuilder. Let me try this now and get back with the results.

**Comment by Pulkitg64** [comment]
https://github.com/apache/lucene/pull/15003#issuecomment-3494558786

Continuing from above benchmark run: With higher deletes we are regressing on recall with out any benefits on force merge time. This suggest it is required to add some guardrails like we had earlier in previous revision i.e select only those graphs which have less than x% deletes and skip others. (**Maybe 20-30% deletes is fine**)

| Experiment |         | Baseline |                | Candidate |                | Change        |                        |
|------------|---------|----------|----------------|-----------|----------------|---------------|------------------------|
| Delete %   | MaxConn | Recall   | ForceMergeTime | Recall    | ForceMergeTime | Recall Change | ForceMerge Time Change |
| 60         | 8       | 0.83     | 28.69          | 0.8       | 23.81          | -3.61%        | 1.2x            |
| 60         | 16      | 0.919    | 42.81          | 0.894     | 39.07          | -2.72%        | 1.1x           |
| 60         | 32      | 0.949    | 51.6           | 0.926     | 50.38          | -2.42%        | 1.02x            |
| 60         | 64      | 0.955    | 54.78          | 0.945     | 52.98          | -1.05%        | 1.03x            |
| 70         | 8       | 0.843    | 20.84          | 0.788     | 18.39          | -6.52%        | 1.13x            |
| 70         | 16      | 0.927    | 30.39          | 0.886     | 29             | -4.42%        | 1.04x            |
| 70         | 32      | 0.955    | 36.05          | 0.936     | 34.85          | -1.99%        | 1.03x            |
| 70         | 64      | 0.961    | 37.15          | 0.945     | 38.41          | -1.66%        | 0.96x            |
| 80         | 8       | 0.864    | 12.95          | 0.732     | 10.69          | -15.28%       | 1.2x            |
| 80         | 16      | 0.94     | 18.23          | 0.821     | 16.42          | -12.66%       | 1.1x            |
| 80         | 32      | 0.965    | 20.98          | 0.914     | 19.18          | -5.28%        | 1.1x            |
| 80         | 64      | 0.968    | 21.69          | 0.916     | 20.23          | -5.37%        | 1.07x            |
| 90         | 8       | 0.893    | 5.37           | 0.39      | 2.67           | -56.33%       | 2.01x            |
| 90         | 16      | 0.961    | 7.26           | 0.562     | 5.14           | -41.52%       | 1.4x            |
| 90         | 32      | 0.978    | 8.04           | 0.678     | 6.43           | -30.67%       | 1.2x            |
| 90         | 64      | 0.98     | 8.02           | 0.758     | 6.48           | -22.65%       | 1.23x            |

**Comment by Pulkitg64** [comment]
@benwtrent on your suggestion of using a different threshold for reconnecting nodes, I re-ran benchmark with 100k docs, maxConn=32 ,10% random deletions, just to see the impact on time and recall. Here is the comparison:


* Baseline: (Current mainline code):
    * Recall: 0.926
    * forceMergeTime: 136.24
    
* Candidate-0: (This PR) (Considering nodes as disconnected for which number of connections is less than 50% of maxConn):
    * No. of disconnected nodes at level 0: 64571 (64k out of 90k)
    * Recall: 0.923
    * forceMergeTime: 99.66 sec
    
* Candidate-1  (Considering nodes as disconnected for which number of connections is less than 50% of their previous number of connections):
    * No. of disconnected nodes at level 0: 17 
    * Recall: 0.876
    * forceMergeTime: 1.24 sec
    
* Candidate-2   (Considering nodes as disconnected for which number of connections is less than 75% of their previous number of connections):
    * No. of disconnected nodes at level 0: 855
    * Recall: 0.880
    * forceMergeTime: 2.48 sec
    
* Candidate-3  (Considering nodes as disconnected for which number of connections is less than 25% of maxConn):
    * No. of disconnected nodes at level 0: 35495
    * Recall: 0.907
    * forceMergeTime: 50.95
    

(Just for comparison)
* Without any reconnection (Just dropping nodes):
    * Recall: 0.877
    * forceMergeTime:  1.12 sec
  
  
We are seeing about 4% drop in recall when switching to candidate-1 since we have almost no nodes (only 17) for reconnection.

**Comment by benwtrent** [comment]
Thank you for all the work @Pulkitg64 

OK, I think your plan from this comment (50% check) is the way to go for now. My intuition is that the relative threshold is likely better, but we would need to make the percentage lower (like 25%, or 20%, like your other benchmarks show). 

However, the 50% shows really nice improvements already, so lets move forward with that. 

As for "Reconstruct completely", yes, lets focus on 40% or fewer deleted total docs? Does that align with your benchmarking results?

If so, I think we have our two thresholds and we can move forward with this PR. 


Thank you for this practical, and powerful improvement for vector indexing!

**Comment by Pulkitg64** [comment]
> OK, I think your plan from this comment (50% check) is the way to go for now. My intuition is that the relative threshold is likely better, but we would need to make the percentage lower (like 25%, or 20%, like your other benchmarks show).

Since in my approach, even if graph has no deletes, it will still try to reconnect nodes which will make merging slower, hence I agree with you that "relative threshold" approach is likely better.  Sharing benchmark results in some time

> As for "Reconstruct completely", yes, lets focus on 40% or fewer deleted total docs? Does that align with your benchmarking results?

Keeping this threshold at 40% now.

**Comment by Pulkitg64** [comment]
Taking a sweet spot to adjust between recall and indexing performance by taking 15% as threshold for relative reduction in connection for considering a node as disconnected, I am getting below results:
|Experiment | | Baseline | | Candidate | | Change |  |
|----------|---------|--------|----------------|--------|----------------|---------------|------------------------|
| Delete % | MaxConn | Recall | ForceMergeTime | Recall | ForceMergeTime | Recall Change | ForceMerge Time Change |
| 10       | 8       | 0.796  | 73.97          | 0.755  | 18.82          | -5.15%        | 3.9x           |
| 10       | 16      | 0.89   | 109.05         | 0.853  | 28.09          | -4.16%        | 3.8x         |
| 10       | 32      | 0.926  | 136.24         | 0.923  | 31.81          | -0.32%        | 4.2x           |
| 10       | 64      | 0.935  | 148.34         | 0.935  | 35.64          | 0.00%         | 4.1x            |
| 20       | 8       | 0.8    | 63.94          | 0.783  | 38.02          | -2.13%        | 1.6x           |
| 20       | 16      | 0.895  | 95.34          | 0.881  | 63.26          | -1.56%        | 1.5x              |
| 20       | 32      | 0.931  | 116.78         | 0.916  | 90.51          | -1.61%        | 1.29x            |
| 20       | 64      | 0.938  | 130.33         | 0.928  | 102            | -1.07%        | 1.27x           |
| 30       | 8       | 0.808  | 55.11          | 0.797  | 43.72          | -1.36%        | 1.26x              |
| 30       | 16      | 0.9    | 83.11          | 0.889  | 71.38          | -1.22%        | 1.16x            |
| 30       | 32      | 0.935  | 100.64         | 0.927  | 94.53          | -0.86%        | 1.06x            |
| 30       | 64      | 0.942  | 111.51         | 0.935  | 108.52         | -0.74%        | 1.02x           |
| 40       | 8       | 0.814  | 46.78          | 0.795  | 40.57          | -2.33%        | 1.15x            |
| 40       | 16      | 0.906  | 74.6           | 0.897  | 64.11          | -0.99%        | 1.16x            |
| 40       | 32      | 0.94   | 85.76          | 0.932  | 81.03          | -0.85%        | 1.05x            |
| 40       | 64      | 0.947  | 96.93          | 0.94   | 98.95          | -0.74%        | 0.97x           |
|          |         |        |                |        |                |               |                        |

**Review by benwtrent — COMMENTED** [review]
OK, reading through it all, I really like where it ended up.

My main concern is testing. Maybe `HnswGraphTestCase.testRandomReadWriteAndMerge` is enough? Could you add some logic there for "delete few, delete many"? Just to make sure we adequately exercise this logic.


This really is great work! Thank you for all the iteration. The numbers are great!

File: lucene/core/src/java/org/apache/lucene/util/hnsw/IncrementalHnswGraphMerger.java [hunk 10]
```diff
@@ -98,7 +111,21 @@ public IncrementalHnswGraphMerger addReader(
         candidateVectorCount = vectorValues.size();
       }
     }
-    graphReaders.add(new GraphReader(reader, docMap, candidateVectorCount));
+
+    GraphReader graphReader = new GraphReader(reader, docMap, candidateVectorCount);
+
+    int graphSize = graph.size();
+    int deletePct = ((graphSize - candidateVectorCount) * 100) / graphSize;
```
kaivalnp, line 107: IIUC, this is attempting to count the percent of deleted vectors in a segment? (which will trigger a fresh graph creation if <= `DELETE_PCT_THRESHOLD`)

If so, `candidateVectorCount` is populated from `byteVectorValues.size()` / `vectorValues.size()` -- which does not incorporate deletes -- we'll need to count that separately (like in #15429)

Pulkitg64: Thanks @kaivalnp 
Fixing this in next revision along with the testing which Ben has suggested.

lucene/core/src/java/org/apache/lucene/util/hnsw/InitializedHnswGraphBuilder.java
```diff
@@ -53,49 +110,338 @@ public static InitializedHnswGraphBuilder fromGraph(
...
+  }
 
-  public InitializedHnswGraphBuilder(
-      RandomVectorScorerSupplier scorerSupplier,
-      int beamWidth,
-      long seed,
-      OnHeapHnswGraph initializedGraph,
-      BitSet initializedNodes)
+  /**
+   * Fixes disconnected nodes at a specific level by performing graph searches from their existing
```
mccullocht, line 285: https://www.vldb.org/pvldb/vol18/p5166-upreti.pdf has an in-place delete algorithm (search for algorithm 6). It may be worth investigating for this use case as it could be less expensive than repairing disconnected nodes, and the paper indicates they've successfully used this algorithm with no observed loss in accuracy.

Worth noting that this algorithm does not guarantee the removal of all in-edges to the deleted node in a directed graph, so even after performing this kind of deletion we'd have to sweep the graph and remove any dead edges.

Pulkitg64: Thanks @mccullocht for idea and sharing this paper. The idea of using the neighbors and neighbor of neighbors of the deleted nodes for reconnection seems like a good idea (definitely faster) and I think worth pursuing. 
But I wonder if we should do it as part of this PR or different one. I can work on this idea immediately as part of different issue if that makes sense to you.

mccullocht: It looks like this PR has been thoroughly tested and I didn't want to block it but trying this algorithm would be a good follow up.

Pulkitg64:
Hi @benwtrent 
Is there anything else we should add in this PR before merging it?

benwtrent:
@Pulkitg64 nope! I am back from eating a bunch of 🦃 . I will merge and backport this. Super excited for this change. Its a practical and very useful optimization for HNSW!

mikemccand:
Note that this caused some CI failures: https://github.com/apache/lucene/issues/15467 -- let's hold off on backporting until we resolve that.

benwtrent:
@mikemccand I already backported, however a fix is in flight.

mikemccand:
Wow, fast!  Thanks @benwtrent.

PR #15011, opened 2025-07-30
Add AcceptDocs abstraction for accepted KNN docs

shubhamvishu:
### Description

Addresses [this comment](https://github.com/apache/lucene/pull/14963#issuecomment-3092568029) on the PR #14963 to allow both sequential and random access consumption for accepted docs in KNN search.


lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
```diff
@@ -0,0 +1,194 @@
...
+ * random-access (Bits) or sequential (DocIdSetIterator) pattern.
+ *
+ * @lucene.experimental
+ */
+public abstract class AcceptDocs {
+
+  /**
+   * Random access to the accepted documents.
+   *
+   * @return Bits instance for random access, or null if not available
```
jpountz: I think we should apply the usual contract that a `null` Bits instance means that all docs are accepted?

```suggestion
   * @return Bits instance for random access, or null if all documents are accepted
```

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
```diff
@@ -0,0 +1,194 @@
...
+
+  /**
+   * Create AcceptDocs from a DocIdSetIterator.
+   *
+   * @param iterator a DocIdSetIterator iterator
+   * @return AcceptDocs wrapping the iterator
+   */
+  public static AcceptDocs fromIterator(DocIdSetIterator iterator, int maxDoc) {
+    if (iterator == null) {
+      return new AcceptAllDocs();
```
jpountz: In other places (e.g. `Weight#scorer`), a `null` `DocIdSetIterator` means that there are no matches. I think we should either reject null iterators or treat it as an `AcceptNoDocs`

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
```diff
@@ -0,0 +1,194 @@
...
+        if (bits instanceof BitSet bitSet) {
+          iterator = new BitSetIterator(bitSet, cost);
+        } else {
+          FixedBitSet bitSet = new FixedBitSet(bits.length());
+          for (int i = 0; i < bits.length(); i++) {
+            if (bits.get(i)) {
+              bitSet.set(i);
+            }
+          }
+          iterator = new BitSetIterator(bitSet, bitSet.cardinality());
```
jpountz: I wouldn't materlialize a `FixedBitSet`. Let's return a `FilteredDocIdSetIterator` on top of a `DocIdSetIteartor.all(bits.length())` that excludes non-matching docs?

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    @Override
+    public DocIdSetIterator getIterator() {
+      return DocIdSetIterator.all(Integer.MAX_VALUE);
+    }
+
+    @Override
+    public long cost() {
+      return Integer.MAX_VALUE;
```
jpountz: Maybe the constructor should take a `maxDoc` argument like the iterator-based impl so that we could return it here?

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
```diff
@@ -0,0 +1,194 @@
...
+   * @return approximate cost
+   */
+  public abstract long cost();
+
+  /**
+   * Create AcceptDocs from a Bits instance.
+   *
+   * @param bits the Bits instance for random access
+   * @return AcceptDocs wrapping the Bits
+   */
```
jpountz, line 74: Can you say in the javadocs that if the bits are null then this will match all docs?

lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene90/Lucene90HnswVectorsReader.java
```diff
@@ -260,7 +261,7 @@ public void search(String field, float[] target, KnnCollector knnCollector, Bits
             vectorValues,
             fieldEntry.similarityFunction,
             getGraphValues(fieldEntry),
-            getAcceptOrds(acceptDocs, fieldEntry),
+            getAcceptOrds(acceptDocs != null ? acceptDocs.getBits() : null, fieldEntry),
```
jpountz: Should we set the contract of this method so that `acceptDocs` is not allowed to be `null`? This would help save all these (annoying) null checks, and it wouldn't be a great burden on the caller side?

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
```diff
@@ -0,0 +1,194 @@
...
+  public static AcceptDocs fromConjunction(List<DocIdSetIterator> disiList, int maxDoc) {
+    if (disiList.isEmpty()) {
+      return new AcceptAllDocs();
+    }
+    if (disiList.size() == 1) {
+      return new IteratorAcceptDocs(disiList.getFirst(), maxDoc);
+    }
+    return new IteratorAcceptDocs(ConjunctionDISI.createConjunction(disiList, List.of()), maxDoc);
+  }
```
jpountz, line 91: The only call site of this factory method calls it with exactly two iterators. For simplicity, I'd rather remove this factory method and change the call site to call `fromIterator` with a `ConjunctionDISI`?

jpountz (commented):
Thank you for looking into it, it looks like this refactoring is working well. I left a first round of suggestions.

jpountz:
@shubhamvishu I hope you don't mind, I addressed my feedback directly on your branch.

jpountz:
We will also need to move the decision between approximate and exact to the codec, but it's probably best done as a follow-up PR? I think that this change is good as-is.

shubhamvishu:
> I hope you don't mind, I addressed my feedback directly on your branch.

@jpountz Not at all, thank you so much for taking care of it! Sorry I couldn’t get to it sooner. I really appreciate the assist.

&nbsp;
> We will also need to move the decision between approximate and exact to the codec, but it's probably best done as a follow-up PR? I think that this change is good as-is.

I agree completely , that is unrelated to this specific change and better to keep as separate PR.

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
shubhamvishu: I like this simplification! <3

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
```diff
@@ -0,0 +1,181 @@
...
+
+    @Override
+    public DocIdSetIterator iterator() throws IOException {
+      return new BitSetIterator(bitSet, cost());
+    }
+
+    @Override
+    public synchronized int cost() {
+      if (cardinality == -1) {
+        cardinality = bitSet.cardinality();
```
shubhamvishu: Do we need this check since we never seem to update `cardinality` value from -1?

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
```diff
@@ -0,0 +1,181 @@
...
+      return new BitSetAcceptDocs(bitSet);
+    }
+    return new BitsAcceptDocs(bits, maxDoc);
+  }
+
+  /**
+   * Create AcceptDocs from a DocIdSetIterator, optionally filtered by live documents.
+   *
+   * @param iterator a DocIdSetIterator iterator
+   * @param liveDocs Bits representing live documents, or {@code null} if
```
shubhamvishu:
```suggestion
   * @param liveDocs Bits representing live documents, or {@code null} if no deleted docs
```

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
jpountz: We do it in AbstractKnnVectorQuery when deciding between exact search and approximate search, don't we?

shubhamvishu: Oh, that's correct. I missed that.

jpountz (approved):
I think it's ready, I plan on merging in the next few days if there are no objections. cc @msokolov @benwtrent @vigyasharma since you commented on the PR that triggered this sub PR

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java
```diff
@@ -332,14 +333,15 @@ private void search(
...
+    Bits accepted = acceptDocs.bits();
+    final Bits acceptedOrds = scorer.getAcceptOrds(accepted);
     HnswGraph graph = getGraph(fieldEntry);
     boolean doHnsw = knnCollector.k() < scorer.maxOrd();
     // Take into account if quantized? E.g. some scorer cost?
     int filteredDocCount = 0;
     // The approximate number of vectors that would be visited if we did not filter
     int unfilteredVisit = HnswGraphSearcher.expectedVisitedNodes(knnCollector.k(), graph.size());
-    if (acceptDocs instanceof BitSet bitSet) {
+    if (accepted instanceof BitSet bitSet) {
```
benwtrent: Shouldn't we use `acceptDocs.cost()` instead of casting here?

benwtrent (commented):
I like this new interface. I have one comment, I think this in combination with sealing the Bits class might break Hnsw99 search. 

We should use the new interface to approximate cost, not casting the bits.

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java
jpountz: I agree it's better to do this in this PR than in a follow-up. @shubhamvishu would you like to fix this?

shubhamvishu: Sure, I'll fix this

lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99HnswVectorsReader.java
```diff
@@ -332,20 +332,19 @@ private void search(
...
-      // size as that is illogical
-      filteredDocCount = Math.min(bitSet.approximateCardinality(), graph.size());
-      if (unfilteredVisit >= filteredDocCount) {
-        doHnsw = false;
-      }
+    // Use approximate cardinality as this is good enough, but ensure we don't exceed the graph
+    // size as that is illogical
+    filteredDocCount = Math.min(acceptDocs.cost(), graph.size());
+    if (unfilteredVisit >= filteredDocCount) {
+      doHnsw = false;
```
benwtrent: ❤️

benwtrent (approved):
I really like this new API!

shubhamvishu:
@jpountz @benwtrent The [test failures](https://github.com/apache/lucene/actions/runs/16915890887/job/47929591014) seems is unrelated to this change and a side effect of [bulk scoring change](https://github.com/apache/lucene/pull/14978). The test fails on main also when HNSW is disabled to force exhaustive search and we score bulk score 64 docs now. I wonder if we still want to respect the `visitedLimit` with bulk scoring which it seems we are not doing anymore? If yes, then it seems like a existing bug which we should fix separately?. Otherwise if not, we could only fix the assertion here. Let me know what you think?

benwtrent:
@shubhamvishu could you open a test failure issue?

@ChrisHegarty what do you think?

shubhamvishu:
@benwtrent I opened an issue #15057 with the details of the test failure

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
ChrisHegarty: Was this resolved? I do think that "or null if all documents are accepted" is more clear.

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
```diff
@@ -0,0 +1,181 @@
...
+ */
+public abstract class AcceptDocs {
+
+  /**
+   * Random access to the accepted documents.
+   *
+   * @return Bits instance for random access, or null if not available
+   * @throws IOException if an I/O error occurs
+   */
+  public abstract Bits bits() throws IOException;
```
ChrisHegarty, line 43: I don't see any requirement for either this `bits` or `iterator` to throw IOException. It can be removed unless there is some other usage not currently in this PR that needs it?

ChrisHegarty (approved):
This is generally very nice. I left a few small comments.

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
```diff
@@ -0,0 +1,181 @@
...
+      return bitSet;
+    }
+
+    @Override
+    public DocIdSetIterator iterator() throws IOException {
+      return new BitSetIterator(bitSet, cost());
+    }
+
+    @Override
+    public synchronized int cost() {
```
ChrisHegarty: why is this `synchronized`?  I there any expectation of thread-safety of AcceptedDocs ?

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
```diff
@@ -0,0 +1,181 @@
...
+        return bitSet;
+      } else {
+        FilteredDocIdSetIterator filterIterator =
+            new FilteredDocIdSetIterator(iterator) {
+              @Override
+              protected boolean match(int doc) {
+                return liveDocs == null || liveDocs.get(doc);
+              }
+            };
+        return BitSet.of(filterIterator, maxDoc); // create a sparse bitset
```
ChrisHegarty:
```suggestion
        if (liveDocs != null) {
          iterator =
              new FilteredDocIdSetIterator(iterator) {
                @Override
                protected boolean match(int doc) {
                  return liveDocs.get(doc);
                }
              };
        }
        return BitSet.of(iterator, maxDoc); // create a sparse bitset
```

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
jpountz: I guess I was thinking of lazily creating the bitset in this method at first, so it would need the `IOException`. But actually we always need the cardinality of the bitset to decide between random access and sequential access now, so maybe we could simplify things and make `AcceptDocs` a record around `Bits` and a `DocIdSetIterator`. So no exception, no lazy computation of `cost()` (to you other comment).

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
jpountz: AcceptDocs should be confined to a single thread so synchronized shouldn't be required indeed.

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
jpountz: Hmm, I spoke too quickly. A flat scorer should never compute the `cost()` or use the `Bits`. So we need to both compute the cost and the `Bits` lazily.

jpountz:
One of Chris' comment helped me realize that we would still load filter matches into a `BitSet` with a flat scorer, which is undesirable. So I pushed a change that makes the loading into a bitset lazy, so that if a flat `KnnVectorsReader` only calls `AcceptDocs#iterator` but neither `#cost()` nor `#bits()` then we'd never load matches into a bitset. I had to restrict the contract of `AcceptDocs` so that it's illegal ta call `#cost()` nor `#bits()` after the iterator has advanced.

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
```diff
@@ -0,0 +1,221 @@
...
+ * random-access (Bits) or sequential (DocIdSetIterator) pattern.
+ *
+ * @lucene.experimental
+ */
+public abstract class AcceptDocs {
+
+  /**
+   * Random access to the accepted documents.
+   *
+   * <p><b>NOTE</b>: This must not be called if the {@link #iterator()} has already been used.
```
msokolov: what's the contract if `bits()` is called first, and then `iterator()`?  Is it possible the iterator will have been consumed in the process of creating the `Bits`?

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
```diff
@@ -0,0 +1,221 @@
...
+
+    /**
+     * The {@link BitSet} is computed lazily so that it is never created in case only the {@link
+     * #iterator} is needed
+     */
+    private void loadIntoBitSetIfNecessary() throws IOException {
+      // Usage of AcceptDocs should be confined to a single thread, so this doesn't need
+      // synchronization.
+      if (bitSet == null) {
+        if (iterator.docID() != -1) {
```
msokolov: should we check this even if bitSet is not null, for consistency with  the API contract??  As it is we are exposing some internal details -- it's OK to call bits(), then iterator(), consume some of the iterator, and then call bits() again ...

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
jpountz: You are right. The original iterator will be consumed to create the bit set and replaced with an iterator over the bit set.

So we should disallow calling bits() or cardinality() after iterator()?

shubhamvishu: Also if we first call the `iterator()` then we directly pass the input iterator as-is in `DocIdSetIteratorAcceptDocs` unless the `bits()` is called which then updates it to a `BitSetIterator`. How will we do that if we disallow bits() after iterator()?

jpountz: It looks like forbidding `#cost()` or `#bits()` to be called after `#iterator()` works with current impls, so I pushed this change.

If you call `#iterator()` first, you will get the original iterator, if you call it after one of the two other methods, the original iterator will be loaded into a bit set and `#iterator()` will return an iterator over the bit set.

@shubhamvishu Sorry, I don't get your point?

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
jpountz: It's not checked all the time.

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
shubhamvishu: I understand from your above main comment that we don't want to load the matches in the bitset incase of flat scorer but it seems to me that allowing would make the api contract simple(but at the cost of doing unneccessary work)....Is that correct? If yes, could we somehow avoid only undesirable loading at the same time keeping the contract simple(with no restrictions)?

Basically I was considering that if user doesn't call `bits()` or `cost()` before calling `iterator()`, we return the same iterator(without loading or creating  the BitSet), and the iterator wouldn't filter out the deletes i.e. the iterator wouldn't be a conjunction over live docs, as we previously decided. So was thinking if we could allow the caller to access both `bits` and `iterator` as needed? maybe its not that simple or the extra cost to pay if not worth?

jpountz: > it seems to me that allowing would make the api contract simple(but at the cost of doing unneccessary work)....Is that correct?

This is correct! I agree with your sentiment, I got there because I wanted to avoid some performance traps, but it has gone too far, the API contract is too complex. I pushed a change so that it's legal to call the methods in any order, it just happens to be a bit more efficient to call bits() before iterator() than iterator() before bits().

shubhamvishu: This looks good to me!

shubhamvishu:
@jpountz Thanks for the changes. It looks overall good to me. I just made 2 simple refactoring changes on top :
1) Moved `FilteredDocIdSetIterator` creation to a separate fn
2) Removed `BitSetAcceptDocs` to further simplify and merged it with `BitSetAcceptDocs`(with no overhead added and since it was not doing any actual work-more of a record kind). 

Let me know if looks good or feel free to revert any changes. I think this looks in good shape to be merged soon (if nobody has concerns)?.

----------------------------
Also ran the luceneutil benchmarks to confirm there is no regression or any sort. Looks all good.

Candidate

```
 0.515       11.386  11.121        0.977  500000   100      50       64        250     4 bits    209.48       2386.85             7         1693.01      1649.857      185.013       HNSW
 0.876       10.755  10.479        0.974  500000   100      50       64        250     7 bits    557.37        897.06             3         1872.24      1832.962      368.118       HNSW
 0.976       13.502  13.153        0.974  500000   100      50       64        250         no    236.32       2115.77             8         1504.68      1464.844     1464.844       HNSW
 ```

Baseline

```
 0.514       11.563  11.291        0.976  500000   100      50       64        250     4 bits    214.63       2329.62             7         1693.17      1649.857      185.013       HNSW
 0.878       11.002  10.710        0.973  500000   100      50       64        250     7 bits    557.80        896.37             3         1872.39      1832.962      368.118       HNSW
 0.976       13.821  13.462        0.974  500000   100      50       64        250         no    235.64       2121.88             8         1504.54      1464.844     1464.844       HNSW
 ```

lucene/core/src/java/org/apache/lucene/search/AcceptDocs.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    @Override
+    public Bits bits() {
+      return bits;
+    }
+
+    @Override
+    public DocIdSetIterator iterator() {
+      return iterator;
```
jpountz: With my most recent commit, this is now supposed to return a new iterator, so we need to move the iterator instantiation from the ctor to here?

shubhamvishu: Done. Thanks!

jpountz: Hmm, we also need to create a new iterator in the case when it's backed by DocIdSetIterator#all?

shubhamvishu: I kept it as-is as we were not creating in the case of BitsAcceptDocs in the earlier commit. But makes sense to return a new iterator irrespective.

shubhamvishu: Okay, I updated it. We always return a fresh iterator now with both APIs..

jpountz (approved):
It looks good! I added some small tests, I'll merge soon.

kaivalnp:
> We will also need to move the decision between approximate and exact to the codec

@jpountz do you mean moving the [`exactSearch`](https://github.com/apache/lucene/blob/8e8e37d9e94c290cf8d02e9f318e601baedf28bc/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java#L302) functionality down to the codec?

If so, I think it could be useful in optimizations like #12820 too, where the main overhead is passing / re-creating information from ordinal-space to docid-space (i.e. the "visited" `BitSet` in our case) -- which could be avoided if exact search was done in ordinal-space..

shubhamvishu:
@kaivalnp Yes, we want to move it down to the Codec. I'll soon post a PR for this change.

benwtrent:
FYI, for exact search to be fast, it would also be good to have AcceptDocs be able to provide you a "size estimate" that doesn't require evaluating the entire iterator. Right now I think it does that and pretty much defeats the purpose of lazy evaluation of the iterator.

Issue #13797, opened 2024-09-16
Relax Lucene Index Upgrade Policy to Allow Safe Upgrades Across Multiple Major Versions


markrmiller:
### Description

TLDR: Relax index upgrade policy across major versions to only be as strict as necessary.

Here is an attempted summary of a recent discussion about this.

Currently, Lucene's policy requires a full reindex when upgrading across more than one major version, which can create significant friction for users with large indexes. We propose relaxing this policy to allow upgrades across multiple major versions when it is safe to do so. The goal is to provide flexibility for users without compromising data integrity or flexibility.

Proposed Changes:

- Modify Upgrade Policy: Allow upgrades across multiple major versions, replacing the existing restriction with a configurable MIN_SUPPORTED_MAJOR version in Version.java.

- Controlled Version Bumping: Bump MIN_SUPPORTED_MAJOR only when necessary, due to index format changes that prevent safe upgrades (e.g., changes to norms encoding).

- Improved Documentation: Clearly document which versions can be safely upgraded to the current version without reindexing.

- Retain Reindexing When Necessary: Ensure that reindexing is still required when necessary to maintain correctness or prevent the propagation of corruption.

Benefits:

- Reduces friction and operational overhead for users with large indexes.

- Facilitates more frequent major releases by reducing mandatory reindexing.

- Maintains safety and integrity by reindexing only when required.

Implementation Plan:

- Modify Version.java to use a configurable MIN_SUPPORTED_MAJOR.

- Update the index upgrade logic to check against MIN_SUPPORTED_MAJOR rather than just the previous major version.

- Enhance documentation to provide clear guidelines on safe upgrade paths and scenarios requiring reindexing.

Request for Feedback: We welcome feedback from the community on this proposal, especially regarding its potential impact, implementation details, and any concerns about safety and backward compatibility.

jpountz:
I have had many discussions on this topic of file format bw compat over the years, because users would ideally like to think of their indexes as never expiring. If this is the problem that should be solved, then there are two main options that I can think of:
 - increasing backward compatibility of already written data,
 - performing a periodic transparent background reindexing.

[...]

PR #15012, opened 2025-07-30
Relax Lucene Index Upgrade Policy to Allow Safe Upgrades Across Multi…

markrmiller:
…ple Major Versions #13797

### DIFF
--- a/dev-tools/scripts/releaseWizard.yaml
+++ b/dev-tools/scripts/releaseWizard.yaml
@@ -300,6 +300,42 @@ groups:
     user_input: !UserInput
       prompt: Enter end date of feature freeze (YYYY-MM-DD)
       name: feature_freeze_date
+  - !Todo
+    id: evaluate_min_supported_major
+    title: Evaluate MIN_SUPPORTED_MAJOR policy
+    types:
+    - major
+    description: |
+      For major releases, evaluate whether MIN_SUPPORTED_MAJOR in Version.java should be bumped.
+      This constant controls the relaxed index upgrade policy - when format-breaking changes occur,
+      it should be updated to the new minimum supported version.
+
+      **Evaluation Checklist:**
+
+      1. **Review index format changes**: Have any format-breaking changes been introduced since the current MIN_SUPPORTED_MAJOR?
+         - Codec version changes that break backward compatibility
+         - SegmentInfo format changes
+         - New mandatory index structures
+         - Changes to core file formats
+
+      2. **Check compatibility span**:
+         - Current MIN_SUPPORTED_MAJOR: Review Version.java
+         - Proposed new value (if changes occurred): Should match the oldest version that can still be opened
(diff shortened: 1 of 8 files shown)


dweiss:
Hi Mark. https://github.com/apache/lucene/blob/main/help/formatting.txt#L32-L40 - if you have the tool installed, eclint -fix should also apply the required fixes (if possible).

dweiss:
(in fact, if you have eclint and run ```./gradlew tidy -Plucene.tool.eclint=eclint```, it'll also run the fixes.)

markrmiller:
Yeah, this isn't something that is about to be committed any second now. This is something for those looking at the Github issue to take a look it - hopefully not for the formatting at this point....

But I'll do my best not to bee line around the failing checks shown above if it does end up getting committed, though I won't deny the temptation.

dweiss:
It's fine, I just mentioned it because it's a relatively new thing and while tests may pass locally, they'll fail on gh because of extra checks performed only there.

anshumg:
The change looks good and is a rather non-invasive way to improve back-compat in future versions.

anshumg:
@markrmiller  - Unless there are fundamental objections to this approach, which doesn't seem to be the case, let's merge this in?

rahulgoswami:
@markrmiller  This is a great starting point for alleviating some of the pain with lucene version upgrades. One question - since I am not aware of any breaking changes post the norms change in 7.x/8.x (and correct me if I am wrong),  is there anything preventing setting MIN_SUPPORTED_MAJOR = 9 ?

markrmiller:
I don't think that there were any breaking changes that would prevent setting it to 9. Though of course, if this goes into 10x, in that case, you would already have the ability to upgrade from 9. For 11x, the 9x codec would have to be included (which wouldn't appear to be unusual given that 10x includes the 8x codec.) I did mention the idea of possibly setting the min major version to lower than 10 in the GitHub issue as a possibility, but I was hoping that someone more on top of whether or not there might have been a breaking change would bless it. I am pretty sure there was not one, but of course, one could still go in.

rahulgoswami:
I could try floating this up again on the Github issue to try to get someone to chime in. If people are happy with this PR and it is ripe for merging, maybe you can go ahead get this in for now(?) And then if there is agreement on setting MIN_SUPPORTED_MAJOR to 9, that can come in as another PR. Based on talks on the Solr channel, there might be a possibility that Solr 11 could closely follow Solr 10, while the latest is still 9.x. Which means having this min version set as far back as reasonable would help those installations who want to adopt Solr 11 down the line, but would have just finished reindexing for 9 (assuming they don't reindex while on 10 because they don't have to).

markrmiller:
> maybe you can go ahead get this in for now(?) And then if there is agreement on setting MIN_SUPPORTED_MAJOR to 9

+1

PR #15014, opened 2025-07-30
Parent block join knn vector query test case test random 10x

msokolov:
This is a followon from https://github.com/apache/lucene/pull/15013 rebased on branch 10x. I added grouping by block using updateDocuments so that parent/child docs would stay together. I haven't seen any failures?


msokolov:
pushed via command line

PR #15015, opened 2025-07-31
MultiIndexMergeScheduler: a production multi-tenant merge scheduler

corecursion:
This PR provides `MultiIndexMergeScheduler` for multi-tenant merge scheduling for Lucene. A version of `MultiIndexMergeScheduler` has been in production use successfully at MongoDB for many years.

Almost no changes are made to `ConcurrentMergeScheduler` by this PR.

`MultiIndexMergeScheduler` builds on existing functionality in `ConcurrentMergeScheduler` by automatically tracking merge sources and merge threads and the index `Directory` that each applies to, and then shunting all of them into a single `ConcurrentMergeScheduler` instance.

The multi-tenant merge scheduling works easily by creating a `MultiIndexMergeScheduler` instance for each index and then using each instance normally, the way you would currently use a lone `ConcurrentMergeScheduler`.

Solves issue [13883](https://github.com/apache/lucene/issues/13883).


lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java
```diff
@@ -671,6 +670,7 @@ protected synchronized MergeThread getMergeThread(MergeSource mergeSource, OneMe
     final MergeThread thread = new MergeThread(mergeSource, merge);
     thread.setDaemon(true);
     thread.setName("Lucene Merge Thread #" + mergeThreadCounter++);
+    mergeThreads.add(thread);
```
corecursion, line 673: This one line that adds a thread to the `mergeThreads` collection is moved inside the `getMergeThread()` function, making that function symmetric with the `removeMergeThread()` function that removes a thread from `mergeThreads`.

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
```diff
@@ -0,0 +1,203 @@
...
+        while (true) {
+          MergeThread toSync = null;
+          synchronized (this) {
+            for (MergeThread t : this.mergeThreads) {
+              // In case a merge thread is calling us, don't try to sync on
+              // itself, since that will never finish!
+              if (t.isAlive()
+                  && t != Thread.currentThread()
+                  // Only wait for merge threads for the current index to finish
+                  && ((InstrumentedMergeThread) t).getDirectory().equals(directory)) {
```
vigyasharma: If we used `((TaggedMergeSource) t.mergeSource).getDirectory()` instead, could we do away with `InstrumentedMergeThread` altogether?

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
```diff
@@ -0,0 +1,203 @@
...
+      } finally {
+        // finally, restore interrupt status:
+        if (interrupted) {
+          Thread.currentThread().interrupt();
+        }
+      }
+    }
+
+    @Override
+    protected synchronized ConcurrentMergeScheduler.MergeThread getMergeThread(
```
vigyasharma: We might not need to override this if we didn't need `InstrumentedMergeThread` ?

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  /**
+   * CombinedMergeScheduler is used internally by MultiIndexMergeScheduler to balance resources
+   * across multiple indices. Normally you don't need to use this.
+   *
+   * <p>For testing purposes, or if partitioning of tenants into groups is needed for some reason, a
+   * CombinedMergeScheduler can be provided to the MultiIndexMergeScheduler constructor.
+   *
+   * <p>CombinedMergeScheduler should <b><i>not</i></b> be passed directly to IndexWriter.
```
vigyasharma, line 98: Should we make this class `private static`?

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
```diff
@@ -0,0 +1,203 @@
...
+    // MultiIndexMergeScheduler to
+    // initialize the wrapped per index CombinedMergeScheduler.
+    public void setInfoStream(InfoStream infoStream) {
+      this.infoStream = infoStream;
+    }
+
+    // The implementation is copied from sync() in ConcurrentMergeScheduler in Lucene code, with
+    // only one additional check:
+    //     ((InstrumentedMergeThread) t).getDirectory().equals(directory).
+    public void close(Directory directory) {
```
vigyasharma: This doesn't really _close_ the merge scheduler. It syncs on directory specific threads. How about renaming this to `sync(Directory directory)` ?

lucene/CHANGES.txt
```diff
@@ -31,6 +29,8 @@ API Changes
 
 New Features
 ---------------------
+* GITHUB#15015: MultiIndexMergeScheduler: a production multi-tenant merge scheduler (Shawn Yarbrough)
```
vigyasharma: I think we can ship this into 10.3.

vigyasharma (commented):
Thanks for sharing this implementation, @corecursion. It's a pretty neat way to share a single CMS across multiple indexes!

By using the same CMS across indexes, you're able to prioritize and throttle merges across those indexes with a common CMS configuration. This is useful for multi-index applications in a single JVM, where indexes might have vastly different indexing rates.
This addresses the first part of the problem. If you follow the [discussions](https://github.com/apache/lucene/pull/14900#issuecomment-3084596936) on  #14900 , we also want the ability to share a threadpool b/w indexing and merging. 

I tend to think that having a shared CMS across indexes is a useful addition on its own. We can add changes for a shared indexing/merging thread pool in a separate independent change. I suggested some minor code styling comments, otherwise this looks great!

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
corecursion: I think you're right. `InstrumentedMergeThread` also did metrics gathering that I removed for this Lucene PR.

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
corecursion: Currently, the user needs access to the shared CMS for calling configuration functions such as `setDefaultMaxMergesAndThreads()` or `setMaxMergesAndThreads()` or possibly others. I wasn't able to get the unit tests to pass without that setup configuration.

There may be some better way to handle the shared configuration as a matter of UI design, that would then allow us to make `CombinedMergeScheduler` private, but I'm still thinking about what could work.

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
corecursion: renamed in [1f761b3](https://github.com/apache/lucene/pull/15015/commits/1f761b3f6a6180f6e023c448b285e73ddf99e94d)

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
corecursion: removed `InstrumentedMergeThread` in [1f6e301](https://github.com/apache/lucene/pull/15015/commits/1f6e3016991f9b475f09ddcf165248a3945188fb)

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
corecursion: removed `getMergeThread()` in [1f6e301](https://github.com/apache/lucene/pull/15015/commits/1f6e3016991f9b475f09ddcf165248a3945188fb)

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
```diff
@@ -0,0 +1,169 @@
...
+        while (true) {
+          MergeThread toSync = null;
+          synchronized (this) {
+            for (MergeThread t : this.mergeThreads) {
+              // In case a merge thread is calling us, don't try to sync on
+              // itself, since that will never finish!
+              if (t.isAlive()
+                  && t != Thread.currentThread()
+                  // Only wait for merge threads for the current index to finish
+                  && ((TaggedMergeSource) t.mergeSource).getDirectory().equals(directory)) {
```
vigyasharma, line 182: Maybe also check `t.mergeSource instanceof TaggedMergeSource` ?

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
vigyasharma: What if we exposed the internal CMS through a descriptively named getter like: 'ConcurrentMergeScheduler getSharedConcurrentMergeScheduler();`. Since it returns a CMS, the additional wrappers of CombinedCMS remain private, while users get access to the configuration setters?

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  @Override
+  public Directory wrapForMerge(MergePolicy.OneMerge merge, Directory in) {
+    return this.combinedMergeScheduler.wrapForMerge(merge, in);
+  }
+
+  /** Close this scheduler for one directory/index. Called automatically by IndexWriter. */
+  @Override
+  public void close() throws IOException {
```
vigyasharma, line 77: This makes sense. Any thoughts on how to close the internal CMS when we're finally done and shutting down? It has an intraMerge executor that needs to be closed. But there could be other indexes that are still using the CMS.

lucene/CHANGES.txt
vigyasharma: Let's move this to 10.3 section in the `CHANGES` file.

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
vigyasharma: We could keep an atomic reference counter in CombinedMS, exposing the singleton instance via an `acquire()` method that also updates the counter. MultiIndexMS would `release()` the singleton on `close()` which would also decrement the counter. CombinedMS.close() can then check that no references are still held before closing the underlying cms.

vigyasharma:
Let us also make this class `@lucene.experimental`; in case we need to change some interfaces or class structure as we progress with #13883

lucene/CHANGES.txt
corecursion: already done in [ac0c902](https://github.com/apache/lucene/pull/15015/commits/ac0c902374f68b71e70b8ff084d3fc20ee6464c1), I should have mentioned

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
corecursion: yes, sounds good, I'll push a commit for it

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
corecursion: I think this is okay (I had the same idea) but I'm wondering if the `CombinedMergeScheduler` should come with reasonable defaults that are usable for most people so they don't have to call `setMaxMergesAndThreads()` etc.? (Same logic could be applied to the vanilla CMS, I suppose.)

corecursion: If users really do need to specify a `setDefaultMaxMergesAndThreads()` `true`/`false` type choice, maybe they can call a static function such as a `MergeScheduler.setDefaultMaxMergesAndThreads()` to specify their process default instead of having to track down and configure each CMS instance.

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
corecursion: added in [b3942dd](https://github.com/apache/lucene/pull/15015/commits/b3942dd015b38d748d51a3ca39fb85962937bf20)

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
vigyasharma: Setting max threads and merges is a common requirement, I think. We override it for the CMS in Amazon Search, and I suspect many other Lucene users do as well. Static setters don't fit well with CMS which is created per IW. Adding them to CombinedMS wouldn't work for users who setup index groups that share a CMS, by passing CombinedMS in the ctor.

I think it is okay to let users directly work with public APIs for the underlying CMS. We can keep them informed via doc strings and mark the getter as a "expert" functionality.

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
```diff
@@ -0,0 +1,171 @@
...
+
+      @Override
+      public void merge(MergePolicy.OneMerge merge) throws IOException {
+        this.in.merge(merge);
+      }
+    }
+
+    // Same functionality as calling initialize() in MergeScheduler, used by
+    // MultiIndexMergeScheduler to
+    // initialize the wrapped per index CombinedMergeScheduler.
```
vigyasharma: This is helpful to understand the PR, but we can remove these comments from the final code before merging.

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
```diff
@@ -0,0 +1,171 @@
...
+    public void setInfoStream(InfoStream infoStream) {
+      this.infoStream = infoStream;
+    }
+
+    // The implementation is copied from sync() in ConcurrentMergeScheduler in Lucene code, with
+    // only these additional checks:
+    //     && t.mergeSource instanceof TaggedMergeSource
+    //     && ((TaggedMergeSource) t.mergeSource).getDirectory().equals(directory).
```
vigyasharma: Same for these comments, we can remove them from the code.

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
corecursion: removed in [7a4efa1](https://github.com/apache/lucene/pull/15015/commits/7a4efa18fcc43b1cfb097de66725f64d6193f8d9)

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
corecursion: removed in [7a4efa1](https://github.com/apache/lucene/pull/15015/commits/7a4efa18fcc43b1cfb097de66725f64d6193f8d9)

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
corecursion: Making the shared CMS class be `private static` also breaks the unit tests and the `MultiIndexMergeScheduler(
      Directory directory, CombinedMergeScheduler combinedMergeScheduler)` constructor which might have value for user customization such as if the user wants to manage tenants in groups for some reason.

I'm still willing to do it but then I'll have to rethink the unit tests.

vigyasharma: Don't want to overcomplicate things. It's fine to keep it public too. We probably need it to let people create index groups that share a CMS? Let's just add the "expert" getter for internal CMS object that backs a MultiIndexMS.

corecursion:
added `@lucene.experimental` tag in [d0eb877](https://github.com/apache/lucene/pull/15015/commits/d0eb877745932740eeab964b70a65236ca66ab9b)

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
corecursion: Added reference counting for the singleton CMS in: [677ae60](https://github.com/apache/lucene/pull/15015/commits/677ae60457bf0bade5ba31296e9673616763f904)

The refcount is synchronized because, for this class, atomics weren't worth the complexity when considering that the singleton needs to be recreated if the refcount drops to zero and then is incremented again.

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
```diff
@@ -0,0 +1,206 @@
...
+   * <p>For testing purposes, or if partitioning of tenants into groups is needed for some reason, a
+   * CombinedMergeScheduler can be provided to the MultiIndexMergeScheduler constructor.
+   *
+   * <p>CombinedMergeScheduler should <b><i>not</i></b> be passed directly to IndexWriter.
+   */
+  static class CombinedMergeScheduler extends ConcurrentMergeScheduler {
+    private static CombinedMergeScheduler singleton = null;
+    private static int singletonRefCount = 0;
+
+    private static synchronized CombinedMergeScheduler incrementSingletonReference() {
```
vigyasharma: Minor: I feel this method name should indicate that it's returning the singleton. How do you feel about renaming it to something like "acquireSingleton" / "fetchSingleton"  / "getSingleton" ? And the corresponding decrement function could be "releaseSingleton" ?

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
```diff
@@ -0,0 +1,206 @@
...
+      return singleton;
+    }
+
+    private static synchronized void decrementSingletonReference() throws IOException {
+      if (singletonRefCount < 1) {
+        throw new IllegalStateException("decrementSingletonReference() called too many times");
+      }
+      singletonRefCount--;
+      if (singletonRefCount == 0) {
+        singleton = null;
```
vigyasharma, line 122: I think we should call `singleton.close()` before we dereference it, so that base class resources like intraMergeExecutor thread pool can be released.

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
```diff
@@ -0,0 +1,206 @@
...
+  private final boolean manageSingleton;
+
+  /** The main MultiIndexMergeScheduler constructor -- use this one. */
+  public MultiIndexMergeScheduler(Directory directory) {
+    this.directory = directory;
+    this.combinedMergeScheduler = CombinedMergeScheduler.incrementSingletonReference();
+    this.manageSingleton = true;
+  }
+
+  /** Alternate MultiIndexMergeScheduler constructor for unit testing or tenant grouping. */
```
vigyasharma: It feels trappy to allow users to create and tenant share a CombinedMergeScheduler without the reference counting safety that we've added for the singleton. Can we skip supporting tenant groups and remove this ctor? Or maybe make it pkg private for testing (not exactly safe but makes the intent explicit).

vigyasharma (commented):
Thanks for all the iterations @corecursion, this looks great! I have one main comment about calling `close` when refCount drops to 0, and a couple of minor suggestions, otherwise, we're almost there.

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
corecursion: Ha, yes, calling `close()` was the point of the reference counting, wasn't it? Fixed in: [d4eec7a](https://github.com/apache/lucene/pull/15015/commits/d4eec7a7a4ad66ddac4c27a5a3325a028075272a)

(The `close()` got lost when the code was switched from atomics to synchronized.)

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
corecursion: Renamed in: [4bb674c](https://github.com/apache/lucene/pull/15015/commits/4bb674c7885b68c8d65bff9cc50c16824b2ce694)

lucene/core/src/java/org/apache/lucene/index/MultiIndexMergeScheduler.java
corecursion: Made the alternate constructor be package private in: [7a6ee71](https://github.com/apache/lucene/pull/15015/commits/7a6ee719bb700e2a199a0aa5c9c3558d2370b911)

Also added a comment that the CMS will not be closed.

lucene/CHANGES.txt
```diff
@@ -7,8 +7,6 @@ http://s.apache.org/luceneversions
 
 API Changes
 ---------------------
-* GITHUB#11023: Removing deprecated parameters from CheckIndex. (Jakub Slowinski)
```
vigyasharma: We shouldn't be removing this, is this a merge / rebase side-effect?

vigyasharma: This is also why the bot is setting this PR milestone to 11.0. Let's fix this and we can merge.

corecursion: Fixed in: [f2f1d4b](https://github.com/apache/lucene/pull/15015/commits/f2f1d4bad32dc910b32f5cf9392bd560ba2d6191)

Not sure what happened there?

vigyasharma (approved):
Wonderful! Appreciate all the work on this Shawn.

corecursion:
Thanks for all of the helpful feedback, @vigyasharma .

dweiss:
Hello. There is a regression caused by this patch that affects one of the tests - this is a racy concurrency issue so a patch needs to be applied to showcase it. I do see some changes in this patch applied to concurrent merge scheduler that I don't fully understand - perhaps it's introduced a regression that causes that other test to fail?

PR #15016, opened 2025-07-31
Refactor computeCommonPrefixLengthAndBuildHistogram

HUSTERGS:
### Description

Since jdk24 is required (#14533) maybe it's safe to revert #12905, which was introduced to work around a JDK bug.

This pr essentially reverts #12905 (but kept the added crash test), so the related code under `MSBRadixSorter` and 
`RadixSelector` can be more clear


HUSTERGS:
Since the `branch_10x` still has minimal jdk requirement of 21, and it's not a important change, so I'd like to close this PR until things change.

PR #15019, opened 2025-07-31
Don't revisit knn query results unless optimistic collection was used…

### LINKED ISSUE #15005: Optimistic knn query breaks nested vector search
Early testing of Elasticsearch with the branch_10x, currently accumulating changes for Lucene 10.3, shows that knn search over nested vector fields returns fewer than expected k hits. I believe that this is related to the recent change for optimistic knn query execution [1], but I've not yet confirmed this. For now, I have a reproducer in Elasticsearch (see [2]), but am working to extract that out into a Lucene-only test.

[1] https://github.com/apache/lucene/issues/14962
[2] https://github.com/elastic/elasticsearch/issues/131749

msokolov:
fixes #15005, I think

@ChrisHegarty if you have a chance to cherry-pick this and see if it addresses the issue you raised, that would be great. I think it will. Anyway it fixes something related. I chose to go with the option that disables revisitation for all KnnCollectorManagers except the one specifically used in the default implementation. In the case of the parent-join collection, the revisitation is just going to be wasted work, even if we were to make it function correctly, and in general, any custom collector is expected to have its own collection logic, so we shouldn't be making assumptions about it.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
@@ -106,9 +106,9 @@ abstract class AbstractKnnVectorQuery extends Query {
       filterWeight = null;
     }
 
+    KnnCollectorManager knnCollectorManagerInner = getKnnCollectorManager(k, indexSearcher);
     TimeLimitingKnnCollectorManager knnCollectorManager =
-        new TimeLimitingKnnCollectorManager(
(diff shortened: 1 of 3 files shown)


tteofili (approved):
LGTM

ChrisHegarty (approved):
I applied the fix to a local branch_10x and verified it against the reproducer in Elasticsearch, and the fix resolves the issue. 

LGTM.  Thanks @msokolov

msokolov:
thanks!

PR #15021, opened 2025-08-01
Return the max score from `RandomVectorScorer.bulkScore`

mccullocht:
Return a little extra information from the `bulkScore()` as this will allow callers to skip results that are
below the minimum similarity needed to appear in the result set. Utilize this in hnsw and exhaustive search
to skip any topN heap manipulation for any bulk scored batches that do not meet the minimum score.

See #14013


benwtrent:
Yeah! This is what I had in mind. I would expect the underlying bulk scorers should be able to keep track of their max score pretty easily (without much overhead), allowing for the caller to bypass looping over uncompetitive scores

mccullocht:
Integrated this into the new off heap bulk scorers.

My intuition was that I'd want to use vector instructions for this but layering it is kind of tricky since we don't want to use the vector API in any public interfaces. We could expose a VectorUtil.arrayMax() but it might be faster to use scalar instructions as we traverse the array.

PR #15035, opened 2025-08-06
Add some basic tests and infra to allow to write small sanity tests for benchmarks

ChrisHegarty:
This commit adds some basic tests and infra to allow to write small sanity tests for benchmarks.

I'm proposing this since I ran into issues when working on a separate change where this would have been helpful. Additionally, it allows to keep a number of assertions and sanity checks out of the benchmark code itself.


ChrisHegarty:
@dweiss I'm adding a new dependency from the benchmarks to the test-framework in this PR, so I ran `gradlew writeLocks`. Do the changes in the versions.locks look right? Or did my machine / setup bork this?  (this is the first time that I'm doing this)

dweiss (commented):
Yes, the lock file look ok - the changes include increased transitive counts (and thus hashes of identical configuration groups).

Extending LuceneTestCase will have side-effects (random setup of components). Is this expected for a benchmark?

ChrisHegarty:
Oh yeah, that's a good point.  I mostly want it for test infra (not the benchmark itself), so I moved it to `moduleTestImplementation`.

benwtrent (approved):
Yes! We need to make sure our benchmark results are correct. Tests are good :)

dweiss (approved):

PR #15040, opened 2025-08-08
Change default ReadAdvice from RANDOM to NORMAL

### LINKED ISSUE #14408: Examine the affects of MADV_RANDOM when MGLRU is enabled in Linux kernel
With the relatively recent capability to call `madvise` in Lucene, we've started to use `MADV_RANDOM` in several places where it makes conceptual sense, e.g. for accessing vector data when navigating the graph. The memory access is truly random, but we've seen several reports of performance regressions that appear as a result of this.

Of particular concern is the interaction of `MADV_RANDOM` with Multi-Gen LRU [1]. From my reading of the code, and someone please correct me,  the semantics of `MADV_RANDOM` has changed in the kernel with MGLRU, and results in pages being proactively reclaimed more eagerly, even when there is no memory pressure. Specifically after https://github.com/torvalds/linux/commit/8788f6781486769d9598dcaedc3fe0eb12fc3e59.

This Elasticsearch issue has the more of the lower-level details, https://github.com/elastic/elasticsearch/issues/124499.

This issue may also have some connection, https://github.com/apache/lucene/issues/14281.

I opened this issue to help facilitate a discussion and hopefully converge on a potential direction to mitigate the possibility of performance regressions. For example, one possible mitigation would be to expose the `ReadAdivce` that will be used as part of the API, so that callers can have more fine-grained control over whether or not to use `MADV_RANDOM`.

[1] https://docs.kernel.org/admin-guide/mm/multigen_lru.html

ChrisHegarty:
This commit changes the default ReadAdvice from RANDOM to NORMAL. 

After this change MMapDirectory will not set any specific read advice out-of-the-box. Rather there is a straightforward way to enable either previous behaviour ( in 10.2 ), by setting either:
1.  `dir.setReadAdvice(MMapDirectory.ADVISE_BY_CONTEXT)`, or
2. setting the system property `-Dorg.apache.lucene.store.defaultReadAdvice=RANDOM`.

closes #14408

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/store/TestMMapDirectory.java
+++ b/lucene/core/src/test/org/apache/lucene/store/TestMMapDirectory.java
@@ -21,6 +21,7 @@ import static java.util.stream.Collectors.toList;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


msokolov:
This seems like good progress.  I'm also unsure about `Lucene99FlatVectorsReader.finishMerge()`  where we call 
`this.vectorData.updateReadAdvice(ReadAdvice.RANDOM);` and related code in that class that was added in #13985 -- going back to read that now

msokolov (approved):

msokolov:
Yeah, I worry that this now conflicts with what we did in the vectors reader, since the default will be normal, but after merging any vectors segments we will switch back to random?

ChrisHegarty:
@msokolov I agree - there is more needed here. Let's get #14844 in first, then we can revaluate if/how to do this properly in this PR.

lucene/core/src/test/org/apache/lucene/store/TestMMapDirectory.java
```diff
@@ -131,6 +132,57 @@ public void testWithNormal() throws Exception {
     }
   }
 
+  public void testAdviseByContextFunc() throws IOException {
+    var func = MMapDirectory.ADVISE_BY_CONTEXT;
+    var flush = IOContext.flush(new FlushInfo(1, 2));
+    var merge = IOContext.merge(new MergeInfo(1, 2, true, 4));
+    var random = new DefaultIOContext(DataAccessHint.RANDOM);
```
thecoop, line 140: Can we iterate over the various hint type arrays rather than hard-coding so this doesn't need to be modified if more are added?

ChrisHegarty: Good idea. Done.  5cfe3fe

thecoop (approved):

PR #15045, opened 2025-08-11
Use FixedBitSet#cardinality for counting liveDocs in CheckIndex

easyice:
This uses `FixedBitSet#cardinality` to speed up counting liveDocs in CheckIndex and some assert implementations, instead of checking bits one by one.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -239,6 +239,8 @@ Optimizations
 
 * GITHUB#14980: Add bulk off-heap scoring for float32 vectors (Chris Hegarty)
 
+* GITHUB#15045: Use FixedBitSet#cardinality for counting liveDocs in CheckIndex (Zhang Chao)
+
 Changes in Runtime Behavior
 ---------------------
 * GITHUB#14823: Decrease TieredMergePolicy's default number of segments per
--- a/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
+++ b/lucene/core/src/java/org/apache/lucene/index/CheckIndex.java
@@ -1297,12 +1297,10 @@ public final class CheckIndex implements Closeable {
         if (liveDocs == null) {
           throw new CheckIndexException("segment should have deletions, but liveDocs is null");
         } else {
-          FixedBitSet bitSet = new FixedBitSet(liveDocs.length());
-          bitSet.set(0, liveDocs.length());
-          liveDocs.applyMask(bitSet, 0);
-          if (bitSet.cardinality() != numDocs) {
+          int numLive = bitsCardinality(liveDocs);
+          if (numLive != numDocs) {
             throw new CheckIndexException(
-                "liveDocs count mismatch: info=" + numDocs + ", vs bits=" + bitSet.cardinality());
+                "liveDocs count mismatch: info=" + numDocs + ", vs bits=" + numLive);
           }
         }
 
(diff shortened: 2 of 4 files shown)


jpountz (commented):
FYI I have an in-progress PR that would break this optimization: #14996. Furthermore, in the typical case, live docs are not as instance of `FixedBitSet` but of `FixedBits` (the result of `FixedBitSet#asReadOnlyBits`) so I don't think it would help much?

easyice:
Thanks for the review Adrien, sorry for not making it clear, this change also use `instanceof FixedBits` to enable the optimization, it will call `fixedBits.bitSet` to get the `FixedBitSet` instance, but just placed the new method `cardinality(Bits, int, int)` in class `FixedBitSet`, so it is also compatible with #14996 in my understand? 

I had considered another approach: placing the new `cardinality` method in `Bits`, similar to `applyMask`, but it seems that only a util method is needed here, WDYT?

jpountz (commented):
Thanks for explaining. I'm not not too fond of the approach, it looks like you'd really want to add `int Bits#cardinality()`, but also don't want to add it to keep `Bits` lean (which I appreciate). But it looks a bit odd to me.

If we'd like to speed these things up, maybe we should allocate a `FixedBitSet(1024)`, copy the content of the `Bits` into this `FixedBitSet` using `applyMask` and then call `cardinality()` on the `FixedBitSet`?

easyice:
It s a nice idea!  although it requires allocating an `FixedBitSet(bits.length())`, it is still much faster than checking bits one by one.

Here are some JMH numbers:

```
Benchmark                                     (density)  (size)   Mode  Cnt   Score   Error   Units
FixedBitSetBenchmark.countWithCardinality           0.5    1024  thrpt    5  53.244 ± 3.754  ops/us
FixedBitSetBenchmark.countWithFallbackGet           0.5    1024  thrpt    5   1.912 ± 0.090  ops/us
FixedBitSetBenchmark.countWithFixedBitSetGet        0.5    1024  thrpt    5   1.758 ± 0.047  ops/us
```

`countWithCardinality` this approach, use `applyMask` and `cardinality`
`countWithFallbackGet`  uses a subclass of `Bits` to mock a `Bits` object that is not an instance of `FixedBitSet`
`countWithFixedBitSetGet` use `FixedBitSet#get`


<details>
<summary >Code</summary>

```java

@BenchmarkMode(Mode.Throughput)
@OutputTimeUnit(TimeUnit.MICROSECONDS)
@State(Scope.Benchmark)
@Warmup(iterations = 3, time = 3)
@Measurement(iterations = 5, time = 5)
@Fork(1)
public class FixedBitSetBenchmark {

  @Param({"1024"})
  private int size;

  @Param({"0.5"})
  private float density; // the percentage of 1 in the bitset

  private Bits bits; // FixedBitSet#asReadOnlyBits 

  Bits fallbackBits; // will not use FixedBitSet

  @Setup(Level.Trial)
  public void setup() {
    FixedBitSet bitSet = new FixedBitSet(size);
    int numSet = (int) (size * density);

    if (numSet == size) {
      bitSet.set(0, size);
    } else if (numSet > 0) {
      Random random = new Random(0);
      for (int i = 0; i < numSet; i++) {
        bitSet.set(random.nextInt(size - 1));
      }
    }
    bits = bitSet.asReadOnlyBits();

    fallbackBits =
        new Bits() {
          @Override
          public boolean get(int index) {
            return index % 2 == 0;
          }

          @Override
          public int length() {
            return size;
          }
        };
  }

  @Benchmark
  public void countWithCardinality(Blackhole bh) {
    int count = 0;
    FixedBitSet bitSet = new FixedBitSet(size);
    bitSet.set(0, size);

    bits.applyMask(bitSet, 0);
    count = bitSet.cardinality();
    bh.consume(count);
  }

  @Benchmark
  public void countWithFixedBitSetGet(Blackhole bh) {
    int count = 0;
    for (int i = 0; i < bits.length(); i++) {
      if (bits.get(i)) {
        count++;
      }
    }
    bh.consume(count);
  }

  @Benchmark
  public void countWithFallbackGet(Blackhole bh) {
    int count = 0;
    for (int i = 0; i < fallbackBits.length(); i++) {
      if (fallbackBits.get(i)) {
        count++;
      }
    }
    bh.consume(count);
  }
}

```
</details>

jpountz:
We don't actually need to allocate a FixedBitSet of size maxDoc, we could copy slices of 1024 bits into a FixedBitSet(1024) to do the counting?

easyice:
No problem. I will update it.

easyice:
The new approach is similar to #14998, so I reused part of the code. 

The changes touch `PendingDeletes` and `CheckIndex`, which would introduce some duplicate code (the new `bitsCardinality` method).  Since there doesn’t seem to be a suitable util class for this method, and the modifications in `PendingDeletes` were only related to assertions and not essential, I reverted those changes.  

The optimization is now applied only to `CheckIndex` to keep the code clean.

jpountz (approved):

PR #15033, opened 2025-08-05
Removing unused parameter in MemorySegmentIndexInput

jainankitk:
### Description

Cleans up unused parameter in `MemorySegmentIndexInput`

PR #15047, opened 2025-08-11
Fixing build failure due to unnamed variable

### LINKED ISSUE #15046: lucene/distribution.tests failing for branch_10x
### Description

Build is failing with the following error:
After backporting the change #15033 on `branch_10x`, build is failing due to unnamed variables error. Working on fixing that

```
/Users/runner/work/lucene/lucene/lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java:330: error: unnamed variables are a preview feature and are disabled by default.
> Task :lucene:core:compileMain21Java FAILED
    } catch (IndexOutOfBoundsException _) {
gradle/actions: Writing build results to /Users/runner/work/_temp/.gradle-actions/build-results/__run-1754950346869.json
                                       ^
  (use --enable-preview to enable unnamed variables)
1 error


[Incubating] Problems report is available at: file:///Users/runner/work/lucene/lucene/build/reports/problems/problems-report.html

10 actionable tasks: 10 executed
* What went wrong:
Execution failed for task ':lucene:core:compileMain21Java'.
  /Users/runner/work/lucene/lucene/lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java:330: error: unnamed variables are a preview feature and are disabled by default.
      } catch (IndexOutOfBoundsException _) {
                                         ^
    (use --enable-preview to enable unnamed variables)
  1 error

* Try:
> Check your code and dependencies to fix the compilation error(s)

BUILD FAILED in 1m 38s
```

_No response_

jainankitk:
### Description
Resolves #15046. Related to #15033

### DIFF
--- a/lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
+++ b/lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInput.java
@@ -327,7 +327,7 @@ abstract class MemorySegmentIndexInput extends IndexInput
         this.curSegment = seg;
       }
       this.curPosition = Objects.checkIndex(pos & chunkSizeMask, curSegment.byteSize() + 1);
-    } catch (IndexOutOfBoundsException _) {
+    } catch (IndexOutOfBoundsException unnamedVariable) {
       throw handlePositionalIOOBE("seek", pos);
     }
   }
@@ -443,7 +443,7 @@ abstract class MemorySegmentIndexInput extends IndexInput
     try {
       final int si = (int) (pos >> chunkSizePower);
       return segments[si].get(LAYOUT_BYTE, pos & chunkSizeMask);
-    } catch (IndexOutOfBoundsException _) {
+    } catch (IndexOutOfBoundsException unnamedVariable) {
       throw handlePositionalIOOBE("read", pos);
     } catch (NullPointerException | IllegalStateException e) {
       throw alreadyClosed(e);
@@ -485,7 +485,7 @@ abstract class MemorySegmentIndexInput extends IndexInput
         curAvail = segments[si].byteSize();
       }
       MemorySegment.copy(segments[si], LAYOUT_BYTE, pos, b, offset, len);
-    } catch (IndexOutOfBoundsException _) {
+    } catch (IndexOutOfBoundsException unnamedVariable) {
       throw handlePositionalIOOBE("read", pos);
     } catch (NullPointerException | IllegalStateException e) {
       throw alreadyClosed(e);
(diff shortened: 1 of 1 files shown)


uschindler:
Hi,
sorry why did you backport the original PR at all. The usused variable was exactly added to suppress the warnings.

In Java 24 there is the unnamed variable `_`, but in Java 21 the original code was much easier to read. Please revert this and the original PR and restore the state which was totally applicable for Java 21. The unused variable was just there for better readability and has no performance impacts as this is exception handling code only.

After reverting please move the changes entry to 11.0 section.

Thanks.

uschindler (commented):
Please revert this and previous backported PR.

uschindler:
I reverted that in 10.x branch.

jainankitk:
I see that the change has already been reverted. I backported the change without realizing that unnamed variable is preview feature in JDK21. And then fixed by replacing the same with dummy variable name, instead of reverting the change itself. 

Although, I am wondering why `RuntimeException unused` as method parameter does not throwing warning, whereas `IndexOutOfBoundsException unnamedVariable` does. I saw we are suppressing few other places using the annotation `@SuppressWarnings("unused")`, but doesn't seem to be the case here? Is it being suppressed some other way?

uschindler:
Unused parameters only cause warnings for private methods. In all other cases they don't. That's easy to explain: only if something is fully private the method parameters can't be used outside of the class. If the method is visible the parameters are part of (semi-)public signature. Then they are "relevant".

jainankitk:
> If the method is visible the parameters are part of (semi-)public signature. Then they are "relevant".

Ah, that's smart. Thanks for the explanation! While trying to understand this, I noticed quite a few occurrences of `@SuppressWarnings("unused")` which can be replaced with unnamed variable for 11.0 atleast. I will raise few PRs to address that. First of many of those - https://github.com/apache/lucene/pull/15102

uschindler:
I approved this PR already.

The problem with the 10.x branch and changes in MemorySegment code are that we don't run ecj linter checks there (in the MRJAR, as ECJ not easily supports those apijar stub JARs). But when I opened the 10.x branch in Eclipse on my local machine I was flooded with those "unused" warnings caused by your fix. Therefore I asked to revert.

Actually, the code like it was before (with the usused variable) was also not "bad" from beginning as it was consistent! The MMap code relies on that "exception handler logic" and has always same pattern in exception handling. For better readbility, I refactored the "catch" block to be separate methods. All those methods take the thrown exception as first parameter. Just the one you fixed here did not use the parameter (but it could at a later version of the code). The other catch handler method handling NullPointerException and other RuntimeExceptions uses the exception parameter to make decissions if the index input was closed or if it is some "real" problem.

So from the code consistency the code in 10.x is more consistent now than the code in main. But I am fine with the main change, just the backport broke warnings.

PR #15050, opened 2025-08-12
Use switch expression for optimized type checks in PointRangeQueryBuilder

sandeshkr419:
### Description

`PointRangeQueryBuilder` code refactoring:
- Replace multiple `if-else` with `switch` expression
- Store `type` as lowercase to avoid checking with ignore-case later

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -63,6 +63,8 @@ Optimizations
 
 * GITHUB#14874: Improve off-heap KNN byte vector query performance in cases where indexing and search are performed by the same process. (Kaival Parikh)
 
+* GITHUB#15050: Use switch expression for optimized type checks in PointRangeQueryBuilder (Sandesh Kumar)
+
 Bug Fixes
 ---------------------
 * GITHUB#14049: Randomize KNN codec params in RandomCodec. Fixes scalar quantization div-by-zero
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -63,8 +63,6 @@ Optimizations
 
 * GITHUB#14874: Improve off-heap KNN byte vector query performance in cases where indexing and search are performed by the same process. (Kaival Parikh)
 
-* GITHUB#15050: Use switch expression for optimized type checks in PointRangeQueryBuilder (Sandesh Kumar)
-
 Bug Fixes
 ---------------------
 * GITHUB#14049: Randomize KNN codec params in RandomCodec. Fixes scalar quantization div-by-zero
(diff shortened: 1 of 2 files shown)


lucene/CHANGES.txt
```diff
@@ -63,6 +63,8 @@ Optimizations
 
 * GITHUB#14874: Improve off-heap KNN byte vector query performance in cases where indexing and search are performed by the same process. (Kaival Parikh)
 
+* GITHUB#15050: Use switch expression for optimized type checks in PointRangeQueryBuilder (Sandesh Kumar)
+
```
jainankitk: This changelog should be part of 10.3? Also, I will move it to the `Other` section as it probably does not have any measurable impact on the benchmark?

jainankitk (approved):
While the code change looks good to me, test failed for some reason:

Could not repro it on my mac (maybe since the failure is for windows):
```
./gradlew :lucene:core:test --tests "org.apache.lucene.TestMergeSchedulerExternal.testSubclassConcurrentMergeScheduler" -Ptests.file.encoding=UTF-8 -Ptests.gui=true -Ptests.jvmargs= -Ptests.jvms=2 -Ptests.seed=509A6FBF278B9673 -Ptests.vectorsize=256
```

Have retried the tests, will wait for checks to be green!

lucene/CHANGES.txt
sandeshkr419: Thanks. Moved to 10.3 / Other.

asimmahmood1 (approved):

jainankitk:
Backported into 10x:

```
To github.com:apache/lucene.git
   9135700a59f..3a99fae956f  branch_10x -> branch_10x
```

lucene/queryparser/src/java/org/apache/lucene/queryparser/xml/builders/PointRangeQueryBuilder.java
```diff
@@ -75,32 +76,33 @@ public Query getQuery(Element e) throws ParserException {
     String field = DOMUtils.getAttributeWithInheritanceOrFail(e, "fieldName");
     final String lowerTerm = DOMUtils.getAttribute(e, "lowerTerm", null);
     final String upperTerm = DOMUtils.getAttribute(e, "upperTerm", null);
+    String type = DOMUtils.getAttribute(e, "type", "int").toLowerCase(Locale.getDefault());
```
msfroh, line 79: Note that this is locale-specific, so for example, if the type is `INT` on a Turkish machine, it will be lowercased to `ınt` (the dotless i).

This should be lower-cased using `Locale.ROOT` to make it locale-insensitive.

sandeshkr419: Got it, thanks for pointing - I remember our offline discussion on this now. Let me fix it up in a follow-up.

Here: https://github.com/apache/lucene/pull/15076

sandeshkr419: https://github.com/apache/lucene/pull/15076

PR #15053, opened 2025-08-13
Cleanup DOMUtils

sandeshkr419:
### Description

Code cleanup of `DOMUtils` class

- Remove unused methods
- Remove unnecessary null checks
- Removed unnecessary boolean value boxing: `Boolean#valueOf` changed to `Boolean#parseBoolean`
- Rename: `deflt` -> `defaultValue`

### DIFF
--- a/lucene/queryparser/src/java/org/apache/lucene/queryparser/xml/DOMUtils.java
+++ b/lucene/queryparser/src/java/org/apache/lucene/queryparser/xml/DOMUtils.java
@@ -16,13 +16,8 @@
  */
 package org.apache.lucene.queryparser.xml;
 
-import java.io.Reader;
-import javax.xml.parsers.DocumentBuilder;
-import javax.xml.parsers.DocumentBuilderFactory;
-import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.Node;
-import org.xml.sax.InputSource;
 
 /** Helper methods for parsing XML */
 public class DOMUtils {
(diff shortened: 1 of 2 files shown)


lucene/queryparser/src/java/org/apache/lucene/queryparser/xml/DOMUtils.java
```diff
@@ -161,45 +133,12 @@ private static void getTextBuffer(Node e, StringBuilder sb) {
...
-
-    try {
-      db = dbf.newDocumentBuilder();
-    } catch (Exception se) {
-      throw new RuntimeException("Parser configuration error", se);
-    }
-
-    // Step 3: parse the input file
-    org.w3c.dom.Document doc = null;
-    try {
```
jainankitk, line 196: I am assuming this method is unused, right? Even then, we cannot remove public method from public class without deprecating, as it might break clients

lucene/queryparser/src/java/org/apache/lucene/queryparser/xml/DOMUtils.java
```diff
@@ -99,45 +89,27 @@ public static String getAttributeWithInheritance(Element element, String attribu
...
-
-  /* Convenience method to append a new child with text*/
-  public static Element insertChild(Element parent, String tagName, String text) {
-    Element child = parent.getOwnerDocument().createElement(tagName);
-    parent.appendChild(child);
-    if (text != null) {
-      child.appendChild(child.getOwnerDocument().createTextNode(text));
-    }
-    return child;
-  }
```
jainankitk, line 116: Same goes for these public methods

PR #15066, opened 2025-08-13
Reduce memory needed for multi-vector tests

vigyasharma:
Tests for late interaction field rescorer have seen some [build failures](https://jenkins.thetaphi.de/job/Lucene-main-Linux/57727/) due to OOM. Looks like they can sometimes exceed 500mb requirements when random tests pick higher values. Updated the tests to keep memory needs low.


jainankitk (approved):
Thanks @vigyasharma for addressing these build failures!

PR #14836, opened 2025-06-23
Sometimes use `FixedBitSet` when doing HNSW searches

benwtrent:
For smaller graphs, the overhead cost of a SparseFixedBitSet shows up in the performance metrics.

This adjusts the bitset creation logic to be more similar to how we utilize [Sparse]FixedBitSet elsewhere. 

The `expectedVisitedNodes` is both empirical and intuitive. This could possibly be refined given the number of connections within the graph, but I think this is "good enough" for now.

PR #15070, opened 2025-08-14
Use expectedVisitedNodes estimator for filtered vector search

akhilesh-k:
**Description**

This PR updates the decision logic in filtered vector search to leverage the new `expectedVisitedNodes(int k, int size)` heuristic introduced in #14836.

The existing flow for filtered vector search is discussed in this issue comment: https://github.com/apache/lucene/issues/14845#issue-3175422512

The previous heuristic `filterCost <= perLeafK` for triggering exact search could cause unnecessary approximate searches which may stop early due to `filterCost <= expectedVisitedNodes(perLeafK, size)` and then require a fallback exact search.

By using `expectedVisitedNodes`, which estimates the number of visited nodes based on both k and graph size, we can more accurately predict query cost and bypass unnecessary approximate searches.

### DIFF
--- a/lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/VectorScorerFloat32Benchmark.java
+++ b/lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/VectorScorerFloat32Benchmark.java
@@ -78,7 +78,7 @@ import org.openjdk.jmh.annotations.Warmup;
       "-XX:+AlwaysPreTouch",
       "--add-modules=jdk.incubator.vector"
     })
-/**
+/*
  * Benchmark to compare the performance of float32 vector scoring using the default and optimized
  * scorers. While there are benchmark methods for each of the similarities, it is often most useful
  * to compare equivalent subsets, e.g. .*dot.*
@@ -170,7 +170,6 @@ public class VectorScorerFloat32Benchmark {
     // type pollution on virtual calls, etc.
     float[] vec = randomVector(size, random);
     var opt = FlatVectorScorerUtil.getLucene99FlatVectorsScorer();
-    var scorer = opt.getRandomVectorScorer(DOT_PRODUCT, values.copy(), vec);
 
     for (int i = 0; i < 2; i++) {
       dotProductOptScorer();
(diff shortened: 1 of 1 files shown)


akhilesh-k:
Hi @jpountz would it be possible to review this PR please?

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -200,17 +202,17 @@ private TopDocs getLeafResults(
     if (scorer == null) {
       return NO_RESULTS;
     }
-
-    BitSet acceptDocs = createBitSet(scorer.iterator(), liveDocs, reader.maxDoc());
+    int documentSize = reader.maxDoc();
```
jpountz: Can you call it `maxDoc` instead?

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -200,17 +202,17 @@ private TopDocs getLeafResults(
...
     int perLeafTopK = perLeafTopKCalculation(k, leafProportion);
-
-    if (cost <= perLeafTopK) {
-      // If there are <= perLeafTopK possible matches, short-circuit and perform exact search, since
-      // HNSW must always visit at least perLeafTopK documents
+    int expectedVisitedNodesEstimate = expectedVisitedNodes(perLeafTopK, documentSize);
+    if (cost <= expectedVisitedNodesEstimate) {
+      // If there are <= expectedVisitedNodes possible matches, short-circuit and perform exact
+      // search, since
+      // HNSW must always visit at least expectedVisitedNodes documents
```
jpountz: Let's fix this comment to say that HNSW likely needs to visit this many documents, rather than guaranteeing that it needs to visit this many vectors?

jpountz (commented):
Thank you, the logic looks good to me.

My main concern is that `AbstractKnnVectorQuery`, which should be agnostic of the vector search index, now imports some HNSW-specific logic. We should be able to fix this when @shubhamvishu completes the work of pushing down the decision logic for exact vs. approximate to the codec (see discussion on #14963). I'd suggest to wait until this other work is complete before merging this PR.

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
akhilesh-k: Yes makes sense, have updated the PR to address this.

akhilesh-k:
Thank you for your review @jpountz. I will wait for the PR https://github.com/apache/lucene/pull/14963 to be merged and will fix up conflicts later.

PR #12027, opened 2022-12-21
Remove IOContext from Directory#openChecksumInput

zacharymorn:
### Description

Remove IOContext from `Directory#openChecksumInput` as suggested in https://github.com/apache/lucene/issues/11933

PR #15074, opened 2025-08-14
Use READONCE to read segment infos

ChrisHegarty:
This commit uses READONCE to read segment infos.

relates #15068

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -277,6 +277,8 @@ Bug Fixes
 * GITHUB#14922: Fix for IndexedDISI.Dense#intoBitSetWithinBlock to take into account the provided bitset size
   and avoid throwing IndexOutOfBoundsException (Panagiotis Bailis)
 
+* GITHUB#15074: Fix to use READONCE when reading segment infos (Chris Hegarty)
+
 Build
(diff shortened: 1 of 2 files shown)


uschindler:
Why is this problem only visible in 9.x? This fix is for main, but Mike said it is working in 10.x?

uschindler (approved):

ChrisHegarty:
> Why is this problem only visible in 9.x? This fix is for main, but Mike said it is working in 10.x?

I dunno yet, but we should be using READONCE for this anyway, since it can be more efficient. I'll continue to investigate why this is different on 10x.

ChrisHegarty:
In 10x `Lucene99SegmentInfoFormat:read` uses ` dir.openChecksumInput(fileName)`, which will do the right thing - use READONCE. It effectively ignores the passed in context.   This may not be the case for other segment info formats, so we should pass in READONCE. So this explains the difference that Mike is seeing.

mikemccand:
Thank you for the fast fix here!!

Yeah [this change (10.x, 11.x only)](https://github.com/apache/lucene/pull/12027) fixed `Directory.openChecksumInput` to always pass `IOContext.READONCE`.

PR #15075, opened 2025-08-14
Fix assertion in testSearchWithVisitedLimit

shubhamvishu:
### Description

Closes https://github.com/apache/lucene/issues/15057

### DIFF
--- a/lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseKnnVectorsFormatTestCase.java
+++ b/lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseKnnVectorsFormatTestCase.java
@@ -42,6 +42,7 @@ import org.apache.lucene.codecs.KnnVectorsFormat;
 import org.apache.lucene.codecs.KnnVectorsReader;
 import org.apache.lucene.codecs.KnnVectorsWriter;
 import org.apache.lucene.codecs.hnsw.HnswGraphProvider;
+import org.apache.lucene.codecs.lucene99.Lucene99HnswVectorsReader;
 import org.apache.lucene.codecs.perfield.PerFieldKnnVectorsFormat;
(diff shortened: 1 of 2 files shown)


lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseKnnVectorsFormatTestCase.java
```diff
@@ -1492,7 +1493,11 @@ public void testSearchWithVisitedLimit() throws Exception {
                   .searchNearestVectors(
                       fieldName, randomNormalizedVector(dimension), k, liveDocs, visitedLimit);
           assertEquals(TotalHits.Relation.GREATER_THAN_OR_EQUAL_TO, results.totalHits.relation());
-          assertEquals(visitedLimit, results.totalHits.value());
+          int size = Lucene99HnswVectorsReader.EXHAUSTIVE_BULK_SCORE_ORDS;
+          assertTrue(
+              visitedLimit == results.totalHits.value()
+                  || ((visitedLimit + size - 1) / size) * ((long) size)
```
msokolov, line 1499: the second one is always looser than the first one, so we don't need both? Can we simplify to

    assertEquals(((visitedLimit + size - 1) / size) * ((long) size), results.totalHits.value());

shubhamvishu: We haven't added the bulk scoring feature to the old codecs/readers, which causes the test to fail for others. This aims to address both those scenarios.

msokolov: ah, I see.

msokolov (approved):

jpountz:
I backported to branch_10x.

PR #15076, opened 2025-08-14
[bug] Fix Locale usage during lower-casing

sandeshkr419:
### Description

Fixes: https://github.com/apache/lucene/pull/15050#discussion_r2277714025
Thanks @msfroh for preventing a bug in future.

Lower-casing using `Locale.ROOT` to make it locale-insensitive instead of `Locale.getDefault()`.


msfroh:
Can you add a test case that covers the `type` attribute? 

It looks like none of the existing `PointRangeQuery*.xml` tests do anything with type.

msfroh (approved):

sandeshkr419:
@msfroh - Thanks for merging this. I'll get to adding the test coverage in a follow-up PR.

@jainankitk Do we need to backport this as well to 10.x - otherwise 10.3 might be with bug since you merged https://github.com/apache/lucene/pull/15050 to 10.x branch?

PR #15077, opened 2025-08-15
Add cleaner asserts to track unreleased RefCountedSharedArenas.

dweiss:
This is not supposed to be merged. Should help with #15068

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/RefCountedSharedArena.java
+++ b/lucene/core/src/java/org/apache/lucene/store/RefCountedSharedArena.java
@@ -18,6 +18,7 @@ package org.apache.lucene.store;
 
 import java.lang.foreign.Arena;
 import java.lang.foreign.MemorySegment;
+import java.lang.ref.Cleaner;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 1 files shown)


lucene/core/src/java/org/apache/lucene/store/RefCountedSharedArena.java
```diff
@@ -49,6 +51,17 @@ final class RefCountedSharedArena implements Arena {
...
   private static final int ACQUIRE_DECREMENT = REMAINING_UNIT - 1; // 0xffff
 
+  /** True iff JVM assertions are enabled for this class. */
+  private static final boolean ASSERTS_ENABLED;
+
+  /** Cleaner only created when assertions are enabled (avoids overhead otherwise). */
+  private static final Cleaner CLEANER;
+
+  static {
+    ASSERTS_ENABLED = RefCountedSharedArena.class.desiredAssertionStatus();
```
uschindler, line 61: I'd prefer not to use that function and use the assert keyword to trigger this:

```java
boolean enabled;
assert enabled = true;
ASSERTS_ENABLED = enabled;
```

PR #15080, opened 2025-08-15
Use DocValuesSkipper for count() in SortedNumericDocValuesRangeQuery

romseygeek:
If a field has a skipper enabled then we can shortcut range queries where the 
requested range is entirely outside of the skipper, or if the skipper is fully included 
within the requested range and there are no deleted documents or documents with 
no value in the field.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/document/SortedNumericDocValuesRangeQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/document/SortedNumericDocValuesRangeQuery.java
@@ -195,11 +195,10 @@ final class SortedNumericDocValuesRangeQuery extends Query {
           return 0;
         }
         if (skipper.docCount() == context.reader().maxDoc()
-            && context.reader().hasDeletions() == false
             && skipper.minValue() >= lowerValue
             && skipper.maxValue() <= upperValue) {
 
-          return context.reader().maxDoc();
+          return context.reader().numDocs();
         }
         return -1;
       }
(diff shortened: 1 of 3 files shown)


iverase (approved):
LGTM

jainankitk (approved):
Thanks @romseygeek, this is a nice improvement. I am wondering if the same logic can be applied to rewrite `SortedNumericDocValuesRangeQuery` into `MatchAllQuery` or `MatchNoDocsQuery` as well?

lucene/core/src/java/org/apache/lucene/document/SortedNumericDocValuesRangeQuery.java
```diff
@@ -184,6 +184,25 @@ public float matchCost() {
...
+        }
+        if (skipper.minValue() > upperValue || skipper.maxValue() < lowerValue) {
+          return 0;
+        }
+        if (skipper.docCount() == context.reader().maxDoc()
+            && context.reader().hasDeletions() == false
+            && skipper.minValue() >= lowerValue
+            && skipper.maxValue() <= upperValue) {
+
+          return context.reader().maxDoc();
```
jpountz: You could make it slightly more general by removing the check on `hasDeletions()` and returning `context.reader().numDocs()`.

jpountz (approved):

lucene/core/src/java/org/apache/lucene/document/SortedNumericDocValuesRangeQuery.java
romseygeek: Good shout, I'll update!

romseygeek:
> Thanks @romseygeek, this is a nice improvement. I am wondering if the same logic can be applied to rewrite SortedNumericDocValuesRangeQuery into MatchAllQuery or MatchNoDocsQuery as well?

Skippers are per-segment so it would be trickier to do this at the top level, unfortunately.

jpountz:
I think that @jainankitk 's point was not an either/or, just pointing out that we could also use the skipper to try to rewrite to MatchAll/MatchNone using similar logic, which has other benefits.

jpountz (approved):

romseygeek:
> I think that @jainankitk 's point was not an either/or, just pointing out that we could also use the skipper to try to rewrite to MatchAll/MatchNone using similar logic, which has other benefits.

Oh yes, I see what you mean.  It's a nice idea, I'll open a follow up.

PR #15039, opened 2025-08-08
Better vectorize score computations.

jpountz:
Existing auto-vectorization of scores is a bit fragile since it relies on `SimScorer#score` being inlined in the for loops where it is called. This is currently the case in nightly benchmarks, but may not be the case in the real world where more implementations of `SimScorer` may be used, in particular those from `FeatureField`.

Furthermore, existing auto-vectorization has some room for improvement as @gf2121 highlighted at
https://github.com/apache/lucene/pull/14679#discussion_r2093432298.

PR #15082, opened 2025-08-17
Make calls to BM25Scorer#score inlinable.

jpountz:
I ran experiments locally that suggest that some of the performance decrease from type pollution (https://github.com/mikemccand/luceneutil/pull/436) can be attributed to calls to `SimScorer#score` no longer being inlinable since they are polymorphic. This change helps `BM25Scorer` remain inlinable using similar tricks that we are applying for `Bits#get` and `ImpactsEnum#nextDoc`/`ImpactsEnum#advance`.

Hopefully changes such as #15039 will help improve performance with other similarities as well in the future.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/CombinedFieldQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/CombinedFieldQuery.java
@@ -289,8 +289,8 @@ public final class CombinedFieldQuery extends Query implements Accountable {
         CollectionStatistics pseudoCollectionStats = mergeCollectionStatistics(searcher);
         TermStatistics pseudoTermStatistics =
             new TermStatistics(new BytesRef("pseudo_term"), docFreq, Math.max(1, totalTermFreq));
-        this.simWeight =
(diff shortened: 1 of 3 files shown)


rmuir:
Why not just commit #15039 instead of this, which will also solve this problem and make it bimorphic?

jpountz:
I was hesitating which one to commit first. #15039 helps less in my benchmarks for now, likely because it only helps the first clause of conjunctive queries, because it doesn't help with computing impact scores, and because it only helps when scores are computed via the BulkScorer API instead of Scorer API (so e.g. AndHighOrMedMed doesn't see a speedup). I can start with #15039 plus follow-ups and see how much this PR still helps when I get stuck.

rmuir:
But this PR is also incomplete and ooesn't hook in the logic for things like SynonymQuery, PhraseQuery, which led to my confusion :)

jpountz:
I'm closing: The main queries that this change now helps are `AndMedOrHighHigh` and `AndHighOrMedMed`, which I don't think are important enough to warrant this complexity.

PR #15083, opened 2025-08-18
Rewrite SortedNumericDocValuesRangeQuery using DocValuesSkippers

romseygeek:
If DocValuesSkippers are enabled on a field, we can check at rewrite time
if a range query will match all or none of the documents in all segments,
and rewrite to a corresponding MatchAllDocsQuery or MatchNoDocsQuery.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
@@ -18,6 +18,7 @@ package org.apache.lucene.index;
 
 import java.io.IOException;
 import org.apache.lucene.search.DocIdSetIterator;
+import org.apache.lucene.search.IndexSearcher;
 
 /**
  * Skipper for {@link DocValues}.
@@ -121,5 +122,57 @@ public abstract class DocValuesSkipper {
       }
       advance(maxDocID + 1);
     }
+  }
+
+  /**
+   * Returns the minimum value for a field across all segments, or {@link Long#MIN_VALUE} if not available
+   * @param searcher  a searcher over the index
+   * @param field     the field to retrieve values for
+   */
+  public static long globalMinValue(IndexSearcher searcher, String field) throws IOException {
+    long minValue = Long.MAX_VALUE;
+    for (LeafReaderContext ctx : searcher.getLeafContexts()) {
+      DocValuesSkipper skipper = ctx.reader().getDocValuesSkipper(field);
+      if (skipper == null) {
+        minValue = Long.MIN_VALUE;
+      } else {
+        minValue = Math.min(minValue, skipper.minValue());
+      }
+    }
+    return minValue;
+  }
+
+  /**
+   * Returns the maximum value for a field across all segments, or {@link Long#MIN_VALUE} if not available
+   * @param searcher  a searcher over the index
+   * @param field     the field to retrieve values for
+   */
+  public static long globalMaxValue(IndexSearcher searcher, String field) throws IOException {
+    long maxValue = Long.MIN_VALUE;
+    for (LeafReaderContext ctx : searcher.getLeafContexts()) {
+      DocValuesSkipper skipper = ctx.reader().getDocValuesSkipper(field);
+      if (skipper == null) {
+        maxValue = Long.MAX_VALUE;
+      } else {
+        maxValue = Math.max(maxValue, skipper.maxValue());
+      }
+    }
+    return maxValue;
+  }
+
+  /**
+   * Returns the total skipper document count for a field across all segments
+   * @param searcher  a searcher over the index
+   * @param field     the field to retrieve values for
+   */
+  public static int globalDocCount(IndexSearcher searcher, String field) throws IOException {
+    int docCount = 0;
+    for (LeafReaderContext ctx : searcher.getLeafContexts()) {
+      DocValuesSkipper skipper = ctx.reader().getDocValuesSkipper(field);
+      if (skipper != null) {
+        docCount += skipper.docCount();
+      }
+    }
+    return docCount;
   }
 }
--- a/lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
+++ b/lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
@@ -125,9 +125,11 @@ public abstract class DocValuesSkipper {
   }
 
   /**
-   * Returns the minimum value for a field across all segments, or {@link Long#MIN_VALUE} if not available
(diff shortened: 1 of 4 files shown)


jpountz (commented):
It looks a bit complex to me with the state management, what about computing the global min/max/doc_count first like `PointRangeQuery`? It wouldn't exit early like your impl but if it's good for `PointRangeQuery`, it should be good here too?

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
```diff
@@ -122,4 +123,61 @@ public final void advance(long minValue, long maxValue) throws IOException {
...
+    for (LeafReaderContext ctx : searcher.getLeafContexts()) {
+      DocValuesSkipper skipper = ctx.reader().getDocValuesSkipper(field);
+      if (skipper == null) {
+        minValue = Long.MIN_VALUE;
+      } else {
+        minValue = Math.min(minValue, skipper.minValue());
+      }
+    }
+    return minValue;
+  }
```
jpountz, line 148: If one segment doesn't have the given field (ie. no field info for the given field), then this will return MIN_VALUE. Should it skip such segments instead? (And likewise for globalMaxValue?)

jpountz (commented):
Two minor comments, otherwise LGTM.

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
```diff
@@ -122,4 +123,61 @@ public final void advance(long minValue, long maxValue) throws IOException {
...
+  public static int globalDocCount(IndexSearcher searcher, String field) throws IOException {
+    int docCount = 0;
+    for (LeafReaderContext ctx : searcher.getLeafContexts()) {
+      DocValuesSkipper skipper = ctx.reader().getDocValuesSkipper(field);
+      if (skipper != null) {
+        docCount += skipper.docCount();
+      }
+    }
+    return docCount;
+  }
```
jpountz, line 188: IMO the contract would be a bit trappy if it returned -1 if the field exists in one segment (as per field infos) but has no skip index (e.g. `MultiReader` allows mixing segments that have a skip index on the field and segments that don't).

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
romseygeek: I think skipping could yield incorrect minimum values for the MultiReader case where some segments have skippers and some don't?

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
romseygeek: In general I'd expect this to be used to check that all documents in a reader are included in the skipper (ie, there aren't any documents with no value in this field).  If you have a MultiReader which includes segments with skippers and segments without, then this will return a value that is less that the top-level reader's `maxDoc()`, so you would know not to use the global values as a shortcut for an all-values-match check.

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
romseygeek: Reading this back I realise that 'skip' is an overloaded term here :). To clarify: ignoring values for segments that don't have a DocValuesSkipper enabled might mean that we incorrectly return a global minimum value.  Let's say we have a MultiReader over two segments, one with DVS enabled and values of [10-20], and one without with values of [0-5].  If we skip the second segment when calculating, we return a global minimum value of 10, and so a count() shortcut for the range [2-8] would assume that no documents match, when in fact some of the documents in the second segment do.

jpountz: I was thinking of ignoring segments that don't have the field in their field infos (meaning that no docs have a value for this field) and returning MIN_VALUE if the field exists in field infos but it has no skip index?

romseygeek: OIC - yes, that sounds like a good plan.  Will update.

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
jpountz: I agree it's fine for this use-case, I was more thinking of when it may start getting used for other purposes

jpountz (approved):

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
```diff
@@ -122,4 +123,67 @@ public final void advance(long minValue, long maxValue) throws IOException {
...
+   */
+  public static long globalMinValue(IndexSearcher searcher, String field) throws IOException {
+    long minValue = Long.MAX_VALUE;
+    for (LeafReaderContext ctx : searcher.getLeafContexts()) {
+      if (ctx.reader().getFieldInfos().fieldInfo(field) == null) {
+        continue; // no field values in this segment, so we can ignore it
+      }
+      DocValuesSkipper skipper = ctx.reader().getDocValuesSkipper(field);
+      if (skipper == null) {
+        minValue = Long.MIN_VALUE;
```
jainankitk, line 142: Thanks @romseygeek for this change. I missed this PR earlier, just noticed that we can exit early if skipper is null for one of the leaves. Opened small followup with this change - https://github.com/apache/lucene/pull/15127

lucene/core/src/java/org/apache/lucene/index/DocValuesSkipper.java
```diff
@@ -122,4 +123,67 @@ public final void advance(long minValue, long maxValue) throws IOException {
...
   }
+
+  /**
+   * Returns the minimum value for a field across all segments, or {@link Long#MIN_VALUE} if not
+   * available
+   *
+   * @param searcher a searcher over the index
+   * @param field the field to retrieve values for
+   */
+  public static long globalMinValue(IndexSearcher searcher, String field) throws IOException {
```
bugmakerrrrrr, line 134: hi @romseygeek , I'm wondering if we can change the parameter of the global* method from `IndexSearcher ` to `IndexReader`, similar to the stats method in `PointValues`. I think this would make it more convenient for upper-level applications to use.

romseygeek: Sounds reasonable - I'm not really sure why I didn't go with `IndexReader` in the first place 🤷 .  Do you want to open an PR?

PR #15085, opened 2025-08-18
Hunspell suggestions: Ensure candidate roots are not worse before updating

p3rmiashkin:
### Description

This PR suggest some performance improvements. CPU profiling showed that some time is taken by `PriorityQueue`'s `poll` / `add` operations.

Changes made:
* Ensure current wosrt root `isWorseThan` candidate before adding it to the queue

Before:
Screenshot 2025-08-18 at 7 15 35 PM

After:
Screenshot 2025-08-18 at 7 15 23 PM

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -65,6 +65,8 @@ Optimizations
 
 * GITHUB#15043: Vectorize L2 normalize function. (Ramakrishna Chilaka)
 
+* GITHUB#15085: Hunspell suggestions: Ensure candidate roots are not worse before updating. (Ilia Permiashkin)
+
 Bug Fixes
 ---------------------
 * GITHUB#14049: Randomize KNN codec params in RandomCodec. Fixes scalar quantization div-by-zero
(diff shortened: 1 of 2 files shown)


lucene/CHANGES.txt
```diff
@@ -65,6 +65,8 @@ Optimizations
 
 * GITHUB#15043: Vectorize L2 normalize function. (Ramakrishna Chilaka)
 
+* GITHUB#15085: Ensure candidate roots are not worse before updating. (Ilia Permiashkin)
```
donnerpeter: Add something like `Hunspell suggestions: ` for curious readers who don't know everything in Lucene?

donnerpeter:
Do you have any performance numbers before/after?

lucene/CHANGES.txt
donnerpeter: Mentioning Hunspell would also be helpful in the commit message and PR name

p3rmiashkin: Did force push to squash and update the commit message

p3rmiashkin:
> Do you have any performance numbers before/after?

Yes, I do have some numbers: `GeneratingSuggester#suggest` took 2440 ms before the changes (`poll` took 780ms, `add` took 130ms), and 1200 ms after the changes (`poll` took 50ms, `add ` took 10ms)

donnerpeter (approved):

PR #15089, opened 2025-08-18
Refactor GroupVIntUtil functional interface lambda which does not inline correctly in MemorySegmentIndexInput

### LINKED ISSUE #15079: Understand 2025/08/06 nightly benchy regression in KNN indexing
[Spinoff from [this comment](https://github.com/apache/lucene/pull/14932#issuecomment-3167938331)]

We recently merged a [nice opto to use GroupVarInt instead of VarInt for encoding each node's neighbors list in the HNSW (KNN) graph](https://github.com/apache/lucene/pull/14932).  Benchmarks showed improvements to search latency, which we expect since GroupVarInt decodes four ints at once and puts less branch prediction stress on the CPU.

But oddly nightly vector indexing benchmarks seemed to slow down, on 2025/08/06, and it's not yet obvious why -- it isn't one of the random "glitches" that that recover after a day or two (those are separately annoying!  [`beast3` benchy box is old now](https://blog.mikemccandless.com/2021/01/apache-lucene-performance-on-128-core.html), ~4.5 years, which equates to 135 years in humans years!  (1 computer year = 30 human years)  Maybe it is getting a bit flakey in its later years...).

I'll try to dig some more ... but if I get nowhere (likely!), I think we should temporarily revert GroupVarInt and see if benchy recovers?  Especially because this might become a blocker for 10.3 release...

uschindler:
The original code doies not optimize well due to type pollution in combination with a capturing lambda for GroupVInts decoding.

The new code uses a verhandle which is adapted for the call site. The caller passes backing storage (called referent as untyped parameter) and the offset (long) as coordinates.

For byte buffers the offset coordinate is casted top int like in old code 

This (hopefully) fixes #15079

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -166,6 +166,9 @@ Improvements
 
 * GITHUB#14816: Expose search strategy in KNN query
 
+* GITHUB#15089, GITHUB#15079: Refactor GroupVIntUtil functional interface lambda
+  which does not inline correctly in MemorySegmentIndexInput (Uwe Schindler, Robert Muir)
+
 Optimizations
 ---------------------
 * GITHUB#14932: Switched to GroupVarInt Encoding for HNSW Graph edges, added backwards compatibility (Akira Lonske)
--- a/lucene/core/src/java/org/apache/lucene/store/BufferedIndexInput.java
+++ b/lucene/core/src/java/org/apache/lucene/store/BufferedIndexInput.java
@@ -154,7 +154,13 @@ public abstract class BufferedIndexInput extends IndexInput implements RandomAcc
   public void readGroupVInt(int[] dst, int offset) throws IOException {
     final int len =
         GroupVIntUtil.readGroupVInt(
-            this, buffer.remaining(), p -> buffer.getInt((int) p), buffer.position(), dst, offset);
+            this,
+            buffer.remaining(),
+            ByteBuffersDataInput.VH_BUFFER_GET_INT,
+            buffer,
+            buffer.position(),
+            dst,
+            offset);
     if (len > 0) {
       buffer.position(buffer.position() + len);
     }
--- a/lucene/core/src/java/org/apache/lucene/store/BufferedIndexInput.java
+++ b/lucene/core/src/java/org/apache/lucene/store/BufferedIndexInput.java
@@ -156,7 +156,7 @@ public abstract class BufferedIndexInput extends IndexInput implements RandomAcc
         GroupVIntUtil.readGroupVInt(
             this,
             buffer.remaining(),
-            ByteBuffersDataInput.VH_BUFFER_GET_INT,
+            GroupVIntUtil.VH_BUFFER_GET_INT,
             buffer,
             buffer.position(),
             dst,
--- a/lucene/core/src/java/org/apache/lucene/store/ByteBuffersDataInput.java
+++ b/lucene/core/src/java/org/apache/lucene/store/ByteBuffersDataInput.java
@@ -48,9 +48,8 @@ public final class ByteBuffersDataInput extends DataInput
       MethodHandles.filterCoordinates(
           MethodHandles.byteBufferViewVarHandle(int[].class, ByteOrder.LITTLE_ENDIAN),
           0,
-          MethodHandles.explicitCastArguments(
-              MethodHandles.identity(Object.class),
-              MethodType.methodType(ByteBuffer.class, Object.class)),
+          MethodHandles.identity(Object.class)
+              .asType(MethodType.methodType(ByteBuffer.class, Object.class)),
           MethodHandles.explicitCastArguments(
               MethodHandles.identity(long.class), MethodType.methodType(int.class, long.class)));
 
--- a/lucene/core/src/java/org/apache/lucene/store/ByteBuffersDataInput.java
+++ b/lucene/core/src/java/org/apache/lucene/store/ByteBuffersDataInput.java
@@ -18,9 +18,6 @@ package org.apache.lucene.store;
 
 import java.io.EOFException;
 import java.io.IOException;
-import java.lang.invoke.MethodHandles;
-import java.lang.invoke.MethodType;
-import java.lang.invoke.VarHandle;
 import java.nio.BufferUnderflowException;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
@@ -41,17 +38,6 @@ import org.apache.lucene.util.RamUsageEstimator;
  */
 public final class ByteBuffersDataInput extends DataInput
     implements Accountable, RandomAccessInput {
-
-  // also used by BufferedIndexInput, this needs to cast the long parameter to the ByteBuffer's int
-  // coordinate
-  static final VarHandle VH_BUFFER_GET_INT =
-      MethodHandles.filterCoordinates(
-          MethodHandles.byteBufferViewVarHandle(int[].class, ByteOrder.LITTLE_ENDIAN),
-          0,
-          MethodHandles.identity(Object.class)
-              .asType(MethodType.methodType(ByteBuffer.class, Object.class)),
-          MethodHandles.explicitCastArguments(
-              MethodHandles.identity(long.class), MethodType.methodType(int.class, long.class)));
 
   private final ByteBuffer[] blocks;
   private final FloatBuffer[] floatBuffers;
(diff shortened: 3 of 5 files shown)


uschindler:
I need to add documentation. I also noticed that one of the methods GroupVIntUtil is dead code. Only the one that decodes to int[] is actually used in it's optimized form. The other ones are only used by older codecs.

rmuir (approved):
No regressions in the microbenchmark. 

main:
```
GroupVIntBenchmark.benchByteArrayDataInput_readGroupVInt                 64  thrpt    5  8.434 ± 0.022  ops/us
GroupVIntBenchmark.benchByteArrayDataInput_readVInt                      64  thrpt    5  4.851 ± 0.011  ops/us
GroupVIntBenchmark.benchByteBuffersIndexInput_readGroupVInt              64  thrpt    5  1.342 ± 0.002  ops/us
GroupVIntBenchmark.benchByteBuffersIndexInput_readGroupVIntBaseline      64  thrpt    5  1.297 ± 0.001  ops/us
GroupVIntBenchmark.benchMMapDirectoryInputs_readGroupVInt                64  thrpt    5  4.077 ± 0.009  ops/us
GroupVIntBenchmark.benchMMapDirectoryInputs_readGroupVIntBaseline        64  thrpt    5  4.076 ± 0.008  ops/us
GroupVIntBenchmark.benchMMapDirectoryInputs_readVInt                     64  thrpt    5  5.091 ± 0.011  ops/us
GroupVIntBenchmark.benchNIOFSDirectoryInputs_readGroupVInt               64  thrpt    5  3.983 ± 0.007  ops/us
GroupVIntBenchmark.benchNIOFSDirectoryInputs_readGroupVIntBaseline       64  thrpt    5  3.982 ± 0.007  ops/us
GroupVIntBenchmark.bench_writeGroupVInt                                  64  thrpt    5  2.385 ± 0.006  ops/us
```

patch:
```
GroupVIntBenchmark.benchByteArrayDataInput_readGroupVInt                 64  thrpt    5  8.442 ± 0.012  ops/us
GroupVIntBenchmark.benchByteArrayDataInput_readVInt                      64  thrpt    5  4.851 ± 0.007  ops/us
GroupVIntBenchmark.benchByteBuffersIndexInput_readGroupVInt              64  thrpt    5  1.300 ± 0.003  ops/us
GroupVIntBenchmark.benchByteBuffersIndexInput_readGroupVIntBaseline      64  thrpt    5  1.301 ± 0.002  ops/us
GroupVIntBenchmark.benchMMapDirectoryInputs_readGroupVInt                64  thrpt    5  4.078 ± 0.004  ops/us
GroupVIntBenchmark.benchMMapDirectoryInputs_readGroupVIntBaseline        64  thrpt    5  4.064 ± 0.008  ops/us
GroupVIntBenchmark.benchMMapDirectoryInputs_readVInt                     64  thrpt    5  5.088 ± 0.006  ops/us
GroupVIntBenchmark.benchNIOFSDirectoryInputs_readGroupVInt               64  thrpt    5  3.971 ± 0.006  ops/us
GroupVIntBenchmark.benchNIOFSDirectoryInputs_readGroupVIntBaseline       64  thrpt    5  4.125 ± 0.013  ops/us
GroupVIntBenchmark.bench_writeGroupVInt                                  64  thrpt    5  2.378 ± 0.002  ops/us
```
I had to re-run `main` branch multiple times because I had varying stability in the mmap case (only 1 fork is used). 

```
Score   Error   Units
3.092 ± 3.404  ops/us
```

This PR did not show any stability issues.

uschindler:
Thanks @rmuir for benchmarking in parallel to my development. I will cleanup the PR a bit.

The instability of the original code in the MMap case is caused by the broken inlining and/or escape analysis of the captured lambda.

The main problem of the code is that the lambda is captured (it refers to a local variable in the implementation) and this breaks performance if escape analysis does not work as it should. This also explains why @mikemccand has seen high GC activity.

uschindler:
Final version now ready. I will merge this to main branch soon, so Mike can do benchmarking.

Unfortunately the code does not yet work with Java 21, as Java 22 introduces `MethodHandles#filterCoordinates()`. I will try to find a solutions for Java 21. In worst case I need to fall back to MethodHandles in 10.x --- DAMN!

uschindler:
I will backport this using a reflection hack if this solves the issue seen in #15079.

The problem is that filterCoordinates is preview API on java 21 as it was introduced for Panama FFI and MemorySegments

uschindler:
I have a backport ready and will open a PR when it confirms to work. The backport is two fold:

- The ByteBuffer varhandle is generated refelctively
- The MemorySegment one is in the Java 21 Sourceset, so it can use the new Method. I just had to extract the API of MethodHandles class, too.

So basically it is still minimal-invasive in 10.x.

dweiss (commented):
LGTM. It's a bit scary how deep into the jvm this has to go in order to achieve good performance.

uschindler:
> I have a backport ready and will open a PR when it confirms to work. The backport is two fold:
> 
> * The ByteBuffer varhandle is generated refelctively
> * The MemorySegment one is in the Java 21 Sourceset, so it can use the new Method. I just had to extract the API of MethodHandles class, too.
> 
> So basically it is still minimal-invasive in 10.x.

Small correction: The backuport for the MemorySegemnt part is more complicated, because retrieving the VarHandle in the preview API is different that in the released Java 22+ API. I havea solution, will post a PR for backporting later.

But let's see how it works for main (if it worth backporting).

The issue is described in this helper method:

```java
  /**
   * Hack: The preview API changed after release in Java 22:
   * <ul>
   * <li>In Java 21 the preview API requires {@code MethodHandles#memorySegmentViewVarHandle(layout)}.
   * <li>In the released API (Java 22) it requires to use {@code layout.varHandle()}, which includes the offset coordinate.
   * </ul>
   * The new API existed before, but with different semantics, so we lookup the first one and
   * if the method does not exist (we can compile against our stubs and see it, but at runtime
   * it may have disappeared), we use the new API.
   */
  private static VarHandle getLayoutVarHandle(ValueLayout layout) {
    try {
      return MethodHandles.memorySegmentViewVarHandle(layout);
    } catch (@SuppressWarnings("unused") NoSuchMethodError e) {
      return layout.varHandle();
    }
  }
```

uschindler:
Here is the backport PR: https://github.com/apache/lucene/pull/15093

jpountz:
Since there are only two directories that optimize group-varint decoding, I wonder if we should duplicate the code instead of using lambdas or method handles.

uschindler:
> Since there are only two directories that optimize group-varint decoding, I wonder if we should duplicate the code instead of using lambdas or method handles.

it is three! And actually no MethodHandles involved, it's all VarHandles. MemorySegment internally also only does VarHandles.

uschindler:
I reverted that one in https://github.com/apache/lucene/commit/4ad598191aed8243a9fc21cb269c929b64014ce7.

uschindler:
Maybe we can autogenerate the code with a template for the three implemttations? @jpountz

jpountz:
I'm not even sure we need to script it, it's very little code?

uschindler:
Actually there are multiple problems: We need to fix this also in 10.x branch. But as the stupid method (which I would never have been put into IndexInput/DataInput) may be implemented by subclasses (I know many who implemented their own MemorySegmentIndexInput and consorts) where update might break, so we should be a bit careful. I'd like to remove the GroupVInt stuff completely from IndexInput and IndexOutput. I was already against the whole lambda and the implementation in general 2 years ago. The funny thing: The improvement by the 

In short: If you ask me: Let's remeove the "improved" method. It is a bit faster in microbenchmarks but the maintenance cost is too expensive and the risk of failing in benchmarks is too hight. Actually Mike's benchmark sees no difference at all because the "microbenchmark" is so specialized and due to callsite pollution it was never working correctly.

I Im still against adding the burden to downstream users to implement a customly optimized readGroupVInt method in their own directory implementation that have never shouwn any use in real benchmarks. This few percent of improvement by calling readGroupVInt in a mincrobenchmark in isolation is a no-go for good code-quality. And that code in those three IndexInputs that implement our specialcase is a mess and I am against it.

Let's remove the mmap/nio/bytebuffers directory optimization, commit it and let Mike benchmark it. If we see no difference larger than 5%, lets kill it forever!

uschindler:
@rmuir - do you have an opinion, too?

uschindler:
I have an idea how to solve the whole thing and remove all spezializations from IndexInput/DataInput. IMHO the GroupVInt code should solely be in GroupVIntUtil!

Basically, the problem of the optimization is that it needs absolute access because it needs t go forward backwards during decoding. So Actually it need RandomAccessInput.

Let's imply rewrite the code that it does instanceof RandomAccessInput (it shoudl not create a new one!!!!!) and if thats the case, just user the readInt(long pos)  methods.

If an IndexInput does not impelment RandomAcccess it would use the unoptimized code. Basically this would automatically improve MMap and ByteBuffers dircetory (the only important IndexInputs).

NIOFSDir's IndexInput does not implement RandomAccessInput, so it won't get an optimization, but that ok - i think? @rmuir @jpountz what do you think. I can give it a try!

uschindler:
Here is the PR solely using RandomAccessInput without any specialization in subclasses: #15116

PR #15093, opened 2025-08-19
Backport of #15089: Refactor GroupVIntUtil functional interface lambda which does not inline correctly in MemorySegmentIndexInput

### LINKED ISSUE #15079: Understand 2025/08/06 nightly benchy regression in KNN indexing

uschindler:
This is a backport of #15089: It is more complicated than expected:

- The method MethodHandles#filterCoordinates(...) method is preview in Java 21 as it is related to FFI and Panama. As it is called exactly one time I rewrote the code a bit to reflectively call it so the preview flag does not matter. Moving the whole code to the Java 21 part is not nice and requires additional lookups (also refelctive), so the cleast solution for calling that method is to just call it reflectively using a MathodHandle.
- retrieving the VarHandle to access `MemorySegments` in the preview API (Java 21) is different than in the released Java 22+ API
  - In Java 21 the preview API requires `MethodHandles#memorySegmentViewVarHandle(layout)`.
  - In the released API (Java 22) it requires to use `layout.varHandle()`, which includes the offset coordinate.
The new API existed before, but with different semantics, so we lookup the first one and if the method does not exist (we can compile against our stubs and see it, but at runtime it may have disappeared), we use the new API.

This fixes #15079 in 10.x branch.

I have tested the code with Java 21 (default), Java 22 and Java 24.


lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
```diff
@@ -33,6 +39,46 @@ public final class GroupVIntUtil {
...
+   * VarHandle} and {@code ByteBuffer} storage parameter.
+   *
+   * @see #readGroupVInt(DataInput, long, VarHandle, Object, long, int[], int)
+   * @see #readGroupVInt(DataInput, long, VarHandle, Object, long, long[], int)
+   */
+  public static final VarHandle VH_BUFFER_GET_INT = getBufferVarHandle();
+
+  /**
+   * In Java 21 the {@code MethodHandles#filterArguments(...)} method is preview only, so we invoke
+   * it reflective. For this code it does not make sense to move the whole code to the Java 21
```
dweiss:
```suggestion
   * it reflectively. For this code it does not make sense to move the whole code to the Java 21
```

dweiss (approved):

uschindler:
Thanks for the review! I still wait for Mike's benchmarks to succeed!

uschindler:
Checks on main branch have shown that this does now work well.

Issue #15095, opened 2025-08-19
Try to diagnose stalled gradle builds on policeman jenkins


dweiss:
### Description

There have been quite a few builds that oddly just... hung:
```
https://jenkins.thetaphi.de/job/Lucene-main-Linux/57884/
```

<img width="1838" height="976" alt="Image" src="https://github.com/user-attachments/assets/310c3f06-b4cc-4ab3-b7cb-b5c1359f6da7" />

I've no idea if it's got something to do with gradle or jenkins but I'll try to add some code on main to dump gradle jvm internals on inactivity exceeding a few minutes. Maybe we can figure out what's happening.

dweiss:
@uschindler - you may want to take a look at the policeman jenkins machine, independently. I don't know how these builds/vms are organized but something is definitely not right.

uschindler:
On my local windows build it also hung several times in one of the tests. Have to look it up. The test stopped and never ended, was always same class.

With normal console you see which test is running, but on Jenkins you don't see at all which module and which test...

It was always same test in Lucene core, will go through my local logs.

uschindler:
It is also happening on all vms. Mac, windows, plain Linux. So it's unrelated to virtualization.

I am 100% sure it's this hanging test.

uschindler:
One addition: Locally I was testing 10.x branch and that test stalled, too. On Policeman Jenkins all operating systems and 10.x and main branch have stalled builds.

I was very busy with testing the VarHandle stuff today so I just pressed Ctrl+C to kill the builds .... and later closed the terminal window. I will try to reproduce. At least in termin you see which test was hanging, as I have 5 lines showing the actually running tets and a 6th one with the total executed tests. When it hangs only one line is actively showing a running test (the last JVM which executed tests and the total number).

I will try to reproduce tomorrow, but I think easiest is to run `gradlew :lucene:core:test` in a row - main branch or 10.x does not matter.

uschindler:
P.S.: If you see a "hanging build", just ping me on Google Chat, I can dump stacks from all JVMs.

dweiss:
I do think it's the hanging build, yes. But it's weird that sometimes it does stop on its own (timeout on the suite) and other times it's just hanging. That gradle no-progress dumping routine works (https://jenkins.thetaphi.de/job/Lucene-main-Linux/57891/consoleText) - 

[...]

PR #15096, opened 2025-08-19
More gradle convention plugins converted to Java

dweiss:
This also piggybacks a plugin that should allow us to diagnose #15095 and enables it for builds
on main. Let's see what happens.


build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/hacks/DumpGradleStateOnStalledBuildsPlugin.java
```diff
@@ -0,0 +1,282 @@
...
+        jvmDiags = "Failed to collect jvm diagnostics: " + t.getMessage();
+      }
+      LOGGER.warn(jvmDiags);
+    }
+
+    /** Collects heap stats and a full thread dump into a single String. */
+    public static String collectJvmDiagnostics() {
+      StringBuilder sb = new StringBuilder(32_768);
+
+      sb.append("====== GRADLE JVM DIAGNOSTICS ======\n");
```
dweiss, line 155: I've generated this method using AI. The rest is written by hand.

dweiss:
I'll allow myself to commit this without waiting - maybe something will show up through the night.

dweiss:
image

Confirmed that gradle stall diagnostics code works.

PR #15098, opened 2025-08-20
Fix TestMultiIndexMergeScheduler hanging

### LINKED ISSUE #15091: TestMultiIndexMergeScheduler hanging occasionally
### Description

I have seen this new test (TestMultiIndexMergeScheduler) hang occasionally both on my local machine and on the CI. Here is a recent example:

https://jenkins.thetaphi.de/job/Lucene-nightly-main-Linux/523/consoleText

I didn't look deep but something is definitely racey there - the seed that was hanging for me didn't reproduce.

#15015 

_No response_

dweiss:
The test was decrementing the doc count before it was checking for thread termination condition (==0). The document assignment is random and it could happen certain threads started with 0 documents to add.

Separately from the fix, the test itself is fishy to me. I didn't look MultiIndexMergeScheduler but the test uses concurrent queues, while all worker threads are under one big monitor lock... Makes to sense to me. Either it's a remnant of some refactoring or the test code should be revisited and cleaned up.

Fixes #15091

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/index/TestMultiIndexMergeScheduler.java
+++ b/lucene/core/src/test/org/apache/lucene/index/TestMultiIndexMergeScheduler.java
@@ -16,10 +16,12 @@
  */
 package org.apache.lucene.index;
 
+import com.carrotsearch.randomizedtesting.annotations.Timeout;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


dweiss:
Ping, @corecursion

dweiss:
Will commit this and backport to 10x.

uschindler:
The test was added Aug 8, this goes in line with the hangs on Jenkins. There were no changes on the test, it is completely new. The original PR is that one: https://github.com/apache/lucene/pull/15015

I think we should really review the code. It is good that you added a timeout to the test for safety.

dweiss:
Yes, it's new code. Like I said - the test... it's a bit odd-looking. A second pair of eyes on that original patch would be probably good.

corecursion:
Thanks for looking at this.

PR #15099, opened 2025-08-20
Fix the remaining missing SuppressWarnings("preview") to silence IDEs like Eclipse. Also remove redundant interface.

uschindler:
This PR has no impact on the Gradle build of the 10.x branch, but silences wanrings emitted by Eclipse or other IDEs due to the fact that those does not know about our apijars and see the preview flag.

2 warnings are due to javadocs in the main branch (I could have replaced them by `{@code ...}` but referring to the preview APIs in Javadocs is good for the reader, so I had to add the warning. Not sure why javadoc does not complain, but ECJ in Eclipse does.

It also fixes a redundant interface declaration that was removed in main branch already.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/store/FSDirectory.java
+++ b/lucene/core/src/java/org/apache/lucene/store/FSDirectory.java
@@ -86,6 +86,7 @@ import org.apache.lucene.util.IOUtils;
  *
  * @see Directory
  */
+@SuppressWarnings("preview") // javadocs only
 public abstract class FSDirectory extends BaseDirectory {
 
   protected final Path directory; // The underlying filesystem directory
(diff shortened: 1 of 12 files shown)


uschindler:
The main issue why we do not see the issues in 10.x branch is beacuse we do not run ECJ on the MR-JAR sourcesets. The normal Javadoc compiler does not cmplain.

uschindler:
P.S.: With this applied my Eclipse IDE is warning-free in both 10.x and main branch.

rmuir:
> P.S.: With this applied my Eclipse IDE is warning-free in both 10.x and main branch.

@uschindler How do you manage fix/avoid the 1 bogus warning around IndexReaderContext/LeafReaderContext? It's the one error that, when you go visit it, it then disappears.

uschindler:
> > P.S.: With this applied my Eclipse IDE is warning-free in both 10.x and main branch.
> 
> @uschindler How do you manage fix/avoid the 1 bogus warning around IndexReaderContext/LeafReaderContext? It's the one error that, when you go visit it, it then disappears.

That's an error in the Eclipse IDE therefore my comment above is correct: All warnings gone. 🤓

This one is a Heisenbug. Sometimes disappears when you open file, sometimes does not come up at all. Depends also on Eclipse version. Nowadays it's sticky.

Will look into it again. Maybe some setting to avoid it or tun it into warning so we can suppress.

rmuir:
glad it isn't just me. The warning itself makes no sense either :)

uschindler:
I opened a bug report about this Heisenbug: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4331

PR #15102, opened 2025-08-21
Replacing unused variable suppress warning with _

jainankitk:
### Description

I noticed quite a few occurrences of `@SuppressWarnings("unused")` which can be replaced with unnamed variable for 11.0. Will get this done across few PRs to not make any individual PR too big.

PR #15109, opened 2025-08-21
Replacing unused variable suppress warning with _ (Part 3/5)

jainankitk:
### Description
I noticed quite a few occurrences of @SuppressWarnings("unused") which can be replaced with unnamed variable for 11.0. Getting this done across few PRs to not make any individual PR too big. Related to https://github.com/apache/lucene/pull/15102, https://github.com/apache/lucene/pull/15105

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/internal/hppc/TestLongArrayList.java
+++ b/lucene/core/src/test/org/apache/lucene/internal/hppc/TestLongArrayList.java
@@ -279,7 +279,7 @@ public class TestLongArrayList extends LuceneTestCase {
 
     count = 0;
     list.resize(0);
-    for (@SuppressWarnings("unused") LongCursor cursor : list) {
+    for (LongCursor _ : list) {
       count++;
     }
     assertEquals(0, count);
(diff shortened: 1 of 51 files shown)


lucene/core/src/test/org/apache/lucene/internal/hppc/TestLongArrayList.java
```diff
@@ -279,7 +279,7 @@ public void testIterable() {
 
     count = 0;
     list.resize(0);
-    for (@SuppressWarnings("unused") LongCursor cursor : list) {
+    for (LongCursor _ : list) {
```
uschindler, line 282: this is funny, but the test obviously only wants to count here to comapre the result of counting the iterable and compare it with real size.

As we check for "0 length" here, we could replace the whole loop with a simple `assertFalse(list.iterator().hasNext()`.

But let's keep this for now.

uschindler (approved):
Most of it is pure mechanical code. Go for it!

The funny loop (everything else is catch blocks) is a separate discussion.

lucene/core/src/test/org/apache/lucene/internal/hppc/TestLongArrayList.java
jainankitk: Yeah, I have also been reviewing these changes and noticed the funny loop a few times. But, just changed to unnamed variable for consistency as part of this PR

PR #15112, opened 2025-08-22
Fix infinite loop when RefCountedSharedArena's underlying Arena#close fails due to concurrent usage of segments

### LINKED ISSUE #15106: RefCountedShatredArena is causing infinite (spin-)loop on close of IndexInput/IndexReader/... due to exception handling
### Description

One user reported and I was able to explain this behaviour: https://github.com/apache/lucene/issues/15068#issuecomment-3210953067

Basically the RefCountedSharedArena throws an `IllegalStateException` when the refcounting is wrong. Tis should only happen when something is wrong, like you close an arena too often. But due to the official API of Arena, throwing IllegalStateException is the worst thing to do, because the documentation exactly specifies when this may happen:

> Throws:
> - [IllegalStateException](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/lang/IllegalStateException.html) - if the arena has already been closed
> - [IllegalStateException](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/lang/IllegalStateException.html) - if a segment associated with this arena is being accessed concurrently, e.g. by a downcall method handle .... _addition by Uwe:_ it is enough if any code actually invokes a method on a memory segment in that shared arena while another thread is calling close!

Throwing IllegalStateException is perfectly fine theoretically, if and only if the underlying `Scope` is already closed. But we don't check this at all and just throw the exception if the close method was called to often.

If now the second documented `IllegalStateException` is thrown by the Arena we delegate to, the refcount was already decremented.

Now the "official" way to handle the second case is this spinloop, which is implemented in `MemorySegmentIndexInput.close()`:

https://github.com/apache/lucene/blob/cb16075c5fd5c4bbcb0fab561e185fe2e3f8e939/lucene/core/src/java/org/apache/lucene/store/MemorySegmentIndexInput.java#L716-L724

This spinloop checks if the arena's scope is still open and if that is still the case it repeats the close call on `IllegalStateException` _until_ the arena's scope returns closed. This code was negotiated with the OpenJDK list and we also have a multithreaded test for it. Of course this test does not use RefCountedArenas as it uses a test file with a name which is not grouped.

Normal case:
- MemorySegmentIndexInput code calls close for first time (inner arenas scope still open). If the arena is still open, all is fine and refcount is decremented and inner arena is not touched. All fine

What now happens if the inner Arena's close throws that `IllegalStateException`: 
- MemorySegmentIndexInput calls close on the last open file in that shared group.
- The refcount is lowered, no exception thrown
- **For bad luck, some other thread acceses an IndexInput**
- The inner Arena's close throws IllegalStateException in the final cleanup
- MemorySegmentIndexInput (above code) catches exception and enters spinloop
- MemorySegmentIndexInput repeats close
- The refcount is now already 0 and therefor the RefCountedSharedArena's code throws IllegalStateException (while inner Arena's close is no longer called at all).
- MemorySegmentIndexInput (above code) catches exception and enters spinloop
- and this repeats forever, the inner arena is never closed

Basically to fix this, the following should be done:
- The code here should be removed: https://github.com/apache/lucene/blob/b1c1770e8052a74fd30e3520e4fd662e2724bd1c/lucene/core/src/java/org/apache/lucene/store/RefCountedSharedArena.java#L106-L108
- Instead if the refcount reaches 0 it should stay 0 (or go negative, no matter).
- The code should run the cleanup again and therefore call inner Arena's close again. If this arena throws IllegalStateException it can be handled by the caller.

So basically I'd remove the two parts of the refocunting and change the code to trivially do:
- decrement refcount
- if refcount>0, do nothing
- otherwise call inner Arena's close

If for some reason the inner scope is already closed the Arena#close() will throw IllegalState anyways, we do not need to duplicate the refcounting!

Severity: The bug is severe, if and only if somebody closes an IndexInput/IndexReader while it is used in another thread. This may lead to a spinloop if the concurrent acces is activcely accessing a segment..

SearcherMananger and normal Lucene use should not be affected, but the bug may still causehalting  the JVM due to a infinite loop when something goes wrong on close!

@ChrisHegarty 

Affected versions: Lucene 9.12, 10.x, main branch

uschindler:
This fixes #15106.

This basically only removes the check when an arena's refcount is already closed. The problem is if a previous close on the inner arena has failed due to concurrent access to the scoped memory segments, the refcount is already in state `CLOSED`, so another call from the "retry" spinloop in `MemorySegmentIndexInput` will hit our extra check and throws `IllegalStateException` again, leading to an endless spinloop.

The extra check is not needed, because if the inner Arena was already closed, we do not need to check this before, the inner Arena's close withh throw correct exception anyways.

This affects all users which have incorrect code, if your code is sane, you won't be affected by the spinloop bug, so it is not so serious like it looks.

@ChrisHegarty I did minimal changes here, but maybe we can no simplify the whole refcounting even more? Do we still need the upper and lower 16 bits of the atomic state?

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -302,6 +302,9 @@ Bug Fixes
   resulted in query branches being wrongly ignored if the QueryIndex was
   built by a different JVM. (Luke Kot-Zaniewski)
 
+* GITHUB#15106: Fix infinite loop when RefCountedSharedArena's underlying
+  Arena#close fails due to concurrent usage of segments.  (Uwe Schindler)
+
 Build
 ---------------------
 * Upgrade forbiddenapis to version 3.9.  (Uwe Schindler)
--- a/lucene/core/src/java/org/apache/lucene/store/RefCountedSharedArena.java
+++ b/lucene/core/src/java/org/apache/lucene/store/RefCountedSharedArena.java
@@ -50,7 +50,11 @@ final class RefCountedSharedArena implements Arena {
   private static final int ACQUIRE_DECREMENT = REMAINING_UNIT - 1; // 0xffff
 
   private final String segmentName;
+
+  // onClose could be called multiple times when the inner Arena's close fails with
+  // IllegalStateException
   private final Runnable onClose;
+
   private final Arena arena;
 
   // high 16 bits contain the total remaining acquires; monotonically decreasing
@@ -99,14 +103,10 @@ final class RefCountedSharedArena implements Arena {
 
   /** Decrements the ref count. */
   void release() {
-    int value;
     while (true) {
-      value = state.get();
+      final int value = state.get();
       final int count = value & 0xFFFF;
-      if (count == 0) {
-        throw new IllegalStateException(value == CLOSED ? "closed" : "nothing to release");
-      }
-      final int newValue = count == 1 ? CLOSED : value - 1;
+      final int newValue = count <= 1 ? CLOSED : value - 1;
       if (this.state.compareAndSet(value, newValue)) {
         if (newValue == CLOSED) {
           onClose.run();


dweiss (approved):
LGTM, although I also feel like the double counter is somehow superfluous here.

ChrisHegarty (approved):
LGTM. Thanks @uschindler.

ayinresh:
Are there any plans to backport this PR considering the severity of the bug?

uschindler:
Hi @ayinresh,
the fix was already backported to Lucene 10.3 which will come out soon. I don't think we'll release a 10.2.x bugfix variant, as we generally would only do this for major branches. If (and only if) we release another 9.12.x we can for sure include this bugfix, as it is serious for affected people (hangs completely with spinning CPU).

About the "severity" of the bug: Yes, indeed it is severe, but only if your code is already broken. In Lucene 9 with older Java versions before 19 broken code using IndexReaders while they are closed would have been crushed the whole JVM, as we forcefully unmap the direct ByteBuffers - leading to a SIGSEGV. I think this is also the reason why we see only few people hit this bug.

In one year this was the only encounter of this bug reported. It is also very seldom that you actually hit this bug: You need to close a thread at exactly the same time while an access to a mapped MemorySegment is done. In most cases you will succeed to close the Arena and later code will hit AlreadyClosedException. The bug is also hard to reproduce, but without the spinloop you can make one of the MMapDirectory tests fail (that was written exactly for that). If you watch closer at this thread, in most cases the Arena#close call succeed, but very seldomly they hit the IllegalStateException and enter the spinloop (normally only one or two iterations - depends on load).

So my opinion: It is not so serious for correctly written code. If your code is broken you may hang instead of getting correct AlreadyCloseExceptions.

There is always a workaround: set MMapDirectory.setGrouping(NO_GROUPING) which restores the older 9.x behaviour.

ayinresh:
Hi @uschindler,

This totally makes sense. Thank you again for the fix and help debugging.

Regards,
@ayinresh

Issue #15113, opened 2025-08-22
Move long[] group varint to backward-codecs


jpountz:
### Description

This is a follow-up of #15104. The benchmark was accidentally calling legacy code that writes the output of a varint group into a long[]. This wouldn't have happened if this code - which is only used by backward codecs - was in lucene/backward-codecs.

Exporterhe:
Hi, I'm new to open source and would like to start contributing. I'd like to work on this issue — could you please assign it to me or guide me on how to get started?

jpountz:
@Exporterhe I'd suggest for #15116 to be complete before starting working on this. FWIW I prefer to not assign issues, just open a PR when you have time and link this issue? The odds that someone else is working on the same issue at the same time are very low.

jpountz:
@Exporterhe You can look into it now.

SAKSHIC:
@jpountz, I have raised the PR #15186 
Please review. I will respond to the review comments as quickly as I can.

SAKSHIC:
@jpountz Gentle reminder to review https://github.com/apache/lucene/pull/15186

subhramit:
Hi, this seems to have been abandoned. May I take it? @uschindler, @easyice / @jpountz

easyice:
@subhramit Hi, thanks for looking into this, feel free to take it forward! Would it make sense to continue on PR #15186?

subhramit:
> @subhramit Hi, thanks for looking into this, feel free to take it forward! 

Hey, thanks! 

>Would it make sense to continue on PR #15186?

That would require me to have write access to their fork, which, given the inactivity, might be a greater hassle. Since the codebase has also changed over the past year, there also seem to be merge conflicts on that PR's state. 
I'd prefer making a fresh PR from scratch, but I'll keep any existing review comments in mind if relevant.

PR #15116, opened 2025-08-22
Rewrite of the GroupVInt optimization without lambdas, varhandles and no code in subclasses

uschindler:
Hi,
this is a rewrite of the GroupVInt optimization. The results are great for MMapDirectory (more than twice as fast than the baseline implementation. The code has no risk of varhandles or lambda optimization and no callsite pollution issue, because it basically requires no optimized implementation in subclasses.

The idea is the following: Basically for GroupVIntReading in the branchless optimization we need random access to the IndexInput. The workaround in the currently existing code is to have a lambda or VarHandle to do random access as a callback.

But actually we have the correct interface available out of box: `RandomAcessInput`.

I simply changed the `GroupVIntUtil.readGroupVInt()` method to check if the `DataInput` passed in also implements `RandomAccessInput` and allows seeking (`IndexInput`). If theres enough free space till end of file _and_ it implements both interfaces the optimized impl is used:
- it casts the DataInput to RandomAccessInput for random access
- it casts the DataInput to IndexInput to get access to filesize and file pointer and seeking to cleanup position after read.

If any condition does not apply it falls back to the sequential impl with branches.

The good thing with this patch;
- No need to implement anything in custom IndexInputs.
- Out of box huge speedup without any custom impl in IndexInput if it implements random access

Small downsides:
- NIOFSDir no longer has the optimization (I don't care) because its IndexInput does not support random access. You see this in the benchmark - same numbers!
- ByteBuffersDirectory implement RandomAccess, but the code of ByteBuffersIndexInput has too many bounds checks in the positional accesses. IMHO, we should for now ignore that and rewrite ByteBuffersDirectory in main branch to reuse MemorySegmentIndexInput and the try/catch logic. This allows us also to have off-heap directory with easy release of memory (the current ByteBuffersDirectory has the downside that offheap needs GC to clean up offheap memeory as we cannot free it explicit. So when we reimplement ByteBuffersDirectory using the same code like MemorySegmentIndexInput just with MemorySegments from on-heap or shared off heap arenas we have the great performance of MMapDir also for NRTCcachingDirectory - I will open a PR for main branch to get this as a new Directoryimpl into core.

Now the numbers:
```
# JMH version: 1.37
# VM version: JDK 24, OpenJDK 64-Bit Server VM, 24+36-3646
# VM invoker: C:\Program Files\Java\jdk-24\bin\java.exe
# VM options: --add-modules=jdk.unsupported
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 3 iterations, 5 s each
# Measurement: 5 iterations, 20 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Throughput, ops/time
# Parameters: (size = 64)

GroupVIntBenchmark.benchByteArrayDataInput_readGroupVInt                 64  thrpt    5  2,154 ± 0,356  ops/us
GroupVIntBenchmark.benchByteArrayDataInput_readVInt                      64  thrpt    5  1,943 ± 0,170  ops/us
GroupVIntBenchmark.benchByteBuffersIndexInput_readGroupVInt              64  thrpt    5  0,612 ± 0,024  ops/us
GroupVIntBenchmark.benchByteBuffersIndexInput_readGroupVIntBaseline      64  thrpt    5  0,763 ± 0,044  ops/us
GroupVIntBenchmark.benchMMapDirectoryInputs_readGroupVInt                64  thrpt    5  4,983 ± 0,362  ops/us
GroupVIntBenchmark.benchMMapDirectoryInputs_readGroupVIntBaseline        64  thrpt    5  1,993 ± 0,455  ops/us
GroupVIntBenchmark.benchMMapDirectoryInputs_readVInt                     64  thrpt    5  2,544 ± 0,115  ops/us
GroupVIntBenchmark.benchNIOFSDirectoryInputs_readGroupVInt               64  thrpt    5  2,461 ± 0,317  ops/us
GroupVIntBenchmark.benchNIOFSDirectoryInputs_readGroupVIntBaseline       64  thrpt    5  2,027 ± 0,422  ops/us
GroupVIntBenchmark.bench_writeGroupVInt                                  64  thrpt    5  1,347 ± 0,138  ops/us
```

Some final TODOs: 
- DataInput is now free of any GroupVint specialization or method siganures. When backporting to 10.x we must keep the old methods as noop for people who have subclassed and implemented (OpenSearch does this for their encryption directory).
- DataOutput still has write methods for both GroupVInt variants (long/int). We should reallay really remove that and implement it only in GroupVIntUtil!

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
+++ b/lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
@@ -28,8 +28,8 @@ import org.apache.lucene.store.RandomAccessInput;
  * @lucene.internal
  */
 public final class GroupVIntUtil {
-  // the maximum length of a single group-varint is 4 integers + 1 byte flag.
-  public static final int MAX_LENGTH_PER_GROUP = 17;
+  // the maximum length of a single group-varint is 1 byte flag and 4 integers.
+  public static final int MAX_LENGTH_PER_GROUP = Byte.BYTES + 4 * Integer.BYTES;
 
   private static final int[] INT_MASKS = new int[] {0xFF, 0xFFFF, 0xFFFFFF, ~0};
 
@@ -58,7 +58,7 @@ public final class GroupVIntUtil {
    * @param dst the array to read ints into.
    * @param offset the offset in the array to start storing ints.
    */
-  public static void readGroupVInt(DataInput in, int[] dst, int offset) throws IOException {
+  private static void readGroupVInt(DataInput in, int[] dst, int offset) throws IOException {
     final int flag = in.readByte() & 0xFF;
 
     final int n1Minus1 = flag >> 6;
@@ -70,7 +70,7 @@ public final class GroupVIntUtil {
     // we use a branch-less implementation:
     if (in instanceof RandomAccessInput rin && in instanceof IndexInput iin) {
       long pos = iin.getFilePointer();
-      if (iin.length() - pos >= MAX_LENGTH_PER_GROUP) {
+      if (iin.length() - pos >= MAX_LENGTH_PER_GROUP - 1) {
         dst[offset] = rin.readInt(pos) & INT_MASKS[n1Minus1];
         pos += 1 + n1Minus1;
         dst[offset + 1] = rin.readInt(pos) & INT_MASKS[n2Minus1];
@@ -93,7 +93,18 @@ public final class GroupVIntUtil {
   }
 
   /** DO not use! Only visible for benchmarking purposes! */
-  public static void readGroupVInt$Baseline(DataInput in, int[] dst, int offset)
+  public static void readGroupVInts$Baseline(DataInput in, int[] dst, int limit)
+      throws IOException {
+    int i;
+    for (i = 0; i <= limit - 4; i += 4) {
+      readGroupVInt$Baseline(in, dst, i);
+    }
+    for (; i < limit; ++i) {
+      dst[i] = in.readVInt();
+    }
+  }
+
+  private static void readGroupVInt$Baseline(DataInput in, int[] dst, int offset)
       throws IOException {
     final int flag = in.readByte() & 0xFF;
 
--- a/lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
+++ b/lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
@@ -59,17 +59,6 @@ public final class GroupVIntUtil {
    * @param offset the offset in the array to start storing ints.
    */
   public static void readGroupVInt(DataInput in, int[] dst, int offset) throws IOException {
-    readGroupVInt(true, in, dst, offset);
-  }
-
-  /** DO not use! Only visible for benchmarking purposes! */
-  public static void readGroupVInt$Baseline(DataInput in, int[] dst, int offset)
-      throws IOException {
-    readGroupVInt(false, in, dst, offset);
-  }
-
-  private static void readGroupVInt(boolean optimized, DataInput in, int[] dst, int offset)
-      throws IOException {
     final int flag = in.readByte() & 0xFF;
 
     final int n1Minus1 = flag >> 6;
@@ -79,7 +68,7 @@ public final class GroupVIntUtil {
 
     // if our DataInput implements RandomAccessInput for absolute access and IndexInput for seeking,
     // we use a branch-less implementation:
-    if (optimized && in instanceof RandomAccessInput rin && in instanceof IndexInput iin) {
+    if (in instanceof RandomAccessInput rin && in instanceof IndexInput iin) {
       long pos = iin.getFilePointer();
       if (iin.length() - pos >= MAX_LENGTH_PER_GROUP) {
         dst[offset] = rin.readInt(pos) & INT_MASKS[n1Minus1];
@@ -97,6 +86,22 @@ public final class GroupVIntUtil {
     }
 
     // fall-through: default impl
+    dst[offset] = readIntInGroup(in, n1Minus1);
+    dst[offset + 1] = readIntInGroup(in, n2Minus1);
+    dst[offset + 2] = readIntInGroup(in, n3Minus1);
+    dst[offset + 3] = readIntInGroup(in, n4Minus1);
+  }
+
+  /** DO not use! Only visible for benchmarking purposes! */
+  public static void readGroupVInt$Baseline(DataInput in, int[] dst, int offset)
+      throws IOException {
+    final int flag = in.readByte() & 0xFF;
+
+    final int n1Minus1 = flag >> 6;
+    final int n2Minus1 = (flag >> 4) & 0x03;
+    final int n3Minus1 = (flag >> 2) & 0x03;
+    final int n4Minus1 = flag & 0x03;
+
     dst[offset] = readIntInGroup(in, n1Minus1);
     dst[offset + 1] = readIntInGroup(in, n2Minus1);
     dst[offset + 2] = readIntInGroup(in, n3Minus1);
--- a/lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
+++ b/lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
@@ -19,6 +19,8 @@ package org.apache.lucene.util;
 import java.io.IOException;
 import org.apache.lucene.store.DataInput;
 import org.apache.lucene.store.DataOutput;
+import org.apache.lucene.store.IndexInput;
+import org.apache.lucene.store.RandomAccessInput;
 
 /**
  * This class contains utility methods and constants for group varint
@@ -41,7 +43,7 @@ public final class GroupVIntUtil {
   public static void readGroupVInts(DataInput in, int[] dst, int limit) throws IOException {
     int i;
     for (i = 0; i <= limit - 4; i += 4) {
-      in.readGroupVInt(dst, i);
+      readGroupVInt(in, dst, i);
     }
     for (; i < limit; ++i) {
       dst[i] = in.readVInt();
@@ -57,6 +59,17 @@ public final class GroupVIntUtil {
    * @param offset the offset in the array to start storing ints.
    */
   public static void readGroupVInt(DataInput in, int[] dst, int offset) throws IOException {
+    readGroupVInt(true, in, dst, offset);
+  }
+
+  /** DO not use! Only visible for benchmarking purposes! */
+  public static void readGroupVInt$Baseline(DataInput in, int[] dst, int offset)
+      throws IOException {
+    readGroupVInt(false, in, dst, offset);
+  }
+
+  private static void readGroupVInt(boolean optimized, DataInput in, int[] dst, int offset)
+      throws IOException {
     final int flag = in.readByte() & 0xFF;
 
     final int n1Minus1 = flag >> 6;
@@ -64,6 +77,26 @@ public final class GroupVIntUtil {
     final int n3Minus1 = (flag >> 2) & 0x03;
     final int n4Minus1 = flag & 0x03;
 
+    // if our DataInput implements RandomAccessInput for absolute access and IndexInput for seeking,
+    // we use a branch-less implementation:
+    if (optimized && in instanceof RandomAccessInput rin && in instanceof IndexInput iin) {
+      long pos = iin.getFilePointer();
+      if (iin.length() - pos >= MAX_LENGTH_PER_GROUP) {
+        dst[offset] = rin.readInt(pos) & INT_MASKS[n1Minus1];
+        pos += 1 + n1Minus1;
+        dst[offset + 1] = rin.readInt(pos) & INT_MASKS[n2Minus1];
+        pos += 1 + n2Minus1;
+        dst[offset + 2] = rin.readInt(pos) & INT_MASKS[n3Minus1];
+        pos += 1 + n3Minus1;
+        dst[offset + 3] = rin.readInt(pos) & INT_MASKS[n4Minus1];
+        pos += 1 + n4Minus1;
+
+        iin.seek(pos);
+        return;
+      }
+    }
+
+    // fall-through: default impl
     dst[offset] = readIntInGroup(in, n1Minus1);
     dst[offset + 1] = readIntInGroup(in, n2Minus1);
     dst[offset + 2] = readIntInGroup(in, n3Minus1);
@@ -81,55 +114,6 @@ public final class GroupVIntUtil {
       default:
         return in.readInt();
     }
-  }
-
-  /**
-   * Provides an abstraction for read int values, so that decoding logic can be reused in different
-   * DataInput.
-   */
-  @FunctionalInterface
-  public static interface IntReader {
-    int read(long v);
-  }
-
-  /**
-   * Faster implementation of read single group, It read values from the buffer that would not cross
-   * boundaries.
-   *
-   * @param in the input to use to read data.
-   * @param remaining the number of remaining bytes allowed to read for current block/segment.
-   * @param reader the supplier of read int.
-   * @param pos the start pos to read from the reader.
-   * @param dst the array to read ints into.
-   * @param offset the offset in the array to start storing ints.
-   * @return the number of bytes read excluding the flag. this indicates the number of positions
-   *     should to be increased for caller, it is 0 or positive number and less than {@link
-   *     #MAX_LENGTH_PER_GROUP}
-   */
-  public static int readGroupVInt(
-      DataInput in, long remaining, IntReader reader, long pos, int[] dst, int offset)
-      throws IOException {
-    if (remaining < MAX_LENGTH_PER_GROUP) {
-      readGroupVInt(in, dst, offset);
-      return 0;
-    }
-    final int flag = in.readByte() & 0xFF;
-    final long posStart = ++pos; // exclude the flag bytes, the position has updated via readByte().
-    final int n1Minus1 = flag >> 6;
-    final int n2Minus1 = (flag >> 4) & 0x03;
-    final int n3Minus1 = (flag >> 2) & 0x03;
-    final int n4Minus1 = flag & 0x03;
-
-    // This code path has fewer conditionals and tends to be significantly faster in benchmarks
-    dst[offset] = reader.read(pos) & INT_MASKS[n1Minus1];
-    pos += 1 + n1Minus1;
-    dst[offset + 1] = reader.read(pos) & INT_MASKS[n2Minus1];
-    pos += 1 + n2Minus1;
-    dst[offset + 2] = reader.read(pos) & INT_MASKS[n3Minus1];
-    pos += 1 + n3Minus1;
-    dst[offset + 3] = reader.read(pos) & INT_MASKS[n4Minus1];
-    pos += 1 + n4Minus1;
-    return (int) (pos - posStart);
   }
 
   private static int numBytes(int v) {
--- a/lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
+++ b/lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
@@ -70,7 +70,7 @@ public final class GroupVIntUtil {
     // we use a branch-less implementation:
     if (in instanceof RandomAccessInput rin && in instanceof IndexInput iin) {
       long pos = iin.getFilePointer();
-      if (iin.length() - pos >= MAX_LENGTH_PER_GROUP - 1) {
+      if (iin.length() - pos >= 4 * Integer.BYTES) {
         dst[offset] = rin.readInt(pos) & INT_MASKS[n1Minus1];
         pos += 1 + n1Minus1;
         dst[offset + 1] = rin.readInt(pos) & INT_MASKS[n2Minus1];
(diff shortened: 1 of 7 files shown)


lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
```diff
@@ -57,13 +59,44 @@ public static void readGroupVInts(DataInput in, int[] dst, int limit) throws IOE
...
+    readGroupVInt(true, in, dst, offset);
+  }
+
+  /** DO not use! Only visible for benchmarking purposes! */
+  public static void readGroupVInt$Baseline(DataInput in, int[] dst, int offset)
+      throws IOException {
+    readGroupVInt(false, in, dst, offset);
+  }
+
+  private static void readGroupVInt(boolean optimized, DataInput in, int[] dst, int offset)
```
rmuir: can we still benchmark this thing properly but avoid this `optimized` boolean which is not a constant? the method is pretty low-level and i know it is just one branch but...

lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
```diff
@@ -57,13 +59,44 @@ public static void readGroupVInts(DataInput in, int[] dst, int limit) throws IOE
...
     final int flag = in.readByte() & 0xFF;
 
     final int n1Minus1 = flag >> 6;
     final int n2Minus1 = (flag >> 4) & 0x03;
     final int n3Minus1 = (flag >> 2) & 0x03;
     final int n4Minus1 = flag & 0x03;
 
+    // if our DataInput implements RandomAccessInput for absolute access and IndexInput for seeking,
+    // we use a branch-less implementation:
+    if (optimized && in instanceof RandomAccessInput rin && in instanceof IndexInput iin) {
```
rmuir: should/can we fix RandomAccessInput to be be an IndexInput? Is there any example where it is not? Maybe its not to be done in this PR but as a separate one, just a question.

rmuir: just wondering, if e.g. RandomAccessInput implied IndexInput, maybe we could even put the readGroupVint() there on RandomAccessInput instead of DataInput. let it always be optimized and there's no other slow way? At least we should look to see what is possible to remove checks and just have more sanity.

I don't know if there are cases in the java classes that would make this not doable.

lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
uschindler: That was just a quick hack to make it work. Keep in mind this is a draft.

lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
uschindler: We should not do this because it is separated regarding the position. Random access input is stateless.

If we want to change anything this would be main only and needs more thoughts.

The instanceof checks are optimized away so is not critical.

uschindler: But we could really do a review about Index inputs and the others for Lucene11.

I am am for example not sure why we need plain DataInputs. It should be an interface which is implemented by classes.

Maybe we can change the code of GroupVIntUtil completely to use IndexInput and drop DataInput.

I just wanted quick results if it's doable and this PR verifies: it works and is doable.

lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
uschindler: It was not easy to implement the baseLine only in the benchamrk as it would have been too much code duplication. I will think about a better way to separate those. I also do not like the public `readGroupVInts$BaseLine()` method in the class. It can't be pkg-private because benchmark is (in contrast to tests) in separate package. And SharedSecrets is even more complicated...

uschindler:
I removed the boolean and duplicated the code. No changes in benchmark.

lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
uschindler: I removed it.

lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
uschindler: Basically if you are afraid of the two instanceof checks which do not seem to have an impact if all is optimized (which is the case for the benchmark as it is isolated with only exactly one implementation), then my original idea was:

Instead of having GroupVIntUtil a purely static class, i'd make in have 2 implementation (base class sealed + abstract). This is not only cleaner to implement and we don't need to expose the baseline, but we can also make the code compile and inline even better, because for reading you could then allocate an instance for reading GroupVInts on top of DataInput when you open the index once. It is just a wrapper around the IndexInput. There will be exactly 2 implementations available, so the overhead is minimal.

Basically in all users of this class (codecs), we do:

```java
// open index file
final (Data|Index)Input input = ....
final GroupVIntUtil gvintInput = GroupVIntUtil.getReadInstance(input);

// code that reads:
gvintInput.readGroupVInt(....)
```

The static `getReadInstance()` method would return one of both implementations that is optimized or not. To benchmark baseline we can add a method to retun an unoptimized instance on request.

That would fit the current usage patterns perfectly.

uschindler: But for the future my idea would be: Enforce all `IndexInput` to allow positional access and remove the RandomAccessInterface.

This requires a bit work for the downstream implementor, but it would avoid stuff like the currently unoptimized NIOFSDir. All good implementations already implement random access, and all the other ones create clones when random access is requested. This random acess is implemented by some stupid seek/read implementation. The clone is needed because of the seeking to not touch the state.

If all IndexInputs needs to support random access all have to implement it correctly (also BufferedIndexInput can do this easily).

If IndexInput is guranteed to support random access, then we can have 2 GroupVInt implementations: One for plain DataImports and an overload for IndexInput that delegate to non-random one if theres not enough space till EOF.

uschindler: So my proposal for main would be to open an issue to remove RandomAcessInput and the accessors and move the requirement to all IndexInputs (without a default impl).

uschindler: ... and let us use this implementation for 10.x

rmuir: yeah i dont want to fix any of this on this PR, it is just an idea. If we clean up the types and remove the instanceof checks, it makes for cleaner code. 

I was not really thinking about performance: we have a lot of implementations of this compression today, and i'm worried some are just theoretical and not really needed.

lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
```diff
@@ -64,6 +66,42 @@ public static void readGroupVInt(DataInput in, int[] dst, int offset) throws IOE
     final int n3Minus1 = (flag >> 2) & 0x03;
     final int n4Minus1 = flag & 0x03;
 
+    // if our DataInput implements RandomAccessInput for absolute access and IndexInput for seeking,
+    // we use a branch-less implementation:
+    if (in instanceof RandomAccessInput rin && in instanceof IndexInput iin) {
+      long pos = iin.getFilePointer();
+      if (iin.length() - pos >= MAX_LENGTH_PER_GROUP) {
```
jpountz: There is an off-by-one error since `MAX_LENGTH_PER_GROUP` includes the byte flag that is read a few lines above?

jpountz (commented):
The approach is fine with me. I agree it's fine if `NIOFSDirectory` is not optimized. I'd also be fine if `ByteBuffersDirectory` was not optimized, since it's typically used for the smaller segments only (no action item, just in case this gives you more ideas).

There's some assymetry, with `vint` or `zint` being available on `DataInput` while group-vint requires reading through another class. Should vint/zint be exposed through a helper class too? (Again, I'm thinking out loud, let's not do something like that in this PR.)

lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
easyice: +1

easyice:
+1 for move read/write methods from DataInput/DataOutput to GroupVIntUtil. should we address #15113 in this PR?

easyice (commented):
Can this method be made private now? It is only used by `readGroupVInts`. maybe the JavaDoc for this method should be updated as well.
```
public static void readGroupVInt(DataInput, int[], int)
```

lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
uschindler: Actually this is not a real bug, it just causes that it fallbacks to the slower version for the end of the file.

I think correct would be `>` instead of `>=`, correct?

uschindler:
> Can this method be made private now? It is only used by `readGroupVInts`. maybe the JavaDoc for this method should be updated as well.
> 
> ```
> public static void readGroupVInt(DataInput, int[], int)
> ```

I was also thinking about this. At least in main branch it should be! This would also make the extra javadocs comment superfluous.

lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
uschindler: Nah, we must -1. I'd like to get rid of the extra calculation, any better suggestion? Or just substract one from the constant?

uschindler: I checked, the constant is no longer used anywhere outside of this class. We should maybe remove it from the usitility class in main branch!

jpountz: Maybe do `4 × Integer.BYTES` instead of the constant?

uschindler: Locally I already changed the constant to use the `Integer:BYTES`.... I think I will remove the constant completely, izt is only used at 2 (err 3 because of `long[]` write) places.

uschindler: I changed the code, but left the constant for now. The constant is still needed for writing when we refactor the code to guess the maximum scratch size needed.

uschindler:
> > Can this method be made private now? It is only used by `readGroupVInts`. maybe the JavaDoc for this method should be updated as well.
> > ```
> > public static void readGroupVInt(DataInput, int[], int)
> > ```
> 
> I was also thinking about this. At least in main branch it should be! This would also make the extra javadocs comment superfluous.

Changed the method to private.

uschindler:
Should we maybe merge this branch to main and wait for Mike's benchmarks. If we backport, we need to make sure people customizing IndexInputs don't fail too worse or we need to add a documented way (like remove custom imp of method XY in IndexInput) to make code work.

I have seen on mailing list that there are plans to relaese 10.3, so maybe let's give this a try on main (we can revert easily).

uschindler:
Of course in 10.x I would keep the methods now made private. Basically on 10.x the methods in IndexInput would stay alive but are no longer called. So code implementing would not fail hard with `@Override`.

uschindler:
Last version had more or less same performance on my laptop:

```
GroupVIntBenchmark.benchByteArrayDataInput_readGroupVInt                 64  thrpt    5  2,128 ± 0,358  ops/us
GroupVIntBenchmark.benchByteArrayDataInput_readVInt                      64  thrpt    5  2,030 ± 0,218  ops/us
GroupVIntBenchmark.benchByteBuffersIndexInput_readGroupVInt              64  thrpt    5  0,576 ± 0,068  ops/us
GroupVIntBenchmark.benchByteBuffersIndexInput_readGroupVIntBaseline      64  thrpt    5  0,746 ± 0,146  ops/us
GroupVIntBenchmark.benchMMapDirectoryInputs_readGroupVInt                64  thrpt    5  4,518 ± 0,617  ops/us
GroupVIntBenchmark.benchMMapDirectoryInputs_readGroupVIntBaseline        64  thrpt    5  1,877 ± 0,126  ops/us
GroupVIntBenchmark.benchMMapDirectoryInputs_readVInt                     64  thrpt    5  2,300 ± 0,283  ops/us
GroupVIntBenchmark.benchNIOFSDirectoryInputs_readGroupVInt               64  thrpt    5  2,398 ± 0,465  ops/us
GroupVIntBenchmark.benchNIOFSDirectoryInputs_readGroupVIntBaseline       64  thrpt    5  1,908 ± 0,168  ops/us
GroupVIntBenchmark.bench_writeGroupVInt                                  64  thrpt    5  1,304 ± 0,191  ops/us
```

lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
```diff
@@ -56,7 +58,54 @@ public static void readGroupVInts(DataInput in, int[] dst, int limit) throws IOE
...
+    final int n1Minus1 = flag >> 6;
+    final int n2Minus1 = (flag >> 4) & 0x03;
+    final int n3Minus1 = (flag >> 2) & 0x03;
+    final int n4Minus1 = flag & 0x03;
+
+    // if our DataInput implements RandomAccessInput for absolute access and IndexInput for seeking,
+    // we use a branch-less implementation:
+    if (in instanceof RandomAccessInput rin && in instanceof IndexInput iin) {
+      long pos = iin.getFilePointer();
+      if (iin.length() - pos >= MAX_LENGTH_PER_GROUP - 1) {
```
jpountz: FWIW the following reads a bit easier to me. It's very clear since it reads 4 integers a few lines below. Versus understanding what this constant is and why we need to subtract one.

```suggestion
      if (iin.length() - pos >= 4 * Integer.BYTES) {
```

jpountz (approved):
+1 to merge this and see the impact on nightlies

uschindler:
I will add a change log when backporting this. I am not yet sure how to not break implementations, so let's wait a bit and let it bake om main.

jpountz:
FWIW we break implementations quite regularly, which I think is fine. I wouldn't worry about it. We should be careful with not breaking users who call our (non-internal) APIs. In my opinion, as soon as you start implementing interfaces / extending classes, you're on your own.

uschindler:
Hi,
the last benchmark on Mike's machine showed..... NO CHANGE AT ALL.

So I am confident that this code simplification does not bring any harm but for normal query execution also no large impacts. So basically I am not even sure if we really need an optimization at all!

jpountz:
I see a ~3.5% speedup on prefix queries: https://benchmarks.mikemccandless.com/Prefix3.html, which is consistent with what we saw when we introduced group-varint in for tail postings: multi-term queries that process many relatively short postings lists (so that tail postings are a non-negligible share of the total number of processed postings) got a speedup.

uschindler:
Thanks Adrien. It is strange that the overview page of Mike's bench did not show this. It said 0% everywhere.

Anyways, the explanation makes sense and confirms that the impact of the group vint is minimal for most queries: we only use group vint for the tail of postings, so only queries with very short postings make it visible, otherwise the impact of change is hidden in the large postings that use (P)FOR encoding for most of their postings.

So in short: let's keep this and backport it?

Interestingly the change did not make a difference for Mike's doubt with introducing group vint in vector indexing. I think the allocation profile he has seen was mainly caused by warmup. Lambdas always see a longer warmup time and the allocations caused by the capturing lambda are always seen during warmup until the escape analysis has removed them.

Nevertheless it's better without capturing lambda. And code is simpler anyways.

In followup I will try to reimplement ByteBuffersDirectory for main branch with a new MemorySegment variant based on current mmap code. But that's unrelated, it will only improve the NRTDirectory. Robert and I had some ideas. Nowadays we can make it also offheap without the garbage collector issues due to missing unmapping.

jpountz:
I see it on the overview page here: https://benchmarks.mikemccandless.com/2025.08.25.18.04.16.html, maybe you were looking at the wrong day?

Capture d’écran 2025-08-27 à 09 36 29

> So in short: let's keep this and backport it?

This is fine with me.

> In followup I will try to reimplement ByteBuffersDirectory for main branch with a new MemorySegment variant based on current mmap code. But that's unrelated, it will only improve the NRTDirectory.

FWIW I also liked your other idea of merging IndexInput and RandomAccessInput together.

uschindler:
> I see it on the overview page here: https://benchmarks.mikemccandless.com/2025.08.25.18.04.16.html, maybe you were looking at the wrong day?
> 
> Capture d’écran 2025-08-27 à 09 36 29

Yes I was on same page. The stupid thing is that you don't see the percentage change to previous run, you have to go through all numbers one by one to see the impact.

I just checked the green numbers and all showed 0. Can anybody explain what they mean? What is the X? Why can't it show with more decial digits, so we see a factor like 1.035?

> > So in short: let's keep this and backport it?
> 
> This is fine with me.

Will do. I will still keep a bit of backwards compatibility so people don't fail when they implement the method in IndexInput. It is just not used.

> > In followup I will try to reimplement ByteBuffersDirectory for main branch with a new MemorySegment variant based on current mmap code. But that's unrelated, it will only improve the NRTDirectory.
> 
> FWIW I also liked your other idea of merging IndexInput and RandomAccessInput together.

Yeah, that's something long overdue. The RandomAccessIndexInput was only introduced at Lucene 3.x times when backwards compatibility was a high priority. At that time adding a method without a useful default implementation (which was not possible due to the seeking issue needing a separate instance) was not "allowed". So we had no other chance than adding a new interface that can be optionally implemented. We also added a getter to retrieve a "view" of the IndexInput as "random". Most "good" implementations just return itsself, but the default backwards compatibility layer returned a new clone instance which does seek/read pairs (and therefor modifies state). BTW, NIOFS and all BufferedIndexInputs have exactly the problem of requiring a clone at moment.

I will use a step-by-step process:
- First let IndexInput implement RandomAccessInput from beginning, so any implementing code fails to compile
- Let the RandomAccess getter method return `this`
- Then add impl for all failing subclasses
- Then remove the getter and adapt code everywhere to fix types
- Finally remove the interface

I will need some time to do this, but will work on it!
Uwe

msokolov:
> I just checked the green numbers and all showed 0. Can anybody explain what they mean? What is the X? Why can't it show with more decial digits, so we see a factor like 1.035?

The green zeros are "p-values" telling you the probability that the difference measured was due to random variation. If it is zero that means it is certainly (if you believe in probability) a real change.  I agree the 1.0X are confusing - not showing enough significant digits, probably?

uschindler:
> > I just checked the green numbers and all showed 0. Can anybody explain what they mean? What is the X? Why can't it show with more decial digits, so we see a factor like 1.035?
> 
> The green zeros are "p-values" telling you the probability that the difference measured was due to random variation. If it is zero that means it is certainly (if you believe in probability) a real change. I agree the 1.0X are confusing - not showing enough significant digits, probably?

I don't trust statistics, only real facts :-)

The p-value is last column. The green "1 X" column says "% change" on its title:
image

And you still dont explain what the "X" means.  If it is a percentage like title of column says it's plain wrong, as the improvement is 3%. If it is a factor (1.03 X; X like times???) it has too few digits.

uschindler:
Backport is here: https://github.com/apache/lucene/pull/15138

msokolov:
> And you still dont explain what the "X" means. If it is a percentage like title of column says it's plain wrong, as the improvement is 3%. If it is a factor (1.03 X; X like times???) it has too few digits.

1X (that is my new way of agreeing "100%")

uschindler:
> > And you still dont explain what the "X" means. If it is a percentage like title of column says it's plain wrong, as the improvement is 3%. If it is a factor (1.03 X; X like times???) it has too few digits.
> 
> 1X (that is my new way of agreeing "100%")

Then title of column is wrong. If it is 100% change it would be great!

mikemccand:
> > > And you still dont explain what the "X" means. If it is a percentage like title of column says it's plain wrong, as the improvement is 3%. If it is a factor (1.03 X; X like times???) it has too few digits.
> > 
> > 
> > 1X (that is my new way of agreeing "100%")
> 
> Then title of column is wrong. If it is 100% change it would be great!

I just pushed a fix so at least we get more than 1 digit of precision there!  `Prefix3` improvement on [08/25 run](https://benchmarks.mikemccandless.com/2025.08.25.18.04.16.html) will be more like `1.0362 X` going forwards...  but, yeah, the column header is wrong too!!  Long ago I used `X` (multiple-of) because we were tinkering with improvements that were many-fold hah.  Now that we are excited about 3-5% improvements I will fix the units!

mikemccand:
In the [8/25 run](https://benchmarks.mikemccandless.com/2025.08.25.18.04.16.html), Pre and PostFilteredVectorSearch also saw nice gains!

4.8606994665086% and 3.21647379169503%, respectively ;)

uschindler:
> In the [8/25 run](https://benchmarks.mikemccandless.com/2025.08.25.18.04.16.html), Pre and PostFilteredVectorSearch also saw nice gains!
> 
> 4.8606994665086% and 3.21647379169503%, respectively ;)

Yeah, I saw this too (used a calculator entering the numbers in first and 3rd column by hand. :-)

So basically you have an improvement for the GroupVInt encoding when using MMapDir on prefix queries with many seldom terms and for vectors.

The backport is ongoing, see https://github.com/apache/lucene/pull/15138

PR #15118, opened 2025-08-24
LUCENE-14611: Fix TestTermInSetQuery.testDuel OOM by reducing large r…

Narendra9793:
This PR addresses the issue where TestTermInSetQuery.testDuel occasionally ran into OutOfMemoryError during randomized stress tests.

🔧 Changes Made:

Updated testDuel to limit the size of randomly generated terms and queries.

Added safeguards to avoid unbounded memory usage during stress test execution.

Ensured the test remains effective in validating correctness while being memory-efficient.

✅ Verification:

Ran affected test multiple times locally to confirm no OutOfMemoryError occurs.

Verified test correctness and consistency across different seeds.

### DIFF
--- a/lucene/core/src/test/org/apache/lucene/search/TestTermInSetQuery.java
+++ b/lucene/core/src/test/org/apache/lucene/search/TestTermInSetQuery.java
@@ -112,16 +112,23 @@ public class TestTermInSetQuery extends LuceneTestCase {
   public void testDuel() throws IOException {
     final int iters = atLeast(2);
     final String field = "f";
+    final int maxTerms = 256;
     for (int iter = 0; iter < iters; ++iter) {
       final List<BytesRef> allTerms = new ArrayList<>();
-      final int numTerms = TestUtil.nextInt(random(), 1, 1 << TestUtil.nextInt(random(), 1, 10));
+      // final int numTerms = TestUtil.nextInt(random(), 1, 1 << TestUtil.nextInt(random(), 1, 10));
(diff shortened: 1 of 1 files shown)


lucene/core/src/test/org/apache/lucene/search/TestTermInSetQuery.java
```diff
@@ -112,16 +112,23 @@ public void testAllDocsInFieldTerm() throws IOException {
...
     final String field = "f";
+    final int maxTerms = 256;
     for (int iter = 0; iter < iters; ++iter) {
       final List<BytesRef> allTerms = new ArrayList<>();
-      final int numTerms = TestUtil.nextInt(random(), 1, 1 << TestUtil.nextInt(random(), 1, 10));
+      // final int numTerms = TestUtil.nextInt(random(), 1, 1 << TestUtil.nextInt(random(), 1, 10));
+      final int numTerms = Math.min(
+          TestUtil.nextInt(random(), 1, 1 << TestUtil.nextInt(random(), 1, 10)),
+          maxTerms
+      );
```
dweiss, line 122: I'd reuse RandomizedTest.randomIntBetween(...) instead of this code.

dweiss (changes requested):

dweiss (commented):
I'd like to see the reason why this test OOMs first - what's causing the memory blowout? Limiting the ranges doesn't seem like a good fix for the underlying problem - you also increase the number of documents in this patch, which is counterintuitive to me.

PR #15120, opened 2025-08-25
Add estimatedByteSizes to merges kicked off by IndexWriter.addIndexes(CodecReader[])

### LINKED ISSUE #14993: Merges kicked off by IndexWriter.addIndexes(CodecReader[]) fail to set estimated size
I was debugging an issue with `addIndexes`, turned on `IndexWriter`'s `InfoStream` logging, and `ConcurrentMergeScheduler` periodically logs juicy details like this:

```
MS 0 [2025-07-24T19:54:48.516458703Z; pool-4-thread-2]: updateMergeThreads ioThrottle=false targetMBPerSec=20.0 MB/sec
merge thread Lucene Merge Thread #19 estSize=1137.2 MB (written=287.1 MB) runTime=15.6s (stopped=0.0s, paused=0.0s) rate=unlimited
  leave running at Infinity MB/sec
merge thread Lucene Merge Thread #20 estSize=0.0 MB (written=0.0 MB) runTime=0.0s (stopped=0.0s, paused=0.0s) rate=unlimited
  leave running at Infinity MB/sec
merge thread Lucene Merge Thread #21 estSize=0.0 MB (written=0.0 MB) runTime=0.0s (stopped=0.0s, paused=0.0s) rate=unlimited
  leave running at Infinity MB/sec
merge thread Lucene Merge Thread #22 estSize=0.0 MB (written=0.0 MB) runTime=0.0s (stopped=0.0s, paused=0.0s) rate=unlimited
  leave running at Infinity MB/sec
```

The top merge (`Lucene Merge Thread #19`) was a natural merge and it correctly sets the `estSize` (`OneMerge.estimatedMergeBytes`), [here](https://github.com/apache/lucene/blob/branch_10_2/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java#L4896-L4906).

But the other merges all (incorrectly) show `estSize=0` because their merges bypass `IndexWriter.registerMerge` and instead [register here](https://github.com/apache/lucene/blob/branch_10_2/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java#L3295-L3299) (in `AddIndexesMergeSource`).

This estimated size is useful for debugging ... e.g. [segment tracing tools](https://githubsearch.mikemccandless.com/segments_15.html) use it.  Also, `ConcurrentMergeScheduler` uses this when it throttles merges.

I think it should be a simple fix -- maybe factor out that code that computes the `estimatedMergeSize` for natural merges today, so it can be shared for merges triggered by `addIndexes` too?

10.2.1 -- NOTE it is possible this has been fixed in 10.2.2 / 10.x -- I have not checked!

cwperks:
### Description

This PR abstracts the logic to add estimatedByteSizes on merges to a method called `addEstimatedBytesToMerge` so that it can be re-used in multiple places.

This PR adds a simple test to ensure that estimated bytes are added to the merge after `addEstimatedBytesToMerge(MergePolicy.OneMerge merge)` is called

The test can be run with `./gradlew :lucene:core:test --tests TestIndexWriterMerging.testAddEstimatedBytesToMerge -i`

Resolves https://github.com/apache/lucene/issues/14993

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -21,6 +21,7 @@ import static org.apache.lucene.util.ByteBlockPool.BYTE_BLOCK_SIZE;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.time.Instant;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -3287,8 +3288,8 @@ public class IndexWriter
     public void registerMerge(MergePolicy.OneMerge merge) {
       try {
         addEstimatedBytesToMerge(merge);
-      } catch (IOException _) {
-        // ignore and append to pending merges
+      } catch (IOException e) {
+        throw new UncheckedIOException(e);
       }
       synchronized (IndexWriter.this) {
         pendingAddIndexesMerges.add(merge);
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -3284,8 +3284,10 @@ public class IndexWriter
       this.writer = writer;
     }
 
-    public void registerMerge(MergePolicy.OneMerge merge) throws IOException {
-      addEstimatedBytesToMerge(merge);
+    public void registerMerge(MergePolicy.OneMerge merge){
+      try {
+        addEstimatedBytesToMerge(merge);
+      } catch (IOException ignore) { }
       synchronized (IndexWriter.this) {
         pendingAddIndexesMerges.add(merge);
       }
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -3284,7 +3284,7 @@ public class IndexWriter
       this.writer = writer;
     }
 
-    public void registerMerge(MergePolicy.OneMerge merge){
+    public void registerMerge(MergePolicy.OneMerge merge) {
       try {
         addEstimatedBytesToMerge(merge);
       } catch (IOException ignore) {
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -4782,6 +4782,7 @@ public class IndexWriter
     closeMergeReaders(merge, true, false);
   }
 
+  /** Compute {@code estimatedMergeBytes} and {@code totalMergeBytes} for a merge. */
   public void addEstimatedBytesToMerge(MergePolicy.OneMerge merge) throws IOException {
     assert merge.estimatedMergeBytes == 0;
     assert merge.totalMergeBytes == 0;
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -3287,7 +3287,9 @@ public class IndexWriter
     public void registerMerge(MergePolicy.OneMerge merge){
       try {
         addEstimatedBytesToMerge(merge);
-      } catch (IOException ignore) { }
+      } catch (IOException ignore) {
+        ignore.printStackTrace(System.err);
+      }
       synchronized (IndexWriter.this) {
         pendingAddIndexesMerges.add(merge);
       }
--- a/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
+++ b/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
@@ -3287,8 +3287,8 @@ public class IndexWriter
     public void registerMerge(MergePolicy.OneMerge merge) {
       try {
         addEstimatedBytesToMerge(merge);
-      } catch (IOException ignore) {
-        ignore.printStackTrace(System.err);
+      } catch (IOException _) {
+        // ignore and append to pending merges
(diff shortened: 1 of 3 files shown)


msokolov:
so this moves the accounting from deep in `registerMerge` to the top? Did the test fail before this change?

cwperks:
> so this moves the accounting from deep in `registerMerge` to the top? Did the test fail before this change?

@msokolov merges registered in `public void registerMerge(MergePolicy.OneMerge merge)` did not get the estimated bytes added like those registered in `private synchronized boolean registerMerge(MergePolicy.OneMerge merge)` do. This leads to [information missing in log statements](https://github.com/apache/lucene/issues/14993) like the one's pointed out in the issue description.

I added a new unit test for the new method `testAddEstimatedBytesToMerge`, but I think it may also be good to ensure merges get this information regardless of how they're registered in an integration test. Do you know the proper place to add a test like that?

msokolov:
Lucene doesn't really separate unit tests from integration tests, because it doesn't ship a service. But you can write unit tests at different levels of abstraction. I think what you've done here (create an index, add some docs, trigger merges, check the output) is about what we'd expect. But if you want to also test the size estimation of merges kicked off by "internal" policies, maybe the best way is to use RandomIndexWriter, which will sometimes randomly kick off some merges of its own?  TBH this isn't an area of the code I know super well, and I commented mostly to understand what was going on!

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -4777,6 +4782,21 @@ private void abortOneMerge(MergePolicy.OneMerge merge) throws IOException {
     closeMergeReaders(merge, true, false);
   }
 
+  /** Compute {@code estimatedMergeBytes} and {@code totalMergeBytes} for a merge. */
+  public void addEstimatedBytesToMerge(MergePolicy.OneMerge merge) throws IOException {
```
dweiss: Do we need to make this method public?

cwperks: Changed this to package-private

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMerging.java
```diff
@@ -461,4 +461,33 @@ public void run() {
...
+    MergePolicy.OneMerge merge = new MergePolicy.OneMerge(segmentInfos.asList());
+
+    writer.addEstimatedBytesToMerge(merge);
+
+    assertTrue(merge.estimatedMergeBytes > 0);
+    assertTrue(merge.totalMergeBytes > 0);
+    assertTrue(merge.estimatedMergeBytes <= merge.totalMergeBytes);
+
+    writer.close();
+    dir.close();
```
dweiss: A try-with-resources block on both, perhaps? This is important for clean up if a test fails.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -3285,6 +3285,11 @@ public AddIndexesMergeSource(IndexWriter writer) {
     }
 
     public void registerMerge(MergePolicy.OneMerge merge) {
+      try {
+        addEstimatedBytesToMerge(merge);
+      } catch (IOException _) {
+        // ignore and append to pending merges
```
dweiss: I would throw an UncheckedIOException here. If something is wrong, it'll be wrong. Let the caller know early.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -4777,6 +4782,21 @@ private void abortOneMerge(MergePolicy.OneMerge merge) throws IOException {
     closeMergeReaders(merge, true, false);
   }
 
+  /** Compute {@code estimatedMergeBytes} and {@code totalMergeBytes} for a merge. */
+  void addEstimatedBytesToMerge(MergePolicy.OneMerge merge) throws IOException {
```
dweiss, line 4787: Can this method be static too? I wonder if it uses any IW's state at all.

cwperks: I tried to make it static, but I was not able to because it calls `numDeletedDocs()` which is non-static. When I tried to make `numDeletedDocs()` static I ran into other issues. I've been trying to contribute more to Lucene lately, but still very much coming up to speed on the repo so opted to start small. Do you know if `numDeletedDocs()` and any other methods it calls can be changed to static?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
cwperks: Changed to throw UncheckedIOException

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMerging.java
cwperks: Changed to try-with-resources here.

Do you think we should create a mechanism to enforce this pattern repo wide? I see a number of tests cases (included ones in this suite) that do not follow the try-with-resources pattern.

dweiss: The project has 20+ years of history, this isn't surprising. I think we should fix these when we touch the code involved - one huge patch is probably not worth the effort here. New code should stick to new conventions (and utilities) available in the language.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
dweiss: Ok, let it stay as it is now then.

dweiss (approved):

PR #15123, opened 2025-08-25
Reusing count() for minor refactor in SortedNumericDocValuesRangeQuery

jainankitk:
### Description
Reusing count() for minor refactor in SortedNumericDocValuesRangeQuery

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/document/SortedNumericDocValuesRangeQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/document/SortedNumericDocValuesRangeQuery.java
@@ -129,10 +129,10 @@ final class SortedNumericDocValuesRangeQuery extends Query {
         }
 
         int maxDoc = context.reader().maxDoc();
-        int count = docCount(context);
+        int count = docCountIgnoringDeletes(context);
         if (count == 0) {
           return null;
-        } else if (count == 1) {
+        } else if (count == maxDoc) {
           return ConstantScoreScorerSupplier.matchAll(score(), scoreMode, maxDoc);
         }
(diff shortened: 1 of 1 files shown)


jpountz:
Javadocs of `LeafReader#numDocs` warn that it may run in O(maxDoc) time. I wonder if we should instead extract a method that tells whether it matches all or nothing (without calling numDocs()) and then use it from count() (plus calling numDocs()) and scorerSupplier()?

jainankitk:
That's a good point. I was initially thinking of doing that by adding `relate` method similar to `PointRangeQuery`, but thought this could be simpler. I overlooked this key detail regarding `LeafReader#numDocs` potentially taking O(maxDoc) time.

lucene/core/src/java/org/apache/lucene/document/SortedNumericDocValuesRangeQuery.java
```diff
@@ -199,19 +194,30 @@ public float matchCost() {
...
+          case 1 -> context.reader().numDocs();
+          default -> cnt;
+        };
+      }
 
-          return context.reader().numDocs();
+      /* Returning 1 instead of LeafReader#numDocs as it may run in O(maxDoc)
+       * which is unnecessary when docCount is invoked from ScorerSupplier
+       */
+      private int docCount(LeafReaderContext context) throws IOException {
```
jpountz: I wonder if calling it `docCountIgnoringDeletes` and returning `maxDoc` instead of 1 would make the contract of this method a bit less awkward?

jpountz (approved):

Issue #15097, opened 2025-08-19
Incorrect size estimation of BooleanQuery in LRUQueryCache causing heap exhaustion


sgup432:
### Description

We recently discovered a bug in OpenSearch where LRUQueryCache was taking ~80% of the heap, though it is capped at 10%(of heap) from OpenSearch side. While analyzing heap dump, we found out that there were bunch of large Boolean queries still lingering around in the cache bloating its size beyond its max size. 

While looking at the LRUQueryCache [code](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java#L428-L433), it seems like we calculate the query size via `accountableQuery.ramBytesUsed()` only if the desired query implements `Accountable` interface and provide its own implementation. In this case, BooleanQuery doesn't provide such implementation and we default to 1kb, leading to underestimation of query size causing the cache size to bloat up with such large queries.

Below is a snapshot from MAT where it shows LRUCache holding ~23gb of heap in an instance where we had 32gb of heap, and LRUCache is configured to be ~3.2gb in size in this case. And majority of being held by ~6k BooleanQuery each around ~3.2mb in size.

<img width="1042" height="137" alt="Image" src="https://github.com/user-attachments/assets/75d6403a-dfd6-4562-8510-ff454d899895" />

### Solution
Below are the solutions I can think of:

1. I think we can make BooleanQuery to implement Accountable interface and calculate rough estimate of ramBytesUsed from its desired clauses? Still better than using 1kb as default.
2. ~~Or Disable caching BooleanQueries with very high number of clauses. Though this still won't be accurate enough.~~ We already have the logic to not cache BooleanQuery having >16 clauses.
3. Or have a different way to calculate cache's key, value size. 

### Version and environment details

Affects all versions

jhinch-at-atlassian-com:
This seems like it would be true not only for `BooleanQuery` but for any query which is cacheable but does not implement `Accountable`. `PhraseQuery` and `MultiPhraseQuery` stand out as implementations to fix as well.

sgup432:
Yup it holds true for other queries as well. So will have to figure out a generic way to solve this.

sgup432:
I did some research, I don't see a nice way to solve this until unless we implement Accountable for most of the queries(like TermQuery etc) and BooleanQuery eventually. Might be too intrusive if it is only required in the cache.

[...]

PR #15124, opened 2025-08-26
[QueryCache] Use RamUsageEstimator instead of defaulting to 1024 bytes for non-accountable query size

sgup432:
### Description

Related issue - https://github.com/apache/lucene/issues/15097

Instead of using the default 1024 bytes for query size, we try to use RamUsageEstimator to calculate approx size. As seen in above issues, using 1024 as a default can cause issues(like heap exhaustion) due to underestimation, and even cause overestimation leading to fewer queries getting cached.

`RamUsageEstimator.sizeOf()` is usually pretty fast, I made a sample BooleanQuery with 15 clauses with a size of around 3kb. It took around 14959ns to calculate its size via `RamUsageEstimator.sizeOf()`.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
@@ -134,10 +134,6 @@ public class BooleanQuery extends Query implements Iterable<BooleanClause> {
   private BooleanQuery(int minimumNumberShouldMatch, BooleanClause[] clauses) {
     this.minimumNumberShouldMatch = minimumNumberShouldMatch;
     this.clauses = Collections.unmodifiableList(Arrays.asList(clauses));
-    this.mustClauseQueries = new ArrayList<>();
-    this.mustNotClauseQueries = new ArrayList<>();
-    this.filterClauseQueries = new ArrayList<>();
-    this.shouldClauseQueries = new ArrayList<>();
     clauseSets = new EnumMap<>(Occur.class);
     // duplicates matter for SHOULD and MUST
     clauseSets.put(Occur.SHOULD, new Multiset<>());
@@ -149,6 +145,11 @@ public class BooleanQuery extends Query implements Iterable<BooleanClause> {
     // calls(like in QueryCache), we need to iterate over clauseSets.keySet() which allocates
     // HashMap$HashSetIterator.<init>
     // for every call causing some performance impact
+
+    List<Query> mustClauseQueries = new ArrayList<>();
+    List<Query> mustNotClauseQueries = new ArrayList<>();
+    List<Query> filterClauseQueries = new ArrayList<>();
+    List<Query> shouldClauseQueries = new ArrayList<>();
     for (BooleanClause clause : clauses) {
       switch (clause.occur()) {
         case MUST:
@@ -166,6 +167,10 @@ public class BooleanQuery extends Query implements Iterable<BooleanClause> {
       }
       clauseSets.get(clause.occur()).add(clause.query());
     }
+    this.mustClauseQueries = List.copyOf(mustClauseQueries);
+    this.mustNotClauseQueries = List.copyOf(mustNotClauseQueries);
+    this.filterClauseQueries = List.copyOf(filterClauseQueries);
+    this.shouldClauseQueries = List.copyOf(shouldClauseQueries);
   }
 
   /** Gets the minimum number of the optional BooleanClauses which must be satisfied. */
--- a/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
@@ -126,10 +126,18 @@ public class BooleanQuery extends Query implements Iterable<BooleanClause> {
   private final List<BooleanClause> clauses; // used for toString() and getClauses()
   // WARNING: Do not let clauseSets escape from this class as it breaks immutability:
   private final Map<Occur, Collection<Query>> clauseSets; // used for equals/hashCode
+  private final List<Query> mustClauseQueries;
+  private final List<Query> mustNotClauseQueries;
+  private final List<Query> filterClauseQueries;
+  private final List<Query> shouldClauseQueries;
 
   private BooleanQuery(int minimumNumberShouldMatch, BooleanClause[] clauses) {
     this.minimumNumberShouldMatch = minimumNumberShouldMatch;
     this.clauses = Collections.unmodifiableList(Arrays.asList(clauses));
+    this.mustClauseQueries = new ArrayList<>();
+    this.mustNotClauseQueries = new ArrayList<>();
+    this.filterClauseQueries = new ArrayList<>();
+    this.shouldClauseQueries = new ArrayList<>();
     clauseSets = new EnumMap<>(Occur.class);
     // duplicates matter for SHOULD and MUST
     clauseSets.put(Occur.SHOULD, new Multiset<>());
@@ -137,7 +145,25 @@ public class BooleanQuery extends Query implements Iterable<BooleanClause> {
     // but not for FILTER and MUST_NOT
     clauseSets.put(Occur.FILTER, new HashSet<>());
     clauseSets.put(Occur.MUST_NOT, new HashSet<>());
+    // We store the queries per clauses in a list beforehand. As otherwise during repeated visit()
+    // calls(like in QueryCache), we need to iterate over clauseSets.keySet() which allocates
+    // HashMap$HashSetIterator.<init>
+    // for every call causing some performance impact
     for (BooleanClause clause : clauses) {
+      switch (clause.occur()) {
+        case MUST:
+          mustClauseQueries.add(clause.query());
+          break;
+        case FILTER:
+          filterClauseQueries.add(clause.query());
+          break;
+        case SHOULD:
+          shouldClauseQueries.add(clause.query());
+          break;
+        case MUST_NOT:
+          mustNotClauseQueries.add(clause.query());
+          break;
+      }
       clauseSets.get(clause.occur()).add(clause.query());
     }
   }
@@ -649,15 +675,23 @@ public class BooleanQuery extends Query implements Iterable<BooleanClause> {
   @Override
   public void visit(QueryVisitor visitor) {
     QueryVisitor sub = visitor.getSubVisitor(Occur.MUST, this);
-    for (BooleanClause.Occur occur : clauseSets.keySet()) {
+    for (Occur occur : Occur.cachedValues()) {
       if (clauseSets.get(occur).size() > 0) {
         if (occur == Occur.MUST) {
-          for (Query q : clauseSets.get(occur)) {
+          for (Query q : mustClauseQueries) {
             q.visit(sub);
           }
         } else {
           QueryVisitor v = sub.getSubVisitor(occur, this);
-          for (Query q : clauseSets.get(occur)) {
+          List<Query> queryList;
+          queryList =
+              switch (occur) {
+                case FILTER -> filterClauseQueries;
+                case SHOULD -> shouldClauseQueries;
+                case MUST_NOT -> mustNotClauseQueries;
+                case MUST -> mustClauseQueries;
+              };
+          for (Query q : queryList) {
             q.visit(v);
           }
         }
--- a/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
@@ -126,10 +126,6 @@ public class BooleanQuery extends Query implements Iterable<BooleanClause> {
   private final List<BooleanClause> clauses; // used for toString() and getClauses()
   // WARNING: Do not let clauseSets escape from this class as it breaks immutability:
   private final Map<Occur, Collection<Query>> clauseSets; // used for equals/hashCode
-  private final List<Query> mustClauseQueries;
-  private final List<Query> mustNotClauseQueries;
-  private final List<Query> filterClauseQueries;
-  private final List<Query> shouldClauseQueries;
 
   private BooleanQuery(int minimumNumberShouldMatch, BooleanClause[] clauses) {
     this.minimumNumberShouldMatch = minimumNumberShouldMatch;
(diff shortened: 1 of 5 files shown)


lucene/core/src/test/org/apache/lucene/search/TestLRUQueryCache.java
```diff
@@ -1062,6 +1063,38 @@ public void testBooleanQueryCachesSubClauses() throws IOException {
     dir.close();
   }
 
+  public void testCacheRamBytesWithALargeTermQuery() throws IOException {
+    Directory dir = newDirectory();
```
cwperks: FYI I recently received a comment to use try-with-resources for these closeable resources: https://github.com/apache/lucene/pull/15120#discussion_r2298924395

It can be updated to:

```
try (Directory dir = newDirectory(); RandomIndexWriter w = new RandomIndexWriter(random(), dir)) {
    ...
}
```

sgup432: Yeah I am fine doing this instead of closing it manually. Though test anyway fails if you don't close these resources.

sgup432:
@msfroh Would you mind taking a look at this?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -426,10 +425,8 @@ public void clear() {
   }
 
   private static long getRamBytesUsed(Query query) {
-    return LINKED_HASHTABLE_RAM_BYTES_PER_ENTRY
-        + (query instanceof Accountable accountableQuery
-            ? accountableQuery.ramBytesUsed()
-            : QUERY_DEFAULT_RAM_BYTES_USED);
+    long queryRamBytesUsed = RamUsageEstimator.sizeOf(query, 0);
+    return LINKED_HASHTABLE_RAM_BYTES_PER_ENTRY + queryRamBytesUsed;
```
benwtrent: I realize you have a small benchmark that you did, but real world performance is still a concern of mine here.

I am not sure how to benchmark this, but my concern is that non-accountable queries just got more expensive, we calculate memory used for a query on cache and eviction

sgup432: Yeah I just timed one 3kb boolean query.

>I am not sure how to benchmark this, but my concern is that non-accountable queries just got more expensive

The only way I see it is to micro-benchmark `RamUsageEstimator` with different query types and size, might give more idea. I had tried to do same but with only term and boolean queries.

>but my concern is that non-accountable queries just got more expensive

I am not sure whether it would be that expensive(atleast in absolute terms). Like for `TermQuery` and similar cheaper queries, RamUsageEstimator should be pretty fast considering we already cache ramBytesUsed for the underlying terms etc. In my opinion, complex BooleanQueries with many nested clauses might be more expensive, as we need to visit all the underlying children for size calculation. To handle that, we already have a limit of 16 clauses beyond which we can cannot cache, and it can be refined further if needed. 

I don't see any other good way of calculating query size until we implement Accountable for all queries, but it would too intrusive for just the caching use-case.

sgup432: @benwtrent So I wrote a simple code to micro-benchmark `putIfAbsent` method which is the one which calls `getRamBytesUsed(Query query)` during caching and eviction. 
Created a cache with `MAX_SIZE = 10000` and `MAX_SIZE_IN_BYTES = 1048576`, and created N sample boolean queries and the logic looks like something below
```
for (int i = 0; i < MAX_SIZE; i++) {
      TermQuery must = new TermQuery(new Term("foo", "bar" + i));
      TermQuery filter = new TermQuery(new Term("foo", "quux" + i));
      TermQuery mustNot = new TermQuery(new Term("foo", "foo" + i));
      BooleanQuery.Builder bq = new BooleanQuery.Builder();
      bq.add(must, BooleanClause.Occur.FILTER);
      bq.add(filter, BooleanClause.Occur.FILTER);
      bq.add(mustNot, BooleanClause.Occur.MUST_NOT);
      queries[i] = bq.build();
    }
```

JMH method to test 100% writes workload for 10 threads
```
@Benchmark
  @Group("concurrentPutOnly")
  @GroupThreads(10)
  public void testConcurrentPuts() {
    int random = ThreadLocalRandom.current().nextInt(MAX_SIZE);
    queryCache.putIfAbsent(
        queries[random], this.sampleCacheAndCount, cacheHelpers[random & (SEGMENTS - 1)]);
  }
```


Baseline numbers
```
Benchmark                               Mode  Cnt        Score       Error  Units
QueryCacheBenchmark.concurrentPutOnly  thrpt   15  4102080.220 ± 80816.546  ops/s
```

My changes
```
Benchmark                               Mode  Cnt       Score       Error  Units
QueryCacheBenchmark.concurrentPutOnly  thrpt   15  901925.345 ± 38059.230  ops/s
```

So it became ~4.5x slower. There were lot of evictions as well. Though this probably one of the worst case scenario with only write workload and boolean query. For a mixed read/write or simple filter queries, this might be way less.
And this is the method profile (taken from JFR). Kind of expected. 

Screenshot 2025-09-17 at 3 53 18 PM



I don't know if there is a way out or we can further improve query visitor logic itself.

sgup432: Seems like I was doing 
```
long queryRamBytesUsed = RamUsageEstimator.sizeOf(query, 0);
```

`long sizeOf(Query q, long defSize) ` Here defSize represents the shallow size of the query, and if 0 is passed, it calculates the shallow size during runtime which was doing a lot of object allocations(adding to higher CPU%) and therefore adding to latency.

I changed to `RamUsageEstimator.sizeOf(query, 32);`, Here 32 is a rough shallow bytes for a term/boolean query which I calculated manually by calling that method

After running with above change, performance improved but still 2.5x slower than the baseline.
```
Benchmark                               Mode  Cnt        Score        Error  Units
QueryCacheBenchmark.concurrentPutOnly  thrpt   15  1608680.127 ± 229218.026  ops/s
```

benwtrent: I wonder if the query cache should "cache" the query sizes? It seems we do this more expensive calculation twice for no reason. Why not keep track of the query size in the cache as well, only calculating when adding it?

sgup432: >I wonder if the query cache should "cache" the query sizes?

Yeah good point. Let me try that approach as well and update here.
I was able to further improve the performance earlier by making changes in BooleanQuery.visit() so that we store clause level queries explicitly in a list beforehand, which made it 62% faster, overall 1.5x slower than baseline

benwtrent: 1.5x is REALLY good!

sgup432: Okay, I did the change to cache query size as well in the existing map

**Baseline (100% write)**
```
Benchmark                               Mode  Cnt        Score        Error  Units
QueryCacheBenchmark.concurrentPutOnly  thrpt    9  4057602.444 ± 380153.561  ops/s
```

**My changes (100% write)**  - 25% slower than baseline
```
Benchmark                               Mode  Cnt        Score        Error  Units
QueryCacheBenchmark.concurrentPutOnly  thrpt    9  3028840.412 ± 483817.248  ops/s
```

--------------

**Baseline - 60% read, 40% write**  (6 read threads, 4 write threads)
```
Benchmark                                                           Mode  Cnt        Score         Error  Units
QueryCacheBenchmark.concurrentGetAndPuts                           thrpt    9  8074623.026 ± 1229792.902  ops/s
```

**My changes - 60% read, 40% write** 
```
Benchmark                                                           Mode  Cnt        Score         Error  Units
QueryCacheBenchmark.concurrentGetAndPuts                           thrpt    9  8262227.277 ± 1224929.417  ops/s
```


So with 100% writes, which is the worst case for any cache, my changes(to add RamUsageEstimator and BooleanQuery.visit() changes) performs 25% worse than the baseline. But this isn't a ideal workload for any cache.

So I did another benchmark to test with 60% reads, 40% writes which represents more ideal distribution for the cache.

Here both are almost equivalent, and no degradation is seen.

benwtrent: so great! I really like this.

sgup432: @benwtrent I have pushed my final changes, let me know how it looks.
On a side note, I am also going to raise a PR soon(based on my idea [here](https://github.com/apache/lucene/issues/14222)) to refactor query cache to improve performance by atleast 2-3x :)

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -705,6 +710,17 @@ public long ramBytesUsed() {
     }
   }
 
+  // pkg-private for testing
+  static class QueryMetadata {
```
benwtrent: record?

lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
```diff
@@ -137,9 +141,36 @@ private BooleanQuery(int minimumNumberShouldMatch, BooleanClause[] clauses) {
...
     clauseSets.put(Occur.MUST_NOT, new HashSet<>());
+    // We store the queries per clauses in a list beforehand. As otherwise during repeated visit()
+    // calls(like in QueryCache), we need to iterate over clauseSets.keySet() which allocates
+    // HashMap$HashSetIterator.<init>
+    // for every call causing some performance impact
+
+    List<Query> mustClauseQueries = new ArrayList<>();
+    List<Query> mustNotClauseQueries = new ArrayList<>();
+    List<Query> filterClauseQueries = new ArrayList<>();
+    List<Query> shouldClauseQueries = new ArrayList<>();
```
benwtrent: Thank you for digging into this performance impact. 

How much does this gain us in the "normal" case?

lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
```diff
@@ -649,15 +680,23 @@ public Query rewrite(IndexSearcher indexSearcher) throws IOException {
   @Override
   public void visit(QueryVisitor visitor) {
     QueryVisitor sub = visitor.getSubVisitor(Occur.MUST, this);
-    for (BooleanClause.Occur occur : clauseSets.keySet()) {
+    for (Occur occur : Occur.cachedValues()) {
```
benwtrent: If the cost is truly creating the iterator, maybe we can switch to something that still uses the key set, like `clauseSets.entries().forEach((occur, clauseSet) -> {clauseSet.forEach(clauseValue -> {...})})`;

It does seem weird to iterate the keys only to immediately use them to rehash and get the values again.

lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
sgup432: >How much does this gain us in the "normal" case?

I am not sure on this. Might have to run some existing benchmark to get an idea.

lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
sgup432: So even doing `clauseSets.forEach()` will create a iterator for the first time(as can be seen in the code and profile below)

With this change, the 100% write ops/s dropped to 2M compared to the change as part of this PR (which was 3M)
```
Benchmark                               Mode  Cnt        Score        Error  Units
QueryCacheBenchmark.concurrentPutOnly  thrpt    9  2179508.622 ± 117501.478  ops/s
```
In my case, because I am frequently calling `visit()` and subsequently `forEach()`, it allocates many of these for different boolean queries showing some impact in the micro benchmark. Also I start profiling after creating these boolean queries

Screenshot 2025-09-19 at 2 55 50 PM

lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
sgup432: @benwtrent Do you know how can I measure the performance for this in normal case? Any existing benchmark you can suggest?

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
```diff
@@ -705,6 +710,17 @@ public long ramBytesUsed() {
     }
   }
 
+  // pkg-private for testing
+  static class Record {
```
benwtrent: No, i mean this should be a `record`, not named such. 

https://docs.oracle.com/en/java/javase/17/language/records.html

lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
benwtrent: I really don't think we should be adding so much complexity to boolean. My concern is that it isn't worth it. This will add significant overhead to boolean queries and we shouldn't do it.



With caching the query size, and all your other improvements, is this now within 2-3x slower in the worst case?

sgup432: Yeah makes sense. I too think it may not bring much benefit outside this usecase. I can remove all the boolean query changes in that case
>With caching the query size, and all your other improvements, is this now within 2-3x slower in the worst case?

Yeah ~2.5x slower in the worst case.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
sgup432: Ah damn, my bad! Let me look into this again

lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
sgup432: I have removed these Boolean query changes, so resolving.

lucene/core/src/java/org/apache/lucene/search/BooleanQuery.java
sgup432: I have removed these boolean query changes now.

lucene/core/src/java/org/apache/lucene/search/LRUQueryCache.java
sgup432: Changed to record type, keeping the same name as before. Hopefully it looks okay now.

sgup432:
@benwtrent I have removed boolean query changes now.
Ran the benchmarks again. 

**Baseline (100% write)**
```
Benchmark                               Mode  Cnt        Score        Error  Units
QueryCacheBenchmark.concurrentPutOnly  thrpt    9  4057602.444 ± 380153.561  ops/s
```

**My changes (100% write) - 47% (1.9x) slower than baseline**
```
Benchmark                               Mode  Cnt        Score        Error  Units
QueryCacheBenchmark.concurrentPutOnly  thrpt    9  2133320.577 ± 335226.223  ops/s
```

**Baseline - 60% read, 40% write (6 read threads, 4 write threads)**
```
Benchmark                                                           Mode  Cnt        Score         Error  Units
QueryCacheBenchmark.concurrentGetAndPuts                           thrpt    9  8074623.026 ± 1229792.902  ops/s
```

**My changes - 60% read, 40% write - ~10% slower than baseline**
```
Benchmark                                                           Mode  Cnt        Score         Error  Units
QueryCacheBenchmark.concurrentGetAndPuts                           thrpt    9  7263435.185 ±  828131.018  ops/s
```


Overall mixed load benchmark looks fine to me, considering I was only testing with boolean queries. In a lot of cases, we might be directly caching `TermInSetQuery` or `TermsQuery` for which we already cache ramBytesInUsed, so this 10% slowness will also go away overall.

benwtrent (approved):
This looks great. The performance numbers show not too horrible a slow down, with much more accurate query sizes. 

Leaving this open for dissenting opinions for a bit, but I will merge in the next day or so :)

sgup432:
@benwtrent Do we want to merge this now or wait a bit more?

BaurzhanSakhariev:
At CrateDB we are facing a similar issue that is described in https://github.com/apache/lucene/issues/15097
Are there any plans to backport this to 10.3?

PR #15125, opened 2025-08-26
Handle inconsistent schema on flush with index sorts

dnhatn:
If index sorts are configured and a document's schema is inconsistent with existing segments, IndexWriter may encounter a NullPointerException when building sort comparators.

I considered fixing the perField hash (remove fields without fieldInfos), but this would be unnecessary and costly for large batches of malformed documents with many fields. This change adds a null check for writer before accessing its fieldInfo, consistent with how other fields are handled in IndexingChain.

```
java.lang.NullPointerException: Cannot invoke "org.apache.lucene.index.FieldInfo.getDocValuesType()" because "pf.fieldInfo" is null
	at org.apache.lucene.core@10.2.2/org.apache.lucene.index.IndexingChain$1.getSortedSetDocValues(IndexingChain.java:207)
	at org.apache.lucene.core@10.2.2/org.apache.lucene.index.DocValues.getSortedSet(DocValues.java:319)
	at org.apache.lucene.core@10.2.2/org.apache.lucene.search.SortedSetSortField.getValues(SortedSetSortField.java:194)
	at org.apache.lucene.core@10.2.2/org.apache.lucene.index.IndexSorter$StringSorter.getDocComparator(IndexSorter.java:427)
	at org.apache.lucene.core@10.2.2/org.apache.lucene.index.IndexingChain.maybeSortSegment(IndexingChain.java:261)
	at org.apache.lucene.core@10.2.2/org.apache.lucene.index.IndexingChain.flush(IndexingChain.java:274)
	at org.apache.lucene.core@10.2.2/org.apache.lucene.index.DocumentsWriterPerThread.flush(DocumentsWriterPerThread.java:456)
	at org.apache.lucene.core@10.2.2/org.apache.lucene.index.DocumentsWriter.doFlush(DocumentsWriter.java:502)
	at org.apache.lucene.core@10.2.2/org.apache.lucene.index.DocumentsWriter.maybeFlush(DocumentsWriter.java:456)
	at org.apache.lucene.core@10.2.2/org.apache.lucene.index.DocumentsWriter.flushAllThreads(DocumentsWriter.java:649)
	at org.apache.lucene.core@10.2.2/org.apache.lucene.index.IndexWriter.prepareCommitInternal(IndexWriter.java:3677)
	at org.apache.lucene.core@10.2.2/org.apache.lucene.index.IndexWriter.commitInternal(IndexWriter.java:4125)
	at org.apache.lucene.core@10.2.2/org.apache.lucene.index.IndexWriter.commit(IndexWriter.java:4087)
	at org.elasticsearch.server@9.2.0/org.elasticsearch.index.engine.InternalEngine.commitIndexWriter(InternalEngine.java:3103)
	at org.elasticsearch.stateless@9.2.0/co.elastic.elasticsearch.stateless.engine.IndexEngine.commitIndexWriter(IndexEngine.java:333)
	at org.elasticsearch.server@9.2.0/org.elasticsearch.index.engine.InternalEngine.flushHoldingLock(InternalEngine.java:2301)
	at org.elasticsearch.stateless@9.2.0/co.elastic.elasticsearch.stateless.engine.IndexEngine.flushHoldingLock(IndexEngine.java:324)
	at org.elasticsearch.server@9.2.0/org.elasticsearch.index.engine.Engine.flush(Engine.java:1496)
	at 
```


javanna:
Hey, thanks for fixing this! 

I checked out your branch and verified the fix, yet I am not very familiar with `IndexingChain`, hence I have some basic questions.

To make sure I understand the issue 100%: when there is an inconsistency, an IAE is thrown as part of `initializeFieldInfo`, which though leaves the `PerField` instance without field info set, which when accessed later causes problems, especially when consulting the doc values type to sort the segments if index sorting is enabled. Correct?

Are there other scenarios where the same could happen, that a `PerField` instance has no field info set to it, and is index sorting the only place where this causes issues? This code seems pretty delicate and it seems like things may go wrong in other ways, but I can't tell exactly.

Your fix handles the null field info by checking the doc values writer. Could you expand on why checking docValuesWriter instead of checking field info directly? Are there circumstances where you'd expect the former to be null but the latter non null?

Do we have a chance to consider a more comprehensive fix like delay adding the `PerField` to the `perFieldHash` until field info is set to it? I can't tell if there may be side effects of that though. 

Last but not least: with your fix, what is the effect on index sorting of returning null when field info is null?

dnhatn:
Thanks Luca!

> To make sure I understand the issue 100%: when there is an inconsistency, an IAE is thrown as part of initializeFieldInfo, which though leaves the PerField instance without field info set, which when accessed later causes problems, especially when consulting the doc values type to sort the segments if index sorting is enabled. Correct?

In other places before accessing fieldInfo, we check if the corresponding writer (e.g., pointValuesWriter, docValuesWriter) is null. We overlooked this null check when building comparators for index sorting.

> Are there other scenarios where the same could happen, that a PerField instance has no field info set to it, and is index sorting the only place where this causes issues? This code seems pretty delicate and it seems like things may go wrong in other ways, but I can't tell exactly.

I've reviewed the IndexingChain class, and this appears to be the only place where we missed the check.

> Your fix handles the null field info by checking the doc values writer. Could you expand on why checking docValuesWriter instead of checking field info directly? Are there circumstances where you'd expect the former to be null but the latter non null?

The fieldInfo is set just before the corresponding writer is initialized. Therefore, if the writer is non-null, the fieldInfo will also be non-null.

> Do we have a chance to consider a more comprehensive fix like delay adding the PerField to the perFieldHash until field info is set to it? I can't tell if there may be side effects of that though.

Yes, I did consider that. I also explored the option of removing PerField that doesn't have fieldInfos after hitting an exception. However, both could be expensive and significantly slow down, especially when documents contain many fields. For this reason, I opted for the current approach.

> Last but not least: with your fix, what is the effect on index sorting of returning null when field info is null?

Returning null for an index sort field can occur in two scenarios:

1. The value for the sort field never appears in the documents of the current DWPT.
2. A schema exception is encountered, and other documents do not contain the sort field.

Since we do not index fields from invalid documents, the second case is safely treated the same as the first - meaning there is no value for the sort field.

javanna:
Thanks for the answers @dnhatn they are very clear and make sense to me. 

Maybe the only doubt I have left is whether checking for field info null as opposed to writer makes more sense long term. I am worried that things may change where the initialization of the two happens differently.

> Yes, I did consider that. I also explored the option of removing PerField that doesn't have fieldInfos after hitting an exception. However, both could be expensive and significantly slow down, especially when documents contain many fields.

Would delaying adding `PerField` to the array be more expensive? Do you have an instinct of the problems it may cause? That feels like a more long term solution to me, but I may miss implications of it.


I don't mean to be a pain, the fix is good, I am good with merging it as-is. We could potentially improve on it but I don't have strong opinions on it. I would love to play with the idea of delaying adding the instance to the array further but I don't have time at the moment, sadly.

dnhatn:
> Maybe the only doubt I have left is whether checking for field info null as opposed to writer makes more sense long term.

@javanna I pushed https://github.com/apache/lucene/pull/15125/commits/a8c22e87d214e90e45e7a3793b2d0b2d8a22d51d to check the fieldInfo instead. Thanks for reviewing.

PR #15127, opened 2025-08-26
Early exit while computing globalMin/Max using skipper

jainankitk:
### Description
Skipper should exit early while computing globalMin/Max if it is not for some of the leaves as min/max cannot be computed correctly using skipper


jpountz (approved):
The change looks good. Maybe try to put the comment on the line before the return statement so that formatting looks less awkward?

PR #15131, opened 2025-08-27
Restrict visibility of `TieredMergePolicy.score()`

mccullocht:
This method uses both `MergeScore` and `SegmentSizeAndDocs` but neither of these are visible enough to
provide a meaningful custom implementation of `score()`. Given that no one has complained that is broken,
restrict visibility of this method and `MergeScore` instead.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -31,7 +31,7 @@ API Changes
 
 * GITHUB#14844: Change IndexInput.updateReadAdvice to take an IOContext instead (Simon Cooper)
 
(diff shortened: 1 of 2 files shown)


jpountz:
Good question. I've had little success with customizing TieredMergePolicy in the past, because all its features tend to interact with one another. The `score()` method may be one exception though, it should be possible to customize it without almost certainly breaking something else?

That said, if it's not actually possible to customize it today and nobody asked for this feature / complained about it, maybe it's better to further lock it down by making score() private. It will make it easier to evolve this merge policy in the future.

jpountz (approved):
Good call on migrating SegmentSizeAndDocs to a record!

mccullocht:
If someone would merge this that would be appreciated.

PR #15135, opened 2025-08-28
early exit in SortedNumericDocValuesRangeQuery & SortedSetDocValuesRangeQuery

RamakrishnaChilaka:
Skipping segments if segment's min and max are out of range of the query

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/document/SortedNumericDocValuesRangeQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/document/SortedNumericDocValuesRangeQuery.java
@@ -237,6 +237,9 @@ final class SortedNumericDocValuesRangeQuery extends Query {
       return null;
     }
 
+    if (skipper.maxValue() < lowerValue || skipper.minValue() > upperValue) {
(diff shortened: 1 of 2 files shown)


jpountz:
This doesn't seem to be saving much work, does it?

RamakrishnaChilaka:
Thanks for review @jpountz.

For example, if doc values are sorted in ascending order, and both lowerValue and upperValue are greater than the segment’s max, we end up calling advance in [SortedNumericDocValuesRangeQuery, L#259]( https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/document/SortedNumericDocValuesRangeQuery.java#L259), which exhausts the skipper. After that, nextDoc is almost a no-op, and advance for upperValue is almost a no-op too. we save the `skipper.advance` call. I agree skipper.advance is relatively cheap, but orders of magnitude more expensive than just doing two long comparisons.

TL;DR: In this PR, for sorted segments, if the query range is outside the segment, we skip skipper.advance which would otherwise exhaust the skipper.

PR #15138, opened 2025-08-29
Backport of #15116: Rewrite of the GroupVInt optimization without lambdas, varhandles and no code in subclasses

uschindler:
This backport basically uses the same code like main branch, it just preserves the method stubs of the optimiaztion, so code subclassing DataInput don't fail compilation.

All code no longer called throws UnsupportedOperationException and is tagged `@Deprecated`.

Revision: ebfd863ac728f9abfcaa42fec7fce90b62900df5
Author: Uwe Schindler <uschindler@apache.org>
Date: 25.08.2025 22:58:28
Message:
Rewrite of the GroupVInt optimization without lambdas, varhandles and no code in subclasses (#15116)

* Rewrite of the GroupVInt optimization without lamdas, varhandles and no code in subclasses

* Remove optimized flag and duplicate code

* apply suggestions/fixes: fix maximum size; cleanup baseline method and hide internal details

* Apply suggestions from code review

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
+++ b/lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
@@ -51,14 +51,18 @@ public final class GroupVIntUtil {
   }
 
   /**
-   * Default implementation of read single group, for optimal performance, you should use {@link
-   * GroupVIntUtil#readGroupVInts(DataInput, int[], int)} instead.
+   * Read single group, for optimal performance, you should use {@link
+   * GroupVIntUtil#readGroupVInts(DataInput, int[], int)} to decode a full group including tails
+   * instead.
    *
    * @param in the input to use to read data.
    * @param dst the array to read ints into.
    * @param offset the offset in the array to start storing ints.
+   * @deprecated Use {@link #readGroupVInts(DataInput, int[], int)} to decode a full group including
+   *     tails instead
    */
-  private static void readGroupVInt(DataInput in, int[] dst, int offset) throws IOException {
+  @Deprecated
+  public static void readGroupVInt(DataInput in, int[] dst, int offset) throws IOException {
     final int flag = in.readByte() & 0xFF;
 
     final int n1Minus1 = flag >> 6;
@@ -137,6 +141,7 @@ public final class GroupVIntUtil {
    * DataInput.
    *
    * @deprecated No longer used.
+   * @see #readGroupVInt(DataInput, long, IntReader, long, int[], int)
    */
   @Deprecated
   @FunctionalInterface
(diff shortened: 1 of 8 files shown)


uschindler:
The changes entry will be chekrry picked in main branch.

lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
```diff
@@ -86,7 +135,10 @@ private static int readIntInGroup(DataInput in, int numBytesMinus1) throws IOExc
   /**
    * Provides an abstraction for read int values, so that decoding logic can be reused in different
    * DataInput.
+   *
+   * @deprecated No longer used.
```
jpountz, line 143: Maybe add a link to GroupVIntUtil so that users know what to use instead?

jpountz (approved):

lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
uschindler: I think there's enough explanation in the method using this signature, so theis comment should be enough

uschindler: I added a `@see` to give the explanation why it was deprecated.

I also made one acidentally hidden method public again. I think now its ready! Will merge soon.

uschindler:
Thanks for review, will merge soon!

uschindler:
I added changes and migrate information in main: https://github.com/apache/lucene/commit/839425eae6bc46024970b7316f7e9c07b82d4070

lucene/core/src/java/org/apache/lucene/util/GroupVIntUtil.java
jpountz: Ah sorry, I commented in the wrong place, I meant to say this about `DataInput#readGroupVInt`.

uschindler: Now I understand your intention. The problem is now that `DataInput#readGroupVInt()` throws UnsupportedEx.

I can fix this easily (it delegates to `GroupVIntUtil#readGroupVInt()`). Let me fix this in main and add a comment!

uschindler: The problem is not people implementing the method, it could be that some user code called the method to read a single VInt group! Now this fails with UOE. My proposal above fixes it.

uschindler: Fixed in 10x branch by https://github.com/apache/lucene/commit/d664abb181a8299d4cdf671dd70348095d1a835e

jpountz: Thank you!

PR #15148, opened 2025-09-02
Add support for uint8 distance comparison

mccullocht:
Add distance functions that treat `byte[]` as unsigned and use them in `ScalarQuantizer` code paths. `ScalarQuantizer`
assumes that all math will be unsigned but this can't be true when all 8 bits of a `byte` are used. In all cases where this
is used values are widened to `short` or `int` before any operations occur so we can perform unsigned widening instead.

Fixes are required in a few other places to support unsigned extension, in particular this is required when recalculating
the vector offset as part of merging.

This will be useful for supporting 8 bit OSQ in #15064

Microbenchmark results:
```
Benchmark                                        (size)   Mode  Cnt   Score   Error   Units
VectorUtilBenchmark.binaryCosineScalar             1024  thrpt   15   0.977 ± 0.006  ops/us
VectorUtilBenchmark.binaryCosineVector             1024  thrpt   15   3.682 ± 0.050  ops/us
VectorUtilBenchmark.binaryDotProductScalar         1024  thrpt   15   2.973 ± 0.037  ops/us
VectorUtilBenchmark.binaryDotProductUint8Scalar    1024  thrpt   15   2.976 ± 0.025  ops/us
VectorUtilBenchmark.binaryDotProductUint8Vector    1024  thrpt   15   6.545 ± 0.098  ops/us
VectorUtilBenchmark.binaryDotProductVector         1024  thrpt   15   6.670 ± 0.108  ops/us
VectorUtilBenchmark.binaryHalfByteScalar           1024  thrpt   15   2.982 ± 0.019  ops/us
VectorUtilBenchmark.binaryHalfByteScalarPacked     1024  thrpt   15   2.971 ± 0.029  ops/us
VectorUtilBenchmark.binaryHalfByteVector           1024  thrpt   15  20.898 ± 0.489  ops/us
VectorUtilBenchmark.binaryHalfByteVectorPacked     1024  thrpt   15  17.602 ± 0.045  ops/us
VectorUtilBenchmark.binarySquareScalar             1024  thrpt   15   2.995 ± 0.006  ops/us
VectorUtilBenchmark.binarySquareUint8Scalar        1024  thrpt   15   2.987 ± 0.015  ops/us
VectorUtilBenchmark.binarySquareUint8Vector        1024  thrpt   15   6.715 ± 0.028  ops/us
VectorUtilBenchmark.binarySquareVector             1024  thrpt   15   6.720 ± 0.038  ops/us
```

I ran luceneutil mostly to sanity check on recall, I don't expect significant differences in performance between 7 and 8 bits because packing is the same but we should see a modest recall improvement from having twice as much space to quantize into.

luceneutil, 500k vectors, dot product distance:
```
Results:
 0.843        1.363   1.335        0.980  500000    10     100       32        250     7 bits      0.00      Infinity            0.07             1         1869.59      1832.962      368.118       HNSW
 0.872        1.347   1.321        0.981  500000    10     100       32        250     8 bits      0.00      Infinity            0.07             1         1869.34      1832.962      368.118       HNSW
```

500k vector, squared l2 distance:
```
Results:
 0.905        1.180   1.156        0.979  500000    10     100       32        250     7 bits    145.52       3435.98          146.40             1         1851.79      1832.962      368.118       HNSW
 0.932        1.166   1.147        0.983  500000    10     100       32        250     8 bits    132.39       3776.75          125.74             1         1851.71      1832.962      368.118       HNSW
```

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -54,8 +54,6 @@ Improvements
   SIGSEGV with the old unsafe "mmap hack", which has been replaced by MemorySegments.
   (Uwe Schindler, Robert Muir)
 
-# GITHUB#15148: Add support uint8 distance and allow 8 bit scalar quantization (Trevor McCulloch)
-
 Optimizations
 ---------------------
 * GITHUB#14011: Reduce allocation rate in HNSW concurrent merge. (Viliam Durina)
@@ -120,7 +118,7 @@ New Features
 
 Improvements
 ---------------------
-(No changes)
+# GITHUB#15148: Add support uint8 distance and allow 8 bit scalar quantization (Trevor McCulloch)
 
 Optimizations
 ---------------------
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -54,6 +54,8 @@ Improvements
   SIGSEGV with the old unsafe "mmap hack", which has been replaced by MemorySegments.
   (Uwe Schindler, Robert Muir)
 
+# GITHUB#15148: Add support uint8 distance and allow 8 bit scalar quantization (Trevor McCulloch)
+
 Optimizations
 ---------------------
 * GITHUB#14011: Reduce allocation rate in HNSW concurrent merge. (Viliam Durina)
--- a/lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/VectorUtilBenchmark.java
+++ b/lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/VectorUtilBenchmark.java
@@ -120,6 +120,17 @@ public class VectorUtilBenchmark {
   }
 
   @Benchmark
+  public int binaryDotProductUint8Scalar() {
+    return VectorUtil.uint8DotProduct(bytesA, bytesB);
+  }
+
+  @Benchmark
+  @Fork(jvmArgsPrepend = {"--add-modules=jdk.incubator.vector"})
+  public int binaryDotProductUint8Vector() {
+    return VectorUtil.uint8DotProduct(bytesA, bytesB);
+  }
+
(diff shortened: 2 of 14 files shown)


benwtrent:
Ah, interesting! For sure, with panama vector enabled, I would expect to see zero slow down in Lucene. 

Looking forward to benchmarks :).

One tricky part would be that we don't mix up the `signed bytes` that a user could provide and the `unsigned byte` that we would gather from quantization. 

I think you have this handled with the `signed` boolean you are passing around. Just wanting to call out the concern.

mccullocht:
I suspect that this won't be slower in the scalar path either, as IIUC java will implicitly widen to `int` on any arithmetic operation and processors will typically have both signed and unsigned widen. The `boolean signed` is deliberately not part of any public API to minimize confusion. Hopefully the JVM uses this to monomorphize the functions.

I'm getting 403s downloading the luceneutil testdata today but hopefully I'll be able to get some numbers soon.

benwtrent (approved):
We should have a changes entry. I like how this doesn't really effect public APIs and is only a minor change in our vector comparison code.

mccullocht:
Added to CHANGES for 11. This would probably backport to 10.4 just fine but I also don't know if 10.4 will be happening.

mccullocht:
I tried picking this change into branch_10x and there's only a minor conflict in `PanamaVectorUtilSupport` so I've moved it to 10.4 in CHANGES

benwtrent:
I would assume there would be a 10.4. Even if its to shore things up before another major. We didn't really nail down if there would be another major before 10.3 was cut.

mccullocht:
I think this is ready to merge

benwtrent:
I agree @mccullocht I can merge and back port. Thank you for digging into this :)

PR #15149, opened 2025-09-03
Add bulk-retrieval API to NumericDocValues.

jpountz:
Lucene recently got very good performance improvements by introducing APIs that apply to batches of doc IDs at once: `DocIdSetIterator#intoBitSet`, `PostingsEnum#nextPostings`, `Scorer#nextDocsAndScores` and `SimScorer#score`. This helps better amortize the cost of virtual function calls across many doc IDs, and also apply additional optimizations, e.g. it's more efficient to bulk-iterate set bits in a `FixedBitSet` than to iterate them one-by-one via `FixedBitSet#nextSetBit`.

This PR introduces bulk retrieval for numeric doc values. It is currently only implemented on norms and used to retrieve norms for doc IDs to score, but I tried to design the API in a way that also works for numeric doc values and is sustainable. Specifically, I'm thinking that optimizing the single-valued and dense case should go a very long way, so I did not try to help users retrieve information about which docs have a value or not. In some cases, this is not even needed. E.g. if you want to compute the sum of the values of a field, returning 0 for docs that don't have a value is good. In the event when knowing which docs have a value is important (such as Lucene's `HistogramCollector`), it is still possible to optimize the case when there are long runs of docs with a value with something like below:

```java
void doSomethingWith(int size, int[] docs, NumericDocValues values) {
  if (size > 0 && values.advanceExact(docs[0]) && values.docIDRunEnd() > docs[size - 1]) {
    // All docs are guaranteed to have a value
    long[] longValues = new long[size];
    values.longValues(size, docs, longValues, 0L);
    // do something with the `longValues` array
  } else {
    // use #advanceExact / #longValue directly
    for (int i = 0; i < size; i++) {
      if (values.advanceExact(docs[i])) {
        // do something with values#longValue
      } else {
        // handle the case when docs don't have a value
      }
    }
  }
}
```

### DIFF
--- a/lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/PolymorphismBenchmark.java
+++ b/lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/PolymorphismBenchmark.java
@@ -1,0 +1,191 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.lucene.benchmark.jmh;
+
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@State(Scope.Benchmark)
+@Warmup(iterations = 5, time = 1)
+@Measurement(iterations = 5, time = 1)
+@Fork(
+    value = 1,
+    jvmArgsAppend = {"-Xmx1g", "-Xms1g", "-XX:+AlwaysPreTouch"})
+public class PolymorphismBenchmark {
+
+  public abstract static class IntList {
+
+    public abstract int size();
+
+    public abstract int get(int index);
+
+    public final long sum1() {
+      long sum = 0;
+      for (int i = 0; i < size(); ++i) {
+        sum += get(i);
+      }
+      return sum;
+    }
+
+    public long sum2() {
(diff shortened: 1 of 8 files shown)


lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/PolymorphismBenchmark.java
```diff
@@ -0,0 +1,191 @@
...
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@State(Scope.Benchmark)
+@Warmup(iterations = 5, time = 1)
+@Measurement(iterations = 5, time = 1)
+@Fork(
+    value = 1,
+    jvmArgsAppend = {"-Xmx1g", "-Xms1g", "-XX:+AlwaysPreTouch"})
+public class PolymorphismBenchmark {
```
martijnvg, line 41: Is this benchmark intended to be included? Given that it doesn't benchmark any Lucene classes. 
Or is this benchmark used to proof that invoking super method can help performance? Like is done in norms producer?

martijnvg (approved):
Thanks Adrien, LGTM

lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/PolymorphismBenchmark.java
jpountz: The latter indeed. I suspect that seeing `@Override public void doSomething() { super.doSomething(); }` would raise some eyebrows as being unnecessary, so I wanted to have a benchmark ready that could be used to demonstrate why it helps.

dnhatn (approved):
LGTM. Thanks Adrien!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90NormsProducer.java
```diff
@@ -394,6 +394,14 @@ public long longValue() throws IOException {
             public long longValue() throws IOException {
               return slice.readByte(doc);
             }
+
+            @Override
+            public void longValues(int size, int[] docs, long[] values, long defaultValue)
+                throws IOException {
+              // Delegate to help performance: when the super call inlines, calls to
+              // #advanceExact/#longValue become monomorphic.
```
jainankitk, line 402: I will probably call this _identity/dummy wrapper to help performance_. Also I am assuming that `super.longValues` will always inline, right?

jainankitk (approved):
Nice inclusion. LGTM!

lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90NormsProducer.java
jpountz: I am not sure if your first sentence is a suggestion to update the code comment?

The JVM gives no guarantees about what may inline or not, but I confirmed that this inlines when the `NumericDocValues#longValues` call site is polymorphic in some hand-made benchmarks.

jainankitk: Yeah, it was to update the comment. I was initially confused by the sub/super class, but this is the subclass so `delegate` does make sense.

> The JVM gives no guarantees about what may inline or not

With some of these recent optimizations around inlining, I am wondering how do we know that we are getting the optimal performance or not. Is there a way to instrument the code in a manner that provides signal if a particular optimization is being leveraged or not?

jpountz: I'm not sure. I usually rely on nightly benchmarks, e.g. the fact that we got a good speedup with annotation JH at https://benchmarks.mikemccandless.com/OrHighHigh.html suggests that the `doScore` calls are properly getting inlined in BM25's `BulkSimScorer`. Also nightly benchmarks report whether methods are inlined, compiled or interpreted in the produced profiles: https://benchmarks.mikemccandless.com/2025.09.07.18.04.39.html#profiler_searching_1_cpu, for instance you can find this line `1.18%         184793        org.apache.lucene.search.similarities.BM25Similarity$BM25Scorer#doScore() [Inlined code]
` that further confirms that `doScore` gets inlined.

PR #14979, opened 2025-07-21
Simplify ForDeltaUtil's prefix sum.

jpountz:
I remember benchmarking prefix sums quite extensively, and unrolled loops performed significantly better than their rolled on counterpart, both on micro and macro benchmarks:

```java
private static void prefixSum(int[] arr, int len) {
  for (int i = 1; i < len; ++i) {
    arr[i] += arr[i-1];
  }
}
```

However, I recently discovered that rewriting the loop this way performs much better, and almost on par with the unrolled variant:

```java
private static void prefixSum(int[] arr, int len) {
  int sum = 0;
  for (int i = 0; i < len; ++i) {
    sum += arr[i];
    arr[i] = sum;
  }
}
```

I haven't checked the assembly yet, but both a JMH benchmark and luceneutil agree that it doesn't introduce a slowdown, so I cut over prefix sums to this approach.

PR #15156, opened 2025-09-05
Optimize a bit two more prefix sums.

jpountz:
This applies the same change as #14979 to two more prefix sums. I was not able to measure speedups (or slowdowns) with luceneutil, but I think it's still better to write our prefix sums this way.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
@@ -204,10 +204,11 @@ public final class Lucene103PostingsReader extends PostingsReaderBase {
     }
   }
 
-  static void prefixSum(int[] buffer, int count, long base) {
-    buffer[0] += base;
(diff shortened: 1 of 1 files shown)


gf2121 (approved):

RamakrishnaChilaka:
@jpountz, Lucene104PostingsReader still uses the older prefix-sum pattern at[ line 193](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104PostingsReader.java#L193); I’ll open a small follow-up PR to apply the same tweak, if you agree..

RamakrishnaChilaka:
similar change on [Line 622](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104PostingsReader.java#L622C11-L622C51).

jpountz:
Thanks for noticing! Please go ahead!

dungba88:
@jpountz I also raised https://github.com/apache/lucene/pull/15194 for the other places in code.

PR #15157, opened 2025-09-05
Support JDK 25 in Panama Vectorization Provider

ChrisHegarty:
This commit updates the Vectorization Provider to support JDK 25. The API has not changed so the changes minimally bump the major JDK check, and enable the incubating API during testing.

JDK 25 is past rampdown phase 2.

I tested this locally with:
```
export RUNTIME_JAVA_HOME=/Users/chegar/binaries/jdk-25.jdk/Contents/Home/
export JAVA_HOME=/Users/chegar/binaries/jdk-24.0.2.jdk/Contents/Home/

CI=true ./gradlew check -Ptests.vectorsize="default" -Ptests.forceintegervectors=false 
```

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -184,6 +184,8 @@ New Features
 
 * GITHUB#15110: PostingsDecodingUtil: interchange loops to enable better memory access and SIMD vectorisation. (Ramakrishna Chilaka)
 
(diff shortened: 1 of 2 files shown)


uschindler (approved):
Go for it!

uschindler:
Do we want to respond because you added 10.3.0 ?

uschindler:
Ah I have seen mailing list post. 👍

PR #15161, opened 2025-09-05
chore: Seal BitSet to only allow FixedBitSet and SparseFixedBitSet

AdityaTeltia:
Closes #15038

- Made BitSet a sealed abstract class permitting only FixedBitSet and SparseFixedBitSet.
- Declared SparseFixedBitSet as final to comply with sealed hierarchy.
- Removed test-only subclass JavaUtilBitSet which was no longer permitted.
- Refactored BaseBitSetTestCase to introduce an abstract fromJavaUtilBitSet(...) factory.
- Implemented fromJavaUtilBitSet(...) in TestFixedBitSet and TestSparseFixedBitSet to build the
  correct BitSet implementation from a java.util.BitSet for testing.
  
This ensures that only Lucene’s optimized implementations of BitSet are allowed, preventing 
accidental performance regressions from custom subclasses, while keeping existing tests valid.


benwtrent:
I am not sure this can go in v10. Possibly v11 (though I am gonna miss being able to see "nextBitSet" and such for Bits interfaces when possible...)

benwtrent:
10.3 is cut, so cannot go there. Possibly 10.4. But I would expect this to be a v11 change.

lucene/core/src/java/org/apache/lucene/util/BitSet.java
```diff
@@ -24,7 +24,8 @@
  *
  * @lucene.internal
  */
-public abstract class BitSet implements Bits, Accountable {
+public abstract sealed class BitSet implements Bits, Accountable
```
jpountz, line 30: Can you add a small comment explaining that we're doing this to help call sites of methods on the `BitSet` class be bimorphic at most, to help with performance?

lucene/CHANGES.txt
```diff
@@ -215,6 +215,8 @@ Improvements
 * GITHUB#15063: Use optimistic-with-checking KNN Query strategy for `DiversifyingChildren*` queries.
   (Ben Trent)
 
+* GITHUB#15038: BitSet is now a sealed abstract class permitting only FixedBitSet and SparseFixedBitSet.
+
```
jpountz: Let's move this to 11.0?

lucene/test-framework/src/java/org/apache/lucene/tests/util/BaseBitSetTestCase.java
```diff
@@ -75,7 +73,7 @@ protected void assertEquals(BitSet set1, T set2, int maxDoc) {
   public void testCardinality() throws IOException {
     final int numBits = 1 + random().nextInt(100000);
     for (float percentSet : new float[] {0, 0.01f, 0.1f, 0.5f, 0.9f, 0.99f, 1f}) {
-      BitSet set1 = new JavaUtilBitSet(randomSet(numBits, percentSet), numBits);
+      BitSet set1 = fromJavaUtilBitSet(randomSet(numBits, percentSet), numBits);
```
jpountz: E.g. doing just `java.util.BitSet set1 = randomSet(numBits, percentSet);` should be good enough here since we're only comparing cardinalities?

lucene/test-framework/src/java/org/apache/lucene/tests/util/BaseBitSetTestCase.java
```diff
@@ -65,6 +61,8 @@ static java.util.BitSet randomSet(int numBits, float percentSet) {
     return randomSet(numBits, (int) (percentSet * numBits));
   }
 
+  protected abstract T fromJavaUtilBitSet(java.util.BitSet set, int numBits);
```
jpountz, line 64: I worry that copying from a `java.util.BitSet` defeats the purpose of the test, since most tests would be comparing two same implementations of `BitSet` instead of `JavaUtilBitSet` vs. a `BitSet` implementation. Can we try to refactor tests to work directly against a `java.util.BitSet` without copying to a `BitSet`?

lucene/core/src/java/org/apache/lucene/util/BitSet.java
AdityaTeltia: added small comment explaining this.

lucene/test-framework/src/java/org/apache/lucene/tests/util/BaseBitSetTestCase.java
AdityaTeltia: tried refactoring it using java.util.Bitset.
Had to hack through two issues.
1. `java.util.BitSet.nextSetBit()` returns  `-1` when there are no more set bits while lucene Bitset return `DocIdSetIterator.NO_MORE_DOCS` i.e. 2147483647.
2. `java.util.BitSet.clear()` throws error if from > to while lucene does nothing.

AdityaTeltia:
@jpountz  @benwtrent  Can we merge this now?

PR #15162, opened 2025-09-05
build: require Python 3.12+, add smoke-test Make target

### LINKED ISSUE #14556: Smoke tester requiring Python 3.12+
#14326 added a [line in scriptutil](https://github.com/apache/lucene/blob/92d79d47cbd238137ec136f6947c0c9e86003ce0/dev-tools/scripts/scriptutil.py#L26) that imports `override` from `typing`, which was [introduced in Python 3.12](https://github.com/python/cpython/issues/101561).

Running the smoke tester with 3.11, errors out like so:
```
Traceback (most recent call last):
  File "/local/home/voditas/ws/open/lucene/dev-tools/scripts/smokeTestRelease.py", line 40, in <module>
    import scriptutil
  File "/local/home/voditas/ws/open/lucene/dev-tools/scripts/scriptutil.py", line 26, in <module>
    from typing import Self, override
ImportError: cannot import name 'override' from 'typing' (/usr/local/lib/python3.11/typing.py)
```

Maybe we can make it clearer what version Python is required?

rmuir:
Update Makefile and README to require Python 3.12 or newer, allow configurable Python executable, and add a 'smoke-test' target for smoketester. Update setup instructions and usage examples accordingly.

Closes #14556

Example error (with minimum bumped to 3.14):

Screen_Shot_2025-09-05_at_14 40 29

If we can just use the `make` command, it will make life easier than user having to install python dependencies. It also won't mess up any other python stuff on their computer because it uses a virtual environment, etc.

### DIFF
--- a/dev-tools/scripts/Makefile
+++ b/dev-tools/scripts/Makefile
@@ -13,11 +13,14 @@ export PYTHONDONTWRITEBYTECODE=true
 # don't self-check, don't ask questions
 PIP_INSTALL_ARGS=--disable-pip-version-check --no-input --upgrade
 
+# override to specific python executable if you need
+PYTHON?=python3
+
 # venv with dependencies in the standard location
 VENV=${PWD}/.venv
 
 # don't behave strangely if these files exist
-.PHONY: lint format reformat autofix ruff ruff-fix pyright env clean
+.PHONY: lint format reformat autofix ruff ruff-fix pyright env clean smoke-test
 
 # list of directories we check
 SOURCES=$(wildcard *.py)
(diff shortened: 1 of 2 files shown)


dweiss (approved):
There should be also a related change to run-nightly-smoketester.yml, right? We currently call it explicitly:
```
python3 -u dev-tools/scripts/smokeTestRelease.py
```

dweiss:
Ok, nevermind - I think it's fine even if we don't use venv there in that workflow, the container should have an up to date version of all dependencies. I'll follow up if it breaks.

rmuir:
I think it would be an improvement to call it in the action. The goal here was just to present a less error-prone way of running the smoke-tester, so it is ideal to test that way too:
* doesn't install a bunch of python libs globally on your machine which may mess stuff up
* give better error messages if something is wrong
* possibility to improve later to make it more reliable: e.g. use lockfile

dweiss:
Yes, it's fine with me - feel free to merge, I'll follow-up with workflows changes.

PR #15165, opened 2025-09-08
 Optimize PForUtil.encode() with histogram-based bit selection

RamakrishnaChilaka:
PForUtil: Replace IntHeap with histogram approach for OptPFOR

* Use `int[32]` histogram to track bit width distribution
* Add early termination when exceptions exceed `MAX_EXCEPTIONS`.
* Maintain backward compatibility with existing format.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene103/PForUtil.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene103/PForUtil.java
@@ -55,25 +55,18 @@ final class PForUtil {
       maxBitsRequired = Math.max(maxBitsRequired, bits);
     }
 
-    int bestCost = Integer.MAX_VALUE;
-    int patchedBitsRequired = maxBitsRequired;
-    int numExceptions = 0;
-
     // We store patch on a byte, so we can't decrease bits by more than 8
     final int minBits = Math.max(0, maxBitsRequired - 8);
     int cumulativeExceptions = 0;
+    int patchedBitsRequired = maxBitsRequired;
+    int numExceptions = 0;
+
     for (int b = maxBitsRequired; b >= minBits; --b) {
-      // Early termination if too many exceptions
-      if (cumulativeExceptions > MAX_EXCEPTIONS) break;
-
-      // storage cost
-      int cost = (ForUtil.BLOCK_SIZE * b) + (cumulativeExceptions << 4);
-      if (cost < bestCost) {
-        bestCost = cost;
-        patchedBitsRequired = b;
-        numExceptions = cumulativeExceptions;
+      if (cumulativeExceptions > MAX_EXCEPTIONS) {
+        break;
       }
-
+      patchedBitsRequired = b;
+      numExceptions = cumulativeExceptions;
       cumulativeExceptions += histogram[b];
     }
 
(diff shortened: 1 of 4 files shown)


RamakrishnaChilaka:
As it's a marginal improvement and code quality change, should we skip change log? I can add change log if you prefer.

jpountz:
I would skip the changelog too for such a change, it has no user-facing impact? I have been wondering about this bit of code recently as well, and I wonder if we should swap the `(Int|Long)Heap` with an `int[32]` that stores the number of values for each required number of bits per value?

RamakrishnaChilaka:
Thanks for the review @jpountz.

> I would skip the changelog too for such a change, it has no user-facing impact?

Yes, there's no user-facing impact.

> I have been wondering about this bit of code recently as well, and I wonder if we should swap the (Int|Long)Heap with an int[32] that stores the number of values for each required number of bits per value?

I agree.
Currently we use `IntHeap`/`LongHeap` to extract the top 8 (7 exceptions + 1 max value (other than the exceptions) from the supplied ints) elements and compare their required bits. However, the max_exception value is static.

Using an int[32] histogram of values by bit width would simplify the logic.
In addition, it could give us flexibility to make the number of exceptions dynamic rather than fixed, which might lead to better compression in some cases. However, encoding might become slower as compared to now due to additional work of computing bpv for all the supplied ints.

jpountz:
What do you think of moving to an histogram of the number of bits per value in this PR?

RamakrishnaChilaka:
I’ll extend this PR to replace the `(Int|Long)Heap` logic with an `int[32]` histogram that tracks the count of values by required bits. I’ll also update the PR title and description once I push the changes.

RamakrishnaChilaka:
> What do you think of moving to an histogram of the number of bits per value in this PR?

Please re-check the PR @jpountz, I have updated PR title and description.

I have pushed the following changes.
* Use int[32] histogram to track bit width distribution
* Implement cost-based optimization: (block_size * base_bits) + (exceptions * 16)
* Add early termination when exceptions exceed MAX_EXCEPTIONS.
* Maintain backward compatibility with existing format.

lucene/core/src/java/org/apache/lucene/codecs/lucene103/PForUtil.java
```diff
@@ -46,34 +45,39 @@ static boolean allEqual(int[] l) {
...
+    // We store patch on a byte, so we can't decrease bits by more than 8
+    final int minBits = Math.max(0, maxBitsRequired - 8);
+    int cumulativeExceptions = 0;
+    for (int b = maxBitsRequired; b >= minBits; --b) {
+      // Early termination if too many exceptions
+      if (cumulativeExceptions > MAX_EXCEPTIONS) break;
+
+      // storage cost
+      int cost = (ForUtil.BLOCK_SIZE * b) + (cumulativeExceptions << 4);
+      if (cost < bestCost) {
```
jpountz: This condition is always true, right? Because going to the previous number of bits per value saves 128 bits (1 bit per integer)? And in the worst-case scenario, you go from 0 to MAX_EXCEPTIONS exceptions, which increases storage for exceptions by 7*16=112 bits?

RamakrishnaChilaka: I agree that it is always true i.e., cost will never swing with 7 Exceptions. However, I have initially added it as we might increase NUM_EXCEPTIONS beyond 7 (with update to the on-disk codec to store num_exceptions in more than 3 bits). I have removed the cost logic in the latest commit.

jpountz (approved):

PR #15169, opened 2025-09-09
Add a new codec to implement OSQ for 4 and 8 bit quantized vectors

mccullocht:
Unlike the existing ScalarQuantizer selects a mode based on an enum to quantize to unsigned bytes or
packed nibbles using the same packing scheme as the existing scalar quantized codec. Seven bits is also
supported for anyone interested in backward compatibility, but this setting is discouraged.

This is separate from `Lucene102BinaryQuantizedVectorsFormat` as we need a larger value to store
the component sum for each vector owing to larger quantized values.

This closes #15064 

luceneutil benchmark results. OSQ results are bits -4 and -8.
```
Results:
recall  latency(ms)  netCPU  avgCpuCount     nDoc  topK  fanout  maxConn  beamWidth  quantized  index(s)  index_docs/s  force_merge(s)  num_segments  index_size(MB)  vec_disk(MB)  vec_RAM(MB)  indexType
 0.875        0.858   0.855        0.996  1000000    10     100       32        250    -4 bits    203.69       4909.49          168.11             1         3349.44      3311.157      381.470       HNSW
 0.954        1.222   1.217        0.996  1000000    10     100       32        250    -8 bits    333.87       2995.20          193.90             1         3717.10      3677.368      747.681       HNSW
 0.450        1.881   1.816        0.965  1000000    10     100       32        250     4 bits    510.04       1960.65          285.58             1         3346.30      3299.713      370.026       HNSW
 0.928        1.245   1.241        0.997  1000000    10     100       32        250     8 bits    325.14       3075.58          207.41             1         3705.70      3665.924      736.237       HNSW
```


lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorScorer.java
```diff
@@ -0,0 +1,204 @@
...
+        switch (scalarEncoding) {
+          case UNSIGNED_BYTE -> VectorUtil.uint8DotProduct(quantizedQuery, quantizedDoc);
+          case PACKED_NIBBLE -> VectorUtil.int4DotProductPacked(quantizedQuery, quantizedDoc);
+        };
+    OptimizedScalarQuantizer.QuantizationResult indexCorrections =
+        targetVectors.getCorrectiveTerms(targetOrd);
+    float scale = SCALE_LUT[scalarEncoding.getBits() - 1];
+    float x1 = indexCorrections.quantizedComponentSum();
+    float ax = indexCorrections.lowerInterval();
+    // Here we assume `lx` is simply bit vectors, so the scaling isn't necessary
```
benwtrent:
```suggestion
    // Here we must scale according to the bits
```

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorsFormat.java
```diff
@@ -0,0 +1,206 @@
...
+
+  /**
+   * Allowed encodings for scalar quantization.
+   *
+   * <p>This specifies how many bits are used per dimension and also dictates packing of dimensions
+   * into a byte stream.
+   */
+  public enum ScalarEncoding {
+    /** Each dimension is quantized to 8 bits and treated as an unsigned value. */
+    UNSIGNED_BYTE(0, (byte) 8),
```
benwtrent: for uniformity with the old format, could we also allow `7` bits?

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorsWriter.java
```diff
@@ -0,0 +1,858 @@
...
+
+  private void writeField(
+      FieldWriter fieldData, float[] clusterCenter, int maxDoc, OptimizedScalarQuantizer quantizer)
+      throws IOException {
+    // write vector values
+    long vectorDataOffset = vectorData.alignFilePointer(Float.BYTES);
+    writeVectors(fieldData, clusterCenter, quantizer);
+    long vectorDataLength = vectorData.getFilePointer() - vectorDataOffset;
+    float centroidDp =
+        !fieldData.getVectors().isEmpty() ? VectorUtil.dotProduct(clusterCenter, clusterCenter) : 0;
```
benwtrent, line 178:
```suggestion
    float centroidDp =
        fieldData.getVectors().isEmpty() ? 0 : VectorUtil.dotProduct(clusterCenter, clusterCenter);
```
not a big deal, but I have found `!` or negatives in general to be rife with silly bugs.

benwtrent (commented):
This is awesome! Thank you for going through the slog. Doing format changes is always a challenge and requires a ton of ceremony. 

I will need to read through this a couple of times to grok all of it (but most of it seems pretty standard for a knn format)

I realize we support unsigned 8 byte now, but for BWC, I would still like to provide 7 bit if possible. 

With this change, we should also move all existing quantized formats to BWC. But that would be yet another thousand+ LOC. So, maybe that can be done in a follow up.

benwtrent:
It is interesting to me that `int4` in the new format has such better latency as well! 

I wonder why that is? Is it simply because the scoring quality is higher, and thus we can exit searching the graph more quickly?

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorsFormat.java
mccullocht: Added a SEVEN_BITS enum, but documented that 8 bits is preferred.

mccullocht (commented):
> With this change, we should also move all existing quantized formats to BWC. But that would be yet another thousand+ LOC. So, maybe that can be done in a follow up.

Marked the old codecs as deprecated. I'd prefer not to do backwards codecs here, this change is already much larger than I'd like but I couldn't figure out how to factor it into smaller pieces.

> It is interesting to me that int4 in the new format has such better latency as well!
> I wonder why that is? Is it simply because the scoring quality is higher, and thus we can exit searching the graph more quickly?

I re-ran the luceneutil changes because I guess it had partially finished last time. Indexing performance is also _much_ better. I think of this in terms of lossy score compression -- for example trivial bit quantization and hamming distance produces just (dimensions + 1) possible scores, so it can be difficult to distinguish between results in some cases. It could be that the old quantizer produces fewer vector representations and output scores.

More generally I think it would be useful to track and surface `KnnCollector` stats in indexing and search paths. Being able to distinguish between "comparisons are faster" and "comparisons are fewer" would be helpful for analyzing this, and also other algorithmic and data layout changes (would be really curious to see this for binary partitioning reordering).

benwtrent:
> Marked the old codecs as deprecated. I'd prefer not to do backwards codecs here, this change is already much larger than I'd like but I couldn't figure out how to factor it into smaller pieces.

Yeah, I am cool with that. Format changes are always huge!

> More generally I think it would be useful to track and surface KnnCollector stats in indexing and search paths. Being able to distinguish between "comparisons are faster" and "comparisons are fewer" would be helpful for analyzing this, and also other algorithmic and data layout changes (would be really curious to see this for binary partitioning reordering).

We don't really track this during indexing for sure. But you can expose the vector comparisons in Lucene Util.

mccullocht:
Average visited count in the query path actually is exposed in luceneutil today, it just appears in the iteration summary and not the overall summary. TIL. I've extracted it for my most recent run here:
```
-4 3794
-8 3857
 4 4294
 8 3849
```

4 bit is doing ~10% more comparisons than 8 bit for the same fanout. More work in 4 bit but doesn't explain the size of the win in OSQ4. Workload CPU usage and latency is mostly driven scoring costs so let's start by looking at microbenchmark results for dot product on the same hardware:
```
VectorUtilBenchmark.binaryDotProductVector         128  thrpt   15   47.741 ±  0.502  ops/us
VectorUtilBenchmark.binaryDotProductVector         256  thrpt   15   26.198 ±  0.608  ops/us
VectorUtilBenchmark.binaryDotProductVector         300  thrpt   15   22.857 ±  0.130  ops/us
VectorUtilBenchmark.binaryDotProductVector         512  thrpt   15   13.864 ±  0.443  ops/us
VectorUtilBenchmark.binaryDotProductVector         702  thrpt   15   10.003 ±  0.205  ops/us
VectorUtilBenchmark.binaryDotProductVector        1024  thrpt   15    6.795 ±  0.098  ops/us
VectorUtilBenchmark.binaryHalfByteVectorPacked     128  thrpt   15   72.590 ±  1.089  ops/us
VectorUtilBenchmark.binaryHalfByteVectorPacked     256  thrpt   15   50.962 ±  0.197  ops/us
VectorUtilBenchmark.binaryHalfByteVectorPacked     300  thrpt   15   39.587 ±  0.159  ops/us
VectorUtilBenchmark.binaryHalfByteVectorPacked     512  thrpt   15   31.660 ±  0.187  ops/us
VectorUtilBenchmark.binaryHalfByteVectorPacked     702  thrpt   15   22.251 ±  0.135  ops/us
VectorUtilBenchmark.binaryHalfByteVectorPacked    1024  thrpt   15   18.117 ±  0.129  ops/us
```

At 702 dimensions (next closest to our test data set) half byte is about twice as fast. The performance different between -4 and -8 makes sense with this context. I don't know why 4 is so slow, these numbers suggest it shouldn't be worse than 8 yet somehow it is 🤷. Not sure this is worth figuring out.

benwtrent (approved):
Really great stuff.

The existing scalar formats can go to the backwards codecs in a separate PR.

mccullocht:
I confused myself about the repository state because I'm still using my fork so I broke the build due to a conflict with #14863  :/. Fix in #15199

I will also backport this to 10.4 tonight.

mikemccand:
> Average visited count in the query path actually is exposed in luceneutil today, it just appears in the iteration summary and not the overall summary. TIL.

Oooh -- maybe open luceneutil issue to carry this into the overall summary?  It's an important metric (number of nodes visited/explored by the query) for our HNSW impl.

mccullocht:
I have an open PR to support 8 bit, I'll expand that to use the new OSQ codecs and the print the visited count

mccullocht:
@mikemccand https://github.com/mikemccand/luceneutil/pull/461 prints the overall summary and includes support for this codec in luceneutil.

mikemccand:
Thank you @mccullocht!  This is an exciting improvement for 10.4.0.

navneet1v:
@mccullocht there was a parameter called as `confidence_interval` in the older format how do we map it to new format now? I don't see anything related to that.

mccullocht:
The confidence interval parameter is no longer needed; the new format does not "train" on a sample of the documents. Encoding is vector local except for the center, which is the mean vector in the segment

navneet1v:
> The confidence interval parameter is no longer needed; the new format does not "train" on a sample of the documents. Encoding is vector local except for the center, which is the mean vector in the segment

@mccullocht how as a consumer of Lucene I am supposed to migrate to new format and still keep on getting consistent results. One way I can think of is copy the code for BWC which was removed. and then for all new indices which are getting created start using the new format.

PR #15171, opened 2025-09-09
Add bulk scoring option to VectorScorer interface

benwtrent:
Now that we have bulk scoring for our RandomScorer's, let's enable it for our VectorScorer interface. 

I only implemented it for float32 right now, as that is the only place where we are actually doing bulk scoring. But once the interface is in place, I imagine it can be easily implemented elsewhere.

This shows a marginal performance improvement:

baseline
```
recall  latency(ms)  netCPU  avgCpuCount     nDoc  topK  fanout  maxConn  beamWidth  quantized  index(s)  index_docs/s  force_merge(s)  num_segments  index_size(MB)  vec_disk(MB)  vec_RAM(MB)  indexType
 1.000        0.880   0.760        0.864  1000000   500     550       16        100         no      0.00      Infinity            0.06             0            0.00         0.000        0.000       HNSW
```
candidate:
```
recall  latency(ms)  netCPU  avgCpuCount     nDoc  topK  fanout  maxConn  beamWidth  quantized  index(s)  index_docs/s  force_merge(s)  num_segments  index_size(MB)  vec_disk(MB)  vec_RAM(MB)  indexType
 1.000        0.780   0.640        0.821  1000000   500     550       16        100         no      0.00      Infinity            0.06             0            0.00         0.000        0.000       HNSW
```

But I think we should consider unlocking this for all our "rescorer" interfaces as well. This will likely require more API changes (e.g. maybe add a "rescorerVectorScorer" function to FloatVectorValues or something...).


lucene/core/src/java/org/apache/lucene/search/VectorScorer.java
```diff
@@ -38,4 +40,37 @@ public interface VectorScorer {
...
+      }
+      buffer.size = size;
+      if (liveDocs != null) {
+        buffer.apply(liveDocs);
+      }
+    };
+  }
+
+  interface Bulk {
+    void nextDocsAndScores(int nextCount, Bits liveDocs, DocAndFloatFeatureBuffer buffer)
```
vigyasharma: Let's add a docstring to describe what these params do, and that the final doc id and score will be stored in the `buffer`?

lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -309,21 +309,24 @@ protected TopDocs exactSearch(
...
         break;
       }
-      assert vectorIterator.docID() == doc;
-      float score = vectorScorer.score();
-      if (score > topDoc.score) {
-        topDoc.score = score;
-        topDoc.doc = doc;
-        topDoc = queue.updateTop();
+      // iterator already takes live docs into account
+      bulkScorer.nextDocsAndScores(64, null, buffer);
```
vigyasharma: nit: move to a final constant?

vigyasharma (approved):

PR #15176, opened 2025-09-10
Adds new simplified rescorer API to [Float|Byte]VectorValues

benwtrent:
Rescoring against a vector field should be simple and allow for lower level bulk scoring optimizations (e.g. scoring floating point values off heap, possibly bulk prefetching, etc.).

However, the current API always returns a scorer against the fastest fidelity vectors, not the most accurate. 

I thought a simple `rescorer` method that also just used the `VectorScorer` object would suffice. While our `DoubleValues` logic cannot take advantage of bulk scoring yet, it might be able to one day.

The other API idea I had in mind was adding something like `VectorScorer#asRescorer` instead. But it didn't seem any better than just adding a simple hook on our `KnnVectorValues` implementations.


tteofili (approved):

mccullocht:
Oh this should be nice for implementing two level lvq or for scoring an f32 vector directly against the quantized representation

PR #15183, opened 2025-09-12
Fix off by one error in Lucene99MemorySegmentFloatVectorScorer

mccullocht:
This was introduced in #15021 where we don't correctly account the max score when there are 3 remaining
vectors in a bulk scoring call.

Fixed #15180


mccullocht:
FYI I don't believe CHANGES are necessary here as the problematic PR has been in main for only a few days and hasn't been released.

lucene/core/src/java24/org/apache/lucene/internal/vectorization/Lucene99MemorySegmentFloatVectorScorer.java
```diff
@@ -114,7 +114,7 @@ public float bulkScore(int[] nodes, float[] scores, int numNodes) throws IOExcep
       }
       if (remaining > 2) {
         scores[i + 2] = normalizeRawScore(scratchScores[2]);
-        maxScore = Math.max(maxScore, scores[i + 1]);
+        maxScore = Math.max(maxScore, scores[i + 2]);
```
benwtrent, line 117: Sneaky!!!!

benwtrent (approved):
Hurrah for randomized tests!

PR #15194, opened 2025-09-17
Optimize prefix sum

dungba88:
### Description

Follow-up of https://github.com/apache/lucene/pull/15156, more optimizations for other places in codebase.


kaivalnp:
LGTM overall, do you have an estimate of performance impact? (JMH or luceneutil)

Issue #15114, opened 2025-08-22
Drop owasp tasks and plugins from the build


dweiss:
### Description

I do have an API key for owasp but the reality is that it takes _forever_ to check. I think there are better tools out there now (dependabot, etc.) and realistically I don't believe anybody is running this owasp support in Lucene. 

I plan to remove this code without a replacement from main. If anybody disagrees, please shout out.

rmuir:
It is outdated by dependabot for the `main` branch. 

We have a bot that tells us what is out of date: https://github.com/apache/lucene/pulls?q=is%3Apr+is%3Aopen+dependabot

If we want to improve dependency analysis, IMO a better path would be to look into "dependency submission API" so that github's dependency graph knows about all of our dependencies. 

Currently it doesn't know the java/gradle ones without us submitting the data (maybe there's a simple gradle plugin for that?), it only knows about `pip` and `actions` ecosystems: https://github.com/apache/lucene/network/dependencies

Anyway its just an idea, we could make another issue for that and maybe someone wants to take it on. personally i have not tackled the dependency submission API with github yet, but if we fixed this then we'd have more security features from github working for us. Its pretty cool: it really does work as a "graph", with alerts on your transitive deps that make sense, SBOM generation, etc.

And a warning: maybe in the future Github improves their gradle processing to just "work", like it does for pip and actions, without any effort on our part. Then we'd avoid any dependency submission logic.

dweiss:
I did play with automated dependency submission but this created a huge graph of dependencies (from all the plugins involved, etc.). I didn't find a way to limit this to just dependencies. Shouldn't be too difficult though.

PR #15195, opened 2025-09-17
Move more gradle plugins to java. Remove Apache Rat dependency.

dweiss:
This moves more gradle script plugins to plain Java, also updating some of them to not use deprecated APIs.

It also removes apache rat dependency (rewritten to a custom, incremental task, #15185). I've removed owasp support entirely (#15114).

I've removed the code that checks that the license precedes opening tag in XMLs. I don't see the benefit of doing this here. Maybe we can do this on the CI with ast-grep (it doesn't support xml out of the box though).


build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/licenses/CheckLicensesTask.java
```diff
@@ -0,0 +1,297 @@
...
+ * license patterns. Files that don't have any permitted licenses will trigger an error.
+ */
+@CacheableTask
+public abstract class CheckLicensesTask extends DefaultTask {
+  /** The default number of leading characters scanned in each file. */
+  private static final Integer DEFAULT_SCANNED_HEADER = 1024;
+
+  private record LicenseFamily(String code, String name, Predicate<String> matcherPredicate) {}
+
+  static final List<LicenseFamily> LUCENE_ACCEPTED_LICENSES =
```
dweiss, line 64: This is the list of allowed licenses and their fixed-string detection patterns.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/licenses/CheckLicensesTask.java
```diff
@@ -0,0 +1,297 @@
...
+    String key = root.relativize(p).toString();
+    if (File.separatorChar != '/') {
+      key = key.replace(File.separatorChar, '/');
+    }
+    return key;
+  }
+
+  private LicenseFamily detectLicense(File file, StringBuilder buffer, char[] scratch)
+      throws IOException {
+    // I assume all files are in UTF8... This is verified elsewhere (eclint).
```
dweiss, line 256: I assume all files are in UTF8... This is verified elsewhere (eclint)

dweiss:
Before:
image
After:
image

It's also incremental so once it runs, the cost of subsequent checks is close to zero.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/licenses/CheckLicensesPlugin.java
```diff
@@ -0,0 +1,126 @@
...
+  private void configureCheckLicenses(CheckLicensesTask task) {
+    Project project = task.getProject();
+
+    task.getReportFile().set(project.getLayout().getBuildDirectory().file("licenses-report.txt"));
+    task.getFiles()
+        .from(
+            project.fileTree(
+                ".",
+                tree -> {
+                  // Exclude build outputs, ide files, .git.
```
rmuir, line 57: In the future could we instead please consider using `git ls-files` for this?

I am afraid of when these long long lists of hardcoded excludes fall out of sync with .gitignore files, then checks become slow, or have false failures, or both.

I am only referring to .git/build/etc exclusions which should not be necessary. In 2025 all checkers should respect .gitignore :) We should only have the "real" excludes here particular to the specific check (things in source control)

rmuir (approved):

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/licenses/CheckLicensesPlugin.java
rmuir: I think the thing i am highly suspicious of: is this `project.fileTree`. Potential forbidden-apis candidate.

Some old build processes got slower over months/years in exactly this way. This can go unnoticed, for example if they only **descend** into a python venv, because there are often thousands and thousands of directories files in there. They don't have to actually match any of the files to slow everything down, that's a big part of the trap.

Maybe there is a gradle or java plugin we could use that behaves like https://docs.rs/ignore/latest/ignore/ or we can just use `git ls-files`.

Sorry for the rant, I see this as the number one cause of slow checkers these days, and it is easy to avoid.

dweiss: So there are two downsides I see - one is that it'd no longer work on a source-only package (without .git). The other is that when you run a check but haven't added your files to git yet, it won't see them. I understand what you're after but it's a double edged sword - you then need to remember to git add (or commit) before you run the checks.

Definitely worth a thought though.

rmuir: First problem is solved by passing some additional parameters. For example, try `git ls-files -c -o --exclude-standard`.

rmuir: Second problem is solved by running `git init` before running the checks. It hurts nothing if there is already a repo initialized (and documents this fact)

dweiss: Makes sense - thanks! I think it's doable but I'll keep it for another patch. Makes sense to me though - reuse .gitignore as the source of truth about versioned files.

dweiss: #15200

PR #15200, opened 2025-09-18
Use git-ls-files as the source of file paths for license checks.

dweiss:
Rob's idea. Works quite well, actually. I couldn't pinpoint why the set of paths is slightly different from before... gradle does have a "default exclusion" pattern set so it's possible it excluded something on its own (?).

### DIFF
--- a/build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/gitinfo/GitFileListValueSource.java
+++ b/build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/gitinfo/GitFileListValueSource.java
@@ -22,21 +22,14 @@ import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
 import javax.inject.Inject;
-import org.apache.tools.ant.taskdefs.condition.Os;
 import org.gradle.api.GradleException;
-import org.gradle.api.file.DirectoryProperty;
 import org.gradle.api.provider.ValueSource;
-import org.gradle.api.provider.ValueSourceParameters;
-import org.gradle.api.services.BuildServiceParameters;
 import org.gradle.process.ExecOperations;
 import org.slf4j.LoggerFactory;
 
 /** Returns a list of versioned and non-versioned (but not ignored) git files. */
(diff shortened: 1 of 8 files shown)


build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/gitinfo/GitFileListValueSource.java
```diff
@@ -0,0 +1,91 @@
...
+                        "-o",
+                        // apply .gitignore exclusions
+                        "--exclude-standard",
+                        // don't quote paths, 0-terminate strings.
+                        "-z");
+                  });
+      out.flush();
+
+      // Assume the output is UTF-8, even if it has 0 eols.
+      String gitOutput = baos.toString(StandardCharsets.UTF_8);
```
dweiss, line 66: This seems... correct, right? If there are broken paths (non-utf8 characters) in git then we're screwed but it doesn't seem like the place to do any sophisticated recovery here...

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/licenses/CheckLicensesPlugin.java
```diff
@@ -48,79 +53,96 @@ public void apply(Project project) {
...
+            "**/*.bin",
+            "**/*.brk",
+            "**/*.bz2",
+            "**/*.dat",
+            "**/*.gif",
+            "**/*.gz",
+            "**/*.png",
+            "**/*.svg",
+            "**/*.xls",
+            "**/*.zip",
```
rmuir, line 78: as a followup any of these we have, maybe we can configure in .gitattributes. at least `git grep` will then recognize it as binary, and possibly github. Especially useful for .svg which isnt "really" binary but is annoying AF in search results and diffs.

rmuir (approved):
thanks!

dweiss:
https://jenkins.thetaphi.de/job/Lucene-main-Windows/16867/consoleText

Seems like something is not working on windows after all. I'll revert this and reapply.

PR #15210, opened 2025-09-22
Introduce support for pruning and skipping to FirstPassGroupingCollector

alexmm-amzn:
### Description

Extends the `FirstPassGroupingCollector` to support pruning (for numeric sort fields using `competitiveIterator`) and skipping of non-competitive documents (for relevance score sorting using `Scorable#setMinCompetitiveScore`).

Both optimizations are enabled automatically, thereby reducing the hit count of the collector if circumstances allow.

@jainankitk Are we fine with enabling this by default, or do we need this configurable (e.g. configurable hit threshold)?

Benchmark results using `luceneutils` for the `TermBGroup1M` scenario (combines first and second pass grouping) using a modified `wikimedium.10M.nostopwords.tasks` job. This scenario uses sort by relevance score.

```
> grep TermBGroup1M tasks/wikimedium500.tasks > tasks/wikimedium.10M.nostopwords.tasks
> python src/python/localrun.py -source wikimediumall 
```

Running on `m6a.2xlarge` using Corretto 24:

```
                        PKLookup      161.09     (12.8%)      156.56     (11.2%)   -2.8% ( -23% -   24%) 0.460
                    TermBGroup1M       11.47     (14.8%)       13.44     (13.7%)   17.1% (  -9% -   53%) 0.000
```

=> ~17% overall performance improvement (first+second pass).

@jpountz I'm getting some rare test failures for `TestGrouping` caused by the `assert canSetMinCompetitiveScore` assertion in `AssertingScorer#setMinCompetitiveScore`, even though the `FirstPassGroupingCollector` uses `ScoreMode.TOP_SCORES` in all configurations when it calls `Scorable#setMinCompetitiveScore`. Is this a known issue?

Reproduce with: `gradlew test --tests TestGrouping.testRandom -Dtests.seed=EC2EC279F564DD82 -Dtests.locale=de-AT -Dtests.timezone=America/St_Thomas -Dtests.asserts=true -Dtests.file.encoding=UTF-8`

edit//Seems to be caused by the `Weight` that gets instantiated by the unit tests with either `ScoreMode.COMPLETE` or `ScoreMode.COMPLETE_NO_SCORES` regardless of the actual collectors. I updated the code to instantiate a new `Weight` instance for every collector that is in line with the collector `ScoreMode`.

### DIFF
--- a/lucene/grouping/src/java/org/apache/lucene/search/grouping/FirstPassGroupingCollector.java
+++ b/lucene/grouping/src/java/org/apache/lucene/search/grouping/FirstPassGroupingCollector.java
@@ -50,9 +50,10 @@ public class FirstPassGroupingCollector<T> extends SimpleCollector {
   private final LeafFieldComparator[] leafComparators;
   private final int[] reversed;
   private final int topNGroups;
-  private final boolean needsScores;
   private final HashMap<T, CollectedSearchGroup<T>> groupMap;
   private final int compIDXEnd;
+  private final ScoreMode scoreMode;
+  private final boolean canSetMinScore;
 
   // Set once we reach topNGroups unique groups:
   /**
@@ -62,6 +63,9 @@ public class FirstPassGroupingCollector<T> extends SimpleCollector {
 
   private int docBase;
   private int spareSlot;
+  private Scorable scorer;
+  private int bottomSlot;
+  private float minCompetitiveScore;
 
   /**
    * Create the first pass collector.
@@ -83,7 +87,6 @@ public class FirstPassGroupingCollector<T> extends SimpleCollector {
     // and specialize it?
 
     this.topNGroups = topNGroups;
-    this.needsScores = groupSort.needsScores();
     final SortField[] sortFields = groupSort.getSort();
     comparators = new FieldComparator<?>[sortFields.length];
     leafComparators = new LeafFieldComparator[sortFields.length];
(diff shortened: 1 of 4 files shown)


lucene/grouping/src/java/org/apache/lucene/search/grouping/FirstPassGroupingCollector.java
```diff
@@ -91,19 +95,34 @@ public FirstPassGroupingCollector(
     for (int i = 0; i < sortFields.length; i++) {
       final SortField sortField = sortFields[i];
 
+      final Pruning pruning;
+      if (i == 0) {
+        pruning = compIDXEnd >= 0 ? Pruning.GREATER_THAN : Pruning.GREATER_THAN_OR_EQUAL_TO;
```
jainankitk, line 100: Isn't pruning always equal to `Pruning.GREATER_THAN` since `compIDXEnd = sortFields.length - 1`?

lucene/grouping/src/java/org/apache/lucene/search/grouping/FirstPassGroupingCollector.java
```diff
@@ -91,19 +95,34 @@ public FirstPassGroupingCollector(
...
+      comparators[i] = sortField.getComparator(topNGroups + 1, pruning);
       reversed[i] = sortField.getReverse() ? -1 : 1;
     }
 
+    if (SortField.FIELD_SCORE.equals(sortFields[0]) == true) {
+      scoreMode = ScoreMode.TOP_SCORES;
+      canSetMinScore = true;
+    } else {
+      scoreMode = groupSort.needsScores() ? ScoreMode.TOP_DOCS_WITH_SCORES : ScoreMode.TOP_DOCS;
+      canSetMinScore = false;
```
jainankitk, line 116: We don't need `ScoreMode.COMPLETE / ScoreMode.COMPLETE_NO_SCORES` here, since we don't have `totalHitsThreshold` like parameter?

jainankitk (approved):
Mostly looks correct to me. Will wait for couple of days, in case, someone has concerns with this change!

PR #15211, opened 2025-09-22
[Tessellator] Fix failure due to hole bridge being coplanar with polygon edge.

iverase:
Polygon might fail to tesselate if the bridge used for merging the hole with the polygon gets removed due to coplanarity between the hole and the polygon edges.  The fix changes the strategy for removing co-planar points to do it once after all holes are removed instead of doing it for each removed hole.

fixes https://github.com/apache/lucene/issues/15205


lucene/core/src/test/org/apache/lucene/geo/TestTessellator.java
```diff
@@ -374,7 +374,7 @@ public void testComplexPolygon17() throws Exception {
     IllegalArgumentException ex =
         expectThrows(IllegalArgumentException.class, () -> Tessellator.tessellate(polygon, true));
     assertEquals(
-        "Polygon ring self-intersection at lat=40.8278688 lon=14.1990929", ex.getMessage());
+        "Polygon ring self-intersection at lat=40.8277665 lon=14.1993975", ex.getMessage());
```
iverase, line 377: This change has changed the returned point here, that should be ok.

craigtaverner (approved):
LGTM

ayinresh:
@iverase + @craigtaverner thanks so much for the fast fix!

ayinresh:
@iverase I see you backported this to [9.12.3](https://github.com/apache/lucene/commit/911b518aa8aff32240499012718bc666dd68b291). Is there any way it can also be backported to 10.2.x? I tried to just copy over the fix into our code base, but the important classes (i.e. `Triangle`) are not public.

ayinresh:
@iverase @craigtaverner - it looks like this fix may have missed a spot - another similar but valid geometry is failing, reported in https://github.com/apache/lucene/issues/15554

PR #15198, opened 2025-09-17
Vectorise expand8 in ForUtil using JDK Vector API

RamakrishnaChilaka:
This PR optimizes the expand8 routine by leveraging the JDK Vector API. 

#### Benchmarks
I have validated performance using a standalone benchmark (see [postings_expand_benchmark](https://github.com/RamakrishnaChilaka/-postings_expand_benchmark)) for block_size: 256. Key take-aways are as follows. Benchmarks ran on i5-13600k and 256 bit vectors.

| Benchmark                     | Mode | Cnt |   Score   |  Error  | Units |
|-------------------------------|------|-----|-----------|---------|-------|
| expand16 (Scalar)             | thrpt|  5  | 112.842   | ± 0.221 | ops/us |
| expand16 (Vector)             | thrpt|  5  | 105.594   | ± 1.307 | ops/us |
| expand8 (Scalar)              | thrpt|  5  |  66.726   | ± 0.452 | ops/us |
| expand8 (Vector)              | thrpt|  5  | 105.821   | ± 0.272 | ops/us |

* **expand8**: Vectorized version is ~59% faster than scalar (66.7 → 105.8 ops/us).
* **expand16**: Scalar slightly outperforms vector (112.8 vs 105.6 ops/us).

### Lucene Microbenchmarks

```

baseline
Benchmark                                (bpv)   Mode  Cnt   Score   Error   Units
PostingIndexInputBenchmark.decode            2  thrpt   15  35.409 ± 0.120  ops/us
PostingIndexInputBenchmark.decode            3  thrpt   15  29.128 ± 0.017  ops/us
PostingIndexInputBenchmark.decode            4  thrpt   15  41.492 ± 0.305  ops/us
PostingIndexInputBenchmark.decode            5  thrpt   15  32.205 ± 0.350  ops/us
PostingIndexInputBenchmark.decode            6  thrpt   15  31.237 ± 0.245  ops/us
PostingIndexInputBenchmark.decode            7  thrpt   15  29.984 ± 0.582  ops/us
PostingIndexInputBenchmark.decode            8  thrpt   15  56.366 ± 0.134  ops/us
PostingIndexInputBenchmark.decode            9  thrpt   15  22.802 ± 0.077  ops/us
PostingIndexInputBenchmark.decode           10  thrpt   15  23.502 ± 0.037  ops/us
PostingIndexInputBenchmark.decodeVector      2  thrpt   15  53.151 ± 0.070  ops/us
[...]

PR #15218, opened 2025-09-23
add JavaDoc for expand8 in VectorUtil

RamakrishnaChilaka:
### Description 

Follow up from https://github.com/apache/lucene/pull/15198

Add JavaDoc for `expand8` in VectorUtil.java.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/VectorUtil.java
+++ b/lucene/core/src/java/org/apache/lucene/util/VectorUtil.java
@@ -492,6 +492,12 @@ public final class VectorUtil {
     return IMPL.filterByScore(docBuffer, scoreBuffer, minScoreInclusive, upTo);
   }
 
+  /**
+   * Expands 64 integers in-place into a 256-element array by extracting individual bytes. Each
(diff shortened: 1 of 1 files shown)


lucene/core/src/java/org/apache/lucene/internal/vectorization/VectorUtilSupport.java
```diff
@@ -145,11 +145,5 @@ float recalculateScalarQuantizationOffset(
 
   float[] l2normalize(float[] v, boolean throwOnZero);
 
-  /**
```
uschindler, line 148: We can keep this, not sure about it. Some methods have Java docs, others not.

Maybe to be safe keep both.

uschindler (approved):

lucene/core/src/java/org/apache/lucene/internal/vectorization/VectorUtilSupport.java
RamakrishnaChilaka: I agree that it's safer to keep JavaDocs for both. Reverted the change to remove JavaDoc from VectorUtilSupport.java.

PR #15221, opened 2025-09-23
Failing test for early termination in DrillSideways

tflobbe:
### Description
Failing test that shows there is no early termination in DrillSideways when the hit collector throws `CollectionTerminatedException`. This test would pass in Lucene 8.10. Related to #15220


tflobbe:
Added an implementation using a dedicated collector for DrillSideways that can bubble up the early termination. This works for the sequential case, but it won't work for the concurrent case, since in concurrent, we are collecting directly with the sideways collector. For the concurrent case, each sideways facet collector should early terminate on it's own. 
A FacetsCollector that early terminates on it's own is something that would even work today without any changes, it's just sometimes difficult to have FacetsCollector know when to terminate, depending on the logic that's triggering that

PR #15223, opened 2025-09-23
Move legacy scalar formats to backwards_codecs

benwtrent:
Slowly moving the legacy formats to the backwards codecs.

I have most of the logic moved, but there are additional things to figure out. 

I don't think we can easily move the `Lucene99ScalarQuantizedVectorScorer` just yet, but we should be able to prevent users from using the old quantized formats.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorsReader.java
@@ -53,9 +53,12 @@ import org.apache.lucene.util.RamUsageEstimator;
 import org.apache.lucene.util.hnsw.OrdinalTranslatedKnnCollector;
 import org.apache.lucene.util.hnsw.RandomVectorScorer;
 import org.apache.lucene.util.quantization.OptimizedScalarQuantizer;
+import org.apache.lucene.util.quantization.QuantizedVectorsReader;
+import org.apache.lucene.util.quantization.ScalarQuantizer;
 
 /** Reader for scalar quantized vectors in the Lucene 10.4 format. */
-class Lucene104ScalarQuantizedVectorsReader extends FlatVectorsReader {
+class Lucene104ScalarQuantizedVectorsReader extends FlatVectorsReader
+    implements QuantizedVectorsReader {
 
   private static final long SHALLOW_SIZE =
       RamUsageEstimator.shallowSizeOfInstance(Lucene104ScalarQuantizedVectorsReader.class);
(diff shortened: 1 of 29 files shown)


benwtrent:
EDIT: these have now been fixed, minor oversight on my part.

🤔 

some failures:

The rescore ones are weird to me...

```
  - org.apache.lucene.search.TestFullPrecisionFloatVectorSimilarityValuesSource.testFullPrecisionVectorSimilarityDVS (:lucene:core)
    Test output: /Users/benjamintrent/Projects/lucene/lucene/core/build/test-results/test/outputs/OUTPUT-org.apache.lucene.search.TestFullPrecisionFloatVectorSimilarityValuesSource.txt
    Reproduce with: gradlew :lucene:core:test --tests "org.apache.lucene.search.TestFullPrecisionFloatVectorSimilarityValuesSource.testFullPrecisionVectorSimilarityDVS" -Ptests.file.encoding=ISO-8859-1 -Ptests.gui=false "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1" -Ptests.jvms=5 -Ptests.seed=B5A4FBB04FB944A9 -Ptests.vectorsize=256

  - org.apache.lucene.search.TestRescoreTopNQuery.testRescoreField (:lucene:core)
    Test output: /Users/benjamintrent/Projects/lucene/lucene/core/build/test-results/test/outputs/OUTPUT-org.apache.lucene.search.TestRescoreTopNQuery.txt
    Reproduce with: gradlew :lucene:core:test --tests "org.apache.lucene.search.TestRescoreTopNQuery.testRescoreField" -Ptests.file.encoding=ISO-8859-1 -Ptests.gui=false "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1" -Ptests.jvms=5 -Ptests.seed=B5A4FBB04FB944A9 -Ptests.vectorsize=256

  - org.apache.lucene.backward_codecs.lucene99.TestLucene99HnswQuantizedVectorsFormat.testQuantizedVectorsWriteAndRead (:lucene:backward-codecs)
    Test output: /Users/benjamintrent/Projects/lucene/lucene/backward-codecs/build/test-results/test/outputs/OUTPUT-org.apache.lucene.backward_codecs.lucene99.TestLucene99HnswQuantizedVectorsFormat.txt
    Reproduce with: gradlew :lucene:backward-codecs:test --tests "org.apache.lucene.backward_codecs.lucene99.TestLucene99HnswQuantizedVectorsFormat.testQuantizedVectorsWriteAndRead" -Ptests.file.encoding=UTF-8 -Ptests.gui=false "-Ptests.jvmargs=-XX:TieredStopAtLevel=1 -XX:+UseParallelGC -XX:ActiveProcessorCount=1" -Ptests.jvms=5 -Ptests.seed=B5A4FBB04FB944A9 -Ptests.vectorsize=512
```

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorsReader.java
```diff
@@ -53,9 +53,12 @@
 import org.apache.lucene.util.hnsw.OrdinalTranslatedKnnCollector;
 import org.apache.lucene.util.hnsw.RandomVectorScorer;
 import org.apache.lucene.util.quantization.OptimizedScalarQuantizer;
+import org.apache.lucene.util.quantization.QuantizedVectorsReader;
+import org.apache.lucene.util.quantization.ScalarQuantizer;
 
 /** Reader for scalar quantized vectors in the Lucene 10.4 format. */
-class Lucene104ScalarQuantizedVectorsReader extends FlatVectorsReader {
+class Lucene104ScalarQuantizedVectorsReader extends FlatVectorsReader
+    implements QuantizedVectorsReader {
```
mccullocht, line 61: What is the purpose of implementing this interface? It gives access to the raw vector bytes but not enough information to interpret them.

mccullocht (approved):
Ah so the fix here is to move the test for the scorer to backwards codecs but leave the class in place.

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorsReader.java
benwtrent: It satisfies some bwc checks. We need to adjust our bwc tests, but right now it assumes that a format indicates if its quantized by providing this reader interface.

benwtrent:
> Ah so the fix here is to move the test for the scorer to backwards codecs but leave the class in place.

Yeah, I think so for now. Maybe we can remove the scorers as well, but that will be complicated.

I also think we should remove the old ScalarQuantizer eventually.

But I wanted to move the formats first to then figure out whats unused and can also be moved. These types of things is always tricky.

mccullocht:
Agreed, we should remove the old ScalarQuantizer, but I'd like to get off heap and bulk scoring for OSQ in first.

navneet1v:
@benwtrent, @mccullocht  I see that we are removing the old Scaler Format here `Lucene99ScalarQuantizedVectorsFormat` to a new one, but the new format is backward compatible with parameters like Confidence interval etc. Is the expectation here that consumers should create their own implementation of `Lucene99ScalarQuantizedVectorsFormat` writers?

benwtrent:
@navneet1v 

All vector formats are labeled experimental I think. If somebody is providing there own custom set of parameters for the format, they will need to adjust the format and the constructors. This is the case with any format upgrade in Lucene.

The new quantization format doesn't require confidence interval parameter, nor does it allow compression changes.


As an aside, looking at the code, I see that Lucene104 format isn't flagged as experimental and it should be. We need to adjust the docs there, that is definitely an oversight.

navneet1v:
> @navneet1v
> 
> All vector formats are labeled experimental I think. If somebody is providing there own custom set of parameters for the format, they will need to adjust the format and the constructors. This is the case with any format upgrade in Lucene.

@benwtrent 
Sure that make sense. Frankly speaking adjusting is least of my question here. What I wanted to raise was 
1. Since we have the new format and in a lucene index which has old segment with `Lucene99ScalarQuantizedVectorsFormat` and new segments with `Lucene104HnswScalarQuantizedVectorsFormat` may give different scores for different vectors. That really freaks me out because as a user at least I should expect scores to be same for same vector or is that not even guaranteed ? :( 

> If somebody is providing there own custom set of parameters for the format, they will need to adjust the format and the constructors.

even for defaults a vector in old vs new segment doesn't guarantee to give same score since we changed the algorithm behind the scene.

PR #15224, opened 2025-09-23
Add asynchronous prefetch for DirectIO directory

benwtrent:
This adds prefetching to directIO. The idea is pretty simple,

 - configure a number of "prefetch buffers" that are the same size as the directIO buffer
 - calling prefetch will start a prefetch virtual thread to fill an available buffer
 - On read, DirectIO will attempt to refill from any prefetched buffers that match the position before attempting to do directIO itself.

When doing many prefetches and handling things in batches, this can significantly improve throughput.

### DIFF
--- a/lucene/misc/src/java/org/apache/lucene/misc/store/DirectIODirectory.java
+++ b/lucene/misc/src/java/org/apache/lucene/misc/store/DirectIODirectory.java
@@ -18,6 +18,7 @@ package org.apache.lucene.misc.store;
 
 import static java.nio.ByteOrder.LITTLE_ENDIAN;
 
+import java.io.Closeable;
 import java.io.EOFException;
 import java.io.IOException;
 import java.io.UncheckedIOException;
@@ -27,9 +28,16 @@ import java.nio.file.Files;
 import java.nio.file.OpenOption;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
+import java.util.ArrayDeque;
 import java.util.Arrays;
+import java.util.Deque;
 import java.util.Objects;
 import java.util.OptionalLong;
+import java.util.TreeMap;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.zip.CRC32;
 import java.util.zip.Checksum;
 import org.apache.lucene.store.AlreadyClosedException;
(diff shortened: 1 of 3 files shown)


lucene/misc/src/java/org/apache/lucene/misc/store/DirectIODirectory.java
```diff
@@ -554,8 +642,178 @@ public IndexInput slice(String sliceDescription, long offset, long length) throw
...
+    private final long[] prefetchPos;
+    private final Future<?>[] prefetchThreads;
+    private final TreeMap<Long, Integer> posToSlot;
+    private final Deque<Integer> slots;
+    private final ByteBuffer[] prefetchBuffers;
+    private final IOException[] prefetchExceptions;
+    private final int prefetchBytesSize;
+    private final Deque<Long> pendingPrefetches = new ArrayDeque<>();
+    private final FileChannel channel;
+    private final ExecutorService executor = Executors.newVirtualThreadPerTaskExecutor();
```
mccullocht, line 668: Is the executor something you would want to share within a directory or potentially even across directories? I can't find any documentation that indicates that this pattern would be a problem.

dungba88:
Do you have some number on throughput change?

mikemccand:
This is neat -- if Lucene implements enough top-down prefetch hinting, it might eventually be that DirectIO, alone, is sufficient for good query latency/throughput?  I.e. we could stop entirely relying on OS to do its prefetching/caching (buffer cache), maybe, in very cold indices?

Isn't `DirectIODirectory` today only inserting itself for merge context reading & writing?

benwtrent:
> Isn't DirectIODirectory today only inserting itself for merge context reading & writing?

Correct, its only used in certain scenarios. We are experimenting using it in more areas (e.g. vector rescoring, to keep from polluting the off-heap cache with rescoring vectors)

> it might eventually be that DirectIO, alone, is sufficient for good query latency/throughput?

Its not quite there yet. I have seen this improve throughput by more than 2x though depending on the read patterns. MMAP still has TONS of advantages (direct memory segment access being a HUGE one for vectors). 

Virtual threads make this VERY easy, but I am sure there is a lot of headroom for improvements.

benwtrent:
I also think that NIOFS could benefit of a prefetch implementation as well.

mccullocht:
If you used direct io for everything you would want to introduce an explicit disk cache somewhere, even with prefetching I don't think performance would meet expectations for a lot of workloads if most reads resulted in a syscall.

benwtrent:
> If you used direct io for everything you would want to introduce an explicit disk cache somewhere, even with prefetching I don't think performance would meet expectations for a lot of workloads if most reads resulted in a syscall.

100% agreed. I think we are a long ways away from making IO super cheap. 

Again, MMAP has many benefits still. But virtual threads do make this way easier than it would have been before!

benwtrent:
I am gonna close this. Honestly, all benchmarking found the tradeoffs here to not be worth it. I think for anything direct IO focused that wants to handle throughput, we need to use some native things...or just stick with mmap and be ok with the page faults.

FWIW, in benchmarking we found just doing pre-fetch with MMAP and multiple threads reached IO saturation, while this PR never really did, and consequently read throughput was always worse :(

PR #15227, opened 2025-09-23
Empty Terms's prepareSeekExact'get()  always false

LuXugang:
When constructing TermStats, if a segment does not contain the target field, we don’t need to return an empty Term, because in the second loop the corresponding `prepareSeekExact` method will always return false

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/index/TermStates.java
+++ b/lucene/core/src/java/org/apache/lucene/index/TermStates.java
@@ -101,7 +101,10 @@ public final class TermStates {
     if (needsStats) {
       PendingTermLookup[] pendingTermLookups = new PendingTermLookup[0];
       for (LeafReaderContext ctx : context.leaves()) {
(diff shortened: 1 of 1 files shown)


lucene/core/src/java/org/apache/lucene/index/TermStates.java
```diff
@@ -101,7 +101,10 @@ public static TermStates build(IndexSearcher indexSearcher, Term term, boolean n
     if (needsStats) {
       PendingTermLookup[] pendingTermLookups = new PendingTermLookup[0];
       for (LeafReaderContext ctx : context.leaves()) {
-        Terms terms = Terms.getTerms(ctx.reader(), term.field());
+        Terms terms = ctx.reader().terms(term.field());
+        if (terms == null) {
+          continue;
+        }
```
jainankitk, line 107: I am wondering if we can instead keep the original and check for `terms == EMPTY`. Although, both read okay to me.

jainankitk (approved):

PR #15232, opened 2025-09-25
BENCHMARK(benchmark-jmh): Add benchmark for Operations.determinize()

Jacky040124:
Introduces a new JMH benchmark for Operations.determinize() to provide performance data for future optimizations, addressing issue #11025. 

The benchmark measures the performance of the NFA-to-DFA conversion against a curated set of regular expressions.


rmuir:
For some of these regexps, they may now come out deterministic or even minimal+deterministic to begin with at the parsing phase (this is a good thing!): we may have to get more creative with the regexps to force the determinize() to do actual work.

In such a case, the `Operations.determinize()` is a no-op, which is why I think you see some of the crazy-fast numbers here.

Best way to check out the regexes is to just write little throwaway unit-tests similar to:
https://github.com/apache/lucene/blob/002094613418c4bc6a7e335a8edca82fd26ac03d/lucene/core/src/test/org/apache/lucene/util/automaton/TestRegExpParsing.java#L527-L533

Basically, if the result from `toAutomaton()` passes `assertCleanDFA()`, then you know it is already a DFA and determinize() wont do anything. See the assertions here:

https://github.com/apache/lucene/blob/de1ed71261d579fdd3cf71b0734f30ea799c4b1f/lucene/test-framework/src/java/org/apache/lucene/tests/util/automaton/AutomatonTestUtil.java#L394-L413

Jacky040124:
@rmuir Thank you so much for your help.

I confirmed the hypothesis that the original benchmark patterns were already optimised to DFAs by Lucene's parser, causing Operations.determinize() to be a no-op.

```
Pattern: a+                    → Clean DFA
Pattern: .*                    → Clean DFA
Pattern: [0-9]+                → Clean DFA 
Pattern: a(b+|c+)d             → Clean DFA
Pattern: (cat|dog|bird|fish|mouse) → Clean DFA 
Only pattern: (a+)+           → NFA (0.003 ms/op)
```

I've written a thrown away test script using the assertCleanDFA method as you suggested, extracted and tested 27 patterns from OpenJDK's test suite (from the SO post), and found 11 NFA patterns that force determinization work:
```
^(a)?a, ((a|b)?b)+, (aaa)?aaa, ^(a(b(c)?)?)?abc,
(a+)+, (a*)+, (b+)+, (|f)?+, (y+)*,
(foo|foobar)*, (aa+|bb+)+
```

Also notice something very interesting, the two patterns explicitly marked as "Nondeterministic group" in OpenJDK's test suite seems to be optimized to DFAs by Lucene's parser (≈ 10⁻⁶ ms/op), contradicting OpenJDK's "nondeterministic" label

```java
// Nondeterministic group
(a+b)+
(a|b)+
```

I've updated RegexDeterminizeBenchmark.java with the 11 verified NFA patterns, ran the benchmark, and committed the results. All patterns now show meaningful determination work (0.001-0.014 ms/op vs ≈ 10⁻⁶ ms/op previously). The benchmark appears to be working correctly now. Please let me know if any additional changes are needed. 

Thank you again for your help, please let me know if any extra change is needed !

PR #15236, opened 2025-09-26
Upgrade to Gradle 9.1.0

### LINKED ISSUE #15215: java 25 support
### Description

Java 25 is released, we need to upgrade to the Gradle 9.1 to be able to use it: https://docs.gradle.org/9.1.0/release-notes.html

Because it is a new LTS release, package managers are already pushing the upgrade: java 24 is EOL. We should bump the minimum of `main` to match as well.

dweiss:
This is a work in progress patch.

I've tried to assess how much is not working after an upgrade to gradle 9.1.0 and it seems most of the infrastructure works fine. Regeneration doesn't work (I know why). I'll try to push it forward but it seems like it's not going to be as complicated as I thought.

Issues to resolve:

* [x] (fixed by [15240](#15240))  this passes for me locally but only because the build directory is created first, racy;
```
Configuring project ':lucene:build-tools' without an existing directory is not allowed. The configured projectDirectory '/home/runner/work/lucene/lucene/lucene/build-tools' does not exist, can't be written to or is not a directory.
```
* [x] (fixed by [15244](#15244))  regeneration tasks don't work (project.exec removed)
* [x] ensure everything works with jdk24 after all the changes have been applied
* [x] temporarily disable forbidden-apis checker until it supports jdk 25 (tracking issue [15246](#15246 ))
* [x] wait for ecj support of Java 25 or use a custom copy of the latest drop. (as it was in the now removed https://github.com/apache/lucene/commit/015b3bbf9c3aa389c497ea708aedb4ad8557fc8f#diff-d8369391f4d26d4041b621158bf7f0b934a40c4dd98850e3d734fc985d38e257)
* [x] update gh workflows to include jdk 25

Related patch #15238


dweiss:
I've made some reasonable progress. I'll try to make ecj work tomorrow and see what's left.

dweiss:
I think everything works... I checked everything locally and the gh workflow that verifies gradle upgrade sanity is also passing. We can wait for Uwe to release forbidden-apis or merge this in and remove the temporary code that turns forbidden apis off (#15246).

rmuir:
Wow, I think merge now to reduce any hassles and let's smooth anything out as a followup?

dweiss:
Merging this will also need to be coordinated with switching all jenkins workflows to jdk 25... which I'm not sure is installed on worker machines (asf infra, uwe's box)? I'll create a sub-issue to investigate this.

kaivalnp (approved):
Nice!

build.gradle
```diff
@@ -57,3 +56,11 @@ plugins {
 }
 
 description = 'Grandparent project for Apache Lucene Core'
+
+// Turn off all forbidden-apis tasks until it supports JDK 25;
+// https://github.com/policeman-tools/forbidden-apis/pull/272
+allprojects {
+  tasks.withType(CheckForbiddenApis.class).configureEach {
+    enabled = false
```
uschindler, line 64: just reenable it like this: `targetVersion = "24"`

Then it runs, but with older signatures!

uschindler (changes requested):

README.md
```diff
@@ -40,7 +40,7 @@ comprehensive documentation, visit:
 
 ### Basic steps:
 
-1. Install [OpenJDK 24](https://jdk.java.net/archive/).
+1. Install [OpenJDK 25](https://jdk.java.net/archive/).
```
uschindler, line 43: the link here only shows java 24 as latest. Maybe only point to home page.

uschindler: Correct link is: https://jdk.java.net/25/ (but it may not be persistent)

build.gradle
dweiss: I'll do this as an interim commit under #15246

PR #15237, opened 2025-09-26
Fix SmartChinese to only deserialize data from classpath with a native array filter applied and never cache dictionaries from custom locations

uschindler:
This PR improves SmartChinese analyzer to only load Java serialized `.mem` files from classpath and never ever cache custom dictionaries in serialized form.

This also adds a filter to the `ObjectInputStream` used for loading the legacy dictionary format from classpath to only accept (arrays of) native types.

At a later stage we should completely remove the serialized data and also add the original sources of the dictionaries (bigram and core) to the Lucene repo and use `DataInput/DataOutput` to write our custom file format.

### DIFF
--- a/lucene/test-framework/src/java/org/apache/lucene/tests/util/LuceneTestCase.java
+++ b/lucene/test-framework/src/java/org/apache/lucene/tests/util/LuceneTestCase.java
@@ -53,7 +53,6 @@ import java.io.Closeable;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.ObjectInputFilter;
 import java.io.PrintStream;
 import java.lang.StackWalker.StackFrame;
 import java.lang.annotation.Documented;
@@ -497,14 +496,6 @@ public abstract class LuceneTestCase extends Assert {
       defaultValue |= systemPropertyAsBoolean(property, false);
     }
     LEAVE_TEMPORARY = defaultValue;
-  }
-
-  /* Make sure we do not allow Java serialization without custom filters anywhere while running tests.
-   * We could have activated that already in the test runner Gradle plugin, but Gradle itsself uses
-   * unfiltered serialization on startup.
-   */
-  static {
-    ObjectInputFilter.Config.setSerialFilter(_ -> ObjectInputFilter.Status.REJECTED);
   }
 
   /** Filesystem-based {@link Directory} implementations. */
--- a/lucene/test-framework/src/java/org/apache/lucene/tests/util/LuceneTestCase.java
+++ b/lucene/test-framework/src/java/org/apache/lucene/tests/util/LuceneTestCase.java
@@ -53,6 +53,7 @@ import java.io.Closeable;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.ObjectInputFilter;
 import java.io.PrintStream;
 import java.lang.StackWalker.StackFrame;
 import java.lang.annotation.Documented;
@@ -496,6 +497,14 @@ public abstract class LuceneTestCase extends Assert {
       defaultValue |= systemPropertyAsBoolean(property, false);
     }
     LEAVE_TEMPORARY = defaultValue;
+  }
+
+  /* Make sure we do not allow Java serialization without custom filters anywhere while running tests.
+   * We could have activated that already in the test runner Gradle plugin, but Gradle itsself uses
+   * unfiltered serialization on startup.
+   */
+  static {
+    ObjectInputFilter.Config.setSerialFilter(_ -> ObjectInputFilter.Status.REJECTED);
   }
 
   /** Filesystem-based {@link Directory} implementations. */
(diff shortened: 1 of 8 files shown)


uschindler:
Just as comment: I tried to rewrite the serialization and desericalization with DataOutput. For the bigram dict with ZigZag encoding its one megabyte smaller (but JAR file size did not change much), but the main dictionary is hard to serialize easily as we have no methods to save sparse multidimensional char/int arrays.

So I kept the serialization code for now, we may look into the complete rewrite later. For now the code is 100% safe due to the serialization filter. I see no reason to change the data format for now, unless we can start from scratch with the original dictionary source code.

uschindler:
Here was my quick hack to serialize the bigram dictionary with DataOutput, size was a megabyte smaller:

```java
  private void saveToObj(Path serialObj) throws IOException {
    try (OutputStreamDataOutput output = new OutputStreamDataOutput(new BufferedOutputStream(Files.newOutputStream(serialObj), 8192))) {
      CodecUtil.writeHeader(output, HEADER, VERSION);
      output.writeVInt(bigramHashTable.length);
      for (int i = 0; i < bigramHashTable.length; i++) {
        output.writeZLong(bigramHashTable[i]);
      }
      output.writeVInt(frequencyTable.length);
      for (int i = 0; i < frequencyTable.length; i++) {
        output.writeVInt(frequencyTable[i]);
      }
      // log.info("serialize bigram dict.");
    }
  }
```

I posted this here as a first idea for future (e.g., main branch).

uschindler:
I added back the code to regenerate the serialized file, but commented out.

lucene/test-framework/src/java/org/apache/lucene/tests/util/LuceneTestCase.java
```diff
@@ -498,6 +499,14 @@ static int defaultRandomMultiplier() {
     LEAVE_TEMPORARY = defaultValue;
   }
 
+  /* Make sure we do not allow Java serialization without custom filters anywhere while running tests.
+   * We could have activated that already in the test runner Gradle plugin, but Gradle itsself uses
+   * unfiltered serialization on startup.
+   */
+  static {
+    ObjectInputFilter.Config.setSerialFilter(_ -> ObjectInputFilter.Status.REJECTED);
```
uschindler: This is something for @rmuir: This early setup in LuceneTestCase makes sure that no test anywhere or any class loaded after the LTC was initialized is allowed to do *unfiltered* serialization.

Ideally we would pass `-Djdk.serialFilter=!*` as test runner command line, but Gradle uses unfiltered serialization on its own child runner setup, so we delay the "deny all" rule up to initialization of LTC.

rmuir (approved):

lucene/test-framework/src/java/org/apache/lucene/tests/util/LuceneTestCase.java
uschindler: Hi, I implemented a better way to check this without breaking downstream users by applying a global filter.

I added code to install a filter factory into the test runner JVM from Gradle. The installed filter allows all Gradle-internal deserialization in a similar way like `TestSecurityManager` allowed `System.exit()` by test runner. It checks stack trace for correct class up in stack.

uschindler:
Hi @rmuir, @dweiss 
I changed the serialization filter installation to work in the same way like SecurityManager in the past. It gets installed by the Gradle infrastructure by system property into the JVM. To still allow Gradle to do serialization, it is whitelisted in the same way like Gradle was allowed to exit with System.exit().

The whole thing works by a `ObjectInputFilterFilter`-factory that is installed via cmd line (its implemented in test framework) and it takes care of installing a deny-all-only-allow-gradle deserialization in cases that the code using deserialization is not installing a filter on its own.

See documentation here: [ObjectInputFilter.Config#setSerialFilterFactory()](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/io/ObjectInputFilter.Config.html#setSerialFilterFactory(java.util.function.BinaryOperator))

Can you review?

dweiss:
LGTM. Which issue does the de-serialization filtering come from? Why was it a problem? Sorry, I wasn't paying attention.

uschindler:
> LGTM. Which issue does the de-serialization filtering come from? Why was it a problem? Sorry, I wasn't paying attention.

See security list :-)

In general this is an improvement to not allow deserialization anywhere (in addition to forbiddenapis). This will also disallow it in external code.

This makes sure that we only use Java's deserialization if it is properly guarded (with class whitelists and other filtering). Unfortunately Gradle is not doing this. Somebody should open a security issue there. If somebody intercepts the internal gradle runner communication it can inject bad data leading to RCE or similar.

uschindler:
I added a test for the filter. I think it's ready.

Backporting is harder as the Gradle infra changed, but it should still be simple copypaste...

dweiss:
> This makes sure that we only use Java's deserialization if it is properly guarded (with class whitelists and other filtering).

I think it's always been the case that deserialization via object input streams could lead to odd things happening. I thought the direction was to move away from trying to secure Java and instead rely on external guardrails (removing the security manager being a prime example of this). Thanks for the pointer though.

uschindler:
Backport was done in https://github.com/apache/lucene/commit/6c9c2df54708dcae7e38534f6079deb9223ef59d

There were some changes needed regarding SecurityManager: AccessController was needed for serialFilter. The permission was added to changes text.

PR #15243, opened 2025-09-27
ci: add CodeQL analysis workflow for java, python and github actions

rmuir:


rmuir:
Screen_Shot_2025-09-29_at_09 09 15

rmuir:
I'm aware the java job takes 5 minutes, I think it is just the nature of a big codebase.

If there are resource issues around this, let's just disable the java language (or reduce its queries back down from security extended).

rmuir:
I'll tone down the queries for java language as we now have hundreds of alerts and a lot of them are not security-related (e.g. implicit narrowing of compound assignments), instead just pure noise. Gimme a bit, ill make a PR.

PR #15245, opened 2025-09-27
Move away from project.javaexec

### LINKED ISSUE #15244: Move away from project.javaexec
### Description

As part of #15215 

`Project#javaexec` was removed in Gradle 9: https://docs.gradle.org/current/userguide/upgrading_version_8.html#deprecated_project_exec -- we need to switch its usages in Lucene ([example](https://github.com/apache/lucene/blob/8defe9316f362a149883b5662ad307fbdb116ea1/build-tools/build-infra/src/main/groovy/lucene.regenerate.icu.gradle#L97))

kaivalnp:
### Description

Fixes #15244

### DIFF
--- a/build-tools/build-infra/src/main/groovy/lucene.regenerate.antlr.gradle
+++ b/build-tools/build-infra/src/main/groovy/lucene.regenerate.antlr.gradle
@@ -15,6 +15,8 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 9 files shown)


dweiss (approved):
I had a plan to do this slightly differently in the long term but in the short term it's a good workaround, I think. I'll merge it in.

PR #15248, opened 2025-09-29
Clarify error message in BackwardsCompatibilityTestBase and fix resource leak in existence check

uschindler:
This PR fixes the confusing (and wrong) error message printed by backwards compatibility tests if the previous major latest version (which does not exist) was not added.

It also fixes a resource leak in existence checks. It should use `Class.getResource() != null` for resource existence checks.


thecoop (approved):

dweiss (approved):

uschindler:
Merged into all three branches (main, 10x, 9.12)

PR #15252, opened 2025-09-29
Try to fix eclipse configuration after upgrading to gradle 9.1.0. 

dweiss:
Fixes #15250, hopefully.

### DIFF
--- a/build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/ide/EclipseSupportPlugin.java
+++ b/build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/ide/EclipseSupportPlugin.java
@@ -32,7 +32,7 @@ import org.apache.lucene.gradle.plugins.java.EcjLintPlugin;
 import org.apache.tools.ant.filters.ReplaceTokens;
 import org.gradle.api.JavaVersion;
 import org.gradle.api.Project;
-import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.artifacts.Dependency;
 import org.gradle.api.artifacts.VersionCatalog;
 import org.gradle.api.plugins.JavaPlugin;
 import org.gradle.api.plugins.JavaPluginExtension;
@@ -44,7 +44,6 @@ import org.gradle.plugins.ide.eclipse.GenerateEclipseJdt;
 import org.gradle.plugins.ide.eclipse.model.AbstractClasspathEntry;
 import org.gradle.plugins.ide.eclipse.model.Classpath;
 import org.gradle.plugins.ide.eclipse.model.ClasspathEntry;
-import org.gradle.plugins.ide.eclipse.model.EclipseClasspath;
 import org.gradle.plugins.ide.eclipse.model.EclipseModel;
 import org.gradle.plugins.ide.eclipse.model.SourceFolder;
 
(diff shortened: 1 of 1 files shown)


build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/ide/EclipseSupportPlugin.java
```diff
@@ -172,23 +153,57 @@ private void configureEclipseIdeSettings(Project project, String eclipseJavaVers
...
         .configure(
             task -> {
+              var classpath = task.getClasspath();
+              classpath.setDefaultOutputDir(project.file("build/eclipse"));
+              classpath.setDownloadJavadoc(false);
+              classpath.setDownloadSources(true);
+
+              // Collect dependencies from selected configurations
+              // and copy them over to a local configuration.
+              List<Dependency> subdependencies = new ArrayList<>();
```
dweiss, line 173: I don't think it's 1:1 the same as before - dependency resolution can end up with different outputs... in theory at least. But it seems to work for me.

uschindler: that was my question, too. You create a clone of the config, but how could they differ. I would say it would resolve the config separately (like a new project with all merged configs, correct?).

It may differ, I am opening the project with Groovy, works!

uschindler:
It looks like this works. It creates a new one-time configuration by copying all config entries into it. It's then its own config and can resolve it as wanted. Nice!

Now I see 2 new errors in Eclipse next to the heisenbug with LeafReaderContext:
image

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/ide/EclipseSupportPlugin.java
dweiss: I think you can theoretically imagine a situation with two configurations having a transitive dependency in different versions - then it would resolve this differently... I think. I wouldn't worry about it though, it's a hack already.

uschindler:
This is a bug in the ecj version that ships with my Eclipse release. The field is definitively initialized in the `this()` ctor called before. Maybe its another kind of Heisenbug!

uschindler:
There's already a bug open at ECJ and fixed already: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4416

uschindler:
We don't see the bug in the Gradle build, because we have ECJ I20250926-1800, which has the fix already.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/ide/EclipseSupportPlugin.java
```diff
@@ -172,23 +153,57 @@ private void configureEclipseIdeSettings(Project project, String eclipseJavaVers
               task.dependsOn(luceneEclipseJdt);
             });
 
+    // Add gradle plugin portal to the source repository list and
+    // apply any ecj source repository hackery the same way as everywhere.
+    project.getRepositories().gradlePluginPortal();
```
uschindler, line 158: why do we need this?

uschindler:
I found another bug with ECJ: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4449

We can't trust Java 25 compilation with ECJ at the current stage. So all errors or non-errors should be carefully inspected.

build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/ide/EclipseSupportPlugin.java
dweiss: There are plugin dependencies (in the build-infra) that wouldn't be resolved from maven central alone and this configuration for Eclipse is set up in the root project (which only has mavenCentral as the source repository).

PR #15254, opened 2025-09-29
ci: further tune the codeql

rmuir:
Some of the security-extended checks were actually useful, it only has one extremely noisy rule, just like the default queries have one extremely noisy rule.

Disable both of the noisy rules via configuration file instead.


rmuir:
Sorry for all the noise here, just a hassle to tune this thing to be helping us instead of being useless and annoying.

PR #15255, opened 2025-09-29
Fix backcompat tests.

dweiss:
I think a constant for the new unreleased minor version is needed. I think. Because I honestly don't grasp how these tests are supposed to work, sorry.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/Version.java
+++ b/lucene/core/src/java/org/apache/lucene/util/Version.java
@@ -182,6 +182,13 @@ public final class Version {
   @Deprecated public static final Version LUCENE_9_12_3 = new Version(9, 12, 3);
 
   /**
+   * Match settings and bugs in Lucene's 9.12.4 release.
+   *
+   * @deprecated Use latest
+   */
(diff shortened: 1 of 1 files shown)


uschindler:
Yes, it is needed (as the new error message says). After the last 9.12.2 release, @ChrisHegarty added the 9.12.3 release. So the same must happen now.

In 9.12 branch the constant was also added.

IMHO, this is not nice because we list a release which is not there, but required at moment to make the iterative releases work. We should possibly remove this at some point in future, but then we need a special case for main branch. In main branch we always need a pointer to the next after stable release to make sure we have all codecs in order, but there it does not hurt, as master is never released. In 10.x branch it is not so nice to have the non-existent release in the constants.

This is all horrible, but it is how it is. The worst was the wrong error message we have seen last night.

In main branch we don't need that constant as we don't list 9.x releases anymore.

uschindler (approved):

jainankitk (approved):
Thanks @dweiss . You just beat me to it! Somehow missed this change for `branch_10x`, and the previous error message did not help understand what was missing.

PR #15260, opened 2025-09-30
Fix smoke tester test workflow to run with Java 25

dweiss:


rmuir (approved):

jainankitk:
Sorry, I am slightly confused. Why do we need to continue adding unsupported 9.12.3 index zip files?

dweiss:
> Sorry, I am slightly confused. Why do we need to continue adding unsupported 9.12.3 index zip files?

There are smoke tester tests that somehow rely on it... I honestly don't remember - there is a long tail of legacy code behind all these checks. It would be nice if somebody found the time to review all these backcompat tests and checks and properly describe what's actually needed and where. It's currently scattered between the release manager script, distro smoke tester and Java tests...

PR #15151, opened 2025-09-03
Use the bulk SimScorer#score API to compute impact scores.

jpountz:
In #15039 we introduced a bulk `SimScorer#score` API and used it to compute scores with the leading conjunctive clause and "essential" clauses of disjunctive queries. With this PR, we are now also using this bulk API when translating (term frequency, length normalization factor) pairs into the maximum possible score that a block of postings may produce.

To do it right, I had to change the impacts API to no longer return a List of (term freq, norm) pairs, but instead two parallel arrays of term frequencies and norms that could (almost) directly be passed to the `SimScorer#score` bulk API. Unfortunately this makes the change quite big since many backward formats had to be touched.

PR #15263, opened 2025-09-30
[10.3] Fix returned Impacts when frequencies are not indexed

iverase:
Return an Impact object with freqs equal to 1 instead of Integer.MAX_VALUE.

Note that this bug does not exists in main as it was fixed in https://github.com/apache/lucene/pull/15151

Iam working on adding a test

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -7,7 +7,9 @@ http://s.apache.org/luceneversions
 
 Bug Fixes
 ---------------------
-(No changes)
+* GITHUB#15263: Fix the returned Impact returned from Lucene103PostingsReader when frequencies
+  are not indexed. It was returning a wrong frequency in that case affecting scoring which
+  might led to performance issues. (Ignacio Vera)
 
 ======================= Lucene 10.3.0 =======================
 
--- a/lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
+++ b/lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
@@ -68,9 +68,14 @@ import org.apache.lucene.util.VectorUtil;
 public final class Lucene103PostingsReader extends PostingsReaderBase {
 
   static final VectorizationProvider VECTORIZATION_PROVIDER = VectorizationProvider.getInstance();
-
-  private static final List<Impact> DUMMY_IMPACTS_NO_FREQ =
-      Collections.singletonList(new Impact(1, 1L));
+  // Dummy impacts, composed of the maximum possible term frequency and the lowest possible
+  // (unsigned) norm value. This is typically used on tail blocks, which don't actually record
+  // impacts as the storage overhead would not be worth any query evaluation speedup, since there's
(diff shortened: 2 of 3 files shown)


lucene/core/src/java/org/apache/lucene/codecs/lucene103/Lucene103PostingsReader.java
```diff
@@ -1382,7 +1379,8 @@ public List<Impact> getImpacts(int level) {
                 return readImpacts(level1SerializedImpacts, level1Impacts);
               }
             }
-            return DUMMY_IMPACTS;
+            // Max freq is 1 since freqs are not indexed
+            return DUMMY_IMPACTS_NO_FREQ;
```
benwtrent: I am not sure about this, I think we should return `DUMMY_IMPACTS` from within the `if (indexHasFreq)` predicate but return `DUMMY_IMPACTS_NO_FREQ` if outside of that predicate.

iverase: you are totally right, I have removed DUMMY from IMPACTS_NO_FREQ as it feels there are not really dummy

iverase:
Added a single test here and open https://github.com/apache/lucene/pull/15266 for a bigger coverage in the active branches.

benwtrent (approved):
This seems correct to me.

curious as to what @gf2121 thinks

ChrisHegarty (approved):
LGTM

ChrisHegarty:
If not already the case, let’s forward port the test to main.

iverase:
@ChrisHegarty see https://github.com/apache/lucene/pull/15266

lucene/CHANGES.txt
```diff
@@ -7,7 +7,9 @@ http://s.apache.org/luceneversions
 
 Bug Fixes
 ---------------------
-(No changes)
+* GITHUB#15263: Fix the returned Impact returned from Lucene103PostingsReader when frequencies
+  are not indexed. It was returning a wrong frequency in that case affecting scoring which
+  might led to performance issues. (Ignacio Vera)
```
jpountz, line 12: nit: maybe replace "wrong" with "suboptimal" since it's correct to return impact scores that are much higher than they should, just suboptimal as it hurts dynamic pruning?

jpountz:
For the record, this looks good to me!

lucene/CHANGES.txt
iverase: done in https://github.com/apache/lucene/commit/25d6374b4981073547bc96173b1c27922f4ea0f3

msfroh:
Wasn't this already fixed in https://github.com/apache/lucene/pull/14511 ?

benwtrent:
yes @msfroh apparently in all the shuffle with the new postings, it was accidentally overwritten and then fixed again in main, but never backported. Quite the kerfuffle :D

msfroh:
Ahh -- looks like https://github.com/apache/lucene/pull/14557 backported the 101 postings reader to 10_x, but missed the 103 postings reader.

PR #15265, opened 2025-09-30
Enable branch and tag protection.

### LINKED ISSUE #15264: Set up branch protection against force-pushes on main, branch_10x
### Description

Prevent accidental screw ups.

dweiss:
Fixes #15264 

I didn't add all the past branch_* branches. Should I?


.asf.yaml
```diff
@@ -22,6 +22,16 @@ github:
     issues: true
     projects: false
 
+protected_branches:
+    main: {}
+    branch_10x: {}
+    branch_9x: {}
```
uschindler, line 28: we should add all older `branch_*` branches (maybe export the list from github) and in parallel ask infra how to setup this with a glob.

Maybe it works like:
```
protected_branches:
    "branch_*": {}
```

uschindler: I think we can only try that out and try to force push as test

uschindler:
why does linting fail now, has to do with main branch problem?

dweiss:
Let me take a look, weird.

uschindler:
In yaml you can have quotes in keys, so globs should work.

.asf.yaml
dweiss: We can try on a some test branch, maybe?

dweiss:
I'd say - let's merge this and then test on test_* if it works.

uschindler:
I am in contact with ASF infra in slack channel!

PR #15266, opened 2025-09-30
Add test to check return values of frequency and norm when there are no frequencies

iverase:
It has happened a couple of times that we returned a wrong value here so let's make sure we have it covered.


tteofili (approved):

ChrisHegarty (approved):
LGTM

lucene/test-framework/src/java/org/apache/lucene/tests/index/BasePostingsFormatTestCase.java
```diff
@@ -1798,4 +1800,31 @@ public void testDocIDRunEnd() throws Exception {
     }
     dir.close();
   }
+
+  protected boolean indexFakeImpacts() {
+    return false;
+  }
```
jpountz, line 1806: should we remove this and fix all postings formats that don't pass the test?

iverase: That sounds a good idea although my knowledge about them is very shallow, e.g. I am not sure what is the right way to detect that a field has no frequencies.

jpountz: Typically you would check FieldIndex#indexOptions. If the value is DOCS, then frequencies are not indexed.

PR #15271, opened 2025-10-01
Extend Lucene104ScalarQuantization format to support quant asymmetry

benwtrent:
We should honestly just have a single quantization format. having one for asymmetric binary and one for symmetric byte seems unnecessary.

This extends the new SQ format to support asymmetry. The initial goal is just parity with the binary format in Lucene 10.2 (single bit docs with 4 bit queries). But I could see this being extended further to support:

(stored <-> queried):

 - 1 <-> 1, 2, 4, 8
 - 2 <-> 4, 8
 - 4 <-> 8
 - 8 <-> 16, 32 (e.g. bfloat16 & float32 query). 

However, those will all take time, and some might not even be useful.

Obviously, this will cause conflicts with @mccullocht off-heap scoring work :(.


lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorsFormat.java
```diff
@@ -186,16 +203,35 @@ public Lucene104ScalarQuantizedVectorsFormat() {
...
+    this(encoding, encoding);
+  }
+
+  /** Creates a new instance with the chosen asymmetric quantization encoding. */
+  public Lucene104ScalarQuantizedVectorsFormat(
+      ScalarEncoding encoding, ScalarEncoding queryEncoding) {
     super(NAME);
     this.encoding = encoding;
+    this.queryEncoding = queryEncoding;
+    // until we have optimized scorers for various other asymmetric encodings, maybe we only allow 1
```
mccullocht: Should ScalarEncoding enumerate both the query and index vector coding? It would turn this error into a compile-time error.

Using `PACKED_NIBBLE` for asymmetric 4:1 is a bit weird because it's packed in a very different way.

FWIW agree about being able to do other combinations, in particular I think 8 -> N for all values of N we want to support is a good idea. I know for 4:1 we write two vector sets during index building to improve scores/graph quality and I'm not sure if that's necessary in other cases but I'm also not sure how much it would hurt.

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorsWriter.java
```diff
@@ -190,22 +196,20 @@ private void writeVectors(
...
-            [OptimizedScalarQuantizer.discretize(
-                fieldData.fieldInfo.getVectorDimension(), encoding.getDimensionsPerByte())];
+        new byte[encoding.getDiscreteDimensions(fieldData.fieldInfo.getVectorDimension())];
     byte[] vector =
         switch (encoding) {
-          case UNSIGNED_BYTE -> scratch;
-          case SEVEN_BIT -> scratch;
-          case PACKED_NIBBLE ->
-              new byte[encoding.getPackedLength(fieldData.fieldInfo.getVectorDimension())];
+          case UNSIGNED_BYTE, SEVEN_BIT -> scratch;
```
mccullocht, line 198: [not actionable] ooh nice, didn't realize you could combine cases like this.

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorsWriter.java
```diff
@@ -190,22 +196,20 @@ private void writeVectors(
...
+          case PACKED_NIBBLE, SINGLE_BIT -> new byte[encoding.getPackedLength(scratch.length)];
         };
     for (int i = 0; i < fieldData.getVectors().size(); i++) {
       float[] v = fieldData.getVectors().get(i);
       OptimizedScalarQuantizer.QuantizationResult corrections =
           scalarQuantizer.scalarQuantize(v, scratch, encoding.getBits(), clusterCenter);
       if (encoding == ScalarEncoding.PACKED_NIBBLE) {
         OffHeapScalarQuantizedVectorValues.packNibbles(scratch, vector);
+      } else if (encoding == ScalarEncoding.SINGLE_BIT) {
```
mccullocht: might be worth converting this back into a switch, i made it an if statement initially because intellij complained that there weren't enough cases. this would help folks adding new encodings in the future.

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorsWriter.java
```diff
@@ -251,22 +255,20 @@ private void writeSortedVectors(
...
+          case UNSIGNED_BYTE, SEVEN_BIT -> scratch;
+          case PACKED_NIBBLE, SINGLE_BIT -> new byte[encoding.getPackedLength(scratch.length)];
         };
     for (int ordinal : ordMap) {
       float[] v = fieldData.getVectors().get(ordinal);
       OptimizedScalarQuantizer.QuantizationResult corrections =
           scalarQuantizer.scalarQuantize(v, scratch, encoding.getBits(), clusterCenter);
       if (encoding == ScalarEncoding.PACKED_NIBBLE) {
         OffHeapScalarQuantizedVectorValues.packNibbles(scratch, vector);
+      } else if (encoding == ScalarEncoding.SINGLE_BIT) {
```
mccullocht: same here

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorsWriter.java
```diff
@@ -385,6 +389,63 @@ static DocsWithFieldSet writeVectorData(
...
+              quantizationScratch,
+              new byte[] {INDEX_BITS, QUERY_BITS},
+              centroid);
+      // pack and store document bit vector
+      packAsBinary(quantizationScratch[0], toIndex);
+      binarizedVectorData.writeBytes(toIndex, toIndex.length);
+      binarizedVectorData.writeInt(Float.floatToIntBits(r[0].lowerInterval()));
+      binarizedVectorData.writeInt(Float.floatToIntBits(r[0].upperInterval()));
+      binarizedVectorData.writeInt(Float.floatToIntBits(r[0].additionalCorrection()));
+      binarizedVectorData.writeInt(r[0].quantizedComponentSum());
```
mccullocht, line 421: This will result in two extra bytes per vector compared to the 102 codec, just checking that's fine (I personally don't have a problem with this).

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorScorer.java
```diff
@@ -131,14 +205,17 @@ public float score(int node) throws IOException {
         public void setScoringOrdinal(int node) throws IOException {
           var rawTargetVector = targetValues.vectorValue(node);
           switch (values.getScalarEncoding()) {
-            case UNSIGNED_BYTE -> targetVector = rawTargetVector;
-            case SEVEN_BIT -> targetVector = rawTargetVector;
+            case UNSIGNED_BYTE, SEVEN_BIT -> targetVector = rawTargetVector;
```
mccullocht, line 200: Can this class be folded into the asymmetric implementation and removed? I think when I wrote this `VectorUtil.int4DotProductBothPacked` didn't exist so I needed to unpack the value, but that shouldn't be necessary anymore.

mccullocht (changes requested):
Overall lgtm, I'd been thinking about something similar because of the amount of overlap with the 102 codec.

No worries about the conflicts, I don't think I'm close to merging that change anyway.

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorsFormat.java
benwtrent: > I know for 4:1 we write two vector sets during index building to improve scores/graph quality and I'm not sure if that's necessary in other cases but I'm also not sure how much it would hurt.

The reason for writing out both, is that during graph building, we don't want to re-quantize query vectors on every calculation for diversity calculation. Possibly, we don't need it for diversity calculation? But we would need to test this to ensure graph quality doesn't drop significantly.

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorsWriter.java
benwtrent: I think this is OK. two bytes is two bytes, for sure, but I would like to err on the side of simplicity and speed for now.

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorScorer.java
benwtrent: let me see

lucene/core/src/java/org/apache/lucene/codecs/lucene104/Lucene104ScalarQuantizedVectorsFormat.java
benwtrent: > Should ScalarEncoding enumerate both the query and index vector coding? It would turn this error into a compile-time error.

I switched to this, it cleaned some stuff up. Did slightly complicate as now we must specify "doc" vs. "query" in various things in the format, but this isn't nearly as bad as I thought it would be.

lucene/core/src/test/org/apache/lucene/codecs/lucene104/TestLucene104HnswScalarQuantizedVectorsFormat.java
```diff
@@ -47,14 +48,32 @@
...
-  private static final KnnVectorsFormat FORMAT = new Lucene104HnswScalarQuantizedVectorsFormat();
+  private KnnVectorsFormat format;
+  private ScalarEncoding encoding;
+
+  @Before
+  @Override
+  public void setUp() throws Exception {
+    var encodingValues = ScalarEncoding.values();
+    encoding = encodingValues[random().nextInt(encodingValues.length)];
+    // always assume asymmetric for now. Eventually make this more general
```
mccullocht: nit: no longer needed

mccullocht (approved):

PR #15272, opened 2025-10-01
Performance regression in vector computations due to call site pollution

kaivalnp:
I was trying to figure out why #14863 adversely affected indexing performance of byte quantized vectors in nightly benchmarks (see https://github.com/apache/lucene/pull/14863#issuecomment-3318974886), when it was supposed to speed things up by doing vector computations off-heap -- and _may_ have found something interesting!

### DIFF
--- a/lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/VectorScorerBenchmark.java
+++ b/lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/VectorScorerBenchmark.java
@@ -33,6 +33,7 @@ import org.apache.lucene.store.IndexInput;
 import org.apache.lucene.store.IndexOutput;
 import org.apache.lucene.store.MMapDirectory;
 import org.apache.lucene.util.IOUtils;
+import org.apache.lucene.util.VectorUtil;
 import org.apache.lucene.util.hnsw.RandomVectorScorer;
 import org.apache.lucene.util.hnsw.RandomVectorScorerSupplier;
 import org.apache.lucene.util.hnsw.UpdateableRandomVectorScorer;
@@ -91,6 +92,11 @@ public class VectorScorerBenchmark {
             .getRandomVectorScorerSupplier(DOT_PRODUCT, vectorValues)
             .scorer();
     scorer.setScoringOrdinal(0);
+
+    // Pollute call site!
+    for (int i = 0; i < 10_000_000; i++) {
+      VectorUtil.dotProduct(vec1, vec2);
+    }
   }
 
   @TearDown
@@ -106,6 +112,17 @@ public class VectorScorerBenchmark {
   @Benchmark
   @Fork(jvmArgsPrepend = {"--add-modules=jdk.incubator.vector"})
   public float binaryDotProductMemSeg() throws IOException {
+    return scorer.score(1);
+  }
+
+  @Benchmark
+  @Fork(
+      jvmArgsPrepend = {
+        "--add-modules=jdk.incubator.vector",
+        "-XX:+UnlockDiagnosticVMOptions",
+        "-XX:CompilerDirectivesFile=lucene/benchmark-jmh/src/resources/vector-directives.json"
+      })
+  public float binaryDotProductMemSegWithVectorDirectives() throws IOException {
     return scorer.score(1);
   }
 
--- a/lucene/benchmark-jmh/src/resources/vector-directives.json
+++ b/lucene/benchmark-jmh/src/resources/vector-directives.json
@@ -1,0 +1,9 @@
+[
+  {
+    "match": [ "org.apache.lucene.internal.vectorization.PanamaVectorUtilSupport::*" ],
+    "inline": [
+      "-org.apache.lucene.internal.vectorization.PanamaVectorUtilSupport$ArrayLoader::*",
+      "-org.apache.lucene.internal.vectorization.PanamaVectorUtilSupport$MemorySegmentLoader::*"
+    ]
+  }
+]


kaivalnp:
We use the [same underlying function](https://github.com/apache/lucene/blob/cd88a97a63a97dc857d692ec3eb73b229bb5f372/lucene/core/src/java25/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java#L387) for a [variety of combinations](https://github.com/apache/lucene/blob/cd88a97a63a97dc857d692ec3eb73b229bb5f372/lucene/core/src/java25/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java#L361-L385) of on and off-heap vectors -- which may lead to its call site being polluted, and compiled functions being sub-optimal

To demonstrate this, I added some type pollution to [`VectorScorerBenchmark`](https://github.com/apache/lucene/blob/main/lucene/benchmark-jmh/src/java/org/apache/lucene/benchmark/jmh/VectorScorerBenchmark.java)

Benchmark without pollution:
```
Benchmark                                                         (size)   Mode  Cnt  Score   Error   Units
VectorScorerBenchmark.binaryDotProductMemSeg                        1024  thrpt   15  7.543 ± 0.062  ops/us
```

Benchmark with pollution:
```
Benchmark                                                         (size)   Mode  Cnt  Score   Error   Units
VectorScorerBenchmark.binaryDotProductMemSeg                        1024  thrpt   15  6.327 ± 0.057  ops/us
```

kaivalnp:
On printing some JVM inlining internals, I came across entries like:
```
 @ 35   org.apache.lucene.internal.vectorization.PanamaVectorUtilSupport$ArrayLoader::length (6 bytes)   inline (hot)   callee changed to  org.apache.lucene.internal.vectorization.PanamaVectorUtilSupport$MemorySegmentLoader::length (13 bytes)   inline (hot)   callee changed to  org.apache.lucene.internal.vectorization.PanamaVectorUtilSupport::dotProductBody (250 bytes)    \-> TypeProfile (1028/43516 counts) = org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport$MemorySegmentLoader   callee changed to  org.apache.lucene.internal.vectorization.PanamaVectorUtilSupport::dotProductBody (250 bytes)    \-> TypeProfile (42488/43516 counts) = org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport$ArrayLoader
```

..so I turned off JVM method inlining for those [internal classes](https://github.com/apache/lucene/blob/cd88a97a63a97dc857d692ec3eb73b229bb5f372/lucene/core/src/java25/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java#L323-L359) using some compiler directives (so that their callers are compiled separately, into more type-appropriate methods)

Here are the results:
```
Benchmark                                                         (size)   Mode  Cnt  Score   Error   Units
VectorScorerBenchmark.binaryDotProductMemSeg                        1024  thrpt   15  6.374 ± 0.026  ops/us
VectorScorerBenchmark.binaryDotProductMemSegWithVectorDirectives    1024  thrpt   15  7.401 ± 0.005  ops/us
```

Looks like we can regain most of the performance drop from call site pollution!

kaivalnp:
Long-term, we probably want to rewrite `PanamaVectorUtilSupport` so that these directives aren't needed?

In the short-term, how can we make users aware of this issue / workaround?
For example, the directives probably need to be added to files like [this](https://github.com/opensearch-project/OpenSearch/blob/main/distribution/src/config/jvm.options) (for OpenSearch) or [this](https://github.com/elastic/elasticsearch/blob/main/distribution/src/config/jvm.options) (for ElasticSearch)

kaivalnp:
Also wanted to note: we fixed a similar JVM inlining issue in #14874 -- and the `PanamaVectorUtilSupport$ArrayLoader` and `PanamaVectorUtilSupport$MemorySegmentLoader` classes referenced in this PR were actually added there^

I undid those changes locally (move away from [`ByteVectorLoader`](https://github.com/apache/lucene/blob/bace2920b4487384e4195a6914c8074c498ae4b5/lucene/core/src/java25/org/apache/lucene/internal/vectorization/PanamaVectorUtilSupport.java#L315), wrap `byte[]` in a `MemorySegment` and have all functions work on two `MemorySegment`s), but that degrades performance further:

```
Benchmark                                                         (size)   Mode  Cnt  Score   Error   Units
VectorScorerBenchmark.binaryDotProductMemSeg                        1024  thrpt   15  1.592 ± 0.065  ops/us
```

i.e. the above PR was net-net positive, but left some _more_ performance to be reclaimed by sidestepping call site pollution (which we hope to do here)

msokolov:
It's annoying that the JVM only allows this kind of control via command-line args rather than annotations that can be placed in code. In fact there are such annotations (@Inline and @DontInline) but they are not available to lowly users, only for JVM internal code.

mccullocht:
It's possible that wrapping `byte[]` in `MemorySegment` is just moving the inlining issue around. I tried something similar in my PR for off-heap scoring and got similar results.

It might be best to use a script to generate code for the 3 different input cases.

mccullocht:
I tried adding a ByteVectorLoader implementation that stores both possible representations and switches between them in the loop (think `if (segment != null) { <segment stuff> } else { <array stuff> }`. This was not successful.

With pollution:
```
Benchmark                                     (size)   Mode  Cnt   Score   Error   Units
VectorScorerBenchmark.binaryDotProductMemSeg    1024  thrpt   15  25.176 ± 2.890  ops/us
```

Without pollution:
```
Benchmark                                     (size)   Mode  Cnt   Score   Error   Units
VectorScorerBenchmark.binaryDotProductMemSeg    1024  thrpt   15  32.198 ± 0.077  ops/us
```

With union vector representation:
```
Benchmark                                     (size)   Mode  Cnt   Score   Error   Units
VectorScorerBenchmark.binaryDotProductMemSeg    1024  thrpt   15  28.138 ± 0.108  ops/us
```

I also tried sealing ByteVectorLoader and switching on the class type. On most run this is faster than the union vector representation, but on some jvm runs it is ~10x slower :/.

kaivalnp:
Turns out these compiler directives make things worse in actual benchmarks (Cohere vectors, 768d):

baseline

```
recall  latency(ms)  netCPU  avgCpuCount    nDoc  topK  fanout  maxConn  beamWidth  quantized  visited  index(s)  index_docs/s  force_merge(s)  num_segments  index_size(MB)  vec_disk(MB)  vec_RAM(MB)  indexType
 0.915        1.531   1.523        0.995  200000   100      50       32        250     8 bits     4583     12.41      16113.44           18.41             1          751.21       735.474      149.536       HNSW
```

candidate (baseline + the following JVM arguments)

```
-XX:CompileCommand=dontinline,org.apache.lucene.internal.vectorization.PanamaVectorUtilSupport$ArrayLoader::*
-XX:CompileCommand=dontinline,org.apache.lucene.internal.vectorization.PanamaVectorUtilSupport$MemorySegmentLoader::*
```

```
recall  latency(ms)  netCPU  avgCpuCount    nDoc  topK  fanout  maxConn  beamWidth  quantized  visited  index(s)  index_docs/s  force_merge(s)  num_segments  index_size(MB)  vec_disk(MB)  vec_RAM(MB)  indexType
 0.917        9.765   9.745        0.998  200000   100      50       32        250     8 bits     4592     13.03      15346.84           96.83             1          751.29       735.474      149.536       HNSW
```

PR #15295, opened 2025-10-07
Switching to a fixed CFS threshold

shubhamsrkdev:
### Problem
Lucene uses CFS ratio (which is by default 10%) to determine whether to use Compound Files.

### Solution
In this PR we are doing the following:

- Switching from a CFS ratio to a fixed threshold which is:

  - 64MB for byte-size-based merge policies
  - 65,536 docs for doc-based merge policies

- Moved CFS configuration to CompoundFormat.java from Merge policies.

Fixes #14959


lucene/core/src/java/org/apache/lucene/codecs/CompoundFormat.java
```diff
@@ -34,6 +36,152 @@ protected CompoundFormat() {}
...
+      return false;
+    }
+
+    // Check if segment exceeds maximum allowed size for CFS
+    if (mergedInfoSize > maxCFSSegmentSize) {
+      return false;
+    }
+
+    // Apply appropriate threshold based on merge policy type
+    if (mergePolicy instanceof LogDocMergePolicy) {
```
epotyom: It would be great if we can avoid customizing it for specific policies, otherwise it might be tricky to maintain in the future, if e.g. there is another policy that is based on doc size not bytes.

Maybe we can add a enum and a method to MergePolicy which returns its unit (`bytes`/`docs`) , and use it here to decide which threshold to use?

Or do we want to always choose compound format based on size in bytes even for LogDocMergePolicy? In this case we might be able to use `merge.getMergeInfo().sizeInBytes()` when we call this method and avoid relying on MergePolicy#size all together?

shubhamsrkdev: +1 to the idea of Enums, I will wait if anyone else has other suggestions here but having an enum makes most sense to me.

lucene/core/src/java/org/apache/lucene/codecs/CompoundFormat.java
```diff
@@ -34,6 +36,152 @@ protected CompoundFormat() {}
   // TODO: this is very minimal. If we need more methods,
   // we can add 'producer' classes.
 
+  /** Default document count threshold for using compound files with LogDocMergePolicy */
+  static final int DEFAULT_CFS_THRESHOLD_DOC_SIZE = 65536; // docs
```
stefanvodita, line 39: Why do we believe these are good thresholds? Do we need a follow-up item about tweaking them?

lucene/core/src/java/org/apache/lucene/index/LiveIndexWriterConfig.java
```diff
@@ -368,9 +368,7 @@ public InfoStream getInfoStream() {
    *
    * <p>Use <code>false</code> for batch indexing with very large ram buffer settings.
    *
-   * <p><b>Note: To control compound file usage during segment merges see {@link
-   * MergePolicy#setNoCFSRatio(double)} and {@link MergePolicy#setMaxCFSSegmentSizeMB(double)}. This
-   * setting only applies to newly created segments.</b>
+   * <p><b>Note: To control compound file usage during segment merges.</b>
```
stefanvodita: Do we want to say more in this comment?

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
```diff
@@ -5336,7 +5342,13 @@ public int length() {
...
       boolean useCompoundFile;
       synchronized (this) { // Guard segmentInfos
-        useCompoundFile = mergePolicy.useCompoundFile(segmentInfos, merge.info, this);
+        useCompoundFile =
+            merge
+                .getMergeInfo()
+                .info
+                .getCodec()
+                .compoundFormat()
+                .useCompoundFile(mergePolicy.size(merge.info, this), mergePolicy);
```
stefanvodita: Nit: Sometimes we call `getMergeInfo`, sometimes we address `info` directly.

lucene/core/src/test/org/apache/lucene/document/TestFeatureField.java
```diff
@@ -62,9 +62,13 @@ public void testBasics() throws Exception {
     Directory dir = newDirectory();
     RandomIndexWriter writer =
         new RandomIndexWriter(
-            random(),
-            dir,
-            newIndexWriterConfig().setMergePolicy(newLogMergePolicy(random().nextBoolean())));
+            random(), dir, newIndexWriterConfig().setMergePolicy(newLogMergePolicy()));
+    writer
```
stefanvodita, line 66: A lot of times we chain these access methods to eventually set compound files. I don't think it's a blocker, but maybe in the future we might add some convenience methods. No strong opinion on it though, I might actually prefer the way it's done now.

lucene/core/src/test/org/apache/lucene/index/TestAddIndexes.java
```diff
@@ -1450,11 +1455,11 @@ public void testNonCFSLeftovers() throws Exception {
...
     IndexWriterConfig conf =
-        new IndexWriterConfig(new MockAnalyzer(random())).setMergePolicy(newLogMergePolicy(true));
-    MergePolicy lmp = conf.getMergePolicy();
+        new IndexWriterConfig(new MockAnalyzer(random())).setMergePolicy(newLogMergePolicy());
     // Force creation of CFS:
-    lmp.setNoCFSRatio(1.0);
-    lmp.setMaxCFSSegmentSizeMB(Double.POSITIVE_INFINITY);
+    conf.getCodec().compoundFormat().setShouldUseCompoundFile(true);
+    conf.getCodec().compoundFormat().setMaxCFSSegmentSizeMB(Double.POSITIVE_INFINITY);
+    conf.setUseCompoundFile(true);
```
stefanvodita: Wouldn't we still benefit from the `true` default?

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterDelete.java
```diff
@@ -862,9 +862,6 @@ public void eval(MockDirectoryWrapper dir) throws IOException {
                 .setReaderPooling(false)
                 .setMergePolicy(newLogMergePolicy()));
 
-    MergePolicy lmp = modifier.getConfig().getMergePolicy();
-    lmp.setNoCFSRatio(1.0);
```
stefanvodita, line 866: We don't need to replicate this?

lucene/test-framework/src/java/org/apache/lucene/tests/util/LuceneTestCase.java
```diff
@@ -3149,7 +3132,6 @@ protected static IndexWriterConfig ensureSaneIWCOnNightly(IndexWriterConfig conf
       // and might use many per-field codecs. turn on CFS for IW flushes
       // and ensure CFS ratio is reasonable to keep it contained.
       conf.setUseCompoundFile(true);
-      mp.setNoCFSRatio(Math.max(0.25d, mp.getNoCFSRatio()));
```
stefanvodita, line 3222: Do we have an equivalent setting?

stefanvodita (commented):
I had looked at this PR before, but I see it's gone stale. Hoping to revive it.
Surprised how many changes we needed for this!

lucene/core/src/java/org/apache/lucene/codecs/CompoundFormat.java
shubhamsrkdev: I took them directly from the original issue, but makes sense to have a followup item to tweak them, I can create one.

lucene/test-framework/src/java/org/apache/lucene/tests/util/LuceneTestCase.java
shubhamsrkdev: We don't need to - by default, it will be a fixed threshold.

lucene/core/src/java/org/apache/lucene/index/IndexWriter.java
shubhamsrkdev: Hmm, makes sense, I prefer accessing through methods, let me change this.

lucene/core/src/test/org/apache/lucene/index/TestAddIndexes.java
shubhamsrkdev: Right, removing.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterDelete.java
shubhamsrkdev: No, as we have switched to a fix threshold.

lucene/core/src/java/org/apache/lucene/index/LiveIndexWriterConfig.java
shubhamsrkdev: Added path to the CF file.

lucene/core/src/java/org/apache/lucene/index/LiveIndexWriterConfig.java
```diff
@@ -368,9 +368,8 @@ public InfoStream getInfoStream() {
    *
    * <p>Use <code>false</code> for batch indexing with very large ram buffer settings.
    *
-   * <p><b>Note: To control compound file usage during segment merges see {@link
-   * MergePolicy#setNoCFSRatio(double)} and {@link MergePolicy#setMaxCFSSegmentSizeMB(double)}. This
-   * setting only applies to newly created segments.</b>
+   * <p><b>Note: To control compound file usage during segment merges. More here:
+   * lucene/core/src/java/org/apache/lucene/codecs/CompoundFormat.java</b>
```
stefanvodita: Can we use the `@link` syntax?

stefanvodita (commented):
Thank you for addressing my comments! There are two questions left for me from the comments:
1. Whether the thresholds we're starting with are good enough for a first iteration.
2. Whether we can avoid branching logic based on which merge policy is in use.

lucene/core/src/java/org/apache/lucene/index/LiveIndexWriterConfig.java
shubhamsrkdev: Done!

lucene/test-framework/src/java/org/apache/lucene/tests/index/BaseCompressingDocValuesFormatTestCase.java
```diff
@@ -46,6 +46,7 @@ static long dirSize(Directory d) throws IOException {
   public void testUniqueValuesCompression() throws IOException {
     try (final Directory dir = new ByteBuffersDirectory()) {
       final IndexWriterConfig iwc = new IndexWriterConfig(new MockAnalyzer(random()));
+      iwc.getCodec().compoundFormat().setShouldUseCompoundFile(false);
```
shubhamsrkdev, line 49: We had to disable compound files here. The reason being, before this change we had a CFS ratio (10%) which in this case would not result in compound file creation. However since we switched to a fixed threshold (64 MB now) these files are way below the threshold leading to compound file creation and thus increase in file sizes. The increase is minimal (4 bytes for me below the original threshold) but enough to fail the tests. Threefore, I have disabled compound files here.

stefanvodita (approved):
Thank you for following up on the remaining comments. I'll try to merge soon to avoid more merge conflicts for this large PR.

romseygeek:
This looks like it's causing failures in some backwards-codecs tests:

```
TestLucene87StoredFieldsFormatHighCompression > testMergeStability FAILED
--
java.lang.AssertionError: expected:<{fnm=77, fdt=37004, fdx=64, fdm=159}> but was:<{cfs=37366, cfe=150}>
at __randomizedtesting.SeedInfo.seed([DF7E25E7C981E11B:AB3263C8C46BE3AD]:0)
at junit@4.13.2/org.junit.Assert.fail(Assert.java:89)
at junit@4.13.2/org.junit.Assert.failNotEquals(Assert.java:835)
at junit@4.13.2/org.junit.Assert.assertEquals(Assert.java:120)
at junit@4.13.2/org.junit.Assert.assertEquals(Assert.java:146)
at org.apache.lucene.test_framework@11.0.0-SNAPSHOT/org.apache.lucene.tests.index.BaseIndexFileFormatTestCase.testMergeStability(BaseIndexFileFormatTestCase.java:306)
```

```
gradlew test --tests TestLucene87StoredFieldsFormatHighCompression.testMergeStability -Dtests.seed=DF7E25E7C981E11B -Dtests.nightly=true -Dtests.locale=ig-Latn-NG -Dtests.timezone=Asia/Qostanay -Dtests.asserts=true -Dtests.file.encoding=UTF-8
```

I think these are nightly-only so may have been missed by normal test runs.

stefanvodita:
Thank you for pointing out the issue @romseygeek and thank you for addressing it quickly @shubhamsrkdev!

mikemccand:
This change is 11.0 (`main` branch) only right?  Are we making a 10.x change to also introduce this new API, and deprecate the old one, so users have migration phase?  And did we update `MIGRATE.txt` in `main` to explain the migration?  Sorry I haven't dug into the change to answer for myself... just wanted to get these questions out there.

shubhamsrkdev:
> Are we making a 10.x change to also introduce this new API, and deprecate the old one, so users have migration phase? 

Hmm, good catch! I can add the previous methods with `@Deprecated` annotation and add the new methods.

> And did we update MIGRATE.txt in main to explain the migration? 

I have it ready, was planning to raise a PR soon, will do so!

PR #15305, opened 2025-10-08
Modernize and move error prone gradle plugin to Java

dweiss:
This is yet another round of the same but this one is more complex. I've rewritten the plugin for error prone so that it always cross-checks the list of checks we have declared in the plugin against the imported error-prone version. So when you upgrade, it'll fail if something is removed or added.


build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/java/ErrorPronePlugin.java
```diff
@@ -0,0 +1,962 @@
...
+
+    // double check we have all the checks from the current version of errorprone covered.
+    private void ensureCheckListIsComplete() {
+      Set<String> allCanonicalNames = new TreeSet<>();
+      Map<String, String> altToCanonical = new HashMap<>();
+
+      Stream.of(
+              BuiltInCheckerSuppliers.ENABLED_ERRORS,
+              BuiltInCheckerSuppliers.ENABLED_WARNINGS,
+              BuiltInCheckerSuppliers.DISABLED_CHECKS)
```
dweiss, line 138: I added all checks - even those disabled by default in error-prone. Those that were missing are turned off in the list below too.

PR #15310, opened 2025-10-08
test: remove slow test for java 21 bug

rmuir:
This test never runs on my system (I never let tests get a C2, its too slow). But its very slow in CI: probably because it makes many writers, probably fsyncs a lot, etc. Given that `main` is on java 25 I think we can remove the java 21 test?


dweiss (approved):
Nuke it.

PR #15320, opened 2025-10-09
build: gradle seems to not run error-prone

rmuir:
Enabling a bunch of checks that should probably be causing failures. I can't see what is happening, and gradle as usual is 100% mute :(

Something isn't right I think, maybe flags are not being passed to compiler?

I insist this build should be failing... I turned on several noisy checks and ran a GH actions job to confirm i'm not losing my mind.

My local cmdline: 
```
./gradlew clean
./gradlew check -x test -Pvalidation.errorprone=true
```

Background: I'm trying to evaluate class deadlock detector to see what needs to be done to prevent these...


dweiss:
Let me see, could be related to my recent changes to the plugin (although I did check it worked for me).

dweiss:
Yes, my bad. I'll fix it asap.

dweiss:
I did check it. And then I did some last minute cleanups that added this beauty:
```
-          .filter(e -> e.getValue() == CheckSeverity.OFF)
+          .filter(e -> e.getValue() != CheckSeverity.OFF)
```

Eh. I've committed a fix to main.

dweiss:
Fixed by 8f68736e75609d13053420450ad451e52cba107d

rmuir:
thank you!!!

PR #15322, opened 2025-10-10
build: enable ClassInitializationDeadlock, SuppressWarnings the problems

rmuir:
Enable error-prone's ClassInitializationDeadlock check, which can statically detect bugs such as #15317

All problems were just suppressed with a 'FIXME' message, this check will at least prevent any new ones from showing up.

There aren't that many problems, it just happens that some are caused by generated code, so you get a violation for each bit-per-value.


dweiss (approved):

ChrisHegarty:
Awesome! Thanks @rmuir

PR #15326, opened 2025-10-10
Add matomo tracker for javadoc

pabloem:

### DIFF
--- a/build-tools/build-infra/src/main/groovy/lucene.documentation.render-javadoc.gradle
+++ b/build-tools/build-infra/src/main/groovy/lucene.documentation.render-javadoc.gradle
@@ -391,7 +391,23 @@ abstract class RenderJavadocTask extends RenderJavadocTaskBase {
     LuceneBuildGlobalsExtension buildGlobals = project.rootProject.extensions.getByType(LuceneBuildGlobalsExtension)
     opts << [
       '-bottom',
-      "<i>Copyright &copy; 2000-${buildGlobals.buildYear} Apache Software Foundation. All Rights Reserved.</i>"
+      """<!-- Matomo -->
+<script>
+  var _paq = window._paq = window._paq || [];
+  /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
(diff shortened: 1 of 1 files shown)


dweiss:
I don't think we should add such trackers to javadoc. To the website - yes, but not to the javadoc. I don't think it'll be useful.

rmuir:
Maybe it could tell us something about the versions of the documentation being looked up

dweiss:
I'm reluctant about this. It gets added to all javadocs everywhere - even if you generate them locally. I'm not so fond of those trackers, maybe that's why I feel uneasy about adding them here.

rmuir:
If you install ublock origin lite, the network request will never even happen.

dweiss:
I do have it. It's in my head though (the tracker).

pabloem:
hi folks!
I was adding this since the website is minimal, and I figured we'd
understand the usage patterns of the docs better (e.g. most useful classes
/ versions being used) - however I understand your reluctance. Also, I
agree it wouldn't make sense to have reporting for local builds of the
javadoc.

A counter-argument for the version question (having used versioned javadocs
myself) is that I often don't care much about the version and will just
click whatever Google gives me unless the thing I'm using was added
recently.

Happy to wait for more feedback, or just close this.

dweiss:
Would it be ok to add it optionally? So, for example, we can add it only to releases?

pabloem:
I moved the tracker to be added when building release javadoc! (depending on lucenedocurl, which is only defined on releases). LMK what y'all think @dweiss

pabloem:
thoughts @dweiss ?

dweiss (approved):
I still don't see the need for this in javadocs but if it helps anybody - here's my approval to merge.

dweiss:
Hmm... I had to revert this as it caused build failures. I'll take another look at why it's so.

dweiss:
I did take another look.

Javadoc's parameters are simple strings or values passed via @file.options. There are a lot of quirks around escaping and this tracking code is full of quote/double quotes, etc. I don't know how to handle this reliably, sorry.

PR #15331, opened 2025-10-13
fix(apache#15309): Optimize PriorityQueue.remove() function

vijaykriishna:

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/PriorityQueue.java
+++ b/lucene/core/src/java/org/apache/lucene/util/PriorityQueue.java
@@ -19,8 +19,12 @@ package org.apache.lucene.util;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
+import java.util.Map;
 import java.util.NoSuchElementException;
+import java.util.Set;
 import java.util.function.IntFunction;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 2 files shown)


dweiss:
PQs are so heavily used in Lucene that any extra cost added to add() will be probably a no-no. Can you run the benchmarks and see how it affects performance?

benwtrent:
I don't think any code actually uses remove...I don't think making it faster is worth making anything else more expensive.

vijaykriishna:
@dweiss @benwtrent Are you suggesting to get rid of `remove` function once and for all ?

vijaykriishna:
`PriorityQueue.remove` function is used by DiversifiedTopDocsCollector. Ref. https://github.com/apache/lucene/blob/71e822e6240878018a6ff3c28381a0d88bebdc72/lucene/misc/src/java/org/apache/lucene/misc/search/DiversifiedTopDocsCollector.java#L154.

Hence, felt to not remove it altogether.

benwtrent:
@vijaykriishna I would much prefer the single use gets optimized than adding cost to the priority queue to everything.

dweiss:
Yes, I think we should remove it. It's only used in one place and this place could probably do away with something else instead.

dweiss:
@vijaykriishna if you'd like to continue - feel free to review DiversifiedTopDocsCollector to see if it can be implemented in a different way, then nuke that remove method from the PriorityQueue class! I'll be away this week but it'd be great to push this forward.

dweiss:
Superseded by #15493

PR #15332, opened 2025-10-13
Add PhraseQuery.Builder.setMaxTerms() method to limit the maximum number of terms and excessive memory use

linyunanit:
### Description

Add configurable term threshold for PhraseQuery#Builder to solve the problem of excessive memory usage in ultra-long text search case.

### Key Changes
- Add termThreshold variables  in PhraseQuery.Builder.
- Check whether the current terms size reaches the threshold.

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/PhraseQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/PhraseQuery.java
@@ -74,12 +74,14 @@ public class PhraseQuery extends Query {
   public static class Builder {
 
     private int slop;
+    private int termLimit;
     private final List<Term> terms;
     private final IntArrayList positions;
 
     /** Sole constructor. */
     public Builder() {
       slop = 0;
+      termLimit = -1;
       terms = new ArrayList<>();
       positions = new IntArrayList();
     }
@@ -91,6 +93,14 @@ public class PhraseQuery extends Query {
      */
     public Builder setSlop(int slop) {
       this.slop = slop;
+      return this;
+    }
+
+    /**
+     * Set the term limit.
+     */
+    public Builder setTermLimit(int value) {
+      this.termLimit = value;
       return this;
     }
 
@@ -127,6 +137,10 @@ public class PhraseQuery extends Query {
                 + term.field()
                 + " and "
                 + terms.get(0).field());
+      }
+      if (termLimit > 0 && terms.size() >= termLimit) {
+        throw new IllegalArgumentException("The current value of terms is " +
+            terms.size() + ", which exceeds the limit of " + termLimit);
       }
       terms.add(term);
       positions.add(position);
--- a/lucene/core/src/java/org/apache/lucene/search/PhraseQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/PhraseQuery.java
@@ -97,7 +97,7 @@ public class PhraseQuery extends Query {
     }
 
     /**
-     * Set the term limit.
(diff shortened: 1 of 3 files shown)


lucene/core/src/java/org/apache/lucene/search/PhraseQuery.java
```diff
@@ -128,6 +136,13 @@ public Builder add(Term term, int position) {
                 + " and "
                 + terms.get(0).field());
       }
+      if (termThreshold > 0 && terms.size() >= termThreshold) {
+        throw new IllegalArgumentException(
+            "The current value of terms is "
```
dweiss:
```suggestion
            "The current number of terms is "
```

lucene/core/src/test/org/apache/lucene/search/TestPhraseQuery.java
```diff
@@ -743,6 +743,20 @@ public void testBackwardPositions() throws Exception {
         });
   }
 
+  public void testPhraseQueryTermLimit() throws Exception {
+    PhraseQuery.Builder builder = new PhraseQuery.Builder();
+    int termLimit = 1000;
```
dweiss: Use a smaller termLimit (5) - it makes little sense to make the test run for 1000 terms. Also, I'd change the loops to not use positions 0..termLimit-1 and then check for failure on termLimit - seems more intuitive.

dweiss (changes requested):
Please add changes entry as well.

lucene/core/src/test/org/apache/lucene/search/TestPhraseQuery.java
linyunanit: Good points, @dweiss! I've made the following changes based on your feedback:
- Reduced `termLimit` from 1000 to 5 to make the test more lightweight
- Changed the position logic to use `0..termLimit-1` and check for failure at position `termLimit`

This does make the test more intuitive. Thanks for the review!

lucene/core/src/java/org/apache/lucene/search/PhraseQuery.java
linyunanit: Thanks for the suggestion, Dawid! 👍

dweiss (approved):

dweiss (commented):
I'm sorry this comes in multiple requests. Could we also rename the issue and changes entry to make it clear to people as to what's changed? I'd suggest "Add PhraseQuery.Builder.setMaxTerms() method to limit the maximum number of terms and excessive memory use." 

I would also rename the method accordingly and add javadocs to it that say what happens when you exceed the threshold (IllegalArgumentException is thrown). Thank you.

linyunanit:
> I'm sorry this comes in multiple requests. Could we also rename the issue and changes entry to make it clear to people as to what's changed? I'd suggest "Add PhraseQuery.Builder.setMaxTerms() method to limit the maximum number of terms and excessive memory use."
> 
> I would also rename the method accordingly and add javadocs to it that say what happens when you exceed the threshold (IllegalArgumentException is thrown). Thank you.

Thanks for the review, @dweiss! Great suggestions. I've made the following changes:

1. **Renamed the method**: Changed `setTermThreshold(int value)` to `setMaxTerms(int maxTerms)` as suggested.
2. **Updated Javadoc**: Added clear documentation explaining that `IllegalArgumentException` is thrown when the threshold is exceeded.
3. **Updated the issue/changes entry**: Modified to make it clear to people as to what's changed.

Looking forward to getting this merged!

PR #15334, opened 2025-10-13
Fix potential NPE for KnnVectorQuery

benwtrent:
When using a particular leaf reader (possibly for a MemoryIndex), `parent` can be `null`. 

Let's verify that and prevent an NPE

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
+++ b/lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
@@ -262,6 +262,7 @@ abstract class AbstractKnnVectorQuery extends Query {
         int visitedLimit, KnnSearchStrategy searchStrategy, LeafReaderContext context)
         throws IOException {
       // The delegate supports optimistic collection
+      // ctx.parent could be null if this is a MemoryIndex
       if (delegate.isOptimistic() && context.parent != null) {
         @SuppressWarnings("resource")
         float leafProportion = context.reader().maxDoc() / (float) context.parent.reader().maxDoc();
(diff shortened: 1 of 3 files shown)


lucene/core/src/java/org/apache/lucene/search/AbstractKnnVectorQuery.java
```diff
@@ -255,7 +262,7 @@ public KnnCollector newCollector(
         int visitedLimit, KnnSearchStrategy searchStrategy, LeafReaderContext context)
         throws IOException {
       // The delegate supports optimistic collection
-      if (delegate.isOptimistic()) {
+      if (delegate.isOptimistic() && context.parent != null) {
```
jainankitk, line 266: I am assuming the same reasoning applies here as well? `// ctx.parent could be null if this is a MemoryIndex`. If yes, can be add the comment here as well?

jainankitk (approved):

msokolov:
If this is a real problem, why hasn't it surfaced in unit test failures? Or ... maybe it did?

benwtrent:
@msokolov I am not sure, I noticed it in ES:

https://github.com/elastic/elasticsearch/issues/136427

Maybe we just exit early there (which I think we will do, e.g. InMemoryIndex just cannot have kNN happen), but it seems to me better to fix it here as well.

msokolov:
I guess what I'm wondering is does this point a testing gap in Lucene? I thought that we would randomize the index directory type when running tests ...

benwtrent:
Looking at MemoryIndex:
```
    @Override
    public void searchNearestVectors(
        String field, float[] target, KnnCollector knnCollector, AcceptDocs acceptDocs) {}

    @Override
    public void searchNearestVectors(
        String field, byte[] target, KnnCollector knnCollector, AcceptDocs acceptDocs) {}
```

So, returning no hits and forcing exact (or the caller to change), is the way. But we should definitely not throw an NPE. I can add a test really quick to make sure this won't throw an NPE.

PR #15340, opened 2025-10-16
Support multiple delimiters for path tokenization

dadoonet:
### Description

This adds a new ctor for the `PathHierarchyTokenizer`, which supports an array of delimiters instead of a single one: `PathHierarchyTokenizer(AttributeFactory, int, char[], char, int)`.

This allows having two different delimiters instead of a single one.

Closes #15196.


dadoonet:
This is not needed as per [my comment](https://github.com/apache/lucene/issues/15196#issuecomment-3432392265) on #15196.

PR #15342, opened 2025-10-17
ci: fix missing permission for the stale workflow

rmuir:
Permissions were restricted (following documentation of actions/stale) in c5d82423934db5e5512b83d6319944be3f33e221, but the `actions: write` is missing

This permission is only mentioned in the top of documentation along with many other "optional" permissions that depend upon which features you use... but it is needed for the action to manipulate the actions cache the way that it does.

When operations limit is exceeded, it "stops" and "picks up from where it left off" in the next run, and actions cache is used to do this.


stefanvodita (approved):
Good find! Thank you for fixing this.

PR #15343, opened 2025-10-20
Use cached cardinality for AcceptDocs cost

benwtrent:
We are unnecessarily calling `BitSet.cardinality` more than once when returning the cost from `AcceptDocs`.


kaivalnp (approved):
Makes sense to me!

PR #15350, opened 2025-10-22
Move documentation gradle scripts to java

dweiss:
This is larger and not so straightforward. I didn't try to improve it a lot but there are some minor tweaks here and there.


build-tools/build-infra/src/main/java/org/apache/lucene/gradle/plugins/java/RenderJavadocPlugin.java
```diff
@@ -0,0 +1,766 @@
...
+            "prettify/prettify.css");
+
+        // append prettify to scripts
+        Provider<RegularFile> customScript = getOutputDir().file("script-files/lucene-script.js");
+        concat(
+            customScript, getTaskResources().dir("prettify"), "prettify.js", "inject-javadocs.js");
+
+        opts.add(List.of("--add-script", customScript.get().getAsFile().toString()));
+        opts.add(List.of("--add-stylesheet", customCss.get().getAsFile().toString()));
+      }
```
dweiss, line 578: This is new - it uses standard doclet's options to pass custom script/css instead of appending over existing scripts (which move around from jdk to jdk version).

PR #15356, opened 2025-10-23
Improve docs for some internal classes related to https://github.com/apache/lucene/issues/15225

mpbarano:
### Description

This change adds documentation to 4 @lucene.internal classes in the geo package.

This is my first contribution to Lucene. I have read the CONTRIBUTING.md. This change is part of https://github.com/apache/lucene/issues/15225. This change started with Cursor LLM, but I have read everything. I have tried my best to validate that the information is accurate. And I altered the language to remove unnecessary adverbs or other awkward language.

The provided docs are verbose, but they seem useful. I would be happy to have feedback about whether the amount of content is appropriate or whether they should be more terse.

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -317,6 +317,8 @@ Improvements
 * GITHUB#15063: Use optimistic-with-checking KNN Query strategy for `DiversifyingChildren*` queries.
   (Ben Trent)
 
+* GITHUB#15356: Improve docs for 4 internal classes in geo package (Matt Baranowski)
+
 Optimizations
 ---------------------
 * GITHUB#14932: Switched to GroupVarInt Encoding for HNSW Graph edges, added backwards compatibility (Akira Lonske)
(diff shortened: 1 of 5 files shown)


dweiss (approved):

dweiss:
I agree these docs are... verbose. I keep wondering if it makes sense to add llm-generated documentation back to the project - seems like we'd be feeding the snake with its own tail if there's something wrong in there. :) This seems ok though.

jainankitk (approved):
Looks fine to me. Although, I am wondering if we really need verbose documentation for classes annotated ` @lucene.internal`. But, I guess it shouldn't hurt!

jainankitk:
Okay, I was wondering why the github workflow did not automatically add milestone even though CHANGES entry is present. It seems to be added in the 10.3.0 section. @mpbarano - Can you move the CHANGES entry to 10.4.0 section?

benwtrent:
I am not gonna block this, but this list of "key features", and the repeating of concepts over and over again (e.g. Haversin Distance) smell of LLMs. 

LLMs continually over-repeat things because it's cheap for them. It's not for humans.

Also, the `containsLine`, etc. should be comments on those methods themselves or something. I am not sure why every possible method is enumerated

dweiss:
Thank you for your contribution but the feeling seems to be that AI-generated docs belong better downstream - where they can be regenerated as the code evolves. Let's try to keep comments inside the codebase less verbose and more to the point.

PR #15375, opened 2025-10-28
Rewrote more Gradle plugins to java

dweiss:
This rewrites even more gradle scripts/ plugins to Java. Includes a clean rewrite of jar checksum/ license validation and other tweaks.


rmuir (approved):
Thanks again for these, it really makes the build more accessible

Issue #15197, opened 2025-09-17
Ways to implement prefetching in BKD trees


abiesps:
I see that Lucene has added prefetching    https://github.com/apache/lucene/issues/13179 in number of data structures to decouple search concurrency with IO concurrency.

In the linked issue, there is an unchecked item on prefetching support to points. I want to know if lucene is thinking of adding prefetching support to BKD trees in near future. 

I was also exploring the ways to achieve this. At a high level, I was thinking that for range queries, point-lookup etc we can traverse the BKD tree in same fashion as we do. But instead of visiting leaf nodes (kdd files). We can call prefetch on matching leaf nodes file pointers. Iterator can also save the matching leaf nodes during traversal. 
Once the traversal of BKD tree is complete, we can actually make a call to visit doc IDs or doc Values from the matching leaf nodes. Hopefully they may be prefetched by this time. 

Opensearch uses similar variation of BKD tree traversal for range with an exception of early termination, if the matching doc id count becomes greater than equal to the size parameter requested in the query. 
For such use cases to support early termination, I was thinking of adding a vInt to inner nodes (kdi files) that denotes the total number of documents below that node. 

Let me know if this make sense or what does community feel about this.

abiesps:
I dont think we need to store matching doc count in inner nodes, probably can get it from size(). Exploring this more for my poc.

abiesps:
Another thing that I realised,  especially relevant for early termination is :
 leafNode may not always exactly match thats why we have two functions in BKDReader visitDocIDs (for the case of cell inside query) and visitDocValues (for the case of cell crossing the query). visitDocValues is the case where leaf node may not match in case of more than one dimension in BKD tree. And we know for sure by actually reading the leaf Node (I/O). Whereas in case of visitDocIDs, all document in the leaf always matches the range in the query. 

If we don't want early termination - then prefetching leaf nodes from both kind of matches is fine, otherwise we can keep on incrementing match count only in case of "r == PointValues.Relation.CELL_INSIDE_QUERY". With this we may do some wasteful traversal of BKD tree but saving on doing IO in query path (through prefetching leaf nodes). 

And size() returns matching points not necessarily doc count.

With above consideration, let me explore more on this and raise a draft PR in couple of days or so.

abiesps:
Ok, I got a working solution. Code is really in very early stage (I need to clean up a lot) - I need more time to raise a PR. 

**Approach** 

At a high level, this is what I am doing:  
[...]

Issue #10659, opened 2020-11-22
Move Points from a visitor API to a cursor-style API? [LUCENE-9619]


asfimport:
Points' visitor API work well but there are a couple things we could make better if we moved to a cursor API, e.g.
- Term queries could return a DocIdSetIterator without having to materialize a BitSet.
- Nearest-neighbor search could work on top of the regular API instead of casting to BKDReader https://github.com/apache/lucene-solr/blob/6a7131ee246d700c2436a85ddc537575de2aeacf/lucene/sandbox/src/java/org/apache/lucene/sandbox/document/FloatPointNearestNeighbor.java#L296
- We could optimize counting the number of matches of a query by adding the number of points in a leaf without visiting documents where there are no deleted documents and a leaf fully matches the query.

---
Migrated from [LUCENE-9619](https://issues.apache.org/jira/browse/LUCENE-9619) by Adrien Grand (@jpountz), updated Dec 06 2021
Pull requests: https://github.com/apache/lucene/pull/486

jpountz:
This has been implemented.

PR #15376, opened 2025-10-28
Proposal for alternate way to traverse BKD tree with prefetching matching leaves for IO concurrency

abiesps:
### Description

I would appreciate feedback on BKD traversal with prefetching. 

At a high level, this is what I am doing:

Start traversing the BKD tree as is,

If cell is inside the query
a) Call goes to prepareDocIDs method variation where if isLeaf is true. 
b) Do not read the leaf node yet. Instead, get the leafOrdinal, if this leaf ordinal is in continuation to last matching leaf ordinal (which I am storing in visitor) i.e leafOrdinal == visitor.lastMatchingLeafOrdinal() + 1. Do not call prefetch as this should be hopefully taken care by kernel readaheads. Otherwise if its not a continuous match i.e leafOrdinal != visitor.lastMatchingLeafOrdinal() + 1, then call prefetch on this leaf node file pointer. Also store this leaf node fp in visitor for visting matching doc IDs later.
c) If its not a leafNode continue with recursion as is.
d) Visit the cached leaf node file pointers after recursion.

For remaining traversal code remains unchanged.

[Related Issue](https://github.com/apache/lucene/issues/15197)

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
+++ b/lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
@@ -17,7 +17,9 @@
 package org.apache.lucene.util.bkd;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 import org.apache.lucene.codecs.CodecUtil;
 import org.apache.lucene.index.CorruptIndexException;
 import org.apache.lucene.index.PointValues;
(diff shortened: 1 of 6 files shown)


abiesps:
I am looking for overall feedback on high level idea for this. I do want to make the api interfaces added/modified more cleaner and possibly minimal.
Based on the directional feedback, I'll also add more tests

iverase:
In general, I am not in favour of adding a new intersects method to the PointValues API like you are doing with `PointValues#intersectWithPrefetch`. Prefetching should be an implementation detail and not leak to the API.

abiesps:
Thanks for the feedback @iverase !! I'll hide prefetching detail into existing intersect api. That would need adding some new methods to existing Visitor.

iverase:
> That would need adding some new methods to existing Visitor.

As far as we can add some default implementation, we should be ok.

abiesps:
@iverase I removed the new intersect api in PointValues. However I have to create a new visitor TwoPhaseVisitor because I want to cache the prefetched leaf blocks so that I can visit them after recursion. This is a new abstract class with some state. Let me know your thoughts on this.

iverase:
I have been thinking in this change and I think the approach makes too many assumptions about how points are implemented under the hood. In general, I don't think prefetching belongs to this API 

The PointValues API is a [cursor-style API](https://github.com/apache/lucene/issues/10659). It allows to navigate a tree-like structure and apply some operations on the nodes the "cursor" is currently positioned. This PR is introducing the concept of applying those operations on a node when we are positioned on a different node via `visitDocIDs(long pos, IntersectVisitor visitor)`. This introduced position does not belong to the current API. The TwoPhaseIntersectVisitor introduces even more stuff.

In general, if you want to introduce something like that, you need to propose a new API for the PointTree trying to be as agnostic as possible from the current implementations and reason the need for it.

IMHO, prefetching is an implementation detail of the BKD tree and most of the changes should be in that area while the changes to the PointValues API should be minimal.

abiesps:
hmm, thanks for the input @iverase . Let me try moving all the methods to BKDReader class. This would mean having the TwoPhaseIntersectVisitor definition in BKDReader and an intersect method in BKDReader that takes in TwoPhaseIntersectVisitor. 
With this change there wont be any changes to the PointValues class.

abiesps:
Let me know if above approach looks ok.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
```diff
@@ -1076,4 +1137,123 @@ public long cost() {
...
+  /**
+   * Base implementation of {@link TwoPhaseIntersectVisitor} that maintains a list of deferred
+   * blocks from first phase of traversal and visits them in the second phase.
+   *
+   * @lucene.experimental
+   */
+  public abstract static class BaseTwoPhaseIntersectVisitor implements TwoPhaseIntersectVisitor {
+
+    int lastDeferredBlockOrdinal = -1;
+    List<Long> deferredBlocks = new ArrayList<>();
```
iverase, line 1175: This is a bit worrisome. An unbounded list of boxed longs does not feel right in here.

iverase:
I am a bit torn with this change as I have never seen io access as an issue in the bkd tree implementation. Data are always read sequentially during intersects so I expect read ahead to be enough. I would like to see some benchmarks that prove the benefits of this approach.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
abiesps: This list can grow to a maximum size of (num of leaves / 2) +1 in worst case.

abiesps:
let me get back on benchmarks.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
abiesps: Actually list can grow upto total number of leaf nodes in worst case. Num of leaves /2 +1 is the maximum number of prefetch calls in worst case.

lucene/core/src/java/org/apache/lucene/util/bkd/BKDReader.java
```diff
@@ -1076,4 +1137,123 @@ public long cost() {
...
+  /**
+   * Finds all documents and points matching the provided visitor. This method does not enforce live
+   * documents, so it's up to the caller to test whether each document is deleted, if necessary.
+   */
+  @Override
+  public final void intersect(IntersectVisitor visitor) throws IOException {
+    final BKDPointTree pointTree = (BKDPointTree) getPointTree();
+    if (visitor instanceof TwoPhaseIntersectVisitor twoPhaseIntersectVisitor) {
+      intersect(twoPhaseIntersectVisitor, pointTree);
+      List<Long> fps = twoPhaseIntersectVisitor.deferredBlocks();
```
navneet1v, line 1220: instead of using `List<Long>` I think it will be good to use `LongArrayList` which uses native arrays to build the list which can avoid autoBoxing of long.

PR #15378, opened 2025-10-28
Return MergeObserver from IndexWriter.forceMergeDeletes() (#14515)

### LINKED ISSUE #14515: IndexWriter forceMergeDeletes should return its MergeSpec

salvatorecampagna:
## Summary

Changes `IndexWriter.forceMergeDeletes()` to return `MergePolicy.MergeObserver` instead of void, allowing applications to monitor merge progress and wait for completion without exposing mutable internal objects.

Fixes #14515

## Motivation

Currently when calling `forceMergeDeletes(false)`, there's no way to monitor whether merges were scheduled or wait for them to complete. This makes it difficult to coordinate merge completion with other operations or implement custom scheduling strategies.

## Changes

Added `MergePolicy.MergeObserver` as a new public API with methods to:
- Query merge count (`numMerges()`)
- Track completion progress (`numCompletedMerges()`)
- Wait synchronously with optional timeout (`await()`, `await(timeout, unit)`)
- Wait asynchronously via CompletableFuture (`awaitAsync()`)

The observer is thread-safe and handles cases where no merges are needed. It provides observability without exposing mutable `OneMerge` objects, maintaining proper encapsulation.

Updated `IndexWriter.forceMergeDeletes()` methods to return the observer instead of void. Also updated `RandomIndexWriter` to propagate the new return type.

## Testing

Added comprehensive tests covering:
- Blocking and non-blocking merge modes
- Timeout handling with proper assertions
- Completion status verification
- Null spec cases (no merges needed)
- Progress tracking via `numCompletedMerges()`

All tests pass with randomized seeds across multiple iterations.

## Backward Compatibility

Fully backward compatible. Existing code that ignores the return value continues to work unchanged. Changing void to a return type is not a breaking change in Java.

## Files Modified

- `MergePolicy.java` - New MergeObserver class with progress tracking
- `IndexWriter.java` - Updated return types and javadoc
- `RandomIndexWriter.java` - Propagated return type
- `TestIndexWriterMerging.java` - Comprehensive test coverage
- `CHANGES.txt` - Added entry under 10.4.0 API Changes

### DIFF
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -69,12 +69,6 @@ Improvements
 
 * GITHUB#15124: Use RamUsageEstimator to calculate size for non-accountable queries. (Sagar Upadhyaya)
 
-* GITHUB#14515: IndexWriter.forceMergeDeletes() now returns MergePolicy.MergeObserver,
-  allowing applications to monitor merge progress, wait for completion (synchronously
-  via await() or asynchronously via CompletableFuture), and inspect individual merges.
-  Backward compatible - existing code that ignores the return value works unchanged.
-  (Salvatore Campagna)
-
 Optimizations
 ---------------------
 * GITHUB#14011: Reduce allocation rate in HNSW concurrent merge. (Viliam Durina)
@@ -150,6 +144,12 @@ API Changes
 
 * GITHUB#15234: Remove unused method which does unsafe XML parsing.
   (Uwe Schindler, Isaac David)
+
+* GITHUB#14515: IndexWriter.forceMergeDeletes() now returns MergePolicy.MergeObserver,
+  allowing applications to monitor merge progress and wait for completion. The observer
+  provides methods to query merge count, check completion status, and wait synchronously
+  (with optional timeout) or asynchronously via CompletableFuture. Backward compatible -
+  existing code that ignores the return value works unchanged. (Salvatore Campagna)
 
 New Features
 ---------------------
--- a/lucene/CHANGES.txt
+++ b/lucene/CHANGES.txt
@@ -69,6 +69,12 @@ Improvements
 
 * GITHUB#15124: Use RamUsageEstimator to calculate size for non-accountable queries. (Sagar Upadhyaya)
 
+* GITHUB#14515: IndexWriter.forceMergeDeletes() now returns MergePolicy.MergeObserver,
+  allowing applications to monitor merge progress, wait for completion (synchronously
+  via await() or asynchronously via CompletableFuture), and inspect individual merges.
+  Backward compatible - existing code that ignores the return value works unchanged.
+  (Salvatore Campagna)
+
 Optimizations
 ---------------------
 * GITHUB#14011: Reduce allocation rate in HNSW concurrent merge. (Viliam Durina)
--- a/lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
+++ b/lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
@@ -966,6 +966,25 @@ public abstract class MergePolicy {
     }
 
     /**
+     * Returns the number of completed merges in this specification. Useful for tracking merge
+     * progress: {@code numCompletedMerges() / numMerges()}.
+     *
+     * @return number of completed merges
+     */
+    public int numCompletedMerges() {
+      if (spec == null) {
+        return 0;
+      }
+      int completed = 0;
+      for (OneMerge merge : spec.merges) {
+        if (merge.mergeCompleted.isDone()) {
+          completed++;
+        }
+      }
+      return completed;
+    }
+
+    /**
      * Waits for all merges in this specification to complete. Returns immediately if no merges were
      * scheduled.
      *
@@ -1003,7 +1022,11 @@ public abstract class MergePolicy {
 
     @Override
     public String toString() {
-      return spec == null ? "MergeObserver: no merges" : spec.toString();
+      if (spec == null) {
+        return "MergeObserver: no merges";
+      }
+      return String.format(
+          "MergeObserver: %d merges (%d completed)", numMerges(), numCompletedMerges());
     }
   }
 }
--- a/lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
+++ b/lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.EnumMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
@@ -1025,7 +1026,7 @@ public abstract class MergePolicy {
       if (spec == null) {
         return "MergeObserver: no merges";
       }
-      return String.format("MergeObserver: %d merges", numMerges());
+      return String.format(Locale.ROOT, "MergeObserver: %d merges", numMerges());
     }
   }
 }
--- a/lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
+++ b/lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
@@ -944,9 +944,8 @@ public abstract class MergePolicy {
    * Observer for merge operations returned by {@link IndexWriter#forceMergeDeletes(boolean)}.
    * Provides methods to query merge status and wait for completion.
    *
-   * <p>When no merges are needed, {@link #hasNewMerges()} returns {@code false} and {@link
-   * #numMerges()} returns 0. In this case, {@link #await()} returns {@code true} immediately since
-   * there is nothing to wait for.
+   * <p>When no merges are needed, {@link #numMerges()} returns 0. In this case, {@link #await()}
+   * returns {@code true} immediately since there is nothing to wait for.
    *
    * @lucene.experimental
    */
@@ -964,15 +963,6 @@ public abstract class MergePolicy {
      */
     public int numMerges() {
       return spec == null ? 0 : spec.merges.size();
-    }
-
-    /**
-     * Returns whether any new merges were scheduled.
-     *
-     * @return {@code true} if merges were scheduled, {@code false} if no merges needed
-     */
-    public boolean hasNewMerges() {
-      return spec != null;
     }
 
     /**
--- a/lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
+++ b/lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
@@ -1025,8 +1025,7 @@ public abstract class MergePolicy {
       if (spec == null) {
         return "MergeObserver: no merges";
       }
-      return String.format(
-          "MergeObserver: %d merges (%d completed)", numMerges(), numCompletedMerges());
+      return String.format("MergeObserver: %d merges", numMerges());
     }
   }
 }
--- a/lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
+++ b/lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
@@ -1026,7 +1026,7 @@ public abstract class MergePolicy {
       if (spec == null) {
         return "MergeObserver: no merges";
       }
-      return String.format(Locale.ROOT, "MergeObserver: %d merges", numMerges());
+      return String.format(Locale.ROOT, "MergeObserver: %d merges\n%s", numMerges(), spec.toString());
     }
   }
 }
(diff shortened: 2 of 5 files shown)


lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -939,4 +939,96 @@ static final class MergeReader {
...
+
+    /**
+     * Returns the merge at the specified index. Caller must ensure {@link #hasNewMerges()} returns
+     * {@code true} and index is within bounds.
+     *
+     * @param i merge index (0 to {@link #numMerges()} - 1)
+     * @return the merge at index i
+     * @throws IndexOutOfBoundsException if index is invalid or no merges exist
+     */
+    public MergePolicy.OneMerge getMerge(int i) {
```
vigyasharma: I'm not sure if providing this method is really useful. The caller can't know the "index" location of a OneMerge unless they already have access to the MergeSpec.

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -939,4 +939,96 @@ static final class MergeReader {
...
+    MergeObserver(MergePolicy.MergeSpecification spec) {
+      this.spec = spec;
+    }
+
+    /**
+     * Returns the number of merges in this specification.
+     *
+     * @return number of merges, or 0 if no merges were scheduled
+     */
+    public int numMerges() {
```
vigyasharma, line 965: How about we also expose the no. of completed merges? You will need to run through all the OneMerge objects in `merges` and check for `mergeCompleted`. It could be useful to track overall merge progress.

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -939,4 +939,96 @@ static final class MergeReader {
...
+      return spec == null ? 0 : spec.merges.size();
+    }
+
+    /**
+     * Returns whether any new merges were scheduled.
+     *
+     * @return {@code true} if merges were scheduled, {@code false} if no merges needed
+     */
+    public boolean hasNewMerges() {
```
vigyasharma: I think this naming can get confusing: is it still a "new" merge once the merge is underway, does it count completed ones, etc.? Instead of this API, can we just use `numMerges() > 0` ? Or how about an `isEmpty` api?

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
vigyasharma: Also, I don't think that the `OneMerge` object is immutable?

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMerging.java
```diff
@@ -333,6 +337,330 @@ public synchronized void merge(MergeSource mergeSource, MergeTrigger trigger)
...
+
+    assertEquals(5, iw.getDocStats().maxDoc);
+    assertEquals(5, iw.getDocStats().numDocs);
+
+    iw.waitForMerges();
+    iw.close();
+    dir.close();
+  }
+
+  public void testMergeObserverGetMerges() throws IOException {
```
vigyasharma: We can remove this test if we remove the getMerges api.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMerging.java
```diff
@@ -333,6 +337,330 @@ public synchronized void merge(MergeSource mergeSource, MergeTrigger trigger)
...
+    MergePolicy.MergeObserver observer = iw.forceMergeDeletes(false);
+
+    // Measure time to detect stuck merges
+    long startNanos = System.nanoTime();
+    assertTrue("await should complete before timeout", observer.await(10, TimeUnit.MINUTES));
+    long elapsedMillis = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startNanos);
+
+    assertTrue(
+        "Merge took too long: " + elapsedMillis + "ms (expected < 30000ms)",
+        elapsedMillis < 30_000);
```
vigyasharma: Is this 30_000 a random magic number for timeout upperbound? Why don't we just use it in the await timeout value? Then `assertTrue` will fail if we don't meet this limit.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMerging.java
```diff
@@ -333,6 +337,330 @@ public synchronized void merge(MergeSource mergeSource, MergeTrigger trigger)
...
+    IndexWriter iw =
+        new IndexWriter(
+            dir,
+            newIndexWriterConfig(new MockAnalyzer(random())).setMergePolicy(newLogMergePolicy()));
+    assertEquals(10, iw.getDocStats().maxDoc);
+    assertEquals(5, iw.getDocStats().numDocs);
+
+    MergePolicy.MergeObserver observer = iw.forceMergeDeletes(true);
+    assertTrue("Should have scheduled merges", observer.hasNewMerges());
+    assertTrue("Should have completed merges", observer.numMerges() > 0);
```
vigyasharma, line 553: You can add an assert here that tests that numMerges == completedMerges

lucene/CHANGES.txt
```diff
@@ -69,6 +69,12 @@ Improvements
 
 * GITHUB#15124: Use RamUsageEstimator to calculate size for non-accountable queries. (Sagar Upadhyaya)
 
+* GITHUB#14515: IndexWriter.forceMergeDeletes() now returns MergePolicy.MergeObserver,
```
vigyasharma: This change can go into 10.4. Let's move this entry to the "API Changes" section under 10.4?

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMerging.java
```diff
@@ -333,6 +337,330 @@ public synchronized void merge(MergeSource mergeSource, MergeTrigger trigger)
     public void close() {}
   }
```
vigyasharma, line 339: Appreciate the thorough tests below, nice work!

vigyasharma (commented):
I like this approach, it let's us observe merge progress without exposing mutable objects. Thanks for making these changes @salvatorecampagna

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
salvatorecampagna: You're right on both points. My original intent was to provide detailed observability, but as you noted, the caller can't meaningfully use index-based access without already having the `MergeSpec`. And `OneMerge` is definitely mutable, so exposing it breaks encapsulation.

The `umMerges()/numCompletedMerges()` approach you suggested is much better, as it provides progress tracking without exposing internals.

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
salvatorecampagna: Excellent suggestion! This is a far better way to provide progress tracking without exposing mutable `OneMerge` objects. I'll add `numCompletedMerges()` that iterates through the merges and checks `mergeCompleted.isDone()`.

Users can then track progress as: `100 * numCompletedMerges() / numMerges()`

~~I'll also update `toString()` to show the completion count for better debugging.~~

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
salvatorecampagna: You're right, the naming is confusing. "New" is ambiguous: does it mean not yetstarted? Does it include in-progress or completed merges?

`numMerges() > 0` is much clearer and eliminates the ambiguity.

I'll remove `hasNewMerges()` and update all test call sites to use `numMerges() > 0` instead.
One less method to maintain and better clarity.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMerging.java
salvatorecampagna: Correct. Calling `await()` without timeout then checking elapsed time separately doesn't make sense. Much cleaner to put the timeout directly in await():
```
assertTrue(observer.await(30_000, MILLISECONDS))
```

This mimics real usage where users will call `await()` with a reasonable timeout. The `30_000ms` is just a test value ensuring `await()` doesn't hang indefinitely. Since we're exposing `await()` APIs, testing they actually work seems reasonable to me.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMerging.java
salvatorecampagna: Sure! Since `forceMergeDeletes(true)` blocks until completion, all merges should be done when it returns. I'll add an assetion. That would verify the blocking behavior actually works.

lucene/core/src/test/org/apache/lucene/index/TestIndexWriterMerging.java
salvatorecampagna: Thank you! I'm a fan of good tests, they're as important as the code itself, documenting behavior and preventing regressions.

This is my first Lucene contribution, so I really appreciate your detailed feedback on the API design. The suggestions made the API much cleaner and safer.

lucene/CHANGES.txt
salvatorecampagna: I see the PR is targeting main (milestone 11.0.0). According to `CONTRIBUTING.md`

```
You should open a pull request against the main branch. Committers will backport it to the maintenance branches once the change is merged into main.
```

Should I keep the `CHANGES.txt` entry in `11.0.0` for now, and then it gets moved to `10.4` in a backport PR if needed? Or should I change it to `10.4` now and update the milestone as well?

vigyasharma: We add the changes entry in the Lucene version we are targeting. All changes go to `main` branch regardless, but the ones for a specific target branch (like 10.4 in this case) additionally get backported to 10.x.

So you should just add the entry to 10.4 in this PR itself and remove it from 11.0. After I merge this into main, i'll backport the commit to 10.x branch. Will reach out to you if I need any help with backport.

salvatorecampagna: Quick question for future contributions: How do I know upfront which version a change is targeting?

vigyasharma: Generally default to the next minor version. If it's a breaking change or a major change in existing behavior, we defer it for a major version release.

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
salvatorecampagna: Later I decided not to include completed merges in `toString` for performance implications with a large number of merges.

lucene/core/src/java/org/apache/lucene/index/MergePolicy.java
```diff
@@ -939,4 +940,93 @@ static final class MergeReader {
...
+          ? CompletableFuture.completedFuture(null)
+          : spec.getMergeCompletedFutures();
+    }
+
+    @Override
+    public String toString() {
+      if (spec == null) {
+        return "MergeObserver: no merges";
+      }
+      return String.format(Locale.ROOT, "MergeObserver: %d merges", numMerges());
```
vigyasharma: We could include `spec.toString()` here. It has some useful info about the segments in each merge.

salvatorecampagna: According to a test I executed locally the outout should be something like. Are we sure we want this level of detail? I guess the output might be long if there are meny merges.

```
MergeObserver: 3 merges
MergeSpec:
  1: _0(11.0.0):c5 _1(11.0.0):c5 into _k
  2: _2(11.0.0):c5 _3(11.0.0):c5 into _l
  3: _4(11.0.0):c5 _5(11.0.0):c5 into _m
```

salvatorecampagna: My concern is that `spec.toString()` was internal before, so we controlled where it's used. Now that `MergeObserver` is public API, users might call `toString()` frequently (logging, monitoring) without realizing it could be expensive with many merges.

vigyasharma (approved):
Looks great, thanks @salvatorecampagna . And congratulations on your first contribution!

PR #15380, opened 2025-10-29
Utilize iterator and not approximation in MaxScoreBulkScorer

benwtrent:
When we utilize the underlying approximation here, its possible that its provided by a TwoPhase iterator. Meaning, we never actually call `matches()` after advancing the approximation and thus the two can get out of sync. This can result in EOFs as described in the related issue.

The other way I could see to fix this is to force calling `matches()` on every item in the essential queue later in the while loop. However, since this was the ONLY place in all of `MaxScoreBulkScorer` that used the approximation for the scorers, I opted to just not use it and use an iterator. 

I welcome discussion, this took a long time to figure out as this part of the code is very interdependent with various other parts of the code :/ and have had many changes over the last couple of minor releases.

Still working on a test...maybe we should add assertions to the TwoPhase iterator that ensures `matches` is called...I am not sure.

closes: https://github.com/apache/lucene/issues/15324

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/DisjunctionScorer.java
+++ b/lucene/core/src/java/org/apache/lucene/search/DisjunctionScorer.java
@@ -108,7 +108,7 @@ abstract class DisjunctionScorer extends Scorer {
     private boolean assertVerifiedMatchesApproximationDocID() {
       int docID = docID();
       for (DisiWrapper w = verifiedMatches; w != null; w = w.next) {
-        if (w.approximation.docID() != docID) {
+        if (w.doc != docID) {
           return false;
         }
       }
--- a/lucene/core/src/java/org/apache/lucene/search/DisjunctionScorer.java
+++ b/lucene/core/src/java/org/apache/lucene/search/DisjunctionScorer.java
@@ -101,7 +101,18 @@ abstract class DisjunctionScorer extends Scorer {
         }
       }
       unverifiedMatches.clear();
+      assert assertVerifiedMatchesApproximationDocID();
       return verifiedMatches;
+    }
+
+    private boolean assertVerifiedMatchesApproximationDocID() {
+      int docID = docID();
+      for (DisiWrapper w = verifiedMatches; w != null; w = w.next) {
+        if (w.approximation.docID() != docID) {
+          return false;
+        }
+      }
+      return true;
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


lucene/core/src/java/org/apache/lucene/search/DisjunctionScorer.java
```diff
@@ -101,9 +101,20 @@ DisiWrapper getSubMatches() throws IOException {
...
 
+    private boolean assertVerifiedMatchesApproximationDocID() {
+      int docID = docID();
+      for (DisiWrapper w = verifiedMatches; w != null; w = w.next) {
+        if (w.approximation.docID() != docID) {
+          return false;
+        }
+      }
+      return true;
+    }
```
benwtrent, line 116: This SEEMS like correct behavior right? If we have a verified match, all their approximations should be on the current docID?

Or do I misunderstand what "approximation" means? Maybe we need to assert against `w.doc`? However, even then, I would expect `w.doc == w.approximation` once things are all verified right?

@jpountz has basically rewritten this entire class over its life time LOL. I am not sure if he is available for feedback here. 

Maybe @romseygeek ?

benwtrent:
Adding some additional context. It seems that `verifiedMatches`, since its used in the disjunction scorers need to all be on the same doc. 

See:

```
  @Override
  protected float score(DisiWrapper topList) throws IOException {
    double score = 0;

    for (DisiWrapper w = topList; w != null; w = w.next) {
      score += w.scorable.score();
    }
    return (float) score;
  }
```

That just assumes the topList (which might be coming from verifiedMatches), are all on the same doc, and that doc matches the assumed iteration position.

We only saw this desynchronizing as an issue when we saw an EOF. But it honestly seems like we are just scoring things incorrectly, fairly often.

Here are some screen shots (sorry for the quality, but wanted to provide some tangible proof of what we are seeing).

Here is one where we are obviously scoring a different document as the verifiedMatches point to a different doc than the iterator. (expected: 13283, verifiedMatch: 15554)

image


Here is one that trips an EOF as the verifiedMatches is iterated to NO_MORE_DOCs, but the iterator assumes its on a valid document (expected: 138666).

image

jpountz (approved):
Thanks for digging this one! Indeed this scorer cannot work with two-phase iterators... It's a pity that tests didn't catch it, I wonder if we could randomly wrap queries with `RandomApproximationQuery` in the test framework to catch such problems.

PR #15382, opened 2025-10-29
Batch virtual calls from DefaultBulkScorer#scoreRange

jainankitk:
### Description

Aims at reducing the overhead of virtual calls from bulk scorer by loading matching docIds into a buffer before flushing the doc buffer into the collector. Related issue #15381

### DIFF
--- a/lucene/core/src/java/org/apache/lucene/search/LeafCollector.java
+++ b/lucene/core/src/java/org/apache/lucene/search/LeafCollector.java
@@ -145,7 +145,6 @@ public interface LeafCollector {
     for (int i = 0; i < count; ++i) {
       collect(docs[i]);
     }
-    ;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


lucene/core/src/java/org/apache/lucene/search/LeafCollector.java
```diff
@@ -123,6 +123,31 @@ default void collect(DocIdStream stream) throws IOException {
...
+   * <p>It is legal for callers to mix calls to {@link #collect(int[], int)}, {@link
+   * #collect(DocIdStream)} and {@link #collect(int)}.
+   *
+   * <p>The default implementation calls {@code for(int i = 0; i < count; ++i) { collect(docs[i]);
+   * }; }.
+   */
+  default void collect(int[] docs, int count) throws IOException {
+    for (int i = 0; i < count; ++i) {
+      collect(docs[i]);
+    }
```
jainankitk, line 147: I was hoping that `LeafCollector` implementations override this method only if they don't need access to the score, but I guess it won't work as the underlying iterator already moved onto the next document

